PR #7, opened 2015-06-15
Bugfix division by zero of the relative convergence measure. Instead divide by some small value

davidsblom:


uekerman:
thanks, david! i think the solution is a bit hacky, though. ok for the moment, but maybe, there is a nicer solution? @scheufks

PR #12, opened 2015-06-29
Bugfix thin plate spline. The log function is used instead of log10

davidsblom:
See http://mathworld.wolfram.com/ThinPlateSpline.html for the definition of the thin plate spline.


uekerman:
right indeed, thanks!

PR #29, opened 2015-08-24
Show the build status of the develop branch in the readme

davidsblom:
Makes it a bit easier to check whether the current commit is working correctly.


floli:
I have integrated that manually, so we have it in just one commit. Thanks for the idea!

PR #61, opened 2016-10-18
Added SU2 adapter and documentation.

arusch:
The adapter for the CFD-Code "Stanford University Unstructured (SU2)" is added along with documentation on how to build the adapter with SU2. It allows for fluid-structure interaction simulations including multiple fluid-solid-interfaces (wet surfaces).


tools/adapters/su2/precice.hpp
```diff
@@ -0,0 +1,200 @@
+/*!
+* \file precice.hpp
+* \brief Header of adapter class for coupling SU2 with preCICE for FSI.
+* \author Alexander Rusch
+*/
+
+
+
+#include "/Users/alexanderrusch/GitHub/precice/src/precice/SolverInterface.hpp"
```
floli: This include is unlikely to work anywhere but on your system. Please use an include relative to the INCLUDE_PATH, which is precice/src

PR #67, opened 2017-04-06
C interface: don't use "extern C" when not in C++.

MakisH:
The header should be able to be included in a C code, compiled with a C compiler. The "extern C" is a C++ statement and triggers an error when trying to compile with a C compiler, while the rest of the code is C. An example is the current CalculiX-adapter.


floli:
Merged. Thanks!

PR #87, opened 2017-12-06
Replace tarch/irr by new XML parser (libxml2)

Pikotee:


uekerman:
Looks good to me. I didn't check every line, but it compiles and the Alya Flap test ran through.Thanks for the contribution! Please don't forget to add the new dependency in the [wiki](https://github.com/precice/precice/wiki/Building).

PR #101, opened 2018-01-30
Add Gitter chat button in README.

MakisH:
Since we have the very nice "building status" button, why not also a Gitter button? I found it in the [tmuxinator](https://github.com/tmuxinator/tmuxinator) repository.


floli:
Please understand that this request won't be merged  immediately as we haven't agreed yet if we proceed with gitter.

PR #107, opened 2018-02-27
compileAndTest: Check for Python 3 and throw a proper error

MakisH:
At the moment, when a user tries to run the compileAndTest script with Python 2, a wild error occures, as the `FileNotFoundError` exception is only defined in Python 3.

I don't know if there are more incompatibilities with Python 2, but if we want to keep it Python 3-only, then this check will throw a human-friendly error if needed.

### DIFF
--- a/tools/compileAndTest.py
+++ b/tools/compileAndTest.py
@@ -7,6 +7,11 @@ To ensure a clean test it can delete and recreate the ./tests directory.
 """
 import argparse, os, shutil, subprocess, sys
 
+# Enforce Python 3 to avoid unexpected errors.
+if sys.version_info[0] != 3:
+    print("This script requires Python version 3.")
+    sys.exit(1)
+
 def run_test(cmd, keep_test):
     if not keep_test:
         try:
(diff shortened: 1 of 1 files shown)


floli:
I am very much against explicit version checks. The first line of the script clearly states, that it is intended for python3.

MakisH:
Then I see the following options:

1. We require at least version 3, or
2. We check if the exception is defined (I don't know if there is anything else incompatible)

floli:
Yes, we require at least version 3 for that scripts, that's what the first line of scripts states.
These scripts are not related to the version the precice binary links against, python2 and 3 work side by side.

MakisH:
But I think this does not work when Python 3 is not available. It will still try to run it with Python 2, without throwing any error. This is what happens to me sometimes e.g. on the MAC Cluster, if I don't load the correct module.

For sure when running the script explicitly with Python 2, it does not complain for the version, it just crashes.

MakisH:
Hm... maybe this is not the case anymore then, because now I only get 

```
/usr/bin/env: python3: No such file or directory
```

which is still ok.

PR #108, opened 2018-02-27
compileAndTest: abort if args.root is not defined.

MakisH:
This is a common problem among new users.
`args.root` defaults to `PRECICE_ROOT`.
If `PRECICE_ROOT` is not defined, the `os.chdir(args.root)` crashes wildly.
This commit checks if it is other than `None` before changing the directory.

### DIFF
--- a/tools/compileAndTest.py
+++ b/tools/compileAndTest.py
@@ -50,7 +50,11 @@ if len(sys.argv) < 2:
     
 args, remainder = parser.parse_known_args()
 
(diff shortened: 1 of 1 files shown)


floli:
Thanks for the suggestions, I have implemented a bit differently using a try... except
```
Easier to ask for forgiveness than permission.
```

MakisH:
Problem solved eitherways! :-)

Issue #129, opened 2018-07-04
cmake missing install stage


olesenm:
As we discussed yesterday, shifting to cmake will certainly help with building. The currently experimental cmake support looks like it does most of the basics, but it only has a build stage and no installation stage.

If we can fix that, spack integration should be no problem. Here is the corresponding spack issue number 
https://github.com/spack/spack/issues/8643 and current state https://github.com/olesenm/spack/tree/precice

floli:
We will be looking into it, but need to get up to speed with regard to cmake installation stuff. 

We are also open to pull requests! :-)

olesenm:
Sure, here's the first bit, but no idea of which top-level includes need to be installed. I figure that someone directly involved with the project will be quicker there.

olesenm:
Also have an `of-precice` package on the same tree: https://github.com/olesenm/spack/tree/precice
Ready to go as soon as the cmake changes are merged.

olesenm:
Closing out with the anticipated merge

PR #131, opened 2018-07-05
CONFIG: add simple cmake install stage (issues #129)

olesenm:
- just the executable and library at the moment

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,14 @@
+#------------------------------------------------------------------------------
 cmake_minimum_required (VERSION 3.5)
+
+# TODO
+# Fail if not using an out-of-source build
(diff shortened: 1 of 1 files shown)


olesenm:
With installation of include files. If this can be merged and tested, it could be used to tag a new release or minor version which the spack packages could use. Or leave as a develop version within spack, which would also be OK.

floli:
I have modified your install rules a bit.
If there no objections, I will merge tomorrow or after the weekend.

Thanks a lot for your contribution!

olesenm:
+1

will you issue a new github release too? Then I can finish up the spack bits and get a merge request do there too.

floli:
We will probably release beginning of August. Our release procedure is not very agile yet ;-)

olesenm:
Ok, then we'll just leave spack as develop version only, which isn't so terrible either. Just need a merge into develop for that.

olesenm:
The travis problem doesn't look like anything from me.

- BTW: both spack packages are now merged in.

PR #138, opened 2018-08-07
MPI Single Ports

floli:
* Change the API of Communication classes. The calling convention is that `acceptConnectionAsServer` and `requestConnectionAsClient` are collective operations now
* Add a class `MPISinglePortsCommunication` and corresponding addition to configuration file options. That class creates just one inter-communicator on the global communicator.
* Unify testing inside com, extend test coverage.


floli:
@fsimonis Thanks for your suggestions, I have taken them into account and did a number of changes to my code.

Though I have added a number of consts, please understand that I haven't followed you everywhere. Adding const to variables which have a purely local scope might not be very necessary.

src/com/MPISinglePortsCommunication.hpp
```diff
@@ -0,0 +1,69 @@
...
+  virtual MPI_Comm &communicator(int rank) override;
+
+  virtual int rank(int rank) override;
+
+  logging::Logger _log{"com::MPISinglePortsCommunication"};
+
+  std::string _addressDirectory;
+
+  /// Remote rank -> communicator map
+  std::map<int, MPI_Comm> _communicators;
```
uekerman: Why do we still need a map here? I thought, we would end up with one communicator, but apparently I misunderstood sth

src/com/MPIDirectCommunication.hpp
```diff
@@ -36,24 +36,28 @@ class MPIDirectCommunication : public MPICommunication
...
+                                int                acceptorRank) override;
 
-  virtual void acceptConnectionAsServer(std::string const &nameAcceptor,
-                                        std::string const &nameRequester,
+  virtual void acceptConnectionAsServer(std::string const &acceptorName,
+                                        std::string const &requesterName,
+                                        int                acceptorRank,
                                         int                requesterCommunicatorSize) override
   {
     ERROR("Not implemented!");
```
uekerman: This should be assertions instead

src/m2n/config/M2NConfiguration.cpp
```diff
@@ -67,12 +67,25 @@ M2NConfiguration::M2NConfiguration(xml::XMLTag &parent)
...
+    doc += "in the same directory.";
+    attrExchangeDirectory.setDocumentation(doc);
+    attrExchangeDirectory.setDefaultValue("");
+    tag.addAttribute(attrExchangeDirectory);
     tags.push_back(tag);
   }
-  
+
   {
-    XMLTag tag(*this, VALUE_MPI_SINGLE, occ, TAG);
```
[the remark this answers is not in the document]
floli: It's like that (values instead of variables) at a couple of other configurations. IMHO it's easier to read and easier to write and extend. If changes are necessary, it's usually only at two places. Furthermore if values are changes, the variable names (which basically are similiar to the values) become inconsistent, which does much more harm to comprehensibility than inconsistency among different configurations of different packages.

src/com/MPIDirectCommunication.hpp
floli: Ok, changed. Also changed one more `ERROR("Not implemented!")`. The one in `ScaleByAreaAction` I left this way, because it's code that can be triggered by the user in a wrong way. Changed in https://github.com/precice/precice/commit/9b394a671f3c4ab48f39be8ab705dd34a038b6bc

src/com/MPISinglePortsCommunication.hpp
floli: `acceptConnection` / `requestConnection` still uses one comm, created on `COMM_SELF`.

``acceptConnectionAsServer` / `requestConnectionAsClient` just uses `communicator[0]`, so it actually creates just one comm.

The reason of that is that the first variant is called in various different ways, mostly just one the accepting/receiving rank. In order to just use communicator, accept/request must be called on the entire global communicator. This would require quite some rework of the calling code. The gains would also be minuscule, because accept/request is mainly (only?) used for 1:1 connection anyways.

The latter, in contrast, is only called from `PointToPointCommunication`, due to that P2P was also heavily modified. But this connection method is the most important, because it does the m2n heavy lifiting.

If we agree, that `acceptConnection` / `requestConnection` just does 1:1 connection, we can rewrite and simplifiy the code.

src/m2n/PointToPointCommunication.cpp
```diff
@@ -410,48 +410,37 @@ void PointToPointCommunication::acceptConnection(std::string const &nameAcceptor
...
 
     // NOTE:
     // Everything is moved (efficiency)!
-    // On the acceptor participant side, the communication object `c'
-    // behaves as a server, i.e. it implicitly accepts multiple connections
-    // to requester processes (in the requester participant). As a result,
-    // only one communication object `c' is needed to satisfy
-    // `communicationMap', and, therefore, for data structure consistency
-    // of `_mappings' with the requester participant side, we simply
-    // duplicate references to the same communication object `c'.
```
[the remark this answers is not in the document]
floli: Yes, put it back in https://github.com/precice/precice/commit/9b394a671f3c4ab48f39be8ab705dd34a038b6bc

src/com/MPISinglePortsCommunication.hpp
uekerman: Currently, `accept/requestConnection` still does 1:N communications for the old server approach. I guess however that these occurrences could be replaced by `acceptConnectionAsServer` / `requestConnectionAsClient`. Anyway, could you simply please add a small comment in line 57 above to explain what you just wrote here.

src/m2n/config/M2NConfiguration.cpp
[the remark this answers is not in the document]
uekerman: OK. I did not know that it is already like that in other places. I agree with the easier-to-read.

src/com/Communication.hpp
```diff
@@ -229,6 +237,10 @@ class Communication
                               int     size,
                               int     rankSender) = 0;
 
+  /// Asynchronously receives a vector of double values.
```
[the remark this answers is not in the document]
floli: No, we use `/// This is a brief comment` for one line comments.

src/com/MPISinglePortsCommunication.cpp
```diff
@@ -0,0 +1,207 @@
...
+  MPI_Open_port(MPI_INFO_NULL, const_cast<char *>(_portName.data()));
+
+  const std::string addressFileName("." + requesterName + "-" + acceptorName + ".address");
+  Publisher::ScopedChangePrefixDirectory scpd(_addressDirectory);
+  ScopedPublisher                        p(addressFileName);
+  p.write(_portName);
+  DEBUG("Accept connection at " << _portName);
+
+  // Connect the first peer, s.t. we can exchange some information about the other side
+  MPI_Comm communicator;
```
fsimonis: Please DRY your code a bit using anonymous namespaces.
This part of the code is essential a function from `(portName, acceptorRank) -> (requesterRank, requsterCommunicatorSize)`
```cpp
      MPI_Comm communicator;
      MPI_Comm_accept(const_cast<char *>(_portName.c_str()), MPI_INFO_NULL, 0, MPI_COMM_SELF, &communicator);
      DEBUG("Accepted connection at " << _portName);

      int    requesterRank      = -1;
      size_t requesterCommunicatorSize = 0;

      MPI_Recv(&requesterRank,             1, MPI_INT, 0, 42, communicator, MPI_STATUS_IGNORE);
      MPI_Recv(&requesterCommunicatorSize, 1, MPI_INT, 0, 42, communicator, MPI_STATUS_IGNORE);
      MPI_Send(&acceptorRank,              1, MPI_INT, 0, 42, communicator);
```

You could return a trivial intermediate such as:
```cpp
struct PeerReply {
    int rank;
    size_t communicatorSize;
};
```
This would make this code super readable and reduce the `MPI_Noise`

src/com/MPISinglePortsCommunication.cpp
```diff
@@ -0,0 +1,207 @@
...
+    MPI_Open_port(MPI_INFO_NULL, const_cast<char *>(_portName.data()));
+    p.write(_portName);
+    DEBUG("Accept connection at " << _portName);
+  }
+  
+  MPI_Comm communicator;
+  MPI_Comm_accept(const_cast<char *>(_portName.c_str()), MPI_INFO_NULL, 0,
+                  utils::Parallel::getGlobalCommunicator(), &communicator);
+  DEBUG("Accepted connection at " << _portName);
+  _communicators[0] = communicator; // all comms are the same
```
fsimonis: _Optional:_ you could assert this in debug builds using:
```cpp
std::all(_communicators.cbegin(), _communicator.cend(), [communicator](const auto& kv){
return kv.second == communicator;
})
```

src/com/MPISinglePortsCommunication.cpp
```diff
@@ -0,0 +1,207 @@
...
+  Publisher::ScopedChangePrefixDirectory scpd(_addressDirectory);
+  Publisher p(addressFileName);
+  _portName = p.read();
+  DEBUG("Request connection to " << _portName);
+
+  MPI_Comm communicator;
+  MPI_Comm_connect(const_cast<char *>(_portName.c_str()), MPI_INFO_NULL, 0,
+                   utils::Parallel::getGlobalCommunicator(), &communicator);
+  DEBUG("Requested connection to " << _portName);
+  _communicators[0] = communicator; // all comms are the same
```
fsimonis: dito

src/com/MPISinglePortsCommunication.cpp
```diff
@@ -0,0 +1,207 @@
...
+}
+
+void MPISinglePortsCommunication::closeConnection()
+{
+  TRACE(_communicators.size());
+
+  if (not isConnected())
+    return;
+
+  for (auto communicator : _communicators) {
```
fsimonis: Use  `auto&` in range-for

src/com/MPISinglePortsCommunication.hpp
```diff
@@ -0,0 +1,83 @@
...
+{
+namespace com
+{
+/**
+ * @brief Provides connection methods based on MPI ports (part of MPI 2.0).
+ *
+ * The two participants to be connected can be run in two process groups started
+ * up individually, i.e. not within the same process group.
+ *
+ * Notes on the implementation:
```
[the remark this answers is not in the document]
floli: Here it does (Doxygen 1.8.14), scroll down to detailed description.

src/com/MPISinglePortsCommunication.cpp
```diff
@@ -0,0 +1,207 @@
...
+
+using precice::utils::Publisher;
+using precice::utils::ScopedPublisher;
+
+namespace precice
+{
+namespace com
+{
+MPISinglePortsCommunication::MPISinglePortsCommunication(std::string const &addressDirectory)
+    : _addressDirectory(addressDirectory)
```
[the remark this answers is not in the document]
floli: `_communicators` is default initialized, there is no use putting it in the c'tors initalizing list.

src/com/MPIPortsCommunication.hpp
```diff
@@ -58,12 +54,14 @@ class MPIPortsCommunication : public MPICommunication
...
 
-  std::vector<MPI_Comm> _communicators;
+  /// Remote rank -> communicator map
+  std::map<int, MPI_Comm> _communicators;
 
   /// Name of the port used for connection.
   std::string _portName = std::string(MPI_MAX_PORT_NAME, '\0');
 
   bool _isAcceptor = false;
+
```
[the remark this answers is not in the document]
floli: Well, would consider this a matter of taste, but ok.

src/com/MPISinglePortsCommunication.cpp
floli: `_communicator.size() == 1` here always. The comment rather says, that all ranks are addressed using the same communicator (when using the `acceptAsServer` and `requestAsClient` methods.

src/com/MPISinglePortsCommunication.cpp
floli: See comment above.

src/com/MPISinglePortsCommunication.cpp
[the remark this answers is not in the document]
fsimonis: This is a clang-tidy warning regarding the following cpp core guideline:
http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c48-prefer-in-class-initializers-to-member-initializers-in-constructors-for-constant-initializers

Looks like this warning is a bit out of place here, as the map is implicitly default initialized.
Lets move on to more important issues.

src/com/MPISinglePortsCommunication.hpp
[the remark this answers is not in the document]
fsimonis: Ran it again after a `git clean -fdx` and now it looks good.

src/com/MPIPortsCommunication.cpp
```diff
@@ -167,66 +137,44 @@ void MPIPortsCommunication::requestConnection(std::string const &nameAcceptor,
...
-  _communicators.push_back(communicator);
   _isConnected = true;
-  _rank        = requesterProcessRank;
 
-  MPI_Send(&requesterProcessRank,      1, MPI_INT, 0, 42, communicator);
+  int acceptorRank = -1;
+  MPI_Send(&requesterRank,             1, MPI_INT, 0, 42, communicator);
   MPI_Send(&requesterCommunicatorSize, 1, MPI_INT, 0, 42, communicator);
+  MPI_Recv(&acceptorRank,              1, MPI_INT, 0, 42, communicator, MPI_STATUS_IGNORE);
+  _communicators[0] = communicator; // should be acceptorRank
```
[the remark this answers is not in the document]
floli: This is not something which can be asserted. It's about the calling convention of the function. The usage of acceptConnection / requestConnection inside preCICE is a bit inconsistent. This "should" needs to addressed in the long run.

src/com/MPISinglePortsCommunication.cpp
fsimonis: The purpose of `setupCommunicatorSelf` is to setup the communicator. It is strictly dependant on the `_portName` and the result is used to register the communicator to `_communicators`. You also mentioned in your PM that there is no access to the member-logger, which is a problem.

So, lets just turn in into a private function: (Not tested)
```cpp
{
  TRACE(acceptorName, requesterName);
  assertion(not isConnected());

  _isAcceptor = true;

  MPI_Open_port(MPI_INFO_NULL, const_cast<char *>(_portName.data()));

  const std::string addressFileName("." + requesterName + "-" + acceptorName + ".address");
  Publisher::ScopedChangePrefixDirectory scpd(_addressDirectory);
  ScopedPublisher                        p(addressFileName);
  p.write(_portName);
  DEBUG("Accept connection at " << _portName);
  // Setup first peer
  const auto firstCommSize = setupCommunicator(acceptorRank);
  
  // Connect all other peers
  for (size_t i = 1; i < firstCommSize; ++i) {
      const auto commSize = setupCommunicator(acceptorRank);
      assertion(firstCommSize == commSize, "Inconsistent communicator sizes!")
  }
  
  _isConnected = true;
}
```
With:
```cpp
size_t MPISinglePortsCommunication::setupCommunicator(int acceptorRank) {
  MPI_Comm communicator;
  MPI_Comm_accept(const_cast<char *>(_portName.c_str()), MPI_INFO_NULL, 0, MPI_COMM_SELF, &communicator);
  DEBUG("Accepted connection at " << _portName);

  int    requesterRank      = -1;
  size_t requesterCommunicatorSize = 0;

  // Exchange information to which rank I am connected and which communicator size on the other side
  MPI_Recv(&requesterRank,             1, MPI_INT, 0, 42, communicator, MPI_STATUS_IGNORE);
  MPI_Recv(&requesterCommunicatorSize, 1, MPI_INT, 0, 42, communicator, MPI_STATUS_IGNORE);
  MPI_Send(&acceptorRank,              1, MPI_INT, 0, 42, communicator);
  
  CHECK(requesterCommunicatorSize > 0, "Requester communicator size has to be > 0!");
  _communicators[requesterRank] = communicator;
  return requesterCommunicatorSize;
}
``` 
This would also allow logging using the class member.
I am not sure about the name of the function though.

fsimonis: The point I am trying to make is that this code handles MPI orchestration, which is hard to debug as problems don't result in compile time errors and lead to runtime deadlocks.
Code duplication in the core of such a function is in my opinion absolutely unacceptable.
That is the core message of my review.
The code examples I provided are only suggestions of how one might DRY the code.

Regarding your feedback:

> * Now that it is a member and modifies the class' state, it has none of the elegance and clearly defined scope of a function (in mathematical sense)

Logging and MPI will always imply a state on this code. I agree that a clean separation would be favourable, but I don't see a possibility here.

> * It deludes the reader into believing that this way of communicator setup is the one default/standard way done in that class, while in fact it is only used in one of four functions that create communicator.

This is an implementation detail, which is what documentation and private functions are for.
That said, the difficult part would be to pick an appropriate name for said function.

> * The net win in terms of lines of code is minimal and the it increases complexity.

Firstly, I don't think that LOC is the appropriate metric here. Secondly, I think that this significantly decreases complexity.
From my perspective it is easier to reason about a small member function than having to track how variables change throughout a long member function that even involves a loop.
Isolating the comm-setup leads to smaller functions with fewer variables and state changes reduced to member variables only.

fsimonis: As an alternative to extracting the section, a `do while` loop could be used to combine the first iteration with the rest:
```cpp
{
  TRACE(acceptorName, requesterName);
  assertion(not isConnected());

  _isAcceptor = true;

  MPI_Open_port(MPI_INFO_NULL, const_cast<char *>(_portName.data()));

  const std::string                      addressFileName("." + requesterName + "-" + acceptorName + ".address");
  Publisher::ScopedChangePrefixDirectory scpd(_addressDirectory);
  ScopedPublisher                        p(addressFileName);
  p.write(_portName);

  size_t peerCurrent = 0; // The current Peer to connect to
  size_t peerCount   = 0; // The total count of peers (initialized in the first iteration)
  do {
    MPI_Comm communicator;
    MPI_Comm_accept(const_cast<char *>(_portName.c_str()), MPI_INFO_NULL, 0, MPI_COMM_SELF, &communicator);
    DEBUG("Accepted connection at " << _portName << " for Peer " << peerCurrent);

    // Handshake with Peer
    int    requesterRank             = -1;
    size_t requesterCommunicatorSize = 0;

    MPI_Recv(&requesterRank, 1, MPI_INT, 0, 42, communicator, MPI_STATUS_IGNORE);
    MPI_Recv(&requesterCommunicatorSize, 1, MPI_INT, 0, 42, communicator, MPI_STATUS_IGNORE);
    MPI_Send(&acceptorRank, 1, MPI_INT, 0, 42, communicator);

    // Check for consistency
    CHECK(requesterCommunicatorSize > 0, "Requester communicator size has to be > 0!");
    CHECK(requesterCommunicatorSize == _communicators.size(),
          "Requester communicator sizes are inconsistent!");
    CHECK(_communicators.count(requesterRank) == 0,
          "Duplicate request to connect by same rank (" << requesterRank << ")!");

    // Register the communicator for this peer connection
    _communicators[requesterRank] = communicator;

    // Initialize the count of peers to connect to
    if (peerCurrent == 0) {
        peerCount = requesterCommunicatorSize;
        DEBUG("Connections per Peer: " << peerCount);
    }
  } while (++peerCurrent < peerCount);

  _isConnected = true;
}
```

floli: I like that way much more! Adapted it a bit
```cpp
void MPISinglePortsCommunication::acceptConnection(std::string const &acceptorName,
                                                   std::string const &requesterName,
                                                   int                acceptorRank)
{
  TRACE(acceptorName, requesterName);
  assertion(not isConnected());

  _isAcceptor = true;

  MPI_Open_port(MPI_INFO_NULL, const_cast<char *>(_portName.data()));

  const std::string addressFileName("." + requesterName + "-" + acceptorName + ".address");
  Publisher::ScopedChangePrefixDirectory scpd(_addressDirectory);
  ScopedPublisher                        p(addressFileName);
  p.write(_portName);

  size_t peerCurrent = 0; // current peer to connect to
  size_t requesterCommunicatorSize = 0;

  do {
    // Connection
    MPI_Comm communicator;
    MPI_Comm_accept(const_cast<char *>(_portName.c_str()), MPI_INFO_NULL, 0, MPI_COMM_SELF, &communicator);
    DEBUG("Accepted connection at " << _portName << " for peer " << peerCurrent);

    int requesterRank = -1;
    // Exchange information to which rank I am connected and which communicator size on the other side
    MPI_Recv(&requesterRank,             1, MPI_INT, 0, 42, communicator, MPI_STATUS_IGNORE);
    MPI_Recv(&requesterCommunicatorSize, 1, MPI_INT, 0, 42, communicator, MPI_STATUS_IGNORE);
    MPI_Send(&acceptorRank,              1, MPI_INT, 0, 42, communicator);

    CHECK(requesterCommunicatorSize > 0,
          "Requester communicator size has to be > 0!");
    CHECK(_communicators.count(requesterRank) == 0,
          "Duplicate request to connect by same rank (" << requesterRank << ")!");
    
    _communicators[requesterRank] = communicator;

  } while (++peerCurrent < requesterCommunicatorSize);
  
  _isConnected = true;
}
```
I took the same scheme for MPIPortsCommunication.

Issue #117, opened 2018-04-11
preCICE version/build information in the output


MakisH:
I think it would be very convenient to have information about the version or the commit number of preCICE in the output. This is particularly important when a user sends log files asking for help.

I will do some research on this and try to implement it. Any ideas are welcome.

floli:
You mean as a kind of startup message at the very top of a log file?

My first shoot at this would be that scons figures out the commit including the latest tag (e.g. from git describe), use that to ```-DPRECICE_VERSION``` which is to be put into the log file.

MakisH:
Yes, thank you for the feedback!

My first approach would also be that, but thenI would need to make this work also without the repository (i.e. for releases). An approach there could be to include this information in an additional file in every release.

floli:
`git describe --long --tags | sed 's/\([^-]*-g\)/r\1/;s/-/./g'` gives a nice string including the version and commit id.

PR #140, opened 2018-08-08
Put version information in separate header

floli:
Related to #117

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,13 +58,11 @@ if (PETSC AND MPI)
   string(REGEX MATCHALL "[0-9]+" PETSC_VERSIONS ${PETSC_VERSION})
   list(GET PETSC_VERSIONS 0 PETSC_MAJOR)
   list(GET PETSC_VERSIONS 1 PETSC_MINOR)
-  add_definitions(-DPETSC_MAJOR=${PETSC_MAJOR} -DPETSC_MINOR=${PETSC_MINOR})
   include_directories(${PETSC_INCLUDES})
 else()
   add_definitions(-DPRECICE_NO_PETSC)
   message(STATUS "PETSc support disabled")
 endif()
-
 
 find_package (Threads REQUIRED)
 
@@ -98,6 +96,10 @@ else()
   message(STATUS "Python support disabled")
 endif()
 
+
+# Create versions.hpp and add binary dir to include dirs
+configure_file("${PROJECT_SOURCE_DIR}/src/versions.hpp.in" "${PROJECT_BINARY_DIR}/versions.hpp")
+include_directories(${PROJECT_BINARY_DIR})
 
 add_subdirectory("src")
 include_directories("src")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,17 +1,8 @@
-#------------------------------------------------------------------------------
 cmake_minimum_required (VERSION 3.5)
-
-# TODO
-# Fail if not using an out-of-source build
-# if (CMAKE_CURRENT_SOURCE_DIR STREQUAL CMAKE_CURRENT_BINARY_DIR)
-#     message(FATAL_ERROR
-#         "In-source builds are not supported.  Please create a build directory "
-#         "separate from the source directory")
-# endif()
 
 project(preCICE LANGUAGES CXX)
 
-list (APPEND CMAKE_MODULE_PATH "${preCICE_SOURCE_DIR}/tools/cmake-modules")
+list (APPEND CMAKE_MODULE_PATH "${PROJECT_SOURCE_DIR}/tools/cmake-modules")
 
 set (CMAKE_CXX_STANDARD 11)
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -30,6 +30,14 @@ endif()
 
 option(BUILD_SHARED_LIBS "Build a shared library" ON)
 
+# Get preCICE version from git (that assumes you are compiling on a git checkout)
+execute_process(
+  COMMAND git describe --tags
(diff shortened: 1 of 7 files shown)


floli:
Can't be merged yet, because it fails with scons.

CMakeLists.txt
```diff
@@ -39,6 +30,14 @@ endif()
 
 option(BUILD_SHARED_LIBS "Build a shared library" ON)
 
+# Get preCICE version from git (that assumes you are compiling on a git checkout)
```
fsimonis: Either assert this assumption or handle the edge case.
Distributing archived source code without `.git` is a common example that does not satisfy this assumption.

CMakeLists.txt
```diff
@@ -99,6 +96,10 @@ else()
 endif()
 
 
+# Create versions.hpp and add binary dir to include dirs
+configure_file("${PROJECT_SOURCE_DIR}/src/versions.hpp.in" "${PROJECT_BINARY_DIR}/versions.hpp")
```
fsimonis: Please restrict substitution using `@ONLY` to enforce consistency in the future.

https://cmake.org/cmake/help/latest/command/configure_file.html

floli:
Just using a plain version number, e.g. 1.2.0 is fine for released version, however, if someone uses a non-released version, it's nicer to also have the commit id.

MakisH:
For the xSDK compliance we need to have a version number for the released versions and a commit id for the development branch. I agree that we should not assume git always.

I still haven't found time to review this, but I will!

fsimonis:
@MakisH does the xSDK specify a version number rule for modified non-git repos?

If that is not the case, we could:

1) try to FindGit
2) try to get the commit (the source might not be a git repo)
3) try to detect a modified build (no tag?)
4) __then__ set the version to the commit hash

If __any__ of these steps fail, we could simply assume that the project is a unmodified version.

We would violate the [cmake version format](https://cmake.org/cmake/help/latest/command/project.html) though:
> If VERSION is specified, given components must be non-negative integers. If VERSION is not specified, the default version is the empty string.

MakisH:
Here is the exact description from xSDK:

> M8.​​ ​Each​ ​package​ ​must​ ​provide​ ​a​ ​runtime​ ​API​,​ ​for​ ​example​ ​a​ ​function​ ​call,​ ​to​ ​return​ ​the​ ​current version​ ​number​​ ​of​ ​the​ ​software.​ ​ ​For​ ​development​ ​versions​ ​of​ ​the​ ​software,​ ​each​ ​package​ ​must provide​ ​the​ ​current​ ​commit​ ​ID​ ​in​ ​the​ ​repository​ ​and​ ​indicate​ ​what​ ​compiler​ ​options​ ​were​ ​used​ ​to​ ​build the​ ​package.​ ​With​ ​this​ ​information​ ​users​ ​should​ ​be​ ​able​ ​to​ ​rebuild​ ​the​ ​package​ ​in​ ​the​ ​same​ ​state.​ ​We do​ ​not​ ​currently​ ​require​ ​that​ ​version​ ​information​ ​for​ ​all​ ​dependent​ ​packages​ ​be​ ​provided,​ ​so​ ​it​ ​may​ ​not be​ ​possible​ ​to​ ​rebuild​ ​the​ ​entire​ ​software​ ​stack​ ​in​ ​the​ ​same​ ​state.

The important message here is to have a way to refer to and reproduce specific states of the code.

fsimonis:
Ok, so lets fiddle this apart:

> Each​ ​package​ ​must​ ​provide​ ​a​ ​runtime​ ​API​,​ ​for​ ​example​ ​a​ ​function​ ​call,​ ​to​ ​return​ ​the​ ​current version​ ​number​​ ​of​ ​the​ ​software.​ 

So, we need to provide runtime access to this information.
As we are using C++11, we could do the following:
```cpp
namespace precice { namespace version {
    static constexpr char * version = "@PRECICE_VERSION@";
    static constexpr uint8_t version_major = @PRECICE_VERSION_MAJOR@;
    static constexpr uint8_t version_minor = @PRECICE_VERSION_MINOR@;
    static constexpr uint8_t version_patch = @PRECICE_VERSION_PATCH@;
}}
```

> ​For​ ​development​ ​versions​ ​of​ ​the​ ​software,​ ​each​ ​package​ ​must provide​ ​the​ ​current​ ​commit​ ​ID​ ​in​ ​the​ ​repository

This reads like it simply needs to be available somewhere.
We could generate a simple file `commit.txt` via cmake only containing the commit hash and install it along with the package.
Alternatively, we could try to incorporate this into the _hopefully future_ `preciceConfig.cmake`.

>​ ​and​ ​indicate​ ​what​ ​compiler​ ​options​ ​were​ ​used​ ​to​ ​build the​ ​package.​ ​With​ ​this​ ​information​ ​users​ ​should​ ​be​ ​able​ ​to​ ​rebuild​ ​the​ ​package​ ​in​ ​the​ ​same​ ​state.​ ​We do​ ​not​ ​currently​ ​require​ ​that​ ​version​ ​information​ ​for​ ​all​ ​dependent​ ​packages​ ​be​ ​provided,​ ​so​ ​it​ ​may​ ​not be​ ​possible​ ​to​ ​rebuild​ ​the​ ​entire​ ​software​ ​stack​ ​in​ ​the​ ​same​ ​state.

The overkill version of this is already available in CMake by telling it to [export the compile commands](https://cmake.org/cmake/help/latest/variable/CMAKE_EXPORT_COMPILE_COMMANDS.html).
The resulting JSON contains the compile command for each translation units, so we could put that in the package too.

floli:
Addendum: `constexpr` are nicer than `#define` but you can't use them for conditional compilation before C++17.

PR #141, opened 2018-08-09
Fix m2n indices left in undefined state after move

fsimonis:
This `std::move()` moves the `std::vector<int>` from the map key and thus leaves it in an undefined state.
This change swaps an empty vector in-place, leaving the map in a altered, but defined state.

### DIFF
--- a/src/m2n/PointToPointCommunication.cpp
+++ b/src/m2n/PointToPointCommunication.cpp
@@ -428,7 +428,8 @@ void PointToPointCommunication::acceptConnection(std::string const &nameAcceptor
 
     c->receive(globalRequesterRank, localRequesterRank);
 
-    auto indices = std::move(communicationMap[globalRequesterRank]);
+    std::vector<int> indices;
+    indices.swap(communicationMap[globalRequesterRank]);
     _totalIndexCount += indices.size();
 
     // NOTE:


floli:
I see the problem. Wouldn't it be more straightforward to simply `erase` the key after it has been moved?

fsimonis:
I had a deeper look at the code and it turns out, that this map is built inside the function and remains untouched until the end of scope. So this move is not a problem. However, I think erasing the undefined value would be a good idea.
We could do the following:

1) Find the key (prevents accidental creation of kv-pair)
2) Assert its existence
3) Move the indices into the new element
4) erase the iter

i.e.:
```cpp
for (size_t localRequesterRank = 0; localRequesterRank < communicationMap.size(); ++localRequesterRank) {
    int globalRequesterRank = -1;

    c->receive(globalRequesterRank, localRequesterRank);

    auto iter = communicationMap.find(globalRequesterRank);
    CHECK(iter != communicationMap.end());
    _totalIndexCount += inter->second.size();
    _mappings.push_back({
        static_cast<int>(localRequesterRank), globalRequesterRank, std::move(iter->second), c, com::PtrRequest(), 0});
    communicationMap.erase(iter);
  }
```

fsimonis:
I am closing this PR in favour of a proper fix.

Issue #144, opened 2018-08-15
Make building of PySolverInterface safe and clean


BenjaminRueth:
https://github.com/precice/precice/commit/f5fa4ed01f30696df37d41523bac1e56daceb6bb makes sure that `setup.py` only deletes `precice/src/precice/bindings/python/build`. Before, it was possible to accidentally delete any folder names `build`, if `setup.py` is not run from `precice/src/precice/bindings/python`. This is especially annoying, if `setup.py` is run from `precice` just after compiling preCICE.

With https://github.com/precice/precice/commit/f5fa4ed01f30696df37d41523bac1e56daceb6bb, `setup.py` can be run from arbitrary folders and only `precice/src/precice/bindings/python/build` is deleted - if it exists. 

However, the new `build` and `PySolverInterface.so` are created in the folder, where `setup.py` is run. `PySolverInterface.cpp` is created in `precice/src/precice/bindings/python`. **This is messy.** 

What do we actually want to have the binaries? I see the following possibilities:

1. do not allow to run `setup.py` outside of `precice/src/precice/bindings/python`.
2. no matter, where `setup.py` is run, put all resulting files into the root folder. (here I do not see any possibility to realize a good cleanup process of temporary files)
3. no matter, where `setup.py` is run, put all resulting files into `precice/src/precice/bindings/python`.

# Progress

- [x] where do we want to have the binaries?
    - [x] temporary in `precice/src/precice/bindings/python`
    - [x] python module at default location (i.e. `site-packages` or `dist-packages`)
- [x] update code influenced by binaries at python default locations
    - [x] elastictube [StructureSolver.py](https://github.com/precice/elastictube1d/blob/master/PythonTube/StructureSolver.py#L10)
    - [x] elastictube [FluidSolver.py](https://github.com/precice/elastictube1d/blob/master/PythonTube/FluidSolver.py#L21)
    - [x] [solverdummy.py](https://github.com/precice/precice/blob/develop/tools/solverdummies/python/solverdummy.py#L13)
- [x] compatibility
    - [x] python 2 (2.7.12)
    - [x] python 3 (3.5.2)
- [x] include building python bindings in preCICE build? yes! --> offloaded issue to https://github.com/precice/precice/issues/182
    - [x] update SCons build process (optional, we want to move towards cmake)
    - [x] update cmake build process
- [x] add relevant changes to [changelog](https://github.com/precice/precice/blob/develop/CHANGELOG.md)

BenjaminRueth:
[...]

Issue #68, opened 2017-05-10
Port python bindings to python3


floli:
Should be rather easy. However, python2 and 3 adaptors should coexist. With some ```from __future__``` calls they may even share the same code.
```
% python3 setup.py build_ext --inplace 
  File "setup.py", line 11
    print "ERROR: PRECICE_ROOT not defined!"
                                           ^
SyntaxError: Missing parentheses in call to 'print'
```

floli:
Needs to be reopened, since we have string decode issues, e.g. with the python solver dummy:
```
Traceback (most recent call last):
  File "solverdummy.py", line 41, in <module>
    interface = PySolverInterface(participantName, solverProcessIndex, solverProcessSize)
  File "PySolverInterface.pyx", line 95, in PySolverInterface.PySolverInterface.__cinit__
    def __cinit__ (self, string solverName, int solverProcessIndex, int solverProcessSize):
  File "stringsource", line 15, in string.from_py.__pyx_convert_string_from_py_std__in_string
TypeError: expected bytes, str found
```

BenjaminRueth:
I got the same error using python 3.5.2.

BenjaminRueth:
Here a few lines of code for reproducing the error from command line:
```
import PySolverInterface  # if PySolverInterface is in PYTHONPATH
PySolverInterface.PySolverInterface("dummy", 1, 0)
```

BenjaminRueth:
With the new commit the decoding issues are gone. However, we now use implicit duck typing instead of explicit typing for strings. To my knowledge there is not a baseclass of byte and string that we could use instead (in python 3)?

BenjaminRueth:
Just tested running `precice/tools/solverdummies/python` with 5b88d7fe7498d72171bce4c7a7c463e0a76764d7. Works with `python3` as well. As soon as #146 is merged into `develop`, we can close this issue, too.

BenjaminRueth:
#146 closed. We can also close this one.

PR #146, opened 2018-08-20
Improving PySolverInterface build

BenjaminRueth:
This resolves the following mentioned in #144. Additionally a smaller issue #68 has been resolved and tested with the dummy solver.

* where do we want to have the binaries? 
    * temporary in `precice/src/precice/bindings/python`
    * python module at default location (i.e. `site-packages` or `dist-packages`)
* update code influenced by binaries at python default locations 
    * python solverdummy
* compatibility
    * python 2 (2.7.12)
    * python 3 (3.5.2)

Issues on moving the build process of the python API into the main build process of preCICE are planned to be addressed in an independent step.


BenjaminRueth:
There seems to be a possibility to add optional arguments to `setup.py`. This might help to allow defining a custom mpi compiler wrapper. See https://stackoverflow.com/questions/677577/distutils-how-to-pass-a-user-defined-parameter-to-setup-py for example.

floli:
Why not give the users the possibility to define the MPI compiler wrapper and use that one? Either via an environment variable, or - better - using an argument to setup.py.

src/precice/bindings/python/setup.py
```diff
@@ -1,62 +1,58 @@
...
 
-# check if PRECICE_ROOT is defined
-if not os.getenv('PRECICE_ROOT'):
-    print("ERROR: PRECICE_ROOT not defined!")
-    exit(1)
-
-precice_root = os.getenv('PRECICE_ROOT')
-python_bindings_path = os.path.join(precice_root,'src','precice','bindings','python')
+python_bindings_path = os.path.dirname(os.path.abspath(__file__))
+precice_root = os.path.join(python_bindings_path,"../../../../../precice")
```
floli: Why not $PRECICE_ROOT here?

It's ok, if you prefer the relative location, not sure if using PRECICE_ROOT is superior here.

BenjaminRueth: I used `$PRECICE_ROOT` before. But this gave me problems, when I had to run `sudo python setup.py install`, if the python is installed in a folder without user access.

BenjaminRueth:
with d8e1923 you can now specify the mpi compiler wrapper through `--mpicompiler=` it should work with `setup.py build/install/build_ext`.

uekerman:
why not `--compiler=`? then it would be consistent with scons, right?

BenjaminRueth:
using `--compiler=` instead of `--mpicompiler=` is no problem. I will push the corresponding commit in a moment

BenjaminRueth:
seems like I promised too much: If I use `--compiler=` instead of `--mpicompiler=`, I get strange error messages. The keyword `compiler` clashes with some other options and `mpicompiler` does not. `install` and `build` do not use the keywork `compiler`, but `build_ext` does. Therefore, I would like to stay with `--mpicompiler`.

floli:
Looks good, with regard to compiling against non-default mpich:
```
% python setup.py build --mpicompiler=/opt/mpich/bin/mpicxx
running build
#####
calling my_build
using --mpicompiler=/opt/mpich/bin/mpicxx
adding extension
#####
running build_ext
#####
calling my_build_ext
using --mpicompiler=mpic++
#####
Compiling /home/lindnefn/precice/src/precice/bindings/python/PySolverInterface.pyx because it changed.
[1/1] Cythonizing /home/lindnefn/precice/src/precice/bindings/python/PySolverInterface.pyx
building 'PySolverInterface' extension
creating build
creating build/temp.linux-x86_64-3.7
creating build/temp.linux-x86_64-3.7/home
creating build/temp.linux-x86_64-3.7/home/lindnefn
creating build/temp.linux-x86_64-3.7/home/lindnefn/precice
creating build/temp.linux-x86_64-3.7/home/lindnefn/precice/src
creating build/temp.linux-x86_64-3.7/home/lindnefn/precice/src/precice
creating build/temp.linux-x86_64-3.7/home/lindnefn/precice/src/precice/bindings
creating build/temp.linux-x86_64-3.7/home/lindnefn/precice/src/precice/bindings/python
gcc -pthread -Wno-unused-result -Wsign-compare -DNDEBUG -g -fwrapv -O3 -Wall -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -flto -fuse-linker-plugin -ffat-lto-objects -flto-partition=none -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -march=x86-64 -mtune=generic -O2 -pipe -fstack-protector-strong -fno-plt -fPIC -I/home/lindnefn/precice/src/precice/bindings/python/../../../../../precice -I/usr/include/python3.7m -c /home/lindnefn/precice/src/precice/bindings/python/PySolverInterface.cpp -o build/temp.linux-x86_64-3.7/home/lindnefn/precice/src/precice/bindings/python/PySolverInterface.o -I/home/lindnefn/precice/src/precice/bindings/python/../../../../../precice -Wall -std=c++11 -g -I/opt/mpich/include -L/opt/mpich/lib -lmpicxx -Wl,-rpath -Wl,/opt/mpich/lib -Wl,--enable-new-dtags -lmpi
creating build/lib.linux-x86_64-3.7
g++ -pthread -shared -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now -flto -fuse-linker-plugin -ffat-lto-objects -flto-partition=none -Wl,-O1,--sort-common,--as-needed,-z,relro,-z,now build/temp.linux-x86_64-3.7/home/lindnefn/precice/src/precice/bindings/python/PySolverInterface.o -L/usr/lib -lpython3.7m -o build/lib.linux-x86_64-3.7/PySolverInterface.cpython-37m-x86_64-linux-gnu.so -L/home/lindnefn/precice/src/precice/bindings/python/../../../../../precice/build/last/ -lprecice -g -I/opt/mpich/include -L/opt/mpich/lib -lmpicxx -Wl,-rpath -Wl,/opt/mpich/lib -Wl,--enable-new-dtags -lmpi
```

```
% ldd build/lib.linux-x86_64-3.7/PySolverInterface.cpython-37m-x86_64-linux-gnu.so | grep mpi
        libmpicxx.so.12 => /opt/mpich/lib/libmpicxx.so.12 (0x00007f99d503c000)
        libmpi.so.12 => /opt/mpich/lib/libmpi.so.12 (0x00007f99d4b8f000)
        libmpifort.so.12 => /opt/mpich/lib/libmpifort.so.12 (0x00007f99ce801000)
```

PR #152, opened 2018-08-29
Adding network socket option for macOS on reference XML

jy1737:
macOS uses different network socket option to Ubuntu. How about we add a comment about this on reference XML (like how SuperMUC socket option is there now)?


jy1737:
I think I've squashed the commits. Can you check please?

MakisH:
Looks good!

Issue #142, opened 2018-08-10
Parallelize the existing initialization of preCICE by a two-level approach


uekerman:
This issue should be an overview on the complete story and link to respective smaller work packages.

This is a draft and should be extended and formatted. 

The following steps need to happen during initialization (in this order):

### 1. [Communicate Bounding Boxes and Generate Connection Information](https://github.com/precice/precice/pull/169)
* package: `partition`
* new function `communicateBoundingBox`: 
  * gather and send bounding boxes at `ProvidedBoundingBox` (finally will be merged to `ProvidedPartition`) 
  * receive set of bounding boxes at `ReceiveBoundingBox` (finally will be merged to `ProvidedBoundingBox`) 
* new function `computeBoundingBox`: 
  * generate a feedback for each rank which includes connected ranks on the orther side
  * gather these feedbacks at Master of `ReceivedPartition` and send it to the other master.
  * Master of `ProvidedPartition` bradcasts the set of feedbacks to its slaves  
  * each rank creates a `CommunicationMap` (member variable of `Mesh`), a map _rank |-> set of vertices. At this level list of vertices has only one dummy member (-1). This list will be udated later.

- [x] Implemented
- [x] Unit tests
- [x] Reviewed 
- [ ] Merged to `develop`

### 2. [Create Communications](https://github.com/precice/precice/pull/180)
* package: `m2n`
* new functions `acceptPreCommunication`, `requestPreCommunication` 
* we get initialConnectionMap from Mesh 
* generate `mappings` by calling accept/requestPreConnection -> without vertex list

- [x] Implemented
- [x] Unit tests
- [ ] Reviewed 
- [ ] Merged to `develop`

### 3. [Broadcast Local Mesh Partitions](https://github.com/precice/precice/pull/181)   
* package: `partition` , `m2n` 
* partition: new function `communicate`  
* each rank on `ProvidedBoundingBox` broadcasts mesh partition to all connected ranks on `ReceivedboundingBox`
* `M2N` needs new methods `accept/requestSlavesPreConnection` to call `accept/requestPreConnection` in `PonitToPointCommunication` 
* two new method in `PonitToPointCommunication` class: broadcastSend/ReceiveMesh 
* `M2N` needs new method `broadcastSendLocalMesh` and `broadcastReceiveLocalMesh`  that in turn calls a new method `broadcastSendMesh` and `broadcastReceiveMesh`  in the respective `PointToPointCommunication`
[...]

PR #169, opened 2018-10-17
WIP issue 142, package 1 :

atotoun:
communicate bb and generate initial communciation map


MakisH:
I got a bit confused and we may need this term written here if in the future we search for pull requests: "bb" stands for "boundary box".

atotoun:
> I got a bit confused and we may need this term written here if in the future we search for pull requests: "bb" stands for "boundary box".

I will explain in the issue, also its "bounding box" ;-)

MakisH:
Aaaand you are perfectly right, that's what I wanted to write! I write "boundary" more often than I write "bounding". :-/

floli:
Please hold back merging this, until MPI Single Ports is merged.

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+    // initializing the _globalBB
+    for (int remoteRank = 0; remoteRank < _remoteParComSize; remoteRank++ )
+    {
+      _globalBB[remoteRank]= _bb;
+    }
+    
+    // receive _globalBB from master
+    com::CommunicateBoundingBox(utils::MasterSlave::_communication).broadcastReceiveBoundingBoxMap(_globalBB);
+    _numberOfRemoteRanks = _globalBB.size();
```
uekerman: Isn't `_remoteParComSize` the same as `_numberOfRemoteRanks` ?

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -0,0 +1,154 @@
...
+        if (compareBoundingBox(_bb,other_rank.second)) {
+          _feedback.push_back(other_rank.first);            
+        }
+      }
+    } 
+
+    //send feedback to master
+    utils::MasterSlave::_communication->send(_feedback, 0);     
+  }
+  else if (utils::MasterSlave::_masterMode)
```
uekerman: It's a bit inconsistent with the rest to now do the master after the slaves.

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -0,0 +1,154 @@
...
+    // sending this participant communicator size to other master
+    _m2n->getMasterCommunication()->send(utils::MasterSlave::_size , 0);
+    utils::MasterSlave::_communication->broadcast(_remoteParComSize);
+    com::CommunicateBoundingBox(utils::MasterSlave::_communication).broadcastSendBoundingBoxMap(_globalBB);
+    _numberOfRemoteRanks = _globalBB.size();
+
+    // initializing _feedbackmap
+    for (int rank_slave=1; rank_slave < utils::MasterSlave::_size ; rank_slave++)
+    {
+      _feedbackMap[rank_slave].push_back(-1);
```
uekerman: Is this necessary?

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -0,0 +1,154 @@
...
+      utils::MasterSlave::_communication->receive(_feedback, rank_slave);
+      _feedbackMap[rank_slave]=_feedback;        
+    }
+
+    // master sends feedbackmap to other master
+    com::CommunicateBoundingBox(_m2n->getMasterCommunication()).sendFeedbackMap(_feedbackMap,0);             
+  }
+
+
+  if (utils::MasterSlave::_masterMode)
```
uekerman: Why do we need this block?

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,117 @@
...
+ *
+ * A mesh is received by the master rank and re-partitioned among all slave ranks.
+ * Afterwards necessary distribution data structures are set up.
+ */
+class ReceivedBoundingBox : public Partition
+{
+public:
+
+  /// Defines the typ of geometric filter used
+  enum GeometricFilter {
```
uekerman: Why is this needed here?

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,117 @@
...
+  friend struct PartitionTests::ReceivedBoundingBoxTests::TestReceivedBoundingBox3D;
+  friend struct PartitionTests::ProvidedBoundingBoxTests::TestProvidedBoundingBox2D;
+  friend struct PartitionTests::ProvidedBoundingBoxTests::TestProvidedBoundingBox3D;
+  friend struct PartitionTests::ProvidedBoundingBoxTests::TestM2NMeshExchange;
+
+  
+private:
+
+  
+  // bool compareBoundingBox(mesh::Mesh::BoundingBox currentBB, mesh::Mesh::BoundingBox receivedBB);
```
uekerman: Where is this method implemented?

atotoun: added, sorry forgot to include in commit!

src/partition/ReceivedBoundingBox.hpp
atotoun: We can use Broadcast_Filter in the bb method, filter_first does not work, I kept it so we can choose to filter after broadcast or not. filter_first can be deleted!

src/partition/ReceivedBoundingBox.cpp
atotoun: we need _vertexCounters for final filtering of vertices. Basically, each member of this vector shows the corresponding rank's maximum global vertex id. We filter the vertices first and afterward to decide where each vertice belong, we compare id with the entries of _vertexCounters.

src/partition/ReceivedBoundingBox.cpp
atotoun: it did not work without initializing! maybe bad implementation from my side, I will check!

src/partition/ReceivedBoundingBox.cpp
atotoun: order changed ;)

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -0,0 +1,154 @@
...
+  }
+
+  if (utils::MasterSlave::_slaveMode)
+  {
+    utils::MasterSlave::_communication->broadcast(_remoteParComSize, 0);
+
+    // initializing the _globalBB
+    for (int remoteRank = 0; remoteRank < _remoteParComSize; remoteRank++ )
+    {
+      _globalBB[remoteRank]= _bb;
```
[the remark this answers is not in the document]
atotoun: corrected!

src/partition/ProvidedBoundingBox.cpp
```diff
@@ -0,0 +1,187 @@
...
+    com::CommunicateBoundingBox(utils::MasterSlave::_communication).broadcastReceiveFeedbackMap(_receivedFeedbackMap);
+
+    for (auto &otherRank : _receivedFeedbackMap)
+    {
+      for (auto &includedRanks: otherRank.second)
+      {
+        if (utils::Parallel::getProcessRank() == includedRanks)
+        {
+          _connectedRanks.push_back(otherRank.first);
+          _mesh->getCommunicationMap()[otherRank.first].push_back(-1);
```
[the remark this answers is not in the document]
atotoun: then we will need to have two data structures and  we will need to call accept/request connection twice! Doing this way, in the second step we only update the list of vertices!
We can discuss this in detail!

src/partition/ProvidedBoundingBox.cpp
```diff
@@ -0,0 +1,187 @@
...
+      for (auto &includedRanks: otherRank.second) {
+        if (utils::Parallel::getProcessRank() == includedRanks) {
+          _connectedRanks.push_back(otherRank.first);
+          _mesh->getCommunicationMap()[otherRank.first].push_back(-1);
+        }
+      }
+    }
+    
+    //add master vertices to vertex counter which is a vector that shows each ranks maximum vertex id
+    for(int i=0; i<numberOfVertices; i++){
```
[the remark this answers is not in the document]
atotoun: Yes, this stuff can be safely moved to the next step! I will do that! 
But we do need that ;)

src/partition/ProvidedBoundingBox.cpp
```diff
@@ -0,0 +1,187 @@
...
+    }
+   
+    _mesh->setGlobalNumberOfVertices(vertexCounter);
+
+    for (int j = 0; j < utils::MasterSlave::_size; j++)
+    {      
+    utils::MasterSlave::_communication->broadcast(_vertexCounters[j]);
+    }
+  }
+  else
```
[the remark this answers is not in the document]
atotoun: yes, but I have added //master and //slave before each branch of if to make it easier. not enough?

src/partition/ProvidedBoundingBox.hpp
```diff
@@ -0,0 +1,84 @@
...
+  
+  bool _hasToSend;
+   
+  int _dimensions;
+  
+  std::vector<int> _connectedRanks;
+  
+  double _safetyFactor;
+  
+  std::vector<int> _vertexCounters;
```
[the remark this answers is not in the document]
atotoun: I have added some comments in the code! 
Also here copying :
'we need _vertexCounters for final filtering of vertices. Basically, each member of this vector shows the corresponding rank's maximum global vertex id. We filter the vertices first and afterward to decide where each vertice belong, we compare id with the entries of _vertexCounters.'

src/partition/ProvidedBoundingBox.cpp
```diff
@@ -0,0 +1,187 @@
...
+      std::vector<int> initialFeedback;
+      initialFeedback.push_back(-1);
+      _receivedFeedbackMap[i]=initialFeedback;
+    }
+    com::CommunicateBoundingBox(_m2n->getMasterCommunication()).receiveFeedbackMap(_receivedFeedbackMap, 0 ); 
+    com::CommunicateBoundingBox(utils::MasterSlave::_communication).broadcastSendFeedbackMap(_receivedFeedbackMap);     
+
+    // master checks which ranks are connected to it
+    for (auto &otherRank : _receivedFeedbackMap) {
+      for (auto &includedRanks: otherRank.second) {
```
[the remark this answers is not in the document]
atotoun: :+1:

src/partition/ProvidedBoundingBox.cpp
```diff
@@ -0,0 +1,187 @@
...
+        if (utils::Parallel::getProcessRank() == includedRanks) {
```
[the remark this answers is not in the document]
atotoun: But it works! will we deactivate this ?

src/partition/ReceivedBoundingBox.cpp
atotoun: yes, I removed  `_numberOfRemoteRanks`.

src/partition/ProvidedBoundingBox.cpp
```diff
@@ -0,0 +1,187 @@
...
+}
+
+void ProvidedBoundingBox::computeBoundingBox()
+{
+  int numberOfVertices = _mesh->vertices().size();
+  if (not utils::MasterSlave::_slaveMode) {//Master
+    assertion(utils::MasterSlave::_size>1);
+    int vertexCounter = 0;
+    //master receives other partition communicator size and also a feedback which is a map:  list of other participant ranks -> connected ranks at this participant  
+    _m2n->getMasterCommunication()->receive(_remoteParComSize, 0);
```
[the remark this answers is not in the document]
atotoun: no need, corrected!

atotoun:
> Please hold back merging this, until MPI Single Ports is merged.

sure!

src/partition/ProvidedBoundingBox.cpp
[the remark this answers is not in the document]
uekerman: At the moment, it's simply not the correct way of doing this. When doing a unit test with more than one participant, this should give you the wrong rank (at least on the second participant).

src/com/CommunicateBoundingBox.hpp
```diff
@@ -32,12 +32,26 @@ class CommunicateBoundingBox
...
+
+  void receiveFeedbackMap(
+      mesh::Mesh::FeedbackMap &fbm,
+      int                         rankSender);
+
   void broadcastSendBoundingBoxMap(
       mesh::Mesh::BoundingBoxMap &bbm);
 
   void broadcastReceiveBoundingBoxMap(
       mesh::Mesh::BoundingBoxMap &bbm);
```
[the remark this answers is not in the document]
atotoun: added!

src/mesh/Mesh.hpp
```diff
@@ -49,10 +49,14 @@ class Mesh : public PropertyContainer, private boost::noncopyable
   typedef utils::ptr_vector<PropertyContainer>   PropertyContainerContainer;
   typedef std::vector<std::pair<double, double>> BoundingBox;
   typedef std::map<int,BoundingBox>              BoundingBoxMap;
```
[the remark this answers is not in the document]
atotoun: added!

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include "com/CommunicateMesh.hpp"
+#include "com/Communication.hpp"
+#include "utils/MasterSlave.hpp"
+#include "m2n/M2N.hpp"
+#include "utils/EventTimings.hpp"
+#include "mapping/Mapping.hpp"
+#include "mesh/Mesh.hpp"
+#include "mesh/Vertex.hpp"
+#include "mesh/Edge.hpp"
+#include "mesh/Triangle.hpp"
```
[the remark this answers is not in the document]
atotoun: No. only mesh/Mesh.hpp is needed.

src/partition/ProvidedBoundingBox.hpp
```diff
@@ -0,0 +1,84 @@
...
+}
+
+
+namespace precice {
+namespace partition {
+
+
+/**
+ * @brief this class is supposed to:
+ * 1- creat bounding boxes around each ranks mesh partition
```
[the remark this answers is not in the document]
atotoun: thanks !

src/partition/ProvidedBoundingBox.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+   /// Constructor
+  ProvidedBoundingBox(mesh::PtrMesh mesh, bool hasToSend, double safetyFactor);
+
+  virtual ~ProvidedBoundingBox() {}
+
+   /// The boundingbox is gathered and sent to another participant (if required)
+  virtual void communicate();
+  virtual void compute();
+  // virtual void communicateBoundingBox();
```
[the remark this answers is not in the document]
atotoun: yes, its a typo. I need these methods for the third package! but  `communicateBoundingBox()` and `computeBoundingBox()` we need now!

src/partition/ProvidedBoundingBox.cpp
```diff
@@ -0,0 +1,187 @@
...
+    _dimensions(mesh->getDimensions()),
+    _safetyFactor(safetyFactor)
+{}
+
+void ProvidedBoundingBox::communicateBoundingBox()
+{
+
+  if (_hasToSend) { 
+  
+    Event e1("creat and gather bounding box");
```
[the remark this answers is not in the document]
atotoun: I am not sure where to put timers, we should discuss it!

src/partition/ProvidedBoundingBox.cpp
```diff
@@ -0,0 +1,187 @@
...
+          com::CommunicateBoundingBox(utils::MasterSlave::_communication).receiveBoundingBox(_bb, rankSlave);       
+          DEBUG("From slave " << rankSlave << ", bounding mesh: " << _bb[0].first
+                << ", " << _bb[0].second << " and " << _bb[1].first << ", " << _bb[1].second);        
+          _globalBB[rankSlave] = _bb;
+        }
+      }
+    }    
+    e1.stop();
+
+    //master sends global bb to the other master
```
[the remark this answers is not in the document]
atotoun: I will try to unify comments!

src/partition/ProvidedBoundingBox.hpp
```diff
@@ -0,0 +1,84 @@
...
+  
+  double _safetyFactor;
+  
+  std::vector<int> _vertexCounters;
+  
+  int _remoteParComSize = 0 ;
+
+  mesh::Mesh::BoundingBox _bb;
+  
+  mesh::Mesh::BoundingBoxMap _globalBB;
```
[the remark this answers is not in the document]
atotoun: yes!

src/partition/ProvidedBoundingBox.cpp
```diff
@@ -0,0 +1,187 @@
...
+logging::Logger ProvidedBoundingBox:: _log ( "precice::partition::ProvidedBoundingBox" );
+
+ProvidedBoundingBox::ProvidedBoundingBox
+(mesh::PtrMesh mesh,
+ bool hasToSend,
+ double safetyFactor)
+:
+    Partition (mesh),
+    _hasToSend(hasToSend),
+    _bb(mesh->getBoundingBox()),
```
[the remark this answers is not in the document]
atotoun: _bb? removed!

src/partition/ProvidedBoundingBox.hpp
```diff
@@ -0,0 +1,84 @@
...
+
+  mesh::Mesh::FeedbackMap _receivedFeedbackMap;
```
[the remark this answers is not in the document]
atotoun: true! corrected!

src/partition/ProvidedBoundingBox.cpp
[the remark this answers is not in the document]
uekerman: I would still prefer to use two data structures here.

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -131,24 +99,83 @@ void ReceivedBoundingBox::computeBoundingBox()
...
+
+    //send feedback to master
+    utils::MasterSlave::_communication->send(_feedback, 0);     
   }  
 }
 
+bool ReceivedBoundingBox::compareBoundingBox(mesh::Mesh::BoundingBox currentBB, mesh::Mesh::BoundingBox receivedBB)
+{
+  //int sizeofBB = currentBB.size();
+  bool intersect=1;
```
uekerman: Please use `true` and `false` instead

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -131,24 +99,83 @@ void ReceivedBoundingBox::computeBoundingBox()
...
+
+  for (int i=0; i < _dimensions; i++) {
+
+    if ((currentBB[i].first < receivedBB[i].first && currentBB[i].second < receivedBB[i].first) || (receivedBB[i].first < currentBB[i].first && receivedBB[i].second < currentBB[i].first) ) {
```
uekerman: Please add a comment for this line explaining what you do

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -131,24 +99,83 @@ void ReceivedBoundingBox::computeBoundingBox()
...
+
+      intersect = 0;
+      i=_dimensions;
```
uekerman: Please use `break` instead

src/partition/ReceivedBoundingBox.hpp
uekerman: But we don't filter any vertices here, right? Is this already something we need in a later WP?

src/partition/ProvidedBoundingBox.cpp
```diff
@@ -0,0 +1,120 @@
...
+}
+
+void ProvidedBoundingBox::computeBoundingBox()
+{
+  int numberOfVertices = _mesh->vertices().size();
+  int remoteParComSize = -1;
+  mesh::Mesh::FeedbackMap receivedFeedbackMap;
+  if (not utils::MasterSlave::_slaveMode) {//Master
+    assertion(utils::MasterSlave::_size>1);
+    int vertexCounter = 0;
```
uekerman: This variable is unused.

src/partition/ProvidedBoundingBox.hpp
```diff
@@ -0,0 +1,49 @@
...
+    
+  bool _hasToSend;
+   
+  int _dimensions;
+  
+  std::vector<int> _connectedRanks;
+  
+  double _safetyFactor;
+  
+  std::vector<int> _vertexCounters;
```
uekerman: needed?

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+  int _dimensions;
+
+  double _safetyFactor;
+
+  int _numberOfVertices = 0;  
+
+  // each element shows the max id of vertices,
+  // e.g. vertexcounters[5]= 25 shows the last global id of rank 6 vertices is 25. 
+  std::vector<int> _vertexCounters;
```
uekerman: needed?

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,93 @@
...
+
+  double _safetyFactor;
+
+  int _numberOfVertices = 0;  
+
+  // each element shows the max id of vertices,
+  // e.g. vertexcounters[5]= 25 shows the last global id of rank 6 vertices is 25. 
+  std::vector<int> _vertexCounters; 
+
+  GeometricFilter _geometricFilter;
```
uekerman: needed?

src/com/CommunicateBoundingBox.hpp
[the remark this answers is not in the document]
atotoun: yes, right. edited! sorry for that!

src/partition/ReceivedBoundingBox.hpp
atotoun: yes, but we need that for next WP. As its an argument of the constructor, I keep it!

src/partition/ReceivedBoundingBox.cpp
atotoun: added!

src/partition/ReceivedBoundingBox.hpp
atotoun: yes, its an argument of the constructor!

src/partition/ReceivedBoundingBox.cpp
atotoun: done!

src/partition/ProvidedBoundingBox.cpp
atotoun: removed for now!

src/partition/ProvidedBoundingBox.hpp
atotoun: removed for now!

src/partition/ReceivedBoundingBox.hpp
atotoun: removed for now!

src/partition/ProvidedBoundingBox.cpp
```diff
@@ -116,4 +125,15 @@ void ProvidedBoundingBox::computeBoundingBox()
   }
 }
 
+// these functions will be implemented in package 3
```
uekerman: Then, in principle, they should not be included here :)

src/mesh/Mesh.hpp
```diff
@@ -47,12 +47,21 @@ class Mesh : public PropertyContainer, private boost::noncopyable
...
   typedef std::vector<PtrData>                   DataContainer;
   typedef utils::ptr_vector<PropertyContainer>   PropertyContainerContainer;
+  /// a vector of pairs that represents the physical range of vertices owned by a rank. 
   typedef std::vector<std::pair<double, double>> BoundingBox;
+  /// set of boundingboxes owned by each rank gathered in the master rank.
+  /// A map : each rank -> boundingbox around it's mesh partition.
   typedef std::map<int,BoundingBox>              BoundingBoxMap;
+  /// a map from rank -> a vector of ranks : shows each rank is connected
+  /// to which ranks of the other participant. 
+  typedef std::map<int, std::vector<int>>        FeedbackMap;
```
uekerman: Is it really necessary to have this data structure as member of the mesh? Couldn't it simply be a local variable during the feedback process. We should only attach data structures to the mesh if they have to be carried from one package to another, here, normally, from partition to m2n.

src/mesh/Mesh.hpp
```diff
@@ -338,6 +359,18 @@ class Mesh : public PropertyContainer, private boost::noncopyable
    */
   int _globalNumberOfVertices = -1;
 
+  /**
+   * @brief each rank stores list of connected ranks. 
+   * Later in M2N package, this will be used to create initial communication channels. 
+   */
+  std::vector<int> _initialCommunicationMap;
```
uekerman: Same as above. I guess this is not needed.

src/mesh/Mesh.hpp
```diff
@@ -258,6 +267,18 @@ class Mesh : public PropertyContainer, private boost::noncopyable
     _globalNumberOfVertices = num;
   }
 
+  /// Returns a vector of connected ranks
+  std::vector<int> & getInitialCommunicationMap()
```
uekerman: Probably not needed

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -118,10 +119,17 @@ void ReceivedBoundingBox::computeBoundingBox()
       if (compareBoundingBox(_bb,other_rank.second)) {
         _feedback.push_back(other_rank.first);            
       }
-    }   
+    }
+
+    // to prevent sending empty vector!
+    if(_feedback.size() == 0)
```
uekerman: I think it's better to simply have an if-block around the send below and sending the size before. It's minimally less efficient, but using dummy data is dangerous.

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -92,8 +93,8 @@ void ReceivedBoundingBox::computeBoundingBox()
     // master receives feedbacks from slaves and put them into the feedbackmap
     for (int rank_slave=1; rank_slave < utils::MasterSlave::_size ; rank_slave++)
     {
-      utils::MasterSlave::_communication->receive(_feedback, rank_slave);
-      _feedbackMap[rank_slave]=_feedback;        
+      utils::MasterSlave::_communication->receive(_feedback, rank_slave);         
+      _feedbackMap[rank_slave]=_feedback;
```
uekerman: Why are `_feedback` and `_feedbackMap` member variables?

src/partition/ProvidedBoundingBox.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+   /// Constructor
+  ProvidedBoundingBox(mesh::PtrMesh mesh, bool hasToSend, double safetyFactor);
+
+  virtual ~ProvidedBoundingBox() {}
+
+  // These functions will be iplemented in 3rd package
+  virtual void communicate();
+  virtual void compute();
+  virtual void createOwnerInformation();
```
uekerman: All three methods are not yet needed

src/partition/ProvidedBoundingBox.hpp
```diff
@@ -0,0 +1,52 @@
...
+  
+  /// The boundingbox is gathered and sent to another participant (if required)
+  virtual void communicateBoundingBox();
+
+  /// The feedback from other participants received here and initial communication map is build 
+  virtual void computeBoundingBox();
```
uekerman: In https://github.com/precice/precice/issues/142, this method is called `compareBB`

src/partition/ProvidedBoundingBox.hpp
```diff
@@ -0,0 +1,52 @@
...
+
+  logging::Logger _log{"partition::ProvidedBoundingBox"};
+  
+//  virtual void createOwnerInformation();
+    
+  bool _hasToSend;
+   
+  int _dimensions;
+  
+  std::vector<int> _connectedRanks;
```
uekerman: Is this variable needed? What does it do?

src/partition/ReceivedBoundingBox.cpp
uekerman: Any update?

src/partition/ReceivedBoundingBox.hpp
uekerman: I know this is pedantic, but the PR will be much easier to read if it is really restricted to the first WP only. Also the constructor can be modified ;)

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,97 @@
...
+  /// bounding box set is received from other participant
+  virtual void communicateBoundingBox();
+
+  /* bounding boxes are compared and feedback sent to other
+     master to construct initial communication map
+  */
+  virtual void computeBoundingBox();
+  
+  /// The mesh is received from another participant.
+  virtual void communicate ();
```
uekerman: Not needed yet

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,97 @@
...
+  
+  /// The mesh is compared and all distribution data structures are set up.
+  virtual void compute ();
```
uekerman: Not needed yet

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,97 @@
...
+  ReceivedBoundingBox (mesh::PtrMesh mesh, double safetyFactor, GeometricFilter geometricFilter);
+  virtual ~ReceivedBoundingBox() {}
+
+  /// bounding box set is received from other participant
+  virtual void communicateBoundingBox();
+
+  /* bounding boxes are compared and feedback sent to other
+     master to construct initial communication map
+  */
+  virtual void computeBoundingBox();
```
uekerman: In https://github.com/precice/precice/issues/142 this is called `compareBB`

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,97 @@
...
+  
+  bool compareBoundingBox(mesh::Mesh::BoundingBox currentBB, mesh::Mesh::BoundingBox receivedBB);
+
+  // void filterMesh(mesh::Mesh& filteredMesh, const bool filterByBB);
+
+  void prepareBoundingBox();
+
+  // bool isVertexInBB(const mesh::Vertex& vertex);
+
+  virtual void createOwnerInformation();
```
uekerman: Not needed yet

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+  // number of other particpant ranks
+  int _remoteParComSize = 0;
+  
+  mesh::Mesh::BoundingBoxMap _globalBB;
+  
+  mesh::Mesh::BoundingBox _bb;
+
+  // feedback from each rank which contains list of connected ranks of the other solver
+  std::vector<int> _feedback;
```
uekerman: Probably not a member variable

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,97 @@
...
+  
+  mesh::Mesh::BoundingBoxMap _globalBB;
+  
+  mesh::Mesh::BoundingBox _bb;
+
+  // feedback from each rank which contains list of connected ranks of the other solver
+  std::vector<int> _feedback; 
+
+  // int : each rank, vect: connected ranks
+  mesh::Mesh::FeedbackMap  _feedbackMap;
```
uekerman: Probably not a member variable

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,97 @@
...
+  mesh::Mesh::BoundingBox _bb;
+
+  // feedback from each rank which contains list of connected ranks of the other solver
+  std::vector<int> _feedback; 
+
+  // int : each rank, vect: connected ranks
+  mesh::Mesh::FeedbackMap  _feedbackMap; 
+
+  // list of connected ranks -> list of vertices
+  mesh::Mesh::FeedbackMap _localCommunicationMap;
```
uekerman: Probably not a member variable

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,97 @@
...
+  mesh::Mesh::FeedbackMap  _feedbackMap; 
+
+  // list of connected ranks -> list of vertices
+  mesh::Mesh::FeedbackMap _localCommunicationMap; 
+
+  int _dimensions;
+
+  double _safetyFactor;
+
+  int _numberOfVertices = 0;
```
uekerman: Probably not needed

src/partition/ReceivedBoundingBox.hpp
uekerman: That's no reason :)

src/partition/ProvidedBoundingBox.cpp
atotoun: we need them to avoid having an abstract class, so we can instantiate from it for testing!

src/partition/ProvidedBoundingBox.cpp
```diff
@@ -0,0 +1,139 @@
...
+    assertion(utils::MasterSlave::_size>1);
+
+    // master receives other partition communicator size
+    _m2n->getMasterCommunication()->receive(remoteParComSize, 0);
+    utils::MasterSlave::_communication->broadcast(remoteParComSize);
+    for (int i=0; i < remoteParComSize; i++)
+    {
+      std::vector<int> initialFeedback;
+      initialFeedback.push_back(-1);
+      receivedFeedbackMap[i]=initialFeedback;
```
fsimonis: You can directly construct a vector here. See [constructor](https://en.cppreference.com/w/cpp/container/vector/vector) number 8.
```suggestion
      receivedFeedbackMap[i] = {-1};
```

src/partition/ProvidedBoundingBox.cpp
```diff
@@ -0,0 +1,139 @@
...
+    // master receives feedback map (list of other participant ranks -> connected ranks at this participant)
+    // from other participants master
+    com::CommunicateBoundingBox(_m2n->getMasterCommunication()).receiveFeedbackMap(receivedFeedbackMap, 0 ); 
+    com::CommunicateBoundingBox(utils::MasterSlave::_communication).broadcastSendFeedbackMap(receivedFeedbackMap);     
+
+    // master checks which ranks are connected to it
+    for (auto &otherRank : receivedFeedbackMap) {
+      for (auto &includedRank: otherRank.second) {
+        if (utils::MasterSlave::_rank == includedRank) {
+          _connectedRanks.push_back(otherRank.first);
```
fsimonis: You add to, but never reset `_connectedRanks`.

Please either reset it, use a `[unordered_]set`, or guarantee (aka assert) that `computeBoundingBox` gets never called twice.

src/partition/ProvidedBoundingBox.cpp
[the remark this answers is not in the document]
fsimonis: I am fine with it if you move the outer variables into the true-false-blocks.

src/partition/ProvidedBoundingBox.cpp
fsimonis: This is nitpicking, but instead of the comment above, you could have added doxygen todos
`/// @todo Implement in next Package`

src/partition/ProvidedBoundingBox.hpp
fsimonis: Should this be a `vector` or an [`unordered_set`](https://en.cppreference.com/w/cpp/container/unordered_set)?

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,97 @@
...
+ *
+ * A mesh is received by the master rank and re-partitioned among all slave ranks.
+ * Afterwards necessary distribution data structures are set up.
+ */
+class ReceivedBoundingBox : public Partition
+{
+public:
+
+  /// Defines the typ of geometric filter used
+  enum GeometricFilter {
```
fsimonis: Please use a [scoped enum](https://en.cppreference.com/w/cpp/language/enum).
```suggestion
  enum class GeometricFilter {
```

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,97 @@
...
+    // @brief undefined
+    UNDEFINED,
+    // @brief No geometric filter used (e.g. for RBF mappings)
+    NO_FILTER,
+    // @brief Filter at master and communicate only filtered mesh.
+    FILTER_FIRST,
+    // @brief Broadcast first and filter then
+    BROADCAST_FILTER
```
fsimonis: From a language perspective both options are not very consistent.

Random thoughts to start some ideas:
* When do we filter?  `FILTER_FIRST` and `FILTER_LATER`
* What is the order? `FILTER_BROADCAST` and `BROADCAST_FILTER`
* What do we do first? `BROADCAST_FIRST` and `FILTER_FIRST`
* Where do we filter? `FILTER_ON_MASTER` and `FILTER_ON_SLAVES`

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,97 @@
...
+  /// The mesh is compared and all distribution data structures are set up.
+  virtual void compute (); 
+    
+private:
+
+  logging::Logger _log{"partition::ReceivedBoundingBox"};
+  
+  bool compareBoundingBox(mesh::Mesh::BoundingBox currentBB, mesh::Mesh::BoundingBox receivedBB);
+
+  // void filterMesh(mesh::Mesh& filteredMesh, const bool filterByBB);
```
fsimonis: Please do not leave things commented out.

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+  void prepareBoundingBox();
+
+  // bool isVertexInBB(const mesh::Vertex& vertex);
```
fsimonis: ditto

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+  virtual void createOwnerInformation();
+
+  // /// Helper function for 'createOwnerFunction' to set local owner information
+  // void setOwnerInformation(const std::vector<int> &ownerVec);
```
fsimonis: ditto

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,97 @@
...
+
+  // number of other particpant ranks
```
fsimonis: Doxygen comments start with three slashes
```suggestion
  /// number of other particpant ranks
```

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,97 @@
...
+  int _remoteParComSize = 0;
+  
+  mesh::Mesh::BoundingBoxMap _globalBB;
+  
+  mesh::Mesh::BoundingBox _bb;
+
+  // feedback from each rank which contains list of connected ranks of the other solver
```
fsimonis:
```suggestion
  /// feedback from each rank which contains list of connected ranks of the other solver
```

src/partition/ReceivedBoundingBox.hpp
fsimonis: `vector` or `[unordered]_set`?

src/partition/ReceivedBoundingBox.hpp
```diff
@@ -0,0 +1,97 @@
...
+  std::vector<int> _feedback; 
+
+  // int : each rank, vect: connected ranks
```
fsimonis: Please explain the reason why we need the member, not (only) the types.
```suggestion
  /// int : each rank, vect: connected ranks
```

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -0,0 +1,200 @@
...
+}
+
+void ReceivedBoundingBox::computeBoundingBox()
+{
+  // handle coupling mode first (i.e. serial participant)
+
+
+  if (not utils::MasterSlave::_slaveMode && not utils::MasterSlave::_masterMode)
+  { 
+    //some code here
```
fsimonis: There is no code here.
```suggestion
    /// @todo implement
```

src/partition/ReceivedBoundingBox.cpp
fsimonis: In case this stays, please use:
```suggestion
      _feedbackMap[rank_slave] = {-1};        
```

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -0,0 +1,200 @@
...
+    // master produces its own feedback
+    for (auto &other_rank: _globalBB)
+    {
+      if (compareBoundingBox(_bb,other_rank.second))
+      {
+        _feedback.push_back(other_rank.first);
+      }          
+    }
+
+    // master first adds ist feedback to feedbackmap
```
fsimonis:
```suggestion
    // master first adds its feedback to feedbackmap
```

src/partition/ReceivedBoundingBox.cpp
fsimonis: I agree

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -0,0 +1,200 @@
...
+
+    //send feedback to master
+    utils::MasterSlave::_communication->send(_feedback, 0);
+    
+  }  
+}
+
+bool ReceivedBoundingBox::compareBoundingBox(mesh::Mesh::BoundingBox currentBB, mesh::Mesh::BoundingBox receivedBB)
+{
+  //int sizeofBB = currentBB.size();
```
fsimonis: Please remove

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -0,0 +1,200 @@
...
+      _feedback.push_back(-1);
+    }
+
+    //send feedback to master
+    utils::MasterSlave::_communication->send(_feedback, 0);
+    
+  }  
+}
+
+bool ReceivedBoundingBox::compareBoundingBox(mesh::Mesh::BoundingBox currentBB, mesh::Mesh::BoundingBox receivedBB)
```
fsimonis: Please find a name a more expressive name than compare.
E.g. `bool overlapping()` or `bool intersecting()`

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -0,0 +1,200 @@
...
+  //int sizeofBB = currentBB.size();
+  bool intersect= true;
+
+  for (int i=0; i < _dimensions; i++) {
+
+    /*
+     * here two boundingboxs are compared to check there is an intersection or not! comparison is done for 
+     * all dimensions and in order to have intersection, at all dimensions there should be a common region!
+     */
+    if ((currentBB[i].first < receivedBB[i].first && currentBB[i].second < receivedBB[i].first) || (receivedBB[i].first < currentBB[i].first && receivedBB[i].second < currentBB[i].first) ) {
```
fsimonis: Can you refactor this into another predicate?

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -0,0 +1,200 @@
...
+    
+  }  
+}
+
+bool ReceivedBoundingBox::compareBoundingBox(mesh::Mesh::BoundingBox currentBB, mesh::Mesh::BoundingBox receivedBB)
+{
+  //int sizeofBB = currentBB.size();
+  bool intersect= true;
+
+  for (int i=0; i < _dimensions; i++) {
```
fsimonis: This may be a candidate for [std::all_of](https://en.cppreference.com/w/cpp/algorithm/all_any_none_of).

src/partition/ReceivedBoundingBox.cpp
```diff
@@ -0,0 +1,200 @@
...
+    if ((currentBB[i].first < receivedBB[i].first && currentBB[i].second < receivedBB[i].first) || (receivedBB[i].first < currentBB[i].first && receivedBB[i].second < currentBB[i].first) ) {
+
+      intersect = false;
+      break;
+    }
+  }
+  return intersect;
+}
+
+void ReceivedBoundingBox::prepareBoundingBox(){
```
fsimonis: Please find a more expressive name.

fsimonis: Could you try to implement this using boost geometry?
Have a look at [return_envelope](https://www.boost.org/doc/libs/1_68_0/libs/geometry/doc/html/geometry/reference/algorithms/envelope/return_envelope_1.html).

src/mesh/Mesh.hpp
atotoun: I think we need this one, because we have to pass it to m2n package to create initial channels! But we agreed to rename it to initialConnectionMap!

src/mesh/Mesh.hpp
atotoun: we need this, because the initialConnectionMap is private! But the name should be changed accordingly!

src/partition/ReceivedBoundingBox.cpp
atotoun: ok, I changed it!

src/partition/ReceivedBoundingBox.cpp
atotoun: not needed! changed!

src/partition/ProvidedBoundingBox.hpp
atotoun: #142  updated!

src/partition/ProvidedBoundingBox.hpp
atotoun: not needed for now, removed!

src/partition/ReceivedBoundingBox.hpp
atotoun: at least not for now! deleted!

src/partition/ProvidedBoundingBox.cpp
```diff
@@ -0,0 +1,139 @@
...
+      com::CommunicateBoundingBox(_m2n->getMasterCommunication()).sendBoundingBoxMap(globalBB,0);
+    }                  
+  }   
+}
+
+void ProvidedBoundingBox::computeBoundingBox()
+{
+  int numberOfVertices = _mesh->vertices().size();
+  int remoteParComSize = -1;
+  mesh::Mesh::FeedbackMap receivedFeedbackMap;
```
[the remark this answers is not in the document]
atotoun: I define the parameters which are used by both master and slaves outside of the splitting region to avoid double definition!

src/partition/ProvidedBoundingBox.cpp
atotoun: right! I applied it, thanks!

src/partition/ReceivedBoundingBox.cpp
atotoun: removed!

src/partition/ReceivedBoundingBox.cpp
uekerman: good point, done

src/partition/ReceivedBoundingBox.cpp
uekerman: i put it on two lines, now it should be easier to read

src/partition/ReceivedBoundingBox.cpp
uekerman: I agree that we could use a better name here, probably sth like `mergeBoundingBoxes`, but this is beyond the scope of this PR. Let's postpone this until the refactoring (work package 6)

src/partition/ProvidedBoundingBox.cpp
```diff
@@ -0,0 +1,139 @@
...
+:
+    Partition (mesh),
+    _hasToSend(hasToSend),
+    _dimensions(mesh->getDimensions()),
+    _safetyFactor(safetyFactor)
+{}
+
+void ProvidedBoundingBox::communicateBoundingBox()
+{
+  if (_hasToSend) {
```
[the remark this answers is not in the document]
uekerman: done

src/partition/ReceivedBoundingBox.hpp
uekerman: beyond this PR, on the list for final refactoring work package

src/partition/ReceivedBoundingBox.hpp
uekerman: beyond this PR, on the list for final refactoring work package

uekerman:
At least two things you should still do:
* Write documentation in `ReceivedBoundingBox.hpp`
* Write more tests and also test edge cases

src/com/CommunicateBoundingBox.cpp
```diff
@@ -88,9 +88,9 @@ void CommunicateBoundingBox::receiveConnectionMap(
     int              rank;
     std::vector<int> connected_ranks;
     _communication->receive(rank, rankSender);
-    assertion((int) i == rank);
     _communication->receive(connected_ranks, rankSender);
     fbm[rank] = connected_ranks;
+    connected_ranks.clear();
```
uekerman: This line is not needed, right?

src/partition/ProvidedBoundingBox.cpp
```diff
@@ -0,0 +1,143 @@
...
+      }
+    }
+
+  } else { // Slave
+
+    utils::MasterSlave::_communication->broadcast(remoteConnectionMapSize, 0);
+
+    for (int i = 0; i < remoteConnectionMapSize; i++) {
+      remoteConnectionMap[i] = {-1};
+    }
```
uekerman: This is probably still wrong, right? Also here, we need to loop of the specific ranks. If true, we should implement a unit test that catches this problem.

PR #173, opened 2018-10-18
Add Doxygen Grouping to SolverInterface

fsimonis:
This PR bundles the members of the SolverInterface into smaller logical groups.

The result:
image
---8<---
image


uekerman:
Good thing

PR #174, opened 2018-10-19
Reduce mesh inner state

floli:
The _enclosingRadius_ and the _center_ of each mesh primitive (Edge, Triangle, Quad) were computed explicitly in `Mesh::computeState`.

Now each primitive computes it inside `getCenter` / `getEnclosingRadius` from its geometry.

This greatly reduces the complexity of  `Mesh::computeState`. It also reduces the inner state of the primitives, and therefore the danger of inconsistencies, e.g. when someone changes coordinates but does not recalculate the center. Last but not least, it's just the right thing to do.

The only possibly downside I see, is that the performance of `getCenter` / `getEnclosingRadius` is somewhat lower, because it's computed upon each request. Still, it's not an expensive computation and it's only used in performance insensitive parts of precice.

### DIFF
--- a/src/mesh/Edge.cpp
+++ b/src/mesh/Edge.cpp
@@ -11,8 +11,7 @@ Edge:: Edge
 :
   _vertices( {&vertexOne, &vertexTwo} ),
   _id ( id ),
-  _normal ( Eigen::VectorXd::Constant(vertexOne.getDimensions(), 0.0) ),
-  _center ( Eigen::VectorXd::Constant(vertexOne.getDimensions(), 0.0) )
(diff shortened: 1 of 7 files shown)


fsimonis:
This definitely looks like a sane change to me.

Is the center calculation of primitives a hot instruction?

floli:
You ask, if it's a potential hot spot? No, as I said, they are rarely used, basically only in computeState, VTK export and irrc in some action. Compute state is only invoked when a mesh changes, i.e. never.

fsimonis:
LGTM

uekerman:
Just missed the merge, but looks good to me, too. There was indeed danger of inconsistencies.

PR #183, opened 2018-11-14
Add MeshRequirement to Actions

### LINKED ISSUE #81: ScaleByAreaAction needs to set meshRequirement
This is currently not the case and apparently a very old bug. 
setMeshEdge does not create edges and, thus, areas are not computed. 
The integration test `testExplicitWithDataScaling` would check for this (currently commented).
Formerly, the test did not see this as a built-in geometry was used (with existing edges).

fsimonis:
This adds a `MeshRequirement` to actions with a default of UNDEFINED.
Adding an Action to a participant upgrades the `MeshRequirement` of the targeted mesh if needed.

This PR fixed #81

### DIFF
--- a/src/action/Action.hpp
+++ b/src/action/Action.hpp
@@ -1,6 +1,7 @@
 #pragma once
 
 #include "mesh/SharedPointer.hpp"
+#include "mapping/Mapping.hpp"
 
 namespace precice
 {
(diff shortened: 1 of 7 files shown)


fsimonis:
@uekerman done

PR #191, opened 2018-11-16
Source Code Modernization

fsimonis:
I modernised various parts of the code by fixing the warnings from compiling some core classes with `-Wall -Wextra -Wpedantic` and additionally running `clang-tidy`.

Main changes include:
* const correctness especially in loops
* replacing `new X[s]`-`delete X[];` pairs with `std::vector`
* using `auto` with `new`
* Pass by value and move instead of pass by const ref and copy-construct
* Fix c style header `<signal.h>`-> `<csignal>`

### DIFF
--- a/src/precice/impl/Participant.cpp
+++ b/src/precice/impl/Participant.cpp
@@ -79,7 +79,7 @@ void Participant:: useMesh
   TRACE(_name,  mesh->getName(), mesh->getID() );
   checkDuplicatedUse(mesh);
   assertion ( mesh->getID() < (int)_meshContexts.size() );
-  MeshContext* context = new MeshContext(mesh->getDimensions());
+  auto context = new MeshContext(mesh->getDimensions());
   context->mesh = mesh;
   context->localOffset = localOffset;
   assertion ( mesh->getDimensions() == context->localOffset.size(),
@@ -105,7 +105,7 @@ void Participant:: addWriteData
 {
   checkDuplicatedData ( data );
   assertion ( data->getID() < (int)_dataContexts.size() );
-  DataContext* context = new DataContext ();
+  auto context = new DataContext ();
   context->fromData = data;
   context->mesh = mesh;
   // will be overwritten later if a mapping exists
@@ -121,7 +121,7 @@ void Participant:: addReadData
 {
   checkDuplicatedData ( data );
   assertion ( data->getID() < (int)_dataContexts.size() );
-  DataContext* context = new DataContext ();
+  auto context = new DataContext ();
   context->toData = data;
   context->mesh = mesh;
   // will be overwritten later if a mapping exists
--- a/src/precice/impl/Participant.cpp
+++ b/src/precice/impl/Participant.cpp
@@ -301,7 +301,7 @@ void Participant:: setClientServerCommunication
   com::PtrCommunication communication )
 {
   assertion ( communication.use_count() > 0 );
-  _clientServerCommunication = communication;
+  _clientServerCommunication = std::move(communication);
 }
 
 com::PtrCommunication Participant:: getClientServerCommunication() const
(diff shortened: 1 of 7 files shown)


floli:
Two remarks:

* It replaces the constructor `X::X(std::string const & s)` by `X::X(std::string s)` causing an additional copy operation. This is mitigated by using `s_(std::move(s))` in the c'tor initalizer list. So we have the same number of copy operations in total. Correct?
What is the net positive effect of this change? I actually prefer having const & as a c'tor argument, because it makes clear that there is no copy.

* I love replacing the new ... delete by an std::vector. Additionally, you could also use `_com->receive(ids, rankSender)` instead of `_com->receive(ids.data(), size, rankSender)`

fsimonis:
@floli good point about the receive calls. I will simplify this.

fsimonis:
Regarding the _pass by reference and move_-idiom:

**TLDR** If a function/constructor needs to create a copy of one of its arguments, then it is generally better to pass by value.

Passing by const reference discards information whether the argument was an rvalue or an lvalue. This gives the compiler a hard time to apply copy/move elision. When moving the intent to copy into the signature of the function then the compiler can perform some magic.

There are two interesting cases when using call by value:
* calling with an lvalue results in a simple copy, which is then available inside the function body. Nothing interesting here.
* calling with an rvalue results (if available) in a move construction which is then available inside the function body. The compiler can elide the move construction and construct the rvalue direct in the memory of the argument.

If the function is a constructor, which moves the argument passed by value into a member in the initialiser list, then the compiler can elide the move and make the caller construct the element directly inside the member.

This optimisation can be chained, so in the best case:
* calling the ctor with an lvalue directly copy-constructs the member using the reference.
* calling the ctor with an rvalue constructs the rvalue itself(!) inside the member.

S0mmerD:
**TLDR** Passing by value is not perfect but it is close enough and the alternatives are worse.

> This optimisation can be chained, so in the best case:
>  - calling the ctor with an lvalue directly copy-constructs the member using the reference.
>  - calling the ctor with an rvalue constructs the rvalue itself(!) inside the member.

Well we won't get this good because of the `s_(std::move(s))`. Unfortunately, this will always move the argument into the member instead of eliding the copy. For an example see 
https://godbolt.org/z/eKl-58
However this is the best approach IMHO. Enabling move elision would require many overloads or really strange universal references. So pass-by-value and `std::move` into member is the best default i guess. Interestingly, the C++ Core Guidelines [recommend const references](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const). However they don't give any reasons on why it should be better than pass-by-value.
This topic is also covered in an interesting [talk](https://youtu.be/PNRju6_yn3o) by Nicolai Josuttis on CppCon 2017.

floli:
Maybe the reason why the recommend pass by reference is:

> Using "unusual and clever" techniques causes surprises, slows understanding by other programmers, and encourages bugs. If you really feel the need for an optimization beyond the common techniques, measure to ensure that it really is an improvement, and document/comment because the improvement may not be portable.

[same link as above to core guidelines](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#f16-for-in-parameters-pass-cheaply-copied-types-by-value-and-others-by-reference-to-const)

fsimonis:
@S0mmerD You are compiling for size with `-Os`, try `-O2`.
When you switch the compiler to `clang(trunk)` you can add an `optimization view` using the + button. There you see what decisions the optimiser take.

The coreguidelines don't even mention the pass-by-value for normal functions.

fsimonis:
@S0mmerD Thanks for the info; that's some food for thoughts.

Can we merge this PR?

PR #194, opened 2018-11-16
Add .clang-tidy with a sane set of checks

fsimonis:
To use:
1) make a build dir `mkdir -p ~/tmp/precice && cd ~/tmp/precice`
2) configure precice using `cmake -DCMAKE_EXPORT_COMPILE_COMMANDS=ON $PRICICE_ROOT` (this generates the needed 'compile_commands.json`)
3) `cd $PRECICE_ROOT`
4) `clang-tidy -p ~/tmp/precice/ src/precice/impl/*.cpp` ( `-p` sets the dir where to find the file `compile_commands.json`) 
5) Wait a long time
6) Reflect on warnings such as [these](https://github.com/precice/precice/files/2588951/clang-tidy.txt).

Limitation: This works only for `cpp` files and their includes headers, which _shouldn't_ be a problem as tests _should_ cover all `hpp` files.


uekerman:
Thanks! Finally had time to look into this. I must admit that some of the warnings, I don't fully understand. Let's discuss some at the next coding days! 
Could you please copy the step-by-step instructions above to doxygen?

PR #195, opened 2018-11-16
WIP: Add missing MPI include to FindPETSc

fsimonis:
The FindPETSc tries to compile tests without including the include path of `mpi.h`.

It now includes the variable `PETSC_MPI_INCLUDE `  into the minimal set of includes.
If it is not set prior to calling the find module, then it attempts to locate it itself.


fsimonis:
I opened a PR regarding this in the original repository: https://github.com/jedbrown/cmake-modules/issues/32

fsimonis:
The author recommends to use the compiler wrapper.

PR #203, opened 2018-11-23
Add Index for all Mesh Primitives

fsimonis:
This PR adds the `PrimitiveRTree` to `precice::mesh::rtree`.

It can be generated using `precice::mesh::rtree::getPrimitiveRTree()` and gets cached just like the `rtree:VertexRTree`.

It is a bit more verbose than the VertexRTree as it needs to index Vertex, Edge, Triangle and Quad at the same time.
For each primitive, it needs to generate an axis-aligned bounding box (AABB) as an Indexable and a `PrimitiveIndex` as Value.
The struct `PrimitiveIndex` contains the index and type of the primitive as an enum for latter lookup.

The functor-like class `AABBGenerator` (parameterised with a `Mesh`) can generate an `AABB` for a given primitive.

This PR is a fragment of #156

### DIFF
--- a/src/mesh/RTree.cpp
+++ b/src/mesh/RTree.cpp
@@ -1,3 +1,5 @@
+#include "impl/RTree.hpp"
+
 #include "RTree.hpp"
 
 namespace precice {
@@ -64,6 +66,18 @@ Box3d getEnclosingBox(Vertex const & middlePoint, double sphereRadius)
   return box;
 }
 
+PrimitiveRTree indexMesh(const Mesh &mesh)
+{
+  using namespace impl;
+
+  AABBGenerator  gen{mesh};
+  PrimitiveRTree tree;
+  indexPrimitive(tree, gen, mesh.vertices());
+  indexPrimitive(tree, gen, mesh.edges());
+  indexPrimitive(tree, gen, mesh.triangles());
+  indexPrimitive(tree, gen, mesh.quads());
+  return tree;
+}
 
 std::ostream &operator<<(std::ostream &out, Primitive val)
 {
--- a/src/mesh/RTree.cpp
+++ b/src/mesh/RTree.cpp
@@ -32,13 +32,12 @@ PtrPrimitiveRTree rtree::getPrimitiveRTree(PtrMesh mesh)
   auto iter = _primitive_trees.find(mesh->getID());
   if (iter != _primitive_trees.end()) {
     return iter->second;
-  } else {
-    auto treeptr = std::make_shared<PrimitiveRTree>(indexMesh(*mesh));
-    _primitive_trees.insert(std::make_pair(
-        mesh->getID(),
-        treeptr));
-    return treeptr;
   }
+  auto treeptr = std::make_shared<PrimitiveRTree>(indexMesh(*mesh));
+  _primitive_trees.emplace(std::piecewise_construct,
+          std::forward_as_tuple(mesh->getID()),
+          std::forward_as_tuple(treeptr));
+  return treeptr;
 }
 
 void rtree::clear(Mesh &mesh)
--- a/src/mesh/RTree.cpp
+++ b/src/mesh/RTree.cpp
@@ -6,7 +6,7 @@ namespace precice {
 namespace mesh {
 
 // Initialize static member
-std::map<int, rtree::PtrRTree> precice::mesh::rtree::trees;
+std::map<int, rtree::PtrRTree> precice::mesh::rtree::_vertex_trees;
 // Initialize static member
 std::map<int, PtrPrimitiveRTree> precice::mesh::rtree::_primitive_trees;
 
@@ -15,7 +15,7 @@ rtree::PtrRTree rtree::getVertexRTree(PtrMesh mesh)
   RTreeParameters params;
   VertexIndexGetter ind(mesh->vertices());
     
-  auto result = trees.emplace(std::piecewise_construct,
+  auto result = _vertex_trees.emplace(std::piecewise_construct,
                               std::forward_as_tuple(mesh->getID()),
                               std::forward_as_tuple(std::make_shared<VertexRTree>(params, ind)));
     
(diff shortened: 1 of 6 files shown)


src/mesh/RTree.cpp
```diff
@@ -5,6 +5,8 @@ namespace mesh {
 
 // Initialize static member
 std::map<int, rtree::PtrRTree> precice::mesh::rtree::trees;
+// Initialize static member
+std::map<int, PtrPrimitiveRTree> precice::mesh::rtree::_primitive_trees;
```
floli: Why store pointer to trees and not the tree itself, like how it's done with VertexTree?

src/mesh/RTree.cpp
```diff
@@ -24,10 +26,25 @@ rtree::PtrRTree rtree::getVertexRTree(PtrMesh mesh)
   return tree;
 }
 
+PtrPrimitiveRTree rtree::getPrimitiveRTree(PtrMesh mesh)
```
floli: Looks like you're doing essentially the same as in `getVertexRTree`. Why not use the same code? e.g. `emplace_back` ...

src/mesh/RTree.cpp
```diff
@@ -24,10 +26,25 @@ rtree::PtrRTree rtree::getVertexRTree(PtrMesh mesh)
...
+        mesh->getID(),
+        treeptr));
+    return treeptr;
+  }
+}
 
-void rtree::clear(Mesh & mesh)
+void rtree::clear(Mesh &mesh)
 {
   trees.erase(mesh.getID());
```
floli: `trees` should be renamed to `vertexTrees` or alike.

src/mesh/RTree.hpp
```diff
@@ -1,21 +1,161 @@
...
+/// A standard print operator for PrimitiveIndex
+std::ostream& operator<<(std::ostream& out, PrimitiveIndex val);
+
+/// The axis aligned bounding box based on the Vertex Type
+using AABB = boost::geometry::model::box<Eigen::VectorXd>;
+
+/** Holding a reference to a Mesh it is a functor for packing
+ *  PrimitiveIndex into AABBs.
+ */
+class AABBGenerator
```
floli: Is there a chance to move that class to the implementation? Or even to `impl/` if it is purely internal use.

src/mesh/RTree.hpp
```diff
@@ -1,21 +1,161 @@
...
+
+/// The shared_ptr convenience type for PrimitiveRTree
+using PtrPrimitiveRTree = std::shared_ptr<PrimitiveRTree>;
+
+/** indexes a container of primitives into an rtree.
+ *
+ * The algorithm indexes every primitive of the given container using the passed generator.
+ * It inserts its result and the PrimitiveIndex into the rtree as a std::pair.
+ *
+ * \param[inout] rtree the rtree to index into
```
floli: `\param[in,out]`

src/mesh/Vertex.hpp
```diff
@@ -31,10 +31,20 @@ class Vertex : public PropertyContainer, private boost::noncopyable
   template<typename VECTOR_T>
   void setCoords ( const VECTOR_T& coordinates );
 
+  /// Sets the coordinates of the vertex.
+  /// RValue variant
```
floli: Not sure if Doxygen eats and digestes the double `///` correctly. I would just put on a single line. `/// Sets the coordinates of the vertex, rvalue variant`

src/mesh/RTree.cpp
fsimonis: The cache of the `VertexRTree` holds `shared_ptr` and I decided to stay consistent with the current implementation.
```cpp
  using PtrRTree = std::shared_ptr<VertexRTree>;
// ...
  static std::map<int, PtrRTree> trees;
```

src/mesh/RTree.cpp
fsimonis: You are right, will change the implementation to using `map emplace`.

src/mesh/RTree.cpp
fsimonis: I agree

src/mesh/RTree.hpp
fsimonis: Thanks, for little typos like this you could try the new suggestion feature.

src/mesh/Vertex.hpp
fsimonis: Will merge the lines.

src/mesh/RTree.hpp
fsimonis: I'll try to move some details things to `impl/RTree.hpp`

fsimonis:
@floli I implemented your feedback.

All in all, I think that both trees use a very consistent interface via `mesh::rtree`.
The fact that the new tree has to support all primitives naturally comes at the cost of increased verbosity.

src/mesh/RTree.cpp
fsimonis: I'll rename `PtrRTree` to `PtrVertexRTree` in #156 as it would affect many places in the code.

src/mesh/RTree.cpp
floli: My bad, you're right.

src/mesh/tests/RTreeTests.cpp
```diff
@@ -9,26 +11,169 @@ namespace bgi = boost::geometry::index;
 
 BOOST_AUTO_TEST_SUITE(MeshTests)
 BOOST_AUTO_TEST_SUITE(RTree)
+BOOST_AUTO_TEST_SUITE(BGAdapters)
+
+BOOST_AUTO_TEST_CASE(VectorAdapter)
```
S0mmerD: What is this Test for?

fsimonis: This is a test for the vertex adaptation to boost geometry.
@floli wrote this test and the one for the eigen vector. I grouped all of the adapter tests into a suite, therefor the diff.

PR #204, opened 2018-11-27
Migrate stacktrace to boost::stacktrace

### LINKED ISSUE #201: Abstract platform-specific Stacktrace Functionality
The following headers in are linux-specific:
https://github.com/precice/precice/blob/3a27a4117937d2b979fceacf045d8edab2b580a5/src/utils/stacktrace.hpp#L3-L4

We should investigate whether [boost::stacktrace](https://www.boost.org/doc/libs/1_65_0/doc/html/stacktrace.html) could be used to implement this for all major platforms.

This is related to #200

fsimonis:
Migrates `utils/stacktrace.hpp` to `boost::stacktrace`.
I moved the implementation to a separate cpp to decouple the compile time overhead of the boost header from the widely used assertion header.

I had to fix missing `<iomanip>`includes in other headers.

Fixes #201 

### TODO

* [ ] Add stacktrace to [wiki](https://github.com/precice/precice/wiki/Dependencies#boost)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,7 @@ All notable changes to this project will be documented in this file. For future 
 - Actions can now specify a `MeshRequirement`, such as the `ScaleByAreaAction`.
 - Many events have been reworked and are now uniformly named. 
 - There is a `syncMode` for events (for detailed performance measurements), configurable and off by default. 
+- Use boost stacktrace for cross platform stacktrace printing. This requires Boost 1.65
 
 ## 1.2.0
 - Make `polynomial=separate` the default setting for PetRBF.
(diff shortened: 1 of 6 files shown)


fsimonis:
Just realised the version dependency.

MakisH:
I guess we could keep this and add it later then. preCICE v2.0.0 would be a nice opportunity to move to a newer boost.

fsimonis:
I reopen this PR as we decided to upgrade Boost.

floli:
Would you please merge develop, which has boost 1.65 for Travis, into that branch and check if it works?

floli:
Please merge again from develop.

fsimonis:
Travis fails due to an issue with boost, which was fixed in Boost 1.65.1rc1 https://github.com/boostorg/stacktrace/issues/29#issuecomment-326656859.

MakisH:
> Travis fails due to an issue with boost, which was fixed in Boost 1.65.1rc1 [boostorg/stacktrace#29 (comment)](https://github.com/boostorg/stacktrace/issues/29#issuecomment-326656859).

Are you fixing this as well? @shkodm you may be interested in this when updating the system tests Dockerfiles.

fsimonis:
I just noticed that we use 1.65.1 in Travis.
It looks like travis reuses an old cached version: https://travis-ci.org/precice/precice/jobs/482336928#L728

fsimonis:
I tried to compile the project with scons and it fails with the same error.
So this problem is build-system related.

floli:
What the status on that? Travis with scons and boost 1.65 compiles fine. Anything holding it back?

PR #208, opened 2018-11-30
Rewrite Handle Iterators

### LINKED ISSUE #80: Rewrite Handler classes to be a proper Iterator
Rewrite EdgeHandler, VertexHandle, ... to be a proper iterator, thus working in range-based for loops.

Maybe with the help of http://www.boost.org/doc/libs/1_65_1/libs/iterator/doc/index.html

fsimonis:
This is a rewrite of the MeshHandle Iterators. They now fulfil the concept of [ForwardIterator](https://en.cppreference.com/w/cpp/named_req/ForwardIterator).

Closes #80 

## TODO
* [x] Add tests

### DIFF
--- a/src/precice/MeshHandle.cpp
+++ b/src/precice/MeshHandle.cpp
@@ -25,7 +25,8 @@ namespace precice {
 
 // VertexIterator
 
-VertexIterator:: VertexIterator(){}
+VertexIterator::VertexIterator() = default;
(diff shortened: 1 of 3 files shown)


fsimonis:
The usage of swap was intentional. Directly calling std swap ignores the
custom swap provided for the iterator types.

PR #209, opened 2018-12-11
Set the logging level from the boost test logging level (--log_level=...)

floli:
Only if no log.conf exists in the current working directory.

The mapping boost test level -> log level is, of course, open to debate. Currently it's:

```
--log_level=X`

X == all or success or test_suite or unit_scope => "%Severity% >= debug"
X == message => "%Severity% >= info"
X == warning => "%Severity% >= warning"
X == error or cpp_exception or system_error or fatal_error, nothing) => "%Severity% >= warning"
```

### DIFF
--- a/src/logging/LogConfiguration.hpp
+++ b/src/logging/LogConfiguration.hpp
@@ -27,6 +27,9 @@ struct BackendConfiguration
 /// Holds the configuration of the logging system
 using LoggingConfiguration = std::vector<BackendConfiguration>;
 
+/// Reads a log configuration file, returns vector of BackEndConfiguration
+LoggingConfiguration readLogConfFile(std::string const & filename);
+
 /// Configures the logging from a log file
 void setupLogging(std::string const & logConfigFile = "log.conf");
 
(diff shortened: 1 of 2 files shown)


floli:
Seems to fail with Boost 1.60 @ Travis.

MakisH:
Useful and more intuitive that the boost level specifiers! I aggree with the changes, I have not tested it yet, though.

floli:
> One remark: Why don't we include trace for all?

I often use `log_level=unit_scope` (which is the same log level as all) to identify a hanging test. (last message with: Entering test case ...). If trace would be active in such a case, preCICE really spewed a tons of messages and the Entering test case message would be gone beyond the horizon already.

That's why I intentionally did non enabled trace here.

Note, that you can always use a log.conf to activate trace and use other, finer grained log control.

uekerman:
I agree, that's a use case, which would be bad with trace output. Having a fifth option here is not possible? `all` and `unit_scope` have to be same technically?

floli:
>  Having a fifth option here is not possible? all and unit_scope have to be same technically?
Yes.

PR #212, opened 2018-12-12
Fix Warning regarding switch with scoped enum Mapping::MeshRequirement

fsimonis:
On clang `-Wreturn-type` complains with `control reaches end of non-void function`.


S0mmerD:
`assert` is a no-op in release builds. This will only work on debug builds.

fsimonis:
@S0mmerD true, thankfully Boost provides the `BOOST_UNREACHABLE_RETURN` macro for this purpose.

PR #213, opened 2018-12-12
Add Error for using PETRBF without PETSc

### LINKED ISSUE #211: Trying to use petrbf without PETSc should give a better error
When trying to use an RBF mapping that depends on PETSc (e.g. `mapping:petrbf-thin-plate-splines`) but preCICE is not built with `petsc=yes`, the user gets the following error (preCICE v1.2.0):

```
[config:MappingConfiguration]:482 in createMapping: ERROR: Unknown mapping type!
```

Better would be to give the user a hint for PETSc. Even better would be to have a list of acceptable mapping types with their respective requirements and check if this is actually the problem.

I added a hint in our [Error Messages wiki page](https://github.com/precice/precice/wiki/Error-Messages#trying-to-use-parallel-rbf-mapping-without-petsc).

fsimonis:
preCICE now tells you of with
> PETRBF mappings are not available as preCICE was build without PETSc.

Fix #211


floli:
Idea: Compare with `type.starts_with("petrbf-");` instead of listing all types explicitly.

fsimonis:
@floli Thanks for the hint. I had to use the C++11 version though.

PR #214, opened 2018-12-12
Improve Doxygen Namespace Documentation

fsimonis:
* Adds missing documentation of main namespaces as a separate file.
* Adds missing namespace comments for _sub-namespaces_

### DIFF
--- a/Doxyfile
+++ b/Doxyfile
@@ -816,7 +816,7 @@ EXCLUDE_PATTERNS       =
 # Note that the wildcards are matched against the file with absolute path, so to
 # exclude all test directories use the pattern */test/*
 
-EXCLUDE_SYMBOLS        =
+EXCLUDE_SYMBOLS        = MappingTests MeshTests CplSchemeTests PreciceTests
 
 # The EXAMPLE_PATH tag can be used to specify one or more files or directories
 # that contain example code fragments that are included (see the \include
--- a/docs/documents/namespaces.dox
+++ b/docs/documents/namespaces.dox
@@ -1,0 +1,71 @@
+/** @namespace precice
+ *  @brief main namespace of the precice library
+ */
+
+/** @namespace precice::action
+ *  @brief contains actions to modify exchanged data
+ */
+
+/** @namespace precice::com
+ *  @brief contains the data communication abstraction layer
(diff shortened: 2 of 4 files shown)


fsimonis:
Doxygen currently generates a namespace for each python file resulting in the entries marked below.
To clean them up, we can:
1) Do nothing :coffee: 
1) Ignore all python files in doxygen
2) Disable the generation of namespaces for python files by setting `EXTRACT_ALL = NO` in the Doxyfile, which has other side effects.
3) Add `__init__.py` to folders containing python files, such as `tools/` and `src/actions/tests/` to give doxygen a hint that they are part of a "package"
4) Move all python files to a separate directory

This is the current state of the namespace documentation.
image

fsimonis:
Let's split this into multiple PRs. The python verbosity is a separate issue.

Doxyfile
```diff
@@ -816,7 +816,7 @@ EXCLUDE_PATTERNS       =
 # Note that the wildcards are matched against the file with absolute path, so to
 # exclude all test directories use the pattern */test/*
 
-EXCLUDE_SYMBOLS        =
+EXCLUDE_SYMBOLS        = MappingTests MeshTests CplSchemeTests PreciceTests
```
floli: Why are them excluded? Why only these test suite? Why are tests at all excluded?

fsimonis: These are the symbols generated by Boost.Test and are only mentioned as forward-, and friend-declarations in our code.
So, they are currently listed as empty entries under [namespaces](http://xgm.de/precice/docs/develop/namespaceCplSchemeTests.html).

Looking at this again I realise that this will not scale in the future.
Maybe it would be better to wrap the fwd/friend declarations in `#ifndef DOXYGEN_SHOULD_SKIP_THIS` blocks.

floli: Ok, now I understand.

I think blacklisting them in the Doxyfile is fine then. Don't like the `#ifndef` stuff, would just clutter the code and the Doxyfile is read only seldom, so we should move clutter there.

fsimonis: I don't like it either. Maybe we can create a single friend helper class to avoid all this clutter.

PR #215, opened 2018-12-12
WIP: Modernize CMakeLists and update Travis

S0mmerD:
This PR is meant to modernize the way preCICE is built. It does the following things:

- Use pkgconfig for PETSc. FindPETSc was always brittle and even Jed Brown, the creator of the FindPETSc module, recommended using pkgconfig
- Create a cmake target for preCICE. This is somehow WIP, as it's not quite clear to me, which dependencies of precice are to be public or private. Packaging preCICE as a target will make using preCICE easier as all that is required to use preCICE will be 
```
find_package(preCICE)
target_link_library(lib preCICE)
```
- Update travis to Ubuntu 16.04 LTS (xenial).
- Use cmake on travis (instead of SCons).


S0mmerD:
@fsimonis
Thanks for the review. I did not know that you were working on this problem too. Would you care to share the branch that you are working on?

I agree to most of your points. `FindPETSc` is badly named. It was only supposed to be a wrapper, instead of a full-fledged search module. About CMakeLists.txt: I would love to see your version. About some of your point specifically.

> I strongly disagree with removing the env variables, the user made a choice and the system has to deal with it.

This only removes the environment variables for Cmake (which should not use them anyway, right?)
>The point of the target-based system is to link targets. There are targets for most libraries and we can add the rest (for old find modules).

This might not be that easy. The Basis is Ubuntu 16.04 LTS (xenial). This has some restrictions on the versions of certain libraries. In Xenial, Cmake is version 3.5.  FindLibXml2 will export Targets only with cmake >= 3.12, Eigen will work with cmake >= 3.3 (This will work). Boost targets work with cmake >= 3.5 (This is fine as well), PETSc does not export targets at all, and FindMPI exports targets with cmake >= 3.9. This means that for Ubuntu 18.04 one would need to provide Wrappers for PETSc, Libxml2 and MPI. 

I will not continue working on this PR, as i think you are more ambitious in this direction, but i would love to collaborate on your PR.

shkodm:
Regarding Travis: I have Scons + CMake ( with builds on 14.04, 16.04 and 18.04 in docker ) Travis setup from some time [here](https://github.com/shkodm/precice). I have not opened a pull request yet since macOS builds are still failing for unknown reasons. I will try to fix it and open a pull request in the near future.

fsimonis:
@S0mmerD I sent you a message via Gitter concerning this straight after I heard that you are working on this. This was even before you opened this PR. My version is on my fork and still in heavy development. I will create the PR as soon as I feel that it is ready for review/discussion.

We had a discussion about ubuntu, boost and cmake. As we require a separate install of boost on 16.04LTS, there is no reason to keep cmake 3.5.1 as a minimum requirement. Instead it makes way more sense to use the next LTS as a baseline. The cmake verision on 18.04LTS is 3.10.2. CMake is _thankfully_ trivial to install and unproblematic to use, thus we decided to target this version. We are a bit more careful increasing the Boost version though.

PR #219, opened 2018-12-19
Fixing relative path

BenjaminRueth:
Before, the relative path assumed that `$PRECICE_ROOT` is called `precice`. However, if the root folder is called differently (e.g. `precice-1.3.0`, if preCICE Version 1.3.0 is used) the compilation of PySolverInterface breaks.

### DIFF
--- a/src/precice/bindings/python/setup.py
+++ b/src/precice/bindings/python/setup.py
@@ -12,7 +12,7 @@ from distutils.command.build import build
 APPNAME = "PySolverInterface"
 
 PYTHON_BINDINGS_PATH = os.path.dirname(os.path.abspath(__file__))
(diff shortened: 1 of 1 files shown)


BenjaminRueth:
Important not to forget: Should we create a precice version 1.3.1 with this bugfix?

floli:
Don't think it's so pressing, just wait for next coding days soon.

fsimonis:
Do we actually need this information at all?

Assuming preCICE was installed properly, then the compiler should find the include and the library. In such a case we wouldn't have to pass neither `-I` nor `-L`.

MakisH:
@fsimonis I don't think we can do this with the current setup (scons). Am I missing something?

fsimonis:
@MakisH This will work with scons and is actually more robust. The Pr just removes an unnecessary level of going back in the relative path `../precice`.

PR #228, opened 2019-01-03
Add State Awareness to setMesh Functions

fsimonis:
This PR adds the flag `_canSetMesh` to SolverInterfaceImpl which keeps track of wether the `setMeshXXX` functions can be called or not.

Calling these functions at the wrong time now results in the error:  
`Cannot set the mesh in the current state!`


src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -624,6 +629,9 @@ int SolverInterfaceImpl:: setMeshVertex
   const double* position )
 {
   TRACE(meshID);
+  if(!_canSetMesh) {
+      ERROR("Cannot set the mesh in the current state!");
+  }
   Eigen::VectorXd internalPosition(_dimensions);
```
uekerman: This (and all other cases) should be replaced by a `CHECK`. 
Also, the error message needs to be more informative. Currently, I guess that users might be puzzled.

uekerman: A perfect error message is no easy task, though

uekerman:
What about making this a property of each mesh instead? Or we change `resetMesh` to `resetMeshes` ?

fsimonis:
@uekerman We really should keep track of this state on a Mesh to Mesh basis.
The big question is whether this state belongs into the solver interface or the mesh.

#### Option A: Extending the SolverInterface
The permission to write to a mesh depends on the state of the solver interface. Thus, it should keep track of the permission using a little multi-lock.
Adding it increases the state of the SolverInterface, but keeps the Mesh class untouched.

```c++
template<class Key>
class MultiLock {
public:
  void add(Key k, bool locked); // called in configure
  void remove(Key k);
  void lock(Key k); 
  void unlock(Key k);      // called from resetMesh(k)
  void lockAll();          // called from advance(), initialize()
  void unlockAll();        // called from configure()
  bool check(Key k) const; // called from `createMesh` functions
};
```

#### Option B: Extending the Mesh
The Mesh class currently does not have any notion of being writeable or not. _(apart from its own constness)_ We could add a flag and the interface to deal with it. The Mesh `add` functions would then check for this state.
Adding it increased the state of the Mesh, which could affect a large part of our test suite.

```c++
class Mesh {
public:
  // ...
  bool canChange() const; // called from `add` functions
  void setChangeable(bool newstate); // called from configure(), initialize(), resetMesh(k), advance()
  // ...
};
```

uekerman:
Just discussed with @fsimonis. We go for option A.

fsimonis:
We push this PR to the next milestone as it breaks _at least_ one test, without a clear reason why.

src/precice/impl/SolverInterfaceImpl.cpp
uekerman: Instead of the meshID, we should give the mesh name as feedback to the user.

src/utils/MultiLock.hpp
```diff
@@ -0,0 +1,105 @@
...
+    }
+  }
+
+  void unlock(const Key &name)
+  {
+    auto iter = _locks.find(name);
+    if (iter == _locks.end()) {
+      throw LockNotFoundException{};
+    } else {
+      iter->second = true;
```
uekerman:
```suggestion
      iter->second = false;
```

src/precice/impl/SolverInterfaceImpl.cpp
fsimonis: I changed the `ERROR` to `CHECK` but had to do a bit more work to print the name of the mesh.

Problem is that the SolverInterface(Impl) provides neither a `hasMesh(int meshID)` nor a `getMeshName(int meshID)`.

To reduce clutter, I added a type MeshInfo which prints a given meshID and its name if the id is actually known by the SolverInterface. The factory function `SolverInterfaceImpl::makeMeshInfo(int meshID)` builds and returns an object of this type tyed to the SolverInterfaceImpl.
The `operator<<` is defined, so it can be inlined into `CHECK`/`ERROR`.


Example output of `cout << makeMeshInfo(meshID)`:
```
1:"MeshOne"
2:<unknown>
```

uekerman: I am not sure if I understand this. Which case is `<unknown>` here? 
Normally, you access the mesh via the participant `_accessor`:
```
_accessor->meshContext(meshID)
```
I have the feeling you are over-complicating things here. Sth like `makeMeshInfo` should not be part of `SolverInterfaceImpl` in my opinion.

uekerman:
You could now also remove

https://github.com/precice/precice/blob/61b777f61f4c3c147b0881cd27b725abe3feeaef/src/precice/impl/SolverInterfaceImpl.cpp#L659-L660

and all similar occurrences.

src/precice/impl/SolverInterfaceImpl.cpp
fsimonis: `<unknown>` is a mesh where the given `meshID` is unknown to the participant aka wrong input.

It is a small utility type to prevent additional control flow around the `ERROR` and  `CHECK` macros to cover the case that the meshID doesn't even exist.

PR #231, opened 2019-01-15
New EventTimings

floli:
* Normalize times inside on rank. Time normalization between participants is done in python postprocessing
* Output JSON instead of CSV data. 
* Add JSON parser from https://github.com/nlohmann/json, licensed under MIT.
* Put all information (timings and state changes) into one file.
* Add a MPI_Mock file to preCICE. Mocks some very basic MPI functions when there is no MPI. Makes it easier to adapt the EventTimings to preCICE if `PRECICE_NO_MPI` is set.
* Split up `EventTimings.*` into `Event.*` (include that file if you want to set an Event in the code) and `EventUtils.*`
* Save the number of iterations of some of the linear solvers used in PetRBF in the respective event. It's not printed at the moment, but logged to the file.


src/utils/EventUtils.cpp
```diff
@@ -0,0 +1,565 @@
...
+  long minTicks;
+  MPI_Reduce(&ticks, &minTicks, 1, MPI_LONG, MPI_MIN, 0, EventRegistry::instance().getMPIComm());
+
+  ticks = localRankData.finalizedAt.time_since_epoch().count();
+  long maxTicks;
+  MPI_Reduce(&ticks, &maxTicks, 1, MPI_LONG, MPI_MAX, 0, EventRegistry::instance().getMPIComm());
+
+  // This assumes the same epoch and ticks rep, should be true for system time
+
+  return std::make_pair(
```
fsimonis:
```suggestion
  return {
```

src/utils/EventUtils.cpp
```diff
@@ -0,0 +1,565 @@
...
+    sys_clk::time_point{sys_clk::duration{minTicks}},
+    sys_clk::time_point{sys_clk::duration{maxTicks}}
+    );
```
fsimonis:
```suggestion
    };
```

src/utils/EventUtils.cpp
```diff
@@ -0,0 +1,565 @@
...
+void RankData::normalizeTo(sys_clk::time_point t0)
+{
+  auto const delta = initializedAt - t0; // duration that this rank initialized after the first rank
+  assertion(t0 <= initializedAt); // t0 should always be before or equal my init time
+
+  for (auto & events : evData) {
+    for (auto & sc : events.second.stateChanges) {
+      auto & tp = sc.second;
+      tp = stdy_clk::time_point(tp - initializedAtTicks + delta);
+      // Deactivated, as it fails for the precice tests, likely because Events are used before initialize
```
fsimonis: I think this should be fine. If an Event was record before t0, then a negative tp is ok as our normalisation is more of an _alignement_.
```suggestion
```

src/utils/EventUtils.cpp
```diff
@@ -0,0 +1,565 @@
...
+      // assert(tp.time_since_epoch().count() >= 0); // Trying to do normalize twice?
```
fsimonis:
```suggestion
```

src/utils/EventUtils.cpp
```diff
@@ -0,0 +1,565 @@
...
+  auto insertion = storedEvents.emplace(std::piecewise_construct,
+                                        std::forward_as_tuple(name),
+                                        std::forward_as_tuple(name, false, false));
+
+  prefix = previousPrefix;
+  return std::get<0>(insertion)->second;
+}
+
+
+void EventRegistry::printAll()
```
fsimonis: This function looks weird to me as it says _print_, but it does _print and writes the log file only from master_ .
Can you please split it up?

Here a suggestion:
```cpp
void EventRegistry::printSummary() { 
  writeSummary(std::cout);
}

void EventRegistry::writeLog() { 
  std::string logFile;
  if (applicationName.empty())
    logFile = "Events.json";
  else
    logFile = applicationName + "-events.json";

  std::ofstream ofs(logFile);
  writeJSON(ofs);
 }

void EventRegistry::logFromMaster() {
  int myRank;
  MPI_Comm_rank(comm, &myRank);

  if (myRank == 0) {
    printSummary();
    writeLog();
  }
}
```

src/utils/EventUtils.cpp
```diff
@@ -0,0 +1,565 @@
...
+    }
+
+    out << endl;
+    printGlobalStats();
+    out << endl << std::flush;
+  }
+}
+
+
+void EventRegistry::writeLog(std::ostream & out)
```
fsimonis:
```suggestion
void EventRegistry::writeJSON(std::ostream & out)
```

src/utils/EventUtils.cpp
```diff
@@ -0,0 +1,565 @@
...
+  else
+    logFile = applicationName + "-events.json";
+
+  writeTimings(std::cout);
+  std::ofstream ofs(logFile);
+  writeLog(ofs);
+}
+
+
+void EventRegistry::writeTimings(std::ostream &out)
```
fsimonis:
```suggestion
void EventRegistry::writeSummary(std::ostream &out)
```

src/utils/EventUtils.hpp
```diff
@@ -0,0 +1,206 @@
...
+  void printAll();
+
+  /// Prints the result table to an arbitrary stream.
+  void writeTimings(std::ostream &out);
+
+  void writeLog(std::ostream & out);
+  
+  void printGlobalStats();
+
+  MPI_Comm & getMPIComm();
```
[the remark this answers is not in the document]
floli: I read your comment as "Shouldn't this be a const reference?" or "Should this be rather a non-const reference?" Correct?

src/utils/EventUtils.cpp
floli: In an ideal world, there is no event recorded prior to t0, i.e., prior to initialization. But obviously, we do not live in an ideal world.

src/utils/EventUtils.cpp
floli: I am not to eager to split it up, since I want /one/ function that is called after finalize and does all the log writing / printing.
`writeTimings` and `writeLog` are ok, they both write to their argument `ostream out`.

The equivalance of master and rank 0 is a precice only thing, and actually, it shouldn't be here either.

fsimonis: _"Split up"_ was a poor formulation.
I really meant that it would be a good idea to _refactor_ those independent parts out of that one function. This makes the code more generic and easier to read as the responsibilities of each function are clearer.

There are 5 concerns here:
1) __generate a json representation__ of the recorded events and write them to a generic ostream  
`EventRegistry::writeJSON(osteam)` (suggested name change in review)
2) __generate a summary__ of the timings of the recorded events and write them to a generic ostream  
Calls `EventRegistry::writeSummary(ostream)`
3) generate and __write the log file__   (suggested name change in review)
`void EventRegistry::writeLog()` (should probably be `writeLogFile()`)  
Calls `EventRegistry::writeJSON(ofsteam)`
4) __print the summary__ to std::out  
`void EventRegistry::printSummary()`
Calls `EventRegistry::writeSummary(std::cout)`
5) Print summary and write log file from master (__preCICE specific__)  
`void logFromMaster()`

_Note: the generation of the file name is a 6th concern, but not important here._


> I want /one/ function that is called after finalize and does all the log writing / printing.

That would be a call to `EventRegistry::instance()::logFromMaster()`.


Now there is only one minor precice specific part of the EventTimings that could be eventually refactored out to further decouple the project from precice:
```cpp
namespace precice::utils {
void logEventsFromMaster() {
  if(utils::Parallel::getProcessRank() != 0) {
    return;
  }
  auto& instance = EventRegistry::instance();
  instance.printSummary();
  instance.writeLog();
}
}
```

PR #241, opened 2019-01-31
Fix XML Validator Leaks

fsimonis:
This PR fixes memory leaks by modernizing the XMLAttribute and Validator codes.

* Modernize _Private undefined_ to `=delete` for constructors
* Modernize owning raw ptrs to `std::unique_ptr`
* Modernize using `override` in derived classes
* Fix leaks in `operator||`
* Fix sanity of `clone()`
* Introduce `makeValidatorEquals` which returns a `unique_ptr`


floli:
Just an idea: Is the explicit specialization of `makeValidatorEquals<std::string>("on")` needed or could we make the compiler put in `std::string` automatically? Or an overload instead?

That would save a lot of keystrokes.

fsimonis:
@floli Good point. I thought of that, but `"on"` is a `const char *`. Deduction guides would solve the problem but they come in `C++17`.

I am thinking of adding a literal suffix so that you could write `"on"_Validator` instead  of `makeValidatorEquals<std::string>("on")`.
I am also considering to migrate this entire machinery to leverage compile time polymorphism.

I'll have a look at the overload option.

fsimonis:
@floli I added an overload for `const char *`, so it works now without template arguments.

floli:
Great, go for merge!

PR #245, opened 2019-02-04
WIP: Add Validator Literals

fsimonis:
This PR adds user-defined literals for string and int validators and applies them where appropriate.

### DIFF
--- a/src/logging/config/LogConfiguration.cpp
+++ b/src/logging/config/LogConfiguration.cpp
@@ -15,6 +15,7 @@ LogConfiguration::LogConfiguration
   precice::logging::setupLogging();
   
   using namespace xml;
+  using namespace xml::literals;
   XMLTag tagLog(*this, "log", XMLTag::OCCUR_NOT_OR_ONCE);
   tagLog.setDocumentation("Configures logging");
 
@@ -26,7 +27,7 @@ LogConfiguration::LogConfiguration
   XMLTag tagSink(*this, "sink", XMLTag::OCCUR_ARBITRARY);
   XMLAttribute<std::string> attrType("type");
   attrType.setDocumentation("Type of sink");
-  attrType.setValidator ( makeValidatorEquals("stream") || makeValidatorEquals("file") );
+  attrType.setValidator ( "stream"_eq || "file"_eq );
   attrType.setDefaultValue(precice::logging::BackendConfiguration::default_type);
   tagSink.addAttribute(attrType);
 
--- a/src/mapping/config/MappingConfiguration.cpp
+++ b/src/mapping/config/MappingConfiguration.cpp
@@ -42,6 +42,7 @@ MappingConfiguration:: MappingConfiguration
 {
   assertion (_meshConfig.use_count() > 0);
   using namespace xml;
+  using namespace xml::literals;
 
   XMLAttribute<double> attrShapeParam ( ATTR_SHAPE_PARAM );
   attrShapeParam.setDocumentation("Specific shape parameter for RBF basis function.");
@@ -62,19 +63,19 @@ MappingConfiguration:: MappingConfiguration
   XMLAttribute<std::string> attrPolynomial("polynomial");
   attrPolynomial.setDocumentation("Toggles use of the global polynomial");
   attrPolynomial.setDefaultValue("separate");
-  attrPolynomial.setValidator(makeValidatorEquals("on")
-                              || makeValidatorEquals("off")
-                              || makeValidatorEquals("separate"));
+  attrPolynomial.setValidator("on"_eq
+                           || "off"_eq
+                           || "separate"_eq);
 
 
   XMLAttribute<std::string> attrPreallocation("preallocation");
   attrPreallocation.setDocumentation("Sets kind of preallocation for PETSc RBF implementation");
   attrPreallocation.setDefaultValue("tree");
-  attrPreallocation.setValidator(makeValidatorEquals("estimate")
-                              || makeValidatorEquals("compute")
-                              || makeValidatorEquals("off")
-                              || makeValidatorEquals("save")
-                              || makeValidatorEquals("tree"));
+  attrPreallocation.setValidator("estimate"_eq
+                              || "compute"_eq
+                              || "off"_eq
+                              || "save"_eq
+                              || "tree"_eq);
 
 
   XMLTag::Occurrence occ = XMLTag::OCCUR_ARBITRARY;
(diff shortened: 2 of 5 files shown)


floli:
I am not really sure if we gain much from this change. Of course, it's convenient, but is less explicit than saying `makeValidatorEquals("foo")` compared to `"foo"_eq'. To understand the latter, how first have to find the code defining the literal and and understand it.

fsimonis:
I think that `makeValidatorEquals("Foo")` is awfully verbose.

Our aim is to provide a simple verification DSL which handles equality and selection. We would like to write:
```cpp
attr.setValidator("Foo" || "Bar" || "Fizz" || NAME_OF_OPTION )
```
We have:
* `ValidatorEquals<std::string>("Foo")` requires the template argument
* `makeValidatorEquals("Foo")` allows us to omit the template parameter but it is verbose

We can reduce the verbosity with these literals `"Foo"_eq`, which is a simple forward to the call above. Name alternatives that come to mind: `"Foo"_equals`, `"Foo"_validator` , `"Foo"_asValidatorEquals`
I am not particularly happy with the literals solution as they do not work with predefined strings, a case we have to handle often in the current codebase.

We could also go another route by stepping back from the DSL approach:
```cpp
attr.setValidator(Validator::Or{"Foo", "Bar", "Fizz", NAME_OF_OPTION}):
```

Summerdave:
**TL;DR** Validator is just a glorified runtime enum, which never uses runtime values. We could replace it by a compile time solution.
## The Validator inheritance is awfully verbose and over-engineered
Some code from `MappingConfiguration.cpp`
```
  ValidatorEquals<std::string> validDirectionWrite ( VALUE_WRITE );
  ValidatorEquals<std::string> validDirectionRead ( VALUE_READ );
  attrDirection.setValidator ( validDirectionWrite || validDirectionRead );

  XMLAttribute<std::string> attrFromMesh(ATTR_FROM);
  XMLAttribute<std::string> attrToMesh(ATTR_TO);

  XMLAttribute<std::string> attrConstraint(ATTR_CONSTRAINT);
  ValidString validConservative(VALUE_CONSERVATIVE);
  ValidString validConsistent(VALUE_CONSISTENT);
  attrConstraint.setValidator(validConservative || validConsistent);
```
notice that earlier there is `using ValidString = ValidatorEquals<std::string>;`. This is not only verbose but also not that easy to understand, as the Validators right now work by recursively descending their inheritance tree.
## Validators can be created entirely at compile time
**There is not a single instance of a Validator in preCICE that was created with a runtime value.** However there is code like 
```
class FooConfig
{
  FooConfig() : VALUE1("value1"), VALUE3("value"),...
...
  const std::string VALUE1;
  const std::string VALUE2 = "value2";
  const std::string VALUE3;
};
```
Instead of 
```
class FooConfig {
  constexpr char VALUE1[] = "value1";
  constexpr char VALUE2[] = "value2";
  constexpr char VALUE3[] = "value3";
}
```
Then we could write code like
```
template <typename... T>
constexpr makeValidator(T... vals) -> Validator<sizeof...(vals), T...>;
```
for creating Validators. `XMLAttribute` is a class template anyway, so we could just overload it with a function that forwards the arguments to makeValidator.
## Less Compile-Time computation, less Variadic Templates?
Another way would be:
```
class ValidatorBase {...};
template <int sz, typename... T>
Validator : public ValidatorBase {...};
unique_ptr<ValidatorBase> makeValidator(...);
```
Which would be less intrusive for `XMLAttribute`.
We could also drop the `Validator` template in favor of explicit template specialization, saving some compilation time.
```
//Validator.hpp
template <typename T>
class Validator 
{
  std::vector<T> allowed;
}
using StringValidator = Validator<std::string>;
using IntValidator = Validator<int>;

//Validator.cpp
template <typename T>
Validator::Validator(...){...}
...
template class Validator<std::string>;
template class Validator<int>;
```
With this solution we would no longer need user-defined literals, as we could simply write
```
attr.setValidator("value1", "value2", "value3");
```

fsimonis:
@Summerdave that's definitely true and I actually created a compile time prototype on my longer-than-expected return from FOSDEM.

You rightly noticed that our pre-C++11 XML code-base desperately needs some serious clean-up.
Most of this validator noise can _and probably should_ be pushed down to where we setup the `XMLAttribute` by calling `setValidator`.

I would prefer a non-invasive solution, which leaves the XMLAttribute mainly untouched. In my prototype it simply holds a `unique_ptr` to a virtual validator base class and that's it.
```cpp
void XMLAttribute::setValidator(std::unique_ptr<ValidatorBase>&& validator);
```
As you said, the validator polymorphism is known at compile time. So, I packed the entire Validator polymorphism into an additional type wrapper, which allows to compose all of that using CRTP. I used [Clara](https://github.com/catchorg/Clara/blob/master/include/clara.hpp#L466) as an inspiration for that.

I created this under the assumption that we want to keep the composability with `operator||` and that we might want to extend the validator logic at some point. That said, I __honestly doubt__ that these assumptions are necessary.

---

At the end of the day, this entire validator mess could just be a `set<string> _options` in `XMLAttribute`. Imagine this:
https://github.com/precice/precice/blob/d601da73c78bf684fa986bb4e2d81861fc96b982/src/precice/config/ParticipantConfiguration.cpp#L127-L141
to look like:
```cpp
const std::string doc = " ... ";
const auto attrGeoFilter = makeXMLAttribute("geometric-filter", VALUE_BROADCAST_FILTER, doc, 
  VALUE_FILTER_FIRST, VALUE_BROADCAST_FILTER, VALUE_NO_FILTER
); 

// or
const std::string doc = /* ... */
const auto attrGeoFilter = makeXMLAttribute("geometric-filter", VALUE_BROADCAST_FILTER).setDoc(doc).setValidator(
  VALUE_FILTER_FIRST, VALUE_BROADCAST_FILTER, VALUE_NO_FILTER);

// or even
const std::set<std::string> filterOptions{VALUE_FILTER_FIRST, VALUE_BROADCAST_FILTER, VALUE_NO_FILTER};
const std::string doc = /* ... */
const auto attrGeoFilter = makeXMLAttribute("geometric-filter", VALUE_BROADCAST_FILTER).setDoc(doc).setValidator(filterOptions); 
```
We can deduce the type of the attribute from the default value and use chaining to set the other values as well as parameter packs for the validators.
And we should generate the documentation and the errors for that on demand. There is simply no reason not to.

fsimonis:
We did not reach consensus on what to do next, but we reached consensus on this being the wrong heading.

Issue #104, opened 2018-02-14
Performance limited by receive buffer capacity (1-way coupling)


MakisH:
Let's assume this scenario: Participant A sends data to Participant B during `advance()` and continues to the next timestep. If Participant B is too slow, then the continuous sending of data will saturate the receive buffer of B. Participant A will then hang until more data can be received.

This problem can be platform-dependent. In order to solve this issue, we need to be able first to find out the size of these buffers. Then we could either configure these buffers to have a different size (if possible), or store the sending data somewhere else before putting them in the system buffers.

In order to reproduce this issue, I use the aste framework (commit [75b800f](https://github.com/precice/aste/commit/75b800f03ecd3f5e1b45a825d481f2f98a684dd3) at the moment), where I added the following lines in the `main.cpp`:

```c++
#include <thread>
#include <chrono>

// ... after advance():

    if (participant == "B") {
        std::this_thread::sleep_for (std::chrono::seconds(999999));
    }
```

and I set the `max-time` in `precice.xml` to a very high number. This way, A keeps sending data, but B never reads them. On my laptop, Participant A paused at the `dt# 4384`, using the default mesh sizes (10x10 elements). I am investigating these further, also in the MAC Cluster and I will update the issue.

floli:
> Let's assume this scenario: Participant A sends data to Participant B during advance() and continues to the next timestep. If Participant B is too slow, then the continuous sending of data will saturate the receive buffer of B. Participant A will then hang until more data can be received.

Minor correction imho: A will hang when receive buffer of B is saturated and send buffer of A.

I have just merged my fork into aste, could you branch and commit your changes, s.t. we can also test it using exact the same code and settings.

MakisH:
I ran it on the MAC-Cluster (partition `snb`) in the following way:

```
# Allocate two nodes:
# We want 4 MPI processes (tasks) for each participant, so 8 in total. Each node has 32 logical processors (CPUs), so we require 8 CPUs per task (64 CPUs in total = 2 nodes).
salloc --partition=snb --ntasks=8 --cpus-per-task=8

# Run participant A with 4 processes, all in one node, redirect the output to log.A and return
mpiexec.hydra -ppn 4 -n 4 ./aste -p A > log.A 2>&1&

# The same for participant B
mpiexec.hydra -ppn 4 -n 4 ./aste -p B > log.B 2>&1&

# Inspect the log.A
tail -f log.A
```

[...]

PR #247, opened 2019-02-06
Fix SocketCommunication aSend with a Queue

Summerdave:
In `SocketCommuncation` there is a bug when mutiple `asio::aSend` invocations are in flight simultaneously. This PR aims to solve this issue by introducing a Queue which ensures that the sends are done serially.


src/com/SocketCommunication.cpp
```diff
@@ -534,11 +534,11 @@ PtrRequest SocketCommunication::aSend(const bool & itemToSend, int rankReceiver)
...
 
   try {
-    asio::async_write(*_sockets[rankReceiver],
-                      asio::buffer(&itemToSend, sizeof(bool)),
-                      [request](boost::system::error_code const &, std::size_t) {
-                        std::static_pointer_cast<SocketRequest>(request)->complete();
-                      });
+    _queue.push(_sockets[rankReceiver],
+                asio::buffer(&itemToSend, sizeof(bool)),
+                  [request]() {
```
fsimonis:
```suggestion
                  [request] {
```

src/com/SocketCommunication.cpp
```diff
@@ -387,11 +387,11 @@ PtrRequest SocketCommunication::aSend(const int *itemsToSend, int size, int rank
...
 
   try {
-    asio::async_write(*_sockets[rankReceiver],
-                      asio::buffer(itemsToSend, size * sizeof(int)),
-                      [request](boost::system::error_code const &, std::size_t) {
-                        std::static_pointer_cast<SocketRequest>(request)->complete();
-                      });
+    _queue.push(_sockets[rankReceiver],
+                asio::buffer(itemsToSend, size * sizeof(int)),
+                  [request]() {
```
fsimonis:
```suggestion
                  [request] {
```

src/com/SocketCommunication.cpp
```diff
@@ -427,11 +427,11 @@ PtrRequest SocketCommunication::aSend(const double *itemsToSend, int size, int r
...
 
   try {
-    asio::async_write(*_sockets[rankReceiver],
-                      asio::buffer(itemsToSend, size * sizeof(double)),
-                      [request](boost::system::error_code const &, std::size_t) {
-                        std::static_pointer_cast<SocketRequest>(request)->complete();
-                      });
+    _queue.push(_sockets[rankReceiver],
+                asio::buffer(itemsToSend, size * sizeof(double)),
+                  [request]() {
```
fsimonis:
```suggestion
                  [request] {
```

src/com/SocketCommunication.cpp
```diff
@@ -451,11 +451,11 @@ PtrRequest SocketCommunication::aSend(std::vector<double> const & itemsToSend, i
...
 
   try {
-    asio::async_write(*_sockets[rankReceiver],
-                      asio::buffer(itemsToSend),
-                      [request](boost::system::error_code const &, std::size_t) {
-                        std::static_pointer_cast<SocketRequest>(request)->complete();
-                      });
+    _queue.push(_sockets[rankReceiver],
+                asio::buffer(itemsToSend),
+                  [request]() {
```
fsimonis:
```suggestion
                  [request] {
```

src/com/SocketSendQueue.cpp
```diff
@@ -0,0 +1,35 @@
+#include "SocketSendQueue.hpp"
+
+namespace precice
+{
+namespace com
+{
+namespace asio = boost::asio;
+
+void SendQueue::push(std::shared_ptr<Socket> sock,
```
fsimonis: This function is called `push` but does `push_and_process`. I think the caller should decide whether to process the queue or not by explicitly calling `process()`.

As this function modifies the underlying queue, so it should acquire the `_queueMutex` first.

src/com/SocketSendQueue.cpp
```diff
@@ -0,0 +1,35 @@
...
+  if (!_ready || _itemQueue.empty())
+    return;
+  auto item = _itemQueue.front();
+  _itemQueue.pop_front();
+  _ready = false;
+  asio::async_write(*(item.sock),
+                    item.data,
+                    [item, this](boost::system::error_code const &, std::size_t) {
+                      item.callback();
+                      this->_ready = true;
```
fsimonis: This is a race condition depending on the asynchronousness of the send. Consider using a `std::atomic_bool` for this.

Summerdave:
`_queueMutex` is a bad name. It came from a previous design where both methods acquired a lock. The idea is that you as a caller don't care about starting/stopping the queue at all. It just runs when it should run (that is: When it is not empty). The way the queue works is to call `process()` enough times to actually process all the items inside the queue. 
About the race condition you remarked: I cannot quite see the problem here. There is at most one request active by design and `push()` is also used by only one thread. This means that a possible race condition will only occur if a request finishes while a new request is pushed onto the queue. In this case one of the threads will acquire the lock, set `_ready` to `false` and the other thread will do nothing once he acquired the lock.

fsimonis:
Ok, you ensure the queue gets fully processed by calling `process` from within the callback.
I just had a look at [asynchronous operations](https://www.boost.org/doc/libs/1_65_1/doc/html/boost_asio/reference/asynchronous_operations.html) and the suspected data race should not occur.

Shouldn't the destructor of the Queue somehow handle the unsent items?

The name of the function does not feel right though, as it does more then the name tells. What about `dispatch()`? That includes both _hand over_ and _process_.

uekerman:
Just to avoid any confusion: Is this PR connected to https://github.com/precice/precice/issues/104 ?

fsimonis:
@Summerdave In the destructor of `SocketsCommunicator` we could check the size of the queue and just log a warning along the lines of `"SocketsCommunicator destructor called with " << _queue.size() << " items in sender queue !"`. You could implement `size()` and `empyt()` as simple `noexcept` proxies.

Summerdave:
After verifying that it solved the problem it was supposed to solve on the SuperMUC, this PR is ready for final review + merge.

uekerman:
Does this PR solve the issues with the buffered communication for uni-directional coupling?
If yes, great and thanks a lot! 
Which tests did you run? @floli mentioned some tests on SuperMUC.  
Thus this mean that we should switch on the buffering again?
https://github.com/precice/precice/blob/develop/src/m2n/PointToPointCommunication.cpp#L568

Summerdave:
> Does this PR solve the issues with the buffered communication for uni-directional coupling?

Yes.

> Which tests did you run? @floli mentioned some tests on SuperMUC.

There was a use case which exhibited the undesired behaviour, on which i tested the PR which resolved the issue. 
> Thus this mean that we should switch on the buffering again?

Yes and i forgot to include this in the PR. Added.

src/com/SocketSendQueue.cpp
```diff
@@ -0,0 +1,43 @@
...
+{
+namespace com
+{
+namespace asio = boost::asio;
+
+void SendQueue::push(std::shared_ptr<Socket> sock, 
+        boost::asio::const_buffers_1 data, 
+        std::function<void()> callback)
+{
+  _itemQueue.push_back({sock, data, callback});
```
fsimonis: I _think_ you can call `emplace_back` here and move the sock and data. Can you please check if that is possible?
```suggestion
  _itemQueue.emplace_back(std::move(sock), std::move(data), callback);
```

Summerdave: Unfortunately this does not work. 
**TL;DR**: It should work but the standard library forbids it.
`SendItem` does not have any constructors other than the implicit ones. In the end `vector::emplace_back` tries to construct a new element by `::new (p) (sock, data, callback)`. This has to do with the way `std::allocator_traits::construct` works. I don't quite know why they don't use the uniform initialization syntax (since it was introduced with C++11 anyway), but my guess is that they want to avoid these strange cases where constructors with `std::initializer_list` take precedence over other constructors. Therefore no aggregate initialization is performed.

Summerdave: In principle i could add the respective constructor to `SendItem` but copying a `shared_ptr` and a `const_buffers_1` is very cheap (the latter is just a `void*` with a `size_t`) anyway.

fsimonis: That's a shame, but good to be aware of that! Thank you for the write-up.

You can move the shared ptr and the data member though :wink:

```suggestion
  _itemQueue.push_back({std::move(sock), std::move(data), callback});
```

floli:
Superseded by #338

PR #240, opened 2019-01-30
CMake Overhaul

fsimonis:
The PR features the overhaul of the CMake build configuration of precice.

I kept the _Interface_ stable, so you can configure using the old way:
```
$ cmake -DMPI=ON -DPETSC=ON -DPYTHON=ON \
        -DCMAKE_BUILD_TYPE=Debug \
        -DBUILD_SHARED_LIBS=ON \
         $PRECICE_ROOT
```
After that, you can:
1) Build `make`
2) Run tests `make test` or `ctest -V`
3) Generate packages `make package` or `cpack`
4) Install precice `[sudo] make install`

For a scoped install:
```
$ cmake -DMPI=ON -DPETSC=ON -DPYTHON=ON \
        -DCMAKE_BUILD_TYPE=Debug \
        -DBUILD_SHARED_LIBS=ON \
        -DCMAKE_INSTALL_PREFIX=/path/to/install/dir \
         $PRECICE_ROOT
$ make install -j $(nproc)
```

## Changes
  - New Options:
    - `PRECICE_Packages` (default ON) to configure CPack
    - `PRECICE_InstallTest` (default ON) to configure installation of tests.  
      This includes the binary `testprecice` and necessary files.
      Use `PREFIX/share/precice` as `PRECICE_ROOT`.
  - Move CMake files from `tools/cmake-modules` to `cmake/` (general scripts) and `cmake/modules` (find modules).
  - Migration from file-globbing to explicit source/interface/test-file lists.  
    Simply run the helper `tools/updateSourceFiles.py` from project-root to update all necessary files.
    __Rational:__ CMake evaluates file glob only during configuration time, adding a file to the source tree will thus not be picked up by the build system. [More Information](https://www.youtube.com/watch?v=bsXLMQ6WgIk&t=16m).
    
  - `install` target installs:
     - the library `PREFIX/lib`.
     - the binaries `PREFIX/bin` and their manfiles `PREFIX/share/man/man1`.
     - the cmake configuration files `PREFIX/lib/cmake/precice`.
     - the necessary files to run testprecice `PREFIX/share/precice`. Use this as `PRECICE_ROOT` on installed system.
  - CTest definition of tests running in isolated working directories:
    - `precice.Base` for the base test suite
    - `precice.MPI2` run on 2 MPI ranks
    - `precice.MPI4` run on 4 MPI ranks
  - CPack configuration of target `package` to generate binary debian, tar and zip packages.
[...]

PR #251, opened 2019-02-07
Fix preCICE on Boost 1.69

Summerdave:
Boost.Test 1.69 is broken on preCICE. There are two issues:

1.  `collector` became `collector_t`
Fixed by this PR with a simple `#if` preprocessor directive.
2. At runtime `testprecice` fails at:
https://github.com/precice/precice/blob/d601da73c78bf684fa986bb4e2d81861fc96b982/src/testing/main.cpp#L47
This is caused by a bug in Boost.Test, which is discussed [there](https://github.com/boostorg/test/issues/199). It only affects preCICE in boost version 1.69, as the author already fixed it. This does, however, pose the question of how we should handle it. Suggestions are welcome :).

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -115,6 +115,7 @@ set(headers
   src/precice/bindings/c/Constants.h)
 set_target_properties(precice PROPERTIES PUBLIC_HEADER "${headers}")
 
+target_include_directories(precice PUBLIC ${Boost_INCLUDE_DIRS})
 target_link_libraries(precice PRIVATE ${PYTHON_LIBRARIES})
 target_link_libraries(precice PUBLIC ${MPI_LIBRARIES})
 target_link_libraries(precice PUBLIC ${Boost_LIBRARIES})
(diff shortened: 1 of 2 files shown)


fsimonis:
@Summerdave It doesn't _feel right_, but we could simply `#if` the problem out of existence for boost 1.69 by printing a __bold__ warning and setting the loglevel to debug.

floli:
The first part of the fix is already included in https://github.com/precice/precice/tree/fix_for_boost_1.69
For the second part, the runtime error, I don't see no good solution at this time.

BTW: What is the CMakeLists.txt change about?

fsimonis:
@floli We are linking against a list of libraries, so the include directories were indeed missing.

floli:
Shouldn't it be rather `PRIVATE`, since boost is not part of our public interface?

I will not merge this PR, instead take over the relevant part to the before-mentioned branch.

@fsimonis Do you take over the missing boost include into your cmake overhaul branch?

fsimonis:
@floli Yes, it should be private, but It wont matter as our dependencies should all be private. I fixed this and the missing include in #240.

PR #253, opened 2019-02-11
Prototype of XMLAttribute Overhaul

fsimonis:
This PR is a major overhaul of the `XMLAttribute` internals. The changes will
* replace the xml Validators with a simple collection of "options" in `XMLAttribute`
* make `setDocumenation`, `setValidator`, and `setDefaultValue` chainable
* provide `makeXMLAttribute(name, default)` that deduced the attribute type from default. (and std::string from `const char *`)
* change `setValidator` to take an `initializer_list<T>` for T convertibe to the attribute type
  This allows to write following `makeXMLAttribute("dimension", 2).setValidator({2, 3})`
* apply this to the configuration


floli:
If I get it correctly, the syntax changed from
```
attrType.setValidator ( makeValidatorEquals("stream") || makeValidatorEquals("file") );
```
to
```
 auto attrType = XMLAttribute<std::string>("type").setValidator({"stream", "file"})
```

So the set validator assumes that validators are combined with OR.

Is there a reason, why the old way, of combining them with operator|| was abolished? I kind of liked the expressiveness.
How can I know combine validators with AND? (do we have that case at all?)

uekerman:
Looks good to me. I only had a brief look, but I also like the brefty.

> How can I know combine validators with AND? (do we have that case at all?)

We don't have such cases. In the end, like you said, the items are just possible options. 

@fsimonis
Did you check whether the same "manual.xml" is generated?

floli:
> > How can I know combine validators with AND? (do we have that case at all?)
> 
> We don't have such cases. In the end, like you said, the items are just possible options.

That's ok, but then the function should be renamed to indicate that more clearly.

fsimonis:
@floli You are absolutely right about the name, so I renamed `setValidator` to `setOptions` to keep the function name small.

I went for a simple options-check as we have simply no use case for AND. It also allows for an extreme simplification: DSL with virtual classes pointers etc vs vector of options.
Note: I used `std::vector` with `std::unique` as it does not require the type to provide a weak ordering (`std::set`) or the type to be hashable (`std::unordered_set`). This is necessary due to the Eigen types.

@uekerman The generated XML looks similar, although the indentation looks very ... [strange](https://github.com/precice/precice/files/2855315/out.xml.txt)

PR #255, opened 2019-02-11
Small cleanup in m2n::PointToPointCommunication

floli:
Should not be problematic, since functionality is untouched. Just wanted to wait for Travis to check it ans will eventually merge it, also without review. Request @uekerman because we planed that change together today.

* Since we hold just one com::Communication instance per P2P object,
move the object out of the mappings vector and save it just once per
class.

* Remove premature resize of vectors. At this point the size of the
indices that are moved into the vector are not yet known. Therefore
the resize will be wrong anyway.

* Reword some of the comments.

### DIFF
--- a/src/m2n/PointToPointCommunication.cpp
+++ b/src/m2n/PointToPointCommunication.cpp
@@ -367,24 +367,20 @@ void PointToPointCommunication::acceptConnection(std::string const &acceptorName
     return;
   }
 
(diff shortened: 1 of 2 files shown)


uekerman:
I just realized that now we even don't need the communication factories any longer. 
I created an issue about that: https://github.com/precice/precice/issues/256

PR #271, opened 2019-02-13
WIP Add a Git verification step to updateSourceFiles

fsimonis:
This PR adds:
* A verification step to the source generation
* IGNORE_PATTERNS to ignore bindings etc.
* CONFIGURED_SOURCES to add the result of `.in` files such as `src/versions.hpp.in`

```console
$ tools/updateSourceFiles.py
Detected:
 sources: 296
 public: 5
 tests: 70
The source tree contains files not tracked by git.
Please do one of the following with these files:
  - track them using 'git add'
  - add them to IGNORE_PATTERNS in this script
  - add them to CONFIGURED_SOURCES in this script!
Files:
  src/precice/bindings/python/BadPythonBoy.hpp
  src/precice/bindings/python/BadPythonBoy.cpp
  src/precice/BadBoy.cpp
VERIFICATION FAILED
```

```console
$ tools/updateSourceFiles.py         
Detected:
 sources: 296
 public: 5
 tests: 70
Git did not run successfully.
These sources are UNVERIFIED!
Generating CMake files
Writing Files
 ./src/sources.cmake
 ./src/tests.cmake
done
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -31,6 +31,7 @@ All notable changes to this project will be documented in this file. For future 
   - CPack configuration of target `package` to generate binary debian, tar and zip packages.
   - Add `CMakeLists.txt` to `tools/solverdummy/cpp`. It is an example of how to link to precice with CMake.
   - Extend the displayed information when configuring.
+- Extend `updateSourceFiles.py` to verify the sources using `git ls-files --full-name` if available.
 
 ## 1.3.0
 - Update of build procedure for python bindings (see [`precice/src/bindings/python/README.md`](https://github.com/precice/precice/blob/develop/src/precice/bindings/python/README.md) for instructions). Note: you do not have to add `PySolverInterface.so` to `PYTHONPATH` manually anymore, if you want to use it in your adapter. Python should be able to find it automatically.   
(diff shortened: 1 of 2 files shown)


floli:
I don't think that's a good idea. If I understand correctly, the script wants to force me to either add the files to git or modify the script in order to compile? Seriously? Sometime I just want to play with some modifications and backup `cp file.cpp file.backup`. Or I add a new source file, but I want to compile before checking in, or have them laying around for some time.

What about issuing the warning and do the generation anyway?

fsimonis:
@floli That's the intention, yes.
In your example, the `file.backup` wouldn't be picked up anyhow as it is not a `.cpp` file. Adding files to the index also does not mean that you commit them.

We could of course display the warning and generate the cmake files anyhow.

floli:
> We could of course display the warning and generate the cmake files anyhow.

Under that condition, I am cool with the changes, but if I read the code correctly, it does not do that right now. (`if not_tracked: return 1`)

Please refrain from enforcing automatism.

fsimonis:
@floli The script now prints the warning and generates the cmake files.

_Is selective automatism really automatism?_

PR #272, opened 2019-02-13
Moved version logging to configure

### LINKED ISSUE #269: Writing preCICE version only after the logger is correcty set up
Currently, we get 
*This is preCICE version 1.3.0. Starting 2019-02-12 11:35:30* 
for every rank because the logger is not already set up. Not that nice if you run (massively) parallel case :wink:

uekerman:
This solves the issue that we always got a "This is preCICE version ..." for every rank and not yet formatted properly. 
See issue #269

### DIFF
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -88,8 +88,8 @@ void SolverInterfaceImpl:: configure
   config::Configuration config;
   xml::configure(config.getXMLTag(), configurationFileName);
   if(_accessorProcessRank==0){
-    INFO("Configuring preCICE with configuration: \"" << configurationFileName << "\"" );
     INFO("This is preCICE version " << PRECICE_VERSION);
+    INFO("Configuring preCICE with configuration: \"" << configurationFileName << "\"" );
   }
   configure(config.getSolverInterfaceConfiguration());
 }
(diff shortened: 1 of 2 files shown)


src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -88,6 +89,7 @@ void SolverInterfaceImpl:: configure
   xml::configure(config.getXMLTag(), configurationFileName);
   if(_accessorProcessRank==0){
     INFO("Configuring preCICE with configuration: \"" << configurationFileName << "\"" );
+    INFO("This is preCICE version " << PRECICE_VERSION);
```
floli: I would definitely put the "This is preCICE" line before "Configuring...".

Also, I think the startup time is useful, what was the reason for removing it?

uekerman: The time is written by the logger anyway, right?
```
(0) 16:36:09 [impl::SolverInterfaceImpl]:91 in configure: Configuring preCICE with configuration: "../precice-config.xml"
(0) 16:36:09 [impl::SolverInterfaceImpl]:92 in configure: This is preCICE version 1.3.0
```

floli: Ah ok, yes. Still, I would also put the date and that regardless of logger configuration. So that the first preCICE thing you see, when looking as a look is also a time stamp of the run.

uekerman: I changed the order of statements. About the date, I am not so sure. What use case would that be for? Wouldn't that be the job of the solver?

floli: Mhh, don't think it can clearly by said it's either the job of the solver or the precice. Anyway, it's fine for me now.

PR #275, opened 2019-02-14
Fortran77 bindings: Fix the missing const declaration of an input var…

nico-krais:
…iable in the precicef_write_data_required interface. This caused the routine to not be available to the Fortran codes that use the library.


MakisH:
Thank you, @nico-krais! Amazing how the compiler missed that one. Please open more Pull Requests if you notice anything else going wrong!

PR #278, opened 2019-02-14
Improve CTest from user perspective

fsimonis:
We currently have a few problematic tests including the `MPI_Ports` and the `PetRBF` Tests.

This PR changes the test suite so that there is a safe base set the user can run.

* If configured without MPI, there is only one test registered running `testprecice`.
* If configured with MPI there are multiple suites:
  * run with `mpiexec -np 4`
    * __Base__ containing all tests except the `MPI_Ports` and the `PetRBF/Paralllel`
    * __MPIPorts__ containing all `@MPI_Ports` tests
    * __PetRBFParallel__ containing the `PetRBF/Parallel` tests
  * run without mpi
    * __NOMPI__ containing all tests

To run:
* all tests: execute `make test`
* the base set: execute `make test_base`

For additional control, use CTest directly `ctest --help`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,11 @@ All notable changes to this project will be documented in this file. For future 
   - Extend the displayed information when configuring.
 - Extend `updateSourceFiles.py` to verify the sources using `git ls-files --full-name` if available.
 - Fix the VTK XML exporter not to write VertexNormals
+- Improve the user-friendliness of the tests.
+  - `make test` will run all tests.
+  - `make test_base` only a unproblematic base-set.
+  - A timeout will kill hanging tests.
+  - All tests sets run in isolated working directories.
 
 ## 1.3.0
 - Update of build procedure for python bindings (see [`precice/src/bindings/python/README.md`](https://github.com/precice/precice/blob/develop/src/precice/bindings/python/README.md) for instructions). Note: you do not have to add `PySolverInterface.so` to `PYTHONPATH` manually anymore, if you want to use it in your adapter. Python should be able to find it automatically.   
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -269,40 +269,6 @@ include(${CMAKE_CURRENT_LIST_DIR}/src/tests.cmake)
 
 
 #
-# CTest
-#
-
-enable_testing()
-
-# Register the Base test
-set(PRECICE_TEST_DIR "${preCICE_BINARY_DIR}/TestOutput")
-file(MAKE_DIRECTORY ${PRECICE_TEST_DIR}/Base)
-add_test(NAME precice.Base
-  COMMAND "$<TARGET_FILE:testprecice>"
-  WORKING_DIRECTORY "${PRECICE_TEST_DIR}/Base"
-  )
-if(MPI AND MPIEXEC_EXECUTABLE)
-  # Register the MPI test for 2 Ranks
-  file(MAKE_DIRECTORY "${PRECICE_TEST_DIR}/MPI2")
-  add_test(NAME precice.MPI2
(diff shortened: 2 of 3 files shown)


MakisH:
> To run:
> 
>     * all tests: execute `make test`
> 
>     * the base set: execute `make test_base`
> 
> 
> For additional control, use CTest directly `ctest --help`.

Why is `make test` the full one (and more prone to failure)? I understood that we wanted to have the default the one that would not fail so easily.

floli:
> Why is `make test` the full one (and more prone to failure)? I understood that we wanted to have the default the one that would not fail so easily.

We should make the most comprehensive test suite the default. Sooner or later people and developers will just use `make test` and the other test will be more and more neglected. So after something fails, the first question from one developer to another will be, if he ran the non-default test. And the reply will likely be "oh no, just the base test".

uekerman:
We decided yesterday to make the default the currently working one. Main concern was the user experience and the support we had to do the last months. Tests that users run should not require to look up the known test issues page. 
Also as a developer, I would do `make test` at the moment as I know that the other ones currently fail on my laptop.

MakisH:
Besides, we have Travis for running the full suite.

fsimonis:
So it turns out that one cannot change the `test` target (`make test`).
Thus, the current scheme is the most viable including:

* `make test` runs all tests starting with `precice.Base`. The other tests have the CTest label `canfail`. So the user gets a comment regarding that. This target actually simply calls `ctest`.
* `make test_base` runs only `precice.Base`
*  `make test_install` _future work_ will eventually configure and build the cpp solverdummy (xSDK) 

 Also, the default target does not enforce any builds (stupid, really). The target `test_base` depends on `testprecice`, so it will ensure that it is actually built.

floli:
>  `make test` runs all tests starting with `precice.Base`. The other tests have the CTest label `canfail`. So the user gets a comment regarding that. This target actually simply calls `ctest`.

Sorry, I don't understand. What tests are run now by `make test`?

Marking some tests as ok to fail is a good solution in contrast to disable them.

fsimonis:
@floli You are right, it was a bit unclear.

# Tests
If configured with MPI `make test` runs 5 tests

1.  all tests except MPI_Ports and PetRBF Parallel on 4 Rank (this is the __base testset__)
2. only MPI_Ports tests on 4 Ranks
3. only PETRBF Parallel on 4 Ranks
4.  all tests on 4 Ranks
5. all tests without MPI

In CTest, tests 2-4 are marked as `canfail` and a separate summary will be displayed for them. 

If configured without MPI `make test` runs 1 test:

1. all tests without MPI (this is the __base testset__)

# Also Important
* Each test runs with a TIMEOUT set to 60 seconds in order to prevent communication tests from waiting forever. _I'll change this to individual timeouts.__
* `make test` simply calls `ctest` (This is build-in and not changeable) 
* `make test_base` calls `ctest` on the __base testset__ in verbose mode. So you actually see the output.
* `make test_base` is the what we tell the users to run.

fsimonis:
The test TIMEOUTS are now 3 minutes for the big MPI testsuites (1,4)  and 20 seconds for the small MPI test suites (2,3).

floli:
> 1. all tests except MPI_Ports and PetRBF Parallel on 4 Rank (this is the **base testset**)
> 2. only MPI_Ports tests on 4 Ranks
> 3. only PETRBF Parallel on 4 Ranks
> 4. all tests on 4 Ranks
> 5. all tests without MPI

Isn't 4 = 1 + 2 + 3? Not that we run most of the tests twice ;-)

fsimonis:
@floli You are right about the full test suite. So lets leave that out.
If you are ok with the PR, feel free to merge.

PR #286, opened 2019-02-15
SCons cannot find PETSc if PETSC_ARCH is needed for full path to PETSc.

ajaust:
I am on Ubuntu 18.04 and have installed PETSc through the package manager. This installs PETSc into `/usr/lib/petscdir/3.7.7/x86_64-linux-gnu-real/`. Following the guide in the wiki, see [wiki page](https://github.com/precice/precice/wiki/Dependencies#petsc-optional), I set

```bash
export PETSC_DIR=/usr/lib/petscdir/3.7.7
export PETSC_ARCH=x86_64-linux-gnu-real
```

However, SCons fails to build preCICE since it does not find PETSc. Line 189 of the `SConstruct` file reads:

```
with open(join(PETSC_DIR, "include/petscversion.h"), "r") as versionfile:
```

This means that it expects that `PETSC_DIR` would point to the full path `/usr/lib/petscdir/3.7.7/x86_64-linux-gnu-real`. This is not an issue when following the other instruction in the wiki, see [wiki page](https://github.com/precice/precice/wiki/Dependencies#ubuntu-1804). There it is suggested to set 

```
export PETSC_DIR=/usr/lib/petscdir/3.7/
```

If one follows this part of the manuels, it works since `/usr/lib/petscdir/3.7/` is a symlink to `/usr/lib/petscdir/3.7.7/x86_64-linux-gnu-real/` on Ubuntu 18.04.

I propose to change line 189 of the SConstruct to (since SCons is still supported):

```python
with open(join( join(PETSC_DIR, PETSC_ARCH), "include/petscversion.h"), "r") as versionfile:
```

*Edit*: I changed it to avoid the double `join` as suggested below.

This will work if `PETSC_ARCH` is empty and `PETSC_DIR` points directly to the PETSc dir and if the full path is given by `$PETSC_DIR/$PETSC_ARCH`. I think both options (`PETSC_ARCH` empty and set) are allowed by newer versions of PETSc.

**Error message**

```
scons: Reading SConscript files ...

(default)  builddir   = build      Directory holding build files. ( /path/to/builddir )
(default)  build      = Debug      Build type (release|debug|Release|Debug)
(default)  libprefix  = /usr       Path prefix for libraries ( /path/to/libprefix )
(default)  compiler   = mpicxx     Compiler to use.
(default)  mpi        = True       Enables MPI-based communication and running coupling tests. (yes|no)
(modified) petsc      = True       Enable use of the PETSc linear algebra library. (yes|no)
(modified) python     = True       Used for Python scripted solver actions. (yes|no)
(default)  gprof      = False      Used in detailed performance analysis. (yes|no)
(default)  platform   = none       Special configuration for certain platforms (none|supermuc|hazelhen)

Checking whether the C++ compiler works... yes
Checking for C++ library pthread... yes
(modified) PETSC_DIR  = /usr/lib/petscdir/3.7.7
(modified) PETSC_ARCH = x86_64-linux-gnu-real
Checking for C++ library petsc... yes
IOError: [Errno 2] No such file or directory: '/usr/lib/petscdir/3.7.7/include/petscversion.h':
  File "/home/jaustar/precice-test/precice-git/SConstruct", line 189:
    with open(join(PETSC_DIR, "include/petscversion.h"), "r") as versionfile:
```

**How to reproduce**
- Under the assumption that you run Ubuntu 18.04 an have installed PETSc through the package manager you can run the following script. It will checkout preCICE and try build it using SCons.

```bash
#/bin/bash -l

precice_ver="git"
nworkers=8
target_dir="precice-git"


if [[ ! -d ${target_dir} ]]; then
  git clone git@github.com:precice/precice.git ${target_dir}
fi

cd ${target_dir}
export PRECICE_ROOT=${PWD}

# Important to reproduce the bug
export PETSC_DIR=/usr/lib/petscdir/3.7.7
export PETSC_ARCH=x86_64-linux-gnu-real

scons petsc=yes python=yes -j ${nworkers}

./tools/compileAndTest.py -t
```

### DIFF
--- a/SConstruct
+++ b/SConstruct
@@ -186,7 +186,7 @@ if env["petsc"]:
     else:
         checkAdd("petsc")
     # Set PETSC_VERSION to correct values 
-    with open(join(PETSC_DIR, "include/petscversion.h"), "r") as versionfile:
+    with open(join( join(PETSC_DIR, PETSC_ARCH), "include/petscversion.h"), "r") as versionfile:
         for line in versionfile:
             tokens = line.split()
             try:
--- a/SConstruct
+++ b/SConstruct
@@ -185,8 +185,15 @@ if env["petsc"]:
         checkAdd("craypetsc_gnu_real")
     else:
         checkAdd("petsc")
+        
+    petsc_path = []
+    if ( FindFile( "petscversion.h", join( PETSC_DIR, PETSC_ARCH, "include/" ) ) == None ):
+      petsc_path = PETSC_DIR
+    else:
+      petsc_path = join( PETSC_DIR, PETSC_ARCH )
+
     # Set PETSC_VERSION to correct values 
-    with open(join( PETSC_DIR, PETSC_ARCH, "include/petscversion.h"), "r") as versionfile:
+    with open(join( petsc_path, "include/petscversion.h"), "r") as versionfile:
         for line in versionfile:
             tokens = line.split()
             try:
--- a/SConstruct
+++ b/SConstruct
@@ -186,7 +186,7 @@ if env["petsc"]:
     else:
         checkAdd("petsc")
     # Set PETSC_VERSION to correct values 
-    with open(join( join(PETSC_DIR, PETSC_ARCH), "include/petscversion.h"), "r") as versionfile:
+    with open(join( PETSC_DIR, PETSC_ARCH, "include/petscversion.h"), "r") as versionfile:
         for line in versionfile:
             tokens = line.split()
             try:


ajaust:
You could also use replace line 189 `with open(join(PETSC_DIR, PETSC_ARCH, "include/petscversion.h"), "r") as versionfile:`. I am not sure what is better and why I broke the Travis build.

floli:
Please use the latter variant, it'd definitely better than the nested join.

The build broke, because when PETSc is downloaded using git (https://www.mcs.anl.gov/petsc/download/index.html), `petscversion.h` is in `$PETSC_DIR/include` not in `$PETSC_DIR/$PETS_ARCH/include`

ajaust:
My PETSc `petscversion.h` lives in:

```
/usr/lib/petscdir/3.7.7/x86_64-linux-gnu-real/include/petscversion.h
```

Ubuntu also has symlinks all over the place. I can also find it in some more directores by adding `include/` to it. Here is a most likely incomplete list:

```
3.7 -> /etc/alternatives/petsc3.7
3.7-real -> /etc/alternatives/petsc3.7-real
/etc/alternatives/petsc -> /usr/lib/petscdir/3.7.7/x86_64-linux-gnu-real
/etc/alternatives/petsc3.7 -> /usr/lib/petscdir/3.7.7/x86_64-linux-gnu-real
/etc/alternatives/petsc3.7-real -> /usr/lib/petscdir/3.7.7/x86_64-linux-gnu-real
```

fsimonis:
@ajaust Thank you for the PR! There are actually two possible locations:

* using a deb install, it is in `PETSC_DIR/PETSC_ARCH/include`
* using a checkout build, it is located in `PETSC_DIR/include/`

Can you change the SConstruct by adding a localisation step based on PETSC_DIR and PETSC_ARCH?

ajaust:
@fsimonis I am sure we could add a localisation step, but would this give us any advantage? If PETSC_ARCH is being used it should be set. If it is not used, as for checkout builds, it is either unset or set to an empty string. Based on this it is safe to join the strings `PETSC_DIR` and `PETSC_ARCH`.

At the moment, this is exactly what the SCons file does for `CPPPATH` and `LIBPATH`, too. See lines 180-183:

```python
    env.Append(CPPPATH = [join(prefix, PETSC_DIR, "include"),
                          join(prefix, PETSC_DIR, PETSC_ARCH, "include")])
    env.Append(LIBPATH = [join(prefix, PETSC_DIR, PETSC_ARCH, "lib"), 
                          join(prefix, PETSC_DIR, "lib")])
```

If I adapt this to distinct between the two cases `PETSC_ARCH` set vs. unset, does it imply this should be changed for `CPPPATH` and `LIBPATH`, too?

ajaust:
@fsimonis I have adapted the SConstruct file. It worked for me locally. Let's see if it works for travis, too.

PR #288, opened 2019-02-15
Updating cpp solver dummy

### LINKED ISSUE #239: Some issues with C++ solver dummy
I encountered a number of minor problems with the [C++ solver dummy](https://github.com/precice/precice/tree/develop/tools/solverdummies/cpp). All these problems could be reproduced by @BenjaminRueth if not specified otherwise.

COMPILING

- The `-o` is missing in the compilation command provided
- Compiling does not seem to work with the command provided: The precice library can't be found, even if `LD_LIBRARY_PATH` is set correctly. It only works when `-L[LIBRARYPATH]` is added before -lprecice

RUNNING

- In the current config file, the default communication should be switched to sockets, since MPI does not work with OpenMPI.
- The command `mpiexec -n 2 ./solverdummy precice-config.xml X` where X=A resp. X=B throws an error complaining that B_Mesh is wrong, probably empty
- The command `mpiexec -n 2 ./solverdummy precice-config.xml X X_Mesh Datax Datay 1 1` where (X,x,y)=(A,1,2) resp. (X,x,y)=(B,2,1) does not throw errors, but for some reason, only the A solver performs proper time stepping, while B gets stuck after the first step. This problem happened to me but could not be reproduced by @BenjaminRueth

BenjaminRueth:
The cpp solver dummy now has the same functionality as the python, c and fortran counterparts. Bugs mentioned in #239 are resolved. Documentation is updated.

As soon as this PR is merged, this closes #239.


BenjaminRueth:
As soon as this PR is merged, this closes #239

PR #289, opened 2019-02-16
Fix petsc building with 64 bit indices

fsimonis:
Build PETSc with 64-bit indices to prevent tests from failing due to overflow.

This already happened to me and leads to crashes with messages along the lines of _"out-of-memory (0 Bytes allocated)"_ and _"Segmentation fault at address 0x0000010"_.


floli:
Does it occurs at Travis? Note that the file is meant to only for building the dependencies for Travis.

But either way, I'm ok with it.

fsimonis:
Is hasn't occurred _yet_ :wink:

PR #292, opened 2019-02-21
Replace use_count and size() == 0

floli:
* Replace use_count() > 0 by operator bool usage. 
* Replace size() == 0 with .empty(), where is makes sense. Can be faster and is semantically clearer.


uekerman:
Nice! Good to merge in my opinion. 
Btw, one decision from the last coding days was to always mention in the PR what you expect from the reviewer, e.g. "please" ... "merge if you agree", "read briefly over it", "check for correctness", "fyi", ...

PR #297, opened 2019-02-26
Hashed connection info files

floli:
+ Writes and reads the connection info using the hash based method
+ All connection files a writen to `addressDirectory/.precice/`
+ For all connection methods
+ Publisher code is removed
+ Compares with the info read by the `utils::Publisher`, uses the address from the hashed file to connect (not anymore, since the `utils::Publisher` was removed)
+ Seems to work so far
+ Works for the 124cores fence test case on SuperMUC.
+ Works for ASTE up to 560 cores on SuperMUC
+ Unit tests pass.
+ All system test pass.

What is left to do, but **not** in this PR is to also take the mesh name into account, s.t. we do not rely on the correct order of connections attempts.

Idea: We could not only write out the connection information, but also from which information the hash was created, i.e., the 3-tuple of `(acceptorName, requesterName, rank)` as a second line in the file. This would make debugging a bit easier. Comments?

Reviewers: Please review and comment, I will do the merging.

## TODO
- [x] Write out textual connection info into files
- [x] Make connection info directory non-hidden.
- [x] Use `precice-run` directory instead of `.precice`


fsimonis:
A few questions, remarks:
1) What is the point of the `.precice` directory?
2) Can you please "unhide" the directories? There is no need for it. It also makes them easier to navigate and remove.
3) Can you please post the directory `tree` of a run? It would be interesting to see the result.

floli:
> A few questions, remarks:
> 
>     1. What is the point of the `.precice` directory?

Storing the subdirectories which contains the address files.

>     2. Can you please "unhide" the directories? There is no need for it. It also makes them easier to navigate and remove.

I am open to discuss that. Just oriented myself at the address files, which were also hidden.
 
>     3. Can you please post the directory `tree` of a run? It would be interesting to see the result.

```
.precice
├── 0b
│   └── 298e44cffb55c3b8926ca172338e738e73
├── 11
│   └── 874f94cf10517085be70484a727d0f7d0f
├── 23
│   └── 16b6e5b7055c74864e89030ce9b5f0b5f0
├── 34
│   └── a5bfc33f5156be94931948f6cc2bb92bb9
├── 3e
│   └── 7f05eb02d657cca803124e01f0d8a0d8a0
├── 49
│   └── e398c184f953bab9652e6ab119971a971a
├── 58
│   └── 6a66747ebc59cfb0e8178d6e21dc15dc15
├── 81
│   └── 0d1a187125521e95638c675a5449384938
├── 86
│   └── 04d3b974395013b1d4726e4daf6f356f35
├── a4
│   └── 039a3d81df515b95db7a94580f23df23df
├── ad
│   └── f28e2d7d4f59e4a130d239ce1741ac41ac
├── b0
│   └── 3c07d18b48503b826548f6c6339def9def
├── b5
│   └── badcb978f85f7daab6b67cef2217311731
├── b6
│   └── 305e9a2be2526eb1f8cd4b4eb233d533d5
├── ba
│   └── f5607826fe5a77ad88c91ddd97b61cb61c
├── c7
│   └── dafada3737581d96bc4f789c6106a306a3
├── d8
│   └── 534ea0f3d75604a4e956941f9fb224b224
├── db
│   └── 1be2fce10f53989513a87e884196ba96ba
└── df
    └── f527a9d70859669e252475da400d010d01
```

src/com/Communication.cpp
```diff
@@ -286,5 +294,70 @@ void Communication::broadcast(std::vector<double> &v, int rankBroadcaster)
...
 }
 
+void Communication::writeConnectionInfo(std::string const & acceptorName,
+                                        std::string const & requesterName,
+                                        int rank,
+                                        std::string addressDirectory,
+                                        std::string addressData)
+{
+  using namespace boost::filesystem;
+  int const firstLevelLen = 2;
```
fsimonis:
```suggestion
  constexpr int firstLevelLen = 2;
```

This should probably be defined in the `Communication` class itself as a `static constexpr int`.

fsimonis:
> Storing the subdirectories which contains the address files

I don't understand why we need the two levels `addressDirectory/.precice`. Isn't one enough?

src/com/Communication.cpp
```diff
@@ -286,5 +294,70 @@ void Communication::broadcast(std::vector<double> &v, int rankBroadcaster)
...
+
+  boost::uuids::string_generator ns_gen;
+  auto ns = ns_gen("af7ce8f2-a9ee-46cb-38ee-71c318aa3580"); // md5 hash of precice.org as namespace
+  
+  boost::uuids::name_generator gen{ns};
+  std::string s = acceptorName + requesterName + std::to_string(rank);
```
fsimonis: As @uekerman stated on the mailing list: To make this future proof, we need the name of the mesh in here too.

src/com/Communication.cpp
```diff
@@ -286,5 +294,70 @@ void Communication::broadcast(std::vector<double> &v, int rankBroadcaster)
...
+  std::string hash = boost::uuids::to_string(gen(s));
+  std::remove(hash.begin(), hash.end(), '-');
```
fsimonis:
```suggestion
  hash.erase(std::remove(hash.begin(), hash.end(), '-'), hash.end());
```

src/com/Communication.cpp
floli: And as I replied on the mailing list, this needs some more invasive changes. Because of that, I do first things first (replace the current system) and fix potential, upcoming problems after that.

src/com/Communication.cpp
floli: Why?

floli:
> > Storing the subdirectories which contains the address files
> 
> I don't understand why we need the two levels `addressDirectory/.precice`. Isn't one enough?

addressDirectory is usually set to something like `.` or `..`, i.e, the case root and I don't think we want to create the directories `0d`, `be`, ... directly in the root of the case.

Please note, that this code is a proof of concept. I honor and value your suggestions, but will wait for finishing until it has proven to work.

src/com/Communication.cpp
floli: Ok with `constexpr`, but I don't like making things `static` for no real reason.

src/com/Communication.cpp
fsimonis: @floli This is just for the record as this is an actual PR.

fsimonis:
> addressDirectory is usually set to something like `.` or `..`, i.e, the case root and I don't think we want to create the directories `0d`, `be`, ... directly in the root of the case.

Got it! I didn't realize that this directory is provided externally. _Reading helps..._

This change makes the entire `ScopedPublisher` and `SuperMUC_WORK` contraption obsolete, correct?

> Please note, that this code is a proof of concept. I honor and value your suggestions, but will wait for finishing until it has proven to work.

It's a PR; I won't push, but I will comment just like Mr Travis :wink:

fsimonis:
Looks like a neatly abstracted and very clean solution!

src/com/SocketCommunication.cpp
```diff
@@ -199,21 +183,14 @@ void SocketCommunication::requestConnection(std::string const &acceptorName,
   TRACE(acceptorName, requesterName);
   assertion(not isConnected());
 
-  std::string address;
-  const std::string addressFileName("." + requesterName + "-" + acceptorName + ".address");
+  ConnectionInfoReader conInfo(acceptorName, requesterName, _addressDirectory);
+  std::string address = conInfo.read();
```
fsimonis:
```suggestion
const std::string address = conInfo.read();
```
Could even be `const auto`

src/com/SocketCommunication.cpp
```diff
@@ -199,21 +183,14 @@ void SocketCommunication::requestConnection(std::string const &acceptorName,
...
+  DEBUG("Request connection to " << address);
+  std::string ipAddress  = address.substr(0, address.find(":"));
```
fsimonis: Please use the (`noexcept`) `char` version of `std::string::find` and refactor out the index to prevent the following line from looking so Frankensteiny.
Feel free to implement some sort of `pair<string, string> splitAt(string, index)` if you feel so inclined.

_Note: Isn't it possible to create a `boost::asio`endpoint straight from a string?_

src/com/SocketCommunication.cpp
```diff
@@ -199,21 +183,14 @@ void SocketCommunication::requestConnection(std::string const &acceptorName,
...
+  std::string portNumber = address.substr(ipAddress.length() + 1, address.length() - ipAddress.length() - 1);
+  _portNumber = static_cast<unsigned short>(std::stoi(portNumber));
```
fsimonis: As the port is unsigned, so lets convert the string to an unsigned straight away.
```suggestion
  _portNumber = static_cast<unsigned short>(std::stoul(portNumber));
```

This could potentially fail in case the file contains rubbish.
Maybe this entire parsing belongs into a `try` block?

src/com/SocketCommunication.cpp
```diff
@@ -267,20 +243,14 @@ void SocketCommunication::requestConnectionAsClient(std::string      const &acce
   
   for (auto const & acceptorRank : acceptorRanks) {
     _isConnected = false;
-    std::string address;
-    const std::string addressFileName("." + requesterName + "-" +
-                                      acceptorName + "-" + std::to_string(acceptorRank) + ".address");
+    ConnectionInfoReader conInfo(acceptorName, requesterName, acceptorRank, _addressDirectory);
+    std::string address = conInfo.read();
```
fsimonis: Same here, therefore a stand-alone function may be a good idea to keep the code dry.

src/utils/SignalHandler.cpp
```diff
@@ -12,20 +11,10 @@ void terminationSignalHandler(int signal)
...
-    std::string file = fs::path(dir->path()).filename().string();
-    if (std::regex_match(file, addrFile)) {
-      fs::remove(dir->path());
-    }
-  }
-
-  std::exit(-1);
+  // Remove connection info files. This is just a guess and will only
+  // work if the address directory is ".".
+  boost::filesystem::remove_all(".precice");
```
fsimonis: Isn't this a bit risky? It is probably a better idea to simply ignore everything.
It is quite simple to `rm` a single directory.

src/com/Communication.cpp
```diff
@@ -1,5 +1,13 @@
 #include "Communication.hpp"
 #include "Request.hpp"
+#include <thread>
```
fsimonis: Please remove unnecessary includes.

src/com/ConnectionInfoPublisher.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include <boost/uuid/string_generator.hpp>
+#include <boost/uuid/uuid_io.hpp>
+#include <boost/filesystem.hpp>
+
+ConnectionInfoPublisher::ConnectionInfoPublisher(std::string acceptorName,
+                                                 std::string requesterName,
+                                                 int rank,
+                                                 std::string addressDirectory)
+  :
+  acceptorName(acceptorName),
```
fsimonis: You should `std::move` the strings straight into the members in this an the following constructor.

```cpp
acceptorName(std::move(acceptorName)),
```

Also consider delegating the strings, part to the following constructor.

src/com/ConnectionInfoPublisher.cpp
```diff
@@ -0,0 +1,93 @@
...
+#include <thread>
+#include <boost/uuid/name_generator.hpp>
+#include <boost/uuid/string_generator.hpp>
+#include <boost/uuid/uuid_io.hpp>
+#include <boost/filesystem.hpp>
+
+ConnectionInfoPublisher::ConnectionInfoPublisher(std::string acceptorName,
+                                                 std::string requesterName,
+                                                 int rank,
+                                                 std::string addressDirectory)
```
fsimonis: Move-constructing strings is `noexcept`.

```suggestion
                                                 std::string addressDirectory) noexcept
```

src/com/ConnectionInfoPublisher.cpp
```diff
@@ -0,0 +1,93 @@
...
+  acceptorName(acceptorName),
+  requesterName(requesterName),
+  rank(rank),
+  addressDirectory(addressDirectory)
+{}
+  
+
+ConnectionInfoPublisher::ConnectionInfoPublisher(std::string acceptorName,
+                                                 std::string requesterName,
+                                                 std::string addressDirectory)
```
fsimonis: Same here
```suggestion
                                                 std::string addressDirectory) noexcept
```

src/com/ConnectionInfoPublisher.cpp
```diff
@@ -0,0 +1,93 @@
...
+  :
+  acceptorName(acceptorName),
+  requesterName(requesterName),
+  addressDirectory(addressDirectory)
+{}
+
+
+std::string ConnectionInfoPublisher::getFilename()
```
fsimonis: I have a few comments here:

1) Doesn't this function return a relative path to the file? (misleading name)
2) Is there a reason why this function doesn't return a `boost::filesystem::path` instead of a `string`?
3) All dependencies of this function are const members, so this can be calculated in the constructor.
4) This is operation is semantically `const`
```suggestion
std::string ConnectionInfoPublisher::getFilename() const
```

src/com/ConnectionInfoPublisher.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+  using namespace boost::filesystem;
+  
+  constexpr int firstLevelLen = 2;
+  boost::uuids::string_generator ns_gen;
+  auto ns = ns_gen("af7ce8f2-a9ee-46cb-38ee-71c318aa3580"); // md5 hash of precice.org as namespace
+  
+  boost::uuids::name_generator gen{ns};
+  std::string s = acceptorName + requesterName + std::to_string(rank);
```
fsimonis:
```suggestion
const  std::string s = acceptorName + requesterName + std::to_string(rank);
```

src/com/ConnectionInfoPublisher.cpp
```diff
@@ -0,0 +1,93 @@
...
+  std::string hash = boost::uuids::to_string(gen(s));
+  hash.erase(std::remove(hash.begin(), hash.end(), '-'), hash.end());
+  
+  path p = path(addressDirectory)
+    / path(".precice")
```
fsimonis: I am still not happy about the hidden directory.
What is the rational to hide it?

src/com/ConnectionInfoPublisher.cpp
```diff
@@ -0,0 +1,93 @@
...
+    / path(hash.substr(0, firstLevelLen))
+    / hash.substr(firstLevelLen);
+  
+  return p.string();  
+}
+
+
+std::string ConnectionInfoReader::read()
```
fsimonis: As strange as it sounds, this call is semantically `const` wrt the object.
```suggestion
std::string ConnectionInfoReader::read() const
```

Considering the name of this class, this should probably also parse the the endpoint and return the ip-port pair.

src/com/ConnectionInfoPublisher.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+  std::string addressData;
```
fsimonis: Please move this under the loop.

src/com/ConnectionInfoPublisher.cpp
```diff
@@ -0,0 +1,93 @@
...
+    ifs.open(path, std::ifstream::in);
+    std::this_thread::sleep_for(std::chrono::milliseconds(1));
+  } while (not ifs);
+
+  ifs >> addressData;
+  return addressData;
+}
+
+
+ConnectionInfoWriter::~ConnectionInfoWriter()
```
fsimonis: It would be great if we could get this destructor `noexcept` for the sake of stability.

src/com/ConnectionInfoPublisher.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+  namespace fs = boost::filesystem;
+  fs::path p(getFilename());
+  boost::filesystem::remove(getFilename());
```
fsimonis: Please use `exists(p)` to prevent ugly errors, especially as you do not create this file in the constructor.
Also move it into the try block to prevent real exceptions.

src/com/ConnectionInfoPublisher.cpp
```diff
@@ -0,0 +1,93 @@
...
+  try {
+    fs::remove(p.parent_path()); // try to also remove parent dir, i.e, first part of hash
```
fsimonis: Please use `exists` and `is_empty` to prevent exceptions here an in the following call.

src/com/ConnectionInfoPublisher.cpp
```diff
@@ -0,0 +1,93 @@
...
+    fs::remove(p.parent_path().parent_path()); // and also the .precice, only if empty
+  }
+  catch (fs::filesystem_error) {
```
fsimonis: Always catch by const ref
```suggestion
  catch (const fs::filesystem_error&) {
```

src/com/ConnectionInfoPublisher.cpp
```diff
@@ -0,0 +1,93 @@
...
+    // directory not empty, no problem
```
fsimonis: Now, we could log any filesystem exceptions and make the dtor `noexcept`

src/com/ConnectionInfoPublisher.cpp
```diff
@@ -0,0 +1,93 @@
...
+  }
+}
+
+
+void ConnectionInfoWriter::write(std::string const & info)
```
fsimonis: 1) Considering the class name, this function should probably take a `boost::asio` endpoint.
2) This function is semantically `const`
```suggestion
void ConnectionInfoWriter::write(std::string const & info) const
```

src/com/ConnectionInfoPublisher.hpp
```diff
@@ -0,0 +1,59 @@
+#include <string>
+
+class ConnectionInfoPublisher
+{
+public:
+  
+  ConnectionInfoPublisher(std::string acceptorName,
+                          std::string requesterName,
+                          int rank,
+                          std::string addressDirectory);
```
fsimonis:
```suggestion
                          std::string addressDirectory) noexcept;
```

src/com/ConnectionInfoPublisher.hpp
```diff
@@ -0,0 +1,59 @@
...
+
+  ConnectionInfoPublisher(std::string acceptorName,
+                          std::string requesterName,
+                          std::string addressDirectory);
```
fsimonis:
```suggestion
                          std::string addressDirectory) noexcept;
```

src/com/ConnectionInfoPublisher.cpp
floli: The connection info is not always an IP-Port pair, e.g. in case of MPI ports.

src/com/ConnectionInfoPublisher.cpp
floli: Connection info are not only about socket communication.

src/com/SocketCommunication.cpp
floli: Ok, about const. Don't like the `auto` here, using `string` makes it more clear to the developer what the actual type is.

src/com/SocketCommunication.cpp
floli: Sorry, I don't see anything wrong with that line.

src/utils/SignalHandler.cpp
floli: We had it with the old connection files and it was added due to explicit user request. Problem is, that when you forget to `rm`, the connection will likely fail.

src/com/ConnectionInfoPublisher.cpp
floli: > All dependencies of this function are const members, so this can be calculated in the constructor.

It's called in both overloads of the constructor, so this will result in code duplication.

src/com/ConnectionInfoPublisher.cpp
floli: Sorry, these type of `const`s in a function of less 15 lines or so make zero sense for me. There is no performance gained, it only adds to human parsing complexity when reading the code.

src/com/ConnectionInfoPublisher.cpp
floli: We can talk about that in the telco, as I said, I looked at how it was done previously.

src/com/ConnectionInfoPublisher.cpp
floli: I could move it into the try-block. Adding an extra `exists` adds no value.

src/com/ConnectionInfoPublisher.cpp
floli: I do not even catch a variable, just a type.

src/com/ConnectionInfoPublisher.cpp
floli: There is no use in logging them, a non-empty directory is a perfect use-case and does not even warrant a INFO.

floli:
@fsimonis I added most of your suggestions, but some are only for the sake of shineness and add no real value imo.

src/com/SocketCommunication.cpp
fsimonis: What about now?
```cpp
const auto sepidx = address.find(':');                     // guaranteed not to throw
const std::string ipAddress  = address.substr(0, sepids);  // shorter
const std::string portNumber = address.substr(sepidx + 1); // no need to specify the end
```

src/com/ConnectionInfoPublisher.cpp
fsimonis: True in this case.

Yes, `const` adds to the complexity of humans parsing the code, but it reduces the complexity of understanding the code and prevents others from breaking your code in the future.
That's why I am trying to push it as much as possible.

src/com/ConnectionInfoPublisher.cpp
fsimonis: Fine for me, especially as this is trivial to change later.

src/com/ConnectionInfoPublisher.cpp
fsimonis: > Adding an extra exists adds no value.

:+1:  I wasn't aware that removing a non-existent path is legal.

src/com/ConnectionInfoPublisher.cpp
fsimonis: > I do not even catch a variable, just a type.

You catch an unnamed exception by value.

src/com/ConnectionInfoPublisher.cpp
fsimonis: If you only remove empty directories by checking `is_empty`, then the caught exceptions are the remaining nasty ones.

Those exceptions may be worth logging.

src/com/ConnectionInfoPublisher.cpp
fsimonis: Sorry, I don't understand what you mean. In the current state, it is called in the destructor as well as the read, and write.
It also doesn't modify the object, so there is no reason not make it `const`.

What I actually meant with point 3 is the following:

By design, this class cannot change its members after construction.
This function depends only on the members of the class.
Thus, once constructed, this member function will always return the same value.
Meaning, you can simply generate this value one in the ctor and use it later.

fsimonis:
@floli I explained my reasoning concerning the destructor and a few other points in the comments.

I am not trying to be a pain, but something about this design bugs me.  
Using a class with only protected const members means that it has no invariant, thus every function without side-effects can be modelled as a pure function.
Maybe all of this could really just be a struct and a few free-standing functions.
In Haskell it could look like this:
```hs
generateFileInfo :: ConnectionInfo -> FileInfo
write :: (FileInfo, Info) -> IO()
read :: FileInfo -> IO(Info)
```
But that's just rambling. This PR removes a _crazy_ special case in the entire build of the library and creates a platform independent solution, so I am very happy to get this in ASAP.

src/com/SocketCommunication.cpp
floli: I have adopted the change to `stoul`. Regarding the try block: If the file contains rubbish, there is no way we can continue in any sensible way, so we need to fail sooner or later, as the error can not be handled gracefully. Furthermore, the files are written be precice and read by precice, if there is rubbish in there, there is a problem with the underlying system, which can produce **any** kind of error, ranging from segfaults to formatting of your hard disk.

floli:
@fsimonis I have tried to adopt the main points of your suggestions, given they were easy to implement and I concurred with them.

I am open to a later rewrite of the design (after all, it's only a handful of lines), right now, I happily agree with you in saying "this PR removes a crazy special case in the entire build of the library and creates a platform independent solution, so I am very happy to get this in ASAP."

If you don't see any major design flaws or bugs, I would ask you to approve the PR.

fsimonis:
@floli I have 2 last requests for this PR:
1) Please use `std::move` in the constructor and mark it `noexcept`.
2) Please use `fs:is_empty` to prevent the `fs::remove` from throwing. If you need additional reasoning:
    1) You are currently using exceptions as a control flow mechanism for handling unexceptional circumstances.
    2) Think about how often this exception will be thrown and caught unnecessarily.

floli:
>     1. Please use `std::move` in the constructor and mark it `noexcept`.

Done
 
>     2. Please use `fs:is_empty` to prevent the `fs::remove` from throwing.

Ok. What do think about the `try` block now? Should it be left in place and a warning is issued? Failure to remove the file is not a problem for preCICE, but may indicate some filesystem trouble (let alone be it delayed synchronization in distributed filesystems)

> You are currently using exceptions as a control flow mechanism for handling unexceptional circumstances.

Yes, I know. I saw nothing wrong with that, as it is a common paradigm in Python. I read up on C++ exception and learned that they come with a performance penalty when thrown.

fsimonis:
@floli One last request:
Can you please reorder the protected members so that they match the initialisation order? (or vice versa)

floli:
Can do that, first I need to find out why Travis with gcc (not clang) breaks at the moment.

BTW: The warnings about initialization order are only emitted when compiling with scons, cmake seems do not have the warnings activated.

fsimonis:
> I like the idea with the connection information in a second line. Will be handy for debugging. If it's easy to add please do so!

@uekerman As @floli and I discussed earlier: the logs should include all the necessary information.
The files are for low level exchange of information only. Adding debug information doesn't come at a noticeable benefit. increases their size and complicates parsing.

I think we should invest time in improving the log output for the communication modules instead.

floli:
> I know we always had the address files hidden and actually I never really questioned it. However, now that we have the discussion on the table, I vote for making them visible. 

I am cool with making the directory non-hidden. Explicit is better than implicit, visible is better than hidden, honey is better than jam.
What are your ideas for the name of directory. Since it will evatually contain other files as well, I think we should not name it `precice-connection-files`, but something more general, like `precice-state` or `precice-run` (the latter resembles the `/var/run` directory. 

> I like the idea with the connection information in a second line. Will be handy for debugging. If it's easy to add please do so!

I also think we should add it. Parsing is simple, as I just put in the second line and ignore the second line when reading in the files. Size of files is not the problem, it was and is the number of files.

uekerman:
> What are your ideas for the name of directory. Since it will evatually contain other files as well, I think we should not name it `precice-connection-files`, but something more general, like `precice-state` or `precice-run` (the latter resembles the `/var/run` directory.
>

 `precice-run` sounds good

src/com/ConnectionInfoPublisher.hpp
```diff
@@ -0,0 +1,115 @@
+#include <string>
```
fsimonis: The alerts from lgtm are not useless:

```suggestion
#pragma once

#include <string>
```

floli:
This PR is basically ready to merge.

> I know we always had the address files hidden and actually I never really questioned it. However, now that we have the discussion on the table, I vote for making them visible. Think about it. If everything goes right the user will never see the address files. But if sth goes wrong and then she sees that there is a new folder that normally is not there, she might think: "mmh, something is odd, maybe it has sth to do with this folder, let's search for this in the trouble-shooting section" and baam, we have one case to debug less! With a hidden folder the information that sth went wrong is also hidden.

Right now, the `./precice-run/[hash]/` directories are not deleted, but the contained files are. There can be multiple files from different ranks in the directories. Experiments shown there is a race condition when trying to delete the dirs as well.

We could add a `boost::filesystem::remove_all("precice-run");` somewhere in the d'tor of solver interface, like it currently is in `terminationSignalHandler`.

This is only a cosmetic issue, the leftover dirs contain no information, they can be deleted after the run or not, doesn't matter.

Issue #296, opened 2019-02-25
Failing Python bindings


MakisH:
We have observed the following error, both on my laptop, @fsimonis laptop, and on Travis, which seems to be originating from our latest changes and must be (Python+MPI)-specific.

```
$ python3 solverdummy.py precice-config.xml SolverOne MeshOne

preCICE: This is preCICE version 1.3.0
preCICE: Configuring preCICE with configuration: "precice-config.xml"
preCICE: Run in coupling mode
[Magdalena:08291] mca: base: component_find: unable to open /usr/lib/openmpi/lib/openmpi/mca_shmem_posix: /usr/lib/openmpi/lib/openmpi/mca_shmem_posix.so: undefined symbol: opal_shmem_base_framework (ignored)
[Magdalena:08291] mca: base: component_find: unable to open /usr/lib/openmpi/lib/openmpi/mca_shmem_sysv: /usr/lib/openmpi/lib/openmpi/mca_shmem_sysv.so: undefined symbol: opal_show_help (ignored)
[Magdalena:08291] mca: base: component_find: unable to open /usr/lib/openmpi/lib/openmpi/mca_shmem_mmap: /usr/lib/openmpi/lib/openmpi/mca_shmem_mmap.so: undefined symbol: opal_show_help (ignored)
--------------------------------------------------------------------------
It looks like opal_init failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during opal_init; some of which are due to configuration or
environment problems.  This failure appears to be an internal failure;
here's some additional information (which may only be relevant to an
Open MPI developer):

  opal_shmem_base_select failed
  --> Returned value -1 instead of OPAL_SUCCESS
--------------------------------------------------------------------------
--------------------------------------------------------------------------
It looks like orte_init failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during orte_init; some of which are due to configuration or
environment problems.  This failure appears to be an internal failure;
here's some additional information (which may only be relevant to an
Open MPI developer):

  opal_init failed
  --> Returned value Error (-1) instead of ORTE_SUCCESS
--------------------------------------------------------------------------
--------------------------------------------------------------------------
It looks like MPI_INIT failed for some reason; your parallel process is
likely to abort.  There are many reasons that a parallel process can
fail during MPI_INIT; some of which are due to configuration or environment
problems.  This failure appears to be an internal failure; here's some
additional information (which may only be relevant to an Open MPI
developer):

[...]

PR #298, opened 2019-02-26
Add systemtests indicator

### LINKED ISSUE #294: Add a system tests status indicator in README
We have this nice "build status" in the preCICE README, which shows if the building and unit/integration tests succeeded/failed.

Since we now have nightly runs of our system tests, let's add another button for the system tests. For example: "tests:passing". Or "system tests:passing", if space allows.

If space is too short, we can remove the licence indicator, which is static and not so useful.

shkodm:
Closes #294

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -6,6 +6,9 @@
 <a style="text-decoration: none" href="https://travis-ci.org/precice/precice" target="_blank">
     <img src="https://travis-ci.org/precice/precice.svg?branch=develop" alt="Build status">
 </a>
+<a style="text-decoration: none" href="https://travis-ci.org/precice/systemtests" target="_blank">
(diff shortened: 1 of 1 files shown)


MakisH:
Let's merge this when we fix the currently failining python bindings tests (#296).

fsimonis:
@shkodm Do you know why the status shield shows `failing` even though it isn't?

MakisH:
@fsimonis I think we need to specify the branch as well, otherwise it picks the latest build, which currently is a failing PR. But it doesn't make sense to specify the branch in the README... :-/ Check the image source on the respective button on the Travis page.

shkodm:
Right, it picks up latest build ( and in that case it was failing PR ).  I haven't though of it. We should point to a master branch, like: 
```
 https://img.shields.io/travis/precice/systemtests/master.svg?label=system%20tests&style=flat

PR #299, opened 2019-02-27
Fix missing mpi4py

fsimonis:


BenjaminRueth:
**How can we reproduce this error?**

If preCICE was compiled **with MPI** and the line https://github.com/precice/precice/blob/dca0c97bfd606cb15a8c82e3e183d04e642e8460/tools/solverdummies/python/solverdummy.py#L6 is missing, an error will be thrown.

If preCICE was compiled **without MPI** the line https://github.com/precice/precice/blob/dca0c97bfd606cb15a8c82e3e183d04e642e8460/tools/solverdummies/python/solverdummy.py#L6 can be spared. Everything works.

PR #302, opened 2019-02-28
Silence Travis warnings from OpenMPI

fsimonis:
Our Travis log is currently flooded with warnings from the OpenMPI headers.
This PR silences these warnings by appending `-Wno-literal-suffix` to the `CXXFLAGS`.

One example from [here](https://travis-ci.org/precice/precice/jobs/499187815#L791):
```
/usr/lib/openmpi/include/mpi_portable_platform.h:374:34: warning: invalid suffix on literal; C++11 requires a space between literal and string macro [-Wliteral-suffix]
              _STRINGIFY(__GNUC__)"."_STRINGIFY(__GNUC_MINOR__)"."_STRINGIFY(__GNUC_PATCHLEVEL__)
```

@floli please review and merge


fsimonis:
Now clang complains about the warning flag:
https://travis-ci.org/precice/precice/jobs/499840279#L781

```
warning: unknown warning option '-Wno-literal-suffix'; did you mean
      '-Wno-literal-range'? [-Wunknown-warning-option]
```

fsimonis:
This silences the warning in gcc but clang now complains about the warnflag being unknown.
So we would have to add the clang flag to disable the warning about unknown warn flags. However, this flag is unknown by gcc.

@shkodm Is there a way to append CXXFLAGs in case of a specific CXX?

This won't propagate into scons, but cmake will pick it up in the future.

MakisH:
Isn't this way enough?

https://github.com/precice/precice/blob/7ac582ace15e3d8f90386c54d4130402525bebd4/.travis.yml#L44

PR #304, opened 2019-03-01
Turn Python Bindings into proper packages

fsimonis:
The python bindings are now packages:

* `PySolverinterface` depends on `precice>=1.3.0`
* `precice` installs `cython` prior to setup
* Readmes explain how to install everything

The python solverdummy now has the following requirements-file:
```
precice
numpy
mpi4py
```

# Result

```
$ pip3 install --user src/precice/bindings/python
$ pip3 install --user src/precice/bindings/PySolverdummy
$ pip3 install --user -r tools/solverdummies/python/requirements.txt
$ // run solver dummies
```

Reviewers: please give feedback


MakisH:
@BenjaminRueth do these packages work with all the scenarios you usually encounter?

MakisH:
Something missing from this PR is that it actually drops Python 2. Correct?

PR #306, opened 2019-03-01
Generate ENV file to similify working with custom install prefixes

fsimonis:
This PR generates the file `precice.env` in the binary directory which sets environment variables to the location of the install prefix. It extends `LD_LIBRARY_PATH`, `LIBRARY_PATH`,  `PATH` and `CPATH`.

Scenario:
```console
$ cmake -DCMAKE_INSTALL_PREFIX=~/software/bugfree/precice/ $PRECICE_ROOT
$ make install
$ . precice.env
$ echo $CPATH
/home/precicefan/software/bugfree/precice/:
```
A user could then either paste the file content into the `.bashrc` or source it from there.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file. For future plans, see our [Roadmap](https://github.com/precice/precice/wiki/Roadmap).
 
 ## develop
+- Generate an env file that can be sourced after installing the library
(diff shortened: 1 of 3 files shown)


fsimonis:
Would it be an idea to generate 2 env files?

1) `precice-install.env` to load after installation
2) `precice-build.env` to load after building


The `precice-build.env` would load:

ENV | dirs to add
--- | ---
CPATH | src in the source and build directory
[LD_]LIBRARY_PATH | binary directory
PATH | binary directory

Issue #311, opened 2019-03-04
Add wasCompiledWithMPI to SolverInterface


BenjaminRueth:
We should provide an API function to be able to find out whether preCICE was compiled with MPI or not. This issue is related to #300 

**Why do we need this functionality?** As an example, the python interface of preCICE requires you to `import mpi4py`, if preCICE has been compiled with MPI (for example https://github.com/precice/precice/pull/299). If preCICE has been compiled without MPI, you do not need this import statement. 

Putting `import mpi4py` everywhere is the safe way to solve this problem. However, this also adds an unnecessary dependency for users that did not compile preCICE with MPI. Ideally, we can use the following lines in the future:

in `__init__.py` (is this the correct solution? See https://github.com/precice/precice/issues/300#issue-415085361)
```
if precice.was_compiled_with_MPI():
    import mpi4py
```

in `setup.py`
```
dependencies = [...]

if precice.was_compiled_with_MPI():
    dependencies.append = ["mpi4py"]

setup(...
      install_requires=dependencies,
      ...)

**Comment by BenjaminRueth** [comment]
For instructions, how to reproduce the error, see https://github.com/precice/precice/pull/299#issuecomment-469421697.

**Comment by uekerman** [comment]
I am not to sure whether we need this. Please keep in mind that maybe 1% of our users will compile preCICE without MPI. This is only needed if you have a closed source code that comes with a pre-compiled MPI which interferes with the MPI dependency of preCICE. I am not even sure if anybody uses this at the moment. And then this closed-source code needs to use the Python bindings. Adding `mpi4py` by default seems to be the easier solution. Any objections to this though? I vote for making this issue a `wontfix`.

**Comment by floli** [comment]
A possible implementation in `SolverInterface.hpp` is just like
```
bool SolverInterfaceImp::wasCompiledWithMPI 
{
    return not #PRECICE_NO_MPI
}
```
true? So next to no effort? @BenjaminRueth stated a use case for that and I concur with him. Especially on weird, but unfortunately not too uncommon systems (windows & proprietary solvers) installing a python library which in turn requires MPI might be some work.

uekerman:
Well, you also need to port this to all bindings, to write tests for it, write documentation ...
and I don't see the use case. Windows & proprietary solvers will not need the Python bindings. Let's please only do this if we really need it.

BenjaminRueth:
[...]

Issue #300, opened 2019-02-27
Automatically import mpi4py when importing the precice bindings


fsimonis:
We saw in #296 that the user has to initialize MPI via the `mpi4py` import.
Maybe it would be a good idea to automatically call this when a user imports the `precice` bindings.

This way, it _just works_.

@BenjaminRueth How can we achieve this? Maybe via the `__init__.py`?

fsimonis:
This depends on how preCICE was built, so it is the responsibility of the user to initialize MPI.

PR #312, opened 2019-03-04
Moving mpi4py into bindings

BenjaminRueth:
Adds `mpi4py` as dependency for all cases. This PR is related to #311 and #300

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,6 +29,7 @@ env:
     - LD_LIBRARY_PATH="$PETSC_DIR/$PETSC_ARCH/lib:$LOCAL_INSTALL/lib:$LD_LIBRARY_PATH"
     - LIBRARY_PATH="$PETSC_DIR/$PETSC_ARCH/lib:$LOCAL_INSTALL/lib:$LIBRARY_PATH"
     - PYTHONPATH="$PETSC_DIR/$PETSC_ARCH/lib:/usr/lib/python2.7:/usr/lib/python2.7/dist-packages:/usr/lib/python2.7/plat-x86_64-linux-gnu"
+    - CXXFLAGS="-Wall -Wextra -Wno-unused-parameter"
 
   matrix:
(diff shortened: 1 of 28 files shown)


BenjaminRueth:
We should first merge #304

fsimonis:
I am still unsure if this is the right way to go. This will require the user to install mpi even if precice does not use it.
It is less work for us, but questionable from the user perspective.

MakisH:
> This will require the user to install mpi even if precice does not use it.

This seems to be enough for a "not the right way to go". What would be the advantages?

uekerman:
See comments in https://github.com/precice/precice/issues/311
I don't think that there is a user that builds precice without mpi, but uses the python bindings. Which solver would that be? All this can be solved yes, I just think that there are just more important work packages we should spend our time with. If a user complains at some point we can still add the API function. But if there is a majority who sees this differently, I am happy to be overruled :)

fsimonis:
I see three solutions:
1) Always init via mpi4py in our bindings, forcing MPI on the user.
2) Conditionally init via mpi4py by probing the preCICE API. This leads to a weird "maybe" dependency of mpi4py and forces us to extend the preCICE API.
2) Ask the user to import it if he compiled preCICE with MPI which will lead to problems.

Maybe we should tie this decision to the fact that we (aim to) provide packages with MPI compiled by default. Thus it make sense to stay consistent with that and put the MPI init into the bindings.

The aim should be that the following would be possible at some time:
```
$ sudo apt-get install precice
$ pip3 install --user precice
```

@uekerman Is there actually _a_ user using preCICE without mpi?

BenjaminRueth:
This decision is related to the build process of the python bindings, therefore let's postpone the decision to a future release.

BenjaminRueth:
I created an issue https://github.com/precice/python-bindings/issues/8 in our new repository for the bindings. Let's try to find an appropriate solution there.

PR #315, opened 2019-03-05
Removed disabled and no-longer-supported tests in com, closes #187

### LINKED ISSUE #187: SendReceiveFourProcessesV2 fails on Ubuntu 18.04
Currently `CommunicationTests/Socket/SendReceiveFourProcessesV2` fails

Behavior is random. Sometimes deadlock, sometimes ...
```
(0) 13:02:01 [com::SocketCommunication]:123 in acceptConnection: ERROR: Accepting connection at 127.0.0.1:55167 failed: boost::filesystem::rename: No such file or directory: "./.B-A.address~", "./.B-A.address"
```

* It fails independent of cmake or scons.
* The integration tests work.
* Also the tests in `M2N` work
* Also, Alya with sockets works.
* It is not related to any preCICE commit. (except the one that added the test)
* It currently works on Ubuntu 16.04  .

uekerman:
* Disabling tests is not enough because they are still executed if specified explicitly, e.g. when running all tests with the `MPI_Ports` flag. 
* We don't support the tested functionality any longer and I currently don't see a situation where we would need that again any time soon.  
Thus, simplest solution: deleting them.


floli:
>  Disabling tests is not enough because they are still executed if specified explicitly, e.g. when running all tests with the MPI_Ports flag.

Just BTW, you can unconditionally deactivate tests using the `Deleted` fixture.

>  We don't support the tested functionality any longer and I currently don't see a situation where we would need that again any time soon. Thus, simplest solution: deleting them.

As the one who disabled the tests, I am ok with removing them.

PR #316, opened 2019-03-05
Update solverdummies to use CMake/pip

BenjaminRueth:
We want to use CMake/pip as the default build system for all solverdummies

- [x] cpp
- [x] c
- [x] python #304
- [x] fortran
- [x] f2003


MakisH:
The Fortran 2003 is omited for now, because we don't support static linking with pkg-config yet. For the Fortran 90, I simply added a Makefile.

MakisH:
Update for the Fortran solverdummies and Fortran bindings:

* Fortran 90: the [bindings](https://github.com/precice/precice/tree/develop/src/precice/bindings/fortran) (which are actually C++ files) are not installed in the `include/` directory, but aparently they are not needed there, as they are not included by the solverdummy (Fortran does not use header files). Linking to the library seems to be enough (or I am doing something terribly wrong). @uekerman @floli is this correct? But I think this is good to go for now.
* Fortran 2003: here we actually need to include the files `ConstantsF2003.f90` and `SolverInterfaceF2003.f90` (don't ask me why `.f90` while they are meant for Fortran2003), which are not installed in the `include`, so I cannot easily get them using `pkg-config`. @fsimonis shouldn't we include these as well?

MakisH:
Anyway, both cases are good for now and we can definitely improve them in future releases.

uekerman:
>     * Fortran 90: the [bindings](https://github.com/precice/precice/tree/develop/src/precice/bindings/fortran) (which are actually C++ files) are not installed in the `include/` directory, but aparently they are not needed there, as they are not included by the solverdummy (Fortran does not use header files). Linking to the library seems to be enough (or I am doing something terribly wrong). @uekerman @floli is this correct? But I think this is good to go for now.

No, that's how Fortran works as far as I know. Also for Alya, we do not need the headers.

Issue #293, opened 2019-02-21
Make tagMesh*Round use boost::geometry


floli:
Right now, `tageMeshFirstRound` and `tagMeshSecondRound` use a bounding box and then iterate through all vertices to check if they are in the bounding book. This is a use case `boost::geometry` is made for.

* Maybe write an adapter to directory use `mesh::BoundingBox`
* Or create a `Box2d` or `Box3d` and enlarge by the support radius. Use the `within` predicate to get all vertices within the box.

uekerman:
@floli probably refers to the methods in `PetRadialBasisFctMapping`. For the other mapping methods, `tagMeshFirst/SecondRound` do different things. 
If there is a neat way to use `boost::geometry` here, there might be many other potential improvements in the `partition` package as well.

PR #322, opened 2019-03-07
Use a Rtree query for tagMesh*Round 

Summerdave:
This PR targets issue #293 by using boost::geometry for lookup.

### DIFF
--- a/src/mapping/PetRadialBasisFctMapping.hpp
+++ b/src/mapping/PetRadialBasisFctMapping.hpp
@@ -770,24 +770,32 @@ void PetRadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::tagMeshFirstRound()
       return; // Ranks not at the interface should never hold interface vertices
 
   // Tags all vertices that are inside otherMesh's bounding box, enlarged by the support radius
-  auto rtree = mesh::rtree::getVertexRTree(filterMesh);
-  mesh::Mesh::VertexContainer insideVerts;
-  if (not _basisFunction.hasCompactSupport()) {
-      for (auto& vert : filterMesh->vertices()){
-        vert.tag();
-      }
-  }
-  else {
-    #if PETSC_MAJOR >= 3 and PETSC_MINOR >= 8
+  if (
+#if PETSC_MAJOR >= 3 and PETSC_MINOR >= 8
+          _basisFunction.hasCompactSupport()
+#else
+          false
+#warning "Mesh filtering deactivated, due to PETSc version < 3.8. \
+          preCICE is fully functional, but performance for large cases is degraded."
+#endif
+     )
+  {
+      auto rtree = mesh::rtree::getVertexRTree(filterMesh);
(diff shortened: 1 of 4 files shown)


src/mapping/PetRadialBasisFctMapping.hpp
```diff
@@ -770,24 +770,25 @@ void PetRadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::tagMeshFirstRound()
...
+          }));
     #else
       #warning "Mesh filtering deactivated, due to PETSc version < 3.8. \
 preCICE is fully functional, but performance for large cases is degraded."
     #endif
-    if (isInside)
-      v.tag();
+  }
+  for (auto& vert : insideVerts){
+    vert.tag();
```
floli: Idea:

Move that loop (`for vert in insideVerts: vert.tag()`) into `if (not _basisFunction.hasCompactSupport)` and use the `make_function_output_iterator` to directly invoke `vert.tag()`. Got the idea?

src/mapping/PetRadialBasisFctMapping.hpp
```diff
@@ -825,18 +826,20 @@ void PetRadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::tagMeshSecondRound()
...
+    bb[d].second += _basisFunction.getSupportRadius();
+  }
+  auto rtree = mesh::rtree::getVertexRTree(mesh);
+  mesh::Mesh::VertexContainer insideVerts;
+  rtree->query(boost::geometry::index::within(bb),
+      boost::make_function_output_iterator([&mesh, &insideVerts](size_t idx){
+        insideVerts.push_back(&mesh->vertices()[idx]);
+      }));
+  for (auto& vert : insideVerts){
+    vert.tag();
```
floli: Same may be possible here.

src/mapping/PetRadialBasisFctMapping.hpp
Summerdave: Clever.

fsimonis:
@Summerdave Looks good.

Can you implement the tagging for the other mappings too? (The title of the PR suggests you change all of them.)

_For the record: I don't like the repetition of the type definition for Mesh::BoundingBox, but I get why you did it. Maybe we could add something like a `mesh_fwd.hpp` in the future?_

Summerdave:
> Can you implement the tagging for the other mappings too? (The title of the PR suggests you change all of them.)

The re-partitioning process for RBF mapping is based on @uekerman`s [thesis](https://mediatum.ub.tum.de/doc/1320661/document.pdf) (page 89). Right now, the partitioning for the other mappings works by computing a mapping and filtering this by whether the point _was used_. So for determining which points to tag a full mapping must be run anyway. @uekerman has also commented on this in #293.

Summerdave:
@floli Found a severe issue in the code which is now fixed. I also added unit tests to ensure `tagMeshFirstRound` and `tagMeshSecondRound` work correctly.

PR #323, opened 2019-03-08
WIP: Migrate travis to xenial

fsimonis:


fsimonis:
So we would have to build openmpi from source, correct?

fsimonis:
We concluded that it is not worth investing more time into this issue.

PR #325, opened 2019-03-11
WIP Parallel initialization, issue 142, work package 5

atotoun:


src/m2n/M2N.hpp
```diff
@@ -78,6 +78,12 @@ class M2N
   void requestSlavesPreConnection(const std::string &acceptorName,
                                   const std::string &requesterName);
 
+  /*
+   * @brief This function must be called by both accepter and requester to update 
+   *        the vertex list in _mappings
```
uekerman: This comment is a bit hard to understand I guess. `_mappings` is a data structure of `PointToPointCommunication`. Can you formulate it in a bit more abstract way. "After preliminary communication channels were set up and after the mesh partitions were communicated locally for every mesh, call this function to update and complete the communication channels for every communicated mesh" ?
Could we also give the function here a different name? Reflecting why it is used. "vertex list" is also not known on this level. I also don't like "update" as this sounds like we could call this function multiple times. "completeSlavesConnection"?

atotoun: I agree completely. I changed the name and updated the documentation.

PR #327, opened 2019-03-11
Add Travis coverage report using codecov

### LINKED ISSUE #307: Integrate a testing coverage tool
Since we have quite some tests in preCICE, it would be useful to also have a testing coverage tool to show us where tests are missing etc. It would also be a nice code quality metric which we can track over time.

A common tool that several projects use is [code coverage](https://codecov.io/) (e.g. see [Spack](https://codecov.io/gh/spack/spack)). Any other ideas/suggestions?

fsimonis:
This PR adds a Travis job which creates a coverage report using gcc and uploads it to codecov.io 

Closes #307

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -63,7 +63,8 @@ matrix:
       env: BUILD_TYPE=SCONS MPI=off PETSC=off
 
     # Test coverage
-    - compiler: gcc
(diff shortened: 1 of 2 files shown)


fsimonis:
Result of the run https://codecov.io/gh/precice/precice

PR #330, opened 2019-03-12
Extends Vector Assignment and Construction

fsimonis:
This PR adds standard constructors and assignments for `precice::utils::Vector`. These model deep copy and ownership transfer of the underlying resource.

Initial motivation for this PR was to [fix this alert by LGTM](https://lgtm.com/projects/g/precice/precice/snapshot/17e1be1c927e6b3cafa21f0d14f48b798947f258/files/src/utils/Petsc.cpp?sort=name&dir=ASC&mode=heatmap#xa02c15719d67faf9:1).

Merging this PR will:
* Add noexcept swap to Vector and a free-standing version as a customization point for the slt
* Add copy construction and copy assignment  
This models a deep copy of the underlying PETSc Vector
* Add move construction and move assignment  
This models ownership transfer of the underlying PETSc Vector
* Remove default argument from `Vector(Vector&, string)` to prevent it from being semantically compatible with copy construction.

_This PR does not add default construction!_


src/utils/Petsc.hpp
```diff
@@ -62,24 +62,40 @@ class Vector
...
+  Vector(Vector&& other);
+
+  /** Move assignement
+   * Destroys the current vector and takes ownership of the other vector.
+   */
+  Vector& operator=(Vector&& other);
 
   /// Duplicates type, row layout etc. (not values) of v.
-  Vector(Vector &v, std::string name = "");  
+  Vector(Vector &v, std::string name);
```
Summerdave: As the comment correctly states this does not (deep) copy the vector. Therefore it is not equivalent to the copy constructor. I would suggest:
```suggestion
  explicit Vector(Vector &v, std::string name = "");  
```

src/utils/Petsc.cpp
```diff
@@ -96,6 +97,33 @@ std::string getName(T obj)
...
+
+Vector::Vector(Vector&& other) {
+  vector = other.vector;
+  other.vector = nullptr;
+}
+
+Vector& Vector::operator=(Vector&& other)
+{
+  PetscErrorCode ierr = 0;
+  ierr = VecDestroy(&vector); [&]{ CHKERRV(ierr); }();
```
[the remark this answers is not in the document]
fsimonis: Its expansion includes an empty return statement which conflicts with the return statement of the function. The lambda eliminates this problem.

src/utils/Petsc.cpp
```diff
@@ -96,6 +97,33 @@ std::string getName(T obj)
...
+Vector::Vector(const Vector &v)
+{
+  PetscErrorCode ierr = 0;
+  ierr = VecDuplicate(v.vector, &vector); CHKERRV(ierr);
+  setName(vector, getName(v.vector));
+}
+
+Vector& Vector::operator=(Vector s)
+{
+    swap(s);
```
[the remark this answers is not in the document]
fsimonis: Cpp reference states that this is the recommend way for classes with swap capability.

Interesting point about the ambiguity though. I'll look into that.

src/utils/Petsc.hpp
fsimonis: Good idea!
We should mark the other ctors explicit too, as they all are candidates for implicit conversions. 

So, making this ctor explicit would remove the semantic compatability with the copy ctor.

However, this "alias" functionality clearly breaks the model of unique ownership wrt the underlying vector.
So, either we go all-in on shared ownership and make them all owners or we remove this alias functionality.

src/utils/Petsc.cpp
```diff
@@ -96,6 +97,33 @@ std::string getName(T obj)
 
 /////////////////////////////////////////////////////////////////////////
 
+Vector::Vector(const Vector &v)
+{
+  PetscErrorCode ierr = 0;
+  ierr = VecDuplicate(v.vector, &vector); CHKERRV(ierr);
```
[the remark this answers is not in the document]
fsimonis: Just noticed the all-caps warning in the doc of `VecDuplicate` :wink:

src/utils/Petsc.hpp
Summerdave: This is no alias. This is just saying "I have a vector here. Create a similar one for me". No pointers are ever copied (the [definition](https://github.com/precice/precice/blob/f71f6e4d0fe29811aaaee04f28d6385067be0277/src/utils/Petsc.cpp#L142)). 
I would also make this take a `const Vector& v` and cast away `const` inside the c`tor.

src/utils/Petsc.cpp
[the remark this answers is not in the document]
floli: Does `CHKERRQ` works`

src/utils/Petsc.hpp
fsimonis: Oh I get it now.

In that case, I argue that this syntax is too close to copy construction.

We could either
1) replace this functionality into a factory function  
`Vector::makeSimilar(Vector& other, string name)`, or
2) add a tag to the constructor  
`Vector(allocate_only_tag, Vector &v, std::string name)`.  
This makes it difficult to keep the constructors consistent in the long run.

I would definitely go for the first option as it make clear that we ask for allocation and not a copy.

src/utils/Petsc.cpp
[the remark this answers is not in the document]
Summerdave: Are you referring to the copy-swap idiom? In this case you wouldn't need `Vector& Vector::operator=(Vector&&)` at all.

src/utils/Petsc.cpp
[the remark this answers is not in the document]
fsimonis: @floli `CHKERRQ` also returns, which results in the same problem.

src/utils/Petsc.hpp
Summerdave: Sounds good.

fsimonis: The ctor is not covered by the [tests](https://codecov.io/gh/precice/precice/src/c428c7f3e791bd39fbdf1507ccc1d500e92c3fb4/src/utils/Petsc.cpp#L108...118) and compiling with `[[deprecated]]` did not raise a warning.

So I presume this is actually dead code.

src/utils/Petsc.hpp
```diff
@@ -62,27 +62,40 @@ class Vector
...
 
-  /// Delete copy and assignment constructor
-  /** Copying and assignement of this class would involve copying the pointer to
-      the PETSc object and finallly cause double destruction of it.
+  /** Creates an uninitialized vector of identical shape.
+   * Duplicates type, row layout etc. (not values) of v.
    */
-  Vector(const Vector&) = delete;
-  Vector& operator=(const Vector&) = delete;
+  Vector makeSimilar(const std::string& name = "") const;
```
Summerdave: This is not really uniform with the previous two methods. I would suggest:
```suggestion
  static Vector makeSimilar(const Vector& templ, const std::string& name = "");
  static Vector makeSimilar(const Matrix& templ, const std::string& name = "", LEFTRIGHT type = LEFT);
  static Vector makeSimilar(const Mat& templ, const std::string& name = "", LEFTRIGHT type = LEFT);
```
Another option would be to move
```
Vector(Matrix &m, std::string name = "", LEFTRIGHT type = LEFT);
```
to `Matrix` and require `Vector(Matrix(Mat))` for the other function.

fsimonis: These methods have different concerns:
* The `Vector` method allocates an identically shaped `Vector`
* The `Matrix` methods allocate a compatible `Vector`

The methods require an existing `Vector` and an existing `Matrix`/`Mat` to work with. Thus I would turn them into member functions of the respective dependency.

What do you PETSc guys think looks better/more natural?
```cpp
Vector v;
auto u = v.makeSimilar("u");
auto u = Vector::makeSimilar(v, "u");

Matrix a;
auto w = Vector::makeSimilar(a, Left);
auto w = a.makeCompatible(Left);
auto w = Matrix::makeCompatible(a, Left);
auto w = Vector::makeCompatible(a, Left);
```

Summerdave: > These methods have different concerns:

They both create a completely new `Vector`. 

Why not go with simple factory functions and distinguish by overloads only?
```suggestion

  static Vector Vector::instance(const std::string& name = "");
  static Vector Vector::instance(const Vector& templ, const std::string& name = "");
  static Vector Vector::instance(const Matrix& templ, const std::string& name = "", LEFTRIGHT type = LEFT);
  static Vector Vector::instance(const Mat& templ, const std::string& name = "", LEFTRIGHT type = LEFT);
```
When creating a `Vector` from another `Vector` one would assume they should be similar. When constructing from a `Matrix`/`Mat` one would assume it is compatible to that as well.
Alternatively you could declare them as `explicit` constructors.

fsimonis: > They both create a completely new Vector.

Ok, I can see that this is an argument why they should be in the Vector class.

The name instance is misleading though. Nouns usually denote properties and especially `instance` is a very common choice for the instance of a singleton.
These functions allocate a similar, or compatible Vector, so let's call them `allocate`.

It feels odd for the aforementioned reasons, but I'll put them in an overload set.

PR #331, opened 2019-03-12
Fix lgtm alerts

fsimonis:
This PR fixes the warnings issues by LGTM.

Most prominent was incomplete control-flow in release builds due to our assertion macro getting removed. Those should never be triggered due to the use of the scoped enum though.


floli:
`tools/plotEventLog.py` is deleted in develop.

PR #333, opened 2019-03-13
Modernize IO and std::endl

fsimonis:
This PR:
* Replaces `std::endl` with newline characters and merges them into adjacent strings is possible.
* Removes `io::TXTWriter`and `io::TXTReader` destructors as the destructor of the streams closes them anyhow.
* Adds `void TXTWriter::flush()`
* Fixes the Tests, which relied on a hidden flush instead of closing the stream by destroying the writer (or calling flush).

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -235,7 +235,7 @@ std::vector<int> BaseCouplingScheme::sendData(m2n::PtrM2N m2n)
   assertion(m2n.get() != nullptr);
   assertion(m2n->isConnected());
   for (const DataMap::value_type &pair : _sendData) {
-    //std::cout<<"\nsend data id="<<pair.first<<": "<<*(pair.second->values)<<std::endl;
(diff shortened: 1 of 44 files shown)


MakisH:
Why replace the `std::endl` with `\n`? Would then this also work on Windows?

fsimonis:
@floli The point of the exercise is that the flush is unnecessary (and often unintended) in the vast majority of cases. Its use creates a additional load on the filesystem and even hides bugs in the handling of streams. (Such as the ones I noticed in the tests.)

Picking what to output is a simple matter of intent:
* Do I intent to output a newline then I output a newline char `<< '\n'`. 
* Do I intent to output a newline and flush the underlying buffer then I output  a `<< '\n' << std::flush` aka `std::endl`.

Also, the newline char is shorter on its own and even shorter when integrated into an adjacent string:
```cpp
out << size << "elements" << std::endl;
out << size << "elements" << '\n';
out << size << "elements\n";
```

You are right about the third-party code. I'll revert that.

floli:
Not want to get into a serious discussion with that. I also find the `endl` more pleasing to the eye, but I am ok with it.

PR #334, opened 2019-03-15
Refactor and Name Condition in ReceivePartition

fsimonis:
This PR refactors and names a verbose condition in the condition in ReceivePartition.

@uekerman please merge if ok

### DIFF
--- a/src/partition/ReceivedPartition.cpp
+++ b/src/partition/ReceivedPartition.cpp
@@ -83,8 +83,7 @@ void ReceivedPartition::compute()
       com::CommunicateMesh(utils::MasterSlave::_communication).sendBoundingBox(_bb, 0);
       com::CommunicateMesh(utils::MasterSlave::_communication).receiveMesh(*_mesh, 0);
 
-      if(hasVerticesAtCouplingInterface()) {
-        // the filtered mesh should still have vertices
+      if(areProvidedMeshesEmpty()) {
         std::string msg = "The re-partitioning completely filtered out the mesh " + _mesh->getName() +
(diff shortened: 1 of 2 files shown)


src/partition/ReceivedPartition.cpp
```diff
@@ -482,6 +476,11 @@ void ReceivedPartition::createOwnerInformation()
   }
 }
 
+bool ReceivedPartition::hasVerticesAtCouplingInterface() const {
+    return (_fromMapping && not _fromMapping->getOutputMesh()->vertices().empty()) ||
+           (_toMapping   && not _toMapping->getInputMesh()->vertices().empty());
+}
+
```
uekerman: Good effort, but the name might be misleading (so was probably the comment before). It's not **this** mesh (the mesh associated to this partition), but the other meshes (connected via mappings) that have vertices at the coupling interface. 
Maybe `areConnectedProvidedMeshesAtCouplingInterface` ?

fsimonis: Misleading comment indeed.

This name looks a bit brutal. What about `isRemoteMeshEmpty`?

uekerman: * `remote`, we never use for meshes, and would again be misleading, because one could argue that the received mesh is the remote one. The right expression here would be `provided`. 
* It can be one or two meshes. 

`areProvidedMeshesEmpty` ?

fsimonis: Nice! Short and expressive.

PR #337, opened 2019-03-18
Fix unused parameters

fsimonis:
This PR uncomments all unused parameters, fixing `-Wunused-parameter`

### DIFF
--- a/src/action/ComputeCurvatureAction.cpp
+++ b/src/action/ComputeCurvatureAction.cpp
@@ -20,10 +20,10 @@ ComputeCurvatureAction::ComputeCurvatureAction(
 }
 
 void ComputeCurvatureAction::performAction(
-    double time,
-    double dt,
-    double computedPartFullDt,
-    double fullDt)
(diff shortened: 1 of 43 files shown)


floli:
I think this is a bad idea.  It only adds visual clutter `/* ... */` all over the place without any benefit.

Yes, it silences the warning. Warnings are to make you think, not to blindly follow and silence them at all costs.

fsimonis:
I kept thinking about this `-Wunused-parameter` and I think this is a warning that actually should not be fixed by hiding the name of the parameter.

1) Removing the name makes the function more difficult to understand as the name is gone. Only exception being tag dispatch and `operator++(int)` where the name is pointless and meaningless.
2) Commenting out unnecessarily messes up formatting.
3) Most important however, is that this messes up the document generation of Doxygen.
4) Silencing the warning actually hides the fact that something is probably wrong with the design of the interface (in case of a virtual class hierarchy).

So I'll close this PR

PR #338, opened 2019-03-18
Correct asynchronous sending from sockets

floli:
Correct asynchronous sending of socket requests.

Work of @Summerdave.

Supersedes #247.

- `of-of` system tests succeeds.


fsimonis:
Are there additions to #247, or is this simply a clean version?

floli:
> Are there additions to #247, or is this simply a clean version?

Simply a clean version without same changes that accidentally sneaked in.

PR #339, opened 2019-03-19
Add Accessor Type to access the SolverInterfaceImpl

fsimonis:
This PR:
* adds the fixture `WhiteboxFixture`. This fixture is a friend of `SolverInterface`.
* adds method `impl(T& obj)` which fetches the `_impl` of obj and returns a lvalue reference to it.

Tests can now access the `SolverInterfaceImpl` by using the `WhiteboxFixture` as a test fixture and by writing `impl(interface).configure(...)` instead of `interface._impl->configure(...)`.
Existing fixtures can simply inherit from `WhiteboxFixture`.
There is no extra friend statement necessary.

### DIFF
--- a/src/precice/SolverInterface.hpp
+++ b/src/precice/SolverInterface.hpp
@@ -14,39 +14,8 @@ namespace precice {
   namespace impl {
     class SolverInterfaceImpl;
   }
-}
-
-// Forward declaration to friend the boost test struct
-namespace PreciceTests {
-  namespace Parallel {
-    struct TestFinalize;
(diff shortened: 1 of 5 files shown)


floli:
I think the name Fixture is misleading, as it stands for something which is executed before and after a test to setup testing conditions.

fsimonis:
@floli I see, what about something like `WhiteboxAccessor` or `ImplAccessor`?

uekerman:
I like it a lot! I always hated those friend statements.

PR #343, opened 2019-03-25
Add Libs.private

### LINKED ISSUE #313: pkg-config: Add Libs.private
In the pkg-config we also need to set the [`Libs.private`](https://people.freedesktop.org/~dbn/pkg-config-guide.html#concepts), to be able to easily link to symbols required when preCICE is used as a static library.

fsimonis:
This PR:
* Keeps track of dependencies building a list of libraries
* Adds the field `Libs.private` to the configured `.pc` file.

# TODO
* [ ] Verify that MPI work like this

Closes #313

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 All notable changes to this project will be documented in this file. For future plans, see our [Roadmap](https://github.com/precice/precice/wiki/Roadmap).
 
 ## develop
+- Add Libs.private field to the pkg-config file
 
 ## 1.4.0
 - The python modules are now proper packages tracking dependencies etc.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -140,13 +140,17 @@ set_target_properties(precice PROPERTIES
   CXX_STANDARD 11
   SOVERSION ${preCICE_VERSION}
   )
+# Initialize list of private libraries for the pkg-config Libs.private field
+set(PRECICE_PKG_PRIVATE_LIBS "")
 
 # Setup Boost
 target_compile_definitions(precice PRIVATE BOOST_ALL_DYN_LINK BOOST_ASIO_ENABLE_OLD_SERVICES)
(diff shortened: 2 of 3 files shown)


MakisH:
OpenMPI also provides a pkg-config module:

```
$ pkg-config --libs --static mpi
-L/usr/lib/openmpi/lib -Wl,-rpath -Wl,/usr/lib/openmpi/lib -Wl,--enable-new-dtags -lmpi -ldl -libverbs -lutil -lm
```

MakisH:
I assume that [we also need `Requires.private`](https://people.freedesktop.org/~dbn/pkg-config-guide.html). However, I tried adding `Requires.private: mpi` to the used `libprecice.pc` file and I still get the same error.

Here are the flags & paths that pkg-config returns on my system:
```
$ pkg-config --cflags libprecice 
-pthread -I/home/makish/inst/include -I/usr/lib/openmpi/include -I/usr/lib/openmpi/include/openmpi -I/usr/lib/openmpi/include/openmpi/opal/mca/event/libevent2021/libevent -I/usr/lib/openmpi/include/openmpi/opal/mca/event/libevent2021/libevent/include

$ pkg-config --libs --static libprecice 
-L/home/makish/inst/lib -L/usr/lib/openmpi/lib -lprecice -lboost_filesystem -lboost_log -lboost_log_setup -lboost_program_options -lboost_system -lboost_thread -lboost_unit_test_framework -lboost_date_time -lboost_regex -lboost_chrono -lboost_atomic -ldl -lxml2 -lpetsc -lpython2.7 -Wl,-rpath -Wl,/usr/lib/openmpi/lib -Wl,--enable-new-dtags -lmpi -ldl -libverbs -lutil -lm
```

MakisH:
We are currently confused about why this is happening but we don't consider it a high priority. We decided to postpone it.

fsimonis:
This is not worth the hassle until we have a real reason to do provide this.
Until then, maybe CMake has a generator for find package files ...

PR #344, opened 2019-03-26
Refactor Post-Processing out of cplscheme

### LINKED ISSUE #154: Refactor Post-Processing out of cplscheme
It has a configuration, it has tests, it has a clean interface, it is many files large ... it is actually a separate package. Let's factor it our into one. This will also help to better overview the `cplscheme` package, which has grown too far. 

The only thing left is to choose a good name. Is `pp` too short?

fsimonis:
This PR is based on #154 and refactors the post-processing and pre-conditioners into a separate module/directory/namespace called `post`.
This effectively splits the massive `cplscheme` module into two equally-sized parts (416K and 454K to be precice).

I chose the name `post` as it is short  while still being expressive enough. _Changing this name is not a big problem._

# Remaining Problems to discuss:
* The tests of `cplscheme` and `post` are strongly interleaved and are often based on the same configuration. 
  _Are we ok with this?_
* What should go into `post::impl`?

# TODO
* [x] Improve documentation
* [ ] ~~Rename xml tag to acceleration~~ postponed to #482
* [x] Rename test and configuration sources to acceleration.

---
_Note: Feel free to edit this post._
Closes #154


floli:
> I chose the name post as it is short while still being expressive enough. Changing this name is not a big problem.

I think just `post` is too generic. `postprocessing` is also very generic, maybe we should come up with a completely different name.

uekerman:
I open the discussion (maybe I'll regret it): I never really liked the name *post-processing*. In our field, it is just used for sth completely different (visualization) and let already to many confusions. The better mathematical name would be *acceleration*. The package name could then be `acc`. Of course, this means we have to change all names, including the configuration, which will also mean a little bit of pain for our users.

Going one step further, we could even rename IQN-ILS to Anderson Acceleration and IQN-IMVJ to Generalized Broyden. The Scientific Computing community knows the methods under these names. The partitioned FSI community probably knows IQN-ILS better. This further step is optional, but we should not change things twice. What do you think? I would also want to get Miriam's opinion on this. I'll put it on the agenda for the next telco.  

In `impl` of the new package should be:
* all preconditioner classes
* `ParallelMatrixOperations`
* `QRFactorizaition`
* `SVDFactorization`

MakisH:
> The better mathematical name would be _acceleration_. The package name could then be `acc`.

I really like the name _acceleration_. Before adopting it, though, we should think if we are ever going to port preCICE to _accelerators_, meaning GPUs, MIC (Xeon Phi), etc. But for now we don't have such a need and we will not have it any time soon, since the solvers are usually the bottleneck.

(Similar terminology problem with _parallel_.)

fsimonis:
I moved the implementation details into `post/impl`; I'll leave the discussion to you guys.

uekerman:
As just discussed in the telco:
* We rename post-processing to acceleration. 
* This affects the package name, the configuration tag, but not the methods' names (so IQN-ILS stays IQN-ILS).
* We don't like abbreviations: the package's name should be `acceleration`.
* We will only merge in v2.0 as this is a crucial change in configuration.  

Side note:
* When updating the wiki, we should also add "the other" names, so IQN-ILS = Anderson Acceleration, IQN-IMVJ = Generalized Broyden.

uekerman:
Not to forget: Also the class and file names `PostProcessing.hpp` etc. need to be updated

MakisH:
We decided to merge this up to the point that it does not change the name of the configuration tag.

src/acceleration/impl/ResidualPreconditioner.cpp
```diff
@@ -1,9 +1,9 @@
-#include "ResidualPreconditioner.hpp"
+#include "acceleration/impl/ResidualPreconditioner.hpp"
```
uekerman:
```suggestion
#include "ResidualPreconditioner.hpp"
```
and a few similar cases. Should not be needed, right?

src/precice/config/Configuration.cpp
```diff
@@ -22,7 +22,7 @@ Configuration:: Configuration()
   _tag.addNamespace("action");
   _tag.addNamespace("server");
   _tag.addNamespace("coupling-scheme");
-  _tag.addNamespace("post-processing");
+  _tag.addNamespace("acceleration");
```
uekerman: This should not yet be part of the PR, right?

src/acceleration/impl/ResidualPreconditioner.cpp
fsimonis: I prefer to keep the path explicit as we add `src/` to our include dirs.
This should prevent conflicts with other headers lying around on the system.

PR #345, opened 2019-03-27
Check Repeated Configuration of CPLSchemes for same Participants

### LINKED ISSUE #155: Configuration: check for multiple coupling scheme between same participants
I noticed that, when (accidentally) having both an explicit and an implicit coupling scheme node, this will lead to weird problems during `advance()`. Moreover, there is no relevant output during `advance()` in the default log level, giving the impression that the solver itself hanged.

It would be nice to either:
1. have a check and don't continue if multiple coupling scheme nodes are available,
2. allow for multiple coupling scheme nodes and provide an additional "switch" node to choose among them. Since this would be needed only for more than two nodes, this would be backwards-compatible. This way, we could also distribute files with all the parameters ready-to-use, without having to use comments.

fsimonis:
This PR:
* Adds `std::set<std::string> getParticipants() const` to the coupling schemes.
* Checks at configuration whether a coupling scheme was already defined for a set of participants.

This produces an error on repeated configuration of a coupling scheme between the same participants and thus fixes #155 

Reviewers: please comment if you are ok with the interface extension.

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -416,6 +416,11 @@ std::vector<std::string> BaseCouplingScheme::getCouplingPartners() const
   return partnerNames;
 }
 
+std::set<std::string> BaseCouplingScheme::getParticipants() const
+{
+    return {_firstParticipant, _secondParticipant};
+}
+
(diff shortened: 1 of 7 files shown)


fsimonis:
@uekerman I looked into your various hints. I'll close this PR considering the complexity / gain of this feature.

PR #348, opened 2019-03-29
Draft constness write functions

Dominanz:
As discussed in [this issue](https://github.com/Dominanz/precice/issues/1), the value arguments of the write functions in were changed into `const` pointers in this branch, since access to them is read-only inside this functions. This allows for safer code and better optimizations in programs that use preCICE, because they are now guaranteed that values are accessed read-only in the write functions.

### DIFF
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -671,10 +671,10 @@ int SolverInterfaceImpl:: setMeshVertex
 
 void SolverInterfaceImpl:: setMeshVertices
 (
-  int     meshID,
-  int     size,
-  double* positions,
-  int*    ids )
+  int           meshID,
+  int           size,
+  const double* positions,
+  int*          ids )
 {
   TRACE(meshID, size);
   if (_clientMode){
@@ -698,10 +698,10 @@ void SolverInterfaceImpl:: setMeshVertices
 
 void SolverInterfaceImpl:: getMeshVertices
 (
-  int     meshID,
-  size_t  size,
-  int*    ids,
-  double* positions )
+  int        meshID,
+  size_t     size,
+  const int* ids,
+  double*    positions )
 {
   TRACE(meshID, size);
   if (_clientMode){
@@ -726,10 +726,10 @@ void SolverInterfaceImpl:: getMeshVertices
 }
 
 void SolverInterfaceImpl:: getMeshVertexIDsFromPositions (
-  int     meshID,
-  size_t  size,
-  double* positions,
-  int*    ids )
+  int           meshID,
+  size_t        size,
+  const double* positions,
+  int*          ids )
 {
   TRACE(meshID, size);
   if (_clientMode){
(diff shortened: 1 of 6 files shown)


fsimonis:
@Dominanz Thank you for your contribution!

Could you please also adjust the C bindings in `src/precice/bindings/c/`?

Dominanz:
@fsimonis I have done and commited this.

fsimonis:
I just looked over this PR again.
The pointers to the index arrays should really be const too. Would you like to add this change to this PR?

Dominanz:
You're totally right, I oversaw that. Thanks, I'm doing this tonight!

Dominanz:
@fsimonis Done. I realized that the same thing can be done for set/get mesh vertex functions, so I updated them too

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1190,11 +1190,7 @@ void SolverInterfaceImpl:: writeVectorData
 # endif
   CHECK(valueIndex >= -1, "Invalid value index (" << valueIndex << ") when writing vector data!" );
   if (_clientMode){
-    Eigen::VectorXd valueCopy(_dimensions);
-    for (int dim=0; dim < _dimensions; dim++){
-      valueCopy[dim] = value[dim];
-    }
-    _requestManager->requestWriteVectorData(fromDataID, valueIndex, valueCopy.data());
```
BenjaminRueth: @floli Can we really delete this block of code? To me it looks fine to remove it, but there might be a reason for the existence of these lines that I do not see...

fsimonis: This change doesn't really fit in with this PR, so I would vouche to fix this in the future.

BenjaminRueth: I reverted the changes in https://github.com/precice/precice/pull/348/commits/426f49e6c4fcdd170b5c19835efd3b420904ea7f. We can deal with this in #356

Dominanz:
Well... I am embarrassed that I didn't notice this earlier, but I just realized that only the implementation (/src/precice/impl/SolverInterfaceImpl.*), the request manager (/src/precice/impl/RequestManager.*) and the c bindings were changed in the merge. The /src/precice/SolverInterface itself remained unchanged, rendering the changes useless.

I am pretty confused because I had thought of these files and was convinced that I had modified them too, but I might well have simply forgotten that.

Whatever... should I open a new pull request for that? Or is it, somehow, possible to reuse this one?

BenjaminRueth:
> Well... I am embarrassed that I didn't notice this earlier, but I just realized that only the implementation (/src/precice/impl/SolverInterfaceImpl._), the request manager (/src/precice/impl/RequestManager._) and the c bindings were changed in the merge. The /src/precice/SolverInterface itself remained unchanged, rendering the changes useless.

No reason to be embarrassed. We also didn't notice it in the review of your PR ;)

> I am pretty confused because I had thought of these files and was convinced that I had modified them too, but I might well have simply forgotten that.

I am not sure whether this change was already missing in your original PR or whether @fsimonis and I removed the changes in `src/precice/SolverInterface` in the rebase before merging the merge. In case we deleted the changes in `src/precice/SolverInterface`, I am sorry for the confusion.

> Whatever... should I open a new pull request for that? Or is it, somehow, possible to reuse this one?

As far as I know, we cannot reuse this PR. Just open a new one. You can link to this PR by referring to #348 and possibly your comment https://github.com/precice/precice/pull/348#issuecomment-482654976 in the body of the new PR.

BenjaminRueth:
See https://github.com/precice/precice/pull/357

PR #349, opened 2019-03-29
Refactor Logger

fsimonis:
This PR refactors the current logger.

## Problem:
* Due to the Boost dependencies, the cost of including `Logger.hpp` is around 2.6 seconds, including:
   * 1500ms for `boost/log/expressions.hpp`via `LogMacros.hpp`
   * 810ms for `boost/log/trivial.hpp` aka the actual logger
   * ~250ms for `MasterSlave.hpp`
   * 100ms for `prettyprint.hpp` (Depends strongly on the call-site)
* This header is included in nearly every translation unit, which slows the entire build process down.

## My Approach:
* Use a pimpl to move the boost dependencies into a single translation unit `Logger.cpp`
* Use a minimal struct to forward the log location to the impl.
* Materialize passed log message to a string and pass that to the impl.
* Use functions for the log levels to prevent a boost dependency for the log-level in the interface.

## Side-Effects:
* The Logger itself is now a mere pointer, thus doesn't bloat up the class it is member of.
* Accessing the logger is always an indirection.

## Result:
* The cost of including `Logger.hpp` now converges against `prettyprint.hpp`.
* The build runs notably faster (Debug + `make -j4` on i7-7700 in tempfs):

|  | Real | User | Sys |
| --- | --- | --- | --- |
Before | 3m43s |  11m29s | 0m37s
After | 2m6s | 6m25s | 0m20s

### DIFF
--- a/src/logging/Logger.hpp
+++ b/src/logging/Logger.hpp
@@ -1,15 +1,35 @@
 #pragma once
 
 #include <string>
-#include <boost/log/trivial.hpp>
+#include <memory>
 
 namespace precice {
 namespace logging {
 
-class Logger : public boost::log::sources::severity_logger<boost::log::trivial::severity_level>
-{
+struct LogLocation {
+    const char * file;
+    int line;
+    const char * func;
+};
+
+class Logger {
 public:
   explicit Logger(std::string module);
+  Logger(const Logger& other);
+  Logger& operator=(Logger other);
+  ~Logger();
+
+  void swap(Logger& other) noexcept;
+
+  void error(LogLocation loc, const std::string& mess);
+  void warning(LogLocation loc, const std::string& mess);
+  void info(LogLocation loc, const std::string& mess);
+  void debug(LogLocation loc, const std::string& mess);
+  void trace(LogLocation loc, const std::string& mess);
+
+private:
+  class LoggerImpl;
+  std::unique_ptr<LoggerImpl> _impl;
 };
 
 }} // namespace precice, logging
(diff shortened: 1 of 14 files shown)


src/logging/Logger.hpp
```diff
@@ -1,15 +1,45 @@
...
+/// This class provides a leightweight logger.
+class Logger {
 public:
+  /** creates a logger for a given module.
+   * @param[in] the name of the module 
+   */
   explicit Logger(std::string module);
+
+  Logger(const Logger& other);
+  Logger& operator=(Logger other);
```
Summerdave: Rule of 5: There should be a move constructor somewhere, right? Why does Logger write its own copy/move anyway? Shouldn't the implicit ones suffice?

fsimonis: The default is implicitly deleted as, `std::unique_ptr` has a deleted copy ctor.

Will add the move.

fsimonis:
@Summerdave 
> std::ostream /* or whatever BOOST_LOG... returns */ Logger::debug(LogLocation loc){

That was my initial thought too. The [documentation](https://www.boost.org/doc/libs/1_65_0/libs/log/doc/html/log/detailed/sources.html#log.detailed.sources.severity_level_logger) shows an example of how to do something similar.

This however requires to return a proxy object, which holds an opened record and pushes the record on destruction. This proxy has to inherit from `std::ostream` and needs to be returned as a value, meaning the proxy needs to be a known type. To prevent boost in the header, I would have to create a pimpl.  _it goes on and on and I haven't reached the end yet_

The materialization happens somewhere anyhow and it is a minimally invasive and very simple solution.

floli:
Haven't read it thoroughly, but so far it LGTM. ;-)

PR #357, opened 2019-04-14
Fix Interface constness

fsimonis:
This PR fixes the constness of the SolverInterface including the SolverInterfaceImpl, and RequestManager.

This change boils down to marking pointers of `[in]` parameters `const`.


BenjaminRueth:
See https://github.com/precice/precice/pull/348 and especially https://github.com/precice/precice/pull/348#issuecomment-482654976.

Dominanz:
Looks all fine. I'd appreciate it if this PR could be merged soon. Thanks @fsimonis !

PR #358, opened 2019-04-15
Refactor SolverInterfaceImpl

### LINKED ISSUE #356: Cleanup SolverInterfaceImpl w.r.t usage of Eigen
There are several places, where we can improve SolverInterfaceImpl and make the code more understandable and less verbose.

Let's collect possible improvements in this issue.

fsimonis:
Applying this PR will:
* clean up the use Eigen in the SolverInterfaceImpl (fixes #356)
* Use `Eigen::Map<Eigen::MatrixXd>` to interact with raw buffers from the interface.
* Merge unnecessary repetitions of debug statements
* Refactor the validation of Vertex and Edge IDs into the `Mesh` class as `isValidVertexID()` / `isValidEdgeID()`.
* Refactor check and create for edges into the `Mesh` class as `createUniqueEdge()`
* Removes a few leftover comments

@uekerman please have a look at the mesh interface extension and accept if ok.

### DIFF
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -1130,14 +1130,7 @@ void SolverInterfaceImpl:: mapReadDataTo
                    << "\" to mesh \"" << context.mesh->getName() << "\"");
       assertion(mappingContext.mapping==context.mappingContext.mapping);
       mappingContext.mapping->map(inDataID, outDataID);
-#     ifndef NDEBUG
-      int max = context.toData->values().size();
-      std::ostringstream stream;
-      for (int i=0; (i < max) && (i < 10); i++){
-        stream << context.toData->values()[i] << " ";
-      }
-      DEBUG("First mapped values = " << stream.str());
-#     endif
+      DEBUG("First mapped values = " << context.toData->values());
     }
   }
   mappingContext.hasMappedData = true;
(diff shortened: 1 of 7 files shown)


src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1133,19 +978,11 @@ void SolverInterfaceImpl:: mapReadDataTo
...
       mappingContext.mapping->map(inDataID, outDataID);
-#     ifndef NDEBUG
-      int max = context.toData->values().size();
-      std::ostringstream stream;
-      for (int i=0; (i < max) && (i < 10); i++){
-        stream << context.toData->values()[i] << " ";
-      }
-      DEBUG("First mapped values = " << stream.str());
-#     endif
+      DEBUG("First mapped values = " << context.toData->values());
```
uekerman: Why changing this? This was always very helpful to not spam the complete debug output.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1667,18 +1486,10 @@ void SolverInterfaceImpl:: mapReadData()
...
       context.mappingContext.mapping->map(inDataID, outDataID);
-#     ifndef NDEBUG
-      int max = context.toData->values().size();
-      std::ostringstream stream;
-      for (int i=0; (i < max) && (i < 10); i++){
-        stream << context.toData->values()[i] << " ";
-      }
-      DEBUG("First mapped values = " << stream.str());
-#     endif
+      DEBUG("First mapped values = " << context.toData->values());
```
uekerman: see above

fsimonis:
@uekerman As the data `values()` is an `Eigen::VectorXd`, I added a function `utils::firstN` which returns a const `Eigen::Map` by mapping the first at max n values of the vector to a row vector.
The result is a neat console output at no extra cost.

Issue #353, opened 2019-04-10
OpenFOAM crashes in re-partitioning for empty ranks 


augustinhugues:
Configuration:
OpenFoam 1812, precice 1.4 (built with cmake), petcs 3.10, CalculiX 2.15
(RAM 126Gb, 32 procs)

When I use `nearest-projection` mapping  with mesh connectivity with C3D10 elements the calculation crashes during the filer mesh by bounding box if I am using more than 12 procs (about 14500 cells per proc). It is working fine for serial and up to 12 procs.

I link the log files and the config files.
[Fluid.log](https://github.com/precice/precice/files/3062949/Fluid.log)

[Solid.log](https://github.com/precice/precice/files/3062950/Solid.log)

[precice-config_parallel.txt](https://github.com/precice/precice/files/3062957/precice-config_parallel.txt)
[config.txt](https://github.com/precice/precice/files/3062958/config.txt)

KyleDavisSA:
In the cases that work, is there a part of the interface on every processor from openFOAM? I have found that there are problems with the bounding box function when running openFOAM and a portion of the coupling interface is not on every processor. I don't have a problem if is split over all processors. This is also specific to nearest-neighbor and nearest-projection mapping that I have checked so far.

augustinhugues:
I split a profile along the Z axis, with a space on each side of the profile. I ve just checked, for the 12 proc case, the first and last proc has a really small part of the profile, I think you are right and if I check for 14 proc (which is not working), at least one of the proc won't contain the coupling interface at all.
I check and keep you updated.

MakisH:
I also wonder if this originates from the OpenFOAM adapter or from preCICE, so I am also very interested in your findings. By the way, how does your geometry look like?

KyleDavisSA:
To the best of my memory, this was not a problem with the heart valves where I know that the interface was not on all processors, but this was with FOAM-FSI. It was also an older preCICE version, so I don't know what exactly is causing it.

In my case, I was using the cylinderFlap tutorial. The NN and NP mapping only works when the interface is on all processors.

augustinhugues:
[wing.zip](https://github.com/precice/precice/files/3085568/wing.zip)

Here is my geometry is a section of airfoil profil the coord is 10m and the length is 5m, but will be 40m for final calculation.

As you expected, as long as every proc has a cell of the interface it is running ok. When I go up to 14 procs, two of them haven't any cell of the coupling interface.
[...]

PR #360, opened 2019-04-17
prevent overflow in bounding box side length computation

uekerman:
Should fix issue #353 

I am a bit puzzled why this only happens with OpenFOAM, but it's understandable that we get a problem if we substract the biggest possible double from the smallest one. 
In particular, the problem is not reproducible with a unit test (there are many tests with empty ranks already). 

I am a bit unhappy with the complete `prepareBoundingBox()`. Further problems might be hidden. If anybody sees room for improvement please directly push to this PR. 

We should think about adding a system test variant which would have caught this problem (the plain OpenFOAM-CCX flap_perp tutorial in parallel with NN mapping).


uekerman:
I am closing the PR as I pushed the changes directly to the branch `bugfix_i353`

PR #362, opened 2019-04-18
Bugfix i353

uekerman:
The new PR for fixing issue #353

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file. For future 
 
 ## develop
 
+## 1.4.1
+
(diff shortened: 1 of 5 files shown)


fsimonis:
Travis on `master` is still based on `scons` as we added the full CMake migration post-release in 34b1d56aca140034fb8cbdf42b029119e702a9c8.

I compiled and tested this branch on my compile server and everything runs smoothly:
* [build.log](https://github.com/precice/precice/files/3097677/build.log) - Note that the warnings are fixed on develop
* [tests.log](https://github.com/precice/precice/files/3097678/tests.log)

uekerman:
I don't fully understand the argument that "Travis on master is still based on scons". 
This should also work with scons, right?

PR #367, opened 2019-04-24
Simplify computePartitions

fsimonis:
This PR:
* reduces the verbosity of accessing `_accessor->usedMeshContexts()`
* replaces the second sort with `std::stable_partition`

@floli please check for sanity and merge if ok.


floli:
Looks good to me. Merging.

Issue #371, opened 2019-04-29
Bug in re-partitioning with nearest-projection mapping


uekerman:
We somehow lose triangles while re-partitioning. One example is described in #368, another one was reported by @threimann for FEAP. For FEAP, I found that the triangles are not lost during the bounding box filtering, but afterwards during the mapping-based filtering. The normal vectors of some vertices looked suspicious. 
 
We should try to reproduce the problem with an integration test.

uekerman:
I can reproduce the problem with an integration test.

PR #373, opened 2019-05-04
WIP: Fix NP mapping problem

uekerman:
I managed to reproduce #371 with an integration test. I am able to locate the problem roughly in the NP mapping, but I am not yet able to figure out where exactly.
@fsimonis I need your help here. Let's look at this problem together.

I used the vtk output of the integration test to visualize the problem. 

Triangulated mesh: the mesh that is repartitioned
The blue circles: the provided mesh 
Wireframe triangle: filtered out
White triangle: not filtered out

np-problem

The leftmost triangle is wrongly filtered out. All blue circles there get assigned to the wrong triangle above (with a negative weight).

PS: There was a second problem here: the safety-factor. 0.1 is too small for NP scenarios where vertices could be relatively far away, but still belong to the closest triangle. As an example look at the very rightmost triangle that is not filtered out. 0.5 should be a more robust value here. Please note that the right vertex of the rightmost white triangle could be safely filtered out for a NN mapping. Increasing the safety factor here appears the best option to me. A proper solution would be too involved at the moment. We could even discuss to make `no-filter` the default option as all projection mappings are now down to `N log(N)`.

### DIFF
--- a/src/precice/config/ParticipantConfiguration.cpp
+++ b/src/precice/config/ParticipantConfiguration.cpp
@@ -110,7 +110,7 @@ ParticipantConfiguration:: ParticipantConfiguration
               " name. The creator has to use the attribute \"provide\" to signal he is "
               "providing the mesh geometry.");
   tagUseMesh.addAttribute(attrFrom);
-  auto attrSafetyFactor = makeXMLAttribute(ATTR_SAFETY_FACTOR, 0.1)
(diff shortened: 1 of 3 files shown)


fsimonis:
The spacial tree used by the NP mapping wraps and saves the axis-aligned bounding-boxes of each primitive.
So the bounding boxes of the two triangles are identical, meaning that one of the two triangles will always get picked over the other.
We need to add some kind of hierarchical check for this case.

PR #379, opened 2019-05-10
refactor precice.pyx

### LINKED ISSUE #364: Python bindings: `read_scalar_data` not working
The current implementation of `read_scalar_data` does not work as Python does not copy a simple double by reference. 

https://github.com/precice/precice/blob/develop/src/precice/bindings/python/precice.pyx#L393-L394

A possible solution would be to change the value to be read to a return value.

### LINKED ISSUE #375: Refactor python bindings: precice.pyx
Avoid code duplication:

https://github.com/precice/precice/blob/6ac4882dd466643fd6d27db652c4ad29b4f4b7fb/src/precice/bindings/python/precice.pyx#L260-L282

https://github.com/precice/precice/blob/6ac4882dd466643fd6d27db652c4ad29b4f4b7fb/src/precice/bindings/python/precice.pyx#L301-L323

Possible additional duplicate: `read_block_vector_data` and `read_block_scalar_data`

BenjaminRueth:
working on #375 here.

closes #375 and closes #364 

# Todos

- [x] solve #375
- [x] ~~add tests~~ see https://github.com/precice/precice/issues/409
- [x] ~~make sure that tests are run by Travis~~ see https://github.com/precice/precice/issues/409
- [x] make bindings more pythonic, but do not break API (by providing wrappers! Preparation for https://github.com/precice/precice/issues/364)
- [x] test with solverdummy
- [x] test with partitioned heat equation
- [x] add test for `read/write_scalar_data` in solverdummy. #364


BenjaminRueth:
This PR is ready for review w.r.t #375 . If you want to make the changes readable, you can ignore whitespace changes in the diff.

Screenshot from 2019-05-10 14-32-40

However, one thing that is still missing are tests for the python bindings: I tested the bindings by running the [partitioned heat equation example](https://github.com/precice/precice/wiki/Tutorial-for-solving-the-heat-equation-in-a-partitioned-fashion-using-FEniCS), which also checks the numerical correctness quite well.

However, it does definitely not use all API functions (`read/write(Block)VectorData`, `read/writeScalarData` and probably some more). Therefore **these code changes are untested**. I think this is a good chance to add some reasonable tests to the python bindings.

BenjaminRueth:
The test with the partitioned heat equation and the fenics adapter failed. Therefore, I moved this PR to `v1.x.x`.

BenjaminRueth:
Current version works with solverdummy and partitioned heat equation.

BenjaminRueth:
I will add a test for `read/write_scalar_data` to make sure these functions are working correctly.

BenjaminRueth:
a similar test in the CPP bindings shows that `read/writeBlockScalar` data works correctly (https://github.com/precice/precice/commit/f194a3c71229061158e36bbf5ca50fda96e46d84).

src/precice/bindings/python_future/test/precice-config.xml
```diff
@@ -8,8 +8,8 @@
 
   <solver-interface dimensions="2" >
    
-    <data:vector name="Forces"  />
-    <data:vector name="Velocities"  />
+    <data:scalar name="Forces"  />
+    <data:scalar name="Velocities"  />
```
MakisH: Was this indeed intended?

BenjaminRueth: Yes. I just wanted to test the functions transferring scalar data.

But I agree, the names are confusing and there should also be a test for the functions dealing with vector data.

PR #382, opened 2019-05-10
Add option to pass custom MPI_Comm

### LINKED ISSUE #21: No library routine should use MPI_COMM_WORLD directly
_Moved from Munich GitLab. Original creator: Alexander Shukaev_

Direct excerpt from MPI documentation:

> [...] no library routine should use `MPI_COMM_WORLD` as the communicator; instead, a duplicate of a user-specified communicator should always be used. For more information, see [Using MPI, 2nd edition](https://bibliotecas.cio.mx/ebooks/e0206.pdf).

**NOTE:** This applies for `MPI_COMM_SELF` and friends as well.

I would also confirm that, in general, any communication done by library should be isolated from consumer (application) code by effective privatization of communication context (see [MPI_Comm_dup](http://www.mpich.org/static/docs/latest/www3/MPI_Comm_dup.html)).

fsimonis:
This PR is a prototype of how we could allow callers to pass a custom MPI communicator to preCICE.

I choose to pass the communicator using a void pointer to prevent exposing the mpi header.
The constructor then sets the global communicator via `utils::Parallel::setGlobalCommunicator`.

Closes #21


uekerman:
As just discussed with @fsimonis: Would be good to add an integration test that calls `MPI_Comm_Split` and then the new constructor (followed by a normal simple test case, maybe with PetRBF and `m2n:mpi` to include all critical cases).

MakisH:
We also need to port this to the various language bindings, but we decide to start from C++ for now.

MakisH:
I added two test cases for this (one with NN and one with PetRBF mapping). The NN one succeeds, but the PetRBF hangs in a deadlock. I am not sure if this is a problem of the tests, of the feature implementation, or of something else.

In other test cases we have:
```c++
    utils::Parallel::splitCommunicator( "SolverOne" );
    utils::Parallel::setGlobalCommunicator(utils::Parallel::getLocalCommunicator()); //needed since this test uses PETSc
```

I am not sure how this trick works, but if I use it (in various possible positions), I either get a segmentation fault or the same deadlock.

Furthermore, `utils::Parallel::clearGroups();` seems to be conflicting with the communicator overwriting. This is why I first copy the communicator:

```c++
    MPI_Comm myComm = utils::Parallel::getLocalCommunicator();
    int myCommSize;
    MPI_Comm_size(myComm, &myCommSize);
    BOOST_TEST(myCommSize == 3);
    utils::Parallel::clearGroups();
```
and then call:
```c++
    SolverInterface interface ( "SolverOne", utils::Parallel::getProcessRank(), 3, &myComm );
```

MakisH:
I just now tried this with the [1D elastic tube tutorial](https://github.com/precice/precice/wiki/Example-for-FSI:-1D-elastic-tube), modifying the [Fluid solver to use the new constructor](https://github.com/precice/elastictube1d/blob/ecd6029b344f332e2111c8fd285fbdce56c1f53a/FluidSolver_Parallel/FluidSolver.cpp#L60):

```c++
  SolverInterface interface(solverName, rank, size); \\ old
  SolverInterface interface(solverName, rank, size, MPI_COMM_WORLD); \\ new
```

When I run the fluid solver:
```bash
$ mpiexec -np 2 ./FluidSolverParallel ./ConfigurationFiles/precice-config-parallel.xml 100 0.01 100
```

I get an `invalid communicator` error:
```
Starting Fluid Solver...
Starting Fluid Solver...
[2019-08-04 10:44:29.717122] [0x00007f7a157b3740] [trace]   Entering operator()
[2019-08-04 10:44:29.717121] [0x00007f982bbd9740] [trace]   Entering operator()
[2019-08-04 10:44:29.717171] [0x00007f7a157b3740] [trace]   Leaving setGlobalCommunicator
[2019-08-04 10:44:29.717176] [0x00007f982bbd9740] [trace]   Leaving setGlobalCommunicator
(0) 10:44:29 [impl::SolverInterfaceImpl]:118 in configure: This is preCICE version 1.4.1
(0) 10:44:29 [impl::SolverInterfaceImpl]:119 in configure: Configuring preCICE with configuration: "./ConfigurationFiles/precice-config-parallel.xml"
(0) 10:44:29 [impl::SolverInterfaceImpl]:162 in configure: Run in coupling mode
(0) 10:44:29 [impl::SolverInterfaceImpl]:162 in configure: Run in coupling mode
[Magdalena:8165] *** An error occurred in MPI_Comm_rank
[Magdalena:8165] *** reported by process [1901723649,1]
[Magdalena:8165] *** on communicator MPI_COMM_WORLD
[Magdalena:8165] *** MPI_ERR_COMM: invalid communicator
[Magdalena:8165] *** MPI_ERRORS_ARE_FATAL (processes in this communicator will now abort,
[Magdalena:8165] ***    and potentially your MPI job)
[Magdalena:08162] 1 more process has sent help message help-mpi-errors.txt / mpi_errors_are_fatal
[Magdalena:08162] Set MCA parameter "orte_base_help_aggregate" to 0 to see all help / error messages
```

Am I using it wrong?

fsimonis:
@MakisH the constructor takes a pointer to a MPI Comm.

So you have to take the address first.

fsimonis:
I think we have add a reset for the communicator in the test fixture as it is stored as static information.

MakisH:
> @MakisH the constructor takes a pointer to a MPI Comm.
> 
> So you have to take the address first.

Ah, indeed. Now the elastictube1d runs, both with NN and PetRBF.



> I think we have add a reset for the communicator in the test fixture as it is stored as static information.

This is already included in the fixture. I see that we still use it sometimes before we create and configure the interface. I tried the same, but it still hangs.

fsimonis:
> Ah, indeed. Now the elastictube1d runs, both with NN and PetRBF.

Great!

> This is already included in the fixture.

@MakisH True, but this seems to be a problem with the tests then as the elastictube1d runs fine :thinking:

fsimonis:
I rebased the PR on develop.

fsimonis:
@uekerman @MakisH I ran the test `PreciceTests/Parallel/UserDefinedMPICommunicator` and got a SEGFAULT from the following location on all 4 Ranks:
```
#0  0x00007ffff2074774 in PMPI_Comm_size () from /usr/lib/libmpi.so.12
#1  0x00007ffff77d5593 in precice::utils::EventRegistry::collect (
    this=0x7ffff7dd5f50 <precice::utils::EventRegistry::instance()::instance>)
    at <snip>/src/utils/EventUtils.cpp:421
#2  0x00007ffff77d53fd in precice::utils::EventRegistry::finalize (
    this=0x7ffff7dd5f50 <precice::utils::EventRegistry::instance()::instance>)
    at <snip>/src/utils/EventUtils.cpp:232
#3  0x000000000078692b in main (argc=2, argv=0x7fffffffce98)
    at <snip>/src/testing/main.cpp:111
```

uekerman:
I am still a bit puzzled. 
To be really sure that this is a test-only problem, we should test a `elastictube1D` or similar that does not use `MPI_Comm_World`, but which calls itself `MPI_Comm_Split` and uses preCICE only on a sub-communicator. That's the real use case we are targeting with this feature anyway, right?

fsimonis:
It turned out that the tests using the constructor called `xml::configure` prior to constructing the SolverInferface. The configure instantiates the mapping and the PetRBF mapping uses PETSc which uses the global communicator.
The the global communicator was not set at that moment and :boom:.

MakisH:
Are the tests now supposed to exit normally or is the segfault still there? Because I always get a segfault when I run any parallel test explicitly (or even all parallel tests: `--run_test=PreciceTests/Parallel/*`).

MakisH:
Tests are now working like a charm! :rocket: 

> I am still a bit puzzled.
> To be really sure that this is a test-only problem, we should test a `elastictube1D` or similar that does not use `MPI_Comm_World`, but which calls itself `MPI_Comm_Split` and uses preCICE only on a sub-communicator. That's the real use case we are targeting with this feature anyway, right?

I now modified the `elastictube1d` to split the communicator, do everything in half the ranks and do nothing on the rest. Since this is very quick&dirty, here is a [gist with the code, config, and log](https://gist.github.com/MakisH/ce52cd2116e1d030d740c1e74033571e).

**Good news:** it runs smoothly with nearest-neighbor.
**Bad news** (or another bug or something I am missing): it hangs when using PetRBF, even when the RBF mapping is defined only for the other solver (StructureSolver):

```
(0) 13:56:16 [config:MappingConfiguration]:301 in createMapping: Entering operator()
  Argument 0: direction == read
  Argument 1: type == petrbf-thin-plate-splines
  Argument 2: timing == 0
  Argument 3: shapeParameter == 0
  Argument 4: supportRadius == 0
(0) 13:56:16 [utils::Petsc]:21 in initialize: Entering operator()
```

The exact same config file works with the original solver. In that case, the Fluid continues after the Solid also synchronizes, so this should have something to do with the communication.

Here is the respective preCICE code:
https://github.com/precice/precice/blob/a0b34e980f4a69cf5f5d78f08123c61204006f58/src/utils/Petsc.cpp#L16-L32

uekerman:
It's a good testcase! 
I can't see any obvious problem in your code. Seems like a bug in preCICE to me. 

> Bad news (or another bug or something I am missing): it hangs when using PetRBF, even when the RBF mapping is defined only for the other solver (StructureSolver):

We still run through the configuration and call the initialization of PETSc (the configuration is always executed for all participants). Maybe we shouldn't do this in the configuration.

fsimonis:
I found ~~the~~ a problem!

In the docs of [PetscInitialize](https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Sys/PetscInitialize.html):
> Collective on MPI_COMM_WORLD or PETSC_COMM_WORLD if it has been set 

And the docs of [PETSC_COMM_WORLD](https://www.mcs.anl.gov/petsc/petsc-current/docs/manualpages/Sys/PETSC_COMM_WORLD.html#PETSC_COMM_WORLD) state:
>  By default PETSC_COMM_WORLD and MPI_COMM_WORLD are identical unless you wish to run PETSc on ONLY a subset of MPI_COMM_WORLD. In that case create your new (smaller) communicator, call it, say comm, and set PETSC_COMM_WORLD = comm BEFORE calling PetscInitialize(), but after MPI_Init() has been called. 

Thus, we have to set `PETSC_COMM_WORLD = Parallel::getGlobalCommunicator();` on line 26 here:
https://github.com/precice/precice/blob/ec01b00f0f28563c12da2223a90f8704963c7503/src/utils/Petsc.cpp#L23-L30

We should have noticed this sooner. ( At least according to the amount of uppercase letters outside of variable names. )

PR #391, opened 2019-05-19
Add PR quality checks on travis and a bot

shkodm:
This adds several checks, that are performed on Travis for pull requests
For now it means that for all changed files following checks are done:
* **Code formatting**
  It will check full file, so at some point the source code should be reformated, so this check does not get triggered with every PR
* **Warnings from clang-tidy**

And if there are enough total lines changed in the code (I am not sure what metric should be set here so now it is magical number of 100 lines, I am open to suggestions ), check whether something added to the `CHANGELOG`.

If some check fails @precice-bot will post a message (e.g. see [ this dummy PR ]( https://github.com/shkodm/precice/pull/5 ) on my fork ), with the reasons of failure. 

This is not the final state, more demonstration of the possibilities, so I am encouraging to propose some ideas of what other checks can be useful or about improvement of the current state.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -73,6 +73,13 @@ matrix:
         - lcov --remove coverage.info '/usr/*' "$LOCAL_INSTALL/*" --output-file coverage.info # filter system-files and local-dependencies
         - lcov --list coverage.info # debug info
         - bash <(curl -s https://codecov.io/bash) -f coverage.info || echo "Codecov did not collect coverage reports"
+    # Additional checks of code quality, etc
+    - name: "Pull request check"
+      if: type = pull_request
+      include:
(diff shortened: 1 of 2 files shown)


fsimonis:
We have to fix the version of `clang-format` to version `8` to ensure consistent results.

This will be a pain on Travis and just one more reason to invest in moving to `github-actions`.
I'll add a corresponding workflow soon.

fsimonis:
This is on fold until we migrated to GitHub Actions.

fsimonis:
This PR is now obsolete as we moved to GitHub actions.
Since then, we added code format checking to the CI #759 #866 #738.
Instead of `clang-tidy` we decided to run coverity-scan #741 on a weekly basis.

PR #404, opened 2019-05-21
Add MPI includes to PETSc discovery

fsimonis:
This PR adds MPI discovery to `FindPETSc` and uses the include flags as part of the minimal include directories. This fixes a major reason for failing configurations of preCICE using CMake. 

This allows the module to work as expected even if the compilers are not set to the mpi wrappers. Which is necessary to discover  `mpi.h` in case of a scoped install e.g. `/usr/include/openmpi/mpi.h`.

### DIFF
--- a/cmake/modules/FindPETSc.cmake
+++ b/cmake/modules/FindPETSc.cmake
@@ -28,6 +28,8 @@ set(PETSC_VALID_COMPONENTS
   C
   CXX)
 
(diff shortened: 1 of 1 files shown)


MakisH:
In other works, this makes this troubleshooting unnecessary: https://github.com/precice/precice/wiki/Building:-Using-CMake#petsc-could-not-be-found

PR #406, opened 2019-05-21
Fix wrong #include

BenjaminRueth:
Don't use `#include precice/impl/SolverInterfaceImpl.hpp`. Use `#include precice/SolverInterface.hpp`.


BenjaminRueth:
Tested modified bindings with `precice/tools/solverdummies/fortran` and it works.

PR #408, opened 2019-05-21
Allow to send a mesh to multiple participants

KyleDavisSA:
supposed to fail as we still need to implement this feature

### DIFF
--- a/src/precice/tests/SerialTests.cpp
+++ b/src/precice/tests/SerialTests.cpp
@@ -1543,7 +1543,9 @@ BOOST_AUTO_TEST_CASE(testSendMeshToMultipleParticipants,
   cplInterface.finalize();
 }
 
+/**
  * @brief Test to reproduce the problem of issue 383, https://github.com/precice/precice/issues/383
+ *
  */
 BOOST_AUTO_TEST_CASE(testPreconditionerBug,
                      * testing::MinRanks(2)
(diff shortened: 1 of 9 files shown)


KyleDavisSA:
The functionality works with OpenFOAM and Calculix. I had two identical fluid participants (FluidA and FluidB with the same mesh and boundary conditions) and one Calculix participant. FluidA and Calculix were serial-implicitly coupled (exchanging forces and displacements) and Calculix and FluidB serial-explicitly coupled (displacements sent to FluidB after the coupling converged for  FluidA and Calculix).

Even though a mesh can be sent to multiple participants, they seem to differ by one time step. The displacements of timestep _i_ for the implicitly coupled case matches the results of timestep _i - 1_ for the explicit coupled case.

uekerman:
> Even though a mesh can be sent to multiple participants, they seem to differ by one time step. The displacements of timestep i for the implicitly coupled case matches the results of timestep i - 1 for the explicit coupled case.

This is most probably a feature of the serial coupling and not a bug. If the solid solver is the "first" participant, forces match between participants at the same time level, but not the displacements.

src/precice/tests/SerialTests.cpp
```diff
@@ -1543,6 +1543,56 @@ BOOST_AUTO_TEST_CASE(testSendMeshToMultipleParticipants,
   cplInterface.finalize();
 }
 
+ * @brief Test to reproduce the problem of issue 383, https://github.com/precice/precice/issues/383
+ */
```
MakisH: Typo

PR #416, opened 2019-05-23
Output of filtered vertices only as aggregated count.

floli:
+ Output as INFO level, because it's happens quite often and is not
critical.
+ All filtered vertices can still be printed when activating debug log output.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to this project will be documented in this file. For future plans, see our [Roadmap](https://github.com/precice/precice/wiki/Roadmap).
 
+## Post 1.5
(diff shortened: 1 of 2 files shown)


fsimonis:
Could also print the total amount of vertices in the mesh? "... 3 / 30 vertices were filtered out..."

floli:
> Could also print the total amount of vertices in the mesh? "... 3 / 30 vertices were filtered out..."

Good idea, now output is like:
```
(0) 15:12:25 [partition::ReceivedPartition]:478 in createOwnerInformation: 47271 of 47549 vertices of mesh MeshA have been filtered out since they have no influence on the mapping.

Forced push commit.
```

I also changed the order, so if you have debug output enabled, you first get the flood of messages and then the summary.

floli:
Applied the requested change.

Issue #409, opened 2019-05-22
Implement automated testing for language bindings


BenjaminRueth:
Our current testing environment only provides very limited testing of the different language bindings.

# What we have

systemtests that run the solverdummies for the different language bindings in [`precice/systemtests/tree/master/Test_bindings`](https://github.com/precice/systemtests/tree/master/Test_bindings).

# What we want

Bindings tests that can be run easily:

* for c, cpp using ctest (either independently or in the main build directory)
* for python using `python3 setup.py test` (compare to [fenics-adapter, testing](https://github.com/precice/fenics-adapter#test-the-adapter))
* for fortran?
* for other?
* run all tests automatically for each commit in Travis (just like the preCICE unit tests)

# The main problem

Currently, we have to run two solverdummies in order to be able to test the bindings. Isolated tests for testing the API calls independently cannot be realized. We do not want to spoil the solverdummies as their purpose is to show how the API can be used, not testing.

BenjaminRueth:
Possible solution: Some kind of mocking, like in [`fenics-adapter`](https://github.com/precice/fenics-adapter#test-the-adapter), where `SolverInterface.hpp` is mocked. It is sufficient to mock each API function separately, because we do not want to test preCICE but only the input-output behaviour of the bindings.

BenjaminRueth:
C++ mocking framework: https://github.com/eranpeer/FakeIt

BenjaminRueth:
#420 and #431 implement automated testing for the python bindings.

fsimonis:
This has been partially implemented for C and Fortran by building and running the solverdummies in #702.

BenjaminRodenberg:
@fsimonis do you think we should still keep this issue open? I don't have an overview over C and Fortran. For python this issue is definitely solved. For the other language bindings we still need tests, but I think this is more an issue for the respective repositories and not so much for this one.

fsimonis:
This issue is still open. The solverdummies only cover a part of the interface. Many other features can break in the bindings without anyone noticing.

Good example was a mismatched definition in the C bindings after the recent rework, which was noticed when we ported the Matlab bindings

PR #420, opened 2019-05-25
Mocking precice in cython

BenjaminRueth:
Related to #409 

This PR is based on #379, therefore #379 should be merged first.

In this PR the CPP preCICE language bindings are mocked in order to allow unit testing for the python language bindings using `python3 setup.py test`

# Todo

- [x] ~~run all tests that are run in `solverdummy.py` using mocked `SolverInterface.hpp`~~ better: provide `SolverInterface.cpp` with mocked implementation, but use original `SolverInterface.hpp`.
- [x] run tests on Travis (see https://github.com/precice/precice/pull/421)
- [x] don't duplicate `SolverInterface.hpp` (see [here](https://github.com/precice/precice/pull/420#issuecomment-495924504)). Use the same header for tests and production runs, but use different implementations.


BenjaminRueth:
# Current state

Mocking of the language bindings is realized with a flag given to the compiler. 

https://github.com/precice/precice/blob/8960fc65d942977cea48f26fac8f05aeb0b23c40/src/precice/bindings/python_future/precice_future.pxd#L9-L12

The flag determines whether the real or the mocked `SolverInterface.hpp` is imported with `cimport`. The flag is set at compile time by `setup.py`. This depends on whether `setup.py` is run with `test` or `install`. A minimal unit test, which uses the mocked interface, is run for `setup.py test`.

This approach has some disadvantages:

* The interface has to be mocked at compile time. We cannot specify the behaviour of the mocked interface from within [the actual tests](https://github.com/precice/precice/blob/mockingPreciceInCython/src/precice/bindings/python_future/test/test_bindings_module.pyx), like it is usually possible if a proper mocking framework is used. See [here](https://github.com/precice/fenics-adapter/blob/567d0ac737e076310156df7af89efc8d6f3ecbac/tests/test_fenicsadapter.py#L58-L61).

* The mocked [`SolverInterface.hpp`](https://github.com/precice/precice/blob/mockingPreciceInCython/src/precice/bindings/python_future/test/SolverInterface.hpp) is a duplication of the real [`SolverInterface.hpp`](https://github.com/precice/precice/blob/mockingPreciceInCython/src/precice/SolverInterface.hpp).

PR #421, opened 2019-05-27
Run bindings tests in Travis

BenjaminRueth:
# Todos before merge

- [x] use `pip`, not `setup.py` [for installation](https://github.com/precice/precice/blob/d0620b2cffb2c64fddecbc1da515e27aaf5cf485/tools/travis-build-test.sh#L32-L34) (main question: how to provide a include and library path to pip?) **we rather use both, because we also document both**
- [x] ~~provide a script for setup of python environment ([here](https://github.com/precice/precice/blob/d0620b2cffb2c64fddecbc1da515e27aaf5cf485/.travis.yml#L99-L102) and [here](https://github.com/precice/precice/blob/d0620b2cffb2c64fddecbc1da515e27aaf5cf485/.travis.yml#L81-L84))~~ *only low gain* ~~and installation of python dependencies to avoid code duplication ([here](https://github.com/precice/precice/blob/d0620b2cffb2c64fddecbc1da515e27aaf5cf485/tools/travis-build-test.sh#L29) and [here](https://github.com/precice/precice/blob/d0620b2cffb2c64fddecbc1da515e27aaf5cf485/.travis.yml#L106))~~ *irrelevant at current state*


BenjaminRueth:
Python bindings are now built and test on Travis. I will add a few more tests, before I merge. However, I am still a bit unsure where to build the bindings and run the test. Should we do this in the main preCICE tests? Or create an independent stage?

*Note:* If we want to build the real bindings (not the ones relying on the mocked Interface), we have to build preCICE.

tools/travis-install-dependencies.sh
```diff
@@ -96,3 +96,4 @@ if [ ! -f $CACHE_CCACHE_TOKEN ]; then
     # Create token
     touch $CACHE_CCACHE_TOKEN
 fi
+
```
shkodm: Nothing should be changed here

src/precice/bindings/python_future/test/test_bindings_module.pyx
```diff
@@ -12,7 +12,7 @@ class TestBindings(TestCase):
     """
     Test suite to check correct behaviour of python bindings.
     """
-    
+
```
shkodm: Nothing should be changed here

.travis.yml
```diff
@@ -99,3 +116,4 @@ after_failure:
   - cd $TRAVIS_BUILD_DIR
   - cat config.log
   - cat -n ./tests/boost-test-output*
+
```
shkodm: Nothing should be changed here

.travis.yml
```diff
@@ -72,9 +85,8 @@ matrix:
         - lcov --capture --directory "$TRAVIS_BUILD_DIR/build" --output-file coverage.info
         - lcov --remove coverage.info '/usr/*' "$LOCAL_INSTALL/*" --output-file coverage.info # filter system-files and local-dependencies
         - lcov --list coverage.info # debug info
-        - bash <(curl -s https://codecov.io/bash) -f coverage.info || echo "Codecov did not collect coverage reports"
-
-
+        - bash <(curl -s https://codecov.io/bash) -f coverage.info || echo "Codecov did not collect coverage reports"      
+
```
shkodm: Nothing should be changed here

shkodm:
>> However, I am still a bit unsure where to build the bindings and run the test. 

@BenjaminRueth , I am not sure. Why do we want to have it tested here and not in the systemtests? Besides, I though there is already a test for bindings there.

BenjaminRueth:
> Besides, I though there is already a test for bindings there.

@shkodm The test in the system tests repository makes sure that the bindings can be installed and the solverdummy is working. However, I think a more fine-grained testing would be good for the bindings, since the solverdummy only shows how a minimal adapter that uses the python bindings looks like. Many cases are not covered by this test. See for example #364.

tools/travis-install-dependencies.sh
BenjaminRueth: Rolled this back in https://github.com/precice/precice/pull/421/commits/5f7809a8c733f007817750b73e41e98d4d06a016

.travis.yml
BenjaminRueth: Rolled back in https://github.com/precice/precice/pull/421/commits/7c2ec4b1f06467d37244bf51177e0eccb134d453

.travis.yml
BenjaminRueth: Rolled back in https://github.com/precice/precice/pull/421/commits/7c2ec4b1f06467d37244bf51177e0eccb134d453

src/precice/bindings/python_future/test/test_bindings_module.pyx
BenjaminRueth: Rolled back in https://github.com/precice/precice/pull/421/commits/7c2ec4b1f06467d37244bf51177e0eccb134d453

PR #438, opened 2019-06-14
WIP: Fix NP Mapping

### LINKED ISSUE #371: Bug in re-partitioning with nearest-projection mapping

fsimonis:
This PR Fixes the RTree-based NP mapping.

Fixes #371 
Closes #373

### DIFF
--- a/src/mapping/NearestNeighborMapping.cpp
+++ b/src/mapping/NearestNeighborMapping.cpp
@@ -35,7 +35,7 @@ void NearestNeighborMapping:: computeMapping()
   
   if (getConstraint() == CONSISTENT){
     DEBUG("Compute consistent mapping");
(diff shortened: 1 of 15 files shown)


MakisH:
In short, what is the fix?

fsimonis:
This PR is a mess. I'll clean it up and reopen another one.

fsimonis:
@MakisH The is is that the initial rtree version got muddled up with overlapping bounding boxes of saved primitives. This bug has been fix by creating separate indices for each primitive.

Issue #261, opened 2019-02-13
Add getVersion to SolverInterface


fsimonis:
The xSDK requirement M8 requires a __runtime API__ to fetch the current version of the library.

MakisH:
Remember to also port this to all bindings.

PR #446, opened 2019-06-26
Add Git Revision Info to Library

### LINKED ISSUE #424: Versioning to differentiate between release, development, and release candidate
Currently, we increase the version number (displayed at runtime), only for releases.

This causes the following problems:

1. Only by looking at the output, we don't know if a user (or we) uses the respective release (e.g. `1.4.1`) or a later state from another branch (e.g. `develop`, which would be `1.4.1 + changes`).

2. While testing our release candidate, it is not easy to say that we are testing against the correct version.

Both problems are prevalent when we have multiple versions of preCICE installed (in different paths), or a broken previous installation.

I propose the following system:

* `1.4.1-develop` for the develop branch after release 1.4.1 (feel free to adjust for other branches). **Question:** Should we use `past-develop`, or `expected-develop`?
* `1.5.0-RC1` for the release candidate that we are testing. Every time a review is re-requested, `RC++`.
* `1.5.0` only for the specific commit that is released. Immediately afterwards, it should change to `1.5.0-develop`.

I don't have any specific examples from other projects in mind, but I think that this system would be reasonable and it would reduce our stress during testing. What do you think?

fsimonis:
This PR adds the described commit info to the library and prints it during configuration.

Scenario | Info
| --- | --- |
No git found | `no-info`
Git fails/No git repo | `Detection failed`
Otherwise | `git describe --tags --broken`

The resulting output looks like this: (C++ Solverdummy)
```
Starting SolverDummy...
preCICE: This is preCICE version 1.5.1
preCICE: Revision info: v1.5.1-27-gc7201c3b-dirty
preCICE: Configuring preCICE with configuration: "../../tools/solverdummies/precice-config.xml"
preCICE: Run in coupling mode
preCICE: Setting up master communication to coupling partner/s
```

Part |  Description
-- | --
v1.5.1 | last tag
27 | commits since last tag
gc7201c3b | hash of last commit
dirty | there are uncommitted changes

Also, one can now grep for the version:
```
grep -a "v[1-9]\.[1-9]\.[1-9]" libprecice.so.1.5.1
```

CMake will automatically update the `src/version.cpp` on each build which results in occasional recompilation and relinking.

Closes #424


floli:
Does it also work with scons? For me it's ok, if there is no info when using scons, but it should not fail.

fsimonis:
@floli Thanks for the reminder! I added the configuration as a no-op to SCons.

fsimonis:
@MakisH Please have a look at the first table of the PR message.

This is a fully automated solution which relies on working git installation and a clone of the repository, which is the setup of every core developer. 

Every other approach requires us to commit information into the repository. This information has to be updated and thus adds additional steps to the release procedure.

IMHO a viable solution to #424 has to be a zero-overhead solution for the developer.

MakisH:
I saw the table already and I understand that we will still not be able to decide if a user uses a released version or an arbitrary state of `develop`.

I would say let's merge this and keep #424 open for now. Maybe we find a better way to solve it down the road.

fsimonis:
@MakisH If a user uses a release, I would expect him to install the debian package, use spack, or download the tar. In this case the version number is sufficient.

If a user uses a develop version, I would expect her to clone the git repo and compile from source. In this case the info is present and will be compiled into the binary.

So this PR solves point 2) of #424 and 1) for cloned repos, which I would assume to be always true for users of develop.

MakisH:
> If a user uses a develop version, I would expect her to clone the git repo and compile from source. In this case the info is present and will be compiled into the binary.

This would be a very reasonable thing to do. But new users (e.g. students without Git experience) often do strange things. One of these strange things is to get a snapshot of the repository in a zip file from GitHub and there are at least two ways to do that:

zip

assets

When this user sends me a log file, I want to immediately know if this is actually "1.5.1" or "1.5.1 + new stuff".

I know I could just ask the user, and that's why I say this solution is good enough for most of the cases.

fsimonis:
Merging this will simplify a solution for #261

MakisH:
I approve this as a concept, but I have not checked it in practice.

PR #450, opened 2019-06-28
WIP: Doxygen style documentation of python_future bindings

IshaanDesai:
Adding doxygen style documentation to python_future bindings. This documentation template is intended to be a pilot setup for further documentation plan of all the bindings.


uekerman:
Do we really want to copy over the full documentation from the C++ API? I would prefer to have it just as we do for Fortran or C, explaining the arguments and referring to the C++ documentation: https://github.com/precice/precice/blob/develop/src/precice/bindings/fortran/SolverInterfaceFortran.hpp
Other opinions?

MakisH:
I see your point (code duplication), but maybe it makes sense to duplicate, because they are different in every language: different types, slightly different names, maybe there are even additional arguments. The "challenge" would be, of course, to keep them up to date. But this is not a strong opinion.

uekerman:
Documenting what the arguments are is, of course, good. But we should avoid to duplicate the "body" of the documentation. Still, going over all bindings and making documentations consistent is a good thing!

IshaanDesai:
I had a discussion with @BenjaminRueth and we came to a similar conclusion that as a first step this will be a template of doxygen documentation for the bindings, but yes I believe it is better to link to the C++ API for the body of the text of each function. I am not sure if we can add hyperlinks for direct linking in the doxygen documentation.

BenjaminRueth:
Like @IshaanDesai already pointed out this PR's main purpose is to find out how to document our bindings. For the python bindings I think we should definitely provide some means of documentation at the current state. The main question is now how to do this (technically and with respect to style).

In order to avoid code duplication, there are some quite nice ways of referencing in Doxygen. See [here](http://www.doxygen.nl/manual/autolink.html).

IshaanDesai:
It appears that doxygen is not able to support cython files with extension `.pyx` as a standard programming language for its documentation toolbox. Problems arise for function definitions of this type:
```
def get_mesh_vertex_size (self, int mesh_id):
     return self.thisptr.getMeshVertexSize(mesh_id)
```
If the extension `.pyx` is specified as a python file then doxygen reads the keyword **int** as a variable and looks for its documentation. One possible solution is to define the doxygen behaviour for cython files for further use.

BenjaminRueth:
This PR should be moved to https://github.com/precice/python-bindings. As soon as it is moved, we can close it.

IshaanDesai:
Refer https://github.com/precice/python-bindings/pull/10 as this PR is now ported to new python bindings repository.

PR #452, opened 2019-07-02
Make sending and receiving in GatherScatter consistent, fix #451

### LINKED ISSUE #451: Bug with master:sockets when coupling against a serial participant
Hi,

I've found that problems occur in preCICE in the following scenario:

- You couple a participant running in parallel against a participant running serially.
- You use the <master:sockets> tag for the parallel participant

An instance of the problem can be reproduced e.g. with the elasticTube1D example: I added a file named `precice-config-hybrid.xml` in the ConfigurationFiles folder. This file is an exact copy of `precice-config-parallel.xml` with the following two exceptions:

- The master tag in the structure solver was removed completely.
- The tag `<master:mpi-single>` in the fluid solver was replaced by `<master:sockets>`.

For convenience, I attached the file. For whatever the hell reason, github told me it doesn't support uploading xml files, so i renamed it to .log: [precice-config-hybrid.log](https://github.com/precice/precice/files/3346140/precice-config-hybrid.log)

Now, if you run the solvers using

- `mpirun -n 2 ./FluidSolverParallel ConfiguonFiles/precice-config-hybrid.xml 100 0.01 100
`
- `mpirun -n 1 ./StructureSolverParallel ConfigurationFiles/precice-config-hybrid.xml 100` or simply `./StructureSolver ConfigurationFiles/precice-config-hybrid.xml 100`

they will run into an incomprehensible error. The message was: "ERROR: Receive failed: vector::_M_default_append".

 If you do the very same without replacing `<master:mpi-single>` by `<master:sockets>` in the fluid solver, it runs just fine.

I had a similar observation with a minimal example I wrote for myself. My example worked properly and as expected with `<master:mpi-single>`. When I switched to `<master:sockets>`, I observed weird mapping behaviour (nearest-neighbor mapping didn't work properly) and the execution terminated on an error stating that `advance()` encountered ambiguous time step lengths. 

I conclude there must be a bug in the implementation of the master tag. I would highly appreciate a hot fix, since I am forced to compile preCICE without MPI due to compatibility problems between MATLAB and MPI.

Thanks in advance!

uekerman:
In `GatherScatterCommunication`, we previously sent a vector of doubles but received an array of doubles. This is wrong for any underlying communication, but only had crucial impact when `SocketCommunication` was used. 

closes #451

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,10 @@ All notable changes to this project will be documented in this file. For future 
 
 ## develop
 
+## 1.5.2
+
+- Fixed faulty communication in `master:sockets` leading to crashes.
+
 ## 1.5.1
 
 - Fixed the exposure of `boost::asio` implementation details leading to version incompatibilities
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,6 +1,6 @@
 cmake_minimum_required (VERSION 3.9.6)
 
-project(preCICE VERSION 1.5.1 LANGUAGES CXX)
+project(preCICE VERSION 1.5.2 LANGUAGES CXX)
 
 #
 # Overview of this configuration
--- a/SConstruct
+++ b/SConstruct
@@ -106,7 +106,7 @@ env.Append(LIBPATH = [('#' + buildpath)])
 env.Append(CCFLAGS= ['-Wall', '-Wextra', '-Wno-unused-parameter', '-std=c++11'])
 
 # ====== PRECICE_VERSION number ======
-PRECICE_VERSION = "1.5.1"
+PRECICE_VERSION = "1.5.2"
 
 
 # ====== Compiler Settings ======
--- a/src/m2n/GatherScatterCommunication.cpp
+++ b/src/m2n/GatherScatterCommunication.cpp
@@ -90,7 +90,7 @@ void GatherScatterCommunication::send(
       DEBUG("Slave Size = " << slaveSize);
       if (slaveSize > 0) {
         std::vector<double> valuesSlave(slaveSize);
-        utils::MasterSlave::_communication->receive(valuesSlave, rankSlave);
+        utils::MasterSlave::_communication->receive(valuesSlave.data(), slaveSize, rankSlave);
         for (size_t i = 0; i < vertexDistribution[rankSlave].size(); i++) {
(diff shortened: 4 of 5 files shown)


BenjaminRueth:
In [`debugMasterSockets`](https://github.com/precice/precice/tree/debugMasterSockets/src/cplscheme/tests/systemtest) there is a minimal example for reproducing the error described in #451.

fsimonis:
@uekerman I would like to question this fix.

From my perspective, sending a raw array of `T` and sending a `std::vector<T>` should be freely interchangeable. At the end of the day, both represent a contiguous block of memory.
This problem is essentially a consistency problem as the `vector` variant sends the size while array variant doesn't.

I would vouch to always send the size first, so that we can validate/adjust the size of the destination at the receiving end.
This would make our communication facilities consistent and easier to understand/use.

uekerman:
I am very open for a different fix of the problem. The current state is surely easy to screw up again.

fsimonis:
@floli @uekerman So the resulting TODO list from this discussion is:

* Change this PR to receiving / sending `std::vector`
* Release it as Hotfix
* Change the implementation of the array functions to communicate and check the size to make them consistent with the vector variants. (New PR)

uekerman:
Well, this PR was merely to show the problem, not to propose the best possible fix. The underlying problem is what @fsimonis described, the inconsistent treatment of `size`.

Direct merge to master as it is a bug. Wrong data is sent. Btw, not only for `master:sockets` but also for `master:mpi-single`. Only the first led to an error. 

Who does a proper fix? We would also need much more unit and integration tests for these methods.

floli:
I will look into the receiving side to also use vector receiving/sending.

fsimonis:
I'll look into the M2N module and polish it up a bit.

BenjaminRueth:
I created an issue #453 related to testing of #451.

fsimonis:
So shall we use this version to hotfix preCICE?
I am fine doing the actual hotfix.

fsimonis:
I added the necessary changes to turn this into a proper bugfix.

BenjaminRueth:
I merged the fix into my systemtest. It seems to work 👍 

In the end the test is reported as failed, but this is due to some technical detail. The solverdummy runs through without throwing any errors. See [here](https://travis-ci.org/precice/systemtests/jobs/554573498#L2667-L2876).

PR #455, opened 2019-07-03
Add additional info in the connection stage

### LINKED ISSUE #444: Waiting for connection: no detailed status
When establishing communication, the user does not get any details on how many partners (and which) it is waiting for:

```
(0) 12:59:10 [impl::SolverInterfaceImpl]:214 in initialize: Setting up master communication to coupling partner/s
(0) 12:59:22 [impl::SolverInterfaceImpl]:219 in initialize: Masters are connected
```

The last message is shown only after every participant is connected (even with logger set with `filter="%Severity% >= trace"`).

In a muli-participant scenario, it would be very helpful to get a message such as `Found participant Inner-Fluid` and then `Found participant Outer-Fluid`. This way, it is easier to debug a case that cannot start.

Even further, it would be nice to have a status of how many participants have been connected (e.g. `Found participant Inner-Fluid [1/2]` and then `Found participant Outer-Fluid [2/2]`). Or, instead of `[1/2]`, write `[1 participant pending]` or `[waiting for participant/s: Outer-Fluid]`.

Printing the address and protocol used to find the masters (together with own address) would also be helpful (at least in the debug/trace mode), as it would give hints for wrong configuration or old address files. 

Example:

```
Setting up master communication to coupling partner/s: Inner-Fluid, Outer-Fluid
My IP address is 1.2.3.5
Looking for participant Inner-Fluid using TCP Sockets at IP address 1.2.3.4
Looking for participant Outer-Fluid using TCP Sockets at IP address 1.2.3.6
Found participant Inner-Fluid
Found participant Outer-Fluid
Masters are connected
```

fsimonis:
This PR adds some more information in the connection stage of the preCICE initialization.

Here is the log from running the solverdummies:

New | SolverOne | SolverTwo 
| --- | --- | --- |
|| Setting up master communication to coupling partner/s | Setting up master communication to coupling partner/s |
:star: | Establishing incoming connection from  master of SolverTwo | Establishing outgoing connection to master of SolverOne |
|| Masters are connected | Masters are connected |
|| Gather mesh MeshOne  | Receive global mesh MeshOne |
|| Send global mesh MeshOne | |
|| Compute partition for mesh MeshOne | Compute partition for mesh MeshTwo|
|| Setting up slaves communication to coupling partner/s | Setting up slaves communication to coupling partner/s|
|:star: | Establishing incoming connection from  slaves of SolverTwo | Establishing outgoing connection to slaves of SolverOne|
|| Slaves are connected | Slaves are connected |

Closes #444

### DIFF
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -213,8 +213,10 @@ double SolverInterfaceImpl:: initialize()
 
     INFO("Setting up master communication to coupling partner/s" );
     for (auto& m2nPair : _m2ns) {
-        m2nPair.second.prepareEstablishment();
-        m2nPair.second.connectMasters();
+        auto& bm2n = m2nPair.second;
+        INFO("Establishing " << (bm2n.isRequesting?"incoming connection from ":"outgoing connection to") <<  " master of " << bm2n.remoteName);
+        bm2n.prepareEstablishment();
+        bm2n.connectMasters();
     }
     INFO("Masters are connected");
 
@@ -222,8 +224,10 @@ double SolverInterfaceImpl:: initialize()
 
     INFO("Setting up slaves communication to coupling partner/s" );
     for (auto& m2nPair : _m2ns) {
-      m2nPair.second.connectSlaves();
-      m2nPair.second.cleanupEstablishment();
+      auto& bm2n = m2nPair.second;
+      INFO("Establishing " << (bm2n.isRequesting?"incoming connection from ":"outgoing connection to") <<  " slaves of " << bm2n.remoteName);
+      bm2n.connectSlaves();
+      bm2n.cleanupEstablishment();
     }
     INFO("Slaves are connected");
 
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -214,9 +214,10 @@ double SolverInterfaceImpl:: initialize()
     INFO("Setting up master communication to coupling partner/s" );
     for (auto& m2nPair : _m2ns) {
         auto& bm2n = m2nPair.second;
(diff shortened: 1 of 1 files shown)


floli:
This prints one additional line per M2N connection, which IIRC can be multiple ones per rank. It might just be too many lines printed on INFO log level for larger number of ranks.

uekerman:
There is one M2N connection per participant pair (i.e. per m2n in the config). INFO is only printed on the master anyway. Should not be soo much INFO logging, but @fsimonis just give it a try.

fsimonis:
@MakisH Thanks for trying this out!

I agree that logging begin and end makes sense, however. the resulting log looks textually very dense.
Should we consider shortening the text a bit?

fsimonis:
@MakisH The new version should output the following:
```
Setting up master communication to coupling partner/s
Awaiting master connection from Inner-Fluid
Connecting master from Inner-Fluid
Established master Connection from Inner-Fluid
Awaiting master connection from Outer-Fluid
Established master Connection from Outer-Fluid
Masters are connected

Setting up slaves communication to coupling partner/s
Awaiting slaves connection from Inner-Fluid
Connecting slaves from Inner-Fluid
Established slaves Connection from Inner-Fluid
Awaiting slaves connection from Outer-Fluid
Established slaves Connection from Outer-Fluid
Slaves are connected
```

MakisH:
What is the difference between "awaiting connection" and "connecting"?

I also don't understand the density argument. In the example above I only added one line to the existing output (the "established" part). Keep in mind that this is a case with three participants. Eitherway, this would only appear once in the beginning.

fsimonis:
@MakisH Sorry, I managed to muddle up the lines.
Its `awaiting` for incoming and `connecting` for outgoing connections.

Density is probably a bad choice of word. I mean that one can simplify and shorten the log statements.

MakisH:
I don't see how this output could be simplified without losing useful information (and complicating the output mechanism). It's fine like this.

PR #460, opened 2019-07-14
Initial commit to add empty string assertion in Cbindings

IshaanDesai:
User needs to be warned if the adapter passes empty strings in place of `participantName` and `configFileName` in the C bindings. This was observed as an issue in the fluent_adapter because FLUENT does not proceed in the event of such a case of SolverInterface initialization.


fsimonis:
This is a very good suggestion!

There is no reason to keep this restrict this to the C bindings. We should move this check to the native C++ interface.

fsimonis:
Closed in favour of a native implementation #461

PR #469, opened 2019-07-25
Add bulk methods for setting Triangles and Edges

### LINKED ISSUE #465: Adding bulk versions for setMeshEdge, setMeshTriangle
The preCICE API offers methods to set mesh vertices, edges and triangles.

We currently provide bulk versions only for adding vertices. We should add bulk versions for all primitives.
Bulk functions allow us to sanitize the entire data once at the interface and thus offers great runtime benefits during mesh creation.

### Status

| Primitve | Singular | Bulk |
| --- | --- | --- |
| Vertex | `setMeshVertex`  | `setMeshVertices` |
| Edge | `setMeshEdge` | :x: | 
| Triangle | `setMeshTriangle` | :x: | 
| Quad | `setMeshQuad` |

There are also `XWidthEdges()` versions of these.

### setMeshEdges
`void setMeshEdges(int meshID, int size, int const * vertexIDs, int * edgeIDs)`

`vertexIDs` has to point to a range of `2*size` vertex ids in the format `e0A e0B e1A e1B ...`
`edgeIDs` has to point to a range of `size` edge ids. The function writes the resulting edge ids to it in the format `e0 e1 e2 ...`

### setMeshTriangles
`void setMeshTriangles(int meshID, int size, int const * edgeIDs)`

`edgeIDs` has to point to a range of `3*size` edge ids in the format `t0AB t0BC t0CA t1AB t1BC ...`

### setMeshQuads
`void setMeshQuads(int meshID, int size, int const * edgeIDs)`

`edgeIDs` has to point to a range of `4*size` edge ids in the format `t0AB t0BC t0CD q0DA q1AB q1BC ...`

fsimonis:
This PR adds bulk counterparts for `setMeshTriangle` and `setMeshEdge`.

Closes #465

Reviewers, please check the changes to the SolverInterface as well as the implementation especially in the server parts.

# TODO
* [ ] Add tests
* [ ] Extend bindings

### DIFF
--- a/src/precice/SolverInterface.cpp
+++ b/src/precice/SolverInterface.cpp
@@ -190,6 +190,16 @@ int SolverInterface:: setMeshEdge
   return _impl->setMeshEdge ( meshID, firstVertexID, secondVertexID );
 }
 
+void SolverInterface:: setMeshEdges
+(
+  int meshID,
+  int size,
+  const int* vertexIDs,
+  int* edgeIDs )
+{
+    _impl->setMeshEdges(meshID, size, vertexIDs, edgeIDs);
+}
+
 void SolverInterface:: setMeshTriangle
 (
   int meshID,
@@ -198,6 +208,15 @@ void SolverInterface:: setMeshTriangle
   int thirdEdgeID )
 {
   _impl->setMeshTriangle ( meshID, firstEdgeID, secondEdgeID, thirdEdgeID );
+}
+
+void SolverInterface:: setMeshTriangles
+(
+  int meshID,
+  int size,
+  const int* edgeIDs )
+{
+    _impl->setMeshTriangles(meshID, size, edgeIDs);
 }
 
 void SolverInterface:: setMeshTriangleWithEdges
--- a/src/precice/SolverInterface.hpp
+++ b/src/precice/SolverInterface.hpp
@@ -466,6 +466,26 @@ public:
     int secondVertexID );
 
   /**
+   * @brief Sets multiple mesh edges from vertexIDs and writes the resulting ids to edgeIDs
+   *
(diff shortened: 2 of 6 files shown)


uekerman:
It was hard work to make all bindings consistent. I would not want to sacrifice that here. Could in principle also got into another PR, but should not be postponed to another release.   
* `c`  
* `f2003`  
* `fortran`  
* `PySolverInterface`  
* `python`  
* `python_future`

Furthermore, we need integration tests and binding tests (where possible, e.g. python).

fsimonis:
We also need tests here.

BenjaminRueth:
> It was hard work to make all bindings consistent. I would not want to sacrifice that here. Could in principle also got into another PR, but should not be postponed to another release.
> 
>     * `c`
> 
>     * `f2003`
> 
>     * `fortran`
> 
>     * `PySolverInterface`
> 
>     * `python`
> 
>     * `python_future`
> 
> 
> Furthermore, we need integration tests and binding tests (where possible, e.g. python).

Did we already decide on a strategy here?

1. add everything to this PR (-> more work in this PR)
2. add support for language bindings in a different PR (-> we should already open an issue, to make sure we do not forget it + make sure that it gets done in the same release like this PR)

For the python bindings I am happy to contribute, even though I would prefer to only update `python_future` and not add the new feature in the (deprecated) `python` bindings.

uekerman:
We can do the bindings in a different PR, but the integration tests for the new methods, we should do here.

fsimonis:
Conclusion:
* Not a must for 2.0.0
* Finish feature after #574 
* Extend all bindings
* Add a test

fsimonis:
I am closing this in favour of a new go.

PR #471, opened 2019-07-30
Replace assertion with BOOST_TEST

BenjaminRueth:
If we do not build preCICE in debug mode, `assertion` will be ignored. Therefore, we should use `BOOST_TEST` in our tests.


fsimonis:
I fixed the conflicts introduced in #476

PR #473, opened 2019-07-31
xSDK - M1 Installation Policies

fsimonis:
This PR is a first prototype of the build system changes required to be compliant with the [xSDK Community Installation Policies](https://figshare.com/articles/xSDK_Community_Installation_Policies_GNU_Autoconf_and_CMake_Options/4495133).

This prototype includes commits to
1) detect `CPP` and `CPPFLAGS`
2) support the `USE_XSDK_DEFAULTS`
3) turn JSON and prettyprint into separate targets _M12_
3) `TPL` overrides for:
    1) Eigen
    2) LibXML2
    3) Boost
    4) Python & NumPy
    5) PETSC
    6) JSON _M12_
    7) prettyprint _M12_

The latter `TPL` overrides force CMake to use a given set of variables to find a single dependency. If CMake cannot detect this dependency it must fail.

Some of these overrides require to set some internal variables which can confuse the find modules.
This is why I added an additional validation step for Python, NumPy and Eigen.


fsimonis:
I already extended our wiki regarding the [xSDK compliance](https://github.com/precice/precice/wiki/Building:-Using-CMake#xsdk-compliance)

fsimonis:
This PR got approved by xSDK. Thus, we merge this PR into develop so that we can use it in the Spack package.

PR #476, opened 2019-08-01
Fix macro namespace

fsimonis:
This PR prefixes all log and error macros with ~~`PRECICE_`~~ `P` which is part of the xSDK policy M9.

We should merge code-heavy PRs first as this will be quite a nightmare.


floli:
I still prefer to find a way to work around this requirement.

+ Use a namespace for macros. Is that possible?
+ Use `P_DEBUG` instead of `PRECICE_DEBUG`. Less stuff to type.

fsimonis:
> Use P_DEBUG instead of PRECICE_DEBUG. Less stuff to type.

Good idea. That should suffice as it is _a_ prefix.

> Use a namespace for macros. Is that possible?

Nope, there are no namespaces in pre-processor land. Therefore the prefixes

fsimonis:
As I touch every `assertion`, we could rename it to `P_ASSERT` to stay consistent with the remaining macros.

floli:
> As I touch every assertion, we could rename it to P_ASSERT to stay consistent with the remaining macros.

Yes, makes it more clear that `assertion` is a macro, too.

MakisH:
> > Use P_DEBUG instead of PRECICE_DEBUG. Less stuff to type.
> 
> Good idea. That should suffice as it is _a_ prefix.

It is _a_ prefix, but doesn't this go against the idea of the policy? If PETSc also defines `P_DEBUG`, we have a problem. I understand that this is a problem of the policy, however.

In any case, it would be of course easy to convert the one to the other once it's there.

fsimonis:
Lets merge this now.
Any other further change regarding the macros becomes trivial.

PR #477, opened 2019-08-02
Add getVersionInformation to API

fsimonis:
This PR is an iteration of #261 and makes it obsolete. It is based on #446.

This PR adds functionality to the build system to generate additional information about the version and configuration of preCICE.
This information is a semicolon-separated list of the format `VERSION;REVISION;CONFIGURATION`.
* `VERSION` is the preCICE version set as the project version and __always present__
* `REVISION` is the revision information from #446 and should always be present.
* `CONFIGURATION` is a semicolon-separated list containing:
    * dependency options:  `MPI=Y;PETSC=N;PYTHON=Y`
    * compiler information: `CXX=Clang` [from CMake](https://cmake.org/cmake/help/v3.10/variable/CMAKE_LANG_COMPILER_ID.html)
    * flags: `CXXFLAGS=-Wall -g;LDFLAGS=-lto=full`

Example from my compile-server:
```
1.5.1;no-info [Git failed/Not a repository];MPI=Y;PETSC=Y;PYTHON=Y;CXX=Clang;CXXFLAGS=-march=native -Wall -Wextra -Wpedantic -Wno-unused-parameter -Wno-deprecated-register -g;LDFAGS=
```


src/precice/SolverInterface.cpp
```diff
@@ -331,6 +332,11 @@ MeshHandle SolverInterface:: getMeshHandle
   return _impl->getMeshHandle ( meshName );
 }
 
+const char * getVersionInformation()
```
floli: Is `const char*` mandated by xSDK? Otherwise, in a C++ interface, I expect a `std::string`.

fsimonis: I agree. This will make it also easier to parse the string at the call-site.

cmake/cmake_refresh_git_revision.cmake
```diff
@@ -11,17 +11,19 @@
 #
 
 find_package(Git REQUIRED)
-set(preCICE_REVISION "no-info [Git failed/Not a repository]")
 execute_process(
   COMMAND ${GIT_EXECUTABLE} describe --tags --broken
```
MakisH: FYI, this does not work on Ubuntu 16.04, but it's fine as it should work for 18.04 (the oprion was added in between):

```
-- Found Git: /usr/bin/git (found version "2.7.4") 
error: unknown option `broken'
usage: git describe [<options>] [<commit-ish>...]
   or: git describe [<options>] --dirty

    --contains            find the tag that comes after the commit
    --debug               debug search strategy on stderr
    --all                 use any ref
    --tags                use any tag, even unannotated
    --long                always use long format
    --first-parent        only follow first parent
    --abbrev[=<n>]        use <n> digits to display SHA-1s
    --exact-match         only output exact matches
    --candidates <n>      consider <n> most recent tags (default: 10)
    --match <pattern>     only consider tags matching <pattern>
    --always              show abbreviated commit object as fallback
    --dirty[=<mark>]      append <mark> on dirty working tree (default: "-dirty")

-- Revision status: Detection failed
[  0%] Built target GitRevision
```

fsimonis: As `--broken` is not necessary, I reverted to using `--dirty`. So the detection will now fail when the repo is broken.
I think that is a negligible corner case and well worth the compatibility with ubuntu 16.04.

PR #481, opened 2019-08-10
Make Pythonactions compatible with Python 2 and 3

fsimonis:
This PR replaced deprecated Python 2 calls.
The result is compatible to both Python 2 and Python 3.

We either need to migrate our tests to python3 or test both Python 2 and 3.


fsimonis:
I cannot reproduce nor understand the error on Travis.

It puzzles me why it cannot load the python action modules, especially as they are compatible with python 2 and 3.

I even used a fresh ubuntu:16.04 container and executed the commands specified in the `.travis.yml` and I could not reproduce the error.
As a last resort, we could override the CMake Python variables with a fixed set.

fsimonis:
As it turns out, it helps immensely to install the correct packages ....

Issue #225, opened 2018-12-29
Brute-force re-initialization


uekerman:
This is my long-promised first attempt to elaborate on the big picture on how to realize a brute-force re-initialization. Feel free to extend the issue.  
It might turn out more complicated than I thought, especially the triggering of the re-initialization, where we need to carefully weigh the consequences of different solutions. 

# The story

"As a user of preCICE, I want to adapt my coupling mesh during a simulation, such that I can handle re-meshing of my solver (or load-balancing or dynamically adaptive mesh refinement) without restarting the simulation."  

* _Re-initialization_ means that user can adapt the mesh during a coupled simulation, define new mesh primitives, delete old ones
* _Brute-force_ means for the user that only the complete mesh can be marked invalid
* _Brute-force_ means for the developer that we don't keep temporary information, but throw away all data structures and set them up again from scratch. This means we don't worry about performance making this the baseline for further improvements. This includes:
  * Re-partitioning
  * Slaves-slaves communication
  * QN information from previous timesteps

## Assumptions for simplification

* Only two coupled participants

## Open questions

* Do we allow re-init within an implicit coupling loop or only after a timestep is completed?
  * I currently don't see any additional complexity when allowing this within an implicit loop.
* Do we allow re-init within subcycling?
  * Probably no big technical issue (we could simply use the last changed mesh)

Anyhow: Let's not consider the general case at first, so no subcycling, no changes during implicit coupling 

## Historical remarks 

* This story was possible with the old serial version of preCICE (way back, see [Bernhard's thesis](https://www5.in.tum.de/pub/Gatzhammer2014_preCICE.pdf))
* It was much simpler back than:
  * No need to update partitions and slaves-slaves communication channels
  * Thus, it was sufficient if only one participant knew about this.

# Status quo

What happens currently in initialize() and what we have to re-do for re-initialization. 

* establish master-master communication -> we could, but we don't want
* compute partitions -> **we have to** (attention: also allocates data values)
* establish slave-slave communication -> **we have to**
* initialize watchpoints
  * find closest vertex -> **we have to** 
  * set up output file -> we cannot
* initialize coupling scheme
  * set start time -> we could, but we don't want  
  * set up convergence measures (coupling data gets assigned, but not set up of data structures) -> we could, but we don't want to
  * set up data matrices (reserve memory for old values for c.m. and extrapolation) -> **we have to**
  * initialize pp (reserve memory, QN columns etc.) -> **we have to**
[...]

PR #486, opened 2019-08-14
Add Watchpoint Reinitialisation

fsimonis:
This PR 
* changes `WatchPoint` so that it can be reinitialized,
* adds the required `TXTTableWriter::reset()`.
* rework watchpoint test to check time series
* add watchpoint test to check reinitialization

This is a part of #225

### DIFF
--- a/src/io/TXTTableWriter.cpp
+++ b/src/io/TXTTableWriter.cpp
@@ -22,13 +22,6 @@ TXTTableWriter::TXTTableWriter(
   _outputStream << std::setprecision(16);
 }
 
-TXTTableWriter::~TXTTableWriter()
-{
-  if (_outputStream.is_open()) {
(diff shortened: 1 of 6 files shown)


uekerman:
Looks correct now. Still, to be on the safe side, it would be good if the `WatchPointTest` covers a case where the modified mesh also modifies the watched value.

fsimonis:
I added the requested case. The tests now actually check the output.

fsimonis:
This PR needs a complete rework after the code reformat and the change of the test framework

fsimonis:
@uekerman I reworked the branch and added a test to check the reinitalization.
The new tests has two watchpoints that project on the point of the same edge. Then it adds an edge so that one of the two points projects to the new edge.

PR #491, opened 2019-08-22
Fix build with PETSc below 3.7 

shkodm:
In 3.7 they changed number of arguments in certain functions ( see [here](https://www.mcs.anl.gov/petsc/documentation/changes/37.html) ), leading to build failure on e.g Ubuntu 16.04 with packaged PETSc. This fixes it by introducing a wrapper that determines number of arguments to to `PetscOptionsSetValue` on compile time, and calls corresponding function. 
I tested on Ubuntu 16.04  (PETSc 3.6.2 ) and Ubuntu 18.04 .


shkodm:
> Could you please add some description of what this does. At some point we will forget.

Done

PR #496, opened 2019-09-03
require n x d arrays and remove all strict types

### LINKED ISSUE #441: Python bindings: How strict do we want to be with input argument types?
Cython supports types in the signature of our functions. This is an optional feature which allows the following two extremes for `write_block_scalar_data`:

1. **Strict types:** `def write_block_scalar_data (self, const int data_id, const int[:] value_indices, const double[:] values)`
2. **No types:** `def write_block_scalar_data (self, data_id, value_indices, values)`

## Strict types

Using the first option, we always know which type of data the bindings have to deal with. This is especially beneficial when it comes to accessing the cpp interface of preCICE that accepts `double*`, where many things can go wrong, if the python bindings provide wrongly typed numpy arrays. However, this means also that the users have to use the types that are accepted by the bindings, since otherwise a runtime error will be thrown. This is very uncommon for python and makes code on the user side harder to write.

## No types

With the second option, we do not know what kind of data the python bindings have to deal with. Often it becomes necessary to reformat the input data before further processing is possible. 

An example of what can go wrong with wrongly formatted data, can be found [here](https://travis-ci.org/precice/precice/jobs/547845333): the input data is not-c-contiguous, which results in wrong behaviour. A possible solution for this problem is shown [here](https://github.com/precice/precice/blob/43a5a8925934a9142e2433db7f571036c83f7abb/src/precice/bindings/python_future/precice_future.pyx#L197-L202): The non-c-contiguous input is copied into a new np.array, which is c-contiguous and typed correctly.

**notes for implementation:**

* using the type `np.ndarray[double, ndim=1, mode="c"] values` checks whether input data is C-contiguous. A `ValueError: ndarray is not C-contiguous` is thrown, if this is not the case.
* one cannot use `const np.ndarray` in cython
* using a memory view `double[:]` allows us to use the keyword `const`, but we cannot check whether input data is C-contiguous.

### LINKED ISSUE #458: Use n x d arrays in python bindings
This issue originates from https://github.com/precice/precice/pull/448#discussion_r301218086.

As the python bindings are returning and accepting numpy arrays, we can easily use 2-D arrays with dimensionality n x d (degrees of freedom times dimension) instead of 1-D arrays of length d*n. In my opinion this would be easier to use and harder to break than 1-D arrays. 

One example where things can go wrong is the that the user has to decide between C- and Fortran order (column vs. row major) when linearizing a 2-D array. From the perspective of the interface it is impossible to tell whether the right decision has been made by the user, but if user input is wrongly formatted preCICE will have a problem. One could argue that for a 2D array the same confusion can happen if d == n. However, usually n >> d. In these cases the interface can easily detect the correct order of traversal. We could even add a warning cases when d == n, which will most likely never be issued.

There are two possible interfaces:

1. only accept 2D arrays
2. accept 1D and 2D arrays, let the bindings check which case we are dealing with and perform consistency checks.

Possibility 1. would be API breaking, while 2. would not.

BenjaminRueth:
closes #458
closes #441

* require n x d arrays for python_future bindings #458
* remove all strict types #441


src/precice/bindings/python_future/precice_future.pyx
```diff
@@ -98,134 +98,140 @@ cdef class Interface:
...
 
     # returns a handle to a created mesh
-    def get_mesh_handle(self, str mesh_name):
+    def get_mesh_handle(self, mesh_name):
         raise Exception("The API method get_mesh_handle is not yet available for the Python bindings.")
 
     # creates a mesh vertex
-    def set_mesh_vertex(self, int mesh_id, position):
+    def set_mesh_vertex(self, mesh_id, position):
+        dimensions = position.size
```
fsimonis: Should be able to handle tuple and list

src/precice/bindings/python_future/precice_future.pyx
```diff
@@ -98,134 +98,140 @@ cdef class Interface:
...
         return self.thisptr.getMeshVertexSize(mesh_id)
 
     # creates multiple mesh vertices
-    def set_mesh_vertices (self, int mesh_id, positions):
-        cdef np.ndarray[double, ndim=1] _positions = np.ascontiguousarray(positions, dtype=np.double)
-        size = _positions.size/self.get_dimensions()
-        assert(size.is_integer())
-        size = int(size)
+    def set_mesh_vertices (self, mesh_id, positions):
+        size, dimensions = positions.shape
```
fsimonis: Should be able to handle tuples and lists

src/precice/bindings/python_future/precice_future.pyx
```diff
@@ -98,134 +98,140 @@ cdef class Interface:
...
+        return _positions.reshape((size, self.get_dimensions()))
 
     # gets mesh vertex IDs from positions
-    def get_mesh_vertex_ids_from_positions (self, int mesh_id, positions):
-        cdef np.ndarray[double, ndim=1] _positions = np.ascontiguousarray(positions, dtype=np.double)
-        size = _positions.size/self.get_dimensions()
-        assert(size.is_integer())
-        size = int(size)
+    def get_mesh_vertex_ids_from_positions (self, mesh_id, positions):
+        size, dimensions = positions.shape
```
fsimonis: Should be able to handle tuples and lists

src/precice/bindings/python_future/precice_future.pyx
```diff
@@ -98,134 +98,140 @@ cdef class Interface:
...
+    def map_read_data_to (self, to_mesh_id):
         self.thisptr.mapReadDataTo (to_mesh_id)
 
-    def map_write_data_from (self, int from_mesh_id):
+    def map_write_data_from (self, from_mesh_id):
         self.thisptr.mapWriteDataFrom (from_mesh_id)
 
-    def write_block_vector_data (self, int data_id, value_indices, values):
+    def write_block_vector_data (self, data_id, value_indices, values):
+        size, dimensions = values.shape
```
fsimonis: Should be able to handle tuples and lists

src/precice/bindings/python_future/precice_future.pyx
```diff
@@ -98,134 +98,140 @@ cdef class Interface:
...
-        cdef np.ndarray[double, ndim=1] _values = np.ascontiguousarray(values, dtype=np.double)
-        assert(_values.size / self.get_dimensions() == _value_indices.size)
+        cdef np.ndarray[double, ndim=1] _values = np.ascontiguousarray(values.flatten(), dtype=np.double)
+        assert(size == _value_indices.size)
         size = value_indices.size
         self.thisptr.writeBlockVectorData (data_id, size, &_value_indices[0], &_values[0])
 
-    def write_vector_data (self, int data_id, value_index, value):
+    def write_vector_data (self, data_id, value_index, value):
+        dimensions = value.size
```
fsimonis: Should be able to handle tuples and lists

src/precice/bindings/python_future/precice_future.pyx
```diff
@@ -98,134 +98,140 @@ cdef class Interface:
...
         self.thisptr.writeBlockScalarData (data_id, size, &_value_indices[0], &_values[0])
 
-    def write_scalar_data (self, int data_id, int value_index, double value):
+    def write_scalar_data (self, data_id, value_index, double value):
         self.thisptr.writeScalarData (data_id, value_index, value)
 
-    def read_block_vector_data (self, int data_id, value_indices):
+    def read_block_vector_data (self, data_id, value_indices):
         cdef np.ndarray[int, ndim=1] _value_indices = np.ascontiguousarray(value_indices, dtype=np.int32)
         size = _value_indices.size
```
fsimonis: Should be able to handle tuples and lists

BenjaminRueth: I do not think that there is any problem in this function.

src/precice/bindings/python_future/precice_future.pyx
BenjaminRueth: done.

src/precice/bindings/python_future/precice_future.pyx
BenjaminRueth: done.

src/precice/bindings/python_future/precice_future.pyx
BenjaminRueth: done.

src/precice/bindings/python_future/precice_future.pyx
BenjaminRueth: done.

src/precice/bindings/python_future/precice_future.pyx
BenjaminRueth: done.

BenjaminRueth:
`git revert 1cee6adba72ce42dbd76fdf30be60adc740e93a2` shows that the new tests even work ;)

src/precice/bindings/python_future/test/SolverInterface.cpp
```diff
@@ -186,7 +186,11 @@ void SolverInterface:: getMeshVertexIDsFromPositions
   int           size,
   const double* positions,
   int*          ids ) const
-{}
+{
+  for(int i = 0; i < size; i++){
```
fsimonis: You could use the infamous [`std::iota`](https://en.cppreference.com/w/cpp/algorithm/iota) here. :wink: 
`std::iota(ids, ids+size, 0)`

src/precice/bindings/python/precice/precice.py
```diff
@@ -57,8 +57,9 @@ def get_data_id(self, data_name, mesh_id):
         return self.interface.get_data_id(data_name, mesh_id)
 
     def set_mesh_vertices(self, mesh_id, size, positions, ids):
-        assert(positions.size == size * self.get_dimensions())
-        out_ids = self.interface.set_mesh_vertices(mesh_id, positions)
+        dimensions = self.get_dimensions()
+        assert(positions.size == size * dimensions)
+        out_ids = self.interface.set_mesh_vertices(mesh_id, positions.reshape((size, dimensions)))
```
uekerman: Maybe this is a stupid comment, but don't you have to always do the inverse operation? So `flatten` here instead of `reshape`?

BenjaminRueth: For the `vertex_ids` (scalar valued) we only get a array of dimension `n`. `flatten` and `reshape` are only needed for vector valued quantities like `positions` or `vector_data`.

src/precice/bindings/python_future/test/SolverInterface.cpp
BenjaminRueth: Done: https://github.com/precice/precice/pull/496/commits/0eafc785ff4dcff403823a74a69ce82fea585520

PR #502, opened 2019-09-09
Fix MPI initialization timing

### LINKED ISSUE #499: Attempt to get MPI_Comm_rank() in serial simulations with PETSc
With ~~https://github.com/precice/precice/pull/382~~ https://github.com/precice/precice/commit/ae1a747aeed7950cc3aa800bbb59ac2024ff760a, an error for serial simulations that use PetRBF was introduced:

```
*** The MPI_Comm_rank() function was called before MPI_INIT was invoked.
*** This is disallowed by the MPI standard.
*** Your MPI job will now abort.
[Magdalena:6012] Local abort before MPI_INIT completed successfully; not able to aggregate error messages, and not able to guarantee that all other processes were killed!
```

Parallel simulations work fine.

You can reproduce this using any solver dummy or adapter and setting PetRBF mapping, e.g.:
```xml
      <mapping:petrbf-thin-plate-splines direction="write" from="MeshTwo" to="MeshOne" constraint="conservative"/>
      <mapping:petrbf-thin-plate-splines direction="read"  from="MeshOne" to="MeshTwo" constraint="consistent" />
```
This also affects our current FSI tutorials. This passed through as we don't yet have a system test that checks PetRBF, the first of which we are going to introduce in https://github.com/precice/systemtests/pull/85.

Could maybe the error originate from this?

https://github.com/fsimonis/precice/blob/51542b11e39f67cb55e60a7a7d0041af96bf0308/src/utils/Petsc.cpp#L66

fsimonis:
Closes #499

### DIFF
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -100,6 +100,7 @@ void SolverInterfaceImpl:: configure
 (
   const std::string& configurationFileName )
 {
+  utils::Parallel::initializeMPI(nullptr, nullptr);
   config::Configuration config;
   xml::configure(config.getXMLTag(), configurationFileName);
   if(_accessorProcessRank==0){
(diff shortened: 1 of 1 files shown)


uekerman:
Makes sense.

PR #508, opened 2019-09-11
WIP: Sort out dependencies of python bindings

BenjaminRueth:
setuptools provides a system to properly define and install dependencies of a python package. Currently, the python bindings do not properly implement this system for all the dependencies.

Our preCICE python bindings have the following dependencies:

* cython (for building)
* numpy (needed by cython to be able to deal with numpy arrays, needed for np.ascontigousarray, probably also at some more places)
* more?

Usually installation of the python bindings works (see systemtests). However, if a system is not set up properly, one might run into problems. Some examples:

* user never installed `numpy` using `sudo apt-get` -> `cython` will not be able to find `arrayobject.h` (currently, all our [precice docker baseimages install numpy](https://github.com/precice/systemtests/blob/32fdbbbc7d35f2395ee394dc8113d538b3dd86a1/precice/Dockerfile.Ubuntu1604.home#L13). Therefore, we never actually see this problem)
* user never installed `numpy` using `pip3` -> when we run the bindings, we might get into trouble.

I already added some new dependency checks in this PR. However, we should review these changes thoroughly and also modify the systemtests correspondingly.

Related: https://github.com/precice/systemtests/issues/94

### DIFF
--- a/src/precice/bindings/python_future/pyproject.toml
+++ b/src/precice/bindings/python_future/pyproject.toml
@@ -1,3 +1,3 @@
 [build-system]
 # PEP 518 - minimum build system requirements
-requires = ["setuptools", "wheel", "Cython>=0.29"]
+requires = ["setuptools", "wheel", "Cython>=0.29", "numpy"]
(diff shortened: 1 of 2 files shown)


BenjaminRueth:
Branch for corresponding systemtests: https://github.com/precice/systemtests/tree/improve_python_bindings_dependencies

BenjaminRueth:
Related: https://github.com/precice/precice/pull/587

BenjaminRueth:
This PR should be moved to https://github.com/precice/python-bindings. As soon as it is moved, we can close it.

BenjaminRueth:
Closing this issue, since we moved the python bindings to https://github.com/precice/python-bindings/.

Issue #474, opened 2019-08-01
Configuration duplication in preCICE and adapters: mesh and read/write data


MakisH:
# Problem

At the moment, there are options that need to be set both in the preCICE configuration file and in every adapter's configuration file, in exactly the same way.

This duplication can lead to errors, requires more checks from preCICE/adapters and reduces usability. Even if the user should understand what she is doing, she should not need to adapt to our technical complexity, but we should make the system more tight and less complex.

# An example

Let's look at an example: the OpenFOAM-OpenFOAM CHT tutorial with nearest-projection mapping. In the Fluid participant, we set:

https://github.com/precice/openfoam-adapter/blob/88dfbbae21a0dbdc9f21357a2a1d5d1954646a61/tutorials/CHT/flow-over-plate/buoyantPimpleFoam-laplacianFoam_nearest-projection/Fluid/precice-adapter-config.yml#L5-L17

The `read-data` and `write-data` are also already known to preCICE, as they are set in the configuration file:

https://github.com/precice/openfoam-adapter/blob/88dfbbae21a0dbdc9f21357a2a1d5d1954646a61/tutorials/CHT/flow-over-plate/buoyantPimpleFoam-laplacianFoam_nearest-projection/precice-config_serial.xml#L31-L36

Ideally, we could directly get groups of parameters, such as: {mesh, read/write, data name}. Although we still need to associate each solver interface with a part of the preCICE configuration file. As soon as this association is established, there should be no duplication.

Moreover, the `provideMeshConnectivity` boolean is a technical thing that the user should not need to know about, as preCICE knows if it needs it. This is described in issue #467. As I wrote on the same issue:
> I actually like this idea also for consistency with the high-level API idea. For example, the user does not need to configure an adapter to keep checkpoints when asking for implicit coupling. It's rather the adapter that asks preCICE if this is needed. Why not do the same here?

This is also related to #311, in the sense that we duplicate information that preCICE already knows.

# Potential solution

Implent additional methods in the API, which return what a participant should read/write on which mesh and which interface.

uekerman:
I agree that we could simplify the life of our users here. 

For simple cases, such as `provideMeshConnectivity`, there are straight-forward solutions that are certainly worth the effort #467. 

For handling coupling meshes and data, things are more complicated. 

Back in the old days, Bernhard had already kind of a solution to this:
* In `Constants.hpp` we define default names for data types, such as `Displacements`, `Forces`, ... and also a default name for the preCICE config. 
[...]

PR #518, opened 2019-09-12
Add API functions to simplify adapter configuration

uekerman:
As discussed in #474.

TODO:

* [ ] Add patch names to configuration
* [ ] Integration test
* [ ] Test with OpenFOAM adapter


uekerman:
Closing since too much out-of-date.
Maybe standardization of adapter configs makes #474 also less painful.

PR #528, opened 2019-09-20
[debug] subroutine interface `precicef_has_data`'s args' order

RPGP1:
change the order of the arguments of subroutine interface
`precicef_has_data` to be consistent to the corresponding c function
declared in `src/precice/bindings/fortran/SolverInterface.hpp`.


uekerman:
## C++
```c++
bool hasData ( const std::string& dataName, int meshID )
```

## C
```c
int precicec_hasData ( const char* dataName, int meshID );
```

## fortran

```c
void precicef_has_data_(const char* dataName, const int* meshID, int* hasData, int lengthDataName);
```

## f2003

* old
```fortran
subroutine precicef_has_data(dataName, hasData, lengthDataName, meshID)
```

* suggested
```fortran
subroutine precicef_has_data(dataName, meshID, hasData, lengthDataName)
```

Sounds good! Thanks for noticing! Maybe it's even wrong currently. 

Probably, we also need to update the rest of the block:
https://github.com/precice/precice/blob/76c08d0c86ec4b3121497c8d2b85267011825af7/src/precice/bindings/f2003/SolverInterfaceF2003.f90#L239-L244

RPGP1:
I don't think additional updates are necessary.
F2003 doesn't change the subroutine's behavier depending on the definition order of variables (including arguments); It's been verified in my project.
However, the readability improves if the order is aligned. I did (pushing).

MakisH:
I am working with the Fortran bindings at the moment, so I will have a detailed look for v2.0.0. Thanks for pointing this out, please keep an eye open for any other similar issues!

PR #494, opened 2019-08-29
Add MATLAB Bindings

BenjaminRueth:
Closes #102.

This branch currently contains all the changes by @Dominanz that were necessary for his thesis. One particularly useful contribution for preCICE are the MATLAB bindings that have been developed by @Dominanz. We want to merge @Dominanz implementation of the MATLAB bindings into https://github.com/precice/precice within this PR.

I would suggest to release the MATLAB bindings as soon as possible in order to be able to provide them to our users. However, I would suggest to explicitly declare them as **experimental** since I expect that we will need to make some experiences until we arrive at the final version of the Interface.

# Requirements

We require the following components for the MATLAB bindings:

* all source code that is necessary for installation of the MATLAB bindings
* documentation, such that a user can install the MATLAB bindings
* a solverdummy using the MATLAB bindings

# Todo

Before we can merge this PR, we have to do the following things:

- [x] get rid of everything that is not directly connected to the MATLAB bindings (see requirements above)
- [x] add documentation, if it is missing (-> documentation is in good shape, more will follow in a later PR)
- [x] write tests for the MATLAB bindings (might also be postponed after merging, since we might need to test the MATLAB bindings using a system test on https://github.com/precice/systemtests/tree/master/Test_bindings)
- [x] update the [Changelog](https://github.com/precice/precice/blob/develop/CHANGELOG.md)
- [x] update the [wiki](https://github.com/precice/precice/wiki/Non%E2%80%93standard-APIs) as soon as PR is merged

PR #530, opened 2019-09-26
Use const double& for writeScalarData

BenjaminRueth:
This change was introduced in https://github.com/precice/precice/commit/be65dcec131ac7f249b82146dd6a75ec8aeba7ad. I first assumed that it was just forgotten in https://github.com/precice/precice/pull/348. However, in https://github.com/precice/precice/commit/e480aef4e3a74ccd405f03a2770bee18cdb548e9 it was intentionally removed.

In any case we should avoid reintroducing this change via #494. And, if we want to introduce the change, introduce it independently. @fsimonis can you enlighten me?

### DIFF
--- a/src/precice/SolverInterface.cpp
+++ b/src/precice/SolverInterface.cpp
@@ -290,9 +290,9 @@ void SolverInterface:: writeBlockScalarData
 
 void SolverInterface:: writeScalarData
 (
-  int    dataID,
-  int    valueIndex,
-  double value )
+  int           dataID,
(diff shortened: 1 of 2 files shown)


fsimonis:
@BenjaminRueth Can you explain the problem?

Passing a single double via const reference is pointless and changing it is backwards-compatible.

BenjaminRueth:
@fsimonis if it is backwards compatible and passing a double value via a const reference is pointless, everything is fine as it is.

PR #533, opened 2019-10-06
Refactor two-level initialization, aka WP6

atotoun:
This package is aimed to refactor the changes in WP1 to WP5 and incorporate them in solver interface in a configurable way.


fsimonis:
The test `PreciceTests/Serial/TestExplicit` runs into the following acception on all ranks:
```
What:  precice::m2n::GatherScatterCommunication::acceptPreConnection(const std::string &, const std::string &) 
Where: /home/ga36qid/fast/precice/repo/src/m2n/GatherScatterCommunication.cpp:169
```

To reproduce and debug run:
```
mpirun -np 2 xterm -e "gdb -ex 'r --log_level=test_suite --run_test=PreciceTests/Serial/TestExplicit' ./testprecice
```

atotoun:
Hi, This package is not ready for review yet. I will let you know whenever it's ready!

src/mesh/Mesh.cpp
```diff
@@ -416,6 +416,42 @@ void Mesh:: computeState()
   computeBoundingBox();
 }
 
+void Mesh:: buildBoundingBox()
```
uekerman: What does this method do compared to `computeBoundingBox()`. Why are both needed? Why are the normals computed here?

src/partition/Partition.hpp
```diff
@@ -39,6 +39,12 @@ class Partition
 
   virtual ~Partition() {}
 
+  /// The mesh is communicated between both master ranks (if required)
+  virtual void communicateBoundingBox() = 0;
+
+  /// The partition is computed, i.e. the mesh re-partitioned if required and all data structures are set up.
+  virtual void computeBoundingBox() = 0;
+
```
uekerman: Please update comments

src/precice/tests/ParallelTests.cpp
```diff
@@ -47,7 +47,7 @@ struct ParallelTestFixture : testing::WhiteboxAccessor {
 BOOST_AUTO_TEST_SUITE(PreciceTests)
 BOOST_FIXTURE_TEST_SUITE(Parallel, ParallelTestFixture)
 
-
+/*
```
uekerman: Why is this test altered and removed? It should still work, right? It is very similar to the tests below, but still could be good to keep it as below we will probably test the new BB initialization and here the old one.

src/m2n/PointToPointCommunication.cpp
```diff
@@ -580,40 +580,45 @@ void PointToPointCommunication::send(double const  *itemsToSend,
...
         buffer->push_back(itemsToSend[index * valueDimension + d]);
       }
-    }
+    }    
     auto request = _communication->aSend(*buffer, mapping.remoteRank);
-    bufferedRequests.emplace_back(request, buffer);
+    bufferedRequests.emplace_back(request, buffer);    
   }
-  checkBufferedRequests(false);
+  checkBufferedRequests(true);
```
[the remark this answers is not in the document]
atotoun: Yes, ture. This was just to check whether buffering has an impact on the order of communcated data. 
No this is just a left over from an experiment.

src/mesh/Mesh.cpp
atotoun: We needed a method to compute bounding boxes only and can be called from solverInterface. If I remember correctly, some time a go (maybe version 1.3) we had only one big method computeState, but now this method calls two other methods: computeBoundBox and computeNormals. At that time I extracted the parts we needed and wrapped them into buildBoundingBox.

In addition, it seems that we do not need the normals here. 

But now, computeBoundingBox() does the samething, the only problem is it is a private method and can't be called from solverInterface. We can either make it public or include the mesh.hpp in solverInterface. right?

uekerman: We should avoid the code duplication. Let's make `computeBoundingBox()` public then. I will anyway iterate over `computeState()` again, I guess.

uekerman:
Ready for review.
Currently two-level init is disabled by default and there are a few cases where it still might crash. I would, therefore, tell users that the feature is still in beta. I collected further TODOs in #633. 
@fsimonis Please have a birds-eye look. Anything we could easily improve? Anything preventing us from merging now?
@atotoun Please review the changes I made. Do your cases still run?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -193,22 +193,38 @@ double SolverInterfaceImpl::initialize()
...
+  }
+
   computePartitions();
 
   PRECICE_INFO("Setting up slaves communication to coupling partner/s");
   for (auto &m2nPair : _m2ns) {
     auto &bm2n = m2nPair.second;
-    PRECICE_DEBUG((bm2n.isRequesting ? "Awaiting slaves connection from " : "Establishing slaves connection to ") << bm2n.remoteName);
-    bm2n.connectSlaves();
+    if (bm2n.m2n->usesTwoLevelInitialization()) {
```
fsimonis: Same as above

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -193,22 +193,38 @@ double SolverInterfaceImpl::initialize()
...
   }
+
   PRECICE_INFO("Masters are connected");
 
+  compareBoundingBoxes();
+
+  PRECICE_INFO("Setting up preliminary slaves communication to coupling partner/s");
+  for (auto &m2nPair : _m2ns) {
+    auto &bm2n = m2nPair.second;
+    if (bm2n.m2n->usesTwoLevelInitialization()) {
```
fsimonis: Accessing the m2n of the BoundM2N is quite ugly. Can we simplify this?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1146,24 +1167,42 @@ void SolverInterfaceImpl::configurePartitions(
...
   //We need to do this in two loops: First, communicate the mesh and later compute the partition.
   //Originally, this was done in one loop. This however gave deadlock if two meshes needed to be communicated cross-wise.
   //Both loops need a different sorting
+
   auto &contexts = _accessor->usedMeshContexts();
 
-  // sort meshContexts by name, for communication in right order.
   std::sort(contexts.begin(), contexts.end(),
-            [](MeshContext const *const lhs, MeshContext const *const rhs) -> bool {
+            [](MeshContext *lhs, const MeshContext *rhs) -> bool {
```
fsimonis: We should keep the first argument const.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1172,6 +1211,8 @@ void SolverInterfaceImpl::computePartitions()
   for (MeshContext *meshContext : contexts) {
     meshContext->partition->compute();
     meshContext->mesh->computeState();
+    if (not meshContext->provideMesh) // received mesh can only compute their bounding boxes here
+      meshContext->mesh->computeBoundingBox();
```
fsimonis: Can you wrap this in braces for the sake of readability?

src/partition/ProvidedPartition.cpp
```diff
@@ -22,122 +24,274 @@ void ProvidedPartition::communicate()
...
-      PRECICE_ASSERT(utils::MasterSlave::getRank() == 0);
-      PRECICE_ASSERT(utils::MasterSlave::getSize() > 1);
+  bool twoLevelInitAlreadyUsed = false;
 
-      for (int rankSlave = 1; rankSlave < utils::MasterSlave::getSize(); rankSlave++) {
-        com::CommunicateMesh(utils::MasterSlave::_communication).receiveMesh(globalMesh, rankSlave);
-        PRECICE_DEBUG("Received sub-mesh, from slave: " << rankSlave << ", global vertexCount: " << globalMesh.vertices().size());
-      }
-    }
+  for (auto m2n : _m2ns) {
```
fsimonis:
```suggestion
  for (auto& m2n : _m2ns) {
```
Otherwise, the m2n pointer gets copied on each iteration.

src/partition/ProvidedPartition.cpp
```diff
@@ -22,122 +24,274 @@ void ProvidedPartition::communicate()
...
-  for (mesh::Vertex &v : _mesh->vertices()) {
-    v.setOwner(true);
+  if (_m2ns.empty())
+    return;
+
+  //@todo coupling mode
+
+  //@todo treatment of multiple m2ns
+  if (not _m2ns[0]->usesTwoLevelInitialization())
+    return;
```
fsimonis: Maybe use an assertion here to make sure that we don't run into this case.

src/partition/ProvidedPartition.cpp
```diff
@@ -22,122 +24,274 @@ void ProvidedPartition::communicate()
...
+
+  // each rank sends its bb to master
+  if (utils::MasterSlave::isSlave()) { //slave
```
fsimonis: Other code handles the master case first.

src/partition/ProvidedPartition.cpp
```diff
@@ -22,122 +24,274 @@ void ProvidedPartition::communicate()
...
+    com::CommunicateBoundingBox(utils::MasterSlave::_communication).sendBoundingBox(_mesh->getBoundingBox(), 0);
+  } else { // Master
+
+    PRECICE_ASSERT(utils::MasterSlave::getRank() == 0);
+    PRECICE_ASSERT(utils::MasterSlave::getSize() > 1);
```
fsimonis: Asserting `isMaster()` should suffice here.

src/partition/ProvidedPartition.cpp
```diff
@@ -22,122 +24,274 @@ void ProvidedPartition::communicate()
...
+
+    // to store the collection of bounding boxes
+    mesh::Mesh::BoundingBoxMap bbm;
+
+    // master stores its bb into bbm
+    bbm[0] = _mesh->getBoundingBox();
+
+    // master receives bbs from slaves and stores them in bbm
+    for (int rankSlave = 1; rankSlave < utils::MasterSlave::getSize(); rankSlave++) {
+      // initialize bbm
```
fsimonis:
```suggestion
      // @todo implement this as a simple assignment
      // initialize bbm
```

src/partition/ProvidedPartition.cpp
```diff
@@ -22,122 +24,274 @@ void ProvidedPartition::communicate()
...
+  std::map<int, std::vector<int>> remoteConnectionMap;
+
+  if (utils::MasterSlave::isMaster()) {
+
+    // master receives feedback map (map of other participant ranks -> connected ranks at this participant)
+    // from other participants master
+    _m2ns[0]->getMasterCommunication()->receive(connectedRanksList, 0);
+    remoteConnectionMapSize = connectedRanksList.size();
+
+    for (auto &rank : connectedRanksList) {
```
fsimonis:
```suggestion
    for (int rank : connectedRanksList) {
```

src/partition/ProvidedPartition.cpp
```diff
@@ -22,122 +24,274 @@ void ProvidedPartition::communicate()
...
+      remoteConnectionMap[rank] = {-1};
+    }
+    if (remoteConnectionMapSize != 0)
+      com::CommunicateBoundingBox(_m2ns[0]->getMasterCommunication()).receiveConnectionMap(remoteConnectionMap, 0);
```
fsimonis: Please use braces to stay consistent with the rest of the code

src/partition/ProvidedPartition.cpp
```diff
@@ -22,122 +24,274 @@ void ProvidedPartition::communicate()
...
+        }
+      }
+    }
+
+  } else { // Slave
+
+    utils::MasterSlave::_communication->broadcast(connectedRanksList, 0);
+
+    if (!connectedRanksList.empty()) {
+      for (auto &rank : connectedRanksList) {
```
fsimonis:
```suggestion
      for (int rank : connectedRanksList) {
```

src/partition/ProvidedPartition.cpp
```diff
@@ -22,122 +24,274 @@ void ProvidedPartition::communicate()
...
+        remoteConnectionMap[rank] = {-1};
+      }
+      com::CommunicateBoundingBox(utils::MasterSlave::_communication).broadcastReceiveConnectionMap(remoteConnectionMap);
+    }
+
+    for (auto &remoteRank : remoteConnectionMap) {
```
fsimonis:
```suggestion
    for (const auto &remoteRank : remoteConnectionMap) {
```

src/partition/ProvidedPartition.cpp
```diff
@@ -22,122 +24,274 @@ void ProvidedPartition::communicate()
...
+      for (auto &includedRanks : remoteRank.second) {
```
fsimonis:
```suggestion
      for (int includedRank : remoteRank.second) {
```

src/partition/ReceivedPartition.cpp
```diff
@@ -33,30 +36,61 @@ ReceivedPartition::ReceivedPartition(
...
-  PRECICE_INFO("Receive global mesh " << _mesh->getName());
-  Event e("partition.receiveGlobalMesh." + _mesh->getName(), precice::syncMode);
-  if (not utils::MasterSlave::isSlave()) {
-    PRECICE_ASSERT(_mesh->vertices().empty());
-    // a ReceivedPartition can only have one communication, @todo nicer design
-    com::CommunicateMesh(_m2ns[0]->getMasterCommunication()).receiveMesh(*_mesh, 0);
+  PRECICE_ASSERT(_m2ns.size() == 1);
+  PRECICE_ASSERT(_mesh->vertices().empty());
+
+  if (_m2ns[0]->usesTwoLevelInitialization()) {
```
fsimonis: Maybe safe a reference to `_m2ns[0]` with an expressive name.

src/partition/ReceivedPartition.cpp
```diff
@@ -33,30 +36,61 @@ ReceivedPartition::ReceivedPartition(
...
+    PRECICE_INFO("Receive global mesh " << _mesh->getName());
+    Event e("partition.receiveGlobalMesh." + _mesh->getName(), precice::syncMode);
+
+    if (not utils::MasterSlave::isSlave()) {
+      // a ReceivedPartition can only have one communication, @todo nicer design
+      com::CommunicateMesh(_m2ns[0]->getMasterCommunication()).receiveMesh(*_mesh, 0);
+      _mesh->setGlobalNumberOfVertices(_mesh->vertices().size());
+    }
+  }
+  if (utils::MasterSlave::isMaster()) {
```
fsimonis: If this is a separate logical block, it may be good to explain its purpose in a comment.
Otherwise we could merge it with the above.

src/partition/ReceivedPartition.cpp
```diff
@@ -116,55 +287,53 @@ void ReceivedPartition::compute()
...
-      com::CommunicateMesh(utils::MasterSlave::_communication).broadcastSendMesh(*_mesh);
+    if (not _m2ns[0]->usesTwoLevelInitialization()) {
+      PRECICE_INFO("Broadcast mesh " << _mesh->getName());
+      Event e("partition.broadcastMesh." + _mesh->getName(), precice::syncMode);
+
+      if (utils::MasterSlave::isSlave()) {
+        com::CommunicateMesh(utils::MasterSlave::_communication).broadcastReceiveMesh(*_mesh);
+      } else { // Master
+        PRECICE_ASSERT(utils::MasterSlave::getRank() == 0);
+        PRECICE_ASSERT(utils::MasterSlave::getSize() > 1);
```
fsimonis: Shouldn't it suffoce to assert `isMaster()`?

src/partition/ReceivedPartition.cpp
```diff
@@ -175,95 +344,125 @@ void ReceivedPartition::compute()
...
+{
+  PRECICE_TRACE();
 
-  // (3) Define which vertices are owned by this rank
-  PRECICE_DEBUG("Create owner information.");
-  createOwnerInformation();
+  // @todo handle coupling mode (i.e. serial participant)
+  // @todo treatment of multiple m2ns
+  PRECICE_ASSERT(_m2ns.size() == 1);
+  if (not _m2ns[0]->usesTwoLevelInitialization())
```
fsimonis: Maybe use a named reference to `*_m2ns[0]`

src/partition/ReceivedPartition.cpp
```diff
@@ -175,95 +344,125 @@ void ReceivedPartition::compute()
...
+    _m2ns[0]->getMasterCommunication()->send(connectedRanksList, 0);
+    if (not connectionMap.empty()) {
+      com::CommunicateBoundingBox(_m2ns[0]->getMasterCommunication()).sendConnectionMap(connectionMap, 0);
+    } else {
+      PRECICE_ERROR("This participant seems to have no mesh partitions for mesh " << _mesh->getName() << " at the coupling interface.");
+    }
+  } else {
+    PRECICE_ASSERT(utils::MasterSlave::isSlave());
+
+    for (auto &remoteBB : remoteBBMap) {
```
fsimonis: As you are not modifying its content:
```suggestion
    for (const auto &remoteBB : remoteBBMap) {
```

src/partition/ProvidedPartition.cpp
```diff
@@ -22,122 +24,274 @@ void ProvidedPartition::communicate()
...
+      // communicate the total number of vertices to the other participants master
+      if (utils::MasterSlave::isMaster()) {
+        _m2ns[0]->getMasterCommunication()->send(_mesh->getGlobalNumberOfVertices(), 0);
       }
-    }
 
-    e1.stop();
+      // the min and max of global vertex IDs of this rank's partition
+      int minGlobalVertexID = _mesh->getVertexOffsets()[utils::MasterSlave::getRank()] - _mesh->vertices().size();
+      int maxGlobalVertexID = _mesh->getVertexOffsets()[utils::MasterSlave::getRank()] - 1;
```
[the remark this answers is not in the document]
uekerman: Mmh, not sure, it's only needed here and only two lines long.

PR #535, opened 2019-10-17
Add Mapping Distance Info

### LINKED ISSUE #160: Warning message for non-matching interface geometries
As pointed out by @msuaeronautics13, it would be helpful to get a warning if two interface meshes describe obviously different geometries, such as a wrong z-plane, etc. 

Currently, with the right decomposition, we have the check:
https://github.com/precice/precice/blob/017e2f3c93bd62280f8031bd9c2ca4c147e57861/src/partition/ReceivedPartition.cpp#L90-L95

But this only fails with the right decomposition. Would be good to have a more general solution.

fsimonis:
This PR adds an info log which displays the minimum and maximum distance between projections for NN/NP mappings.
This is easily extensible to any other features of [boost accumulator](https://www.boost.org/doc/libs/1_65_1/doc/html/accumulators/user_s_guide.html#accumulators.user_s_guide.the_statistical_accumulators_library).
I implemented a wrapper class `precice::utils::statistics::DistanceAccumulator` which makes it simpler to print and extend.

This is a non-intrusive way of giving the user a hint for mismatching meshes without making assumptions of the units of the coordinate systems.

**Test output:**
```
(0) 11:59:01 [mapping::NearestProjectionMapping]:122 in computeMapping: Mapping distance min:0 max:0.707107
```

Closes #160

### DIFF
--- a/src/mapping/NearestNeighborMapping.cpp
+++ b/src/mapping/NearestNeighborMapping.cpp
@@ -1,13 +1,12 @@
 #include "NearestNeighborMapping.hpp"
 #include "query/FindClosestVertex.hpp"
 #include "utils/Helpers.hpp"
+#include "utils/Statistics.hpp"
 #include "mesh/RTree.hpp"
 #include <Eigen/Core>
 #include <boost/function_output_iterator.hpp>
 #include <boost/container/flat_set.hpp>
 #include "utils/Event.hpp"
-
-
 
 namespace precice {
 extern bool syncMode;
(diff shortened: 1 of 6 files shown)


fsimonis:
@uekerman I added the mean, sample variance and the sample count to the output.
This should be enough information for the user.

Emitting a warning mean that we are making assumptions on the unit, something which I highly discourage.
We would make sense to add a max distance/variance as a configuration tag though.

PR #536, opened 2019-10-21
Add configuration context to callback

fsimonis:
The primary goal of this PR is to add the name, rank and com size as a context to the configuration.

The implementation passes an immutable context to the tag callbacks, which is passed to the entire tag-tree. The function `xml::configure()` now takes the context as an additional parameter.

As a direct consequence, the majority of our integration tests had to change.
For this reason I introduces a new test fixture `testing::SlimConfigurator` which builds a context from a solverinterface, generates the configuration and finally configures the solverinterface.
I also had to extend the interface of SolverInterfaceImpl to access the parameters passed to the constructor.
This reduces the verbosity of our test setup a lot!

I also added the test `XML/Context` which detected an old bug in the `ConfigParser` which lead to `Listener::xmlEndTagCallback()` never being called _if a tag is empty_.

_edit: clarified the bug_

### DIFF
--- a/src/testing/Testing.hpp
+++ b/src/testing/Testing.hpp
@@ -220,7 +220,7 @@ struct SlimConfigurator : WhiteboxAccessor {
      * @param[in] configFilename The filename of the configuration file.
      */
     template<typename T>
-    static void slimConfig(T& interface, const std::string& configFilename) {
+    static void slimConfigure(T& interface, const std::string& configFilename) {
         auto & interfaceImpl = impl(interface);
         precice::xml::ConfigurationContext context{
             interfaceImpl.getAccessorName(),
--- a/src/testing/Testing.hpp
+++ b/src/testing/Testing.hpp
@@ -3,6 +3,8 @@
 #include "math/math.hpp"
 #include "utils/Parallel.hpp"
 #include <boost/test/unit_test.hpp>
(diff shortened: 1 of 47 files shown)


src/testing/Testing.hpp
```diff
@@ -206,6 +208,30 @@ struct WhiteboxAccessor {
     }
 };
 
+/// struct extending WhiteboxAccessor by a slim configuration call.
+struct SlimConfigurator : WhiteboxAccessor {
+    /** Configures the interface given a fileName in a slim and quiet manner.
+     *
+     * This first genenrates a ConfigurationContext from the SolverInterface.
```
uekerman:
```suggestion
     * This first generates a ConfigurationContext from the SolverInterface.
```

src/testing/Testing.hpp
```diff
@@ -206,6 +208,30 @@ struct WhiteboxAccessor {
...
+     * It then configures a config::Configuration using the context and the given file.
+     * Finally, it calls SolverInterface::configure with the configuration.
+     * 
+     * @param[in] interface The SolverInterface to configure.
+     * @param[in] configFilename The filename of the configuration file.
+     */
+    template<typename T>
+    static void slimConfig(T& interface, const std::string& configFilename) {
```
uekerman: Why not `slimConfigure`?

Issue #538, opened 2019-10-24
Solver dummies do not include read and write functions


KyleDavisSA:
Currently, the solver dummies do not include any kind of read or write functionality. This is understandable as it is desired to be able to run a python solver dummy with a C solver dummy. But this can also make it difficult to troubleshoot problems when writing a new adapter for a new solver. It would be helpful if all solver dummies had a set input that can read and write data. An example could be:

- Each solver dummy would come with 2 solvers, A and B, and the config file. The solver dummy A can work with any other language solver B.
- Each solver has 3 vertices
- Each reads in data from the other solver and adds 1
- This can be read by the other solver and checked to see if the data transfer was correct (for simple nearest neighbor mapping with exactly the same vertices)

This way, the solver dummy can always be reviewed when writing a new adapter.

uekerman:
> This is understandable as it is desired to be able to run a python solver dummy with a C solver dummy

I guess there is a misunderstanding here. You can still couple different solver dummies when data is exchanged. Is that what you meant?

Still, I agree that extending the dummies here could be helpful. 

> Each solver dummy would come with 2 solvers, A and B, and the config file. 

Just to double-check, both "solvers" would come in the same file, right? So, A or B is passed via a command line argument as it is the case already now.

Rest sounds good.

BenjaminRueth:
> > Each solver dummy would come with 2 solvers, A and B, and the config file.
> 
> Just to double-check, both "solvers" would come in the same file, right? So, A or B is passed via a command line argument as it is the case already now.

See https://github.com/precice/precice/pull/540#issuecomment-546355879

PR #540, opened 2019-10-25
Solver dummy extension for data transfer

KyleDavisSA:
This PR aims to address /issues/538 . The solver names and mesh names are hard coded into the solvers at the moment. The solver dummies should be able to run together i.e. `python solverdummy_A` should run with `fortran solverdummy_B`, but not a `Solver_A`  with another `Solver_A`. 

Each solver has 3 vertices, with the x,y,z coordinate being `1` for vertex `1`, `2` for vertex `2`, and `3` for vertex `3`. There is a pause before advance that waits for a keystroke before continuing to troubleshoot any data transfer issues. `Solver_A` starts with zeros and adds a value of `1` to each vertex. This is sent to `Solver_B`, which should receive the values of `1` and add `1` so that the value is `2`. This is sent back to `Solver_A` which should receive a value of `2` and so forth.


uekerman:
Just discussed with @KyleDavisSA: Let's first find the actual problem and then put the PR in a better shape (see coments in #538). 
@BenjaminRueth Could you please check the Python API?

BenjaminRueth:
I am not so happy with the code duplication introduced by having `solverdummy_A*` and `solverdummy_B*`. I assume that you introduced these two versions for the sake of not always having to enter all the required parameters (`python3 solverdummy.py SolverOne MeshOne ...` is quite lengthy...). 

Maybe you can introduce a parameter that can be set for choosing between default configuration `A` and `B`. This would avoid the code duplication and shorten the required user input to a minimum `python3 solverdummy.py -A` sets `meshName =  "meshOne"` and `solverName= "solverOne"`.

Please make sure that the new implementation is backwards compatible, since otherwise we will also have to update our tests. Of course, we can also update the tests, if otherwise the technical overhead becomes too big, but at least in Python one can quite easily implement these two alternative ways of providing input. I have no idea about Fortran, though.

uekerman:
I would vote for not changing the input.
```
python3 solverdummy.py SolverOne MeshOne
```
is good enough in my opinion.
And let's not call it "A" if it's already called "SolverOne"

New `precice-config.xml` should neither be needed. 

But like said above: let's first find the actual problem.

KyleDavisSA:
The new solverdummy has the same input format of:
 
`python3 solverdummy.py ../precice-config.xml SolverOne MeshOne` for python or 
`./solverdummy ../precice-config.xml SolverTwo MeshTwo` for fortran

The adapter are also inter-usable, i.e. a python solverOne works with a fortran solverTwo.

Issue #545, opened 2019-11-05
Python binding (precice_future): More descriptive assertions


ajaust:
Hi,

I think it would make sense to have more descriptive assertions, especially with the changes in the new Python bindings (`python_future`). A lot of things changed in the API making it likely that things break in user code. I was a bit surprised that the new bindings not only change the number of arguments a preCICE function takes (`set_mesh_vertices` takes 3 arguments instead of 4 in the C++ interface), but also it expects different argument types/shapes. 

In the old bindings `set_mesh_vertices` would accept a list of coordinates where it was basically a 1D list/array of length `numPoints*dimension`. The data passed to tthe function did **not** have to be an numpy array. In `precice_future` it worked for me only when I use 2D numpy.array with `numPoints` entries that have the coordinates (`dimension`).

I started with using an older code with the new interface that used a 1D list (no numpy) of coordinates. This was caught by the exception [in line 135](https://github.com/precice/precice/blob/v1.6.1/src/precice/bindings/python_future/precice_future.pyx#L135), but the exception was not really helpful as there was no explanation what I did wrong. From the documentation it is also not obvious what the function type/dimension the input arguments should have. 

I would suggest to replace the line
```python
assert(dimensions == self.get_dimensions())
```
by something like
```python
assert(dimensions == self.get_dimensions(),"Descriptive message what failed and why (what did the binding expect here?)")
```

This holds true for the other assertions as well. At least for me as a user it was not immediately obvious why it failed and what the binding actually expects. This is also somehow related to the documentation of the bindings as discussed in #450 and #437.

Thank you!

IshaanDesai:
Thank you @ajaust for raising this important issue! I have added error message to the assertions in the new python_bindings here: https://github.com/precice/precice/pull/547

ajaust:
Great! Thank you. Sorry for opening another issue. I guess I only searched for "Python" and not for "Pythonbindings" in the issues before opening a new one.

BenjaminRueth:
> Great! Thank you. Sorry for opening another issue. I guess I only searched for "Python" and not for "Pythonbindings" in the issues before opening a new one.

No need for apologies. You issue actually triggered this PR.

ajaust:
Ah. I missed that. Then I am happy with opening the issue. ;)

Thanks for the quick fix.

IshaanDesai:
[...]

PR #547, opened 2019-11-05
Pythonbindings error messages

IshaanDesai:
Addressing this issue: https://github.com/precice/precice/issues/545 
I am not sure if writing long strings in `assert` statements in python is the best way to provide the user with error information. 
Open question: Is it possible to list a general set of exceptions to which integer values can be passed and then the expressions are raised when errors occur ?


src/precice/bindings/python_future/precice_future.pyx
```diff
@@ -118,7 +118,7 @@ cdef class Interface:
         if not isinstance(position, np.ndarray):
             position = np.asarray(position)
         dimensions = position.size
-        assert(dimensions == self.get_dimensions())
+        assert(dimensions == self.get_dimensions(), "Function set_mesh_vertex expects a 1d numpy array with coordinates of point as position input")
```
BenjaminRueth: I think it would generally be helpful to provide error messages where you give the user information about what was received and what the interface accepts.

Example:

```
assert(dimensions == self.get_dimensions(), "Function set_mesh_vertex(mesh_id, position) received position of type {position_type}, but expects a 1d numpy array with coordinates of point as position input".format(type(position)))
```

For other API calls, we might need to provide additional information, like expected shape etc.

BenjaminRueth: It might even be a good idea to refactor these checks out into some new functions to avoid code duplication.

IshaanDesai: Is there a chance that the `type()` call to a wrongly defined variable or array lead to further more confusing error output ? That was my thinking towards restraining to pass additional arguments to the error message.

BenjaminRueth: It can always become worse, I assume (especially, in python, where you can break everything, if you really want it). But for the normal use case that a user provides a list or another iterable object, while we expect a numpy array, it should work.

ajaust: I like the new assertions much more than the old ones. :)

From my perspective it would be important 

- to know that something failed
- to know why it failed or at least to get a hint that helps me figuring this out. 
- to maybe know where it failed (if I need to know that/helps me solve the problem.) 

The new assertions seem to address these points as far I as I checked.

src/precice/bindings/python_future/error_handling.py
```diff
@@ -0,0 +1,16 @@
+"""
+Functions to check the type of input argument against what is expected
+and raise an Exception if mismatch found
+"""
+
+
+def check1D(func_name, usr_input, true_val):
+    if usr_input != true_val:
+        raise Exception("Function {} received position of type {} but expects a 1D numpy \
+                         array with {} values as position input".format(func_name, usr_input, true_val))
```
BenjaminRueth: I like the new functions for checks. However, I think currently this function is trying to address two issues and this makes it a bit confusing.

What do we want to check?

1. Is the input a numpy array or at least something that we can transform into a numpy array? I.e. can we even access `input.shape`?
2. Does the dimensions match? Does the number of vertices match? Be careful, these are two different things and we should also give the correct error depending on the context.

Currently `check1D` and `check2D` are always used for checking the second requirement. The first requirement is not checked, even though, the error message of `check1D` suggests that it is checking for the type (I assume this is a copy-paste-error).

BenjaminRueth: Additionally, if you split these two concerns, you can raise a `TypeError`, if the type does not match and a different kind of error if the dimension or size is wrong. `ValueError` might be appropriate, but I am not 100% whether this is matching. At least Scipy raises a `ValueError` for dimension mismatch.

IshaanDesai: I agree that we need to separately check if the input array is a numpy array and hence can we even access functionality like `shape`. But then one question is raised, if the user has not provided a numpy array do we convert it and notify the user of the conversion or do we just throw an error and tell the user to convert it ?

BenjaminRueth: I would be strict here: If we do not receive something that we can work with, throw an error.

* We really need `shape`, in the case of 2D data (vector data). Everything else than a `np.array` is invalid input -> throw an error.
* We can also use `len()` instead, for 1D data (scalar data, lists of IDs...). Here, any object that supports `len()` is fine -> accept it, if possible.

IshaanDesai:
This PR is now addressed via [this branch](https://github.com/precice/python-bindings/tree/improvedErrorMessages) in the new bindings repo: https://github.com/precice/python-bindings/tree/develop

PR #548, opened 2019-11-05
delete copy and assign operator (C++11 feature)

ajaust:
Hi,

Instead of making the copy and assignment operator private it would be more easy to read/understand and less error-prone to [delete these operators](https://docs.microsoft.com/en-us/cpp/cpp/explicitly-defaulted-and-deleted-functions?view=vs-2019). This is a C++11 feature thus it is easy to integrate as preCICE requires at least C++11.

It can avoid also compile/link errors if somebody still tries to use the operations (see [Stackoverflow](https://stackoverflow.com/questions/7823845/disable-compiler-generated-copy-assignment-operator)).

Best,
Alex

### DIFF
--- a/src/precice/SolverInterface.hpp
+++ b/src/precice/SolverInterface.hpp
@@ -803,10 +803,10 @@ public:
 
   ///@}
 
-  /// Disable copy construction by making copy constructor private.
+  /// Disable copy construction 
   SolverInterface ( const SolverInterface& copy ) = delete;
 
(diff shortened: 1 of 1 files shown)


fsimonis:
@ajaust Thank you for the PR! I did a few passes over the codebase already. Funny that I missed the most visible header...

ajaust:
No problem. I assumed this slipped through somehow.

PR #551, opened 2019-11-06
Migrate PETSc discovery to pkg-config

fsimonis:
This PR migrates the cmake PETSc discovery to `pkg-config`.

The result is
* we use the well know pkg-config which is provided with petsc source installations
* for source installations, once has to set either `CMAKE_PREFIX_PATH` or `PKG_CONFIG_PATH` to `$PETSC_DIR/$PETSC_ARCH/`
* we get rid of 500 lines of code we don't understand.

**TODO**
* [x] wrap in a `FindPETSc` module to allow override via `PETSc_DIR` or a `PETScConfig.cmake`.
* [x] make xSDK overrides work with said module.


fsimonis:
This PR needs more work as the pkg_config call should be wrapped inside a `FindPETSc` and define a target with the canonical name `PETSc::PETSc`. This way it can coexist with provided `PETScConfig` files.

PR #552, opened 2019-11-08
Add C++11 Library Conformance Check

fsimonis:
This PR adds a check to the early CMake configuration phase, which enforces a c++11 conforming library in terms of time I/O manipulators.

Now, the pre-gcc-5 compilers, which have C++11 compiler support, but only partial library support will fail straight away.


cmake/ValidateCpp11Library.cmake
```diff
@@ -0,0 +1,24 @@
...
+  #include <iomanip>
+  #include <ctime>
+  int main() { std::time_t t = std::time(nullptr); std::tm tm = *std::localtime(&t); std::cout << std::put_time(&tm, \"%c %Z\"); }"
+  )
+
+  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
+  check_cxx_source_compiles("${CPP11CheckSource}" CPP11LibraryConforming)
+  if(NOT "${CPP11LibraryConforming}")
+    message(FATAL_ERROR "The standard library used by your compiler is not C++11 compliant, as it does not support time manipulation!")
```
MakisH: Maybe suggest a next step here?

MakisH: Would it be a good practice to throw a hint about installing preCICE with Spack here? Or refer to our documentation with a link?

fsimonis:
@MakisH I extended the error message. Mentioning spack is probably a bit off-topic, so I just tell the user what compiler versions are supported.

**Example**
```
-- Performing Test CXX11_N2071_N2072
-- Performing Test CXX11_N2071_N2072 - Failed
CMake Error at cmake/CheckCXX11Library.cmake:16 (message):
  

      The standard library used by your compiler Clang 3.1.0 is not C++11 compliant.
      It does not support time manipulation (N2071 and N2072)!
      Please upgrade your compiler at least to:
      * GCC 5
      * LLVM 3.8
      * MSVC 19.0
      * Intel 15 in conjunction with GCC 5
      * PGI 2015 in conjunction with GCC 5
      
Call Stack (most recent call first):
  cmake/CheckCXX11Library.cmake:35 (_check_cxx_11_n2071_n2072)
  cmake/CheckCXX11Library.cmake:38 (check_cxx_11_library_support)
  CMakeLists.txt:29 (include)


-- Configuring incomplete, errors occurred!
See also "/home/simonis/dev/precice/main/build/CMakeFiles/CMakeOutput.log".
See also "/home/simonis/dev/precice/main/build/CMakeFiles/CMakeError.log".
```

PR #554, opened 2019-11-11
Fix assertion in getMeshVertexIDsFromPositions

BenjaminRueth:
`size` of the vertices where we request `vertexIDs` has to be `<=` the `size` of `mesh.vertices`. Not `>=`!

### DIFF
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -742,7 +742,7 @@ void SolverInterfaceImpl:: getMeshVertexIDsFromPositions (
     mesh::PtrMesh mesh(context.mesh);
     PRECICE_DEBUG("Get IDs");
     const auto &vertices = mesh->vertices();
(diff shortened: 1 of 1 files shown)


fsimonis:
Actually, this assertion is unnecessary and should be removed.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -742,7 +742,7 @@ void SolverInterfaceImpl:: getMeshVertexIDsFromPositions (
     mesh::PtrMesh mesh(context.mesh);
     PRECICE_DEBUG("Get IDs");
     const auto &vertices = mesh->vertices();
-    PRECICE_ASSERT(vertices.size() <= size, vertices.size(), size);
+    PRECICE_ASSERT(vertices.size() >= size, vertices.size(), size);
```
fsimonis:
```suggestion
```

fsimonis:
```suggestion
```

PR #558, opened 2019-11-11
Set post-processing occurrence to OCCUR_ONCE

### LINKED ISSUE #524: Add error message for multiple post-processing tags
A `coupling-scheme` can have 0 or 1 `post-processing`. Currently, there is no error message if >=2 `post-processing` tags are used in the config. Probably, the first one is used. We should add a proper check.

huangq1234553:
Updated occurrence flag for 'post-processing' XML Tag in 
precice/src/acceleration/config/AccelerationConfiguration.cpp to OCCUR_ONCE

Fixes #524


fsimonis:
It looks like the code ifself was right as the acceleration tag can occur once or not at all. 
Something must be wrong in the xml configuration parser.

PR #564, opened 2019-11-16
Speedup mesh operations

fsimonis:
This PR speeds-up the operations on meshes by migrating the maps to flat_maps, preventing them to resize and by moving some code around.
I also took the liberty to refactor the `filteredMesh` out of the partition classes and put it into `mesh/Filter.hpp`. This is now templated on the UnaryPredicate used to filter vertices.

This reduces the runtime of these functions for large Meshes ~500000 vertices by 1/3.

Current downside is that we loose the debug output at the beginning and the end of the original `filteredMesh`.

Reviewers: Do we have to keep the debug output?
_update: The output is usefull, so I added it at the call-side_

### DIFF
--- a/src/com/CommunicateMesh.cpp
+++ b/src/com/CommunicateMesh.cpp
@@ -7,6 +7,8 @@
 #include "mesh/Mesh.hpp"
 #include "mesh/Triangle.hpp"
 #include "mesh/Vertex.hpp"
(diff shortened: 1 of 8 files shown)


fsimonis:
I would like to wait for #585 to be merged so that I can integrate it into this PR.

Issue #374, opened 2019-05-04
Splitting of interfaces into patches


KyleDavisSA:
Hi all

we want to add some functionality in preCICE to improve the mapping using RBF. Taking the cylinderFlap tutorial example, there is an interface giving by the name “flap”. The points on this surface is set using 

setMeshVertices (int meshID, int size, double* positions, int* ids) 

However, if a gaussian RBF is used as an example, it selects all points within a specified radius around each point. This could then select points on an opposing surface which is not physically correct. In the picture below, all points on the surface are part of the same interface.

![Screenshot from 2019-05-04 15-43-29](https://user-images.githubusercontent.com/22762795/57180726-cf5d0480-6e8b-11e9-8475-4cc1ced59b22.png)

An idea is to split a SINGLE interface into separate patches. The example above is then split into 3 separate patches, all belonging to the same interface. This could potentially be set with something like:

setMeshVertices (int meshID, int size, double* positions, int* ids, int patchSize, int* PatchIndex)  

In the example below, the single interface is split into 3 patches. In order to perform an RBF mapping, a point on patch 1 can only search for points on patch 1, same for patches 2 and 3. However all patches are part of the same interface used in the acceleration scheme. Therefore the acceleration scheme does not distinguish between patches and only looks at the interface meshID. 

![Screenshot from 2019-05-04 13-59-34](https://user-images.githubusercontent.com/22762795/57180736-f74c6800-6e8b-11e9-97ef-9b7bee4a81c8.png)

The potential benefit is that the meshID, size, positions and ids stay exactly the same. The addition is patchSize is the total number of patches. PatchIndex is an array saying how far down the list of the array positions the patch starts. Therefore the positions array must be input in order of the patches. This is done in the adapter and preCICE does not need to do anything in this regard, and this is possible within the solver adapters.

Lets say patch 1 has 20 points, patch 2 has 30 points and patch 3 has 50 points. The original function works like.

SetMeshVertices (1; 100; 1 2 3,…,298 299 100; 1,…,100) 

the total number of points are 100, the positions are made of of the vertex locations in 3 dimensions and the ids gives an index to the particular vertex. The new function could have the additional number of patches (in this case, 3) and the position down the list where the patch starts.

SetMeshVertices (1; 100; 1 2 3,…,298 299 100; 1,…,100; 3; 0,20,50)

[...]

PR #566, opened 2019-11-20
Refactor meshids

### LINKED ISSUE #559: Remove static state in Mesh
To fully support multiple instances of the SolverInterface in one executable, we need to get rid of some `static` state.

All [`mesh::PropertyContainers`](https://xgm.de/precice/docs/develop/classprecice_1_1mesh_1_1PropertyContainer.html) (`Mesh`, `Vertex`, `Edge`, `Triangle`) contain a static `utils::ManageUniqueIDs`.
This should be owned by the `participant` itself and passed into its `Mesh`es.

Related to #378 and #385

fsimonis:
This PR refactors the static IdManager for Meshes, as well as Mesh-internal Meshes, out of the Mesh.
It is now injected as a dependency via the constructor.
This makes testing easier, as we can use separate IdManagers instead to make assumptions on IDs.
For everything else, we can use a common manager (`testing::meshIDManager()`)

The owner of this IDManager is first the MeshConfiguration and then the actual Participant.
During the configuration creates the Meshes and thus requires this Manager. This is moved to the actual Participant after it was determined.

**However** this manager is only required to set [subIDs](https://xgm.de/precice/docs/develop/classprecice_1_1mesh_1_1Mesh.html#a2e65d1316c0d507ff751bbc3de7a2d05), a function that could theoretically be removed. @uekerman can you comment on that please?

This also reduces the global state of of the program.

Fixes #559

### DIFF
--- a/src/acceleration/test/AccelerationMasterSlaveTest.cpp
+++ b/src/acceleration/test/AccelerationMasterSlaveTest.cpp
@@ -53,7 +53,7 @@ BOOST_AUTO_TEST_CASE(testVIQNILSpp, * testing::OnSize(4) * boost::unit_test::fix
   PtrPreconditioner prec(new ConstantPreconditioner(factors));
   std::vector<int> vertexOffsets {4, 8, 8 , 10};
 
-  mesh::PtrMesh dummyMesh(new mesh::Mesh("DummyMesh", 3, false, testing::meshIDManager()));
+  mesh::PtrMesh dummyMesh(new mesh::Mesh("DummyMesh", 3, false, testing::nextMeshID()));
   dummyMesh->setVertexOffsets(vertexOffsets);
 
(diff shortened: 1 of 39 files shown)


uekerman:
> However this manager is only required to set subIDs, a function that could theoretically be removed.

I did not have a careful look yet, but I think indeed that this feature is currently obsolete. But it could get interesting again: https://github.com/precice/precice/issues/374

fsimonis:
@uekerman There may be a simpler solution to this. I'll comment in the other issue.

This PR will become partially obsolete if the decide to remove this feature, so I'll put it on hold.

fsimonis:
After we merge #585, there will be no submeshes.
Thus there will be no need to pass the manager to the Mesh. Passing a simple id will be sufficient.
Consequently, this will simplify creating meshes and thus simplify this PR.

fsimonis:
The meshes now receive an ID on construction.

PR #568, opened 2019-11-21
Add cleanup step prior to running tests

fsimonis:
This PR adds a cleanup/setup step, which wipes the working directories of tests prior to running them.

This ensures the tests don't clog up due to a stale `precice-run` directory and removes irrelevant files from previous runs.


fsimonis:
This method is too verbose. This will explode in verbosity once #594 is merged.

PR #571, opened 2019-11-22
Add Markdown Reference Generator

fsimonis:
This PR adds a Markdown reference generator. Solves #557 

`./binprecice md > reference.md`

The general format is:

* Heading with tag name
* Tag description
* Table of attributes
* List of the names of subtags with their allowed occurrence setting

Then the same for all subtags with a deeper heading level.
See [an example of the result here](https://github.com/precice/precice/issues/557#issuecomment-557522229).


fsimonis:
I would prefer to completely refactor the printing out of the `XMLTag` and `XMLAttribute` classes.
Their header are includes in many places and should be as lightweight as possible.
Furthermore, formatting is not really a concern of these classes themselves.

PR #572, opened 2019-11-22
Simplify parallel configuration

### LINKED ISSUE #347: Unify preCICE configuration for parallel and serial solvers
When changing a solver from serial to parallel execution, the preCICE configuration has to be adapted at several locations. Some of them are purely technical and hard to understand for our users. We need to improve this situation.

* `master:mpi-single` should be the default if a solver is run in parallel.
* `gather-scatter` should be picked automatically if necessary
* RBF mappings should always be called `rbf` and there should be a clever default depending on whether preCICE was built with PETSc and whether a solver is run in parallel or not. 

*I will add more information to this issue.*

uekerman:
fixes #347 

* [x] make `master:mpi-single` default
* [x] pick `gather-scatter` automatically
* [x] pick correct RBF variant
* ~[ ] Remove coupling mode~ #598

### Direct MPI as default master com

**Summary of behavior**

* If `server` is specified: use server as usual 
* If no tag `master`  is specified and `size` > 1: a `DirectMPICommunication` (`mpi-single`) is created
* If no tag `master`  is specified and `size` == 1: normal serial run, nothing to do
* If `master` tag specified and  `size` > 1: create specified communication
*  If `master` tag specified and  `size` == 1: throw error message
* If no tag `master`  is specified and `size` > 1, but mpi=off: explicit definition of master com is still required (otherwise error)

**Implementation**

* Straight forward with the configuration context implemented in #536 

### gather-scatter

**Summary of behavior**

* Instead of `distribution-type`, there is now a boolean attribute `enforce-gather-scatter`, which defaults to false.
* Meaning, as default, always a `PointToPointCommunication` is created. 
* `GatherScatterCommunication` should now only be relevant for debugging (I could still imagine some very rare cases with performance benefits)
* Only exception is `m2n:mpi-single`, where by default a `GatherScatterCommunication` is created as the only supported option. Standard users should not need this option, but we still rely heavily on it in the integration tests.

**Implementation**
 
* Instead of directly using the master-master communication, a  `DistributedCommunication` is now created even for serial participants. Functionality of `GatherScatterCom` and `PointToPointCom` was extended accordingly.
* This makes live a bit more complicated for unit tests in `cplscheme` as they need inter-participant communication. I decided to not construct the complete overhead of mesh and vertex distribution definition for every test, but offered a loophole through a friend declaration in `M2N` and a fallback to the pure master-master communication. Suggestions welcome!

### RBF

**Summary of behavior**

* There is now only `rbf`, no `petrbf` anymore.
* Instead, there is an attribute `use-lu-decomposition`, which defaults to false.
* If possible and not requested differently a `PetRBFMapping` is created.
* If `use-lu-decomposition="true"` but run in parallel a warning is thrown and the mapping is changed to `PetRBFMapping`.
* If parallel and built with `PETSC=OFF` an error is thrown.

### Further changes

* Removed deprecated (and unused) `ParticipantConfiguration::addParticipant()`
* With the `DistributionCommunication` refactoring, the `server` mode is no longer functional for parallel participants. No problem as we either way want to get rid of it #574. 

### Many things will get simpler when ...

* server is removed #574 
* coupling mode is removed #598
* configuration is refactored such that only accessing participant is configured

### Difficulties in tests

see #597 

* serial integration tests cannot use PETSc-based RBF mappings -> `use-lu-decomposition` always set to true. 
* needed to switch off the collect step in the `EventRegistry` for tests

### Reviewers

* @fsimonis Please have a brief look over the code
* @MakisH Please check the user impact and test some OpenFOAM case (I have already tested many combinations of configs with Alya, so no need to test all tutorials right now). 

### Clean-up TODOs (once PR is merged)

* [ ] all tutorials should only offer one config file
  * [ ] remove `master` tag
  * [ ]  remove `distribution-type` attribute
* [ ] update documentation
  * [ ] use `master` tag only for cases where MPI support is limited.
  * [ ] remove / update `distribution-type` documentation


fsimonis:
The test with MPI and PETSc enabled seems to fail due to the same reasons as #591.
The failing MPI test with PETSc disabled runs on my machine, so it could really be a timeout issue.

src/utils/EventUtils.cpp
```diff
@@ -230,10 +234,11 @@ void EventRegistry::finalize()
   for (auto & e : storedEvents)
     e.second.stop();
 
-  if (initialized) // this makes only sense when it was properly initialized
+  if (initialized && not precice::testMode) // this makes only sense when it was properly initialized
```
fsimonis: It seems a bit odd that this is necessary.

I would rather refactor the events library and the testing framework instead of integrating hacks like these.

src/precice/config/ParticipantConfiguration.hpp
```diff
@@ -119,7 +112,12 @@ class ParticipantConfiguration : public xml::XMLTag::Listener
 
   mapping::PtrMapping getMapping ( const std::string& mappingName );
 
-  void finishParticipantConfiguration ( const impl::PtrParticipant& participant );
+  /// does the config already provide a solution for a parallel participant (master or server)
+  bool _isParallelSolutionDefined = false;
```
fsimonis: I don't understand what this is supposed to mean.

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -616,6 +608,23 @@ void ParticipantConfiguration:: finishParticipantConfiguration
...
   _watchPointConfigs.clear ();
+
+
+  // create default master communication if needed
+  if(context.size > 1 && not _isParallelSolutionDefined && participant->getName() == context.name){
+    #ifdef PRECICE_NO_MPI
+        std::ostringstream error;
+        error << "When building with \"mpi=off\", you need to specify an alternative \"master\" "
+              << "communication (e.g. sockets) for every parallel participant";
+        throw std::runtime_error{error.str()};
```
fsimonis: You can collapse those lines into a single throw:
```cpp
throw std::runtime_error{ "When building with \"mpi=off\", you need to specify an alternative \"master\" "
              "communication (e.g. sockets) for every parallel participant"};
```

Note that following `"c" " string " "literals"` will be joined by the compiler.

src/m2n/M2N.hpp
```diff
@@ -182,6 +191,15 @@ class M2N
   bool _isMasterConnected = false;
 
   bool _areSlavesConnected = false;
+
+
+  /// between two serial participants, only use the master-master com (required only in some unit tests)
+  bool _useOnlyMasterCom = false;
```
fsimonis: Can you explain the rational behind this?

src/m2n/M2N.hpp
```diff
@@ -182,6 +191,15 @@ class M2N
...
+
+  // to set _useOnlyMasterCom to true for unit tests
+  friend struct CplSchemeTests::SerialImplicitCouplingSchemeTests::testConfiguredAbsConvergenceMeasureSynchronized;
+  friend struct CplSchemeTests::CompositionalCouplingSchemeFixture;
+  friend struct CplSchemeTests::ExplicitCouplingSchemeFixture;
```
fsimonis: This invites to explode. A `friend` testing fixture would be better to handle this in the unit tests.

You can have a look at the use of the `WhiteboxAccessor`:
https://github.com/precice/precice/blob/a5bc1e7652ed957d817c9fcb46848fd760a474fc/src/testing/Testing.hpp#L196

src/utils/EventUtils.cpp
uekerman: Yes, it is odd. @fsimonis and me just agreed to fix this not here, but in #597 
I'll add a TODO in the sources.

src/m2n/M2N.hpp
uekerman: Rational is explained in the PR description:

> This makes live a bit more complicated for unit tests in cplscheme as they need inter-participant communication. I decided to not construct the complete overhead of mesh and vertex distribution definition for every test, but offered a loophole through a friend declaration in M2N and a fallback to the pure master-master communication. Suggestions welcome!

uekerman: I'll copy this to the code.

fsimonis:
I created a branch on my fork which runs this branch inside an ubuntu 18.04 container using `github-actions`.
The result is that it  fails similar to the Travis installation [click to see the log](https://github.com/fsimonis/precice/commit/2dbe33aa112e295e76ac0dcfaf3e34b9b6ad44be/checks?check_suite_id=357634235)

src/precice/config/ParticipantConfiguration.hpp
uekerman: I added more explanation in the comment.

src/precice/config/ParticipantConfiguration.cpp
uekerman: Yes, done.

src/m2n/M2N.hpp
uekerman: Done

src/m2n/M2N.hpp
uekerman: Done.

MakisH:
> Edit: I am pretty sure that I use the latest version of your branch, but for some reason I get the following:
> 
> ```
> ---[precice]  This is preCICE version 1.6.1
> ---[precice]  Revision info: v1.3.0-772-g2926429
> ```
> 
> I don't have time to check it now, but just FYI that this review may not be completely correct.

It actually looks like there are no tags after v1.3.0 in your fork, this is why Git reports that my version it 772 commits after v1.3.0. But merging this should be fine.

MakisH:
Important to note in the documentation **and release notes** is that the `distribution-type` does not exist anymore. Shall we add a special error message that hints that this was removed in preCICE v2?

Our current tutorials (and therefore probably many real simulations) use it, so I expect users to stumble upon the following error:
```
ERROR:  Wrong attribute "distribution-type"
```

MakisH:
I also observe similar behavior with our `FSI/3D_Tube/OpenFOAM-CalculiX` (which also uses nearest-projection mapping), **also in serial**.

[Fluid.log](https://github.com/precice/precice/files/3986046/Fluid.log)
[Solid.log](https://github.com/precice/precice/files/3986047/Solid.log)
[precice-config_serial.xml.txt](https://github.com/precice/precice/files/3986053/precice-config_serial.xml.txt)

uekerman:
> Important to note in the documentation **and release notes** is that the `distribution-type` does not exist anymore. Shall we add a special error message that hints that this was removed in preCICE v2?
> 
> Our current tutorials (and therefore probably many real simulations) use it, so I expect users to stumble upon the following error:
> 
> ```
> ERROR:  Wrong attribute "distribution-type"
> ```

Yes on the documentation and the release notes, but it would be overkill to add this as an error message in my opinion. Technically, it would already be hard to add it as this error is an automatic one thrown by the configuration. And the error message is already pretty explicit in my opinion: if we add this to some trouble-shooting section it should be easy to find.

uekerman:
Good catch on the first error! 
You use 
```xml
<mapping:nearest-projection direction="write" from="Solid-Mesh-Nodes" to="Fluid-Mesh-Centers" constraint="consistent" timing="initial"/>
```
which is not supported for a parallel participant. Problem is that the error message did no longer kick in. Fixed it.

uekerman:
Not sure about the second one, but you 
```xml
<m2n:sockets from="Fluid" to="Calculix" distribution-type="gather-scatter"/>
```
which should not work. So, I guess you are testing a different version here?

There is another problem in the CalculiX definition

```xml
<participant name="Calculix">
  <use-mesh name="Calculix_Mesh" provide="yes"/>
  <use-mesh name="Fluid-Mesh-Faces" from="Fluid"/>
  <write-data name="DisplacementDeltas0" mesh="Calculix_Mesh"/>
  <read-data  name="Forces0" mesh="Calculix_Mesh"/>
</participant>
```
as `Fluid-Mesh-Faces` has no mappings defined. This should however result in an error message later. Maybe only if run in parallel and it's ignored in serial.

PR #581, opened 2019-11-27
Remove deprecated features (other than tools and server)

uekerman:
* [X] `acceleration/HierarchicalAitkenAcceleration`
* [X] `action/ModifyCoordinatesAction`
* [X] `io/Import.hpp`
* ~`io/TXTReader`~ (still needed to test TXTWriter)
* [x] `query/FindVoxelContent`
* [x] `query/ExportVTKVoxelQueries`
* [x] `query/ExportVTKNeighbors`

Mesh stuff: 
* ~`mesh/PropertyContainer`~ is already removed in #585  
* ~`mesh/Group`~ let's also remove this in #585 
* conlusion: this PR needs to be merged before #585  

Needs further thought and a careful look into the integration tests
* ~`precice/MeshHandle`~ #588


fsimonis:
I migrated the MeshHandle in #585 https://github.com/precice/precice/pull/585/commits/823df9fcb3be8dc710b8a99e5c75abebaa44906b.

So It works, but the question remains whether it is useful or not.

uekerman:
`MeshHandle` opens our API to read coordinates and values of a received mesh (but not to modify them, which would lead to severe problems). To my best knowledge, historically, this had two use cases:
* (A) In an immersed boundary approach, to update the coordinates of the local Eulerian mesh, we need the values of the received Lagrangian mesh.
* (B) We don't want to define a local mesh, but just use the received mesh.

(A) is no longer supported as we currently don't support dynamically changing meshes. This might [get interesting again](https://github.com/precice/precice/projects/2).
(B) is currently only possible for a serial participant (but not documented). A parallel participant would need a re-partitioning, which is not possible without local information. (Or some alternative like holding a copy of the entire mesh in every rank). 

Furthermore, we still use `MeshHandle` in integration tests.

* as some sort of helper functionality, which could easily be replaced
  * [`testEpxlicitWithBlockDataExchange`](https://github.com/precice/precice/blob/develop/src/precice/tests/SerialTests.cpp#L348)
  * [`testExplicitWithDataExchange`](https://github.com/precice/precice/blob/develop/src/precice/tests/SerialTests.cpp#L185)
* to (implicitly) test (B), but could be replaced again 
  * [`testExplicitWithDataScaling`](https://github.com/precice/precice/blob/develop/src/precice/tests/SerialTests.cpp#L660)
* for white-box testing in [`ParallelTests`](https://github.com/precice/precice/blob/develop/src/precice/tests/ParallelTests.cpp), checking number of triangles after re-partitioning etc.

**Conclusion**
* For white-box testing, the concept is over-engineered in my opinion. Still, it would be good to have some entry point here.
* We should remove use case (B) as it does not easily generalize to a parallel participant and is currently not documented.
* There are still many steps to do before (A) could get interesting again. Also, we have already removed all integration tests for (A) some time ago.

My tendency would be to remove it. @fsimonis How do you see these points?

uekerman:
We decided to merge now, then merge #585 and do `MeshHandle` in a separate issue afterwards #588.

PR #585, opened 2019-11-27
Rework Mesh Internals

fsimonis:
This PR reworks the internals of the Mesh by
* removing PropertyContainer
* turning Mesh and the the Primitives into simple **copyable and/or movable** types (used to be a class hierarchy)
* storing the primitives directly in deque's for better locality
* removing the now-unused mesh::Group
* remove the ability to have submeshes, making #566 partially obsolete

The direct consequences of this PR are:
* dramatically fewer allocations
* dramatically faster iteration over primitives as they are now chunkwise contiguously stored.
* filterMesh and addMesh are faster
* forced the partial rewrite of `MeshHandle`

Notes:
* I had to use `std::deque` instead of `std::vector` as the deque does not invalidate references to contained objects on insertion. `Mesh::createVertex()` etc return a reference to the element.
* The Mesh is not copyable due to the `boost::signal`s.

Related to #581


fsimonis:
In #581, we decided to remove `mesh::Group` in this PR as well.

Issue #597, opened 2019-12-13
Refactor initialization and finalization of tests


uekerman:
We currently start all our tests from a single entry point in `src/testing/main.cpp` with some things initialized and finalized there:

https://github.com/precice/precice/blob/a5bc1e7652ed957d817c9fcb46848fd760a474fc/src/testing/main.cpp#L98-L131

These initialization and finalization efforts however make not sense for every test, e.g. in an integration test, the `EventRegistry` should be initialized with the `GlobalCommunicator` of the participant (so through the preCICE API), not with the `GlobalCommunicator` of the whole tests. 

For other tests, certain initialization is also missing: e.g. for integration tests in `PreciceTests/Serial/*`, we don't restrict the MPI communicator as we do for integration tests in `PreciceTests/Parallel/*`.

https://github.com/precice/precice/blob/a5bc1e7652ed957d817c9fcb46848fd760a474fc/src/testing/Fixtures.hpp#L81-L99

We currently work around this issues with several (historically grown) hacks, explicit setup of environments directly in the tests and prohibiting certain features for certain tests (e.g. PETSc based RBF in serial integration tests). Furthermore, our tests are sensitive to their order, which makes them very hard (thus time consuming) to maintain. 

Let's do a clean separation of different testcases and their initialization / finalization and a clean definition which things we initialize / finalize and which things we treat with `precice::testMode`. Either through different entry points or through fixtures alone. We need to distinguish the following cases:
* **(US)** serial unit tests, example: `MappingTests/NearestNeighborMapping`
* **(UP)** parallel unit tests on a single participant, example: `MappingTests/PetRadialBasisFunctionMapping/Parallel`
* **(UMS)** unit tests with two or more serial participants, example: `CplSchemeTests/SerialImplicitCouplingSchemeTests/*`
* **(UMP)** unit tests with multiple participants and at least one parallel participant, example: `PartitionTests/*`  
* **(IS)** integration tests with one, two or more serial participants, example: `PreciceTests/Serial/*`
* **(IP)** integration tests with multiple participants and at least one parallel participant, example: `PreciceTests/Parallel/*`

PR #600, opened 2019-12-17
WIP: Add PETSc test fixture

fsimonis:
This PR adds the fixture `PETScTestFixture`, which restricts the communicator to a given amount of ranks and then both initializes and finalizes PETSc.

This has two effects:
1. Every PetRBF test now runs in a fresh environment, which makes debugging saner as the test order does not matter any more.
2. The test `main` does not have to initialize or finalize PETSc any more.

This is a first step of #597

### DIFF
--- a/src/testing/Fixtures.hpp
+++ b/src/testing/Fixtures.hpp
@@ -1,6 +1,6 @@
 #pragma once
 
-#include "Testing.hpp"
+#include "testing/Testing.hpp"
 #include "utils/Parallel.hpp"
 #include "utils/MasterSlave.hpp"
 #include "com/Communication.hpp"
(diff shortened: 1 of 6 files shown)


fsimonis:
WIP as I have a more consistent method in mind

BenjaminRueth:
Is this related to https://github.com/precice/systemtests/issues/139?

src/testing/Fixtures.hpp
```diff
@@ -105,4 +107,35 @@ struct SplitParticipantsFixture {
 
 };
 
+#ifndef PRECICE_NO_PETSC
+
+/** Fixture to initialize and finalize PETSc for a given amount of ranks.
```
uekerman:
```suggestion
/** Fixture to initialize and finalize PETSc in a unit test for a given amount of ranks.
```

uekerman:
> Is this related to [precice/systemtests#139](https://github.com/precice/systemtests/issues/139)?

No, it's a pure refactoring to improve maintainability of tests, see #597.

fsimonis:
As it turns out, this PR is a fix for the problem, not a solution.

PR #577, opened 2019-11-26
Improve CMake Packaging

fsimonis:
This PR generally improves packaging with CMake by:

* using the `GNUInstallDirs`
* setting the correct SOVERSION
* generating a  correctly named debian package
* adding a correct changelog for the package
* making lintian happy

Furthermore, this PR:
* Enables C++11 as actual requirement
* Disables compiler extensions
* Fixes the use of an uninitialized (but harmless) variable

Future work:
* Split the package into a runtime and developer package
* Hide symbol names #623 

**TODO**
* [x] Add version and system to package name

PR #605, opened 2020-01-04
Fix the install target dirs to use the CMAKE flags

Scriptkiddi:
Building precice on systems like NixOS requires the usage of the recommended `GNUInstallDirs Variable` and its also useful for somebody that wants to control the build output
https://cmake.org/cmake/help/latest/command/install.html


fsimonis:
Thank you for opening this PR!
We actually have PR https://github.com/precice/precice/pull/577 in process which fixes multiple issues with the installation and packaging process. It also uses `GNUInstallDirs`!
I will review the PR in the upcoming week.

PR #608, opened 2020-01-13
Remove MeshHandle

### LINKED ISSUE #588: Remove `MeshHandle` and replace with easier solution for white-box testing
`MeshHandle` opens our API to read coordinates and values of a received mesh (but not to modify them, which would lead to severe problems). To my best knowledge, historically, this had two use cases:
* (A) In an immersed boundary approach, to update the coordinates of the local Eulerian mesh, we need the values of the received Lagrangian mesh.
* (B) We don't want to define a local mesh, but just use the received mesh.

(A) is no longer supported as we currently don't support dynamically changing meshes. This might [get interesting again](https://github.com/precice/precice/projects/2).
(B) is currently only possible for a serial participant (but not documented). A parallel participant would need a re-partitioning, which is not possible without local information. (Or some alternative like holding a copy of the entire mesh in every rank). 

Furthermore, we still use `MeshHandle` in integration tests.

* as some sort of helper functionality, which could easily be replaced
  * [`testEpxlicitWithBlockDataExchange`](https://github.com/precice/precice/blob/develop/src/precice/tests/SerialTests.cpp#L348)
  * [`testExplicitWithDataExchange`](https://github.com/precice/precice/blob/develop/src/precice/tests/SerialTests.cpp#L185)
* to (implicitly) test (B), but could be replaced again 
  * [`testExplicitWithDataScaling`](https://github.com/precice/precice/blob/develop/src/precice/tests/SerialTests.cpp#L660)
* for white-box testing in [`ParallelTests`](https://github.com/precice/precice/blob/develop/src/precice/tests/ParallelTests.cpp), checking number of triangles after re-partitioning etc.

**Conclusion**
* For white-box testing, the concept is over-engineered in my opinion. Still, it would be good to have some entry point here.
* We should remove use case (B) as it does not easily generalize to a parallel participant and is currently not documented.
* There are still many steps to do before (A) could get interesting again. Also, we have already removed all integration tests for (A) some time ago.

My tendency would be to remove it. @fsimonis How do you see these points?

_Originally posted by @uekerman in https://github.com/precice/precice/pull/581#issuecomment-560326848_

fsimonis:
This PR removes the `precice::MeshHandle` and the function `SolverInterface::getMeshHandle()` from the preCICE API.

The only remaining valid use-case is to use it in white-box testing. For this purpose, I added the following function:
```cpp
const Mesh &  SolverInterfaceImpl::mesh(const std::string & meshName) const
```

Closes #588


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@ All notable changes to this project will be documented in this file. For future
 
 ## develop
 
+- Remove `MeshHandle` and replace only valid use by adding `SolverInterfaceImpl::mesh()`.
```
uekerman:
```suggestion
- Remove `MeshHandle` from API and replace use in integration tests by `SolverInterfaceImpl::mesh()`.
```

src/precice/tests/SerialTests.cpp
```diff
@@ -200,29 +200,31 @@ BOOST_AUTO_TEST_CASE(testExplicitWithDataExchange,
...
+      //for (VertexIterator it = vertices.begin(); it != vertices.end(); it++) {
+      for (auto &vertex : vertices) {
+        Vector3d force(Vector3d::Constant(counter) + vertex.getCoords());
+        cplInterface.writeVectorData(forcesID, vertex.getID(), force.data());
       }
       maxDt = cplInterface.advance(maxDt);
       if (cplInterface.isCouplingOngoing()) {
         i = 0;
-        for (VertexIterator it = vertices.begin(); it != vertices.end(); it++) {
+        //for (VertexIterator it = vertices.begin(); it != vertices.end(); it++) {
```
uekerman:
```suggestion

```

PR #609, opened 2020-01-13
Clarify CMake Variables

### LINKED ISSUE #553: Clarify CMake Variables
# Problem

The main configuration variables for preCICE are: `MPI`, `PETSc`, and `PYTHON`.

While the first two are relatively straight forward, it is not clear that they (hopefully only) affect preCICE.
The latter is a common and very effective source of confusion as it solely relates to the module enabling user-provided python actions. It does not affect the python bindings.

# Proposition

1. Prefix all variables with `PRECICE_`. This clarifies that these variables are related to preCICE and also displays them as a neat group in the CMake GUI.
2. Rename the `PYTHON` variable to `PRECICE_PYTHONACTIONS` to prevent future misunderstandings and confusion.
3. Keep backwards compatibility by detecting the old variables and using their values. We should add a depreciation message in such cases.

### LINKED ISSUE #596: Default values of CMake options are confusing
Hi,

I had some students compiling preCICE today and we ran into some problems. This came from the fact that the students had *not* specified `-DBUILD_SHARED_LIBS=ON` when running CMake. I was not of aware that this could be a problem as I have set this flag in a script that I use for compiling preCICE so I had forgotten about that. 

At the same time, optional dependencies have there flags set to `ON` (PETSc, Python and MPI). Should it not rather be the other way around: Sane standard settings that lead to a reasonable build and optional settings should be turned on by the users if they need them? The option `CMAKE_BUILD_TYPE` only allows for three options, but none of them is specified as default?

Regarding `BUILD_SHARED_LIBS` the wiki in says several times that one should create the shared libraries and that the static library is even unsupported. From my experience today is also that the static library does not work well (or at all?!). Is there a reason why `BUILD_SHARED_LIBS` is not set to `ON` by default?  If it is not possible to set it to `ON` by default I would suggest that the parts about recommending the shared library a bit more visible by making it bold or so.

Parts of this issue have been part of #283, but nothing was changed.

Best,
Alex

fsimonis:
This PR clarifies the CMake variables

* by prefixing them with `PRECICE_`
* by renaming `MPI` to `PRECICE_MPICommunicator`
* by renaming `PETSC` to `PRECICE_PETScMapping`
* by renaming `PYTHON` to `PRECICE_PythonActions`
* by autoselection `Debug` if `CMAKE_BUILD_TYPE` is empty.

**TODO After merge**

* [ ] Update Systemtests
* [ ] Update Spack
* [ ] Update Wiki

Closes #596 
Closes #553


CMakeLists.txt
```diff
@@ -46,18 +46,29 @@ endif()
 # Setup Options
 #
 
-option(MPI "Enables MPI-based communication and running coupling tests." ON)
-option(PETSC "Enable use of the PETSc linear algebra library." ON)
-option(PYTHON "Python support" ON)
+if(NOT CMAKE_BUILD_TYPE)
+  message(WARNING "You did not specify a CMAKE_BUILD_TYPE.
+
+  We will assume you asked for a build a Debug build.")
```
MakisH:
```suggestion
  We will assume you asked for a Debug build.")
```

fsimonis:
@MakisH so you suggest using feature names such as:
```
PRECICE_PETScMapping
PRECICE_MPICommunicator
PRECICE_PythonActions
```

uekerman:
OK for me.
Wouldn't `MPICommunication` be better?

fsimonis:
@uekerman Yes, that sounds better.

---

@uekerman @MakisH can we agree on the following names?

```
PRECICE_PETScMapping
PRECICE_MPICommunication
PRECICE_PythonActions
```

uekerman:
Yes. Ready to merge from my perspective.

MakisH:
Looks good!

fsimonis:
@uekerman hold your horses!

I am going to change the names tomorrow. We can merge then.

cmake/BackwardsCompatibility.cmake
```diff
@@ -0,0 +1,16 @@
+# REMOVE THIS FILE AS SOON AS TESTS ARE FIXED
```
MakisH: Nice workaround to separate the expected from any unexpected regressions!

cmake/XSDKOptions.cmake
```diff
@@ -13,9 +13,9 @@ set(USE_XSDK_DEFAULTS OFF CACHE BOOL "Use XSDK defaults?")
 if(USE_XSDK_DEFAULTS)
   message(STATUS "XSDK defaults are enabled.")
   # 4. Debug and Release
-  set(CMAKE_BUILD_TYPE Debug CACHE STRING "The build type.")
+  set(CMAKE_BUILD_TYPE Debug CACHE STRING "The build type." FORCE)
```
MakisH: Why this `FORCE`?

fsimonis: This forces to override variables already located in the CMake cache.

cmake/BackwardsCompatibility.cmake
fsimonis: We just have to remember deleting it :joy:

Issue #366, opened 2019-04-24
Use a single step setup for C++-API and all bindings


BenjaminRueth:
Currently, some bindings ([C++](https://github.com/precice/precice/wiki/Adapter-Example), [Python](https://github.com/precice/precice/blob/e480aef4e3a74ccd405f03a2770bee18cdb548e9/src/precice/bindings/python/precice.pyx#L112-L122)) use a two step setup, where first the SolverInterface object is created and then the SolverInterface is configured:

```
SolverInterface( String participantName, int rank, int size );
configure( String configuration );
```

Other bindings ([C](https://github.com/precice/precice/blob/e480aef4e3a74ccd405f03a2770bee18cdb548e9/src/precice/bindings/c/SolverInterfaceC.h#L18-L22), Fortran) use a single step setup:

```
precicec_createSolverInterface(participantName, configuration, rand, size);
```

This is not consistent and (at least for me) there is no obvious reason why a two-step setup is necessary. Therefore, let's change all bindings to use a one step setup.

**Note:** If we want to have this change in `v1.x`, we have to maintain the API and therefore have to provide the two step setup. Only in `v2.0` we can entirely get rid of the two-step setup.

uekerman:
I'll remove the `good first issue` label as this has implications everywhere (all integration tests, all adapters, all bindings).

BenjaminRueth:
https://github.com/precice/python-bindings/pull/13 implements this change for the python bindings (which is good). However, https://github.com/precice/python-bindings/pull/13 currently relies on the two-level C++ interface, because the C++ interface does not provide the single step setup yet.

uekerman:
I'll suggest to add an additional constructor 
```c++
SolverInterface( String participantName, String configurationFileName, int rank, int size );
```
to the C++ API, which calls the old constructor and `configure`, and add a deprecated warning to the old constructor (not sure yet how to do this technically). The documentation and all official adapters use the new variant.

Why:
* For backwards compatibility. Breaking the API is much more severe than breaking the configuration in my opinion. We could then remove the old variant with v3.0. I would not want to make it a fixed law to do it this way every time. This time, it is easy to do and critical at the same time.
* Because we might need the split into constructor and `configure` in our integration tests. I suspect that this was also the reason for the design in the first place.

fsimonis:
@uekerman I agree with your suggested constructor.

However, we should definitely break the API now with the upcoming version 2.
[...]

PR #614, opened 2020-01-16
Single-step setup

fsimonis:
This PR changes the two-step setup of the `SolverInterface` to a single-step setup by:
* Removing `SolverInterface::configure()`
* Making `SolverInterfaceImpl::configure()`s private
* Extending the `SolverInterface` constructors with the parameter `const std::string& configurationFileName`.
* Migrating the C and Fortran bindings
* Migrating the C++ Solverdummy to the new interface.
* Removing `testing::SlimConfigurator`
* Migrating the tests to directly use the `SolverInterface` constructor.

## Adapter developers

Please merge the initialization with the configuration of preCICE as such:
```diff
-  SolverInterface interface(solverName, commRank, commSize);
-  interface.configure(configFileName);
+  SolverInterface interface(solverName, configFileName, commRank, commSize);
```

**TO-Update**

* [ ] openfoam https://github.com/precice/openfoam-adapter
* [ ] dealii https://github.com/precice/dealii-adapter
* [ ] su2 https://github.com/precice/su2-adapter

* [ ] elastictube1d https://github.com/precice/elastictube1d
* [ ] ASTE https://github.com/precice/aste
* [ ] fem-shell https://github.com/precice/fem-shell
* [ ] python bindings https://github.com/precice/python-bindings/pull/18
* [ ] matlab bindings https://github.com/precice/matlab-bindings/issues/8

Related to #366

### DIFF
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -51,10 +51,11 @@ bool syncMode = false;
 namespace impl {
 
 SolverInterfaceImpl::SolverInterfaceImpl(
-    std::string participantName,
-    int         accessorProcessRank,
-    int         accessorCommunicatorSize,
-    void *      communicator)
+    std::string        participantName,
+    const std::string &configurationFileName,
+    int                accessorProcessRank,
+    int                accessorCommunicatorSize,
+    void *             communicator)
     : _accessorName(std::move(participantName)),
       _accessorProcessRank(accessorProcessRank),
       _accessorCommunicatorSize(accessorCommunicatorSize)
@@ -77,6 +78,8 @@ SolverInterfaceImpl::SolverInterfaceImpl(
 #endif
 
   logging::setParticipant(_accessorName);
+
+  configure(configurationFileName);
 }
 
 SolverInterfaceImpl::SolverInterfaceImpl(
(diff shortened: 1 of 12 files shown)


src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -51,10 +51,11 @@ bool syncMode = false;
 namespace impl {
 
 SolverInterfaceImpl::SolverInterfaceImpl(
-    std::string participantName,
-    int         accessorProcessRank,
-    int         accessorCommunicatorSize,
-    void *      communicator)
+    std::string        participantName,
+    const std::string &configurationFileName,
```
uekerman: Why do we treat `participantName` and `configurationFileName` differently? i.e. `std::string` vs. `const std::string &`?

fsimonis: We need to store the `participantName` inside the object, so we use call-by-value and move construct the member which holds it in the constructor of the `SolverInterfaceImpl`.

The configuration file name does not need to be stored as we simply call `configure` in the constructor of `SolverInterfaceImpl`. So we simply pass it by const ref.

This is not important and I wouldn't mind changing both to pass by value for the sake of readability.

uekerman: Thanks for the explanation. I merely asked for learning, but you might have a point. A user not deep into C++ might wonder if she needs to treat `participantName` and `configurationFileName` differently when calling the constructor.

fsimonis: The user does not need to treat both differently.
We just do things slightly different in the implementation.

BenjaminRueth:
We want to merge https://github.com/precice/systemtests/pull/148 and https://github.com/precice/systemtests/pull/128 before we merge this one (latest until Monday!).

PR #617, opened 2020-01-17
Add Mesh Name to Connection Info Publisher

### LINKED ISSUE #570: More context required in com::Communication
# Problem

When implementing the 2-phase initialization, there can be cases where the communication build-up for meshes between the same participants mix up. This leads to a deadlock.

An example being the elastic1d-case:

1d

The single M2N has 2 `m2n::DistributedCommunication` instances for the 2 meshes.
Running with 4 solid and 3 fluid ranks, the communication map looks as follows:

Solid | Fluid
--- | ---
0 | 0,1
1 | 0,1,2,3
2 | 2,3

The mix-up happens in this loop in both the accept and request variants:

https://github.com/precice/precice/blob/e7bd717e8e58e034a4a347baa68c3247a7bb7751/src/m2n/M2N.cpp#L135-L138

# Solution

The communication-buildup is lacking context. It knows the participants and the rank of the requester involved in the the connection, but lacks the additional information of what `m2n::DistantCommunication` it is connecting.
We need to forward this information to the Communicator, so that we can incorporate it into the `com::ConncectionInfoPublisher`.

Directly related to #142

fsimonis:
This PR adds the mesh-name to the context used in the connection info publisher.

This required passing the name of the mesh from the distributed communication down to:
1. the `m2n::M2N`
2. the `com::Communication::accept/requestConnection`
3. the `com::ConnectionInfoWriter`

Resolves #570

### DIFF
--- a/src/com/Communication.hpp
+++ b/src/com/Communication.hpp
@@ -76,10 +76,12 @@ public:
    *
    * @param[in] acceptorName Name of calling participant.
    * @param[in] requesterName Name of remote participant to connect to.
+   * @param[in] tag Tag for establishing this connection
    * @param[in] acceptorRank Rank of the accpeting process, usually the calling one.
    */
   virtual void acceptConnection(std::string const &acceptorName,
                                 std::string const &requesterName,
+                                std::string const &tag,
                                 int                acceptorRank) = 0;
 
   /**
@@ -93,11 +95,13 @@ public:
    *
    * @param[in] acceptorName Name of calling participant.
    * @param[in] requesterName Name of remote participant to connect to.
+   * @param[in] tag Tag for establishing this connection
    * @param[in] acceptorRank Rank of accepting server, usually the rank of the current process.
    * @param[in] requesterCommunicatorSize Size of the requester (N)
    */
   virtual void acceptConnectionAsServer(std::string const &acceptorName,
                                         std::string const &requesterName,
+                                        std::string const &tag,
                                         int                acceptorRank,
                                         int                requesterCommunicatorSize) = 0;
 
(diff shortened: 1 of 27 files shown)


src/com/Communication.hpp
```diff
@@ -93,11 +95,13 @@ class Communication {
    *
    * @param[in] acceptorName Name of calling participant.
    * @param[in] requesterName Name of remote participant to connect to.
+   * @param[in] meshName Name of the involved mesh
    * @param[in] acceptorRank Rank of accepting server, usually the rank of the current process.
    * @param[in] requesterCommunicatorSize Size of the requester (N)
    */
   virtual void acceptConnectionAsServer(std::string const &acceptorName,
                                         std::string const &requesterName,
+                                        std::string const &meshName,
```
uekerman: Maybe, instead of `meshName`, we could use a more neutral name. In the end, from a design perspective, there is no mesh needed to establish a communication and the `meshName` only serves as a tag for establishing the communication. What about `std::string const tag`?

src/m2n/PointToPointCommunication.cpp
```diff
@@ -300,7 +300,7 @@ void PointToPointCommunication::acceptConnection(std::string const &acceptorName
     // Establish connection between participants' master processes.
     auto c = _communicationFactory->newCommunication();
 
-    c->acceptConnection(acceptorName, requesterName, utils::MasterSlave::getRank());
+    c->acceptConnection(acceptorName, requesterName, _mesh->getName(), utils::MasterSlave::getRank());
```
uekerman: Not completely sure, but might be dangerous to use the mesh name here as it could interfere with the slaves communication below. It's a bit ugly here as `c` is again a (temporary) master-master communication. What about sth like `"TMP-MASTERCOM-" +_mesh->getName()`?

fsimonis: Got it. I implemented your suggestion.

src/com/Communication.hpp
fsimonis: True, I initially thought about calling it `context`, which didn't quite fit. `tag` is perfect!

Issue #555, opened 2019-11-11
Mention data names in convergence and iterations file


uekerman:
Currently, e.g.
`precice-SOLIDZ-iterations.log`:
```
Timesteps  Total_Iterations  Iterations  Convergence  avgConvRate(0)  avgConvRate(1)  avgConvRate(2)  avgConvRate(3)  avgConvRate(4)  deleted_Columns
1  4  4  1  0.0131513617329646  0.0097355863572098  0.0121224497932821  0.0148052718537714  0.0097533280718913  0
2  8  4  1  0.0149890377161203  0.0131392422263492  0.0306217024085128  0.0316058307866920  0.0189617635359644  1
3  13  5  1  0.0083313444673868  0.0090989214732156  0.0335150447116706  0.0433537858697371  0.0314135230849259  0
```
and
`precice-SOLIDZ-convergence.log`:
```
Timestep  Iteration  resNorm(0)  resNorm(1)  resNorm(2)  resNorm(3)  resNorm(4)
1  1  1.0000000000000000  1.0000000000000000  1.0000000000000000  1.0000000000000000  1.0000000000000000
1  2  0.0028264970995254  0.0018828719443888  0.0014289099428535  0.0039439618931841  0.0021543142030861
1  3  0.0000019852262002  0.0000009146078748  0.0000087277900904  0.0000057307160882  0.0000051453312708
1  4  0.0000000299145789  0.0000000089835596  0.0000000215954161  0.0000000480469191  0.0000000090492238
```

`avgConvRate(0) `, `avgConvRate(1)`, ... and `resNorm(0)`, `resNorm(1)`, ... come in the same order as in the preCICE configuration specified. It would be much easier to understand, however, if we directly specified the names: `avgConvRate(Forces)` or `resNorm(Displacements)`.

While doing this, we could also bring some language consistency into these files (`ResNorm`, `TotalIterations` ...) 

... and also add the names to the convergence INFO statement:
```
preCICE: relative convergence measure: two-norm diff = 1.79666e-09, relative limit = 1.92741e-10, conv = false
preCICE: relative convergence measure: two-norm diff = 4.39099e-09, relative limit = 2.07417e-10, conv = false
preCICE: relative convergence measure: two-norm diff = 0.0336136, relative limit = 2.06253e-05, conv = false
preCICE: relative convergence measure: two-norm diff = 0.0114555, relative limit = 2.06612e-05, conv = false
[...]

PR #620, opened 2020-01-21
Add data names to convergence and iteration file

KyleDavisSA:
This PR aims to resolve https://github.com/precice/precice/issues/555. The previous output went from dataID 0 and upwards:

> resNorm(0)  resNorm(1)  resNorm(2)

printing `convMeasure.data->getID()` does not print the data ID in this order. The different data ID's are looped through by 

https://github.com/precice/precice/blob/fe00f53e059f57e6c390d4fcfa60a25da896fe1c/src/cplscheme/BaseCouplingScheme.cpp#L695-L703

An example output was generated for the cylinderFlap example

```
Timestep  Iterations  ResNorm(3)  ResNorm(2)  
1  1  1.0000000000000000  1.0000000000000000  
1  2  1.0223306624150461  1.2125016658484906  
1  3  1.0104960419007956  1.5949903615016618  
1  4  0.9907038391847834  6.9682477895700625
```
The same dataID is also found when using 
`sstm << "ResNorm(" << _convergenceMeasures[i].data->getID() << ");`

For now, it prints the dataID and not the dataName.

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -575,7 +575,7 @@ bool BaseCouplingScheme::measureConvergence(
   PRECICE_ASSERT(_convergenceMeasures.size() > 0);
   if (not utils::MasterSlave::isSlave()) {
     _convergenceWriter->writeData("Timestep", _timesteps);
-    _convergenceWriter->writeData("Iterations", _iterations);
+    _convergenceWriter->writeData("Iteration", _iterations);
   }
   for (size_t i = 0; i < _convergenceMeasures.size(); i++) {
     ConvergenceMeasure &convMeasure = _convergenceMeasures[i];
@@ -691,8 +691,7 @@ void BaseCouplingScheme::initializeTXTWriters()
     }
 
     if (not doesFirstStep()) {
-      for (size_t i = 0; i < _convergenceMeasures.size(); i++) {
-        ConvergenceMeasure &convMeasure = _convergenceMeasures[i];
+      for (ConvergenceMeasure &convMeasure : _convergenceMeasures) {
 
         // only for fine model optimization, i.e., coupling
         if (convMeasure.level > 0)
@@ -728,8 +727,8 @@ void BaseCouplingScheme::advanceTXTWriters()
     int converged = _iterations < _maxIterations ? 1 : 0;
     _iterationsWriter->writeData("Convergence", converged);
 
-    int i = -1;
     if (not doesFirstStep()) {
+      int i = -1;
       for (ConvergenceMeasure &convMeasure : _convergenceMeasures) {
         i++;
 
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -595,7 +595,7 @@ bool BaseCouplingScheme::measureConvergence(
 
     if (not utils::MasterSlave::isSlave()) {
       std::stringstream sstm;
-      sstm << "ResNorm(" << convMeasure.data->getName() << ")";
+      sstm << "ResNorm(" << convMeasure.data->getID() << ")";
       _convergenceWriter->writeData(sstm.str(), convMeasure.measure->getNormResidual());
     }
 
@@ -692,14 +692,15 @@ void BaseCouplingScheme::initializeTXTWriters()
 
     int i = -1;
     if (not doesFirstStep()) {
-      for (ConvergenceMeasure &convMeasure : _convergenceMeasures) {
-        i++;
+      for (size_t i = 0; i < _convergenceMeasures.size(); i++) {
+        ConvergenceMeasure &convMeasure = _convergenceMeasures[i];
+
         // only for fine model optimization, i.e., coupling
         if (convMeasure.level > 0)
           continue;
         std::stringstream sstm, sstm2;
-        sstm << "AvgConvRate(" << convMeasure.data->getName() << ")";
-        sstm2 << "ResNorm(" << convMeasure.data->getName() << ")";
+        sstm << "AvgConvRate(" << convMeasure.data->getID() << ")";
+        sstm2 << "ResNorm(" << convMeasure.data->getID() << ")";
         _iterationsWriter->addData(sstm.str(), io::TXTTableWriter::DOUBLE);
         _convergenceWriter->addData(sstm2.str(), io::TXTTableWriter::DOUBLE);
       }
@@ -738,7 +739,7 @@ void BaseCouplingScheme::advanceTXTWriters()
           continue;
 
         std::stringstream sstm;
-        sstm << "AvgConvRate(" << convMeasure.data->getName() << ")";
+        sstm << "AvgConvRate(" << convMeasure.data->getID() << ")";
         if (math::equals(_firstResiduumNorm[i], 0.)) {
           _iterationsWriter->writeData(sstm.str(), std::numeric_limits<double>::infinity());
         } else {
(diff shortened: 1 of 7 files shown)


uekerman:
The order of convergence measures is determined by their order in the config. And I guess that's the expected behavior. No need to sort them by ID. 
`_convergenceMeasures[i].data->getName()` should give you the name.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -676,34 +676,35 @@ void BaseCouplingScheme::initializeTXTWriters()
...
         std::stringstream sstm, sstm2;
-        sstm << "avgConvRate(" << i << ")";
-        sstm2 << "resNorm(" << i << ")";
+        sstm << "AvgConvRate(" << convMeasure.data->getID() << ")";
+        sstm2 << "ResNorm(" << convMeasure.data->getID() << ")";
         _iterationsWriter->addData(sstm.str(), io::TXTTableWriter::DOUBLE);
         _convergenceWriter->addData(sstm2.str(), io::TXTTableWriter::DOUBLE);
       }
-      _iterationsWriter->addData("deleted_Columns", io::TXTTableWriter::INT);
+      _iterationsWriter->addData("deletedColumns", io::TXTTableWriter::INT);
```
uekerman: Wouldn't `"DeletedColumns"` be more consistent?

KyleDavisSA:
Previously, the output was 
`resNorm(0) resNorm(1) ...`

If you print the dataID for each data name in the solvers, you find for the Fluid solver:
```
Forces = 0
Displacements = 1
```
and for the Solid solver:
```
Forces = 2
Displacements = 3
```

As an example, the current convergence file will print the dataID by looping through variable `i`:
https://github.com/precice/precice/blob/c0548c720ca681b3e1bac75e8565f14b49b8a8ad/src/cplscheme/BaseCouplingScheme.cpp#L598
```
Timestep  Iterations  ResNorm(0)  ResNorm(1)  
1  1  1.0000000000000000  1.0000000000000000  
1  2  1.0227066777948459  1.2158700458329670  
1  3  1.0113675981800501  1.7151706495425945  
```
By printing the dataID for the associated resNorm measure will print:
https://github.com/precice/precice/blob/3f732ff84066e1cc89b3bd145d10613b3cf34b6d/src/cplscheme/BaseCouplingScheme.cpp#L598
```
Timestep  Iterations  ResNorm(3)  ResNorm(2)  
1  1  1.0000000000000000  1.0000000000000000  
1  2  1.0227066777948459  1.2158700458329670  
1  3  1.0113675981800501  1.7151706495425945 
```
And by printing the dataName:
```
Timestep  Iterations  ResNorm(Displacements)  ResNorm(Forces)  
1  1  1.0000000000000000  1.0000000000000000  
1  2  1.0227066777948459  1.2158700458329670  
1  3  1.0113675981800501  1.7151706495425945 
```
Does the current implementation printout the resNorm for dataID `3` and names it as `ResNorm(0)`?

uekerman:
Not sure about which ID is used exactly in the convergence measure, but we should actually never output IDs. So, the current state is certainly sub-optimal.   
```
ResNorm(Displacements)  ResNorm(Forces)
```
does the job.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -676,34 +676,34 @@ void BaseCouplingScheme::initializeTXTWriters()
...
       _convergenceWriter->addData("Timestep", io::TXTTableWriter::INT);
-      _convergenceWriter->addData("Iteration", io::TXTTableWriter::INT);
+      _convergenceWriter->addData("Iterations", io::TXTTableWriter::INT);
     }
 
-    int i = -1;
     if (not doesFirstStep()) {
-      for (ConvergenceMeasure &convMeasure : _convergenceMeasures) {
-        i++;
+      for (size_t i = 0; i < _convergenceMeasures.size(); i++) {
```
MakisH: Why converting from the ranged-for (previous) to the "classic" for-loop? We don't need the counter `i` anywhere and a ranged-for is safer.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -676,34 +676,34 @@ void BaseCouplingScheme::initializeTXTWriters()
...
 
-    _iterationsWriter->addData("Timesteps", io::TXTTableWriter::INT);
-    _iterationsWriter->addData("Total_Iterations", io::TXTTableWriter::INT);
+    _iterationsWriter->addData("Timestep", io::TXTTableWriter::INT);
+    _iterationsWriter->addData("TotalIterations", io::TXTTableWriter::INT);
     _iterationsWriter->addData("Iterations", io::TXTTableWriter::INT);
     if (hasCoarseModelOptimization) {
-      _iterationsWriter->addData("Total_Iterations_Surrogate_Model", io::TXTTableWriter::INT);
-      _iterationsWriter->addData("Iterations_Surrogate_Model", io::TXTTableWriter::INT);
+      _iterationsWriter->addData("TotalIterations_SurrogateModel", io::TXTTableWriter::INT);
```
MakisH: I understand you are trying to nicely separate between `TotalIterations` and `TotalIterations_SurrogateModel`, but wouldn't just `TotalIterationsSurrogateModel` be good (and consistent) as well? Does [clang-format](http://xgm.de/precice/docs/develop/tooling.html) suggest any change?

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -718,33 +718,35 @@ void BaseCouplingScheme::advanceTXTWriters()
...
     if (hasCoarseModelOptimization) {
-      _iterationsWriter->writeData("Total_Iterations_Surrogate_Model", _totalIterationsCoarseOptimization);
-      _iterationsWriter->writeData("Iterations_Surrogate_Model", _iterationsCoarseOptimization);
+      _iterationsWriter->writeData("TotalIterations_SurrogateModel", _totalIterationsCoarseOptimization);
+      _iterationsWriter->writeData("Iterations_SurrogateModel", _iterationsCoarseOptimization);
     }
     int converged = _iterations < _maxIterations ? 1 : 0;
     _iterationsWriter->writeData("Convergence", converged);
 
+    int i = -1;
```
MakisH: We don't need `i` anywhere, right?

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -575,7 +575,7 @@ bool BaseCouplingScheme::measureConvergence(
   PRECICE_ASSERT(_convergenceMeasures.size() > 0);
   if (not utils::MasterSlave::isSlave()) {
     _convergenceWriter->writeData("Timestep", _timesteps);
-    _convergenceWriter->writeData("Iteration", _iterations);
+    _convergenceWriter->writeData("Iterations", _iterations);
```
MakisH: From the results, I understand that this should be `Iteration` (as before), not `Iterations`. For example:

```
Timestep  Iterations  ResNorm(Displacements0)  ResNorm(Forces0)  
1  1  0.0003191866323491  1.0000000000000000  
1  2  0.0003159904961397  0.0000149565768695  
1  3  0.0000000004065084  0.0014792261490834  
2  1  0.0007445657741119  0.2056384940762610  
2  2  0.0000002404722286  0.0039091692413695  
3  1  0.0005485913384467  0.1860456395794740  
3  2  0.0000000664794586  0.0033499436260277  
```
Looking at the `*-convergence.log` file I usually look if the residual norm decreases with every iteration. `Iterations` is not what is meant here, I guess (even if the variable is named `_iterations`).

Am I misunderstanding something?

KyleDavisSA: I think that would be right. I vote for keep `TotalIterations` the same and make it `Iteration` (singular, refering to the result of a specific iteration). I don't think changing the variable `_iterations` is necessary.

src/cplscheme/BaseCouplingScheme.cpp
KyleDavisSA: `i` is used in calculation of the residual below (line742)

MakisH: Then it's ok, I guess! :-)

fsimonis:
@KyleDavisSA the cplscheme tests fail as their data names don't match on write.

PR #624, opened 2020-01-27
Change default mpi com in m2n to MPISinglePortsCom

uekerman:
`MPISinglePortsCom` is the more efficient variant developed by @floli. `MPIPortsCom` is the old, now kind of deprecated, variant. This PR changes the default setting. `"mpi"` now points to the new implementation and the old variant is accessible via `"mpi-multiple-ports"`. 
In my opinion it is still too early to completely remove the old implementation.

### DIFF
--- a/docs/changelog/624.md
+++ b/docs/changelog/624.md
@@ -1,0 +1,1 @@
+- Change `m2n:mpi` to use the more efficient single-ports implementation. To use the old implementation, use `m2n:mpi-mulitple-ports`.
--- a/src/m2n/config/M2NConfiguration.cpp
+++ b/src/m2n/config/M2NConfiguration.cpp
@@ -162,7 +162,7 @@ void M2NConfiguration::xmlTagCallback(const xml::ConfigurationContext &context, 
     } else if (tag.getName() == "mpi-multiple-ports") {
       std::string dir = tag.getStringAttributeValue(ATTR_EXCHANGE_DIRECTORY);
 #ifdef PRECICE_NO_MPI
-      PRECICE_ERROR("Communication type \"mpi\" can only be used if preCICE was compiled with MPI support enabled. "
+      PRECICE_ERROR("Communication type \"mpi-multiple-ports\" can only be used if preCICE was compiled with MPI support enabled. "
                     "Either switch to a \"sockets\" communication or recompile preCICE with \"PRECICE_MPICommunication=ON\".");
 #else
 #ifdef OMPI_MAJOR_VERSION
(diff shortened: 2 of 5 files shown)


uekerman:
All test involving `MPISinglePortsCom` currently fail, the new unit tests and the integration tests. I am not sure of the reason currently and would therefore move the PR to the next release.

fsimonis:
I rebased this branch on develop ( I had to drop 57861e2 ) and all tests pass.

We now need to restore the naming of test configuration files and maybe also add some more to test both `mpi` and  `mpi-multiple-ports`.

fsimonis:
I pushed the rebased version of this PR.

fsimonis:
Merge this after the release 2.1.0.
There is still more testing to be done.

fsimonis:
Let's rebase this on develop an check if it still works.

BenjaminRodenberg:
This PR is now rebased on top of the current state of develop (1a5f5d9cfb39769534ebbdfe0d4f645feb25f591).

fsimonis:
We should really keep the `mpi-singleports` tag to ensure backwards compatibility. This should not be too difficult.

BenjaminRodenberg:
### Todos

- [ ] implement `mpi-singleports`
- [ ] run a parallel case to make sure this PR does not break anything (i.e. parallel fenics-fenics)
- [ ] add MPICH to github actions and run tests #713

BenjaminRodenberg:
Currently I do not really have time to work on this PR. If somebody else would like to work on it, it's basically free.

fsimonis:
I just rebased this on develop. Let's see how the tests do.

fsimonis:
The mpiports tests still fail with OpenMPI 4.1.1.

So we could either leave it as it is or change `mpi` to use multiple ports for OpenMPI and single-ports for everything else.

uekerman:
> The mpiports tests still fail with OpenMPI 4.1.1.

You mean they fail for the single-port implementation, but not for the multiple-ports implementation?
That is works with the old multiple-ports implementation is new to me.

We already give build-time and run-time warnings if OpenMPI is used, right?

fsimonis:
> You mean they fail for the single-port implementation, but not for the multiple-ports implementation?
That is works with the old multiple-ports implementation is new to me.

Just checked again. Both implementations fail for OpenMPI.

> We already give build-time and run-time warnings if OpenMPI is used, right?

There are warnings in place #797

uekerman:
> Just checked again. Both implementations fail for OpenMPI.

But that's not an argument to *not* merge this PR, right?

fsimonis:
> But that's not an argument to not merge this PR, right?

True.

The last thing that is missing is the backwards-compatibility for the `mpi-singleports' option.

fsimonis:
The configuration part works. Merging

PR #625, opened 2020-01-27
Move MATLAB solverdummy to this repository

BenjaminRueth:


BenjaminRueth:
We decided to move the solverdummies to their respective repository, if they depend on bindings that are distributed independently. This applies to the python and matlab solver dummy.

* matlab solverdummy is fine: see https://github.com/precice/matlab-bindings/tree/develop/solverdummy
* python solverdummy will be done soon: https://github.com/precice/python-bindings/issues/23

Issue #626, opened 2020-01-27
AUR package is missing Eigen


MakisH:
Our [AUR package](https://aur.archlinux.org/cgit/aur.git/tree/PKGBUILD?h=precice) does not specify Eigen as a dependency:
```
depends=('openmpi' 'boost')
```
The fix is easy:
```
depends=('openmpi' 'boost' 'eigen')
```

fsimonis:
We need eigen only at build-time, so the fix would be:
`makedepends=('cmake' 'eigen')`

PR #632, opened 2020-01-28
Add Eigen dependency for AUR package

dinimar:
Add eigen build-time dependency for AUR package mentioned in #626


tools/packaging/Arch_User_Repository/PKGBUILD
```diff
@@ -8,7 +8,7 @@ arch=('x86_64')
 url="https://www.precice.org"
 license=('LGPL3')
 depends=('openmpi' 'boost')
-makedepends=('cmake')
+makedepends=('cmake', 'eigen')
```
MakisH: This is actually not valid, it needs to be:
```suggestion
makedepends=('cmake' 'eigen')
```

MakisH: Other than that, it's ok, I have tested it in Manjaro.

dinimar:
> That's great! Could you also submit the patch to AUR, please?

If all I need to do is listed at this [wiki](https://wiki.archlinux.org/index.php/AUR_submission_guidelines#Publishing_new_package_content) then I can. At least, in wiki it doesn't seem hard.

PR #630, opened 2020-01-27
Restructure tools

fsimonis:
This PR restructures the tools according to https://github.com/precice/precice/issues/512#issuecomment-571528689.

Related to #512

PR #634, opened 2020-01-29
C-bindings: Use precice::SolverInterface instead of precice::impl::SolverInterfaceImpl

MakisH:
We consider the C++ API as main and everything else as an additional layer. Breaking through layers could lead to bugs due to duplication.

I have tested this with the C solverdummy.

This only changes the implementation file, the header / signatures remain the same.

Depends on #630.


extras/bindings/c/SolverInterfaceC.cpp
```diff
@@ -15,149 +15,150 @@ void precicec_createSolverInterface(
 {
   std::string stringAccessorName(participantName);
   std::string stringConfigFileName(configFileName);
-  impl = new precice::impl::SolverInterfaceImpl(stringAccessorName,
-                                                stringConfigFileName,
-                                                solverProcessIndex, solverProcessSize);
+  interface = new precice::SolverInterface(stringAccessorName,
+                                        stringConfigFileName,
+                                        solverProcessIndex,
+                                        solverProcessSize);
```
fsimonis: Can you run format-all on the code-base? This looks odd.

MakisH: Done, thanks!

MakisH:
Rebased on `develop` and looks good now.

PR #635, opened 2020-01-29
Add getVersionInformation to C, Fortran bindings

MakisH:
Similar changes as #477, this time for the C and Fortran bindings.

- [x] C (tested with the C solverdummy)
- [x] Fortran (tested with the Fortran solverdummy)
   - The FASTEST bindings are not ported, as they are a special case.
   - Note for using it: Make sure to define a large-enough character array to store the information (e.g. 250 characters).

Fortran 2003 / "Fortran module" will be added in a separate PR, as the bindings have been already moved.

This depends on #630.


extras/bindings/c/include/precice/SolverInterfaceC.h
```diff
@@ -413,6 +413,17 @@ void precicec_readScalarData(
     int     valueIndex,
     double *dataValue);
 
+/** 
+ * @brief Returns information on the version of preCICE.
+ *
+ * Returns a semicolon-separated C-string containing:
```
MakisH: We can still say that a `char *` is a `C-string`, I guess?

PR #646, opened 2020-01-31
Remove last traces of SCons

MakisH:
While we have removed SCons support for preCICE, there are still a few mentions in the solver dummies and developer documentation. This hopefully removes everything.

The `tools/testing/communication_dummies/SConstruct` is intentionally kept, until we revive this tool.


MakisH:
I also removed SCons from our wiki pages. The old [build preCICE using SCons](https://github.com/precice/precice/wiki/Building:-using-SCons) page still exists, but is not references. In the special systems pages I added an out-of-date warning.

PR #650, opened 2020-01-31
Add safety check for user defined MPI Communicator

### LINKED ISSUE #648: Duplicated (rank, size) info when passing an MPI Communicator to preCICE
Since recently, we have the possibility to pass an MPI Communicator to preCICE.

https://github.com/precice/precice/blob/139118546d992cbf48fa6bd77c0a35f8ca7700d8/src/precice/SolverInterface.hpp#L68-L73

We should either implement a safety `PRECICE_CHECK` that the passed `rank` and `size` matches the communicator or remove these two arguments. The latter would be a breaking change.

uekerman:
closes #648

### DIFF
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -80,6 +80,21 @@ SolverInterfaceImpl::SolverInterfaceImpl(
   logging::setParticipant(_accessorName);
 
   configure(configurationFileName);
+
+// This block cannot be merge with the one above as only configure calls
+// utils::Parallel::initializeMPI, which is needed for getProcessRank.
+#ifndef PRECICE_NO_MPI
+  if (communicator != nullptr) {
+    PRECICE_CHECK(_accessorProcessRank == utils::Parallel::getProcessRank(),
+                  "The passed solverProcessIndex (" << _accessorProcessRank
+                                                    << ") does not match the rank of the passed MPI communicator ("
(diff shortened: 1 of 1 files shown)


src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -80,6 +80,21 @@ SolverInterfaceImpl::SolverInterfaceImpl(
   logging::setParticipant(_accessorName);
 
   configure(configurationFileName);
+
+// This block cannot be merge with the one above as only configure calls
+// utils::Parallel::initializeMPI, which is needed for getProcessRank.
+#ifndef PRECICE_NO_MPI
+  if (communicator != nullptr) {
```
fsimonis: Why not check it straight after setting the communicator here?
https://github.com/precice/precice/blob/3699cc0b3ba1ed6a907a86abc97367df3389584f/src/precice/impl/SolverInterfaceImpl.cpp#L70-L78

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -80,6 +80,21 @@ SolverInterfaceImpl::SolverInterfaceImpl(
   logging::setParticipant(_accessorName);
 
   configure(configurationFileName);
+
+// This block cannot be merge with the one above as only configure calls
```
MakisH:
```suggestion
// This block cannot be merged with the one above, as only configure calls
```

PR #649, opened 2020-01-31
Add option to pass custom MPI communicator to C bindings

MakisH:
This adds the following methods:
- C: `precicec_createSolverInterface_withCommunicator`
   - Tested with the solverdummy and `MPI_COMM_WORLD`.
   - Use it as e.g.
   ```c
   MPI_Comm comm = MPI_COMM_WORLD; // or e.g. a splitted communicator
   precicec_createSolverInterface_withCommunicator(participantName, configFileName, solverProcessIndex, solverProcessSize, &comm);
   ```
- Fortran: More complicated, implemented in https://github.com/precice/precice/pull/651

Related to #382

Issue #520, opened 2019-09-13
Make sure that complete API is supported by bindings


BenjaminRueth:
example for a missing function: `getVersion` #380

# bindings that should be checked

- [x] python refer to https://github.com/precice/python-bindings/issues/16
- [x] MATLAB refer to https://github.com/precice/matlab-bindings/issues/6
- [x] C
- [ ] Fortran90
- [ ] Fortran2003

MakisH:
I compared the C++ API with C, Fortran, and the Fortran 2003 module

## C

- Missing custom MPI communicator (see #479)
- `isTimeWindowComplete` has a different name: `precicec_isCouplingTimeWindowComplete`
- Missing `getMeshIDs`

Everything else looks good.

## Fortran

- Missing custom MPI communicator (see #479)

Many functions have different style names and are in a different order than in C, but everything looks good.

## Fortran 2003 module

I did not go into details, but looks the same as Fortran.

So, this is a duplicate of #479.

IshaanDesai:
There is a name inconsistency in the Fortran bindings (withint the preCICE Repo and not the external module) when compared to the C++ API:
* The function `isActionRequired` in the C++ API is `precicef_action_required` in the Fortran bindings in [this line](https://github.com/precice/precice/blob/develop/extras/bindings/fortran/include/precice/SolverInterfaceFortran.hpp#L170). The consistent name would be `precicef_is_action_required`.
* The function `isCouplingOngoing` in the C++ API is `precicef_ongoing` in the Fortran bindings in [this line](https://github.com/precice/precice/blob/develop/extras/bindings/fortran/include/precice/SolverInterfaceFortran.hpp#L98). The consistent name would be `precicef_is_coupling_ongoing`.

MakisH:
This is now only a duplicate of #479, which is more complicated. Let's close this and address any other inconsistencies in separate issues, in the respective repositories.

PR #651, opened 2020-01-31
Add option to pass custom MPI communicator to Fortran bindings

MakisH:
In #382 we introduced (in the C++ API) the option to pass an MPI communicator. In #649 we extended this for C and tested it with the C solver dummy.

I implement the same for the Fortran bindings here, however this method is untested (and maybe wrong). I was not able to make it work with the Fortran solverdummy and I am unsure how to pass the communicator correctly through the differenet layers and languages.

I tried the following in the solver dummy (built with mpifort):
```fortran
Program main
    use mpi
   ! ... more declarations ...

   integer :: ierror, comm
   call MPI_INIT(ierror)
   call MPI_COMM_SIZE(MPI_COMM_WORLD, commsize, ierror)
   call MPI_COMM_RANK(MPI_COMM_WORLD, rank, ierror)

   ! ... more stuff

   call precicef_create_with_communicator(participantName, config, rank, commsize, MPI_COMM_WORLD)
```

This leads to the following error:

```
preCICE: No coupling scheme exists for the participant
preCICE: No coupling scheme exists for the participant
[Magdalena:9519] *** An error occurred in MPI_Comm_rank
[Magdalena:9519] *** reported by process [1268056065,0]
[Magdalena:9519] *** on communicator MPI_COMM_WORLD
[Magdalena:9519] *** MPI_ERR_COMM: invalid communicator
```

Notice here that in the C solver dummy we cannot directly pass `MPI_COMM_WORLD`, so there I defined another communicator, copied `MPI_COMM_WORLD` to that and passed its reference as `&comm`.

In Fortran, I also tried defining another integer as a copy and use that one instead of `MPI_COMM_WORLD`, but I got the same error. I still need deeper understanding of the situation.

Related to #649, #520, #382.

### DIFF
--- a/extras/bindings/c/include/precice/SolverInterfaceC.h
+++ b/extras/bindings/c/include/precice/SolverInterfaceC.h
@@ -33,6 +33,23 @@ void precicec_createSolverInterface(
     int         solverProcessIndex,
     int         solverProcessSize);
 
+/**
+ * @param[in] participantName Name of the participant using the interface. Has to
+ *        match the name given for a participant in the xml configuration file.
+ * @param[in] configurationFileName Name (with path) of the xml configuration file.
+ * @param[in] solverProcessIndex If the solver code runs with several processes,
+ *        each process using preCICE has to specify its index, which has to start
+ *        from 0 and end with solverProcessSize - 1.
+ * @param[in] solverProcessSize The number of solver processes using preCICE.
+ * @param[in] communicator A pointer to an MPI_Comm to use as MPI_COMM_WORLD.
+ */
+void precicec_createSolverInterface_withCommunicator(
+    const char *participantName,
+    const char *configFileName,
+    int         solverProcessIndex,
+    int         solverProcessSize,
+    void *      communicator);
+
 ///@}
 
 /// @name Steering Methods
--- a/extras/bindings/c/src/SolverInterfaceC.cpp
+++ b/extras/bindings/c/src/SolverInterfaceC.cpp
@@ -22,6 +22,22 @@ void precicec_createSolverInterface(
                                            solverProcessSize);
 }
 
+void precicec_createSolverInterface_withCommunicator(
+    const char *participantName,
+    const char *configFileName,
+    int         solverProcessIndex,
+    int         solverProcessSize,
+    void *      communicator)
+{
+  std::string stringAccessorName(participantName);
+  std::string stringConfigFileName(configFileName);
+  interface = new precice::SolverInterface(stringAccessorName,
+                                           stringConfigFileName,
+                                           solverProcessIndex,
+                                           solverProcessSize,
+                                           communicator);
+}
+
 double precicec_initialize()
 {
   PRECICE_ASSERT(interface != nullptr);
--- a/extras/bindings/fortran/include/precice/SolverInterfaceFortran.hpp
+++ b/extras/bindings/fortran/include/precice/SolverInterfaceFortran.hpp
@@ -35,6 +35,29 @@ void precicef_create_(
     int         lengthConfigFileName);
 
 /**
+ * @brief See precice::SolverInterface::SolverInterface() and ...::configure().
+ *
+ * Fortran syntax:
+ * precicef_create(
+ *   CHARACTER participantName(*),
+ *   CHARACTER configFileName(*),
+ *   INTEGER   solverProcessIndex,
+ *   INTEGER   solverProcessSize,
+ *   INTEGER   communicator )
+ *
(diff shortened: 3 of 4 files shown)


MakisH:
Apparently, there is an MPI method to convert communicators between Fortran and C: https://linux.die.net/man/3/mpi_comm_c2f

I will try if this helps.

fsimonis:
This could be an interesting signature for you from the [MPI-3 Standard](https://www.mpi-forum.org/docs/mpi-3.0/mpi30-report.pdf):


```cpp
 MPI_Comm_accept(const char *port_name, MPI_Info info, int root,MPI_Comm comm, MPI_Comm *newcomm)
```

```fortran
MPI_Comm_accept(port_name, info, root, comm, newcomm, ierror) BIND(C)
  CHARACTER(LEN=*), INTENT(IN) ::  port_name
  TYPE(MPI_Info), INTENT(IN) ::  info
  INTEGER, INTENT(IN) ::  root
  TYPE(MPI_Comm), INTENT(IN) ::  comm
  TYPE(MPI_Comm), INTENT(OUT) ::  newcomm
  INTEGER, OPTIONAL, INTENT(OUT) ::  ierror
```

fsimonis:
This issue describes this problem quite well. Maybe this could help to solve this? https://stackoverflow.com/q/42530620/11633170

MakisH:
This is completely outdated and nobody is looking at it at the moment. In order to close this PR and delete the branch, but potentially touch this again in the future, here is the gist of this PR.

Header of the C bindings:

```cpp
/**
 * @param[in] participantName Name of the participant using the interface. Has to
 *        match the name given for a participant in the xml configuration file.
 * @param[in] configurationFileName Name (with path) of the xml configuration file.
 * @param[in] solverProcessIndex If the solver code runs with several processes,
 *        each process using preCICE has to specify its index, which has to start
 *        from 0 and end with solverProcessSize - 1.
 * @param[in] solverProcessSize The number of solver processes using preCICE.
 * @param[in] communicator A pointer to an MPI_Comm to use as MPI_COMM_WORLD.
 */
void precicec_createSolverInterface_withCommunicator(
    const char *participantName,
    const char *configFileName,
    int         solverProcessIndex,
    int         solverProcessSize,
    void *      communicator);
```

Implementation of the C bindings:

```cpp
void precicec_createSolverInterface_withCommunicator(
    const char *participantName,
    const char *configFileName,
    int         solverProcessIndex,
    int         solverProcessSize,
    void *      communicator)
{
  interface = new precice::SolverInterface(stringAccessorName,
                                           stringConfigFileName,
                                           solverProcessIndex,
                                           solverProcessSize,
                                           communicator);
}
```

Fortran bindings header:

```cpp
/**
 * @brief See precice::SolverInterface::SolverInterface() and ...::configure().
 *
 * Fortran syntax:
 * precicef_create(
 *   CHARACTER participantName(*),
 *   CHARACTER configFileName(*),
 *   INTEGER   solverProcessIndex,
 *   INTEGER   solverProcessSize,
 *   INTEGER   communicator )
 *
 * IN:  participantName, configFileName, solverProcessIndex, solverProcessSize, communicator
 * OUT: -
 */
void precicef_create_with_communicator_(
    const char *participantName,
    const char *configFileName,
    const int * solverProcessIndex,
    const int * solverProcessSize,
    int       * communicator,
    int         lengthAccessorName,
    int         lengthConfigFileName);
```

Implementation of the Fortran bindings:

```cpp
void precicef_create_with_communicator_(
    const char *participantName,
    const char *configFileName,
    const int * solverProcessIndex,
    const int * solverProcessSize,
    int       * communicator,
    int         lengthAccessorName,
    int         lengthConfigFileName)
{
  //cout << "lengthAccessorName: " << lengthAccessorName << '\n';
  //cout << "lengthConfigFileName: " << lengthConfigFileName << '\n';
  //cout << "solverProcessIndex: " << *solverProcessIndex << '\n';
  //cout << "solverProcessSize: " << *solverProcessSize << '\n';
  //cout << "communicator: " << *communicator << '\n';
  int    strippedLength = precice::impl::strippedLength(participantName, lengthAccessorName);
  string stringAccessorName(participantName, strippedLength);
  strippedLength = precice::impl::strippedLength(configFileName, lengthConfigFileName);
  string stringConfigFileName(configFileName, strippedLength);
  //cout << "Accessor: " << stringAccessorName << "!" << '\n';
  //cout << "Config  : " << stringConfigFileName << "!" << '\n';
  impl = new precice::SolverInterface(stringAccessorName,
                                      stringConfigFileName,
                                      *solverProcessIndex, *solverProcessSize, &communicator);
}
```

PR #654, opened 2020-02-01
Update documentation and tools for cmake variables

MakisH:
Related to https://github.com/precice/precice/pull/609

@fsimonis I auto-replaced `-Dold_variable=` to `-Dnew_variable=`. Could you please check if we need to update it in more places?

### DIFF
--- a/docs/documents/070_tooling.dox
+++ b/docs/documents/070_tooling.dox
@@ -84,9 +84,9 @@ $ cmake \
    -DCMAKE_CXX_CPPCHECK="cppcheck;--enable=all" \
    -DCMAKE_CXX_CPPLINT="cpplint.py" \
    -DCMAKE_CXX_INCLUDE_WHAT_YOU_USE="path/to/iwyu;-p;." \
-   -DPYTHON=ON \
-   -DMPI=ON \
-   -DPETSC=ON \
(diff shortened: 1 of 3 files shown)


docs/documents/070_tooling.dox
```diff
@@ -84,9 +84,9 @@ $ cmake \
    -DCMAKE_CXX_CPPCHECK="cppcheck;--enable=all" \
    -DCMAKE_CXX_CPPLINT="cpplint.py" \
    -DCMAKE_CXX_INCLUDE_WHAT_YOU_USE="path/to/iwyu;-p;." \
-   -DPYTHON=ON \
-   -DMPI=ON \
-   -DPETSC=ON \
+   -DPRECICE_PythonActions=ON \
+   -DPRECICE_MPICommunicator=ON \
```
fsimonis: The MPI Variable is called `PRECICE_MPICommunication`, not `PRECICE_MPICommunicator`!

MakisH: I just took it from [the PR](https://github.com/precice/precice/pull/609), but good point!

MakisH: Fixed

Issue #639, opened 2020-01-29
Rename timestep to time window


uekerman:
Leftover TODO from down the rabbit hole of #619 
Update all versions of "timestep" in preCICE, e.g. `timestepsLeft`, `maxTimesteps` and `timesteps` in `BaseCouplingScheme`.

uekerman:
Are there still things to do after #689 ?

BenjaminRueth:
I hope not, but I think yes. We should do a final check after merging #689, since otherwise are probably missing something.

BenjaminRodenberg:
I did `grep -r "timestep" *` (fc9e02ce92d88132d36681cd40b98f46c2d86cf9) and got a list of TODOS for this issue. I already kicked out the cases where "timestep" is correct. I will take care of the remaining TODOS on [branch `i_639`](https://github.com/precice/precice/tree/i_639):

```
src/acceleration/BroydenAcceleration.cpp:    int                     timestepsReused,
src/acceleration/BroydenAcceleration.cpp:    : BaseQNAcceleration(initialRelaxation, forceInitialRelaxation, maxIterationsUsed, timestepsReused,
src/acceleration/MVQNAcceleration.hpp:      int                     timestepsReused,
src/acceleration/IQNILSAcceleration.hpp:      int                     timestepsReused,
src/acceleration/test/PreconditionerTest.cpp:  //should change, since first timestep
src/acceleration/test/AccelerationMasterSlaveTest.cpp:  int              timestepsReused          = 6;
src/acceleration/test/AccelerationMasterSlaveTest.cpp:                        timestepsReused, filter, singularityLimit, dataIDs, prec);
src/acceleration/test/AccelerationMasterSlaveTest.cpp:  int    timestepsReused          = 6;
src/acceleration/test/AccelerationMasterSlaveTest.cpp:                      timestepsReused, filter, singularityLimit, dataIDs, prec, alwaysBuildJacobian,
src/acceleration/test/AccelerationMasterSlaveTest.cpp:  int    timestepsReused          = 0;
src/acceleration/test/AccelerationMasterSlaveTest.cpp:                      timestepsReused, filter, singularityLimit, dataIDs, _preconditioner, alwaysBuildJacobian,
src/acceleration/test/AccelerationMasterSlaveTest.cpp:  int              timestepsReused          = 1;
src/acceleration/test/AccelerationMasterSlaveTest.cpp:                         timestepsReused, filter, singularityLimit, dataIDs, prec);
[...]

PR #656, opened 2020-02-03
Adapt its and conv files to time window naming scheme

uekerman:
We forgot to now also change `Timestep` to `TimeWindow` in the convergence and iterations file. I further added two cases where we still used singular "iteration" instead of plural "iterations". Makes only sense in the convergence file, not in the iterations file.

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -573,7 +573,7 @@ bool BaseCouplingScheme::measureConvergence(
   bool oneSuffices  = false;
   PRECICE_ASSERT(_convergenceMeasures.size() > 0);
   if (not utils::MasterSlave::isSlave()) {
-    _convergenceWriter->writeData("Timestep", _timesteps);
+    _convergenceWriter->writeData("TimeWindow", _timesteps);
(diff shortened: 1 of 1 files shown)


src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -573,7 +573,7 @@ bool BaseCouplingScheme::measureConvergence(
   bool oneSuffices  = false;
   PRECICE_ASSERT(_convergenceMeasures.size() > 0);
   if (not utils::MasterSlave::isSlave()) {
-    _convergenceWriter->writeData("Timestep", _timesteps);
+    _convergenceWriter->writeData("TimeWindow", _timesteps);
```
KyleDavisSA: There is still the issue of `_timesteps`. Going down that rabbit hole?

MakisH:
We then also need to adapt the system tests reference files.

src/cplscheme/BaseCouplingScheme.cpp
uekerman: That part, we postponed. All internal variables are still named "timestep".

uekerman: https://github.com/precice/precice/issues/639

PR #658, opened 2020-02-03
Relax QN integration test to also cope with different Eigen versions

uekerman:
Such that tests also run in spack with different Eigen versions. The design of this test is not overly nice as it is so sensitive, but still very valuable at the moment as we don't test QN over several iterations in any other integration test.

@fsimonis Please merge if the tests now work in your setup.


fsimonis:
Works now on my system.

PR #657, opened 2020-02-03
Release v2.0.0

fsimonis:
## How to work with this template

* [x] assign a release manager, who takes care of the process @fsimonis 
* [ ] assign each point below to a responsible person, before you continue. Use `@member`.

Only the release manager should update this post (even tickboxes, due to race conditions in editing). Everybody else should comment on the PR with the progress.

## Step by step guide
* [x] Look over [`CHANGELOG.md`](https://github.com/precice/precice/blob/develop/CHANGELOG.md) (@fsimonis )
   * Add things, if necessary
   * Extract summary
   * Fix wording
   * Sort the entries lexicographically
* [x] Look over the [Roadmap](https://github.com/precice/precice/wiki/Roadmap) and update entries.
* [x] Merge master to develop (No commits after the release on master)
* [x] Check code base w.r.t code formatting (run [`precice/tools/formatting/check-format`](https://github.com/precice/precice/blob/develop/tools/formatting/check-format)) and reformat if required (run [`precice/tools/formatting/format-all`](https://github.com/precice/precice/blob/develop/tools/formatting/format-all))
* [x] Create branch `release-N` from develop. If needed, `git rebase develop`.
* [x] Open PR from `release-N` to `master` (use [this template](https://github.com/precice/precice/blob/add_PR_template/.github/PULL_REQUEST_TEMPLATE/release_pull_request_template.md))
* [x] Do regression tests using the release branch (specific revision) _list below :arrow_down:_ (all)
* [x] Fix potential problems in develop (all)
* [x] Run `tools/releasing/bumpversion.sh MAJOR.MINOR.PATCH` to bump the version @fsimonis 
* [x] Verify the version changes in: @fsimonis 
   * [x] [CHANGELOG](https://github.com/precice/precice/blob/develop/CHANGELOG.md)
   * [x] [CMakeLists.txt](https://github.com/precice/precice/blob/develop/CMakeLists.txt)
   * [x] [debian changelog](https://github.com/precice/precice/blob/develop/tools/releasing/packaging/debian/changelog)
   * [ ] (do independent release, if necessary!) [Python bindings](https://github.com/precice/python-bindings)
   * [ ] (do independent release, if necessary!) [MATLAB bindings](https://github.com/precice/matlab-bindings)
* [x] Draft message to mailing list @uekerman 
* [x] Update documentation (all)
  * [ ] Update markdown configuration reference in wiki
* [x] Approve the PR with at least two reviews (all)
* [ ] Merge PR to master @fsimonis 
[...]

PR #659, opened 2020-02-03
Update PR template

BenjaminRueth:
Applying changes according to latest version of #657


fsimonis:
Further feedback:
* Order: formatting, release branch, bumpversion, verify versions, release PR, regression tests , fix problems, 
* Add point "external releases" to post-release.
* Move python and matlab releases to post-release/external-releases
* Maybe add spack recepie to regression tests as it tends to reveal problems in the build system

fsimonis:
Another thought is that maybe the PR is not the right place for keeping track of all these steps.
The PR only exists in a subinterval of the actual release workflow, so this template should maybe be an issue template.

BenjaminRueth:
@fsimonis Can you do a review?

One thing I was wondering about while merging: What has happened to the [packages for Ubuntu](https://github.com/precice/precice/blob/885d0de534d2ee35c4d772305e8eba8793aba238/.github/PULL_REQUEST_TEMPLATE/release_pull_request_template.md#post-release)? Do we have to add this to the template or do we already have everything that's required?

Feel free to merge, if your review does not discover any further points.

fsimonis:
@BenjaminRueth You mean this? :wink: 
image

PR #664, opened 2020-02-05
Give details for develop and master branch

BenjaminRueth:
The project status part in our `README.md` file often looks much worse than it acutally is:

Screenshot from 2020-02-05 10-16-32

Often the system tests are shown as failing. However, this is often only the case for the develop branch and for master everything should be fine. But the way we are presenting the project status currently suggests that our tests are failing for the actual release.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -23,12 +23,12 @@
 </a>
 
 ``precice/master``  
-[![MasterBuild](https://travis-ci.org/precice/precice.svg?branch=master)](https://travis-ci.org/precice/precice)
-[![MasterSystemtests](https://img.shields.io/travis/precice/systemtests/master.svg?label=system%20tests&style=flat)](https://travis-ci.org/precice/systemtests)
+[![MasterBuild](https://travis-ci.org/precice/precice.svg?branch=master)](https://travis-ci.org/precice/precice/branches)
+[![MasterSystemtests](https://img.shields.io/travis/precice/systemtests/master.svg?label=system%20tests&style=flat)](https://travis-ci.org/precice/systemtests/branches)
 
 ``precice/develop``  
-[![DevelopBuild](https://travis-ci.org/precice/precice.svg?branch=develop)](https://travis-ci.org/precice/precice)
-[![DevelopSystemtests](https://img.shields.io/travis/precice/systemtests/develop.svg?label=system%20tests&style=flat)](https://travis-ci.org/precice/systemtests)
+[![DevelopBuild](https://travis-ci.org/precice/precice.svg?branch=develop)](https://travis-ci.org/precice/precice/branches)
+[![DevelopSystemtests](https://img.shields.io/travis/precice/systemtests/develop.svg?label=system%20tests&style=flat)](https://travis-ci.org/precice/systemtests/branches)
 
 **Code Quality**  
 [![CodeFactor](https://www.codefactor.io/repository/github/precice/precice/badge)](https://www.codefactor.io/repository/github/precice/precice)
--- a/docs/README.md
+++ b/docs/README.md
@@ -25,13 +25,13 @@
 ``master`` branch: <a style="text-decoration: none" href="https://travis-ci.org/precice/precice" target="_blank">    
     <img src="https://travis-ci.org/precice/precice.svg?branch=master" alt="Build status">
 </a> <a style="text-decoration: none" href="https://travis-ci.org/precice/systemtests" target="_blank">    
-    <img src="https://travis-ci.org/precice/systemtests.svg?branch=master" alt="Build status">
+    <img src="https://img.shields.io/travis/precice/systemtests.svg?label=system%20tests&style=flat&branch=master" alt="Build status">
 </a>
 
 ``develop`` branch: <a style="text-decoration: none" href="https://travis-ci.org/precice/precice" target="_blank">    
     <img src="https://travis-ci.org/precice/precice.svg?branch=develop" alt="Build status">
 </a> <a style="text-decoration: none" href="https://travis-ci.org/precice/systemtests" target="_blank">    
-    <img src="https://travis-ci.org/precice/systemtests.svg?branch=develop" alt="Build status">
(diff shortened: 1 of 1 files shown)


BenjaminRueth:
Note that currently, also the master branch is indicated as "failing":

---

Screenshot from 2020-02-05 10-38-18

---

I guess that this is a result of triggering mechanism that we use. The triggered tests run on the master branch of the systemtests (@Eder-K?):

---

Screenshot from 2020-02-05 10-31-43

---

I'm not sure whether this is a bug or a feature. This is something that we should also revisit. But I assume this is an issue rather on the `precice/systemtests` than on the `precice/precice` side.

uekerman:
Can we call the second badge `systemtests` again?

BenjaminRueth:
> Can we call the second badge `systemtests` again?

Done :+1:

BenjaminRueth:
I see one more problem:

Travis does not offer a direct link to a branch. Therefore, [MasterSystemtests](https://travis-ci.org/precice/systemtests) and [DevelopSystemtests](https://travis-ci.org/precice/systemtests) both point to the same location: The Travis page that shows the status for the build that was triggered last. This is slightly misleading. I see three possibilities:

1. Just accept this and provide the misleading link: [MasterSystemtests](https://travis-ci.org/precice/systemtests) misleading
2. Point users to the page with the branches instead: [MasterSystemtests](https://travis-ci.org/precice/systemtests/branches) tedious to find the right branch
3. Don't provide a link, but just the picture of the badge: MasterSystemtests at least not wrong...

Any opinions on this? Do you have a different idea? If I overlooked the Travis API that gives access to a branch, then I would be very grateful for somebody pointing me to the right link.

MakisH:
>     1. Just accept this and provide the misleading link: [MasterSystemtests](https://travis-ci.org/precice/systemtests) misleading

This is what we do now, isn't it? I think this is the correct way and we just need to fix it.

>     2. Point users to the page with the branches instead: [MasterSystemtests](https://travis-ci.org/precice/systemtests/branches) tedious to find the right branch

This badge will not give time-consistent information, right?

>     3. Don't provide a link, but just the picture of the badge: MasterSystemtests at least not wrong...

I would definitely not do this. It can be misleading (to false positive) / bad scientific practice and it would not be consistent between master and develop.

Let's investigate on why this happens (maybe move this to the systemtests repository). We can also just remove the badge for the system tests, if it is unreliable. We already have a Travis badge there and the important is that **we** know the detailed status at any time.

uekerman:
I want to add that I was really not knowing that the badge in `precice/precice`, also the `build` badge was not pointing to the status of the corresponding branch.

> We already have a Travis badge there and the important is that **we** know the detailed status at any time.

I still don't understand which status of the system tests belongs to which branch of preCICE. :) 

In an ideal world (that's the world I thought we are living in so far), the `build` badge (wouldn't "tests" be a better name?) and the `systemtests` badge should show and link to the status of the corresponding branch. So if you look at the master README, both should point to the status of the master branch. If you look at the develop branch. both should point to the status of the develop branch. For `systemtests`, as "other packages" the respective branches of the adapters and tutorials should be chosen.

But, if I understand @BenjaminRueth correctly sth like this is not possible with Travis.

BenjaminRueth:
> But, if I understand @BenjaminRueth correctly sth like this is not possible with Travis.

Exactly. Currently there are several users requesting this feature: https://travis-ci.community/t/landing-page-for-specific-branch/6710, https://github.com/travis-ci/travis-ci/issues/5024.

MakisH:
Then maybe let's remove this badge for now? We could also make our own badge that is e.g. blue and just links to a general Travis page.

> I still don't understand which status of the system tests belongs to which branch of preCICE. :)

This is something we can improve separately, and @Eder-K has a nice solution already in https://github.com/precice/systemtests/pull/147.

uekerman:
> Then maybe let's remove this badge for now?

That would mean both badges, also the one for the preCICE tests as it is also misleading currently. Not as much as the system tests, but it is still wrong. 
Better then removing would be to just delete the link?
How do other projects using Travis handle this?

MakisH:
Just for clarification: Our badge is currently set up like this, in all branches:

```
<a style="text-decoration: none" href="https://travis-ci.org/precice/precice" target="_blank">
    <img src="https://travis-ci.org/precice/precice.svg?branch=develop" alt="Build status">
</a>
```

This means:
- The status (image) is always the build status of the latest `develop` state (unless something is wrong)
- The link target ("landing page") is always `https://travis-ci.org/precice/precice`. We could easily change it to `https://travis-ci.org/precice/precice/branches`.

BenjaminRueth:
@MakisH 

> 
> This means:
> 
>     * The status (image) is always the build status of the latest `develop` state (unless something is wrong)
What do you exactly mean with the status (image)? Can you provide a screenshot? I get the following:

image

Here, the image for master shows the status of master and the image for develop shows the status for develop. Probably, I am misunderstanding something...
> 
>     * The link target ("landing page") is always `https://travis-ci.org/precice/precice`. We could easily change it to `https://travis-ci.org/precice/precice/branches`.

I added this point in a79b6ad. Linking to the branches page makes everything a bit less confusing in my opinion.

BenjaminRueth:
I will change this PR to a "draft", since I currently do not really see a good solution with a clear improvement comparing to the current state. If we move to github actions (#713) at some point in time, there might be a good solution available.

MakisH:
> What do you exactly mean with the status (image)? Can you provide a screenshot? I get the following:

@BenjaminRueth we always ask for the build status of develop, see the source of the image: `https://travis-ci.org/precice/precice.svg?branch=develop`

MakisH:
How about we simply make this a generic blue "systemtests: check"? I think that with Travis and our current structure, we will not be able to get a consistent and useful badge soon. Most of the times it will be "failing" and, especially in the master branch, this would be misleading. Besides, the README should not be the place that _we_ look to get the status, but the users.

<a style="text-decoration: none" href="https://travis-ci.org/github/precice/systemtests/builds" target="_blank">
    Build status
</a>

#794 implements this approach and closes this PR.

BenjaminRueth:
This PR can now be closes. #794 provides a solution for the problem with the systemtests and #798 describes what's still missing.

PR #690, opened 2020-03-05
Initial implementation of split patches

KyleDavisSA:
This PR aims to implement the changes required for the split patching issue from https://github.com/precice/precice/issues/374. The initial idea is to add a `patchID` variable for each vertex. During the mapping, a vertex on mesh A is only used in the mapping of a vertex in mesh B, if it has the same patch ID of a vertex on the corresponding mesh in the other solver. 

As an example, consider the yellow, blue and red surfaces in https://github.com/precice/precice/issues/374, both solvers need the meshes split into the patches accordingly. Each colour will receive a patch ID of 1, 2 and 3 respectively. For a vertex in the yellow patch, with a patchID of 1, it can only be mapped to a point on the yellow patch of the other solver (also with a patch ID of 1). Any vertex that is not assigned a patchID automatically has a value of 0, making it fair game for any other vertex with a partchID of 0. Therefore if no patchID is set, then mapping functions as normal. 

The adapters would need to be changed to allow multiple surfaces to make up a mesh, but use each surface to define the patch. That is not the purpose of this PR, but only to implement how to use the patchID. 

Note that this does not create a meshing for each patch, but only influences which vertices are used (i.e. still one mesh with one RBF matrix). it has also only been added to the `tree-based preallocation for matrix C` of `PetRadialBasisFctMapping.hpp`. The RBF mapping functions as normal by getting the distance between 2 vertices

```
distance = inVertex.getCoords() - vj.getCoords();
double const norm = distance.norm();
```
For any compact support RBF, this norm must be within the support radius
```
if (supportRadius > norm or col == global_row) {
```
The patchID can then be checked to see if the vertices on both meshes belong to the same patch ID
```
if (inVertex.getPatchID() == vj.getPatchID()) {
```
This is probably not the most efficient implementation, and the patchID check can probably be moved up to avoid doing unnecessary computations, such as determining the distance between vertices.

# Open Todos

- [x] Implement function to set the patchID of a vertex
- [ ] Implement setting block variant `setMeshVerticesPatch`
- [ ] Implement `getPatchID(meshID, patchName)` that can be used with patch names in the config
- [ ] Add patch functions for C
- [ ] Add patch functions for python
- [ ] Add patch functions for fortran
- [ ] Integration tests

### DIFF
--- a/src/mapping/PetRadialBasisFctMapping.hpp
+++ b/src/mapping/PetRadialBasisFctMapping.hpp
@@ -71,6 +71,7 @@ public:
       bool                           xDead,
       bool                           yDead,
       bool                           zDead,
+      int                            fromPatchID,
       double                         solverRtol    = 1e-9,
       Polynomial                     polynomial    = Polynomial::SEPARATE,
       Preallocation                  preallocation = Preallocation::TREE);
@@ -193,6 +194,7 @@ PetRadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::PetRadialBasisFctMapping(
     bool                           xDead,
     bool                           yDead,
     bool                           zDead,
+    int                            fromPatchID,
     double                         solverRtol,
     Polynomial                     polynomial,
     Preallocation                  preallocation)
@@ -824,7 +826,8 @@ void PetRadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::tagMeshFirstRound()
       preCICE is fully functional, but performance for large cases is degraded."
 #endif
   ) {
-    auto rtree    = mesh::rtree::getVertexRTree(filterMesh);
+    int z = 0;
+    auto rtree    = mesh::rtree::getVertexRTree(filterMesh,z);
     namespace bgi = boost::geometry::index;
     auto bb       = otherMesh->getBoundingBox();
     // Enlarge by support radius
@@ -881,7 +884,8 @@ void PetRadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::tagMeshSecondRound()
     bb[d].first -= _basisFunction.getSupportRadius();
     bb[d].second += _basisFunction.getSupportRadius();
   }
-  auto rtree = mesh::rtree::getVertexRTree(mesh);
+  int z = 0;
+  auto rtree = mesh::rtree::getVertexRTree(mesh,z);
   rtree->query(boost::geometry::index::within(bb),
                boost::make_function_output_iterator([&mesh](size_t idx) {
                  mesh->vertices()[idx].tag();
@@ -1346,7 +1350,8 @@ PetRadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::bgPreallocationMatrixC(mesh::
   std::tie(n, std::ignore) = _matrixC.getLocalSize();
   std::vector<PetscInt> d_nnz(n), o_nnz(n);
 
-  auto tree = mesh::rtree::getVertexRTree(inMesh);
+  int z = 0;
+  auto tree = mesh::rtree::getVertexRTree(inMesh,z);
 
   double const supportRadius = _basisFunction.getSupportRadius();
 
@@ -1444,7 +1449,8 @@ PetRadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::bgPreallocationMatrixA(mesh::
 
   PetscInt       ownerRangeABegin, ownerRangeAEnd, colOwnerRangeABegin, colOwnerRangeAEnd;
   PetscInt const outputSize    = _matrixA.getLocalSize().first;
-  auto           tree          = mesh::rtree::getVertexRTree(inMesh);
+  int z = 0;
+  auto           tree          = mesh::rtree::getVertexRTree(inMesh,z);
   double const   supportRadius = _basisFunction.getSupportRadius();
 
   std::tie(ownerRangeABegin, ownerRangeAEnd)       = _matrixA.ownerRange();
--- a/src/mapping/PetRadialBasisFctMapping.hpp
+++ b/src/mapping/PetRadialBasisFctMapping.hpp
@@ -368,9 +368,6 @@ void PetRadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::computeMapping()
     if (not inVertex.isOwner())
       continue;
 
-    // Need the patch ID for the vertex of interest here
-    // int inVertexPatchID = inVertex.getPatchID();
-
     PetscInt colNum = 0; // holds the number of non-zero columns in current row
 
     // -- SETS THE POLYNOMIAL PART OF THE MATRIX --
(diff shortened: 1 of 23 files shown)


src/precice/impl/SolverInterfaceImpl.hpp
```diff
@@ -237,14 +237,24 @@ class SolverInterfaceImpl {
   /**
    * @brief Sets several spatial positions for a mesh.
    *
-   * @param[out] ids IDs for data from given positions.
+   * @param[out] ids IDs fo r data from given positions.
```
uekerman: That's probably not intended.

src/mesh/Vertex.hpp
```diff
@@ -62,10 +68,15 @@ class Vertex {
...
   inline bool operator!=(const Vertex &rhs) const;
 
+  
+
 private:
   /// Unique (among vertices in one mesh) ID of the vertex.
   int _id;
 
+  /// Patch id that a vertex belongs to
+  int _patchid = 0;
```
uekerman:
```suggestion
  int _patchID = 0;
```
would be more consistent?

src/mapping/PetRadialBasisFctMapping.hpp
```diff
@@ -1355,13 +1357,20 @@ PetRadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::bgPreallocationMatrixC(mesh::
           distance[d] = 0;
 
       double const norm = distance.norm();
+      if (inVertex.getPatchID() == vj.getPatchID()){
+        PRECICE_INFO("ID numbers of vertices: inVertexID = " << inVertex.getID() << " and vjID = " << vj.getID() );
+        PRECICE_INFO("Patch ID matches: inVertexPatchID = " << inVertex.getPatchID() << " and vjPatchID = " << vj.getPatchID() );
```
uekerman: These should probably be `Debug` statements, but we probably also don't want all of this in the end.

src/mapping/PetRadialBasisFctMapping.hpp
```diff
@@ -1355,13 +1357,20 @@ PetRadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::bgPreallocationMatrixC(mesh::
...
       if (supportRadius > norm or col == global_row) {
         vertexData[local_row - localPolyparams].emplace_back(vj.getGlobalIndex() + polyparams, norm);
         if (mappedCol >= colOwnerRangeCBegin and mappedCol < colOwnerRangeCEnd)
           d_nnz[local_row]++;
         else
           o_nnz[local_row]++;
       }
+      } else{
+        PRECICE_INFO("ID numbers of vertices: inVertexID = " << inVertex.getID() << " and vjID = " << vj.getID() );
+        PRECICE_INFO("Skip due to patch ID difference: inVertexPatchID = " << inVertex.getPatchID() << " and vjPatchID = " << vj.getPatchID() );
```
uekerman: Probably not needed.

src/mapping/PetRadialBasisFctMapping.hpp
KyleDavisSA: Agreed. I put them in when checking if the patches were selected correctly. Same for the comment below.

KyleDavisSA:
I had thought about how to possibly implement this with NN mapping, but was not too sure how to go about that. My thought is to include a condition in the boost query. At the moment, the coordinates of a point are given and boost queries which coordinate in `input()->vertices()[val];` is closest.

https://github.com/precice/precice/blob/78b266f3165adafc7a9f97b40ee0c2b55b6fca58/src/mapping/NearestNeighborMapping.cpp#L47-L52

In the RBF mapping, there is also a `index::satisfies` condition as well to check if the point is within a specified radius or not:

https://github.com/precice/precice/blob/78b266f3165adafc7a9f97b40ee0c2b55b6fca58/src/mapping/PetRadialBasisFctMapping.hpp#L1344

The idea is to then include the patchid as a condition here for NN mapping with something like

```
const mesh::Mesh::VertexContainer &    inputVertices = input()->vertices();

const Eigen::VectorXd &coords = outputVertices[i].getCoords();
int outputPatchID = outputVertices[i].getPatchID();

rtree->query(boost::geometry::index::nearest(coords, 1) && boost::geometry::index::satisfies([&](size_t const i) {return outputPatchID == inputVertices[i].getPatchID()}),
```
Question is, how can I get boost to check through all the patchID's and not coordinates of the vertices? This seems ugly that it creates a new `VertexContainer`, and I am not sure if this will even loop through all the patchid`s

KyleDavisSA:
As a check, I have run:

```
rtree->query(boost::geometry::index::nearest(coords, 1) and boost::geometry::index::satisfies([&](size_t const i) { return (outputPatchID - input()->vertices()[i].getPatchID()) < 1 ;}),
```

which should enforce that the patchID's must also be equal (considering they are integers). Small tests have shown that the `boost::geometry::index::satisfies` will give the results we are looking for.

The value of `input()->vertices()[i].getPatchID()` is always zero. Does anyone know how to access these values here? I am not too sure what is going on here, and struggling to figure it out with boost documentation.

uekerman:
Why not creating a mapping per patch? I think this would be simpler overall. The only prerequisite would be that we need to know how many patches there are at configuration. But in my opinion this would be a good thing anyway.

Mappings are created here:
https://github.com/precice/precice/blob/develop/src/mapping/config/MappingConfiguration.cpp#L211

KyleDavisSA:
Currently, a patchID number is given to each vertex, i.e. every vertex has a number associated to which patch it may belong to, with a default of 0 if no patches are specified. If no patches are specified, then all vertices on a mesh belong to the same patch. But if a vertex belongs to a specific patch (i.e. ID = 0, or ID = 1 etc.), how should this patch information be passed to preCICE? Should the patch names be an input in the `precice-config-xml` or in the solvers config file? Assuming each solver has some kind of config file that can do this?

As an example, CalculiX has a `config.yml` file that contains
```
        interfaces:
        - nodes-mesh: Calculix_Mesh
          patch: interface
```

This can be modified to include all patches associated to the mesh. This can be given to preCICE through a call `setVertexPatchName(meshID, vertexID, patchName)`. But if some kind of config isn't available from the solver, then this can't work. An alternative is for the preCICE config file to have this information. An initial idea is to add a tag allowing to specify if a mesh is divided into patches, and if so, what are the patchNames with something like:

```
<mesh name="Fluid-Mesh-Nodes">
        <use-data name="DisplacementDeltas0"/>
	<use-patch name="patchSet1"/>
      </mesh>

<patch-set name="patchSet1">
	<patch-name name="patch1"/>
	<patch-name name="patch2"/>
      </patch-set>
```
With this information, the total number of patches per mesh can be determined, and a mapping per patch can be created. If no `use-patch` is specified for a mesh, then the default is just 1 patch and mapping essentially functions as it is now. There will still be a call `setVertexPatchName(meshID, vertexID, patchName)` so that a vertex has a patchName associated with it. Is it better to have this information in the `precice-config.xml` file? As some things such as the patch names will have to be repeated in solver config files.

uekerman:
Having this information in the preCICE config is a safeguard and, therefore, redundant.
It is consistent with how we treat meshes and data fields already. Basically, we can use it to easily check that both participants define the same patches and that patches are define at all, etc. In any case, we also need this information in the adapter configs.

In the preCICE config, it could be even simpler:
```xml
<mesh name="Fluid-Mesh-Nodes">
   <use-data name="DisplacementDeltas0"/>
   <use-patch name="Front"/>
   <use-patch name="Back"/>
   <use-patch name="Tip"/>
</mesh>
```

fsimonis:
Closing as the diff to develop has grown too much.

PR #697, opened 2020-03-18
Adds nearest projection mapping for quad meshes

### LINKED ISSUE #153: Implement and test Quads
Right now, Quads (hexahedrals) are implemented, but not throughout preCICE and not really tested.

KyleDavisSA:
This PR is to provide a method for preCICE to handle quad meshes for nearest projection mapping. The problem is described in https://github.com/precice/precice/issues/153. This PR implements option 2.1 in https://github.com/precice/precice/issues/153, however this is not entirely unique. of the points, connectivity is assumed to be 0 -> 1 -> 2 -> 3 -> 0. 
```
  2 ------- 3
  |         |
  |         |
  |         |
  1 ------- 0
```

This PR splits this mesh into two triangles:

* 0 -> 1 -> 2 -> 0
* 2 -> 3 -> 0 -> 2 

Option 2.2 requires creating a fifth point. However this point has no data assciated with it from the solver, and therefore requires some kind of interpolation in a solver as well. Therefore, the best start is to split the quad into 2 triangles only.

# Open Todos

- [x] Project vertices in 3D coordinates onto a 2D plane
- [x] Check if all four vertices lie in plane
- [x] Check convexity of quad from 4 vertices
- [x] Check ordering of edges
- [x] Check convexity of quad from edges
- [x] Add diagonal for `setMeshQuadWithEdges`
- [x] Add diagonal for `setMeshQuad` 

Resolves #153


src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -781,6 +781,17 @@ void SolverInterfaceImpl::setMeshQuad(
...
+// Create two triangles with the four points
+/*
+  2 ------- 3
+    |\    |
+    | \   |
+    |  \  |
+    |   \ |
+    |    \|
+  1 ------- 0
+*/
```
uekerman: That's an unusual place for a comment. Either place it before the definition in the header file or inside the body of the function.

KyleDavisSA:
Should I change `setMeshQuad` as well, or actually create another function `setMeshQuadWithTriangles` and leave all other quad functions the same? I ran a test with calculix and it is working, so Ill add that soon.

uekerman:
I would vote for adjusting both functions. No new function. Quads are anyway not working currently, so why keeping those nop functions. 
Great that CalculiX already works! How did you check that exactly? A convergence study or looking at the weights? We need to be sure that the NP does not fall back to some random first order method. 
Anyway, we should implement integration tests, a serial and a parallel at least 😬

KyleDavisSA:
Agreed on the integration tests. From CalculiX side, I was printing the values of the elements, faceID (in calculix) and node coordinate values, comparing that to the preciceNodeID's when setting mesh vertices, and then comparing to the nodeID's given to preCICE for the mapping. I do not know yet if preCICE uses those values properly.

fsimonis:
@KyleDavisSA we don't require an ordering of edges or vertices in the interface, thus we cannot make assumptions on them internally and have to reorder them manually. You can check the `setMeshVertex` code for some inspiration on that.

As the triangularisation of the quad is our choice, we should not leak this implementation detail out of the interface. Instead we should leave the interface untouched.

I propose:
1. take 4 edges
2. order them and check for correctness
3. internally create a diagonal edge (maybe use an existing one)
4. create a quad

A note on point 3: we could split the longer diagonal to prevent degenerative meshes.

KyleDavisSA:
Thinking of how to implement this for both `setMeshQuad` and `setMeshQuadWithEdges` to not repeat any unnecessary work. `setMeshQuad` should be easy to check by reordering the edges, checking diagonal distances and then creating the new edge. However when only vertices are given, the quad has to be built essentially (if we can't assume an ordering). 

I think this can be done with gift wrapping algorithm [here](https://en.wikipedia.org/wiki/Gift_wrapping_algorithm). This can firstly build the quad (give us an ordered list of vertices (0 - 1 - 2 - 3 - 0), and also check for convex-ness (if number of points making up the convex hull is less than 4, then the quad is not valid). This can then be checked in `setMeshQuad` too. Complexity is probably O(numberOfQuads * n^2) where n is 4 points per quad.

But what if an adapter is created to give quad information in correct order? Should this be skipped or always done?

KyleDavisSA:
Another point on this, if the case is 3D, each vertex has a coordinate in the x, y and z direction. Quads are inherently 2D, so these 3D vertices need to be projected onto a 2D plane. Therefore each point has a x' and y' coordinate. These new coordinates are used to evaluate the quad. Is there already similar functionality inside or preCICE? Or can someone help steer this in the right direction?

src/mesh/Mesh.hpp
```diff
@@ -191,6 +191,13 @@ class Mesh {
   /// Computes the boundingBox for the vertices.
   void computeBoundingBox();
 
+  /// Computes if a quad is convex and returns and ordered set of vertices.
```
uekerman, line 194:
```suggestion
  /// Computes if a quad is convex and returns an ordered set of vertices.
```
How is the information whether the quad is convex returned?

src/mesh/Mesh.hpp
```diff
@@ -191,6 +191,13 @@ class Mesh {
...
+  void computeQuadConvexityFromPoints(std::array<int,4> &hull) const;
```
uekerman, line 195: A more descriptive name than "hull" would be good.

src/mesh/Mesh.cpp
```diff
@@ -466,5 +466,137 @@ std::ostream &operator<<(std::ostream &os, const Mesh &m)
   return os;
 }
 
+void Mesh::computeQuadConvexityFromPoints(std::array<int,4> &hull) const
+{
+
+  PRECICE_INFO("Entering compute quad: ");
```
uekerman, line 472:
```suggestion
  PRECICE_DEBUG("Entering compute quad");
```

src/mesh/Mesh.cpp
```diff
@@ -466,5 +466,137 @@ std::ostream &operator<<(std::ostream &os, const Mesh &m)
...
+  // All points need to be projected into a new plane with only 2 coordinates, x and y. These are used to check the quad.
+  // These are store in vx and vy. For now keep same valiues
```
uekerman, line 474:
```suggestion
  // These are stored in vx and vy. For now keep same values
```

src/mesh/Mesh.cpp
```diff
@@ -466,5 +466,137 @@ std::ostream &operator<<(std::ostream &os, const Mesh &m)
...
+  Eigen::Vector3d coords[4];
+
+  // Normal of the plane of first three points in the list of vertices
+  Eigen::Vector3d e_1 = vertices()[hull[1]].getCoords() - vertices()[hull[0]].getCoords();
+  Eigen::Vector3d e_2 = vertices()[hull[2]].getCoords() - vertices()[hull[0]].getCoords();
+  Eigen::Vector3d normalPlane = e_1.cross(e_2);
```
uekerman, line 480: A vector is not a plane. `normalVector` or just `normal` ?

src/mesh/Mesh.cpp
```diff
@@ -466,5 +466,137 @@ std::ostream &operator<<(std::ostream &os, const Mesh &m)
...
+  Eigen::Vector3d diff;
```
uekerman, line 481: diff of what? :smirk: 
Maybe move this into the loop below.

src/mesh/Mesh.cpp
```diff
@@ -466,5 +466,137 @@ std::ostream &operator<<(std::ostream &os, const Mesh &m)
...
+  for (int i = 0; i < 4; i++){
+    diff = vertices()[hull[i]].getCoords() - vertices()[hull[0]].getCoords();
+    coords[i][0] = e_1.dot(diff);
+    coords[i][1] = e_2.dot(diff);
+    coords[i][2] = normalPlane.dot(diff);
+  }
+
+  PRECICE_INFO("Vertex IDs are: " << hull[0] << " " << hull[1] << " " << hull[2] << " " << hull[3]);
+  PRECICE_INFO("X coordinates are: " << coords[0][0] << " " << coords[1][0] << " " << coords[2][0] << " " << coords[3][0]);
+  PRECICE_INFO("Y coordinates are: " << coords[0][1] << " " << coords[1][1] << " " << coords[2][1] << " " << coords[3][1]);
```
uekerman, line 493:
```suggestion
  PRECICE_DEBUG("Vertex IDs are: " << hull[0] << " " << hull[1] << " " << hull[2] << " " << hull[3]);
  PRECICE_DEBUG("X coordinates are: " << coords[0][0] << " " << coords[1][0] << " " << coords[2][0] << " " << coords[3][0]);
  PRECICE_DEBUG("Y coordinates are: " << coords[0][1] << " " << coords[1][1] << " " << coords[2][1] << " " << coords[3][1]);
```

src/mesh/Mesh.cpp
```diff
@@ -466,5 +466,137 @@ std::ostream &operator<<(std::ostream &os, const Mesh &m)
...
+  The first edge in the list is treated as the first edge (edge[0]). An edge that does not share a vertex
+  with the frist edge is the 3rd edge (edge[2]). The edge that shares the first vertex with the first
+  edge (shares vertex[0] with edge[0]) is the 4th edge (edge[3]) as long as it does not share
+  the other vertex with edge[0] (then there is a duplicated edge). The edge sharing vertex[1] and not
+  vertex[0] is the 2nd edge (edge[1]). This also provides the vertex ordering to determine the diagonal, 
+  but not convexity.
+  */
+
+  int edgeOrder[4];
+  //edgeOrder[0] = edgeList[0];   //Will order the edges in edgeList in order that they form a closed quad
```
uekerman, line 561: Do we need this or not? If not, let's not keep it here. Only confuses.

src/mesh/Mesh.hpp
```diff
@@ -191,6 +191,13 @@ class Mesh {
...
+
+  /// Computes if a quad is convex and returns and ordered set of vertices.
+  //void computeQuadEdgeOrder(Vertex &verticesOne, Vertex &verticesTwo, Vertex &verticesThree, Vertex &verticesFour) const;
+  void computeQuadEdgeOrder(std::array<int,4> &edgeList, std::array<int,4>  &vertexList) const;
```
uekerman, line 199: How is invalid input treated in this function? Meaning duplicated edges etc.

src/precice/SolverInterface.hpp
```diff
@@ -495,8 +495,9 @@ class SolverInterface {
    * @param[in] secondEdgeID ID of the second edge of the Quad
    * @param[in] thirdEdgeID ID of the third edge of the Quad
    * @param[in] fourthEdgeID ID of the forth edge of the Quad
-   *
-   * @pre edges with firstEdgeID, secondEdgeID, thirdEdgeID, and fourthEdgeID were added to the mesh with the ID meshID
+   * @param[in] fifthEdgeID ID of the fifth edge of the Quad connecting two opposing points
+   * 
+   * @pre edges with firstEdgeID, secondEdgeID, thirdEdgeID, fourthEdgeID and fifthEdgeID were added to the mesh with the ID meshID
```
uekerman, line 500: Comments do not match the signature below.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -773,11 +773,43 @@ void SolverInterfaceImpl::setMeshQuad(
...
+    edgeList[3] = fourthEdgeID;
+
+
+    // Write out a list of all vertex ID's that are eventually connected in V0-V1-V2-V3-V0 order to get diagonal distances.
+    mesh::Vertex *vertices[4];
+    mesh->computeQuadEdgeOrder(edgeList,vertexList);  // returns reordered list of edgeList and vertices in order of quad
+    mesh->computeQuadConvexityFromPoints(vertexList);
+    PRECICE_CHECK(utils::unique_elements(utils::make_array(vertices[0]->getCoords(),
+                                                           vertices[1]->getCoords(), vertices[2]->getCoords(), vertices[3]->getCoords())),
+                  "The coordinates of the vertices must be unique!");
```
uekerman, line 796: We need a better error message here. And no "!" please :wink:

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -773,11 +773,43 @@ void SolverInterfaceImpl::setMeshQuad(
     PRECICE_CHECK(mesh->isValidEdgeID(secondEdgeID), "Given EdgeID is invalid!");
     PRECICE_CHECK(mesh->isValidEdgeID(thirdEdgeID), "Given EdgeID is invalid!");
     PRECICE_CHECK(mesh->isValidEdgeID(fourthEdgeID), "Given EdgeID is invalid!");
-    mesh::Edge &e0 = mesh->edges()[firstEdgeID];
-    mesh::Edge &e1 = mesh->edges()[secondEdgeID];
-    mesh::Edge &e2 = mesh->edges()[thirdEdgeID];
-    mesh::Edge &e3 = mesh->edges()[fourthEdgeID];
-    mesh->createQuad(e0, e1, e2, e3);
+    mesh::Edge *e[5];
```
uekerman, line 776: Let's add some comment here to explain what the fifth edge is.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -803,13 +835,41 @@ void SolverInterfaceImpl::setMeshQuadWithEdges(
...
+    std::array<int,4>  hull;
+    hull[0] = firstVertexID;
+    hull[1] = secondVertexID;
+    hull[2] = thirdVertexID;
+    hull[3] = fourthVertexID;
 
-    mesh->createQuad(*edges[0], *edges[1], *edges[2], *edges[3]);
+    PRECICE_CHECK(utils::unique_elements(utils::make_array(vertices[0]->getCoords(),
+                                                           vertices[1]->getCoords(), vertices[2]->getCoords(), vertices[3]->getCoords() )),
+                  "The first triangles coordinates of the vertices must be unique!");
```
uekerman, line 846: Triangle coordinates?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -803,13 +835,41 @@ void SolverInterfaceImpl::setMeshQuadWithEdges(
...
+
+    // Don't assume connectivity. Use gift wrapping algorithm to check quad quality and connectivity.
+    mesh->computeQuadConvexityFromPoints(hull);
+
+    Eigen::VectorXd distance1(_dimensions),distance2(_dimensions);
+
+    distance1 = vertices[hull[0]]->getCoords() - vertices[hull[2]]->getCoords();
+    distance2 = vertices[hull[1]]->getCoords() - vertices[hull[3]]->getCoords();
+    PRECICE_INFO("Distance 1: " << distance1.norm() << " and Distance 2: " << distance2.norm());
```
uekerman, line 855:
```suggestion
    PRECICE_DEBUG("Distance 1: " << distance1.norm() << " and Distance 2: " << distance2.norm());
```

Use INFO with care. Only if the user really needs to know this.

src/mesh/Mesh.cpp
KyleDavisSA: Do you mean the commented line `//edgeOrder[0] = edgeList[0]`. I have removed it.

src/mesh/Mesh.hpp
KyleDavisSA: I could add a check if a valid convex is true, and make the function a bool that returns true for a valid quad.
```
bool computeQuadConvexityFromPoints(std::array<int,4> &vertices) const;
```
If untrue, the check would fail and give an error about an invalid quad.

src/mesh/Mesh.hpp
KyleDavisSA: Would `vertexList` be better is it contains a list of vertix ID's?

src/mesh/Mesh.hpp
KyleDavisSA: These are now checked in `SolverInterfaceImpl` to ensure they are unique before calling `computeQuadEdgeOrder`.

src/mesh/Mesh.hpp
uekerman: `vertexList` does not carry the information that those are IDs.
`vertexIDs`?

src/mesh/Mesh.cpp
```diff
@@ -466,5 +466,139 @@ std::ostream &operator<<(std::ostream &os, const Mesh &m)
...
+  Eigen::Vector3d e_1 = vertices()[vertexIDs[1]].getCoords() - vertices()[vertexIDs[0]].getCoords();
+  Eigen::Vector3d e_2 = vertices()[vertexIDs[2]].getCoords() - vertices()[vertexIDs[0]].getCoords();
+  Eigen::Vector3d normalVector = e_1.cross(e_2);
+
+  //Transform Coordinates - coord[0] is the origin
+  for (int i = 0; i < 4; i++){
+    Eigen::Vector3d distance = vertices()[vertexIDs[i]].getCoords() - vertices()[vertexIDs[0]].getCoords();
+    coords[i][0] = e_1.dot(distance);
+    coords[i][1] = e_2.dot(distance);
+    coords[i][2] = normalVector.dot(distance);
```
uekerman, line 488:
```suggestion
    Eigen::Vector3d difference = vertices()[vertexIDs[i]].getCoords() - vertices()[vertexIDs[0]].getCoords();
    coords[i][0] = e_1.dot(difference);
    coords[i][1] = e_2.dot(difference);
    coords[i][2] = normalVector.dot(difference);
```
I am also not really happy with `difference`, but is sounds more like a vector than `distance`. Could not find a better name right now.

src/mesh/Mesh.hpp
```diff
@@ -191,6 +191,18 @@ class Mesh {
...
+  bool computeQuadConvexityFromPoints(std::array<int,4> &vertexIDs) const;
+
+  /**
+   * @brief Computes the ordering of edges to form a quad.
+   *
+   * Receives a set of edges and re-orders them to form a quad. A list
+   * of four unique vertices is returned that form the quad
+   *
+   */
+  std::array<int,4> computeQuadEdgeOrder(std::array<int,4> &edgeList) const;
```
uekerman, line 204:
```suggestion
  /**
   * @brief Computes the ordering of edges to form a quad.
   *
   * Receives a set of edge IDs and re-orders them to form a quad. A list
   * of four vertex IDs is returned that form the quad
   *
   */
  std::array<int,4> computeQuadEdgeOrder(std::array<int,4> &edgeList) const;
```

src/mesh/Mesh.cpp
```diff
@@ -466,5 +466,139 @@ std::ostream &operator<<(std::ostream &os, const Mesh &m)
...
+
+  //Ordering of quad is hull 0-1-2-3-0
+  return (validVertexIDCounter == 4);
+}
+
+std::array<int,4> Mesh::computeQuadEdgeOrder(std::array<int,4> &edgeList) const
+{
+  /*
+  The first edge in the list is treated as the first edge (edge[0]). An edge that does not share a vertex
+  with the frist edge is the 3rd edge (edge[2]). The edge that shares the first vertex with the first
```
uekerman, line 555:
```suggestion
  with the first edge is the 3rd edge (edge[2]). The edge that shares the first vertex with the first
```

src/mesh/Mesh.cpp
```diff
@@ -466,5 +466,139 @@ std::ostream &operator<<(std::ostream &os, const Mesh &m)
...
+  but not convexity.
+  */
+  int edgeOrder[4]; // Will contain new order of edges to form a closed quad
+  std::array<int,4> vertexIDs;
+
+  // The first two vertices are the points on the first edge in edgeList. The other edges are built around this
+  vertexIDs[0] = edges()[edgeList[0]].vertex(0).getID();
+  vertexIDs[1] = edges()[edgeList[0]].vertex(1).getID();
+
+  for (int j = 1; j < 4; j++){  // looping thorugh edges 2, 3 and 4 in edgeList
```
uekerman, line 568:
```suggestion
  for (int j = 1; j < 4; j++){  // looping through edges 2, 3 and 4 in edgeList
```

src/mesh/tests/MeshTest.cpp
```diff
@@ -491,5 +491,123 @@ BOOST_AUTO_TEST_CASE(ComputeStateOfNotFullyConnectedMesh)
...
+  Vertex &v2 = mesh.createVertex(coords2);
+  Vertex &v3 = mesh.createVertex(coords3);
+
+  std::array<int,4>  vertexList;
+  vertexList[0] = v0.getID();
+  vertexList[1] = v1.getID();
+  vertexList[2] = v2.getID();
+  vertexList[3] = v3.getID();
+
+  BOOST_TEST(mesh.computeQuadConvexityFromPoints(vertexList) == 1);
```
uekerman, line 519:
```suggestion
  BOOST_TEST(mesh.computeQuadConvexityFromPoints(vertexList));
```

src/mesh/tests/MeshTest.cpp
```diff
@@ -491,5 +491,123 @@ BOOST_AUTO_TEST_CASE(ComputeStateOfNotFullyConnectedMesh)
...
+  Vertex &v2 = mesh.createVertex(coords2);
+  Vertex &v3 = mesh.createVertex(coords3);
+
+  std::array<int,4>  vertexList;
+  vertexList[0] = v0.getID();
+  vertexList[1] = v1.getID();
+  vertexList[2] = v2.getID();
+  vertexList[3] = v3.getID();
+
+  BOOST_TEST(mesh.computeQuadConvexityFromPoints(vertexList) == 0);
```
uekerman, line 551:
```suggestion
  BOOST_TEST(not mesh.computeQuadConvexityFromPoints(vertexList));
```

src/precice/SolverInterface.hpp
```diff
@@ -495,9 +495,9 @@ class SolverInterface {
    * @param[in] secondEdgeID ID of the second edge of the Quad
    * @param[in] thirdEdgeID ID of the third edge of the Quad
    * @param[in] fourthEdgeID ID of the forth edge of the Quad
-   *
-   * @pre edges with firstEdgeID, secondEdgeID, thirdEdgeID, and fourthEdgeID were added to the mesh with the ID meshID
-   *
+   * 
+   * @pre edges with firstEdgeID, secondEdgeID, thirdEdgeID and fourthEdgeID were added to the mesh with the ID meshID.
+   
    * @warning Quads are not fully implemented yet.
```
uekerman, line 501:
```suggestion
   *
   * @pre edges with firstEdgeID, secondEdgeID, thirdEdgeID and fourthEdgeID were added to the mesh with the ID meshID.
   *
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -773,11 +779,54 @@ void SolverInterfaceImpl::setMeshQuad(
...
+    mesh::Edge *e[5];   //Contains the four edges of the quad and the new diagonal that will be created. 
+    e[0] = &mesh->edges()[firstEdgeID];
+    e[1] = &mesh->edges()[secondEdgeID];
+    e[2] = &mesh->edges()[thirdEdgeID];
+    e[3] = &mesh->edges()[fourthEdgeID];
+
+    PRECICE_CHECK(utils::unique_elements(utils::make_array(firstEdgeID, secondEdgeID, thirdEdgeID, fourthEdgeID)),
+                  "The four edge ID's are not unique. Please check that the edges that form the quad are correct.");
+
+    std::array<int,4>  edgeList;  // A list of edge ID numbers
```
uekerman, line 791:
```suggestion
    std::array<int,4>  edgeIDs;  
```
more follow-up changes necessary below

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -773,11 +779,54 @@ void SolverInterfaceImpl::setMeshQuad(
...
+    edgeList[3] = fourthEdgeID;
+
+    // Write out a list of all vertex ID's that are eventually connected in 
+    // V0-V1-V2-V3-V0 order to get diagonal distances.
+    mesh::Vertex *vertices[4];
+    // Reorders the edgeList and sets the vertexIDs order.
+    std::array<int,4> vertexIDs = mesh->computeQuadEdgeOrder(edgeList); 
+    
+    // Computes vertex order for the convex quad and reorders the variable vertexIDs
+    PRECICE_CHECK(mesh->computeQuadConvexityFromPoints(vertexIDs),"Quad is not convex.");
```
uekerman, line 804: Please extend error message.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -773,11 +779,54 @@ void SolverInterfaceImpl::setMeshQuad(
...
+    PRECICE_CHECK(mesh->computeQuadConvexityFromPoints(vertexIDs),"Quad is not convex."); 
+    vertices[0] = &mesh->vertices()[vertexIDs[0]];
+    vertices[1] = &mesh->vertices()[vertexIDs[1]];
+    vertices[2] = &mesh->vertices()[vertexIDs[2]];
+    vertices[3] = &mesh->vertices()[vertexIDs[3]];
+ 
+    PRECICE_CHECK(utils::unique_elements(utils::make_array(vertices[0]->getCoords(), vertices[1]->getCoords(), 
+                                                           vertices[2]->getCoords(), vertices[3]->getCoords())),
+                  "The four vertices that form the quad are not unique. The resulting shape may be a point, line or triangle." \
+                  "Please check that the edges that form the quad are correct.");
```
uekerman, line 813: Shouldn't this check come before the one in line 804?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -803,13 +854,42 @@ void SolverInterfaceImpl::setMeshQuadWithEdges(
...
+    vertexIDs[2] = thirdVertexID;
+    vertexIDs[3] = fourthVertexID;
+
+    PRECICE_CHECK(utils::unique_elements(utils::make_array(vertices[0]->getCoords(), vertices[1]->getCoords(), 
+                                                           vertices[2]->getCoords(), vertices[3]->getCoords() )),
+                  "The four vertices that form the quad are not unique. The resulting shape may be a point, line or triangle." \
+                  "Please check that the vertices that form the quad are correct.");
+
+    // Computes vertex order for the convex quad and reorders the variable vertexIDs
+    PRECICE_CHECK(mesh->computeQuadConvexityFromPoints(vertexIDs),"Quad is not convex.");
```
uekerman, line 869: Extend error message

src/precice/impl/SolverInterfaceImpl.cpp
KyleDavisSA: Yes, but then I would have to create the vertices, check for uniqueness, then reorder them and check for convexity and reset the vertices. Perhaps check for uniqueness of the vertexIDs?

src/mesh/Mesh.cpp
KyleDavisSA: euclideanDistance would be the most accurate?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -707,6 +707,8 @@ void SolverInterfaceImpl::setMeshTriangle(
 {
   PRECICE_TRACE(meshID, firstEdgeID,
                 secondEdgeID, thirdEdgeID);
+  PRECICE_CHECK(_dimensions == 3, "setMeshTriangle is only possible for 3D cases."\
```
KyleDavisSA, line 710: Is it necessary to add an assertion check for dim != 3 when we have a check?

src/precice/impl/SolverInterfaceImpl.cpp
uekerman: > Perhaps check for uniqueness of the vertexIDs?

if that covers all case already, good!

src/precice/impl/SolverInterfaceImpl.cpp
uekerman: The assertions we need in both `mesh` functions. To be sure that those are not used the wrong way in the future by somebody else.

fsimonis:
So, I am currently trying to refactor the code and have a question.

Is it necessary to reorder the vertices and edges whist checking for convexity?

If the 4 given edges form a connected chain and the vertices of that chain are convex, then this should suffice. Right?
As you currently compute the clockwise order of convex vertices, one would have to check if the convex order is the same as the chain's original or reversed order.

KyleDavisSA:
It is necessary with the way we check for the diagonal. When checking which diagonal is longer, we choose either vertex0-vertex2, or vertex1-vertex3. If the ordering is wrong then they won't necessarily be the two opposite points of the quad.

fsimonis:
I refactored the convexity check into `math::geometry::isConvexQuad` and the edge sorting into `mesh::asChain`.
I shifted the unit tests around for individual parts of this PR and added some additional tests.

The tests for the convexity check, however are not succeeding. The results for the reversed unit square/quad are bogus.

fsimonis:
@uekerman If we commit to the triangulation of quads, then there is no reason to keep the dormant quad-machinery inside preCICE. The question is whether we really want to kick them out once and for all, or keep them around?

uekerman:
> @uekerman If we commit to the triangulation of quads, then there is no reason to keep the dormant quad-machinery inside preCICE. The question is whether we really want to kick them out once and for all, or keep them around?

See my check list above. I am in for removing them. I don't see a use case for them any time soon.

fsimonis:
What is the rational of the last part of this error message:
> ERROR: The given quad is not convex. Please check that the adapter send the four correct vertices or that the interface is composed of quads. A mix of triangles and quads are not supported.

What is the problem with mixing triangles and quads? As we triangulate quads, this should make no difference.

fsimonis:
Resolves #153

fsimonis:
@KyleDavisSA Do you have an update on the convexity issue?

KyleDavisSA:
We take the first vertex in the array and use that as the origin of a new 2D plane. However when calculating the points on the 2D plane, we were now calculating them in reference to the new origin of `(0,0,0)` and not the original origin points.

uekerman:
What is the state of the TODOs listed [above](https://github.com/precice/precice/pull/697#pullrequestreview-449666199)?

uekerman:
* [ ] Check if the setMeshQuadXXX are already part of all the bindings as well.

docs/changelog/697.md
```diff
@@ -0,0 +1 @@
+* Added support for planar quads based on triangulation. Quads will be split along the shortest diagonal to maximize the area-to-perimeter ratio. Note that this triangulation will be visible in configured mesh exports.
```
uekerman, line 1:
```suggestion
* Added support for planar quads based on triangulation, needed, for instance for nearest-projection mapping. Quads will be split along the shortest diagonal to minimize the mapping error. Note that this triangulation will be visible in configured mesh exports.
```

fsimonis: Is the use case really worth mentioning here?

uekerman: Yes, that's how users relate to quads. Otherwise we will only get "wait, what do I need quads for?"

src/sources.cmake
```diff
@@ -190,20 +190,20 @@ target_sources(precice
...
     src/mesh/Triangle.cpp
     src/mesh/Triangle.hpp
+    src/mesh/Utils.hpp
     src/mesh/Vertex.cpp
     src/mesh/Vertex.hpp
     src/mesh/config/DataConfiguration.cpp
     src/mesh/config/DataConfiguration.hpp
     src/mesh/config/MeshConfiguration.cpp
     src/mesh/config/MeshConfiguration.hpp
+    src/mesh/impl/BBUtils.hpp
```
uekerman, line 206: Why is this (only) added here? Why did it not give problems earlier?

src/utils/tests/AlgorithmTest.cpp
```diff
@@ -131,6 +131,54 @@ BOOST_AUTO_TEST_CASE(EmptyRange)
...
+  PRECICE_TEST(1_rank);
+  std::array<int, 3> input{3,4,5};
+  std::array<int, 3> order{2,0,1};
+  std::array<int, 3> expected{5,3,4};
+
+  auto reordered = utils::reorder_array(order, input);
+  BOOST_TEST(reordered == expected);
+}
+
+BOOST_AUTO_TEST_SUITE_END() // ReorderArray
```
uekerman, line 180: Should we also add a test where `T` and `Index` don't have the same typ? Or I am misunderstanding sth here?

src/math/geometry.cpp
```diff
@@ -243,6 +243,81 @@ int containedInTriangle(
   return NOT_CONTAINED;
 }
 
+ConvexityResult isConvexQuad(std::array<Eigen::VectorXd, 4> coords)
+{
+  /*
+    All points need to be projected into a new plane with only 2 coordinates, x' and y'. These are used to check 
+    the convexity of the quad. These new coordinates are stored in 'coords'. Vertices in 2D does not need to
+    be transformed.
```
uekerman, line 251:
```suggestion
    All points need to be projected into a new plane with only 2 coordinates, x' and y'. These are used to check 
    the convexity of the quad. These new coordinates are stored in 'coords'.
```
2D makes no sense here. Could you please add an assertion to safeguard against that?

src/math/geometry.cpp
```diff
@@ -243,6 +243,81 @@ int containedInTriangle(
   return NOT_CONTAINED;
 }
 
+ConvexityResult isConvexQuad(std::array<Eigen::VectorXd, 4> coords)
```
uekerman, line 246: coords could now be `const`, right?

src/math/tests/GeometryTest.cpp
```diff
@@ -700,6 +701,173 @@ BOOST_AUTO_TEST_CASE(ContainedInHyperrectangle)
...
+  Eigen::VectorXd coords3(dim);
+  coords0 << 0.5, 0.34, 0.0;
+  coords1 << 0.62, 0.32, 0.0;
+  coords2 << 0.6, 0.24, 0.0;
+  coords3 << 0.3, 0.22, 0.5;
+
+  auto vertexList = utils::make_array(coords0, coords1, coords2, coords3);
+  auto result     = geometry::isConvexQuad(vertexList);
+
+  BOOST_TEST(result.convex);
```
uekerman, line 797: so this is not planar, right? Where do we catch this?

src/math/tests/GeometryTest.cpp
```diff
@@ -700,6 +701,173 @@ BOOST_AUTO_TEST_CASE(ContainedInHyperrectangle)
...
+  BOOST_TEST(utils::unique_elements(result.vertexOrder));
+  BOOST_TEST_MESSAGE("Vertex Order" << result.vertexOrder);
+  BOOST_TEST(result.vertexOrder[0] == 3);
+  BOOST_TEST(result.vertexOrder[1] == 2);
+  BOOST_TEST(result.vertexOrder[2] == 1);
+  BOOST_TEST(result.vertexOrder[3] == 0);
+}
+
+BOOST_AUTO_TEST_CASE(ComputeValidQuadConvexityWithEqualLeftPoint)
```
uekerman, line 806: What do you mean by "equal left point"? Isn't this test identical to `ComputeUnitQuadConvexity`? Copy-paste mistake?

src/mesh/Utils.hpp
```diff
@@ -0,0 +1,140 @@
...
+#include <mesh/Edge.hpp>
+#include <mesh/Mesh.hpp>
+#include <utility>
+
+namespace precice {
+namespace mesh {
+
+/** return a pointer to the shared vertex of 2 edges
+ * @param[in] a pointer to Edge a
+ * @param[in] a pointer to Edge a
```
uekerman, line 14:
```suggestion
 * @param[in] reference to edge a
 * @param[in] reference to Edge b
```

src/mesh/Utils.hpp
```diff
@@ -0,0 +1,140 @@
...
+  Vertex *b0 = &b.vertex(0);
+  Vertex *b1 = &b.vertex(1);
+  if (a0 == b0 || a0 == b1) {
+    return a0;
+  }
+  if (a1 == b0 || a1 == b1) {
+    return a1;
+  }
+  return nullptr;
+}
```
uekerman, line 30: What if both edges are identical (same start and end vertex)?

src/mesh/Utils.hpp
```diff
@@ -0,0 +1,140 @@
...
+ * Also, the first and last vertex of the chain will be the vertices of the first edge of the argument.
+ *
+ * @param[in] edges an array of pointers to edges to chain together
+ *
+ * @returns A \ref Chain of the input
+ */
+template <std::size_t n>
+Chain<n> asChain(std::array<mesh::Edge *, n> edges)
+{
+  static_assert(n > 1, "You already know the answer.");
```
uekerman, line 66: Why do you use `static_assert` here?

src/mesh/Utils.hpp
```diff
@@ -0,0 +1,140 @@
...
+  /*
+  The first edge in the list is treated as the first edge (edge[0]). An edge that does not share a vertex
+  with the first edge is the 3rd edge (edge[2]). The edge that shares the first vertex with the first
+  edge (shares vertex[0] with edge[0]) is the 4th edge (edge[3]) as long as it does not share
+  the other vertex with edge[0] (then there is a duplicated edge). The edge sharing vertex[1] and not
+  vertex[0] is the 2nd edge (edge[1]). This also provides the vertex ordering to determine the diagonal, 
+  but not convexity.
```
uekerman, line 73: This comment sounds a lot like N needs to be 4.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -873,11 +882,42 @@ void SolverInterfaceImpl::setMeshQuad(
...
+    auto coords = mesh::coordsFor(chain.vertices);
+    PRECICE_CHECK(utils::unique_elements(coords),
+                  "The four vertices that form the quad are not unique. The resulting shape may be a point, line or triangle."
+                  "Please check that the adapter sends the four unique vertices that form the quad, or that the mesh on the interface "
+                  "is composed of quads. A mix of triangles and quads are not supported.");
+
+    auto convexity = math::geometry::isConvexQuad(coords);
+    PRECICE_CHECK(convexity.convex, "The given quad is not convex. "
+                                    "Please check that the adapter send the four correct vertices or that the interface is composed of quads. "
+                                    "A mix of triangles and quads are not supported.");
```
uekerman, line 903: Why again the "A mix of triangles and quads are not supported"?

src/sources.cmake
fsimonis: CMake only requires compilation units / cpp files to be listed. The compiler is responsible for finding the right headers.
Listing them will make them appear in IDEs though, so it is generally recommended.

src/math/geometry.cpp
fsimonis: I prefer to pass them by value here, as we need to transform them internally. So this just saves us to explicitly copy them inside the function.
If that was not the case `const T&` would probably make sense.

src/math/tests/GeometryTest.cpp
fsimonis: This should happen in `isConvexQuad` and return `convex == false`.
Good spot!

src/mesh/Utils.hpp
fsimonis: Still returns one shared vertex.
This is valid behaviour, will document it.

src/mesh/Utils.hpp
fsimonis: This comment can be removed entirely,

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -864,6 +870,9 @@ void SolverInterfaceImpl::setMeshQuad(
 {
   PRECICE_TRACE(meshID, firstEdgeID, secondEdgeID, thirdEdgeID,
                 fourthEdgeID);
+  PRECICE_CHECK(_dimensions == 3, "setMeshQuad is only possible for 3D cases."
+                                  " Please set the dimension to 3 in the preCICE configuration file.");
+  PRECICE_ASSERT(_dimensions == 3, _dimensions);
```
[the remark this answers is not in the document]
fsimonis: Good question. It has to go.

src/precice/impl/SolverInterfaceImpl.cpp
fsimonis: I didn't understand it either. Removing (twice).

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -890,27 +930,57 @@ void SolverInterfaceImpl::setMeshQuadWithEdges(
 {
   PRECICE_TRACE(meshID, firstVertexID,
                 secondVertexID, thirdVertexID, fourthVertexID);
+  PRECICE_CHECK(_dimensions == 3, "setMeshQuadWithEdges is only possible for 3D cases."
+                                  " Please set the dimension to 3 in the preCICE configuration file.");
+  PRECICE_ASSERT(_dimensions == 3, _dimensions);
```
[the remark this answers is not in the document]
fsimonis: ditto

src/mesh/Utils.hpp
fsimonis: It tells a silly developer off at compile time.

PR #701, opened 2020-03-26
Fix critical bug in ReceivedPartition

fsimonis:
The current implementation overwrites the local bounding box in filterByBoundingBox when receiving the bounding boxes from the slaves.
This leads to incorrect mappings which affects the repartitioning.

This error was missed due to #700.

### DIFF
--- a/src/partition/ReceivedPartition.cpp
+++ b/src/partition/ReceivedPartition.cpp
@@ -238,6 +238,19 @@ void ReceivedPartition::compute()
   }
 }
 
+namespace {
+bool isVertexInBoundingBox(mesh::Vertex const &vertex, mesh::Mesh::BoundingBox const &bb)
+{
+  const int dim = bb.size();
+  for (int d = 0; d < dim; d++) {
+    if (vertex.getCoords()[d] < bb[d].first || vertex.getCoords()[d] > bb[d].second) {
+      return false;
+    }
+  }
+  return true;
+}
+} // namespace
+
 void ReceivedPartition::filterByBoundingBox()
 {
   PRECICE_TRACE(_geometricFilter);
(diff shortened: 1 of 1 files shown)


src/partition/ReceivedPartition.cpp
```diff
@@ -506,12 +523,7 @@ void ReceivedPartition::prepareBoundingBox()
 
 bool ReceivedPartition::isVertexInBB(const mesh::Vertex &vertex)
```
uekerman:
```suggestion
bool ReceivedPartition::isVertexInBoundingBox(const mesh::Vertex &vertex)
```

src/partition/ReceivedPartition.cpp
```diff
@@ -506,12 +523,7 @@ void ReceivedPartition::prepareBoundingBox()
...
 {
-  for (int d = 0; d < _dimensions; d++) {
-    if (vertex.getCoords()[d] < _bb[d].first or vertex.getCoords()[d] > _bb[d].second) {
-      return false;
-    }
-  }
-  return true;
+  return vertexInBoundingBox(vertex, _bb);
```
uekerman:
```suggestion
  return isVertexInBoundingBox(vertex, _bb);
```

src/partition/ReceivedPartition.cpp
```diff
@@ -238,6 +238,19 @@ void ReceivedPartition::compute()
   }
 }
 
+namespace {
+bool vertexInBoundingBox(mesh::Vertex const &vertex, mesh::Mesh::BoundingBox const &bb)
```
uekerman:
```suggestion
bool isVertexInBoundingBox(mesh::Vertex const &vertex, mesh::Mesh::BoundingBox const &bb)
```

src/partition/ReceivedPartition.cpp
```diff
@@ -279,12 +292,14 @@ void ReceivedPartition::filterByBoundingBox()
...
       for (int rankSlave = 1; rankSlave < utils::MasterSlave::getSize(); rankSlave++) {
-        com::CommunicateMesh(utils::MasterSlave::_communication).receiveBoundingBox(_bb, rankSlave);
+        mesh::Mesh::BoundingBox slaveBB(_bb.size(), std::make_pair(std::numeric_limits<double>::max(), std::numeric_limits<double>::lowest()));
+        com::CommunicateMesh(utils::MasterSlave::_communication).receiveBoundingBox(slaveBB, rankSlave);
 
-        PRECICE_DEBUG("From slave " << rankSlave << ", bounding mesh: " << _bb[0].first
-                                    << ", " << _bb[0].second << " and " << _bb[1].first << ", " << _bb[1].second);
+        PRECICE_DEBUG("From slave " << rankSlave << ", bounding mesh: "
+                                    << slaveBB[0].first << ", " << slaveBB[0].second << " and "
+                                    << slaveBB[1].first << ", " << slaveBB[1].second);
```
uekerman: In 3D, this debug statement feels a bit incomplete, but OK, it's just a debug statement.

fsimonis:
@uekerman I renamed the function and print the bounding box as a vector, so it is universal.

PR #689, opened 2020-03-01
Refactor CouplingScheme class hierarchy

BenjaminRueth:
Major refactoring of CouplingScheme and derived classes:

* Make **all** member variables private. Reduce getters and setters to the necessary minimum.
* Remove unused constructor for `BaseCouplingScheme`
* Remove unused functions in `CouplingScheme`
* Provide `CouplingMode` to constructor of `BaseCouplingScheme`. Makes it more explicit that `MultiCouplingScheme` is always `Implicit`.
* Apply [template method pattern](https://en.wikipedia.org/wiki/Template_method_pattern) to `advance()`, `initialize()`, `initializeData()` and `setupConvergenceMeasures()`. This avoids code duplication for common parts of the functions and provides a template for implementing the different parts in the derived classes.
* Improve encapsulation of `CouplingScheme` w.r.t `MMAcceleration`. Pointer to member variable `_isCoarseModelOptimizationActive` was provided before. `MMAcceleration` changed `_isCoarseModelOptimizationActive` via this pointer. Now `BaseCouplingScheme` provide a common interface (`getIsCoarseModelOptimizationActive`, `activateCoarseModelOptimization`, `deactivateCoarseModelOptimization`) such that `MMAcceleration` and derived coupling schemes can control coarse model optimization.
* Update documentation (@uekerman There are still some open TODOs, where I would need help)
* Consistently use terms time window / time step

# Open Todos

- [x] Fix Documentation
- [x] Time Window / Time Step
- [x] Run clang-format
- [x] Write Changelog entry

PR #703, opened 2020-03-26
Remove Manifold Mapping

uekerman:
After careful discussion we decided to remove the feature "Manifold Mapping". It had just become too deprecated. No unit tests, no integration tests, and no known user. At the same time, it made future develops in coupling scheme (waveform data structures) unnecessarily complicated.

I leave the API function `hasToEvaluateSurrogateModel` and `hasToEvaluateFineModel` is nop stubs. The first returns `false`, the second `true`. This way, adapters that use these functions (if they are any) should still work, but just never do "multi-level".

Reviewers:
@BenjaminRueth Please double check that I did not forget anything. Can you merge this into #689 ?
@fsimonis Building, API, etc all good?

Before merging, we should check the user documentation for any remainders.


src/cplscheme/SerialCouplingScheme.cpp
```diff
@@ -193,129 +193,59 @@ void SerialCouplingScheme::explicitAdvance()
...
+    PRECICE_DEBUG("measure convergence.");
+    // measure convergence of the coupling iteration,
+    convergence = measureConvergence();
+    // Stop, when maximal iteration count (given in config) is reached
+    if (maxIterationsReached())
+      convergence = true;
+
+    // coupling iteration converged for current time window. Advance in time.
+    if (convergence) {
+      if (getAcceleration().get() != nullptr) {
```
fsimonis, line 219: Same as above

src/cplscheme/SerialCouplingScheme.cpp
```diff
@@ -193,129 +193,59 @@ void SerialCouplingScheme::explicitAdvance()
...
-        convergence = true;
-    }
+      newConvergenceMeasurements();
+      timeWindowCompleted();
 
-    // passed by reference, modified in MM acceleration. No-op for all other accelerations
-    if (getAcceleration().get() != nullptr) {
-      getAcceleration()->setCoarseModelOptimizationActive(&_isCoarseModelOptimizationActive);
+      // no convergence achieved for the coupling iteration within the current time step
+    } else if (getAcceleration().get() != nullptr) {
```
fsimonis, line 227: Same as above

src/cplscheme/ParallelCouplingScheme.cpp
```diff
@@ -189,109 +189,51 @@ void ParallelCouplingScheme::explicitAdvance()
...
-      convergence = false;
-      // in case of multilevel PP only: if coarse model optimization converged
-      // steering the requests for evaluation of coarse and fine model, respectively
-      if (convergenceCoarseOptimization) {
-        _isCoarseModelOptimizationActive = false;
-        doOnlySolverEvaluation           = true;
-      } else {
-        _isCoarseModelOptimizationActive = true;
+    if (convergence) {
+      if (getAcceleration().get() != nullptr) {
```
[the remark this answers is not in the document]
uekerman: :+1: This change is already part of #689.

Issue #699, opened 2020-03-25
Summation action


uekerman:
**Please describe the problem you are trying to solve**

In certain three-field coupled problems, (conservative) data values need to be summed up. A typical examples are forces. Solver A receives forces from solvers B and C on the same geometry (let's say electromagnetic forces and mechanical forces). Currently, this can be solved by having two force data values in solver A and summing them up in the adapter of solver A. The drawback of this approach that we need to modify the adapter of A. Having an action would circumvent this modification. The situation is also motivated in [Alban Rousset's talk at the preCICE Workshop 2020](https://www.youtube.com/watch?v=3FP8y1Zjqns).

**Describe the solution you propose.**

Implement an action that sums up n source data values and writes them to a target data value on the same mesh.

```xml
<participant name="SolverA"> 
    <use-mesh name="MeshA" provide="yes"/> 
    <read-data name="ForcesA" mesh="MeshA"/> 
    ...
    <action:sum mesh="MeshA" timing="on-exchange-post">
        <source-data name="ForcesB"/>
        <source-data name="ForcesC"/>
        <target-data name="ForcesA"/>
    </action:sum>
    ...
</participant>
```

Currently, an action can only have one `source-data`. We need to relax this general restriction in the configuration and enforce it instead via `PRECICE_CHECK` for all other actions.

oguzziya:
I am not sure which one is more convenient to implement for the multiple source data in configuration:

- Changing `sourceDataID` to vector, modifying every action accordingly in their implementation
- Changing `sourceDataID` to vector, modifying only `createAction()` method in configuration (seems more maintainable for me )
- Or something other than using vector (not so sure)

uekerman:
`std::vector` is the right data structure. I would recommend the second option. Ideally, you should not need to edit the other actions at all.

In the configuration, you can create a specific tag
https://github.com/precice/precice/blob/3eb896b45ca90cdbe98c5ec2292a26b3bb04e8cf/src/action/config/ActionConfiguration.cpp#L39
for the new action and change `OCCUR_ONCE` to `OCCUR_ONCE_OR_MORE`.

PR #707, opened 2020-03-31
Summation action

oguzziya:
The purpose of this PR to implement "Summation Action" which is discussed in #699 . The changes are:
- Multiple source data support in the action configuration
- Summation action class
- Summation action tests


src/action/SummationAction.hpp
```diff
@@ -0,0 +1,55 @@
...
+	 * 
+	 * @param[in] timing When to apply the action
+	 * @param[in] source1 First source data
+	 * @param[in] source2 Second source data
+	 * @param[in] target Data in which the action will be applied
+	 * 
+	 */
+	SummationAction(
+		Timing				timing,
+		std::vector<int>	sourceDataID,
```
uekerman: Plural would be better here, "sourceDataIDs"

src/action/SummationAction.hpp
```diff
@@ -0,0 +1,55 @@
...
+		int					targetDataID,
+		const mesh::PtrMesh	&mesh);
+	
+	virtual ~SummationAction() {}
+	 
+	/**
+	 * @brief Adds or substracts the source data to target data
```
uekerman: Please update

src/action/SummationAction.hpp
```diff
@@ -0,0 +1,55 @@
...
+		double time,
+		double dt,
+		double computedPartFullDt,
+		double fullDt);
+
+private:
+	logging::Logger _log{"action::SummationAction"};
+
+	mesh::PtrData _targetData;
+	std::vector<mesh::PtrData> _sourceData;
```
uekerman: Maybe "_sourceDataVector" ?

src/action/SummationAction.cpp
```diff
@@ -0,0 +1,44 @@
+#include "SummationAction.hpp"
+#include "mesh/Data.hpp"
+#include "mesh/Edge.hpp"
```
uekerman: Probably not needed.

src/action/SummationAction.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+SummationAction::SummationAction(
+	Timing				timing,
+	std::vector<int>	sourceDataID,
+	int					targetDataID,
+	const mesh::PtrMesh	&mesh)
+	: Action(timing, mesh, mapping::Mapping::MeshRequirement::VERTEX), _targetData(mesh->data(targetDataID))
+{
+
+	for(auto sourceID : sourceDataID){
```
uekerman:
```suggestion
	for(int sourceDataID : sourceDataIDs){
```
I would recommend to avoid `auto` if it is just a simple type, such as `int`. Probably my personal taste, confuses me more than it helps.

src/action/SummationAction.cpp
```diff
@@ -0,0 +1,44 @@
...
+		_sourceData.push_back(mesh->data(sourceID));
+	}
+
+	for(const auto & source : _sourceData) {
+		PRECICE_ASSERT(source->getDimensions() == _targetData->getDimensions(), source->getDimensions(), _targetData->getDimensions());
```
uekerman: We use `PRECICE_ASSERT` for assertions, so to safeguard implementation bugs. `PRECICE_CHECK`, on the other hand checks whether user input was correct. And if not we through an error. I guess here, we actually need a `PRECICE_CHECK` (it might also be wrong in other implemented actions). [Here](https://github.com/precice/precice/blob/develop/src/precice/impl/SolverInterfaceImpl.cpp#L558-L559) is an example.

src/action/SummationAction.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+void SummationAction::performAction(
+	double time,
+	double dt,
+	double computedPartFullDt,
+	double fullDt)
+{
+	PRECICE_TRACE();
+	auto & targetValues = _targetData->values();
+	auto sum = _sourceData.at(0)->values();
```
uekerman: Why not moving this into the loop? Makes it easier to read.

src/action/SummationAction.cpp
```diff
@@ -0,0 +1,44 @@
...
+
+	for(int i = 1; i < _sourceData.size(); ++i){
+		sum += _sourceData.at(i)->values();
+	}
+	targetValues = sum;
```
uekerman: Mmh, don't we need to loop over the complete mesh? Similarly to [here](https://github.com/precice/precice/blob/develop/src/action/ScaleByDtAction.cpp#L43-L45). Or is the "+=" here overloaded?

src/action/config/ActionConfiguration.cpp
```diff
@@ -37,12 +39,15 @@ ActionConfiguration::ActionConfiguration(
   using namespace xml;
   std::string doc;
   XMLTag      tagSourceData(*this, TAG_SOURCE_DATA, XMLTag::OCCUR_ONCE);
-  tagSourceData.setDocumentation("Data to read from.");
+  tagSourceData.setDocumentation("Single data to read from. ");
+  XMLTag      tagMultipleSourceData(*this, TAG_SOURCE_DATA, XMLTag::OCCUR_ONCE_OR_MORE);
+  tagSourceData.setDocumentation("Multiple data to read from.");
   XMLTag tagTargetData(*this, TAG_TARGET_DATA, XMLTag::OCCUR_ONCE);
-  tagSourceData.setDocumentation("Data to read from and write to.");
+  tagTargetData.setDocumentation("Data to read from and write to.");
```
uekerman: Good catch!

src/action/config/ActionConfiguration.cpp
```diff
@@ -88,6 +93,15 @@ ActionConfiguration::ActionConfiguration(
     tag.addSubtag(tagTargetData);
     tags.push_back(tag);
   }
+  {
+    XMLTag tag(*this, NAME_SUMMATION, occ, TAG);
+    doc = "Sums the multiple source data values and writes the result into ";
```
uekerman:
```suggestion
    doc = "Sums up multiple source data values and writes the result into ";
```

src/action/config/ActionConfiguration.cpp
```diff
@@ -198,15 +212,15 @@ void ActionConfiguration::createAction()
   action::Action::Timing timing = getTiming();
 
   // Determine data and mesh
-  int           sourceDataID = -1;
-  int           targetDataID = -1;
+  std::vector<int>  sourceDataID;
```
uekerman:
```suggestion
  std::vector<int>  sourceDataIDs;
```

src/action/tests/SummationActionTest.cpp
```diff
@@ -0,0 +1,200 @@
+#include "action/SummationAction.hpp"
+#include "action/config/ActionConfiguration.hpp"
+#include "mesh/Data.hpp"
+#include "mesh/Edge.hpp"
```
uekerman: Not needed I guess

src/action/tests/SummationActionTest.cpp
```diff
@@ -0,0 +1,200 @@
...
+using namespace precice;
+
+BOOST_AUTO_TEST_SUITE(ActionTests)
+BOOST_AUTO_TEST_SUITE(Summation, *testing::OnMaster())
+
+BOOST_AUTO_TEST_CASE(SummationTwoSources)
+{
+  using namespace mesh;
+  PtrMesh mesh(new Mesh("Mesh", 3, true, testing::nextMeshID()));
+  PtrData sourceData1   = mesh->createData("SourceData1", 1);
```
uekerman:
```suggestion
  int dimension = 1;
  PtrData sourceData1   = mesh->createData("SourceData1", dimension);
```
Makes it easier to read.

src/action/tests/SummationActionTest.cpp
```diff
@@ -0,0 +1,200 @@
...
+  sum.performAction(0.0, 0.25, 1.0, 1.0);
+  BOOST_TEST(sourceValues1[0] == 2.0);
+  BOOST_TEST(sourceValues1[1] == 3.0);
+  BOOST_TEST(sourceValues1[2] == 4.0);
+  BOOST_TEST(sourceValues2[0] == 4.0);
+  BOOST_TEST(sourceValues2[1] == 5.0);
+  BOOST_TEST(sourceValues2[2] == 1.0);
+  BOOST_TEST(targetValues[0] == 6.0);
+  BOOST_TEST(targetValues[1] == 8.0);
+  BOOST_TEST(targetValues[2] == 5.0);
```
uekerman: I don't think we need all these cases with all different dt settings. More important would be to also have a test with vector-valued data, i.e. `dim=3`.

src/action/tests/SummationActionTest.cpp
```diff
@@ -0,0 +1,200 @@
...
+    mesh::PtrDataConfiguration dataConfig(new mesh::DataConfiguration(tag));
+    dataConfig->setDimensions(3);
+    mesh::PtrMeshConfiguration meshConfig(new mesh::MeshConfiguration(tag, dataConfig));
+    meshConfig->setDimensions(3);
+    action::ActionConfiguration config(tag, meshConfig);
+    xml::configure(tag, xml::ConfigurationContext{}, filename);
+    BOOST_TEST(config.actions().size() == 1);
+    action::PtrAction action = config.actions().front();
+    BOOST_TEST(action);
+  }
```
uekerman: I don't think we need both here.

src/action/SummationAction.cpp
oguzziya: When i checked, it was overloaded for Eigen vectors and matrices but I changed it to the loop version anyways since it is not obvious.

oguzziya:
Currently integration test does not work yet. During the testing, `There cannot be two mappings to mesh "MeshTarget"` error is thrown and test fails.

oguzziya:
In order to seperate the mappings, I carried them to source participants and changed them to write (I am not sure whether it is allowed or not). Two mappings error disappeared but still, integration test fails since action is not performed and target values remain zero.

src/precice/tests/SerialTests.cpp
```diff
@@ -1400,6 +1399,171 @@ BOOST_AUTO_TEST_CASE(testPreconditionerBug,
...
+    int idA = cplInterface.setMeshVertex(meshID, coordA.data());
+    int idB = cplInterface.setMeshVertex(meshID, coordB.data());
+    int idC = cplInterface.setMeshVertex(meshID, coordC.data());
+    int idD = cplInterface.setMeshVertex(meshID, coordD.data());
+
+    int idAB = cplInterface.setMeshEdge(meshID, idA, idB);
+    int idBC = cplInterface.setMeshEdge(meshID, idB, idC);
+    int idCD = cplInterface.setMeshEdge(meshID, idC, idD);
+    int idDA = cplInterface.setMeshEdge(meshID, idD, idA);
+    int idCA = cplInterface.setMeshEdge(meshID, idC, idA);
```
uekerman, line 1482: No need to define edges for this testcase. We only need edges for a nearest-projection mapping. For this test we can just use matching meshes together with a nearest-neighbor mapping.

oguzziya:
I just realized that in the action configuration, while reading tags, source tag is always overwritten. Therefore, i changed `sourceData` to `sourceDataVector` and now serial test and summation action test works fine

PR #708, opened 2020-03-31
Simplify FindNumPy

fsimonis:
This PR simplifies the CMake module for finding NumPy.

It now **exclusively** uses the information from the python module:
* `numpy.__version__` for the version, and
* `numpy.get_include()` for the include directory.

The latter is verified via a the `find_path` CMake command using `NO_DEFAULT_PATH`.

This point of this PR is to keep numpy in sync with the detected python interpreter. This is crucial to support virtual environments reliably.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 All notable changes to this project will be documented in this file. For future plans, see our [Roadmap](https://github.com/precice/precice/wiki/Roadmap).
 
-
+- Change the CMake FindNumPy module to only consider information based on the selected python interpreter.
 - Implemented a new action to sum up data values, `action:summation`.
 
 ## 2.0.2
--- a/cmake/modules/FindNumPy.cmake
+++ b/cmake/modules/FindNumPy.cmake
@@ -1,7 +1,7 @@
 #.rst:
 #
-# Find the include directory for ``numpy/arrayobject.h`` as well as other NumPy tools like ``conv-template`` and
-# ``from-template``.
+# Find the include directory for ``numpy/arrayobject.h``.
+# This module exclusively uses the information from the numpy python module.
 #
 # This module sets the following variables:
 #
@@ -11,25 +11,10 @@
 #   The include directories needed to use NumpPy.
 # ``NumPy_VERSION``
 #   The version of NumPy found.
-# ``NumPy_CONV_TEMPLATE_EXECUTABLE``
-#   Path to conv-template executable.
-# ``NumPy_FROM_TEMPLATE_EXECUTABLE``
-#   Path to from-template executable.
 #
 # The module will also explicitly define one cache variable:
 #
 # ``NumPy_INCLUDE_DIR``
-#
-# .. note::
-#
-#     To support NumPy < v0.15.0 where ``from-template`` and ``conv-template`` are not declared as entry points,
-#     the module emulates the behavior of standalone executables by setting the corresponding variables with the
-#     path the the python interpreter and the path to the associated script. For example:
-#     ::
-#
-#         set(NumPy_CONV_TEMPLATE_EXECUTABLE /path/to/python /path/to/site-packages/numpy/distutils/conv_template.py CACHE STRING "Command executing conv-template program" FORCE)
-#
-#         set(NumPy_FROM_TEMPLATE_EXECUTABLE /path/to/python /path/to/site-packages/numpy/distutils/from_template.py CACHE STRING "Command executing from-template program" FORCE)
 #
 
 if(NOT NumPy_FOUND)
@@ -43,10 +28,8 @@ if(NOT NumPy_FOUND)
   find_package(PythonInterp ${_find_extra_args})
   find_package(PythonLibs ${_find_extra_args})
 
-  find_program(NumPy_CONV_TEMPLATE_EXECUTABLE NAMES conv-template)
-  find_program(NumPy_FROM_TEMPLATE_EXECUTABLE NAMES from-template)
-
   if(PYTHON_EXECUTABLE)
+    unset(_numpy_include_dir)
     execute_process(COMMAND "${PYTHON_EXECUTABLE}"
       -c "import numpy; print(numpy.get_include())"
       OUTPUT_VARIABLE _numpy_include_dir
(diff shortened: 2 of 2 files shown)


fsimonis:
@MakisH So after you removed the pip-installed numpy, there is no numpy installed for python3, right? In that case the detection should fail.
Could it be that the system numpy in `/usr/include` is owned by the `python(2)-numpy` package? This cross-version detection is a common reason why preCICE compiles, but the action tests fail. At runtime `import numpy` fails.

You could try and pass the conda python interpreter to cmake via the `-DPYTHON_EXECUTABLE=..../python` option. The  find module should then pick up the numpy provided by conda.

The ultimate goal of this PR is to keep the numpy include consistent with the version that the python interpreter would find.

MakisH:
> @MakisH So after you removed the pip-installed numpy, there is no numpy installed for python3, right? In that case the detection should fail.

Correct, and this works as expected.

> Could it be that the system numpy in `/usr/include` is owned by the `python(2)-numpy` package? This cross-version detection is a common reason why preCICE compiles, but the action tests fail. At runtime `import numpy` fails.

I am not sure how the packages `python-numpy` and `python3-numpy` are related (if at all) in Ubuntu 16.04, but indeed, there is a package for Python2 and a package for Python3. And I have both.

However, configuring with only `python-numpy` (i.e. the Python2 version) installed does not find NumPy (i.e. it does the right thing). Indeed, the previous approach (i.e. `develop`) finds an empty path and configuration succeeds, but the tests timeout.

> You could try and pass the conda python interpreter to cmake via the `-DPYTHON_EXECUTABLE=..../python` option. The find module should then pick up the numpy provided by conda.

This also works as expected during configure, even when I have different NumPy versions installed:
```
-- Found PythonLibs: /home/makish/inst/miniconda3/lib/libpython3.7m.so (found version "3.7.4") 
-- Found PythonInterp: /home/makish/inst/miniconda3/bin/python3 (found version "3.7.4") 
-- Found NumPy: /home/makish/inst/miniconda3/lib/python3.7/site-packages/numpy/core/include (found version "1.18.1") 
-- Validating LibPython
-- Validating LibPython - success
-- Validating NumPy
-- Validating NumPy - success
```
But the tests seem to fail: 
```
test 2
      Start  2: precice.action

2: Test command: /usr/bin/mpiexec "-n" "4" "/home/makish/github/precice_fsimonis/build/testprecice" "--run_test=ActionTests"
2: Environment variables: 
2:  PRECICE_ROOT=/home/makish/github/precice_fsimonis
2: Test timeout computed to be: 20
2: Test setup error: no test cases matching filter or all test cases were disabled
2: Running 6 test cases...
2: Test setup error: no test cases matching filter or all test cases were disabled
2: Test setup error: no test cases matching filter or all test cases were disabled
2: Could not find platform independent libraries <prefix>
2: Could not find platform dependent libraries <exec_prefix>
2: Consider setting $PYTHONHOME to <prefix>[:<exec_prefix>]
2: Fatal Python error: initfsencoding: Unable to get the locale encoding
2: ModuleNotFoundError: No module named 'encodings'
```


I do, also, get warnings about conflicts with packages from Conda, but this is another story that we shouldn't really worry about:
```
-- Configuring done
CMake Warning at CMakeLists.txt:266 (add_library):
  Cannot generate a safe runtime search path for target precice because files
  in some directories may conflict with libraries in implicit directories:

    runtime library [libgcc_s.so.1] in /usr/lib/gcc/x86_64-linux-gnu/5 may be hidden by files in:
      /home/makish/inst/miniconda3/lib
    runtime library [libquadmath.so.0] in /usr/lib/gcc/x86_64-linux-gnu/5 may be hidden by files in:
      /home/makish/inst/miniconda3/lib
    runtime library [libstdc++.so.6] in /usr/lib/gcc/x86_64-linux-gnu/5 may be hidden by files in:
      /home/makish/inst/miniconda3/lib

  Some of these libraries may not be found correctly.
```

> The ultimate goal of this PR is to keep the numpy include consistent with the version that the python interpreter would find.

I understand that. We can keep debugging my very edge-case system (which may actually be quite common among users), or we can ignore the conda stuff.

fsimonis:
> I understand that. We can keep debugging my very edge-case system (which may actually be quite common among users), or we can ignore the conda stuff.

I am aware of at least one user who uses preCICE with conda. So this is definitely worth it.

Could you please:

1. verify that `/home/makish/inst/miniconda3/lib/` is in your `LD_LIBRARY_PATH`
2.  try setting `PYTHON_HOME=/home/makish/inst/miniconda3` before starting the tests

MakisH:
Taking a step back:

> You could try and pass the conda python interpreter to cmake via the -DPYTHON_EXECUTABLE=..../python option. The find module should then pick up the numpy provided by conda.

Although this works, shouldn't CMake prefer the shell defined by Conda and is in my `$PATH` over the global installation? Usually such a setting means that the user want to use this over the global installation.

> Could you please:
> 
>     1. verify that `/home/makish/inst/miniconda3/lib/` is in your `LD_LIBRARY_PATH`
> 
>     2. try setting `PYTHON_HOME=/home/makish/inst/miniconda3` before starting the tests

Even when setting both (after having built), the `action` test keeps failing. I also tried adding `/home/makish/.local/lib/python3.7/site-packages/numpy/core/lib` to my `LD_LIBRARY_PATH`, but it didn't help. Interestingly, that directory only contains:
```
.
├── libnpymath.a
└── npy-pkg-config
    ├── mlib.ini
    └── npymath.ini
```
I am not sure what other path to try.

fsimonis:
> Although this works, shouldn't CMake prefer the shell defined by Conda and is in my $PATH over the global installation?

The documentation of the [`FindPythonInterp`](https://cmake.org/cmake/help/v3.10/module/FindPythonInterp.html) is not very detailed. Looking at [the code](https://github.com/Kitware/CMake/blob/v3.10.3/Modules/FindPythonInterp.cmake), it simply tries to detect python versions one after the other.

The Python Modules have been reworked in [version 3.12.4](https://cmake.org/cmake/help/v3.12/module/FindPython.html). So a future upgrade will help with this.

MakisH:
> Can you please post the above CMakeCache values when configuring with the conda installation set via `-DPYTHON_EXECUTABLE`?

Here you are:
```
PYTHON_EXECUTABLE                /home/makish/inst/miniconda3/bin/python3                                                                                                                                                                                                           
PYTHON_INCLUDE_DIR               /home/makish/inst/miniconda3/include/python3.7m                                                                                                                                                                                                    
PYTHON_LIBRARY                   /home/makish/inst/miniconda3/lib/libpython3.7m.so                                                                                                                                                                                                  
PYTHON_LIBRARY_DEBUG             PYTHON_LIBRARY_DEBUG-NOTFOUND
NumPy_INCLUDE_DIR                /home/makish/.local/lib/python3.7/site-packages/numpy/core/include
```

I get the same with and without conda in the `LD_LIBRARY_PATH`.

fsimonis:
I just installed `miniconda3` and it worked out of the box for me. :thinking: 
@MakisH Shall we put the issue aside and merge this PR?

MakisH:
> @MakisH Shall we put the issue aside and merge this PR?

Sure, we can open another issue if needed.

PR #709, opened 2020-03-31
Add connectMasterSlaves to Communication

fsimonis:
This PR adds `connectMasterSlaves()` to `com::Communication`.

The participant name, size and current rank are all known at the when establishing the MS connection.
Thus it is safer to pass everything to the communication class which then performs the required calls.
The current class design allows a single implementation for all Comunication classes.

### DIFF
--- a/src/com/Communication.cpp
+++ b/src/com/Communication.cpp
@@ -3,6 +3,31 @@
 
 namespace precice {
 namespace com {
+
+void Communication::connectMasterSlaves(std::string const &participantName,
+                                        std::string const &tag,
+                                        int                rank,
+                                        int                size)
+{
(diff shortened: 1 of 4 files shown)


src/com/Communication.cpp
```diff
@@ -3,6 +3,32 @@
 
 namespace precice {
 namespace com {
+
+void Communication::connectMasterSlaves(std::string const &name,
```
uekerman, line 7:
```suggestion
void Communication::connectMasterSlaves(std::string const &participantName,
```

fsimonis:
I rebased on develop, migrated TestContext and renamed the parameter.

Issue #705, opened 2020-03-26
Refactor type alias Bounding Box into class 


uekerman:
What once started as a small addition

https://github.com/precice/precice/blob/6cb67dd1478b79aff87cfcc12e63df19ad9dca9f/src/mesh/Mesh.hpp#L38

has now actually much associated functionality:

(0) computation in 
https://github.com/precice/precice/blob/6cb67dd1478b79aff87cfcc12e63df19ad9dca9f/src/mesh/Mesh.cpp#L210

(1) merging in
https://github.com/precice/precice/blob/6cb67dd1478b79aff87cfcc12e63df19ad9dca9f/src/partition/ReceivedPartition.cpp#L461

(2) intersection in 
https://github.com/precice/precice/blob/6cb67dd1478b79aff87cfcc12e63df19ad9dca9f/src/partition/ReceivedPartition.cpp#L448

(3) testing whether it contains a specific vertex
https://github.com/precice/precice/blob/6cb67dd1478b79aff87cfcc12e63df19ad9dca9f/src/partition/ReceivedPartition.cpp#L507

(4) an own communication wrapper class
https://github.com/precice/precice/blob/develop/src/com/CommunicateBoundingBox.hpp

It's time to bundle most of this functionality ( (1)-(3) yes, (4) no, (0)? my first suggestion) in a separate class within the `mesh` package.
Which relation to implement to `Mesh` needs careful consideration, meaning does a `BoundingBox` have a reference to a `Mesh`?

fsimonis:
To simplify sending and receiving bounding boxes, it would be helpful to implement it in terms of  `std::vector<double>` (instead of pairs) and provide access to the underlying data via
```
double const * BoundingBox::data() const;
```
Also creating a BoundingBox from a `std::vector<double>` would be handy for receiving. Maybe in some static factory function:
```
static BoundingBox fromData(std::vector<double> d);
```

To simplify logging, we should also provide a
```
std::ostream& operator<<(std::ostream & out, Boundingbox const & val);
```

In addition to that, the dimension will always be 2 or 3, thus the vector will never surpass 6 elements. So if we save `int dimension` + `std::array<double. 6>` the size of an BoundingBox object is only 52 Bytes. This would allow for a very dense representation and further simplify communication.

oguzziya:
[...]

PR #712, opened 2020-04-04
Refactoring Bounding Box

oguzziya:
This PR is related to #705, currently under progress


oguzziya:
Currently, all the usage of the `BoundingBox` in whole code is changed. Some cleaning, removal of unused parts, fixing of undefined behaviors and documentation will be done. The current situation of the tests as follows:
```
The following tests FAILED:
	  3 - precice.com (Timeout)
	 13 - precice.partition (Failed)
	 16 - precice.parallel (Timeout)
```
It seems that there is a problem while sending and receiving the `BoundingBox`. While overall cleaning, it will be solved also.

fsimonis:
You can run the tests suites individually with `ctest -V -R precice.com`. The log level can be controlled with `export BOOST_TEST_LOG_LEVEL=all`.

Communicate mesh probably broke during the refactoring.

oguzziya:
Currently receiving and sending bounding boxes works and individual tests pasts. However, I think there is a deadlock while broadcasting a `BoundingBox`, currently I am trying to solve this issue.

uekerman:
Just discussed with @oguzziya on the phone: 
I think it would be better to not offer the direct modification options `setMin`, `setMax` and `setBounds`. It could be better to make modification options more narrow, but this way also more explicit, e.g. through functions such as `enlarge` or a specific constructor for tests.
For all use case in `partition`, this should be possible.

@fsimonis Can you help with the computations in `RTreeAdapter`. How could we avoid the direct modifications there?

src/com/CommunicateBoundingBox.cpp
```diff
@@ -117,11 +111,11 @@ void CommunicateBoundingBox::broadcastReceiveBoundingBoxMap(
...
 
-  for (auto &rank : bbm) {
-    for (auto &dimension : rank.second) {
-      _communication->broadcast(dimension.first, 0);
-      _communication->broadcast(dimension.second, 0);
-    }
+  std::vector<double> receivedData;
+
+  for (int i = 0; i < sizeOfReceivingMap; ++i) {
+    _communication->receive(receivedData,0);
```
uekerman, line 117:
```suggestion
    _communication->broadcast(receivedData,0);
```

oguzziya:
With the help of @uekerman, broadcoasting bug is fixed and currently all the tests are running.
To Do:

- [ ] Remove/Replace the usage of set funcitons for the bounds

oguzziya:
Other than usage of `setMin` and `setMax` in `RTree`, all other `setBounds`functions are removed and for the tests, `modifyForTest`function is introduced. I am still trying to find a way to remove 'setMin` and `setMax`.

fsimonis:
@oguzziya It is probably easier and cleaner to use `BoundingBox::isVertexInBB()` in conjunction with the boost geometry [`statisfies`](https://www.boost.org/doc/libs/1_65_1/libs/geometry/doc/html/geometry/spatial_indexes/queries.html#geometry.spatial_indexes.queries.user_defined_unary_predicate).
This also decouples our code further from boost geometry (no pun intended).

oguzziya:
After struggling a little bit with `boost::geometry` refactoring is now complete. All the transparent set functions are removed and modification of `BoundingBox` is done under the hood.

src/com/CommunicateBoundingBox.cpp
```diff
@@ -54,6 +50,7 @@ void CommunicateBoundingBox::receiveBoundingBoxMap(
   int sizeOfReceivingMap;
   _communication->receive(sizeOfReceivingMap, rankSender);
 
+  PRECICE_CHECK(sizeOfReceivingMap == (int) bbm.size(), "Incoming size of map is not compatible");
   PRECICE_ASSERT(sizeOfReceivingMap == (int) bbm.size());
```
uekerman, line 54:
```suggestion
  PRECICE_ASSERT(sizeOfReceivingMap == (int) bbm.size(), "Incoming size of map is not compatible");
```
This should be an assertion. A user of preCICE should not be able to reach that state. It's merely to safeguard against bugs in preCICE.

src/com/CommunicateMesh.hpp
```diff
@@ -31,12 +31,12 @@ class CommunicateMesh {
...
-      const mesh::Mesh::BoundingBox &bb,
-      int                            rankReceiver);
+      const mesh::BoundingBox &bb,
+      int                     rankReceiver);
 
   void receiveBoundingBox(
-      mesh::Mesh::BoundingBox &bb,
-      int                      rankSender);
+      mesh::BoundingBox &bb,
+      int               rankSender);
```
uekerman, line 39: Why are both these functions needed at all? Couldn't they be replaced with calls to the identical functions of `CommunicateBoundingBox`?

src/mesh/BoundingBox.hpp
```diff
@@ -0,0 +1,96 @@
+#pragma once
+#include "logging/Logger.hpp"
+#include "mesh/Vertex.hpp"
+#include <vector>
+
+namespace precice {
+namespace mesh {
+
+/// Bounding box class for mesh and partions
```
uekerman, line 9:
```suggestion
/// An axis-aligned bounding box around a (partition of a) mesh
```

src/mesh/BoundingBox.hpp
```diff
@@ -0,0 +1,96 @@
...
+  * @param[in] dimension Dimension of the bounding box
+  * 
+  */
+  BoundingBox(int dimension);
+
+  /// Default constructor
+  BoundingBox();
+
+  /// Special function for tests to create custom bounds
+  void modifyForTest(int rank, std::string testName);
```
uekerman, line 35: This construction looks a bit odd. Why not directly using the constructor `BoundingBox(std::vector<double> bounds)` or `createFromData` in the tests?

src/mesh/BoundingBox.cpp
```diff
@@ -0,0 +1,171 @@
...
+}
+
+BoundingBox::BoundingBox()
+{
+  _dimensions = 3;
+  for(int i = 0; i < _dimensions; ++i){
+    _bounds.push_back(std::numeric_limits<double>::max());
+    _bounds.push_back(std::numeric_limits<double>::lowest());
+  }
+}
```
uekerman, line 29: This one should not be needed.

src/mesh/BoundingBox.cpp
```diff
@@ -0,0 +1,171 @@
+#include "BoundingBox.hpp"
+
+namespace precice {
+namespace mesh {
+
+BoundingBox::BoundingBox(std::vector<double> bounds)
+{
+  //PRECICE_CHECK(bounds.size() == 4 || bounds.size() == 6, "Dimension of the bounding box should be 2 or 3.");
```
uekerman, line 8:
```suggestion
  PRECICE_ASSERT(bounds.size() == 4 || bounds.size() == 6, "Dimension of a bounding box can only be 2 or 3.");
```

src/mesh/BoundingBox.cpp
```diff
@@ -0,0 +1,171 @@
...
+}
+
+BoundingBox::BoundingBox(int dimension)
+:_dimensions(dimension)
+{
+  for(int i = 0; i < _dimensions; ++i){
+    _bounds.push_back(std::numeric_limits<double>::max());
+    _bounds.push_back(std::numeric_limits<double>::lowest());
+  }
+}
```
uekerman, line 20: Let's also add an assertion here that dimension has to be 2 or 3.

src/mesh/BoundingBox.cpp
```diff
@@ -0,0 +1,171 @@
...
+    out << "dim: " << d << " min: " << bb._bounds[2 * d] << ", max: " << bb._bounds[2 * d + 1] << "\n";
+  }
+  return out;
+}
+
+bool BoundingBox::operator==(const BoundingBox& otherBB) const
+{
+  if(_dimensions != otherBB._dimensions){
+    return false;
+  }
```
uekerman, line 53: Let's replace this with an assertion. Comparing bounding boxes with different dimensions should not happen.

src/mesh/BoundingBox.cpp
```diff
@@ -0,0 +1,171 @@
...
+  for(int i = 0; i < _dimensions; ++i){
+    if(std::abs(_bounds.at(i) - otherBB._bounds.at(i)) > 1e-6){
```
uekerman, line 55: We could use `math::NUMERICAL_ZERO_DIFFERENCE` see `math/differences.hpp`.
But there should also be a tolerance given through the boost testing. @fsimonis Does this one overwrite `math::NUMERICAL_ZERO_DIFFERENCE` or what would be the best approach here? I guess `==` is only needed in tests.

src/mesh/BoundingBox.cpp
```diff
@@ -0,0 +1,171 @@
...
+      _bounds[2*i + 1] = i + 1;
+    }
+  }
+  if(testName == "partition"){
+    for (int i = 0; i < _dimensions; i++) {
+      _bounds[2*i] = 3 - rank - 1;
+      _bounds[2*i + 1] = 3 - rank;
+    }
+  }
+}
```
uekerman, line 77: Not needed, see above

src/mesh/BoundingBox.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+bool BoundingBox::isVertexInBB(const mesh::Vertex &vertex) const
+{
+  for (int d = 0; d < _dimensions; d++) {
+    if (vertex.getCoords()[d] < _bounds.at(2 * d) || vertex.getCoords()[d] > _bounds.at(2 * d + 1)) {
+      return false;
+    }
+  }
+  return true;
+}
```
uekerman, line 97: Let's add an assertion that the vertex and the bb have the same dimension.

src/mesh/BoundingBox.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+double BoundingBox::getData(int dimension, int type) const
+{
+  return _bounds.at(2*dimension + (type - 1));
+}
```
uekerman, line 102: I think this should not be needed. All functionality could be move to `BoundingBox`, such as `Mesh::getCOG`.

src/mesh/BoundingBox.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+const double* BoundingBox::data() const
+{
+  return _bounds.data();
+}
+
+const std::vector<double> BoundingBox::dataVector() const
+{
+  return _bounds;
+}
```
uekerman, line 117: Why do we need both?

src/mesh/BoundingBox.cpp
```diff
@@ -0,0 +1,171 @@
...
+    if (_bounds.at(2 * d + 1) > _bounds.at(2 * d))
+      maxSideLength = std::max(maxSideLength, _bounds[2 * d + 1] - _bounds[2 * d]);
+  }
+  for (int d = 0; d < _dimensions; d++) {
+    _bounds.at(2 * d + 1) += _safetyFactor * maxSideLength;
+    _bounds.at(2 * d) -= _safetyFactor * maxSideLength;
+    PRECICE_DEBUG("Merged BoundingBox" << *this);
+  }
+  return true;
+}
```
uekerman, line 141: Can we split this in two functions. `expandTo(const BoundingBox &otherBB)` and `addSafetyMargin(double safetyFactor)`?
* `safetyFactor` could be passed as an argument here instead of the private member. Maybe that's easier to understand.
* I don't like the name `mergeBoundingBoxes` as it indicates a symmetric operation. But actually `otherBB` is not touched.

src/mesh/BoundingBox.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+void BoundingBox::expandTo(const Vertex& vertices)
+{
+  for(int d = 0; d < _dimensions; ++d){
+    _bounds.at(2*d) = std::min(vertices.getCoords()[d], _bounds.at(2*d));
+    _bounds.at(2*d+1) = std::max(vertices.getCoords()[d], _bounds.at(2*d+1));
+  }
+}
```
uekerman, line 149: Let's add an assertion to test that the dimensions match

src/mesh/BoundingBox.cpp
```diff
@@ -0,0 +1,171 @@
...
+
+void BoundingBox::enlargeWith(double value)
+{
+  for (int d = 0; d < _dimensions; d++) {
+    _bounds[2*d] -= value;
+    _bounds[2*d + 1] += value;   
+  }
+}
```
uekerman, line 157: See my suggestion at `BoundingBox::mergeBoundingBoxes`

src/mesh/BoundingBox.hpp
```diff
@@ -0,0 +1,96 @@
...
+
+  /// Copy Constructor
+  BoundingBox(const BoundingBox &bb);
+
+  /// Destructor
+  ~BoundingBox();
```
uekerman, line 41: I guess this is added by default. So, normally we don't make it explicit, @fsimonis correct?

src/partition/ReceivedPartition.cpp
```diff
@@ -458,19 +443,6 @@ void ReceivedPartition::compareBoundingBoxes()
...
-  for (size_t i = 0; i < currentBB.size(); i++) {
-    if ((currentBB[i].first < receivedBB[i].first && currentBB[i].second < receivedBB[i].first) ||
-        (receivedBB[i].first < currentBB[i].first && receivedBB[i].second < currentBB[i].first)) {
-      return false;
-    }
-  }
-  return true;
-}
-
 void ReceivedPartition::prepareBoundingBox()
```
uekerman, line 446: Maybe we don't need that function anymore. All could just happen inline. Should be easier to understand.

src/partition/tests/ProvidedPartitionTest.cpp
```diff
@@ -360,38 +360,34 @@ BOOST_AUTO_TEST_CASE(TestCompareBoundingBoxes2D)
...
+    BOOST_TEST(receivedGlobalBB[0].getData(1,1) == 0);
+    BOOST_TEST(receivedGlobalBB[0].getData(1,2) == 3);
+    BOOST_TEST(receivedGlobalBB[1].getData(0,1) == 0);
+    BOOST_TEST(receivedGlobalBB[1].getData(0,2) == 1);
+    BOOST_TEST(receivedGlobalBB[1].getData(1,1) == 3.5);
+    BOOST_TEST(receivedGlobalBB[1].getData(1,2) == 4.5);
+    BOOST_TEST(receivedGlobalBB[2].getData(0,1) == 2.5);
+    BOOST_TEST(receivedGlobalBB[2].getData(0,2) == 4.5);
+    BOOST_TEST(receivedGlobalBB[2].getData(1,1) == 5.5);
+    BOOST_TEST(receivedGlobalBB[2].getData(1,2) == 7.0);
```
uekerman, line 391: Let's compare the complete bounding box.

src/partition/tests/ProvidedPartitionTest.cpp
```diff
@@ -484,24 +477,24 @@ BOOST_AUTO_TEST_CASE(TestSendBoundingBoxes3D)
...
+    BOOST_TEST(receivedGlobalBB[1].getData(1,1) == 3.5);
+    BOOST_TEST(receivedGlobalBB[1].getData(1,2) == 4.5);
+    BOOST_TEST(receivedGlobalBB[1].getData(2,1) == 0);
+    BOOST_TEST(receivedGlobalBB[1].getData(2,2) == 1);
+    BOOST_TEST(receivedGlobalBB[2].getData(0,1) == 2.5);
+    BOOST_TEST(receivedGlobalBB[2].getData(0,2) == 4.5);
+    BOOST_TEST(receivedGlobalBB[2].getData(1,1)== 5.5);
+    BOOST_TEST(receivedGlobalBB[2].getData(1,2) == 7.0);
+    BOOST_TEST(receivedGlobalBB[2].getData(2,1) == 2.5);
+    BOOST_TEST(receivedGlobalBB[2].getData(2,2) == 4.5);
```
uekerman, line 497: Let's compare the complete bounding box.

src/mesh/BoundingBox.cpp
oguzziya: As I see `BoundingBoxMap` which is `std::map` requires a default constructor for the object. I tried to find a way but I could not find yet.

uekerman: @fsimonis What would be good practice here?

fsimonis: The default constructor is only required if we use `std::map::operator[]`, as it default-constructs missing entries.
To fix that, we need to use `.at(key)` for lookup and `.emplace(key, value)` for insertion. ( Which makes the code more readable/expressive too. )

oguzziya: Oh, nice then. I will modify the BoundingBoxMap insertions then. Thank you!

src/mesh/BoundingBox.hpp
```diff
@@ -0,0 +1,92 @@
...
+  static BoundingBox createFromData(std::vector<double> bounds);
+
+  /// Expand bounding box using another bounding box
+  void expandTo(const BoundingBox &otherBB);
+
+  /// Expand bounding box using vertices
+  void expandTo(const Vertex& vertices);
+
+  /// Expand bounding box using value
+  void expandTo(double value);
```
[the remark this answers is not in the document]
oguzziya: That's true, we discussed and it should be expandBy as you said

src/mesh/BoundingBox.cpp
```diff
@@ -0,0 +1,152 @@
...
+BoundingBox::BoundingBox(const BoundingBox &bb)
+{
+  _dimensions = bb._dimensions;
+  _bounds     = bb._bounds;
+}
+
+std::ostream &operator<<(std::ostream &out, const BoundingBox &bb)
+{
+  for (int d = 0; d < bb._dimensions; ++d) {
+    out << "dim: " << d << " min: " << bb._bounds[2 * d] << ", max: " << bb._bounds[2 * d + 1] << "\n";
```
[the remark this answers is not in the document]
uekerman: > Newlines in a single log statement will lead to problems.

Yes, good point.

Why not just `([-5.0,2.3], [-10.0,5.4], [3.0,6.0])`

src/mesh/BoundingBox.hpp
```diff
@@ -0,0 +1,92 @@
...
+  /// Checks whether two bounding boxes are overlapping
+  bool overlapping(const BoundingBox &otherBB);
+
+  /**
+   * @brief Returns the Center Of Gravity of the mesh
+   *
+   * Returns a vector of doubles, size d, each dimension computed as
+   * cog =  (max - min) / 2 + min
+   */
+  std::vector<double> getCOG() const;
```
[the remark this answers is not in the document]
uekerman: Yes, sounds good

src/mesh/BoundingBox.hpp
```diff
@@ -0,0 +1,92 @@
+#pragma once
+#include "logging/Logger.hpp"
+#include "mesh/Vertex.hpp"
+#include <vector>
+
+namespace precice {
+namespace mesh {
+
+/// An axis-aligned bounding box around a (partition of a) mesh
+class BoundingBox {
```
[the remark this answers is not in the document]
uekerman: No, I am actually against this, see comments further above.

src/mesh/BoundingBox.hpp
```diff
@@ -0,0 +1,92 @@
...
+  void expandTo(const BoundingBox &otherBB);
+
+  /// Expand bounding box using vertices
+  void expandTo(const Vertex& vertices);
+
+  /// Expand bounding box using value
+  void expandTo(double value);
+
+  /// Increase the size of bounding box by safety margin
+  void addSafetyMargin(double safetyFactor);
```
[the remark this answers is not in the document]
uekerman: Yes, and margin sounds a lot like an addition, where it is a scaling actually.

src/mesh/BoundingBox.cpp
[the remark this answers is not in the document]
oguzziya: Way more cleaner :+1:

src/partition/tests/ProvidedPartitionTest.cpp
```diff
@@ -360,38 +360,30 @@ BOOST_AUTO_TEST_CASE(TestCompareBoundingBoxes2D)
     BOOST_TEST(context.isNamed("NASTIN"));
 
     mesh::Mesh::BoundingBoxMap receivedGlobalBB;
-    mesh::Mesh::BoundingBox    localBB;
+    mesh::BoundingBox    localBB{dimensions};
+
+    mesh::Mesh::BoundingBoxMap compareBB;
+    compareBB[0] = mesh::BoundingBox::createFromData({-1, 5, 0, 3});
+    compareBB[1] = mesh::BoundingBox::createFromData({0, 1, 3.5, 4.5});
+    compareBB[2] = mesh::BoundingBox::createFromData({2.5, 4.5, 5.5, 7.0});
```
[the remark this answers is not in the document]
oguzziya: Since I removed the constructor which takes `std::vector` and use only `createFromData() based on your previous feedback, I think it can stay as it is.

src/mesh/BoundingBox.cpp
```diff
@@ -0,0 +1,152 @@
...
+    cog[d] = (_bounds[2*d+1] - _bounds[2*d]) / 2.0 + _bounds[2*d];
+  }
+  return cog;
+}
+
+double BoundingBox::getArea(std::vector<bool> deadAxis){
+  double meshArea = 1.0;
+  for (int d = 0; d < _dimensions; d++)
+    if (not deadAxis[d])
+      meshArea *= _bounds[2*d + 1] -_bounds[2*d];
```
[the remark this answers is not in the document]
oguzziya: I added a default state checker and added assertion before these operations. It was what you mean right?

oguzziya:
- Dependency for default constructor is removed by changing functions used for `BoundingBoxMap`
- Constructor which takes `std::vector` is removed and all the functionilaty is preserved in static factor function `createFromData()` since it simplifies the creation most of times when compared to constructor.

src/mesh/BoundingBox.hpp
```diff
@@ -0,0 +1,92 @@
...
+   * Returns a vector of doubles, size d, each dimension computed as
+   * cog =  (max - min) / 2 + min
+   */
+  std::vector<double> getCOG() const;
+
+  /// Calculate the area of bounding box
+  double getArea(std::vector<bool> deadAxis);
+
+  /// Return data as std::vector
+  const std::vector<double> dataVector() const;
```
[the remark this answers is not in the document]
fsimonis: > currently using only the std::vector version, sending and receiving can be performed.

Receiving won't be possible with `data()` and `size()` due to the const specifier, but sending should work fine.
For receiving, you can create a `std::vector<double>`, receive to it and pass/move that to the BoundingBox constructor.

From an interface perspective, the only thing that is missing is the access to the BoundingBox data which is required for sending. Both `data()` and `dataVector()` provide this access, whereas the former is a minimal approach. `size()` is not even required as it simply returns `2*getDimensions()`.


@uekerman opinions on this?

src/mesh/BoundingBox.cpp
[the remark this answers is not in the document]
fsimonis: What you refer to as "uninitialized state" is actually well defined: `[ (max, min), (max, min), (max, min) ]`.
Adding additional state to a class is never a good idea.

Maybe you can add a function `bool empty() const` and use it as a guard ( `if (empty()) return 0;`) in some problematic functions.
Checking for emptiness is simply making sure that for all `d` the following holds: `_bounds[2*d] <= _bounds[2*d+1]`.

src/mesh/BoundingBox.hpp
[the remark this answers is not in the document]
fsimonis: I see. YAGNI, correct? :wink:

fsimonis:
> Constructor which takes std::vector is removed and all the functionialty is preserved in static factory function createFromData() since it simplifies the creation most of times when compared to constructor.

I fail to see you point. Why does a factory function simplify creation of an object?
Constructors are used to construct objects and throw exceptions when this fails.
Factory functions are useful if you want to run some functionality before calling the ctor such as caching. Or if you need virtual behaviour, but those points don't apply here.

Also constructing using a vector is quite simple: deduce the dimensions and `std::move` the vector into `_bounds`.

src/mesh/BoundingBox.hpp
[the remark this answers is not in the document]
uekerman: Yes, wasn't aware that there is a term for this, but now that I read it: yes :grin:

src/mesh/BoundingBox.hpp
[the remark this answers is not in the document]
uekerman: Yes, let's keep it minimal.

> For receiving, you can create a std::vector<double>, receive to it and pass/move that to the BoundingBox constructor.

Sounds good.

And for sending either `data()` or `dataVector`, no strong opinion which one, but not both.

oguzziya:
> Factory functions are useful if you want to run some functionality before calling the ctor such as > caching, or if you need virtual behaviour, but those points don't apply here.

Yes, you are right. I thought using factory function may save few lines in the code but using constructor is more robust and general idea. I will change this implementation to constructor.

src/mesh/BoundingBox.hpp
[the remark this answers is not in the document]
oguzziya: Okay, therefore I will switch it to `data()` then. Since it is not used other than the communication, I won't add the `size()` function, as Frederic mentioned, `getDimension()` should work fine.

fsimonis:
@oguzziya Sorry if I confused you by contradicting [my previous statement](https://github.com/precice/precice/issues/705#issuecomment-609719875).

The work looks good! Let's get this done and shipped.

oguzziya:
Some notes on latest implementations:
- Factory function is replaced with constructor
- Guards added to `meshArea()` , `center()` and `scaleBy()` functions

I struggle to change `std::vector<double> dataVector()` to `double* data()` and I wasted so much time trying to find what is wrong. Therefore, I did not change the implementation of `dataVector()` and related `receive()` and `send()` operations. In the current state of the pr, all tests except `mpiports` is are passed (I use VirtualBox, it can be the problem).

uekerman:
Looks good now :+1: 
One thing I did not realize earlier: we now, of course, need unit test for the bounding box. Sry for not pointing to this earlier :grin: 
Some of the test of `Mesh` could then also be moved there, e.g. COG.

`dataVector()` is OK. 
`mpiports` are known to fail sometimes for OpenMPI. @fsimonis is currently investigating this.

oguzziya:
Unit tests for `BoundingBox` is implemented. I think I covered all functionality in the class. However, I am not fully aware of all the use cases. Therefore if you have additional suggestions for unit tests, I can add them also.

uekerman:
Tests look good :+1:
Small things:
* `ExpandBy_BoundingBox` -> `ExpandByBoundingBox`. We strictly stick to CamelCase for our naming scheme. [More conventions](https://xgm.de/precice/docs/develop/conventions.html)
* Currently all tests work in 3D. Would be good to also have some 2D tests.

oguzziya:
For all cases 2D Tests are also added. Naming convention is fixed. Since the tests are relatively small, instead of creating new test functions, 2D tests are implemented in terms of scopes.

PR #724, opened 2020-04-14
Add error when participant uses mesh from itself

### LINKED ISSUE #488: Improve Error message for wrong configuration: Participant tries to use mesh from itself
The following configuration:

```
    <participant name="SolverTwo">
      <use-mesh name="MeshOne" from="SolverTwo"/>
      <use-mesh name="MeshTwo" provide="yes"/>
      <mapping:nearest-neighbor   direction="write" from="MeshTwo" to="MeshOne" constraint="conservative"/>
      <mapping:nearest-projection direction="read"  from="MeshOne" to="MeshTwo" constraint="consistent" />
      <write-data name="Velocities" mesh="MeshTwo" />
      <read-data  name="Forces"     mesh="MeshTwo" />
    </participant>
```

Note the second line `<use-mesh name="MeshOne" from="SolverTwo"/>`, which should read `<use-mesh name="MeshOne" from="SolverOne"/>`

results in an error message that is not really understandable:

```
preCICE: This is preCICE version 1.5.1
preCICE: Configuring preCICE with configuration: "precice-config.xml"
preCICE: Run in coupling mode
terminate called after throwing an instance of 'std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> >'
Aborted (core dumped)
```

## How to reproduce the error

* Modify [precice-config.xml](https://github.com/precice/precice/blob/dc33024404ef766f4a5c87bb8b2c1e50d3362ac9/tools/solverdummies/precice-config.xml#L30-L37) of the solverdummy, like shown above.
* Run (for example) the [python_future solver dummy](https://github.com/precice/precice/tree/develop/tools/solverdummies/python_future): `python3 python_future/solverdummy.py precice-config.xml SolverOne MeshOne` and `python3 python_future/solverdummy.py precice-config.xml SolverTwo MeshTwo`

fsimonis:
This PR adds a check to the `<use-mesh>` tag which verifies that a participant does not use a mesh `from=` itself.

Closes #488

### DIFF
--- a/src/precice/config/ParticipantConfiguration.cpp
+++ b/src/precice/config/ParticipantConfiguration.cpp
@@ -234,8 +234,11 @@ void ParticipantConfiguration::xmlTagCallback(
       stream << "Safety Factor must be positive or 0";
       throw std::runtime_error{stream.str()};
     }
-    bool          provide = tag.getBooleanAttributeValue(ATTR_PROVIDE);
-    mesh::PtrMesh mesh    = _meshConfig->getMesh(name);
+    bool provide = tag.getBooleanAttributeValue(ATTR_PROVIDE);
+    if (_participants.back()->getName() == from) {
+      PRECICE_CHECK(provide, "Participant \"" << context.name << "\" cannot use mesh \"" << name << "\" from itself. Use the \"from\"-field to specify which participant has to communicate the mesh to \"" << context.name << "\".");
+    }
+    mesh::PtrMesh mesh = _meshConfig->getMesh(name);
     if (mesh.get() == nullptr) {
       std::ostringstream stream;
       stream << "Participant \"" << _participants.back()->getName()
--- a/docs/changelog/724.md
+++ b/docs/changelog/724.md
@@ -1,0 +1,1 @@
+* Add check to prevent `<use-mesh>` from the same participant.


src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -233,6 +233,9 @@ void ParticipantConfiguration::xmlTagCallback(
       throw std::runtime_error{stream.str()};
     }
     bool          provide = tag.getBooleanAttributeValue(ATTR_PROVIDE);
+    if (_participants.back()->getName() == from) {
+      PRECICE_CHECK(provide, "Participant \"" << context.name << "\" cannot use Mesh \"" << name << "\" from itself" << from << ". This is likely a typo in the from-field.");
```
uekerman, line 237: I am so sure about the typo. Normally participants have rather different names. If somebody runs into this I guess he/she did not understand what we mean with "from". 

```suggestion
      PRECICE_CHECK(provide, "Participant \"" << context.name << "\" cannot use Mesh \"" << name << "\" from itself" << from << ". With \"from\" you can mark a mesh that will be communicated from another participant. .");
```

fsimonis: We could also give a hint on what to enter as `from`.
```suggestion
      PRECICE_CHECK(provide, "Participant \"" << context.name << "\" cannot use Mesh \"" << name << "\" from itself. Use \"from\" to mark a mesh provided by another participant.");                         
```

This argument should really not be necessary as the provider is explicitly configured.

uekerman: > This argument should really not be necessary as the provider is explicitly configured.

Yes, good point. This is more involved to get though. Actually, we only have two states, provided or received. So, the config could be changed (eventually) to 
```xml
<use-mesh name="MyMesh" source="provided/received" />
```
or a different name for "source". Would be a breaking change though.

fsimonis: @uekerman Which version do you prefer?

uekerman: No strong opinion. I think some "communicated" in there helps to understand the concept.

fsimonis: I rephrased the error message

Issue #698, opened 2020-03-19
Discuss and improve error messages


uekerman:
I collected all error messages in preCICE (`PRECICE_CHECK`, `PRECICE_ERROR`, `runtime_error`). Not yet sure what to do with them exactly. Let's discuss here. I think we all agree that we need to improve things here.

The best way is probably to directly improve the error messages. Add more information, more context, next steps, etc. For some errors, we probably also need further information in the wiki.

Quite some errors need to be assertions instead. 

Let's also agree on a style and wording. 
* I suggest to define that error messages always have to end with a period ".". "!" is forbidden :smirk: 
* Let's not say "solver" or "accessor", but always "participant" or "this participant".
* Use data names instead of IDs etc.
* More things?
* Format:
  * 1 line (no `\n`)
  * "What you did wrong. Hint or solution to fix it."

Example:  
   ```cpp
  This participant does not provide Mesh "NodeCenters", but attempted to modify it. Please extend the use-mesh tag as follows <use-mesh name="NodeCenters" provide="yes" >.
```

Fur sure this needs to be a common effort (coding days?). We could use the list to check updates and assign developers to blocks.

_Note: Checkboxes in `<detail>` are broken._

# 1. Wrong configuration

## 1.1 Data and mesh config #728 @MakisH 

- [x] Data "MyData" cannot be created twice for mesh "MyMesh"!
- [x] Data "MyData" uses non-unique name!
- [x] Data "MyData" is not available in data configuration!
- [x] Unknown data type!
- [x] Data with name "MyData" is not available during configuration of mesh "MyMesh"

### 1.2.1 use-mesh, read-data, write-data

- [x] Participant "MyParticipant" cannot provide and receive mesh "MyMesh"!
- [x] Participant "MyParticipant" must either provide or receive the mesh "MyMesh"!
- [x] Participant "MyParticipant" cannot receive and provide mesh "MyMesh" at the same time!
- [x] Mesh "MyMesh" cannot be used twice by participant "MyParticipant"!
- [x] Participant "MyParticipant" has to use "MyMesh" in order to write data to it!
- [x] Participant "MyParticipant" has to use mesh "MyMesh" when writing data to it!
- [x] Participant "MyParticipant" has to use "MyMesh" in order to read data from it!
- [x] Data action of participant "MyParticipant" uses mesh which is not used by the participant
- [x] Participant "MyParticipant" defines watchpoint "MyWatchpoint" for mesh "MyMesh" which is not used by him!
[...]

PR #727, opened 2020-04-15
Error message rework 5 - origin or reason unknown

fsimonis:
This PR revises all errors in section 5 **except** the ones deep inside the cplschemes in order not to conflict with #689.

Examples:
```
// getMeshVertexIDsFromPositions
ERROR:  Unable to find a vertex on mesh "MeshOne" at position (0, 1). The request failed for query 2 out of 3.
```

Related to #698


fsimonis:
That's actually quite interesting.

In commit 893b6a0be694eb86c9bb7e3ee01c63d0d0a52e32, I pulled a check from `void ScaleByAreaAction::performAction(double fullDt)` up to the configuration.
The configuration tests now fail as they were implemented for 3D meshes, which is actually a wrong use-case.

This should probably be a lesson to us that failing early is good. :wink:

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -261,7 +261,7 @@ void BaseCouplingScheme::extrapolateData(DataMap &data)
       utils::shiftSetFirst(pair.second->oldValues, values);
     }
   } else {
-    PRECICE_ERROR("Called extrapolation with order != 1,2!");
+    PRECICE_ASSERT(false, "Extrapolation order is invalid.");
```
BenjaminRueth, line 264: Why `PRECICE_ASSERT(false, ...)` and not just `PRECICE_ERROR(...)`?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1454,8 +1464,7 @@ const mesh::Mesh &SolverInterfaceImpl::mesh(const std::string &meshName) const
       return *context->mesh;
     }
   }
-  PRECICE_ERROR("Participant \"" << _accessorName
-                                 << "\" does not use mesh \"" << meshName << "\"!");
+  PRECICE_ASSERT(false, "Participant \"" << _accessorName << "\" does not use mesh \"" << meshName << "\".");
```
BenjaminRueth, line 1467: Why `PRECICE_ASSERT(false,...)` and not just `PRECICE_ERROR(...)`?

src/cplscheme/BaseCouplingScheme.cpp
fsimonis: This case should have been caught in the configuration.
It is thus an precice-internal problem and  should trigger an assertions.

src/precice/impl/SolverInterfaceImpl.cpp
fsimonis: This is only used in integration tests and means that the setup is incorrect. A user will never run into this code path.
Thus I would say that this should be an assertion.

PR #728, opened 2020-04-15
Improve errors: Data and mesh config

MakisH:
```diff
- Data "MyData" cannot be created twice for mesh "MyMesh"!
+ Data "MyData" cannot be created twice for mesh "MyMesh". Please rename or remove one of the use-data tags with name "MyData".
```
```diff
- Data "MyData" uses non-unique name!
+ Data "MyData" has already been defined. Please rename or remove one of the data tags with name="MyData".
```
```diff
- Data "MyData" is not available in data configuration!
+ Data "MyData" is not defined. Please define a data tag with name="MyData".
```
```diff
- Unknown data type!
+ Unknown data type "typeName". Known data types: "scalar", "vector".
```

```diff
- Data with name "MyData" is not available during configuration of mesh "MyMesh"
+ Data with name "MyData" used by mesh "MyMesh" is not defined. Please define a data tag with name="MyData"
```

Related to #698.

### DIFF
--- a/src/mesh/config/MeshConfiguration.cpp
+++ b/src/mesh/config/MeshConfiguration.cpp
@@ -85,8 +85,9 @@ void MeshConfiguration::xmlTagCallback(
     }
     if (not found) {
       std::ostringstream stream;
-      stream << "Data with name \"" << name << "\" is not available during "
-             << "configuration of mesh \"" << _meshes.back()->getName() << "\"";
+      stream << "Data with name \"" << name << "\" used by "
+             << "mesh \"" << _meshes.back()->getName() << "\" is not defined. "
+             << "Please define a data tag with name=\"" << name << "\".";
       throw std::runtime_error{stream.str()};
     }
   }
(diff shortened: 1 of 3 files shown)


src/mesh/config/MeshConfiguration.cpp
```diff
@@ -109,7 +110,7 @@ void MeshConfiguration::addMesh(
         break;
       }
     }
-    PRECICE_CHECK(found, "Data " << dataNewMesh->getName() << " is not available in data configuration!");
+    PRECICE_CHECK(found, "Data " << dataNewMesh->getName() << " is not defined. Please define a data tag with name=\""<< dataNewMesh->getName() << "\".");
```
KyleDavisSA, line 113: Seems like this error won't be run into, but good to keep it? I always hit the error above first.

MakisH: I also don't understand why this check is here, but I also don't see any problem with keeping it. Let's keep it then?

MakisH: @KyleDavisSA @uekerman shall we keep this then?

MakisH:
I rebased on the latest develop. The files still need formatting, but maybe that would be confusing, as this would add also irrelevant changes in the diff.

src/mesh/config/MeshConfiguration.cpp
uekerman: Seems like this function is only called in tests. We could convert it to an assertion.

MakisH: I converted this to an assertion, thanks!

PR #730, opened 2020-04-15
Improve further API errors, Section 2.5

uekerman:


src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -578,16 +578,14 @@ void SolverInterfaceImpl::resetMesh(
     int meshID)
 {
   PRECICE_TRACE(meshID);
+  PRECICE_ASSERT(false, "At the moment, we can no longer and not yet again reset meshes.");
```
MakisH, line 581: - Why this assertion of `false` here?
- What does `we can no longer and not yet again` mean?

I assume this was for debugging purposes and should be removed, right?
```suggestion
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -821,9 +819,11 @@ void SolverInterfaceImpl::mapWriteDataFrom(
   PRECICE_VALIDATE_MESH_ID(fromMeshID);
   impl::MeshContext &   context        = _accessor->meshContext(fromMeshID);
   impl::MappingContext &mappingContext = context.fromMappingContext;
-  if (mappingContext.mapping.use_count() == 0) {
-    PRECICE_ERROR("From mesh \"" << context.mesh->getName()
-                                 << "\", there is no mapping defined");
+  if (not mappingContext.mapping) {
+    PRECICE_ERROR("You attempt to \"mapWriteDataFrom\" mesh "
```
MakisH, line 823: `this function` could maybe be more specific, or `mapWriteDataFrom` should be clear that is a preCICE-internal function:

```suggestion
    PRECICE_ERROR("You attempt to call \"mapWriteDataFrom\" on mesh "
```

even further we could write:
```suggestion
    PRECICE_ERROR("You attempt to call the preCICE-internal function \"mapWriteDataFrom\" on mesh "
```

As an alternative, maybe we could change the "You" subject with "preCICE", to clearly distinguish internal problems from problems originating from wrong API usage. Otherwise, a user could be quite confused if they see this (which normally they should not).

```suggestion
    PRECICE_ERROR("preCICE tried to call \"mapWriteDataFrom\" on mesh "
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -852,8 +852,10 @@ void SolverInterfaceImpl::mapReadDataTo(
   impl::MeshContext &   context        = _accessor->meshContext(toMeshID);
   impl::MappingContext &mappingContext = context.toMappingContext;
   if (mappingContext.mapping.use_count() == 0) {
-    PRECICE_ERROR("From mesh \"" << context.mesh->getName()
-                                 << "\", there is no mapping defined!");
+    PRECICE_ERROR("You attempt to \"mapReadDataTo\" mesh "
```
MakisH, line 855: Same as with `mapWriteDataFrom`.

src/precice/impl/SolverInterfaceImpl.cpp
uekerman: Currently, `resetMesh` is not allowed to be called and also not reachable from `SolverInterface`. 
"We can no longer and not yet again" means that this function was needed for dynamic meshes as implemented by Bernhard, for which we dropped support for the sake of simpler parallelization. And it will presumably be needed again once we have "parallel" dynamic meshes.

src/precice/impl/SolverInterfaceImpl.cpp
uekerman: This is not an internal function. It is part of the API and the error indicates wrong usage of the API together with the configuration.

uekerman:
> We could maybe use a different wording for user-caused errors (i.e. an adapter called an API function wrongly) and for preCICE-internal errors (i.e. we ended up in an code part we shouldn't), reflecting the PRECICE_CHECK and PRECICE_ERROR, to not confuse users.

Looks like there is a misunderstanding. `PRECICE_ERROR` falls within the same category as `PRECICE_CHECK`. It is a real error we want to catch. Not a bug as in `PRECICE_ASSERT`.
Otherwise, I agree. `PRECICE_ASSERT` can follow a different language. Most of them don't even have a message, which I guess is fine. You anyway have to look into and understand the code.

src/precice/impl/SolverInterfaceImpl.cpp
MakisH: So, the assertion wants to say "we shouldn't be here in the current preCICE version".

The "no longer" and "not yet again" I really understood as time of the simulation, not version on preCICE. Maybe this reply would be a nice comment to add next to the assertion.

uekerman: I removed the assertion altogether as still some tests depend on this function. Will be resolved as part of #225

PR #733, opened 2020-04-19
Improve participant config errors

MakisH:
Section 1.2: Participant config
- [x] 1.2.1 use-mesh, read-data, write-data
- [x] 1.2.2 mapping
- [x] 1.2.3 mapping:rbf

related to #698.


MakisH:
Note to future shelf: incorporate the changes from #763 and fix the missing closing quote in https://github.com/precice/precice/pull/763/files#diff-a58a08692da49eee461ca014ee83b5aaR517

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -257,7 +257,7 @@ MappingConfiguration::ConfiguredMapping MappingConfiguration::createMapping(
   } else if (direction == VALUE_READ) {
     configuredMapping.direction = READ;
   } else {
-    PRECICE_ERROR("Unknown direction type \"" << direction << "\"!");
+    PRECICE_ASSERT(false, "Unknown mapping direction type \"" << direction << "\". Please check the documentation for available options.");
```
MakisH, line 260: I am using a (doomed to fail) `PRECICE_ASSERT` here, as this is already checked by the XML parser. I kept the suggestion vague to avoid duplication (we may add more options in the future).

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -244,28 +243,19 @@ void ParticipantConfiguration::xmlTagCallback(
     std::string                                   from         = tag.getStringAttributeValue(ATTR_FROM);
     double                                        safetyFactor = tag.getDoubleAttributeValue(ATTR_SAFETY_FACTOR);
     partition::ReceivedPartition::GeometricFilter geoFilter    = getGeoFilter(tag.getStringAttributeValue(ATTR_GEOMETRIC_FILTER));
-    if (safetyFactor < 0) {
-      std::ostringstream stream;
-      stream << "Safety Factor must be positive or 0";
-      throw std::runtime_error{stream.str()};
-    }
+    PRECICE_CHECK(safetyFactor >= 0, "Participant \"" << context.name << "\" uses mesh \"" << name << "\" with safety-factor=\"" << safetyFactor << "\". Please use a positive or zero safety-factor instead.")
```
MakisH, line 246: I converted from the old system to `PRECICE_CHECK` here and below.

MakisH:
This is now ready for reviewing.

Note that the following two messages do not exist anymore, since we unified `rbf` and `petrbf`:
- `Using RBF data mappings in parallel requires building with PETSc.`
- `RBF mapping is not supported for a participant in master mode, use petrbf instead`

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -283,7 +274,8 @@ void ParticipantConfiguration::xmlTagCallback(
     mesh::PtrMesh mesh     = _meshConfig->getMesh(meshName);
     PRECICE_CHECK(mesh, "Participant "
                             << "\"" << _participants.back()->getName() << "\" has to use "
-                            << "mesh \"" << meshName << "\" in order to read data from it!");
+                            << "mesh \"" << meshName << "\" in order to read data from it. "
```
MakisH, line 277: Somehow this condition seems to not be reached. If I try to **read** from a mesh that is not used, I still get the error message for **writing**.

For example, in the C++ solver dummy:
```xml
      <use-mesh name="MeshOne" provide="yes"/>
      <write-data name="Forces"     mesh="MeshOne" />
      <read-data  name="Velocities" mesh="MeshTwo" />
    </participant>
```
gives (before, similarly now):
```
preCICE:ERROR:  Participant "SolverOne" has to use mesh "MeshTwo" when writing data to it!
```

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -244,28 +243,19 @@ void ParticipantConfiguration::xmlTagCallback(
...
-    if (mesh.get() == nullptr) {
-      std::ostringstream stream;
-      stream << "Participant \"" << _participants.back()->getName()
-             << "\" uses mesh \"" << name << "\" which is not defined";
-      throw std::runtime_error{stream.str()};
-    }
+    PRECICE_CHECK(mesh, "Participant \"" << _participants.back()->getName()
+                            << "\" uses mesh \"" << name << "\" which is not defined. "
+                            << "Please check the use-mesh node with name=\"" << name << "\" or define the mesh.");
     if ((geoFilter != partition::ReceivedPartition::GeometricFilter::ON_SLAVES || safetyFactor != 0.5) && from == "") {
```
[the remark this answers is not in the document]
uekerman: Yes, that's indeed tricky to refactor. 
But the edge case is not so critical. The meaningless safety-factor gets simply ignored.
Let's keep it as it is for the moment.

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -469,7 +472,8 @@ void ParticipantConfiguration::finishParticipantConfiguration(
     int toMeshID = dataContext.mesh->getID();
     PRECICE_CHECK(participant->isMeshUsed(toMeshID),
                   "Participant \"" << participant->getName() << "\" has to use mesh \""
-                                   << dataContext.mesh->getName() << "\" when writing data to it!");
+                                   << dataContext.mesh->getName() << "\" in order to write data to it. "
```
uekerman, line 475:
```suggestion
                                   << dataContext.mesh->getName() << "\" in order to read data from it. "
```

src/precice/config/ParticipantConfiguration.cpp
uekerman: That's due to the check in line 475 (new numbering), see below, I fixed it there.
This check here only fires if the mesh does not exist at all. So, the message is not wrong, but neither perfect.

MakisH: Good point, I did not notice that the previous one was already wrong and simply rephrased it. Thanks!

PR #736, opened 2020-04-20
Platform Specific Default for Loopback Name

### LINKED ISSUE #198: Compile-time changes to enable tests on macOS
We currently work on building and testing macOS on Travis (@shkodm, @JihoYang, is there an issue?). To ultimately get the tests working, we need to differentiate between different os at some point. One example is the m2n network where `lo` is the default for linux systems, but where we need `lo0` on macOS. Other cases might exist as well. (Please report here). 

The cleanest way to get this differentiation is to do this at compile time, meaning to change the network default depending on the system.

fsimonis:
This PR adds a platform-specific default name for the loopback interface.
Unsupported platforms will display a warning at compile-time and default to `""`.

Currently supported are:
* Windows (do nothing)
* Linux `lo`
* Apple `lo0`

Closes #198

### DIFF
--- a/src/utils/networking.cpp
+++ b/src/utils/networking.cpp
@@ -14,7 +14,7 @@ std::string loopbackInterfaceName()
   // Not required as we directly use the 127.0.0.1 under Windows
   return "";
 #else
-#warning "Your target architecture does not define a loopback interface. Please consider reporting this upstream."
+#warning "Your target architecture does not define a loopback interface. Please consider reporting this to the preCICE developers."
   return "";
 #endif
 }
--- a/src/utils/networking.cpp
+++ b/src/utils/networking.cpp
@@ -1,0 +1,24 @@
+#include "utils/networking.hpp"
+
+namespace precice {
+namespace utils {
+namespace networking {
+
+std::string loopbackInterfaceName()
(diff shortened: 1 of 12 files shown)


src/utils/networking.cpp
```diff
@@ -0,0 +1,24 @@
...
+{
+#if defined(__linux__)
+  return "lo";
+#elif defined(__APPLE__)
+  return "lo0";
+#elif defined(__WIN32)
+  // Not required as we directly use the 127.0.0.1 under Windows
+  return "";
+#else
+#warning "Your target architecture does not define a loopback interface. Please consider reporting this upstream."
```
MakisH, line 17: "upstream" could be ambiguous here: is it preCICE or some dependency? What if we use the "preCICE developers" here?

```suggestion
#warning "Your target architecture does not define a loopback interface. Please consider reporting this to the preCICE developers."
```

fsimonis: I don't think that upstream is ambiguous as this warning will only be displayed when compiling preCICE itself.
That said, upstream is not a mainstream term (no pun intended). Your version is understandable by a wider audience. :+1:

Issue #718, opened 2020-04-09
Gather-scatter LU decomposition for parallel RBF mappings


uekerman:
**The Problem**

preCICE currently offers two ways how to solve an RBF mapping system:
* Using GMRES in every call to `map`. To this end, we use PETSc, which also works in parallel.
https://github.com/precice/precice/blob/develop/src/mapping/PetRadialBasisFctMapping.hpp
* Computing an LU decomposition in `computeMapping` and solving one forward and one backward substitution in every `map`. Here, we use Eigen. However, our implementation only works for a serial participant.
https://github.com/precice/precice/blob/develop/src/mapping/RadialBasisFctMapping.hpp

PETSc keeps being a tedious build dependency. Already now, it is optional, but then RBF mappings cannot be used in parallel. 
Furthermore, GMRES in every timestep is not necessarily the best option. Depending on the mesh size and the number of timesteps, LU might also be a valid alternative performance-wise.

**Suggested Solution**

Extend the LU-based Eigen implementation to also handle a parallel participant. This should be done, as a first variant, in a brute-force gather-scatter way:
* Gather both meshes during `computeMapping` and compute LU decomposition on master.
* Gather data during `map`, solve forward and backward substitution on master, scatter data.

Communication should be handled within `RadialBasisFctMapping`, also all global data structures. Meaning, the mesh on the master rank should not contain the global mesh.

Then, we could switch off PETSc as a build dependency by default and in all binaries and users with large cases could still switch it on if needed. `mapping:rbf` should use GMRES / PETSc if built with PETSc, LU / Eigen otherwise or if configured explicitly (so very similarly to how it is handled right now already): https://github.com/precice/precice/blob/develop/src/mapping/config/MappingConfiguration.cpp#L272-L376   

This will be a very slow and inefficient variant if larger meshes are used. This is however not the use case, so we could ignore this at first. Later, we could, however, also look into a proper parallel LU decomposition (such as discussed in the Parallel Numerics Lecture at TUM) and implement it with Eigen (no new dependency if possible!). But this could be formulated in another issue.

PR #739, opened 2020-04-20
Gather-scatter LU decomposition for parallel RBF mappings

oguzziya:
This is PR for issue #718 . Currently under progress


src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -232,65 +266,122 @@ void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::map(
...
-  Eigen::VectorXd &inValues  = input()->data(inputDataID)->values();
-  Eigen::VectorXd &outValues = output()->data(outputDataID)->values();
+  std::vector<double> inValues;
+  inValues.resize(input()->data(inputDataID)->values().size());
+  std::vector<double> outValues;
+  outValues.resize(output()->data(outputDataID)->values().size());
+
+  // Map the data from Eigen::VectorXd to std::vector for send and receive operations
+  Eigen::Map<Eigen::VectorXd>(&inValues[0], input()->data(inputDataID)->values().size())     = input()->data(inputDataID)->values();
+  Eigen::Map<Eigen::VectorXd>(&outValues[0], output()->data(outputDataID)->values().size())  = output()->data(outputDataID)->values();
```
fsimonis, line 276: You could directly construct the vector using the `data()` and `size()` accessors of `Eigen::VectorXd`.
Sadly Eigen does not provide `begin()` and `end()` consistently, so one has to pass pointers instead.

```cpp 
  auto & localInData  = input()->data(inputDataID)->values();
  auto & localOutData = output()->data(outputDataID)->values();
  std::vector<double> inValues(localInData.data(),   localInData.data()  + localInData.size());
  std::vector<double> outValues(localOutData.data(), localOutData.data() + localOutData.size());
```

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -232,65 +266,122 @@ void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::map(
...
 
-      Au  = _matrixA.transpose() * in;
-      out = _qr.solve(Au);
+  // Gather input data
+  if(utils::MasterSlave::isSlave()){
+    utils::MasterSlave::_communication->send(inValues, 0);
+  }
+  else {
+    globalInValues.resize(inValues.size());
+    globalInValues = inValues;
```
fsimonis, line 298: This resizes the vector and then replaces it, thus you can omit the `resize()`.

If you know the total size beforehand then you can `u.reserve(n)` the memory in one go and insert the values from the other vector using `u.insert(u.end(), v.begin(), v.end())`.

oguzziya:
With the latest commit, test case for homogeneous mesh distribution passes but the heteregenous cases fail.

oguzziya:
With the latest commit, consistent and serial cases passes. There is a distribution problem for conservative cases. Mapping calculates the the data correctly but in the distribution I have a unresolved issue. I will have a look at it.

oguzziya:
With the latest commit, all the tests other than tagging (not implemented yet) passes. I have made some assumptions and it would be really nice if they can be confirmed:

- For consistent mapping only the input mesh is distributed, output mesh is local. Therefore, for mesh gathering in `computeMapping()`, it is only enough to filter the input mesh. Vice versa for consistent.
- In consistent mapping since the input data is distributed, while gathering the data, input data is filtered. No filtering is applied for output data. In the conservative mapping, filtering is only applied in the scattering phase for output data.

Since I deduced them from tests, I am not sure about their correctness. So it would be really nice if @uekerman or @fsimonis could confirm them.

As the next step, I will implement mesh tagging and clean up the code a little bit since there ara some unnecessary copies I think.

oguzziya:
Mesh tagging tests are also working now.

uekerman:
For parallel use, we have the restrictions that a **consistent** mapping always needs to be a **read** mapping and a **conservative** mapping a **write** mapping. [More information](https://github.com/precice/precice/wiki/Mapping-Configuration#restrictions-for-parallel-participants). 
This means that the output mesh of a consistent mapping is "provided" (so distributed **without overlap**) and the input mesh is "received" (distributed possibly **with overlap**). For a conservative mapping, input and output are swapped.

I think this is what you mean, right?

uekerman:
Only integration tests with serial participants still fail, should be easy to fix.

oguzziya:
Yes, that's what I meant. And yes, I realized the failing integration test a little bit late, looking into it.

oguzziya:
Currently all the tests are passing.

oguzziya:
Parallel Eigen RBF is activated in the mapping configuration and integration tests. All the tests are passing but there is an interesting situation. 
At the end of the `computeMapping()` matrix is checked whether it is invertable or not. In the integration tests, this check throws an error. When I commented out this check, everything works fine and mapping is accomplished correctly. I am not sure why this is the case. I investigated the vertex coordinates for failing cases and I think they should result in an invertible matrix.

fsimonis:
@oguzziya Your latest commits are not linked to your GitHub account. Can you please check your git config and fix the commits before we merge the PR? Otherwise they won't be associated to you.

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -232,66 +289,230 @@ void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::map(
...
 
-      // Copy mapped data to ouptut data values
-      for (int i = 0; i < out.size(); i++) {
-        outValues(i * valueDim + dim) = out[i];
+      // Data scattering to slaves
+      int beginPoint = outValuesSize.at(0);
+
+      if (utils::MasterSlave::isMaster()) {
+        for (int rank = 1; rank < utils::MasterSlave::getSize(); ++rank) {
+          std::vector<double> sendValues(outputValues.data() + beginPoint, outputValues.data() + beginPoint + outValuesSize.at(rank));
```
fsimonis, line 503: We could reuse the buffer `sendValues` by moving it outside of the loop.

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -318,15 +539,71 @@ Eigen::VectorXd RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::reduceVector(
...
+  // Tags all vertices that are inside otherMesh's bounding box, enlarged by the support radius
+
+  if (_basisFunction.hasCompactSupport()) {
+
+    auto rtree    = mesh::rtree::getVertexRTree(filterMesh);
+    namespace bgi = boost::geometry::index;
+    auto bb       = otherMesh->getBoundingBox();
+    // Enlarge by support radius
+    bb.expandBy(_basisFunction.getSupportRadius());
+    rtree->query(bgi::satisfies([&](size_t const i) { return bb.contains(filterMesh->vertices()[i]); }),
```
fsimonis, line 564: The spatial rtree can execute an intersection query faster than an evaluating a predicate, but this is out of scope. I'll open an issue.

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -318,15 +539,71 @@ Eigen::VectorXd RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::reduceVector(
...
+                 boost::make_function_output_iterator([&filterMesh](size_t idx) {
+                   filterMesh->vertices()[idx].tag();
+                 }));
+  } else {
+    for (auto &vert : filterMesh->vertices())
+      vert.tag();
```
fsimonis, line 570: A function `void mesh::Mesh tagAll();` would be handy here.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -285,19 +285,19 @@ MappingConfiguration::ConfiguredMapping MappingConfiguration::createMapping(
 #endif
 
   bool createPetRBF = false;
+  bool createEigenRBF  = false;
```
fsimonis, line 288: I don't think an additional variable is required.

You could tidy this function up by introducing an enum though:
```cpp
enum struct RBFType { Eigen, PETSc };
RBFType rbftype = RBFType::Eigen;
// ...
if (rbftype == RBFType::Eigen) {
   // create eigen mapping ...
}
```

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -122,84 +126,137 @@ void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::computeMapping()
...
+        deadDimensions += 1;
     }
-    _matrixA(i, inputSize) = 1.0;
-    for (int dim = 0; dim < dimensions - deadDimensions; dim++) {
-      _matrixA(i, inputSize + 1 + dim) = reduceVector(iVertex.getCoords())[dim];
+    int polyparams = 1 + dimensions - deadDimensions;
+    PRECICE_ASSERT(inputSize >= 1 + polyparams, inputSize);
+    int n = inputSize + polyparams; // Add linear polynom degrees
+
+    // LU Decomposition
```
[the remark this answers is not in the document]
oguzziya: If they are member functions, is it necessary to return something? They can be modified inside the function directly

uekerman:
> At the end of the `computeMapping()` matrix is checked whether it is invertable or not. In the integration tests, this check throws an error. When I commented out this check, everything works fine and mapping is accomplished correctly. I am not sure why this is the case. I investigated the vertex coordinates for failing cases and I think they should result in an invertible matrix. If the resulting matrix was wrong, mapping would not work also, but this is not the case.

Sounds interesting indeed. Either we still have a bug here or the test setup is not valid. Which integration test fails exactly? Or are there multiple ones?

oguzziya:
Both 'GlobalRBFPartitioning' and 'LocalRBFPartitioning' tests fail in the parallel tests when the invertible check is opened. Since both of them fails, my suspection is about a bug in my implementation. However, it i interesting to get the correct results.

src/mapping/RadialBasisFctMapping.hpp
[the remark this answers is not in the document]
fsimonis: These functions can be implemented as non-member functions as they don't need access to member variables.

Non-member function have less scope (as in no object) and are therefore easier to understand, maintain and test. Without side-effects, you can treat them as pure functions and create a comprehensive test suite for them.
An additional benefit in this case is that they don't rely on the template. So you can even declare them in `RadialBasisFctMapping.hpp` and define them in `RadialBasisFctMapping.cpp`.

fsimonis:
@oguzziya If you [refactor the matrix generation methods](https://github.com/precice/precice/pull/739#discussion_r431754677) into a separate functions, then you can write small unit tests for your implementation. :wink:

src/mapping/RadialBasisFctMapping.hpp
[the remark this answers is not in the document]
fsimonis: :hand: I overlooked the evaluation of the basis function!

So, you have to pass that as a templated parameter too:
```
template<class BasisFunction>
Eigen::MatrixXd buildMatrixCLU(const precice::mesh::Mesh& in, BasisFunction func) {
  /// --- >8 ---
}

template<class BasisFunction>
Eigen::MatrixXd buildMatrixA(const precice::mesh::Mesh& in, const precice::mesh::Mesh& out, BasisFunction func) {
  /// --- >8 ---
}
```
So let's skip the move to `.cpp`.

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -122,84 +126,137 @@ void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::computeMapping()
...
+      }
+    }
+    // Fill _matrixA with values
+    for (int i = 0; i < outputSize; ++i) {
+      for (int j = 0; j < inputSize; ++j) {
+        difference = globalOutMesh.vertices()[i].getCoords();
+        difference -= globalInMesh.vertices()[j].getCoords();
+        _matrixA(i, j) = _basisFunction.evaluate(reduceVector(difference).norm());
+      }
+      _matrixA(i, inputSize) = 1.0;
```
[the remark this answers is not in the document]
oguzziya: I investigated the matrix and actually we not set the entire column at the end. So one option that I only find is block operations and I think it would be an overkill for this process. However, I moved it to one outer loop, at least it won't be set twice.

oguzziya:
Is it a meaningful approach putting stack trace information also into the functions `mapConservative()` and `mapConsistent()` ?

fsimonis:
@oguzziya Using `PRECICE_TRACE` for member functions is always a good idea especially as it is a noop in release builds.

src/mapping/RadialBasisFctMapping.hpp
fsimonis: Opened the issue: #762

fsimonis:
@uekerman All tests pass when building with and without PETSc. Are there additional problems?

uekerman:
> @uekerman All tests pass when building with and without PETSc. Are there additional problems?

They only pass because the invertible check is commented out for the moment. I'll have a closer look this week.

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -245,69 +238,321 @@ void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::map(
...
+  // Gather input data
+  if (utils::MasterSlave::isSlave()) {
+
+    // For conservative, data is not filtered
+    const auto &localInData  = input()->data(inputDataID)->values();
+    const auto &localOutData = output()->data(outputDataID)->values();
+
+    // Send data
+    utils::MasterSlave::_communication->send(localInData.data(), localInData.size(), 0);
+    utils::MasterSlave::_communication->send(localOutData.data(), localOutData.size(), 0);
```
uekerman, line 262: Why is `localOutData` also communicated here? That should not be needed, right?
Also: a conservative mapping in parallel is always a write mapping. It goes, thus, from a "provided" to a "received" mesh. Meaning, input is non-overlapping, **but output is**.

oguzziya:
The problem that @uekerman mentioned is solved. Apparently, there were too many unnecessary communication of the output data, these were fixed.

fsimonis:
@oguzziya please consider merging develop back into your branch as there have been some fixes regarding the PETSc RBF.

Issue #746, opened 2020-04-29
MPI Ports Communication Problems


fsimonis:
We keep on having problems with both the MPIPorts and the MPISinglePorts communication. The point of this issue of to document and find the root cause of the problems.

### `MPI_Open_port()`
Opens a port and returns a string referring to it.
[OpenMPI](https://www.open-mpi.org/doc/v3.0/man3/MPI_Open_port.3.php) [MPICH](https://www.mpich.org/static/docs/v3.2/www3/MPI_Open_port.html)

### `MPI_Close_port()`
Closes the port and releases owned resources.
[OpenMPI](https://www.open-mpi.org/doc/v3.0/man3/MPI_Close_port.3.php) [MPICH](https://www.mpich.org/static/docs/v3.2/www3/MPI_Close_port.html)

### `MPI_Comm_accept()`
Given a port and a local comm, this creates an intercommunicator to the requesting side.
This call is collective on the local comm.
[OpenMPI](https://www.open-mpi.org/doc/v3.0/man3/MPI_Comm_accept.3.php)[MPICH](https://www.mpich.org/static/docs/v3.2/www3/MPI_Comm_accept.html)

### `MPI_Comm_connect()`
Given a port and a local comm, this creates an intercommunicator to the accepting side.
This call is collective on the local comm and requires that `MPI_Comm_accept()` has been called on all ranks of the requesting communicator.
[OpenMPI](https://www.open-mpi.org/doc/v3.0/man3/MPI_Comm_connect.3.php) [MPICH](https://www.mpich.org/static/docs/v3.2/www3/MPI_Comm_connect.html)

### `MPIPortsCommunication`

Similar to the `SocketCommunication`, this directly connect peers 1-1.
It uses `MPI_COMM_SELF` on the accepting and connecting side of the calls, creating one Communicator with size `1` (the other peer) per connection. As we use `MPI_COMM_SELF`, this never leads to problems as there are essentially no collective calls. This is thus a very robust approach given that the actual underlying communicator are always ignored. Also Master-Master connections, Master-Slave connections and Client-Server connections use the same implementation.

### `MPISinglePortsCommunication`

In contrast to `MPIPortsCommunication` this communication uses a single communicator for the Client-Server communication. It still needs to rely on 1-1 connections for the Master-Master communication.
The Master-Slave connections are currently implement as 1-1 connections, but could use a single communicator by splitting the base comm into MasterComm and SlavesComm. This is currently not implemented.

[...]

PR #702, opened 2020-03-26
Redesign testing fixtures

fsimonis:
This PR is the complete redesign of the testing fixtures to cover all the cases described in #597.
The primary goal is to generate clean and consistent environments for each test. There should be close to no inter-test dependencies.

## Goals
The secondary goals of this PR are:
1. to simplify the notation of tests. Especially the notation of integration tests were highly obscure due to the required setup.
2. to reduce the setup and error checks to be done by the developer.

## Changes
To achieve said goals, I had to:
* Introduce a custom notation for tests that cover integration tests as well as all unit test cases described in #597.
* Redesign the way preCICE handles MPI communicators. This lead to a complete overhaul of `utils/Parallel.h`. It now uses shared linked `CommStates` that free communicators when required.
* Redesign of `MPIDirectCommunciation`, which is now only usable for participants inside a shared communicator. This required a slight modification of `com::Communication`. I also added tests for all communication routines and cases.
* Add a TestContext helper type which exposes all required information to a running test (Name, rank, size, comm) and allows to connect participants.
* Remove the majority of test fixtures and decorators as they are now super-seeded by the TestContext.
* Port and fix all tests.

Additional decisions reflected in this PR:
* Some tests one setup with different configurations. I broke the tests down into tests running one configuration each. This simplifies debugging and improves error messages.
* Tests now require to be run on 4 MPI Ranks and will be disabled when configured without `PRECICE_MPICommunication`. Running tests on less than 4 ranks covers a miniscule part of the library. The tests will still compile without MPI, thus testing compile-time information.
* The solverdummies (c, cpp, fortran) are now build against precice as additional tests. Further test run coupling cases with the solverdummies. This ensures that we don't break the solverdummies and provides some level of testing for preCICE build without MPI.
* Ranks in tests now synchronize at the beginning of the setup, at the end of the setup (individual participants) and at the end of the test. This slows down tests, but improves the resulting logs.

The notation for tests looks as follows:

```cpp
// Integration test
BOOST_AUTO_TEST_CASE(LocalRBFPartitioning)
{
  PRECICE_TEST("SolverOne"_on(3_ranks), "SolverTwo"_on(1_rank));
  // ...
  SolverInterface interface(context.name, configFilename, context.rank, context.size);
  // ...
  if (context.name == "SolverOne") { //...
    if (context.isMaster()) {
      //...
    } else if (context.isRank(1)) {
      //...
    } else {
      //...
    }
  }
}

[...]

PR #747, opened 2020-04-30
Rework MPI[Single]Ports and add Tests

fsimonis:
Adding the MPISinglePorts tests revealed some problems, which required to finally migrate the tests, which revealed some issues in the acceleration and com packages.

This PR will:
* Add tests for MPISinglePorts similar to #624 
* Refactor MPISinglePorts to distinguish between direct and global communicators.
* Fix not properly closed ports in MPIPorts and MPISinglePorts
* Migrate the tests to the context-bases system #702 
* Add a bypass to SocketCommunication to prevent work
* Detect the MPI implementation in CMake
* Disable mpiports tests when OpenMPI was detected #746 
* Refactors the ownership of cyclic comms from MVQN into ParallelMatrixOperations

### DIFF
--- a/src/acceleration/MVQNAcceleration.cpp
+++ b/src/acceleration/MVQNAcceleration.cpp
@@ -1,3 +1,4 @@
+#include "utils/assertion.hpp"
 #ifndef PRECICE_NO_MPI
 
 #include <Eigen/Core>
@@ -10,7 +11,6 @@
 #include "acceleration/impl/QRFactorization.hpp"
 #include "com/Communication.hpp"
 #include "com/MPIPortsCommunication.hpp"
-#include "com/SocketCommunication.hpp"
 #include "cplscheme/CouplingData.hpp"
 #include "mesh/Mesh.hpp"
 #include "mesh/Vertex.hpp"
@@ -48,8 +48,6 @@ MVQNAcceleration::MVQNAcceleration(
       _matrixV_RSLS(),
       _matrixW_RSLS(),
       _matrixCols_RSLS(),
-      _cyclicCommLeft(nullptr),
-      _cyclicCommRight(nullptr),
       _parMatrixOps(nullptr),
       _svdJ(RSSVDtruncationEps, preconditioner),
       _alwaysBuildJacobian(alwaysBuildJacobian),
@@ -67,22 +65,6 @@ MVQNAcceleration::MVQNAcceleration(
 // ==================================================================================
 MVQNAcceleration::~MVQNAcceleration()
 {
-  //if(utils::MasterSlave::isMaster() ||utils::MasterSlave::isSlave()){ // not possible because of tests, MasterSlave is deactivated when PP is killed
-
-  // close and shut down cyclic communication connections
-  if (not _imvjRestart) {
-    if (_cyclicCommRight != nullptr || _cyclicCommLeft != nullptr) {
-      if ((utils::MasterSlave::getRank() % 2) == 0) {
-        _cyclicCommLeft->closeConnection();
-        _cyclicCommRight->closeConnection();
-      } else {
-        _cyclicCommRight->closeConnection();
-        _cyclicCommLeft->closeConnection();
-      }
-      _cyclicCommRight = nullptr;
-      _cyclicCommLeft  = nullptr;
-    }
-  }
 }
 
 // ==================================================================================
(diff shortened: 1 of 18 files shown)


src/com/MPIPortsCommunication.cpp
```diff
@@ -47,20 +47,21 @@ void MPIPortsCommunication::acceptConnection(std::string const &acceptorName,
...
     // Connection
     MPI_Comm communicator;
     MPI_Comm_accept(const_cast<char *>(_portName.c_str()), MPI_INFO_NULL, 0, MPI_COMM_SELF, &communicator);
     PRECICE_DEBUG("Accepted connection at " << _portName << " for peer " << peerCurrent);
 
+    // Which rank is requesting a connection?
     int requesterRank = -1;
-    // Exchange information to which rank I am connected and which communicator size on the other side
     MPI_Recv(&requesterRank, 1, MPI_INT, 0, 42, communicator, MPI_STATUS_IGNORE);
+    // Who big is the communicator of the requester
```
uekerman, line 61:
```suggestion
    // How big is the communicator of the requester
```

src/com/MPIPortsCommunication.cpp
```diff
@@ -89,7 +94,7 @@ void MPIPortsCommunication::acceptConnectionAsServer(std::string const &acceptor
                                                      int                requesterCommunicatorSize)
 {
   PRECICE_TRACE(acceptorName, requesterName, acceptorRank, requesterCommunicatorSize);
-  PRECICE_CHECK(requesterCommunicatorSize > 0, "Requester communicator size has to be > 0!");
+  PRECICE_ASSERT(requesterCommunicatorSize >= 0, "Requester communicator size has to be positive!");
```
uekerman, line 97:
```suggestion
  PRECICE_ASSERT(requesterCommunicatorSize >= 0, "Requester communicator size has to be positive.");
```

src/com/MPISinglePortsCommunication.cpp
```diff
@@ -95,21 +101,30 @@ void MPISinglePortsCommunication::acceptConnectionAsServer(std::string const &ac
...
 
-  ConnectionInfoWriter conInfo(acceptorName, requesterName, tag, _addressDirectory);
-
   _isAcceptor = true;
 
-  if (utils::MasterSlave::getRank() == 0) { // only master opens a port
+  const int rank = utils::Parallel::current()->rank();
+
+  ///@todo verify that getRank() makes sense in this case
+  //if (utils::MasterSlave::getRank() == 0) { // only master opens a port
```
uekerman, line 109: Can we clean up that?

src/com/MPISinglePortsCommunication.hpp
```diff
@@ -76,8 +77,14 @@ class MPISinglePortsCommunication : public MPICommunication {
 
   std::string _addressDirectory;
 
-  /// Remote rank -> communicator map
-  std::map<int, MPI_Comm> _communicators;
+  /// A map of direct communication channels based on MPI_COMM_SELF on both sides
+  std::map<int, MPI_Comm> _direct;
+
+  /// The global intercommunicator that connects all ranks
+  MPI_Comm _global = MPI_COMM_NULL;
```
uekerman, line 84: Why again do we need `_direct` **and** `_global` ?

src/com/SocketCommunication.cpp
```diff
@@ -134,9 +134,15 @@ void SocketCommunication::acceptConnectionAsServer(std::string const &acceptorNa
                                                    int                requesterCommunicatorSize)
 {
   PRECICE_TRACE(acceptorName, requesterName, acceptorRank, requesterCommunicatorSize);
-  PRECICE_CHECK(requesterCommunicatorSize > 0, "Requester communicator size has to be > 0!");
+  PRECICE_CHECK(requesterCommunicatorSize >= 0, "Requester communicator size has to be positve!");
```
uekerman, line 137:
```suggestion
  PRECICE_ASSERT(requesterCommunicatorSize >= 0, "Requester communicator size has to be positve.");
```

fsimonis:
> Wouldn't it be a nice add-on to print a warning if OpenMPI is used and MPIPortsCom or MPISinglePortsCom is chosen in CommunicationConfiguration?

True, but I would prefer to split this into a separate PR. This requires to pass the information from CMake to the configuration, which requires some additional work.

> Does changing the default as in #624 work for you now? Not saying that we should do this, but would be good to know. I cannot test this locally since OpenMPI.

I haven't tried #624 in conjunction with this PR yet, but will give it a try prior to merge.
You can always build the source with spack using:
```
$ cd ~/dev/veryimportant/preCICE
$ spack dev-build --until=build precice@develop^mpich
$ cd spack-build
$ spack build-env precice@develop^mpich -- ctest
```

src/com/MPISinglePortsCommunication.hpp
fsimonis: The `_direct` communication covers the case where P2P communication is required, namely:

* Master-Salves communication
* Master-Master communication

Once `acceptConnectionAsServer()` and `requestConnectionAsClient()` are whilst establishing the client-server (aka p2p, m2n) communication, this class connects both the client and the server communicator.
Once that happened, this `_global` communicator can be used for arbitrary p2p communication as well as collectives.

src/com/SocketCommunication.cpp
fsimonis: This is called on each rank, thus this assertion must hold.

src/com/MPISinglePortsCommunication.cpp
fsimonis, line 109:
```suggestion
```

Yes we can :wink:

src/acceleration/MVQNAcceleration.cpp
```diff
@@ -1,3 +1,4 @@
+#include "utils/assertion.hpp"
```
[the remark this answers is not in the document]
fsimonis:
```suggestion
```

You are right. I moved the requiring part to MatrixOperations

src/com/MPISinglePortsCommunication.hpp
uekerman: :+1: I did not fully get this from the comments above though. Could you please add this?

src/com/SocketCommunication.cpp
uekerman: I am not sure if I understand what you mean. You mean that this should indeed be a `CHECK` and no `ASSERT` ?

fsimonis: The requester size always has to be `>0` as at least the Master calls this.
The requester size has to be checked by the remote SolverInterface, thus this should never be triggered. Therefore I think this should be an assertion.

uekerman: Yes, so the suggestion is good, right? :smirk:

fsimonis: Sorry! I misread the diff! :sweat_smile: 
Yes the suggestion is correct.

PR #749, opened 2020-05-06
Remove PrimitiveRTree

fsimonis:
This PR:
* removes the PrimitiveRTree whose functionality was replaced by Vertex/Edge/TriangleRTrees
* refactors the rtree-related tests and migrated the PrimitiveRTree tests to the various individual indices above.

The implementation of PrimitiveRTree was incompatible with boost 1.73.0, thus this PR fixes the compatibility with the newest boost version too.


src/mesh/RTree.hpp
```diff
@@ -1,5 +1,6 @@
 #pragma once
 
+#include <Eigen/Core>
```
oguzziya, line 3: Not sure but we may not need this since `mesh/Triangle.hpp`

fsimonis: I try to state dependencies explicitly instead of relying on transitive includes.
[This is a good read](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#sf10-avoid-dependencies-on-implicitly-included-names) on why this is a good idea.

Our codebase is a bit of a nightmare in that respect though. Maybe running [includewhatyouuse](https://include-what-you-use.org/ ) on it would be a good first step.

oguzziya: Good to know, thank you then 👍

PR #750, opened 2020-05-06
Modifies solverdummy to include data transfer

### LINKED ISSUE #538: Solver dummies do not include read and write functions

KyleDavisSA:
This PR adds data transfer capability to the c, c++ and fortran solverdummies. This uses 3 vertices for transfering vector data and uses the same commands to run. They are also able to run in combination with each other.

This takes over the solverdummy extension for the previous version of preCICE which will be closed https://github.com/precice/precice/pull/540 .

Resolves #538


examples/solverdummies/c/solverdummy.c
```diff
@@ -10,9 +11,15 @@ int main(int argc, char **argv)
   int     solverProcessSize  = 1;
   int     dimensions         = -1;
   double *vertex;
+  double *readData;
+  double *writeData;
   int     meshID   = -1;
   int     dataID   = -1;
-  int     vertexID = -1;
+  int    *vertexID;
```
uekerman, line 18: should be `vertexIDs`, right?

examples/solverdummies/c/solverdummy.c
```diff
@@ -37,37 +44,74 @@ int main(int argc, char **argv)
...
+  for (int i = 0; i < N; i++){
+    for (int j = 0; j < dimensions; j++) {
+      vertex[j + N*i] = i;
+      readData[j + N*i] = i;
+      writeData[j + N*i] = i;
+    }
+    vertexID[i] = i;
+    printf("DUMMY: Vertex: %f, %f, %f \n",vertex[i*N + 0],vertex[i*N + 1],vertex[i*N + 2]);
+    printf("DUMMY: Read Data: %f, %f, %f \n",readData[i*N + 0],readData[i*N + 1],readData[i*N + 2]);
+    printf("DUMMY: Write Data: %f, %f, %f \n",writeData[i*N + 0],writeData[i*N + 1],writeData[i*N + 2]);
```
uekerman, line 71: That's probably too much output.

examples/solverdummies/c/solverdummy.c
```diff
@@ -37,37 +44,74 @@ int main(int argc, char **argv)
...
+
   free(vertex);
 
   dt = precicec_initialize();
 
   while (precicec_isCouplingOngoing()) {
 
     if (precicec_isActionRequired(writeItCheckp)) {
-      printf("DUMMY: Writing iteration checkpoint");
+      precicec_writeBlockVectorData(writeDataID, N, vertexID, writeData);
```
uekerman, line 83: why calling `precicec_writeBlockVectorData` only when `precicec_isActionRequired(writeItCheckp)` ?

examples/solverdummies/c/solverdummy.c
```diff
@@ -37,37 +44,74 @@ int main(int argc, char **argv)
...
+        printf("DUMMY: Write Data: %f, %f, %f \n",writeData[i*N + 0],writeData[i*N + 1],writeData[i*N + 2]);
+      }
+      printf("DUMMY: Writing iteration checkpoint \n");
       precicec_markActionFulfilled(writeItCheckp);
     }
 
     dt = precicec_advance(dt);
 
     if (precicec_isActionRequired(readItCheckp)) {
+      precicec_readBlockVectorData(readDataID, N, vertexID, readData);
```
uekerman, line 94: wrong location, see above

examples/solverdummies/cpp/solverdummy.cpp
```diff
@@ -32,27 +32,70 @@ int main(int argc, char **argv)
...
+    dataReadName="Forces";
+    dataWriteName="Velocities";
+  }
+  const int readDataID = interface.getDataID(dataReadName,meshID);
+  const int writeDataID = interface.getDataID(dataWriteName,meshID);
+
+  double vertexPos[N * dimensions];
+  double readData[N * dimensions];
+  double writeData[N * dimensions];
+  int vertexIDs[N];
```
uekerman, line 53: We should use `std::vector` here.

examples/solverdummies/cpp/solverdummy.cpp
```diff
@@ -32,27 +32,70 @@ int main(int argc, char **argv)
...
+    std::cout << "DUMMY: Write Data: " << writeData[i*N + 0] << " ; " << writeData[i*N + 1] << " ; " << writeData[i*N + 2] << "\n";
+  }
+  interface.setMeshVertices(meshID,N,vertexPos,vertexIDs);
 
   double dt = interface.initialize();
 
   while (interface.isCouplingOngoing()) {
 
     if (interface.isActionRequired(actionWriteIterationCheckpoint())) {
+      interface.writeBlockVectorData(writeDataID,N,vertexIDs,writeData);
```
uekerman, line 73: wrong location

examples/solverdummies/cpp/solverdummy.cpp
```diff
@@ -32,27 +32,70 @@ int main(int argc, char **argv)
...
       std::cout << "DUMMY: Writing iteration checkpoint\n";
       interface.markActionFulfilled(actionWriteIterationCheckpoint());
     }
 
+
     dt = interface.advance(dt);
 
     if (interface.isActionRequired(actionReadIterationCheckpoint())) {
-      std::cout << "DUMMY: Writing iteration checkpoint\n";
+      interface.readBlockVectorData(readDataID,N,vertexIDs,readData);
```
uekerman, line 85: wrong location

MakisH:
This will change the results of the system tests on purpose @Eder-K.

uekerman:
> One comment: I'm not sure whether we want/need this, but we could also include `initializeData()`.

Let's keep it simple. That's the story for another day (another PR).

MakisH:
Mainly for documentation purposes: what is the motivation behind changing from an explicit to an implicit scheme and from a 2D to a 3D interface?

MakisH:
Ported the changes also to the Fortran module: https://github.com/precice/fortran-module/pull/5
Ready for review, currently waiting for this one to get merged first.

examples/solverdummies/c/solverdummy.c
```diff
@@ -37,37 +44,73 @@ int main(int argc, char **argv)
...
-  dimensions = precicec_getDimensions();
-  vertex     = malloc(dimensions * sizeof(double));
+  if (strcmp(participantName,"SolverOne") == 0){
+    writeDataID = precicec_getDataID("dataOne",meshID);
+    readDataID = precicec_getDataID("dataTwo",meshID);
+  }
+  if (strcmp(participantName,"SolverTwo") == 0){
+    writeDataID = precicec_getDataID("dataTwo",meshID);
+    readDataID = precicec_getDataID("dataOne",meshID);
+  }
```
BenjaminRueth, line 54: It is a bit strange to use `participantName` as a "key" for all the parameters (`writeDataID`, `readDataID`) except `meshName`. I would suggest something similar to the pesudocode below: 

```
  int meshID;
  const char *meshName;
  if (strcmp(participantName,"SolverOne") == 0){
    meshName = "meshOne";
    meshID = precicec_getMeshID(meshName);
    writeDataID = precicec_getDataID("dataOne",meshID);
    readDataID = precicec_getDataID("dataTwo",meshID);
  }
  if (strcmp(participantName,"SolverTwo") == 0){
    meshName = "meshTwo";
    meshID = precicec_getMeshID(meshName);
    writeDataID = precicec_getDataID("dataTwo",meshID);
    readDataID = precicec_getDataID("dataOne",meshID);
  }
  assert(strcmp(argv[3], meshName));
```

I would suggest to not change the signature, since this means we will have to change all our tests, as well. The assertion ensures that everything is still consistent. Let's clean this up later.

If we want to merge this PR now, I would also be totally fine with turning all this into an issue and leave it for now as it is.

BenjaminRueth: We decided not to do the suggested change above, but we will directly solve the problem thoroughly in the future (https://github.com/precice/precice/issues/829).

IshaanDesai:
The modified solver dummies for `C`, `C++` and `Fortran` are tested with the new `Python` solver dummy and each combination works. Python solver dummy is modified in the PR: https://github.com/precice/python-bindings/pull/48

BenjaminRueth:
Also did the cross-check with https://github.com/precice/fortran-module/pull/5 and all solverdummies worked.

Issue #713, opened 2020-04-07
Migration to Github Actions


fsimonis:
GitHub Actions gives us the possibility to run the tests in docker containers. This now opens the question on what to test. Below a list of things we should consider when building our test Matrix.

Questions to the reader:
* What else has to be tested?
* What can be removed from the test matrix?

## preCICE Feature Configurations
* PythonActions - This is a pure add-on, so let's always enable it.
* Fortran/C Bindings - Also a pure add-on, so let's always enable it.
* MPICommunication - This is required to run the actual test suite after #710 
* PETScMapping - This is not a pure add-on as both paths run code. This will be even more significant once we get a fallback implementation for parallel RBF.

Configurations to test (nothing changed):

 MPI | PETSc
 --- | ---
Off | Off
On | Off
On | On

## Build Type
* Debug - Enables assertions and debug output.
* Release - Optimized build which we did not test so far. (we should)

## Compiler
* GCC - This is what the vast majority of our users use.
* Clang - This is what some of the developers use. So we probably don't need to explicitly test this anymore.
* Intel - Significant for clusters, but cumbersome to integrate into testing. Let's ignore this for now.

## MPI Implementations
* Open MPI - default MPI on ubuntu / Arch
* MPICH - no precompiled packages available. We would have to build the dependencies from source. Spack could help.
* Intel MPI - cumbersome to setup. Let's ignore this.

## Systems
* Ubuntu 16.04 LTS - supported until 2021, possibly not worth it.
* Ubuntu 18.04 LTS - supported until 2023, our current baseline for dependencies
* Ubuntu 20.04 LTS - supported until 2025
* Arch Linux - rolling release, thus everything is always at the latest release. Official petsc package missing.
* CentOS 7 - frequently pops up with users.

## Dependency versions:
* Latest release (covered by Arch)
* Oldest supported release (covered by Ubuntu 18.04 LTS)

# Build Matrix

What | Options
--- | ---
Features | 3
Build Type | 2
Compiler | 1
MPI Implementations | 2
Systems | 4
**Product** | **48**

uekerman:
Looks good in my opinion. Thanks for the overview. 
I guess we can still decrease the number a bit:
* I don't think we need the full product of "MPI Implementations" times "Systems". We could just use OpenMPI by default and additionally test MPICH on one selected system (e.g. Ubuntu 18.04) ? 
* We also don't have "Features" times "MPI Implementations". 
[...]

PR #754, opened 2020-05-08
Replace FindPETSc by a pkg-config wrapper

fsimonis:
This PR replaces the internals of the FindPETSc module by a pure pkg-config wrapper.
This new wrapper creates a `PETSc::PETSc` target, which makes linking in preCICE simpler and less error prone. This target follows the canonical naming scheme, so this should be compatible in case PETSc decides to provide a CMake config file in the future.

This PR superseeds #551 

**Todo**
* [x] Test how this method works with the packages provided on various distributions
* [ ] Test how this method works when PETSc is compiled as a static library.


fsimonis:
Let's wait with this until we have adopted #713

fsimonis:
This PR works on
* ubuntu 1804 (installation in `/opt/petsc` with `export PETSC_DIR=/opt/petsc`)
* ubuntu 2004 (system package)

tools/ci/travis-install-dependencies.sh
```diff
@@ -97,3 +98,27 @@ if [ ! -f $CACHE_CCACHE_TOKEN ]; then
     # Create token
     touch $CACHE_CCACHE_TOKEN
 fi
+
+
+# Download and compile pkg-config 0.29.2
```
MakisH, line 103: This is the first time we download a dependency ourselves automatically, right? This is a shift from our current approach and I am not sure if it would work on some clusters. What is the motivation for this decision?

MakisH (commented):
It works on my system, but I found a case in which it does not work now (while it works on develop):
- PETSc 3.13.3 built from source, setting `PETSC_DIR`, `PETSC_ARCH`, `LD_LIBRARY_PATH`, `LIBRARY_PATH`, `CPATH`, `PYTHONPATH`, `PKG_CONFIG_PATH`.
- PETSc 3.6.2 from APT

CMake picks the locally-built PETSc version, but the shared object is linked against the system version. I did not have this issue with the previous approach.

I am not sure if this is a use case we want to address, but we will have this issue as long as we require newer PETSc versions than available by default on some systems.

fsimonis:
@MakisH Thank you for the report!
I provided the detected paths as `PATHS` to `find_library`. The correct way is to provide them as `HINTS`, which are then used as preferred paths instead of additional paths.

tools/ci/travis-install-dependencies.sh
fsimonis: This installs a newer version of pkg-config in the travis runtime. The version from 14.04 is simply to old.
However, this is now obsolete since #843

fsimonis:
This also allows to compile preCICE in slurm jobs.
Previously, we had users who ran into issues attempting this.

MakisH (approved):
In the same problematic scenario I described, all tests work now! :+1:

tools/ci/travis-install-dependencies.sh
MakisH: Oh, for a moment I thought this was in our CMakeLists.txt, sorry.

fsimonis:
Let's merge this after #875

floli:
This blocks update of the AUR build script.

MakisH:
> This blocks update of the AUR build script.

How are these two related? What is the issue?

floli:
Issue is the one in the related comment #878 
```
CMake Error at /usr/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:165 (message):
  PETSc could not be found.  Be sure to set PETSC_DIR and PETSC_ARCH.
  (missing: PETSC_EXECUTABLE_RUNS) (found suitable version "3.14.0", minimum
  required is "3.12")
Call Stack (most recent call first):
  /usr/share/cmake-3.18/Modules/FindPackageHandleStandardArgs.cmake:458 (_FPHSA_FAILURE_MESSAGE)
  cmake/modules/FindPETSc.cmake:345 (find_package_handle_standard_args)
  CMakeLists.txt:217 (find_package)
```
which occurs with `makepkg` the current master's `PKGBUILD`.

fsimonis:
@floli I cannot reproduce this on my machine nor in the `precice/ci-archlinux` image, nor after building an up-to-date version of the image.
This is not surprising, as the lack of reproducibility is one of the reasons why this PR exists in the first place.

We decided to release this change in the next release to allow us to test is for a longer time on develop. Thus, I propose to disable `petsc` one last time for `2.1.1` AUR package. The next release should then finally be a feature-complete release.

PR #757, opened 2020-05-14
Remove support for CPP and CPPFLAGS env

### LINKED ISSUE #756: Incorrect handling of CPP and CPPFLAGS in XSDKCompilerDetect.cmake
**Describe your setup**

This bug was initially discovered when building on our HPC platform (CentOS via EasyBuild), but it can be reproduced on _Ubuntu 19.10_ and probably any other systems.
preCICE Version: 2.0.2

**Describe the problem**

CMake fails when trying to set `CPP` and `CPPFLAGS` under certain conditions.
There are actually 2 bugs:
- `CPP` is a C PreProcessor, not a C++ compiler, so it cannot be used a such
- When environment variable `CPPFLAGS` and CMake option `CMAKE_CXX_FLAGS` are both defined, `list(APPEND)` inserts a `;` in the list of flags (because that's how list works in CMake) which is then interpreted a 2 different commands when CMake tests the C++ compiler.

**Step To Reproduce**

For the `CPP` bug:
```
CPP=cpp cmake ..
```
will just hang because `cpp` waits for the standard input.

For the `CPPFLAGS` bug:
```
CPPFLAGS=-DTEST cmake .. -DCMAKE_CXX_FLAGS='-Wall'
```
will fail with
```
    Building CXX object CMakeFiles/cmTC_b4c75.dir/testCXXCompiler.cxx.o
    /usr/bin/c++    -Wall;-DTEST    -o CMakeFiles/cmTC_b4c75.dir/testCXXCompiler.cxx.o -c /home/xavier/work/precice/build/CMakeFiles/CMakeTmp/testCXXCompiler.cxx
    c++: fatal error: no input files
    compilation terminated.
    /bin/sh: 1: -DTEST: not found
```

**Discussion**

I believe you want to follow the xSDK Community Installation Policies and take into account these flags, but CMake does not support them and I'm not sure there is an easy workaround.
https://gitlab.kitware.com/cmake/cmake/-/merge_requests/3633

For `CPP`, I believe CMake does not support it because it would mean adding a additional step for every compilation of a file, ie first run CPP, and pass the output to CXX to compile it.

For `CPPFLAGS`, it is not that easy to get it correct:
- `list(APPEND)` does not work for the reason above
- `set(CMAKE_CXX_FLAGS "..."` is ignored if the value is already set by the user call CMake
- `set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} $ENV{CPPFLAGS}" CACHE STRING "" FORCE)` will work but the `CPPFLAGS` will keep being added everytime CMake is run.
- Maybe a better logic could be added to get it right.

My feeling is that this issue is not for preCICE to solve, but either:
- by CMake (but for `CPP`, I am not sure it will happen)
- by having xSDK relaxing policies, in particular because CMake is one of their preferred build tool.

How do the other CMake-based xSDK-compatible software handle this?

As an alternative, I could suggest displaying a warning in case one of these variables is defined to inform the user it will not be used.

fsimonis:
This PR removes the buggy support for the `CPP` and `CPPFLAGS` environment variables in CMake.
These variables are used in an unconventional way and not natively supported via CMake.

Closes #756


fsimonis:
We decided to drop the buggy support for `CPP` and `CPPFLAGS` in develop.

PR #759, opened 2020-05-19
Add CI Format Checker

### LINKED ISSUE #171: Automatically check for code style
Problem: Even though we have a [.clang-format](https://github.com/precice/precice/blob/develop/.clang-format) file, contributed code sometimes does not follow it. Therefore, implement an automatic check (maybe similar to #170 ) for conformance of the contributed files.

Side-effect: since our code is not entirely conforming at the moment, we should be careful not to mix style changes with "real" edits, making the reviewing very difficult. Althought this can be implemented now, it needs to be deployed after we fix the complete source base.

@shkodm is currently working on this.

fsimonis:
This PR adds a code format checker for pull requests. It runs only when C/C++ files were changed.
It currently checks all files, which takes about 1 minute in total.

Partially replaces #391 
Closes #171

### DIFF
--- a/.github/workflows/code-style.yml
+++ b/.github/workflows/code-style.yml
@@ -1,0 +1,17 @@
+name: code-style
+on:
+  pull_request:
+    paths:
+      - '**.[ch]pp'
+      - '**.[ch]'
+      - '**.xml'
+
+jobs:
+  formatting:
+    runs-on: ubuntu-latest
+    container: precice/ci-formatting:latest
+    steps:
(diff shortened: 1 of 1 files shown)


fsimonis:
I extended the formatting ci image https://github.com/precice/ci-images/commit/839123c793f4d69cb181f6ed4fb4be355361ec65#diff-7eadfa78e8e3f7bbafb079a0945a95ed and rebased on develop to also use the xml formatter #866. 

This is ready to merge now.

PR #760, opened 2020-05-19
Re-enable reset of written data at the end of advance

uekerman:
For the manifold mapping, some years ago, we had to disable the reset of written data at the end of `advance()`. Since we recently removed manifold mapping this PR re-enables this safeguard.

**Example:**

Before this PR ...

```c++
writeBlockVectorData(forces);
advance();
advance();
```

uses `forces` twice. Now, the preCICE data structures are reset to zero at the end of the first `advance()`. Thus, the second `advance()` uses zero and not `forces`.

**Argument in favor of this safeguard:** it has already revealed a bug in an integration test, which I fixed on-the-fly.

### DIFF
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -381,6 +381,8 @@ double SolverInterfaceImpl::advance(
   PRECICE_DEBUG("Handle exports");
   handleExports();
 
+  resetWrittenData();
+
   _meshLock.lockAll();
   solverEvent.start(precice::syncMode);
   return _couplingScheme->getNextTimestepMaxLength();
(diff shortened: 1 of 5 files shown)


src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -381,6 +381,8 @@ double SolverInterfaceImpl::advance(
   PRECICE_DEBUG("Handle exports");
   handleExports();
 
+  resetWrittenData();
```
fsimonis, line 384: Is this issue severe enough to enforce this safe-guard on all our user?
Maybe it would be an idea to enable it only in Debug builds:

```suggestion
#ifndef NDEBUG
  resetWrittenData();
#endif
```

fsimonis: We decided to leave it enabled all the time. Everything else is surprising for users and inconsistent for developers.

fsimonis:
I committed the changelog onto develop in 583339a9

PR #744, opened 2020-04-27
Improve errors messages of section 1.6

BenjaminRueth:
Improves error messages of section 1.6 in #698.

Notes:

* Trying to move configuration errors into `CplSchemeConfig.cpp`
* Some error messages are skipped until #689 is merged.

PR #764, opened 2020-06-07
Remove check configuration from BaseCouplingScheme

BenjaminRueth:
Preparatory work for #744


BenjaminRueth:
This PR only introduces minimal changes that we do not really need. Will be closed.

PR #768, opened 2020-06-08
Fixing potential memory leak

ajaust:
The pointer to preCICE was never deleted nor was the value of the pointer checked before it was created. I am not sure if it is a valid use case, but a user might finalize preCICE and initialize it again for some reason or somebody calls preCICE's `initialize` functions twice. In such a case I propose that the old object should be either deleted or one should throw an error.

### DIFF
--- a/extras/bindings/c/src/SolverInterfaceC.cpp
+++ b/extras/bindings/c/src/SolverInterfaceC.cpp
@@ -64,6 +64,7 @@ void precicec_finalize()
   if (interface != nullptr)
   {
     interface->finalize();
+    interface.reset();
   }
 }
 
--- a/extras/bindings/c/src/SolverInterfaceC.cpp
+++ b/extras/bindings/c/src/SolverInterfaceC.cpp
@@ -7,7 +7,7 @@ extern "C" {
 #include "precice/impl/versions.hpp"
 #include "utils/assertion.hpp"
 
-static std::unique_ptr<precice::SolverInterface> interface;
+static std::unique_ptr<precice::SolverInterface> interface = nullptr;
 
 void precicec_createSolverInterface_withCommunicator(
     const char *participantName,
@@ -20,12 +20,12 @@ void precicec_createSolverInterface_withCommunicator(
   std::string stringConfigFileName(configFileName);
 
   PRECICE_ASSERT(interface == nullptr);
-  interface = std::unique_ptr<precice::SolverInterface>( new precice::SolverInterface(stringAccessorName,
-                                                                                     stringConfigFileName,
-                                                                                     solverProcessIndex,
-                                                                                     solverProcessSize,
-                                                                                     communicator)
-                                                        );
+  interface.reset( new precice::SolverInterface(stringAccessorName,
+                                               stringConfigFileName,
+                                               solverProcessIndex,
+                                               solverProcessSize,
+                                               communicator)
+                  );
 }
 
 void precicec_createSolverInterface(
--- a/extras/bindings/c/src/SolverInterfaceC.cpp
+++ b/extras/bindings/c/src/SolverInterfaceC.cpp
@@ -7,7 +7,12 @@ extern "C" {
 #include "precice/impl/versions.hpp"
 #include "utils/assertion.hpp"
 
-static std::unique_ptr<precice::SolverInterface> interface = nullptr;
+static std::unique_ptr<precice::SolverInterface> impl = nullptr;
+
+static precice::logging::Logger _log("SolverInterfaceC");
+
+static std::string errormsg = "preCICE has not been created properly. Be sure to call \"precicec_createSolverInterface\" or \"precicec_createSolverInterface_withCommunicator\" before any other call to preCICE.";
+static std::string errormsgCreate = "preCICE has been created already! Be sure to call either \"precicec_createSolverInterface\" or \"precicec_createSolverInterface_withCommunicator\" exactly once.";
 
 void precicec_createSolverInterface_withCommunicator(
     const char *participantName,
@@ -19,8 +24,8 @@ void precicec_createSolverInterface_withCommunicator(
   std::string stringAccessorName(participantName);
   std::string stringConfigFileName(configFileName);
 
-  PRECICE_ASSERT(interface == nullptr);
-  interface.reset( new precice::SolverInterface(stringAccessorName,
+  PRECICE_CHECK(impl == nullptr, errormsgCreate);
+  impl.reset( new precice::SolverInterface(stringAccessorName,
                                                stringConfigFileName,
                                                solverProcessIndex,
                                                solverProcessSize,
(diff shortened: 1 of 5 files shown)


ajaust:
Alternative: Could you use a smart pointer (`std::unique_ptr`) here or would it break compatibility with C?

fsimonis:
@ajaust Well spotted!

Compatabiliy is no issue as we already use string etc. Would you like to modernize the bindings to use `std::unique_ptr`?

ajaust:
> @ajaust Well spotted!
> 
> Compatabiliy is no issue as we already use string etc. Would you like to modernize the bindings to use `std::unique_ptr`?

I will do that, if that is fine for you. I prefer smart pointers over raw pointers and since you require C++14 this should not be a concern.

*Side note*
I did a quick `grep` on the source code of preCICE and I must say you have a lot of raw pointers (or at least a lot of `new`) in the source code. Maybe that is a good project for be fixed by a student some time. ;)

extras/bindings/c/src/SolverInterfaceC.cpp
```diff
@@ -16,6 +16,8 @@ void precicec_createSolverInterface(
 {
   std::string stringAccessorName(participantName);
   std::string stringConfigFileName(configFileName);
+  if (interface != nullptr)
+    delete interface;
```
uekerman, line 20: We should better throw an error message here for the moment. It's currently never allowed to call that function twice, even with a different `participantName`. I could imagine that users expect a different behavior if they see that they can call this twice.

extras/bindings/c/src/SolverInterfaceC.cpp
```diff
@@ -31,6 +33,8 @@ void precicec_createSolverInterface_withCommunicator(
...
```
uekerman, line 37: same as above

extras/bindings/c/src/SolverInterfaceC.cpp
ajaust: Is this done via `PRECICE_ASSERT` or do you have any other way when throwing errors in the C interface?

fsimonis:
@ajaust preCICE still requires C++11, not C++14.

> I did a quick grep on the source code of preCICE and I must say you have a lot of raw pointers (or at least a lot of new) in the source code.

`grep new` is misleading as we often create shared pointers with the new syntax. (We should probably migrate to `std::make_shared`. )
`grep delete` gives you more insight on the code base.

Migrating the code base from shared to unique ownership would be a good project too.

ajaust:
> @ajaust preCICE still requires C++11, not C++14.
> 
> > I did a quick grep on the source code of preCICE and I must say you have a lot of raw pointers (or at least a lot of new) in the source code.
> 
> `grep new` is misleading as we often create shared pointers with the new syntax. (We should probably migrate to `std::make_shared`. )
> `grep delete` gives you more insight on the code base.
> 
> Migrating the code base from shared to unique ownership would be a good project too.

I did `grep delete` as well. Also there you still have some opportunities to use smart pointers ( in `SolverInterfaceFASTEST.cpp`, `SolverInterfaceFortran.cpp`, `Participant.cpp`, `MappingConfiguration.cpp`, `FindClosestTest.cpp`, `ParallelMatrixOperationsTest.cpp`). Grepping for `delete` alone does not give much insight on possibly missing `delete`, thus I usually grep for `new` as well. I would support to make more explicit calls to `make_shared` or `make_unique`. This would also be what the core guidelines recommend.

extras/bindings/c/src/SolverInterfaceC.cpp
uekerman: `PRECICE_ASSERT` here would be consistent with the rest of the class, yes. In principle, we could change all assertions here to errors, but that's the story of another PR. We have no perfect policy yet how to treat such problems in the bindings.

ajaust:
I converted the PR to a draft for now. I will revise soon.

ajaust:
## Update

I have updated the pull request. 

I also changed the Fortran bindings since I missed #776.

@fsimonis I do not use `reset` on the pointers, but mimic what `make_unique` would do (if I remember the implementation of `make_unique` correctly). I am not sure what is the preferred way. (I had forgotten that `c++11` did not have `make_unique` yet. :disappointed: )

I did not use `reset` on in the finalize routine either. Is this a problem with the lifetime of the static object here?


I was digging a bit through the code to see that 
```c++
SolverInterfaceImpl::SolverInterfaceImpl(
    std::string        participantName,
    const std::string &configurationFileName,
    int                accessorProcessRank,
    int                accessorCommunicatorSize)
```
invokes 
```c++
SolverInterfaceImpl::SolverInterfaceImpl(std::move(participantName), configurationFileName, accessorProcessRank, accessorCommunicatorSize, nullptr)
```
in the end. Therefore, I changed the order of `precicec_createSolverInterface` and `precicec_createSolverInterface_withCommunicator` in order to minimize code duplication. I would need some feedback whether the change is safe or not.

## Error 

When running the tests locally, I receive the following error for the C and Fortran solver dummies:

```
terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::log::v2_mt_posix::system_error> >'
```

I am not sure whether this is related to my changes or not. I think I have seen this also in my normal installation of preCICE 2.0.2. I am not sure if I messed something up in my Boost or Linux installation. So far I had no time to investigate this.

## Site note
Regarding the code duplication: Would it not make sense to define the constructor of `class SolverInterfaceImpl`as
```c++ 
  SolverInterfaceImpl(
      std::string        participantName,
      const std::string &configurationFileName,
      int                accessorProcessRank,
      int                accessorCommunicatorSize,
      void *             communicator = nullptr);
```
such one avoids implementing 
```c++
  SolverInterfaceImpl(
      std::string        participantName,
      const std::string &configurationFileName,
      int                accessorProcessRank,
      int                accessorCommunicatorSize);
```
explicitly? The less code one writes the less bugs could pop up. It is the same case for `class SolverInterface`.

fsimonis:
@ajaust `unique_ptr.reset(new )` is exception safe.
`make_unique` is more expressive and can prevent resource leaks in cases where the compiler is allowed to interleave the call to `new` and the construction of the `unique_ptr` with operations that may throw. Arguments passed to a function are the prime example here.

>  terminate called after throwing an instance of 'boost::exception_detail::clone_impl<boost::exception_detail::error_info_injector<boost::log::v2_mt_posix::system_error> >'

This is most likely due to the tracing functionality. It calls the logger in the destructor, which may throw. I fixed this in 7e2cdfc5 as it is dangerous potential pitfall.

> Regarding the code duplication: Would it not make sense to define the constructor of class SolverInterfaceImpl as

We kept the old constructor intentionally to keep the ABI stable and most importantly to avoid confusion about the additional parameter.

> I also changed the Fortran bindings since I missed #776.

Can you please rebase on develop as #776 was already merged?

ajaust:
@fsimonis Yeah, I will rebase and come back here. :-)

ajaust:
I rebased the branch and changed my fix to match the changes in #776 better (use of `reset`).

extras/bindings/c/src/SolverInterfaceC.cpp
```diff
@@ -58,8 +61,11 @@ double precicec_advance(double computedTimestepLength)
 
 void precicec_finalize()
 {
-  PRECICE_ASSERT(interface != nullptr);
-  interface->finalize();
+  if (interface != nullptr)
+  {
+    interface->finalize();
+    interface.reset();
+  }
```
fsimonis, line 68: This is not consistent with the Fortran bindings:

https://github.com/precice/precice/blob/583339a950427b530bc4872649145e14f66cb0ea/extras/bindings/fortran/src/SolverInterfaceFortran.cpp#L70-L75

A `PRECICE_CHECK` should suffice.

extras/bindings/c/src/SolverInterfaceC.cpp
```diff
@@ -1,41 +1,44 @@
...
+    void *      communicator)
 {
   std::string stringAccessorName(participantName);
   std::string stringConfigFileName(configFileName);
-  interface = new precice::SolverInterface(stringAccessorName,
-                                           stringConfigFileName,
-                                           solverProcessIndex,
-                                           solverProcessSize);
+
+  PRECICE_ASSERT(interface == nullptr);
```
fsimonis, line 22: As this is covers wrong API usage, this should be a `PRECICE_CHECK`.
We use assertions only for checking internal pre/post conditions. A user _should_ never run into a `PRECICE_ASSERT`.

Using `CHECK` requires a logger. You can have a look at the fortran bindings.
https://github.com/precice/precice/blob/583339a950427b530bc4872649145e14f66cb0ea/extras/bindings/fortran/src/SolverInterfaceFortran.cpp#L14

ajaust: How should this be changed? The solver interface for C only uses `PRECICE_ASSERT` and not a single `PRECICE_CHECK`. Should I replace all assertions by checks?

uekerman: In my opinion this could also be solved in another PR. Not related to the main theme here.

ajaust: I have just changed `PRECICE_ASSERT` to `PRECICE_CHECK` and the name of the `unique_ptr` to `impl` in 04a637e156673b42032993697f53b7ad87f4a1b1. If you do not like that, I would suggest to drop the commit.

extras/bindings/c/src/SolverInterfaceC.cpp
ajaust: I have changed that to be consistent with the Fortran interface in the sense that there is a `PRECICE_ASSERT` before finalizing preCICE and resetting the `unique_ptr` in 30d69b59149f9073170ddd30884ae7569ca72d39.

fsimonis:
This PR successfully fixes the memory leaks in the C bindings.
Let's bring both Fortran and C bindings into a consistent state in a separate PR.

@ajaust Thank you for pointing out and fixing the leaks!

ajaust:
Great! No problem. I am happy to help. :smile:

Issue #765, opened 2020-06-08
Calculation of shape-parameter Gaussian RBF from support-radius


KyleDavisSA:
**Please describe the problem you are trying to solve.**
Using the Gaussian RBF, it is difficult to specify a shape parameter for any given mesh as this is a function of the Gaussian decay and mesh width.

**Describe the solution you propose.**
A python script is available in preCICE to determine the shape parameter [here](https://github.com/precice/precice/tree/develop/extras/rbfShape). This formula is dependent on the Gaussian decay, number of points that is desired inside a radius around the vertex, and the average mesh width. This equation is desired in situations when you only want a limited amount of vertices to be included in the local RBF interpolation. It would be better for the support radius defined in the `precice-config.xml` file and have preCICE determine the shape parameter internally as a function of the support radius. The Gaussian can then be used by specifying either the shape-parameter or support-radius.

uekerman:
Another interesting use case of letting preCICE determine the mesh width is to only specify the "number of points desired to cover" and then compute an "adaptive" decay for each shape function.

KyleDavisSA:
A naive approach for the mesh width is to find the nearest neighbour for each point and the distance between them, and then set the support radius as a function of this value (say 2 or 3 times this value). Conditioning of the RBF matrix will need to be looked at, but probably will probable be okay if the mesh width changes "gradually".

uekerman:
If the nearest neighbor is easier to get (meaning to implement) and cheaper than the k-nearest neighbors then this could be a good compromise, yes.

fsimonis:
The rbf tagging requires the vertex index tree too. So it has to be generated at some point.
Once it exists, querying it should not be overly expensive.

PR #769, opened 2020-06-09
Simplify the Gaussian Basis Function

fsimonis:
We never change the support radius of the Gaussian via the constructor.

Removing this argument allows to simplify the class quite a bit.
More important, the Gaussian then becomes a RBF with compact support.

We now know for each basis function weather it has compact support.
This allows us to lift this information into the type system.

### DIFF
--- a/src/mapping/impl/BasisFunctions.hpp
+++ b/src/mapping/impl/BasisFunctions.hpp
@@ -139,24 +139,22 @@ public:
  */
 class Gaussian {
 public:
-  Gaussian(const double shape, const double supportRadius = std::numeric_limits<double>::infinity())
-      : _shape(shape),
-        _supportRadius(supportRadius)
+  Gaussian(const double shape)
+      : _shape(shape)
   {
     PRECICE_CHECK(math::greater(_shape, 0.0),
                   "Shape parameter for radial-basis-function gaussian has to be larger than zero!");
-    PRECICE_CHECK(math::greater(_supportRadius, 0.0),
-                  "Support radius for radial-basis-function gaussian has to be larger than zero!");
 
-    _deltaY          = hasCompactSupport() ? evaluate(supportRadius) : 0;
-    double threshold = std::sqrt(-std::log(cutoffThreshold)) / shape;
-    _supportRadius   = std::min(supportRadius, threshold);
+    /// Below that value the function is supposed to be zero. Defines the support radius if not explicitely given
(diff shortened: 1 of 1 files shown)


fsimonis:
Related to #765

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -139,24 +139,22 @@ class VolumeSplines {
...
-    PRECICE_CHECK(math::greater(_supportRadius, 0.0),
-                  "Support radius for radial-basis-function gaussian has to be larger than zero!");
 
-    _deltaY          = hasCompactSupport() ? evaluate(supportRadius) : 0;
-    double threshold = std::sqrt(-std::log(cutoffThreshold)) / shape;
-    _supportRadius   = std::min(supportRadius, threshold);
+    /// Below that value the function is supposed to be zero. Defines the support radius if not explicitely given
+    constexpr double cutoffThreshold = 1e-9;
+
+    _supportRadius = std::sqrt(-std::log(cutoffThreshold)) / shape;
```
KyleDavisSA, line 151: This would still require the user to calculate the shape parameter before hand from the desired support radius? We should have something like 

```
Gaussian(const double supportRadius)

shape = std::sqrt(-std::log(cutoffThreshold)) / _supportRadius;
```
Or am I wrong?

fsimonis: > This would still require the user to calculate the shape parameter before hand from the desired support radius?

Yes. This PR is just a simplification of the rbf code. It does not change anything in the actual program. Currently, passing a support radius to the Gaussian is not used and not tested.

> We should have something like

I think #765 is a more appropriate place to discuss this.

fsimonis:
Let's keep this PR on hold until we know how to proceed.

uekerman:
> Let's keep this PR on hold until we know how to proceed.

I agree. I am bit hesitant to this change. We currently don't use the functionality, but maybe we should. It is used in [our RBF paper](ftp://ftp.informatik.uni-stuttgart.de/pub/library/ncstrl.ustuttgart_fi/INPROC-2017-35/INPROC-2017-35.pdf).

Let's first figure out in #765 what we want.

fsimonis:
During a discussion with @uekerman, we realized that our current implementation always uses a cut-off threshold. Therefore our Gaussian is always compact. I created PR #769 which applies only the simplification without altering the interface.
This PR is thus obsolete.

PR #772, opened 2020-06-12
Fix Resource Finalization of SolverInterface

fsimonis:
This PR refactors and fixes the resource cleanup of the SolverInterface(Impl).

It:
* adds state tracking to the SolverInterfaceImpl
* adds a destructor to SolverInterfaceImpl which calls finalize if it was not called explicitly prior to the destructor
* changes `finalize()` to only finalize the SolverInterface state (accessor, m2ns, cplscheme) if `initialize` has been called
* changes `finalize()` to release ownership of Participants and M2Ns

**The Problem**

The SolverInterface exists in various states:
```cpp
{
  auto interface =  precice::SolverInterface(...);
  // State : constructed
  interface.initialize();
  // State : initialized
  interface.finalize();
  // State : finalized
}  /* end of sope: interface.~SolverInterface() */
// State : gone / deconstructed and deleted
```

The current implementation **exclusively** supports the above lifecycle.

The following results in leaks:
```cpp
{ 
  auto leaky_interface =  precice::SolverInterface(...);
} // At least MPI / PETSc / Events are not finalized here
```

As does this:
```cpp
{ 
  auto leaky_interface =  precice::SolverInterface(...);
  interface.initialize();
} // MPI / PETSc / Events are not finalized
```

Even the correct example can hang and crash as the lifetime of the PETRBF mapping end with the destruction of the SolverInterface. PETSc will be finalized in `finalize`, which can result in errors if MPI is managed by preCICE.

lifecycle

**Additional Context**

Related to #768 

Special thanks to @ajaust for pointing out the lifetime issues!

### DIFF
--- a/docs/changelog/772.md
+++ b/docs/changelog/772.md
@@ -1,3 +1,3 @@
-* Fix value sematics of `precice::SolverInterface`
+* Fix value semantics of `precice::SolverInterface`
 * Fix memory leaks and hanging communication when not calling `precice::SolverInterface::finalize()`
 * Fix occasional errors when using PETRBF with a preCICE-managed MPI Communicator.
--- a/docs/changelog/772.md
+++ b/docs/changelog/772.md
@@ -1,0 +1,3 @@
+* Fix value sematics of `precice::SolverInterface`
+* Fix memory leaks and hanging communication when not calling `precice::SolverInterface::finalize()`
+* Fix occasional errors when using PETRBF with a preCICE-managed MPI Communicator.
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -105,6 +105,13 @@ SolverInterfaceImpl::SolverInterfaceImpl(
 {
 }
 
+SolverInterfaceImpl::~SolverInterfaceImpl()
+{
+  if (_state != State::Finalized) {
+    finalize();
+  }
+}
+
 void SolverInterfaceImpl::configure(
     const std::string &configurationFileName)
 {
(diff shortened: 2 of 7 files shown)


fsimonis:
@uekerman Please review and merge if ok

docs/changelog/772.md
```diff
@@ -0,0 +1,3 @@
+* Fix value sematics of `precice::SolverInterface`
```
uekerman, line 1:
```suggestion
* Fix value semantics of `precice::SolverInterface`
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -401,49 +412,58 @@ void SolverInterfaceImpl::finalize()
...
 
-  PRECICE_DEBUG("Handle exports");
-  for (const io::ExportContext &context : _accessor->exportContexts()) {
-    if (context.everyNTimeWindows != -1) {
-      std::ostringstream suffix;
-      suffix << _accessorName << ".final";
-      exportMesh(suffix.str());
-      if (context.triggerSolverPlot) {
-        _couplingScheme->requireAction(std::string("plot-output"));
+    // PRECICE_CHECK(_couplingScheme->isInitialized(), "initialize() has to be called before finalize()");
```
uekerman, line 417:
```suggestion
    PRECICE_ASSERT(_couplingScheme->isInitialized());
```

src/precice/impl/SolverInterfaceImpl.hpp
```diff
@@ -127,8 +133,25 @@ class SolverInterfaceImpl {
...
+   * - finalizes managed PETSc
+   * - finalizes managed MPI
+   *
+   * @post Closes MasterSlave communication
+   * @post Finalized managed PETSc
+   * @post Finalized managed MPI
+   *
+   * @warning
+   * Finalize is not the inverse of initialize().
+   * It has to be called after constuction.
```
uekerman, line 154:
```suggestion
   * Finalize has to be called after construction.
```

PR #775, opened 2020-06-16
Simplify and Refactor Basis Functions

fsimonis:
This PR:
* simplified the Gaussian similar to #769, but without altering the interface of the ctor.
  As we always use a cut-off threshold for the Gaussian, it always has compact support.
* refactors the compact-supportness of basis functions into 2 classes. This simplifies the implementation and allows for easier lifting of this information to compile time.

Closes #769

### DIFF
--- a/src/mapping/impl/BasisFunctions.hpp
+++ b/src/mapping/impl/BasisFunctions.hpp
@@ -148,15 +148,16 @@ public:
     PRECICE_CHECK(math::greater(_supportRadius, 0.0),
                   "Support radius for radial-basis-function gaussian has to be larger than zero!");
 
-    _deltaY          = hasCompactSupport() ? evaluate(supportRadius) : 0;
+    if (supportRadius < std::numeric_limits<double>::infinity()) {
+      _deltaY = evaluate(supportRadius);
+    }
     double threshold = std::sqrt(-std::log(cutoffThreshold)) / shape;
     _supportRadius   = std::min(supportRadius, threshold);
   }
 
-  /// Compact support if supportRadius is not infinity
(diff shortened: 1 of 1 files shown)


src/mapping/impl/BasisFunctions.hpp
```diff
@@ -148,17 +129,13 @@ class Gaussian {
     PRECICE_CHECK(math::greater(_supportRadius, 0.0),
                   "Support radius for radial-basis-function gaussian has to be larger than zero!");
 
-    _deltaY          = hasCompactSupport() ? evaluate(supportRadius) : 0;
+    if (supportRadius < std::numeric_limits<double>::infinity()) {
```
MakisH, line 132: Why using `std::numeric_limits<double>::infinity()` here but `std::numeric_limits<double>::max()` in the `NoCompactSupportBase`?

fsimonis: Good point.
I could imaging that working with the maximal value of double is a numerical nightmare.
Infinity may be a better pick.

We should probably investigate this in the future.

KyleDavisSA:
I tested all basis functions and compared the L2 error for them. I get the same value with this PR and with the current develop branch.

src/mapping/impl/BasisFunctions.hpp
KyleDavisSA: I agree that `std::numeric_limits<double>::infinity()` would be better.

fsimonis: I'll open an issue regarding this.

PR #776, opened 2020-06-17
Migrate Fortran bindings to use unique_ptr

fsimonis:
This PR:
* Migrates Fortran bindings to use unique_ptr
* Fixes memory leaks in case of missing finalize

Related to #768 

Reviewers: Please test the bindings with your Fortran codes.


uekerman:
Fortran module including solverdummy works as well. Merging.

PR #780, opened 2020-06-29
Add oversubscribe flag to test environment

fsimonis:
This PR sets the environment variable `OMPI_MCA_rmaps_base_oversubscribe=1` when executing the tests.
This allows over-subscription for Open MPI, which is relevant when running the tests on hardware with less than four physical cores.


DavidSCN:
Tested on my machine with only two physical cores and works perfectly fine (apart from the known to fail mpiports).
Thanks for the quick fix @fsimonis!

Issue #706, opened 2020-03-27
macOS compilation error with boost


oguzziya:
Operating system macOS Catalina 10.15.3
preCICE Version: 2.0.1

During the compilation process, the error thrown as:
 ```
#error "Boost.Stacktrace requires _Unwind_Backtrace function. Define _GNU_SOURCE macro or BOOST_STACKTRACE_GNU_SOURCE_NOT_REQUIRED if _Unwind_Backtrace is available without _GNU_SOURCE."
```

When I tried to define `_GNU_SOURCE` or `BOOST_STACKTRACE_GNU_SOURCE_NOT_REQUIRED`, the error disappeared and turned into undefined references.

Here, [in the boost repo](https://github.com/boostorg/stacktrace/pull/70), similar error is reported for different Apple related OS and therefore i am not sure this is directly preCICE related error.

Issue #539, opened 2019-10-24
Compile errors macOS


arusch:
Hi,

after not having compiled preCICE on macOS for quite some time, I recently tried building preCICE (v1.6.1) under macOS 10.14 and 10.15.
I ran into multiple errors of the following type:

```
[ 40%] Building CXX object CMakeFiles/precice.dir/src/mesh/Mesh.cpp.o
In file included from /Users/<MASKED>/Software/precice/src/mesh/Mesh.cpp:1:
In file included from /Users/<MASKED>/Software/precice/src/mesh/Mesh.hpp:3:
In file included from /Users/<MASKED>/Software/precice/src/mesh/Group.hpp:4:
In file included from /Users/<MASKED>/Software/precice/src/utils/PointerVector.hpp:4:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/vector:275:
In file included from /Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/__bit_reference:16:
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/algorithm:1297:14: error: no matching function for call to object of type
      'std::__1::__equal_to<precice::mesh::Vertex *, precice::mesh::Vertex *>'
        if (!__pred(*__first1, *__first2))
             ^~~~~~
/Applications/Xcode.app/Contents/Developer/Toolchains/XcodeDefault.xctoolchain/usr/bin/../include/c++/v1/algorithm:1345:19: note: in instantiation of function template specialization
      'std::__1::is_permutation<boost::iterators::indirect_iterator<std::__1::__wrap_iter<precice::mesh::Vertex *const *>, precice::mesh::Vertex *, boost::use_default, precice::mesh::Vertex
      &, boost::use_default>, boost::iterators::indirect_iterator<std::__1::__wrap_iter<precice::mesh::Vertex *const *>, precice::mesh::Vertex *, boost::use_default, precice::mesh::Vertex
      &, boost::use_default>, std::__1::__equal_to<precice::mesh::Vertex *, precice::mesh::Vertex *> >' requested here
    return _VSTD::is_permutation(__first1, __last1, __first2, __equal_to<__v1, __v2>());
                  ^
[...]

PR #783, opened 2020-07-03
Fix includes of codebase

fsimonis:
This is the result of running include-what-you-use, manual cleanup and
format-all.

Many of our files depended on the internal STL includes, which lead to failing builds on Mac OS.


uekerman:
Probably connected to https://github.com/precice/precice/issues/706 ?

uekerman:
and https://github.com/precice/precice/issues/539 ?

fsimonis:
@uekerman This is likely connected to #539.
The boost stacktrace issue has a different course.

PR #785, opened 2020-07-06
Use RTree intersects() with BoundingBoxes

### LINKED ISSUE #762: Use RTree intersects() with BoundingBox
**Describe your setup**

Operating system (e.g. Linux distribution and version): _Arch Linux_
preCICE Version: develop 657f759

**Describe the problem**
The tagging of RBF mappings currently use `satisfies()` to query vertices contained in a bounding box.
This is slow as the tree has to apply the predicate to every vertex and cannot take advantage of the actual data structure.
We should adopt the BoundingBox type and use `intersects()` instead.

**Expected behaviour**
Take advantage of the rtree.

**Additional context**

https://github.com/precice/precice/blob/9a4652c3b092e0959df76d66bb090aa4136883b4/src/mapping/PetRadialBasisFctMapping.hpp#L824-L827

PR #739 adds a second call-site

fsimonis:
This PR adds a transformation from `mesh::BoundingBox` to `mesh::RTreeBox` and uses it with `boost::geometry::index::intersects()` instead of using `mesh::BoundingBox::contains()` as a predicate.
This restores the lookup complexity of `O(log(n))` which was lost in #712.

The change from `within()` to `intersects()` allows for cleaner usage as the BoundingBox of a mesh now sellects all vertices. This was not the case using `within()`.

Closes #762

### DIFF
--- a/src/mapping/PetRadialBasisFctMapping.hpp
+++ b/src/mapping/PetRadialBasisFctMapping.hpp
@@ -11,6 +11,7 @@
 #include "impl/BasisFunctions.hpp"
 #include "math/math.hpp"
 #include "mesh/RTree.hpp"
+#include "mesh/impl/BBUtils.hpp"
 #include "precice/impl/versions.hpp"
 #include "utils/Petsc.hpp"
 namespace petsc = precice::utils::petsc;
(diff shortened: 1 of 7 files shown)


oguzziya:
Everything seems fine for me 👍

PR #790, opened 2020-07-09
Clarify the CMake output for enabled features

### LINKED ISSUE #789: Confusing CMake output on enabled features

Oh, hi there, I feel flattered to be giving an interview. Yes, aaah, my problem is that I always get confused by the following CMake output:

```
-- === ENABLED FEATURES ===
 * MPICommunication, Enables the MPI communication back-end.

   This enables the MPI communication back-end which is highly recommended on multi-node systems.
   See the documentation of the CMake module FindMPI to control its functionality.

   Set PRECICE_MPICommunication to enable.
```

On the one hand, the header specifies that these features are enabled, on the other hand the message implies that I still need to enable them.

This is defined in our CMakeLists.txt:
https://github.com/precice/precice/blob/1ea9d2dfb46edbf83ad19e820dcd1580e8c6cde2/CMakeLists.txt#L89-L130

I am about to submit a PR to change the formulation to the following:
```
   This feature can be enabled/disabled by setting the PRECICE_MPICommunication CMake option
```

Ignore it because this formulation is important for a reason I don't currently see.

MakisH:
This replaces formulations such as:
```
Set PRECICE_MPICommunication to enable.
```

with:
```
This feature can be enabled/disabled by setting the PRECICE_MPICommunication CMake option.
```

Comments:
- Does not start with a verb, as it should not necesarily trigger an action from the user, it should only inform them about the option.
- Uses both "enable" and "disable", as it can appear in both states.
- Clarifies that this is a CMake option, because unexperienced users may confuse it with an environment variable (which would also work, but this is not the purpose of the instruction).

Closes #789.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -92,14 +92,14 @@ add_feature_info(MPICommunication PRECICE_MPICommunication
    This enables the MPI communication back-end which is highly recommended on multi-node systems.
    See the documentation of the CMake module FindMPI to control its functionality.
 
-   Set PRECICE_MPICommunication to enable.
+   This feature can be enabled/disabled by setting the PRECICE_MPICommunication CMake option.
 ")
 add_feature_info(PETScMapping PRECICE_PETScMapping
   "Enables the PETSc-powered radial basic function mappings.
 
    The radial basis function mappings require MPI and PETSc to work in parallel.
 
-   Set PRECICE_PETScMapping to enable.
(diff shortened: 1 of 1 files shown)


fsimonis:
This is definitely easier to understand now!

> Clarifies that this is a CMake option, because unexperienced users may confuse it with an environment variable (which would also work, but this is not the purpose of the instruction).

Environment variables have no influence on user-defined CMake options. So this PR prevent mistakes :+1:

PR #791, opened 2020-07-09
Add CII Best practices badge

MakisH:
Pointing to https://bestpractices.coreinfrastructure.org/en/projects/3895

This is based on recent input from @fsimonis (mainly) and me (a bit) on the respective "passing" level of the badge. We plan to extend the data to give a clear image on how we also conform to the silver/gold levels.

@uekerman check the location:

Screenshot from 2020-07-09 10-29-09

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -21,9 +21,6 @@
 <a style="text-decoration: none" href="https://github.com/precice/precice/releases/latest" target="_blank">
     <img src="https://img.shields.io/github/release/precice/precice.svg" alt="Release">
 </a>
-<a style="text-decoration: none" href="https://bestpractices.coreinfrastructure.org/projects/3895" target="_blank">
-    <img src="https://bestpractices.coreinfrastructure.org/projects/3895/badge" alt="CII Best Practices">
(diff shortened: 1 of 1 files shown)


uekerman:
Cool thing!
I guess you already had second thoughts about the location and that's why you ask :smile: 

What about making "Code Quality" to "Software Quality" and then adding the badge there?
In a way "codecov" is also not 100% code quality, right?

MakisH:
Indeed, I had second thoughts. It is a quality metric, but it is not a code quality metric per se.

Similarly, the "project status" currently has two kind of items: "lifetime status" and "current health status". If we rename the "code quality" to "project quality", then it would make more sense to move it there. "Software quality" also works.

uekerman:
Both are fine for me, you choose.

MakisH:
@uekerman moved to "Project quality"
@fsimonis any objections?

Screenshot from 2020-07-09 17-21-27

fsimonis:
I don't have a strong opinion on this.

It just looks weird alongside all the code related badges.

PR #794, opened 2020-07-10
Make the system tests badge static

MakisH:
With our current implementation of system tests, it is not easy to get the system tests status for each branch of preCICE. This makes it always blue and "system tests: check", pointing to the build history page: https://travis-ci.org/github/precice/systemtests/builds

Closes #664, offering an alternative solution to the same problem.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -24,8 +24,8 @@
 <a style="text-decoration: none" href="https://travis-ci.org/precice/precice" target="_blank">
     <img src="https://travis-ci.org/precice/precice.svg?branch=develop" alt="Build status">
 </a>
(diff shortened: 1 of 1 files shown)


MakisH:
Then indeed we should indeed create an issue that summarizes the actual problem, with our current insight. @BenjaminRueth since you opened the PR, could you also open the summary issue?  I see #664 as a dead-end at the moment, therefore we should close it.

BenjaminRueth:
I created an issue https://github.com/precice/precice/issues/798. That issue is only about the "delta" between #664 and #794, i.e. the parts that we are missing, if we merge #794 and close #664.

PR #796, opened 2020-07-10
Set BUILD_SHARED_LIBS to True by default

fsimonis:

This PR changes the default value of the option for `BUILD_SHARED_LIBS` to `ON`.

**Motivation**

We constantly run into issues with users (and developers) running into issues with a statically build preCICE library.
Sometimes new users don't read the documentation and sometimes seasoned users simply don't remember it.

Setting the `option(` to `On` has the following effects:
* It does not override the setting if is present in the CMakeCache.
* It defaults to `ON` when the user did not set anything.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,7 +58,7 @@ option(PRECICE_PETScMapping "Enable use of the PETSc linear algebra library." ON
 option(PRECICE_PythonActions "Python support" ON)
 option(PRECICE_Packages "Configure package generation." ON)
 option(PRECICE_InstallTest "Add test binary and necessary files to install target." OFF)
-option(BUILD_SHARED_LIBS "Build shared libraries by default" OFF)
+option(BUILD_SHARED_LIBS "Build shared libraries by default" ON)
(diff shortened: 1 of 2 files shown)


uekerman:
Isn't this contradicting your previous opinion?
"Following standards more important than being simple"

I have a bit mixed feeling to do this in a minor release. I guess users expect that building works the same way for every minor release.

MakisH:
Do we currectly know of any users that build a static library (on purpose)?

Also, since we had already the discussion a few times, which is the standard we are sticking to?

uekerman:
> Do we currectly know of any users that build a static library (on purpose)?

That's information we have never collected. I guess we don't know. At least I don't have a feeling for this at all.

MakisH:
What we know is that we don't test the static library and some adapters should not work with static out of the box.

uekerman:
I am not religious about this. There are good reasons to make the shared lib the default, I agree. I am just a bit puzzled since we had this exact discussion before releasing v2.0 and decided the opposite way :smile: :see_no_evil: 

> Also, since we had already the discussion a few times, which is the standard we are sticking to?

That's @fsimonis' expertise

fsimonis:
This was a bit of a snap reaction to a more in-depth problem. Let's shift it to an issue.

PR #799, opened 2020-07-11
Convergence measure logging overhaul

### LINKED ISSUE #786: False assertion in TXTTableWriter
Is this assertion really needed?

https://github.com/precice/precice/blob/fa5e664ea6dc464f29341b356d27a425787346a1/src/io/TXTTableWriter.cpp#L33

It prevents the definition of multiple convergence measures for one data field, for example:
```xml
         <absolute-convergence-measure limit="1e-5" data="DataTwo" mesh="MeshOne"/>
         <relative-convergence-measure limit="1e-3" data="DataTwo" mesh="MeshOne"/>
```

If yes, we should find a solution to nevertheless allow for that. Or, in the worst case, prevent false usage in the configuration.

uekerman:
In was about time to clean up our convergence measure logging as more and more unsatisfactory compromises piled up. 

This PR ...

1. increases readability of relative convergence measure INFO logs

before:
```
relative convergence measure: two-norm diff = 2.6036e-08, relative limit = 1.0005e-08, conv = true
```
Problems: the relative limit changes in every iteration and the (absolute) two-norm diff does not match the entry in the convergence file. Pointed out by @ajaust.   

after:
```
relative convergence measure: relative two-norm diff = 2.6023e-05, limit = 1e-05, normalization = 0.00100051, conv = false
```

2. adds a unit test for `ResidualRelativeConvergenceMeasure`
3. remove the unnecessary assertion in TXTTableWriter (closes #786)
4. instead, extends the title headers of the convergence file by convergence measure abbreviations

Now, we have for example ...
```
TimeWindow  Iteration  ResRel(Pressure)  ResAbs(CrossSectionLength)  ResDrop(CrossSectionLength)  
1  1  1.0000000000000000  0.0000015151688798  1.0000000000000000  
1  2  0.0092426077637776  0.0000014861770983  0.9808656435350955  
1  3  0.8894974714023499  0.0000000036011503  0.0023767319918400  
1  4  0.0013652823318746  0.0000000000147103  0.0000097086996980  
1  5  0.0000009272349971  0.0000000000000499  0.0000000329390337
```

... for ... 
```xml
      <relative-convergence-measure data="Pressure" mesh="Fluid_Nodes" limit="1e-5"/>
      <absolute-convergence-measure data="CrossSectionLength" mesh="Structure_Nodes" limit="1e-5"/>
      <residual-relative-convergence-measure data="CrossSectionLength" mesh="Structure_Nodes" limit="1e-3"/>  
```

 <min-iteration-convergence-measure data="CrossSectionLength" mesh="Structure_Nodes" min-iterations="2"/>

5. suppresses convergence file logging for min-iterations convergence measures as it really makes no sense here

6. removes logging of average convergence rates in the iterations file. In my opinion, they had no practical use case at all. I don't think that any user ever looked at them and we, ourselves, did never use them in any publication.  

7. instead, extends the iterations file logging by the total and the dropped number of QN columns. In my opinion, this is very helpful information to better understand how well the QN coupling works. 

```
TimeWindow  TotalIterations  Iterations  Convergence  QNColumns  DeletedQNColumns  DroppedQNColumns  
1  5  5  1  4  0  0  
2  9  4  1  6  1  0  
3  12  3  1  7  1  0  
4  15  3  1  9  0  0  
5  18  3  1  10  1  0  
6  21  3  1  11  1  0  
7  24  3  1  12  1  0  
8  27  3  1  13  1  0  
9  30  3  1  13  2  0  
10  33  3  1  14  1  0  
11  36  3  1  15  1  0  
12  39  3  1  15  1  1  
13  42  3  1  15  1  1  
```

Deleted columns are columns that are filtered out. Dropped columns are columns that went out of scope (due to `max-used-iterations value` or `time-windows-reused value`. 5 iterations means 4 new columns (columns are diffs between iterations).

# Reviewers:

* @fsimonis Please check code quality
* @MakisH Please check usability with any tutorial

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -138,7 +138,7 @@ void BaseCouplingScheme::initialize(double startTime, int startTimeWindow)
       // merge send and receive data for all pp calls
       mergeData();
       // setup convergence measures
-      for (ConvergenceMeasure &convergenceMeasure : _convergenceMeasures) {
+      for (ConvergenceMeasureContext &convergenceMeasure : _convergenceMeasures) {
         int dataID = convergenceMeasure.data->getID();
         assignDataToConvergenceMeasure(&convergenceMeasure, dataID);
       }
@@ -476,7 +476,7 @@ void BaseCouplingScheme::setupDataMatrices(DataMap &data)
   PRECICE_TRACE();
   PRECICE_DEBUG("Data size: " << data.size());
   // Reserve storage for convergence measurement of send and receive data values
-  for (ConvergenceMeasure &convMeasure : _convergenceMeasures) {
+  for (ConvergenceMeasureContext &convMeasure : _convergenceMeasures) {
     PRECICE_ASSERT(convMeasure.couplingData != nullptr);
     if (convMeasure.couplingData->oldValues.cols() < 1) {
       utils::append(convMeasure.couplingData->oldValues,
@@ -505,7 +505,7 @@ void BaseCouplingScheme::setAcceleration(
 void BaseCouplingScheme::newConvergenceMeasurements()
 {
   PRECICE_TRACE();
-  for (ConvergenceMeasure &convMeasure : _convergenceMeasures) {
+  for (ConvergenceMeasureContext &convMeasure : _convergenceMeasures) {
     PRECICE_ASSERT(convMeasure.measure.get() != nullptr);
     convMeasure.measure->newMeasurementSeries();
   }
@@ -517,7 +517,7 @@ void BaseCouplingScheme::addConvergenceMeasure(
     impl::PtrConvergenceMeasure measure,
     bool                        doesLogging)
 {
-  ConvergenceMeasure convMeasure;
+  ConvergenceMeasureContext convMeasure;
   convMeasure.data         = std::move(data);
   convMeasure.couplingData = nullptr;
   convMeasure.suffices     = suffices;
@@ -538,7 +538,7 @@ bool BaseCouplingScheme::measureConvergence()
     _convergenceWriter->writeData("Iteration", _iterations);
   }
   for (size_t i = 0; i < _convergenceMeasures.size(); i++) {
-    ConvergenceMeasure &convMeasure = _convergenceMeasures[i];
+    ConvergenceMeasureContext &convMeasure = _convergenceMeasures[i];
 
     PRECICE_ASSERT(convMeasure.couplingData != nullptr);
     PRECICE_ASSERT(convMeasure.measure.get() != nullptr);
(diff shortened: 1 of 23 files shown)


src/cplscheme/impl/ConvergenceMeasure.hpp
```diff
@@ -48,6 +48,12 @@ class ConvergenceMeasure {
   {
     return 0;
   }
+
+  /// Returns an abreviation of the name of the measure for the log file headers
+  virtual std::string getAbbreviation() const
+  {
+    return "";
+  }
```
fsimonis, line 56: You could also force this to be provided by the convergence measure by making it abstract.
```suggestion
  virtual std::string getAbbreviation() const = 0;
```

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -592,13 +590,18 @@ void BaseCouplingScheme::initializeTXTWriters()
...
       for (ConvergenceMeasure &convMeasure : _convergenceMeasures) {
-        std::stringstream sstm, sstm2;
-        sstm << "AvgConvRate(" << convMeasure.data->getName() << ")";
-        sstm2 << "ResNorm(" << convMeasure.data->getName() << ")";
-        _iterationsWriter->addData(sstm.str(), io::TXTTableWriter::DOUBLE);
-        _convergenceWriter->addData(sstm2.str(), io::TXTTableWriter::DOUBLE);
+
+        if (convMeasure.doesLogging) {
+          std::stringstream sstm2;
+          sstm2 << "Res" << convMeasure.measure->getAbbreviation() << "(" << convMeasure.data->getName() << ")";
```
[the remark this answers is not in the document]
uekerman: The data name is not yet there. Would you still do it? I guess there is always the conflict whether things should go into `BaseCouplingScheme::ConvergenceMeasure` or `ConvergenceMeasure`. Already writing this feels awkward.

src/cplscheme/impl/ConvergenceMeasure.hpp
uekerman: I copied from `getNormResidual()`. Problem is that both functions make no sense for `MinIterationsConvergenceMeasure`. Should get cleaner with #800

src/cplscheme/BaseCouplingScheme.cpp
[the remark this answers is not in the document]
fsimonis: This seems similar to the problem of `M2N` and `BoundM2N`.
I definitely agree that the current wording is confusing at best. We should find a better name for the BaseCouplingScheme wrapper.


As a first step, you could recfactor it to
```
std::string BaseCouplingScheme::ConvergenceMeasure::logHeader() const;
```
Returning a string also allows you to change the call-side to a one-liner.

src/cplscheme/impl/ConvergenceMeasure.hpp
fsimonis: Ok, let's refactor both together in the future.

PR #806, opened 2020-07-15
Errors 2.2 - Steering

BenjaminRueth:
Moves `PRECICE_CHECKS` on state into `SolverInterfaceImpl` and either removes unreachable `PRECICE_CHECKS` or turns them into `PRECICE_ASSERT`.

Resolves section 2.2 in https://github.com/precice/precice/issues/698

Reviewers:

@fsimonis Does this PR extend the `State` in an appropriate way?
@uekerman Are the assumptions on states valid? We get some new public functions in `cplscheme`. Is this reasonable?

### DIFF
--- a/src/cplscheme/CompositionalCouplingScheme.cpp
+++ b/src/cplscheme/CompositionalCouplingScheme.cpp
@@ -44,7 +44,7 @@ bool CompositionalCouplingScheme::isInitialized() const
 bool CompositionalCouplingScheme::sendsInitializedData() const
 {
   PRECICE_TRACE();
-  bool sendsInitializedData = true;
+  bool sendsInitializedData = false;
   for (Scheme scheme : _couplingSchemes) {
     sendsInitializedData |= scheme.scheme->sendsInitializedData();
   }
@@ -55,7 +55,7 @@ bool CompositionalCouplingScheme::sendsInitializedData() const
 bool CompositionalCouplingScheme::receivesInitializedData() const
 {
   PRECICE_TRACE();
-  bool receivesInitializedData = true;
+  bool receivesInitializedData = false;
   for (Scheme scheme : _couplingSchemes) {
     receivesInitializedData |= scheme.scheme->receivesInitializedData();
   }
--- a/src/cplscheme/CompositionalCouplingScheme.cpp
+++ b/src/cplscheme/CompositionalCouplingScheme.cpp
@@ -46,7 +46,7 @@ bool CompositionalCouplingScheme::sendsInitializedData() const
   PRECICE_TRACE();
   bool sendsInitializedData = true;
   for (Scheme scheme : _couplingSchemes) {
-    sendsInitializedData &= scheme.scheme->sendsInitializedData();
+    sendsInitializedData |= scheme.scheme->sendsInitializedData();
   }
   PRECICE_DEBUG("return " << sendsInitializedData);
   return sendsInitializedData;
@@ -57,7 +57,7 @@ bool CompositionalCouplingScheme::receivesInitializedData() const
   PRECICE_TRACE();
   bool receivesInitializedData = true;
   for (Scheme scheme : _couplingSchemes) {
-    receivesInitializedData &= scheme.scheme->receivesInitializedData();
+    receivesInitializedData |= scheme.scheme->receivesInitializedData();
   }
   PRECICE_DEBUG("return " << receivesInitializedData);
   return receivesInitializedData;
--- a/src/cplscheme/CompositionalCouplingScheme.cpp
+++ b/src/cplscheme/CompositionalCouplingScheme.cpp
@@ -41,6 +41,28 @@ bool CompositionalCouplingScheme::isInitialized() const
   return isInitialized;
 }
 
+bool CompositionalCouplingScheme::sendsInitializedData() const
+{
+  PRECICE_TRACE();
+  bool sendsInitializedData = true;
+  for (Scheme scheme : _couplingSchemes) {
+    sendsInitializedData &= scheme.scheme->sendsInitializedData();
+  }
+  PRECICE_DEBUG("return " << sendsInitializedData);
+  return sendsInitializedData;
+}
+
+bool CompositionalCouplingScheme::receivesInitializedData() const
+{
+  PRECICE_TRACE();
+  bool receivesInitializedData = true;
+  for (Scheme scheme : _couplingSchemes) {
+    receivesInitializedData &= scheme.scheme->receivesInitializedData();
+  }
+  PRECICE_DEBUG("return " << receivesInitializedData);
+  return receivesInitializedData;
+}
+
 void CompositionalCouplingScheme::initializeData()
 {
   PRECICE_TRACE();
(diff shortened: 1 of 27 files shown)


BenjaminRueth:
PR is ready for review. I provoked all the errors mentioned in #698 with the C++ solverdummy.

I also introduced some new `PRECICE_CHECKS`. I did not check those in detail, but they are mainly sanity checks.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -934,6 +967,8 @@ void SolverInterfaceImpl::writeBlockVectorData(
     const double *values)
 {
   PRECICE_TRACE(fromDataID, size);
+  PRECICE_CHECK(_state != State::Constructed, "writeBlockVectorData(...) has to be called after initialize().");
```
uekerman, line 970:
```suggestion
  PRECICE_CHECK(_state != State::Constructed, "writeBlockVectorData(...) can only be called after initialize().");
```
Otherwise it sounds like the function should only be called once. And that is has to be called.
Same for other examples below.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -934,6 +967,8 @@ void SolverInterfaceImpl::writeBlockVectorData(
...
+  PRECICE_CHECK(_state != State::Finalized, "writeBlockVectorData(...) has to be called before finalize().");
```
uekerman, line 971:
```suggestion
  PRECICE_CHECK(_state != State::Finalized, "writeBlockVectorData(...) cannot be called after finalize().");
```

src/cplscheme/CompositionalCouplingScheme.cpp
```diff
@@ -41,6 +41,28 @@ bool CompositionalCouplingScheme::isInitialized() const
   return isInitialized;
 }
 
+bool CompositionalCouplingScheme::sendsInitializedData() const
+{
+  PRECICE_TRACE();
+  bool sendsInitializedData = true;
+  for (Scheme scheme : _couplingSchemes) {
+    sendsInitializedData &= scheme.scheme->sendsInitializedData();
```
uekerman, line 49:
```suggestion
    sendsInitializedData |= scheme.scheme->sendsInitializedData();
```
right?

src/cplscheme/CompositionalCouplingScheme.cpp
```diff
@@ -41,6 +41,28 @@ bool CompositionalCouplingScheme::isInitialized() const
...
+  PRECICE_DEBUG("return " << sendsInitializedData);
+  return sendsInitializedData;
+}
+
+bool CompositionalCouplingScheme::receivesInitializedData() const
+{
+  PRECICE_TRACE();
+  bool receivesInitializedData = true;
+  for (Scheme scheme : _couplingSchemes) {
+    receivesInitializedData &= scheme.scheme->receivesInitializedData();
```
uekerman, line 60:
```suggestion
    receivesInitializedData |= scheme.scheme->receivesInitializedData();
```
right?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -430,6 +443,7 @@ void SolverInterfaceImpl::finalize()
 {
   PRECICE_TRACE();
   PRECICE_CHECK(_state != State::Finalized, "finalize() may only be called once.")
+  PRECICE_CHECK(_state != State::Constructed, "initialize() has to be called before finalize().")
```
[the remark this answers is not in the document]
BenjaminRueth: Surprisingly no... the test did not fail.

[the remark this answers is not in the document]
BenjaminRueth: Correction: It seems like something was wrong on my side. I ran the tests again and now this test fails (as expected).

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -229,6 +229,7 @@ double SolverInterfaceImpl::initialize()
 {
   PRECICE_TRACE();
   PRECICE_CHECK(_state == State::Constructed, "initialize() may only be called once.");
+  PRECICE_CHECK(_state != State::Finalized, "initialize() cannot be called after finalize().")
```
fsimonis, line 232: This is already covered by the proceeding check.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -361,8 +370,12 @@ double SolverInterfaceImpl::advance(
...
   utils::ScopedEventPrefix sep("advance/");
 
-  PRECICE_CHECK(_couplingScheme->isInitialized(), "initialize() has to be called before advance()");
-  PRECICE_CHECK(isCouplingOngoing(), "advance() cannot be called when isCouplingOngoing() returns false");
+  PRECICE_CHECK(_state != State::Constructed, "initialize() has to be called before advance().");
+  PRECICE_CHECK(_state != State::Finalized, "advance() cannot be called after finalize().")
+  PRECICE_ASSERT(_couplingScheme->isInitialized());
+  PRECICE_CHECK(isCouplingOngoing(), "advance() cannot be called when isCouplingOngoing() returns false.");
+  PRECICE_CHECK((not _couplingScheme->receivesInitializedData() && not _couplingScheme->sendsInitializedData()) || (_state == State::InitializedData),
+                "initializeData() needs to be called before advance if data has to be initialized.");
```
fsimonis, line 378: Can you add the isCouplingOngoing() and reword the coupling data parts in the preconditions part of the documentation? (See SolverInterface.hpp)

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRueth: I would actually suggest to swap the two checks. Then we have more helpful error messages. We would still catch everything by checking for `_state == State::Constructed`, but there are different causes for `_state == State::Constructed` failing:

1. `initialize()` called multiple times
2. `initialize()` called after finalize

I think we should also provide two different error messages.

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRueth: Does https://github.com/precice/precice/commit/eab65f3d394dfb05c1b44e0b8901e92b4e997a12 look ok?

uekerman:
> @uekerman According to this scheme, using read/write data functions before `initialize()` should be illegal too. Is that correct?

Already discussed with @BenjaminRueth, technically that's not a problem, I think, as long as a mesh is filled. Question is whether there is a use case.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -999,8 +990,7 @@ void SolverInterfaceImpl::writeVectorData(
     const double *value)
 {
   PRECICE_TRACE(fromDataID, valueIndex);
-  PRECICE_CHECK(_state != State::Constructed, "writeVectorData(...) has to be called after initialize().");
-  PRECICE_CHECK(_state != State::Finalized, "writeVectorData(...) has to be called before finalize().");
+  PRECICE_CHECK(_state != State::Finalized, "writeVectorData(...) cannot be called before finalize().");
```
uekerman, line 993:
```suggestion
  PRECICE_CHECK(_state != State::Finalized, "writeVectorData(...) cannot be called after finalize().");
```

PR #808, opened 2020-07-15
Errors 1.9 - Technical Config

fsimonis:

* Rewords the error messages related to XML parser, tags and attributes.
* Coordinates in attributes are now required to match the dimensions of the SolverInterface. Previous behaviour was to truncate the last component if the coordinate had more dimensions than required.

**Additional Information**

Related to #698

### DIFF
--- a/src/mesh/config/DataConfiguration.cpp
+++ b/src/mesh/config/DataConfiguration.cpp
@@ -63,7 +63,7 @@ void DataConfiguration::xmlTagCallback(
     int         dataDimensions = getDataDimensions(typeName);
     addData(name, dataDimensions);
   } else {
-    PRECICE_ERROR("Received callback from tag " << tag.getName());
+    PRECICE_ASSERT(false, "Received callback from unknown tag " << tag.getName());
   }
 }
 
--- a/src/xml/ConfigParser.cpp
+++ b/src/xml/ConfigParser.cpp
@@ -1,5 +1,6 @@
 #include "ConfigParser.hpp"
 #include <algorithm>
+#include <exception>
 #include <fstream>
 #include <iterator>
 #include <libxml/SAX.h>
@@ -91,8 +92,8 @@ ConfigParser::ConfigParser(const std::string &filePath, const ConfigurationConte
 
   try {
     connectTags(context, DefTags, SubTags);
-  } catch (const std::string &error) {
-    PRECICE_ERROR(error);
+  } catch (const std::exception &e) {
+    PRECICE_ERROR("And unexpected exception occurred during configuration: " << e.what() << '.');
   }
 }
 
@@ -157,7 +158,7 @@ void ConfigParser::connectTags(const ConfigurationContext &context, std::vector<
         });
 
     if (tagPosition == DefTags.end()) {
-      PRECICE_ERROR("Tag <" + expectedName + "> is unknown");
+      PRECICE_ERROR("The configuration contains an unknown tag <" + expectedName + ">.");
     }
 
     auto pDefSubTag = *tagPosition;
@@ -165,7 +166,7 @@ void ConfigParser::connectTags(const ConfigurationContext &context, std::vector<
 
     if ((pDefSubTag->_occurrence == XMLTag::OCCUR_ONCE) || (pDefSubTag->_occurrence == XMLTag::OCCUR_NOT_OR_ONCE)) {
       if (usedTags.count(pDefSubTag->_fullName)) {
-        PRECICE_ERROR("Tag <" + pDefSubTag->_fullName + "> is already used");
+        PRECICE_ERROR("Tag <" + pDefSubTag->_fullName + "> is not allowed to occur multiple times.");
       }
       usedTags.emplace(pDefSubTag->_fullName);
     }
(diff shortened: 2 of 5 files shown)


src/xml/ConfigParser.cpp
```diff
@@ -91,8 +92,8 @@ ConfigParser::ConfigParser(const std::string &filePath, const ConfigurationConte
 
   try {
     connectTags(context, DefTags, SubTags);
-  } catch (const std::string &error) {
-    PRECICE_ERROR(error);
+  } catch (const std::exception &e) {
+    PRECICE_ERROR("And unexpected exception occurred during configuration: " << e.what() << '.');
```
MakisH, line 96:
```suggestion
    PRECICE_ERROR("An unexpected exception occurred during configuration: " << e.what() << '.');
```

src/xml/XMLAttribute.hpp
```diff
@@ -155,13 +155,13 @@ void XMLAttribute<ATTRIBUTE_T>::readValue(const std::map<std::string, std::strin
...
   if (_read) {
-    PRECICE_ERROR("Attribute \"" + _name + "\" is defined multiple times");
+    PRECICE_ERROR("Attribute \"" + _name + "\" has already been defined. Duplications are not permitted.");
   }
 
   const auto position = aAttributes.find(getName());
   if (position == aAttributes.end()) {
     if (not _hasDefaultValue) {
-      PRECICE_ERROR("Attribute \"" + _name + "\" missing");
+      PRECICE_ERROR("Attribute \"" + _name + "\" is required, but was not defined.");
```
MakisH, line 164: It would be good if here we also had some context: The user only gets a message that "you also need X", without defining where. As often this would not happen because of a simple typo (but because something is really missing), this would be tricky to debug.

MakisH, line 164: Interestingly, omitting the `dimensions` in the `<solver-interface>` does not trigger this. Is there a default value?

src/mesh/config/DataConfiguration.cpp
```diff
@@ -63,7 +63,7 @@ void DataConfiguration::xmlTagCallback(
     int         dataDimensions = getDataDimensions(typeName);
     addData(name, dataDimensions);
   } else {
-    PRECICE_ERROR("Received callback from tag " << tag.getName());
+    PRECICE_ASSERT(false, "Received callback from unknown tag " << tag.getName());
```
MakisH, line 66: Not sure how to test this, but it is already a bit clearer than before. I guess that an assertion is a good idea here, as it should be caught earlier. :+1:

src/xml/XMLAttribute.hpp
```diff
@@ -155,13 +155,13 @@ void XMLAttribute<ATTRIBUTE_T>::readValue(const std::map<std::string, std::strin
 {
   PRECICE_TRACE(_name);
   if (_read) {
-    PRECICE_ERROR("Attribute \"" + _name + "\" is defined multiple times");
+    PRECICE_ERROR("Attribute \"" + _name + "\" has already been defined. Duplications are not permitted.");
```
MakisH, line 158: I cannot trigger this, it is always getting caught earlier, as e.g.: `ERROR: Attribute mesh redefined`

Possible candidate for converting to an assertion.

src/xml/XMLTag.cpp
```diff
@@ -140,10 +140,10 @@ Eigen::VectorXd XMLTag::getEigenVectorXdAttributeValue(const std::string &name,
...
   auto iter = _eigenVectorXdAttributes.find(name);
   PRECICE_ASSERT(iter != _eigenVectorXdAttributes.end());
-  PRECICE_CHECK(iter->second.getValue().size() >= dimensions,
-                "Vector attribute \"" << name << "\" of tag <" << getFullName()
-                                      << "> has less dimensions than required (" << iter->second.getValue().size()
-                                      << " instead of " << dimensions << ")!");
+  const auto size = iter->second.getValue().size();
+  PRECICE_CHECK(size == dimensions,
+                "Vector attribute \"" << name << "\" of tag <" << getFullName() << "> is "
+                                      << size << "D, which does not match the dimension of the " << dimensions << "D SolverInterface.");
```
MakisH, line 146: Suggestion for consistency with the XML tags:
```suggestion
                                      << size << "D, which does not match the dimension of the " << dimensions << "D solver-interface.");
```

src/xml/XMLAttribute.hpp
fsimonis: Yep, it is `2`.

fsimonis: I was thinking about this as well. The problem is that there is no access to the tag from the attribute itself.
So this needs additional work.

src/xml/XMLAttribute.hpp
fsimonis: Good point. I agree with you. This should be an assertion.

src/xml/XMLAttribute.hpp
MakisH: I opened a specific issue (https://github.com/precice/precice/issues/827), let's skip this for now.

PR #810, opened 2020-07-15
Updates error messages for actions 1.3

KyleDavisSA:
This PR improves the error messages for the actions.

### DIFF
--- a/src/action/config/ActionConfiguration.cpp
+++ b/src/action/config/ActionConfiguration.cpp
@@ -235,24 +235,12 @@ void ActionConfiguration::createAction()
       }
     }
   }
-  if (mesh.get() == nullptr) {
-    std::ostringstream stream;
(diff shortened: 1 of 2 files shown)


src/action/config/ActionConfiguration.cpp
```diff
@@ -207,7 +207,7 @@ int ActionConfiguration::getUsedMeshID() const
       return mesh->getID();
     }
   }
-  PRECICE_ERROR("No mesh ID found!");
+  PRECICE_ERROR("No mesh ID found. Please check that the correct mesh name is used.");
```
fsimonis, line 210: This really is internal knowledge and could confuse the user.

```suggestion
  PRECICE_ERROR("No mesh name \"" << _configuredAction.mesh << "\" found. Please check that the correct mesh name is used.");
```

fsimonis:
This can be merged after Travis succeeds.

PR #811, opened 2020-07-15
Bugfix and test for implicit coupling and initialize data

BenjaminRueth:
#689 has introduced a bug that can be reproduced with the integration test included in this PR.


BenjaminRueth:
Note: The actual bugfix is still a todo. But I assume that the bug is somewhere in this function:

https://github.com/precice/precice/blob/53716a549f54ac31b7dc65c198cd6ff307356e09/src/cplscheme/BaseCouplingScheme.cpp#L251-L266

PR #812, opened 2020-07-15
Errors 2.4 - Data access

KyleDavisSA:
This PR improves error messages for data access in `SolverInterfaceImpl`. Related to https://github.com/precice/precice/issues/698

### DIFF
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -958,7 +958,8 @@ void SolverInterfaceImpl::writeBlockVectorData(
   auto &valuesInternal = context.fromData->values();
   for (int i = 0; i < size; i++) {
     const auto valueIndex = valueIndices[i];
-    PRECICE_CHECK(0 <= valueIndex && valueIndex < valuesInternal.size() / context.fromData->getDimensions(), "Value index out of range");
+    PRECICE_CHECK(0 <= valueIndex && valueIndex < valuesInternal.size() / context.fromData->getDimensions(), "Value index out of range. Please check that the size of "
+                  << context.fromData->getName() << " is correct.");
     int offsetInternal = valueIndex * _dimensions;
     int offset         = i * _dimensions;
     for (int dim = 0; dim < _dimensions; dim++) {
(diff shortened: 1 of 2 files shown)


fsimonis:
@KyleDavisSA Can we merge this?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1066,14 +1077,17 @@ void SolverInterfaceImpl::writeScalarData(
   PRECICE_TRACE(fromDataID, valueIndex, value);
   PRECICE_CHECK(_state != State::Finalized, "writeScalarData(...) cannot be called after finalize().");
   PRECICE_VALIDATE_DATA_ID(fromDataID);
-  PRECICE_CHECK(valueIndex >= -1, "Invalid value index (" << valueIndex << ") when writing scalar data!");
+  PRECICE_CHECK(valueIndex >= -1, "Invalid value index (" << valueIndex << ") when writing scalar data. Value index must be >= 0.");
```
MakisH, line 1080: Why changing the phrasing here compared to above? It would also be nice to have the context here, similarly to:

```c++
PRECICE_CHECK(0 <= valueIndex && valueIndex < valuesInternal.size() / context.fromData->getDimensions(), "Value index out of range. Please check that the size of "
<< context.fromData->getName() << " is correct.");
```

MakisH: The same in two more instances below.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1167,14 +1189,17 @@ void SolverInterfaceImpl::readScalarData(
   PRECICE_TRACE(toDataID, valueIndex, value);
   PRECICE_CHECK(_state != State::Finalized, "readScalarData(...) cannot be called after finalize().");
   PRECICE_VALIDATE_DATA_ID(toDataID);
-  PRECICE_CHECK(valueIndex >= -1, "Invalid value index ( " << valueIndex << " )when reading vector data!");
+  PRECICE_CHECK(valueIndex >= -1, "Invalid value index ( " << valueIndex << " ) when reading vector data. Value index must be >= 0.");
```
MakisH, line 1192: Should be `when reading scalar data`. See also the suggestion above.

PR #813, opened 2020-07-15
C bindings: Add missing getMeshIDs

MakisH:
This ports the C++ API method `getMeshIDs` to the C bindings.

The original returns an `std::set`:
https://github.com/precice/precice/blob/493f85ab723fe35d1ad4a484f9cd6d7e08266fd0/src/precice/SolverInterface.hpp#L340-L345

The C bindings return an int pointer as an argument, similar to the style we are using e.g. for `getMeshVertices`. Some serialization is also needed:
```c++
void precicec_getMeshIDs(int *ids)
{
  size_t i = 0;
  for (auto const &id : impl->getMeshIDs()) {
    ids[i++] = id;
  }
}
```

You can check this e.g. by adding the following to the C solverdummy:
```c
int *   meshIDs  = malloc(3 * sizeof(double));

precicec_getMeshIDs(meshIDs);
for (int i = 0; i < 3; i++) {
  printf("### MESH ID: %d \n", meshIDs[i]);
}
```
I am still not really sure about a use case here, especially since in C we would need to somehow know the number of meshes beforehand.

@fsimonis please comment on these snippets and especially the conversion from `std::set` to `int*`.
@uekerman do we actually want to integrate this?

Related to #520.

### DIFF
--- a/extras/bindings/c/src/SolverInterfaceC.cpp
+++ b/extras/bindings/c/src/SolverInterfaceC.cpp
@@ -166,6 +166,14 @@ int precicec_getMeshID(const char *meshName)
   return impl->getMeshID(stringMeshName);
 }
 
+void precicec_getMeshIDs(int *ids)
+{
+  size_t i = 0;
+  for (auto const &id : impl->getMeshIDs()) {
+    ids[i++] = id;
+  }
+}
+
 int precicec_hasData(const char *dataName, int meshID)
 {
   PRECICE_CHECK(impl != nullptr, errormsg);
--- a/docs/changelog/813.md
+++ b/docs/changelog/813.md
@@ -1,0 +1,1 @@
+* Port the API method `getMeshIDs` to the C bindings as `precicec_getMeshIDs`.
--- a/extras/bindings/c/include/precice/SolverInterfaceC.h
+++ b/extras/bindings/c/include/precice/SolverInterfaceC.h
@@ -164,6 +164,13 @@ int precicec_hasMesh(const char *meshName);
 int precicec_getMeshID(const char *meshName);
 
 /**
+ * @brief Returns the ids of all used meshes by this participant.
+ *
+ * @param[out] ids the mesh ids
+ */
+void precicec_getMeshIDs(int *ids);
+
+/**
  * @brief Creates a mesh vertex
  *
  * @param[in] meshID the id of the mesh to add the vertex to.


fsimonis:
Writing a safe interface to something like this in C is actually everything other than trivial.

If the caller manages the maximum size of the object, then we have to respect this maximum. Example: [strncpy](https://en.cppreference.com/w/cpp/string/byte/strncpy)
If we handle the allocated size, then we have to also tell the caller how many elements the result contains.
This can be implemented by a sentinel value such as `\0` in c strings. By returning a size.

So an interface without introducing an additional struct would look like this:
```c
int precicec_getMeshIDs(int ** destination);

int * meshIDs;
int size = precicec_getMeshIDs(&meshIDs);
// use meshIDs and size .... somehow
free(meshIDs);
```

extras/bindings/c/src/SolverInterfaceC.cpp
```diff
@@ -166,6 +166,14 @@ int precicec_getMeshID(const char *meshName)
   return impl->getMeshID(stringMeshName);
 }
 
+void precicec_getMeshIDs(int *ids)
```
fsimonis, line 169: Sadly not safe.

fsimonis:
We could also take do a U-turn here and declare the function as deprecated.
`std::set<std::string> getMeshNames() const` would have an actual use-case.

MakisH:
Even though the interoperability problem is very interesting, I think we risk introducing more bugs with an API function that we don't really want to advertise (as it offers an alternative approach than the one we are teaching to users). So, I upvote the deprecation idea.

I think we cannot avoid passing the mesh name right now: in any case, the adapter would need to read it from its own config, together with information such as what kind of data to write/read through it and which boundaries to associate with it. Unless we manage to completely get rid of the adapter configs (see #474), I don't see a use case for a method that avoids asking the name, or takes all the ids/names together. Do I miss something?

extras/bindings/c/src/SolverInterfaceC.cpp
MakisH: But it is consistent with the rest of our C bindings. Don't we have the same problem there?

If we want to keep this method, let's keep it consistent to the rest. Otherwise, let's rewrite the C bindings to be safer.

uekerman:
Let's deprecate the function in the C++ API and not add it to the bindings.

MakisH:
Let's continue the deprecation discussion here: https://github.com/precice/precice/issues/821

PR #817, opened 2020-07-16
Errors 2.3 - Mesh Definition

fsimonis:

* Reword mesh-related error messages.
* Refactor common error messages into dedicated functions

**Additional Information**

Related to #698

### DIFF
--- a/src/partition/ProvidedPartition.cpp
+++ b/src/partition/ProvidedPartition.cpp
@@ -105,7 +105,7 @@ void ProvidedPartition::communicate()
 
       if (not utils::MasterSlave::isSlave()) {
         PRECICE_CHECK(globalMesh.vertices().size() > 0,
-                      "The provided mesh \"" << globalMesh.getName() << "\" was not defined. Please set the mesh using setMeshXXX() prior to calling initialize().");
+                      "The provided mesh \"" << globalMesh.getName() << "\" is empty. Please set the mesh using setMeshXXX() prior to calling initialize().");
         com::CommunicateMesh(m2n->getMasterCommunication()).sendMesh(globalMesh, 0);
       }
     }
--- a/src/partition/ProvidedPartition.cpp
+++ b/src/partition/ProvidedPartition.cpp
@@ -1,4 +1,3 @@
-#include "partition/ProvidedPartition.hpp"
 #include <algorithm>
 #include <map>
 #include <memory>
@@ -16,6 +15,7 @@
 #include "mesh/Mesh.hpp"
 #include "mesh/Vertex.hpp"
 #include "partition/Partition.hpp"
+#include "partition/ProvidedPartition.hpp"
 #include "utils/Event.hpp"
 #include "utils/MasterSlave.hpp"
 #include "utils/assertion.hpp"
@@ -104,7 +104,8 @@ void ProvidedPartition::communicate()
       Event e("partition.sendGlobalMesh." + _mesh->getName(), precice::syncMode);
 
       if (not utils::MasterSlave::isSlave()) {
-        PRECICE_CHECK(globalMesh.vertices().size() > 0, "The provided mesh " << globalMesh.getName() << " is invalid (possibly empty).");
+        PRECICE_CHECK(globalMesh.vertices().size() > 0,
+                      "The provided mesh \"" << globalMesh.getName() << "\" was not defined. Please set the mesh using setMeshXXX() prior to calling initialize().");
         com::CommunicateMesh(m2n->getMasterCommunication()).sendMesh(globalMesh, 0);
       }
     }
--- a/src/partition/ReceivedPartition.cpp
+++ b/src/partition/ReceivedPartition.cpp
@@ -243,6 +243,18 @@ void ReceivedPartition::compute()
   }
 }
 
+namespace {
+std::string errorMeshFilteredOut(const std::string &meshName)
+{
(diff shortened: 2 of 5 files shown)


src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -859,10 +867,11 @@ void SolverInterfaceImpl::setMeshQuadWithEdges(
...
   if (context.meshRequirement == mapping::Mapping::MeshRequirement::FULL) {
     mesh::PtrMesh &mesh = context.mesh;
-    PRECICE_CHECK(mesh->isValidVertexID(firstVertexID), "Given VertexID is invalid!");
-    PRECICE_CHECK(mesh->isValidVertexID(secondVertexID), "Given VertexID is invalid!");
-    PRECICE_CHECK(mesh->isValidVertexID(thirdVertexID), "Given VertexID is invalid!");
-    PRECICE_CHECK(mesh->isValidVertexID(fourthVertexID), "Given VertexID is invalid!");
+    using impl::errorInvalidVertexID;
+    PRECICE_CHECK(mesh->isValidVertexID(firstVertexID), errorInvalidVertexID(firstVertexID));
+    PRECICE_CHECK(mesh->isValidVertexID(secondVertexID), errorInvalidVertexID(secondVertexID));
+    PRECICE_CHECK(mesh->isValidVertexID(thirdVertexID), errorInvalidVertexID(thridVertexID));
```
uekerman, line 873:
```suggestion
    PRECICE_CHECK(mesh->isValidVertexID(thirdVertexID), errorInvalidVertexID(thirdVertexID));
```

src/partition/ProvidedPartition.cpp
```diff
@@ -104,7 +104,8 @@ void ProvidedPartition::communicate()
       Event e("partition.sendGlobalMesh." + _mesh->getName(), precice::syncMode);
 
       if (not utils::MasterSlave::isSlave()) {
-        PRECICE_CHECK(globalMesh.vertices().size() > 0, "The provided mesh " << globalMesh.getName() << " is invalid (possibly empty).");
+        PRECICE_CHECK(globalMesh.vertices().size() > 0,
+                      "The provided mesh \"" << globalMesh.getName() << "\" was not defined. Please set the mesh using setMeshXXX() prior to calling initialize().");
```
uekerman, line 108:
```suggestion
                      "The provided mesh \"" << globalMesh.getName() << "\" is empty. Please set the mesh using setMeshXXX() prior to calling initialize().");
```
"defined" is a bit ambiguous. We could say that one defines a mesh already in the config.

src/partition/ReceivedPartition.cpp
```diff
@@ -426,11 +420,10 @@ void ReceivedPartition::compareBoundingBoxes()
 
     // send connectionMap to other master
     m2n().getMasterCommunication()->send(connectedRanksList, 0);
-    if (not connectionMap.empty()) {
-      com::CommunicateBoundingBox(m2n().getMasterCommunication()).sendConnectionMap(connectionMap, 0);
-    } else {
-      PRECICE_ERROR("This participant seems to have no mesh partitions for mesh " << _mesh->getName() << " at the coupling interface.");
-    }
+    PRECICE_CHECK(not connectionMap.empty(), "The mesh \"" << _mesh->getName() << "\" of this participant has no partitions at the coupling interface. "
```
uekerman, line 423:
```suggestion
    PRECICE_CHECK(not connectionMap.empty(), "The mesh \"" << _mesh->getName() << "\" of this participant seems to have no partitions at the coupling interface. "
```
It depends who defines where the coupling interface is. This participant or the other one? I prefer the weaker statement here.

src/partition/ReceivedPartition.cpp
```diff
@@ -426,11 +420,10 @@ void ReceivedPartition::compareBoundingBoxes()
...
+                                                                                  "Check that both mapped meshes are describing the same geometry. "
+                                                                                  "If you deal with very coarse meshes, consider increasing the safety-factor in the <use-mesh /> tag.");
```
uekerman, line 425:
```suggestion
                                                                                  "If you deal with very different mesh resolutions, consider increasing the safety-factor in the <use-mesh /> tag.");
```
Just being coarse is no problem I guess?

src/partition/ProvidedPartition.cpp
fsimonis: I am fine with the change.

Normally declaration means "there is a mesh" and definition means "this is how the mesh looks".
But my sight is probably tainted by the programmer inside me.

src/partition/ReceivedPartition.cpp
fsimonis: This check is in the master block of the `ReceivedPartition`, so the side should be clear here.

src/partition/ReceivedPartition.cpp
fsimonis: Your version is what I was looking for :wink:

src/partition/ReceivedPartition.cpp
uekerman: What I meant is: who defines where the coupling interface is? This participant or the other one? The error here means in the end that both sides probably defined different coupling interfaces (concerning the geometry). Maybe we should just reformulate the first sentence completely. 
"Re-partitioning the received mesh XXX resulted in no overlaps with provided meshes."
The suggestions afterwards are good.

PR #818, opened 2020-07-17
Add strict convergence option

### LINKED ISSUE #761: Option to force program termination on divergence/lack of convergence
I want to run a parameter study of different couplings and settings for one of my problems. It is very likely that not all settings work out, i.e. that the coupling will not converge in a time step. If the coupling has not converged, it is very likely that the solution is wrong and the simulation will crash in a few time steps. Moreover, I might not trust the result anymore. Therefore, it would be nice if the simulation would be stopped as soon as the coupling does not converge.

It would be nice, if one could tell preCICE to be strict on the convergence/divergence behavior. This could be a new input parameter in the input file that does not break the current interface. One might consider an option after how many diverged time steps/time windows one wants the simulation to be aborted. 

The parameter could be a new property of the `xxx-convergence-measure` tags. This would give the user the ability to be strict on certain convergence measures they deem important. One possible name would be `strict`. If one wants to allow the coupling to diverge/not converge in at most `n` time steps/windows one could also call it `strict_after_n_time_windows=n`. The default values could be `0` to indicate that preCICE will not abort ever (current behavior).

I would suggest to only implement one of the two options. I came up with the names spontaneously so there is possibly room for improvement. :)

uekerman:
closes #761 
Please see details there. I implemented the "strict overrules suffices" variant.

### DIFF
--- a/src/cplscheme/config/CouplingSchemeConfiguration.cpp
+++ b/src/cplscheme/config/CouplingSchemeConfiguration.cpp
@@ -70,6 +70,7 @@ CouplingSchemeConfiguration::CouplingSchemeConfiguration(
       ATTR_FROM("from"),
       ATTR_TO("to"),
       ATTR_SUFFICES("suffices"),
+      ATTR_STRICT("strict"),
       ATTR_CONTROL("control"),
       VALUE_SERIAL_EXPLICIT("serial-explicit"),
       VALUE_PARALLEL_EXPLICIT("parallel-explicit"),
@@ -175,8 +176,7 @@ void CouplingSchemeConfiguration::xmlTagCallback(
     PRECICE_ASSERT(_config.type == VALUE_MULTI);
     bool        control         = tag.getBooleanAttributeValue(ATTR_CONTROL);
     std::string participantName = tag.getStringAttributeValue(ATTR_NAME);
-    PRECICE_CHECK(std::find(_config.participants.begin(), _config.participants.end(), participantName) == _config.participants.end()
-                      && participantName.compare(_config.controller) != 0,
+    PRECICE_CHECK(std::find(_config.participants.begin(), _config.participants.end(), participantName) == _config.participants.end() && participantName.compare(_config.controller) != 0,
                   "Participant \""
                       << participantName
                       << "\" is provided multiple times to multi coupling scheme. Please make sure that you do not provide the participant multiple times via the <participant name=\""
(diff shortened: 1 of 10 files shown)


src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -625,11 +633,12 @@ void CouplingSchemeConfiguration::addAbsoluteConvergenceMeasure(
                     << "limit=\"" << limit << "\" "
                     << "data=\"" << dataName << "\" "
                     << "mesh=\"" << meshName << "\" "
-                    << "suffices=\"" << suffices << "\"/> subtag in your <coupling-scheme ... /> in the precice-config.xml.");
+                    << "/> subtag in your <coupling-scheme ... /> in the preCICE configuration file.");
   impl::PtrConvergenceMeasure measure(new impl::AbsoluteConvergenceMeasure(limit));
   ConvergenceMeasureDefintion convMeasureDef;
   convMeasureDef.data        = getData(dataName, meshName);
   convMeasureDef.suffices    = suffices;
+  convMeasureDef.strict      = strict;
```
fsimonis, line 641: Does this require special treatment?

Doesn't this mean that all other convergence measures are superfluous.

src/precice/tests/SerialTests.cpp
```diff
@@ -1661,6 +1661,80 @@ BOOST_AUTO_TEST_CASE(testSummationActionTwoSources)
...
+    }
+
+    cplInterface.advance(1.0);
+    ++numberOfAdvanceCalls;
+    ++numberOfIterations;
+
+    if (cplInterface.isActionRequired(actionReadIterationCheckpoint())) {
+      cplInterface.markActionFulfilled(actionReadIterationCheckpoint());
+    } else { //converged
+      BOOST_TEST(numberOfIterations == expectedIterations[timestep]);
```
fsimonis, line 1707: This is undefined behaviour and a SEGFAULT if there are more timesteps then provided.

Using `.at()` throws an exception:
```suggestion
      BOOST_TEST(numberOfIterations == expectedIterations.at(timestep));
```

src/cplscheme/config/CouplingSchemeConfiguration.cpp
uekerman: You mean when you set both strict and suffices to true? Yes, exactly. But this could be helpful for users if they not want to comment out all other convergence measures. And, you could of course mark two measures as strict. Then, "strict" overrules "suffices". Of course, all of that are edge cases, but they should be covered in a meaningful way.

fsimonis: Ok, so there is a use case.

uekerman:
I pushed the changelog directly to develop. https://github.com/precice/precice/commit/5d70ae2443365085da5b958b153d4fb48715c0ff

PR #819, opened 2020-07-17
Weaken residualsum preconditioner check and add additional one

uekerman:
One test introduced in #787 was indeed to strong.
I also turned the assertion in the preconditioner into a check with a rather general error message.

> Your simulation got probably unstable, e.g. produces NAN values.

Not sure if we can write anything more specific.

### DIFF
--- a/src/acceleration/impl/ResidualSumPreconditioner.cpp
+++ b/src/acceleration/impl/ResidualSumPreconditioner.cpp
@@ -43,15 +43,16 @@ void ResidualSumPreconditioner::_update_(bool                   timestepComplete
       sum += norms[k];
       offset += _subVectorSizes[k];
       norms[k] = std::sqrt(norms[k]);
-      PRECICE_CHECK(not math::equals(norms[k], 0.0), "A residual sub-vector in the residual-sum preconditioner became numerically zero. "
-                                                     "Thus, the preconditioner is no longer stable. Please try the value preconditioner instead.");
     }
     sum = std::sqrt(sum);
     PRECICE_ASSERT(sum > 0);
+    PRECICE_CHECK(not math::equals(sum, 0.0), "All residual sub-vectors in the residual-sum preconditioner are numerically zero."
+                                              "Your simulation got probably unstable, e.g. produces NAN values.");
 
     for (size_t k = 0; k < _subVectorSizes.size(); k++) {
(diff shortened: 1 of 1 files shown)


src/acceleration/impl/ResidualSumPreconditioner.cpp
```diff
@@ -43,15 +43,16 @@ void ResidualSumPreconditioner::_update_(bool                   timestepComplete
       sum += norms[k];
       offset += _subVectorSizes[k];
       norms[k] = std::sqrt(norms[k]);
-      PRECICE_CHECK(not math::equals(norms[k], 0.0), "A residual sub-vector in the residual-sum preconditioner became numerically zero. "
-                                                     "Thus, the preconditioner is no longer stable. Please try the value preconditioner instead.");
     }
     sum = std::sqrt(sum);
     PRECICE_ASSERT(sum > 0);
+    PRECICE_CHECK(not math::equals(sum, 0.0), "All residual sub-vectors in the residual-sum preconditioner are numerically zero."
+                                              "Your simulation got probably unstable, e.g. produces NAN values.");
```
MakisH, line 50: Not sure if correct, but feels better:
```suggestion
                                              "Your simulation probably got unstable, e.g. produces NAN values.");
```

src/acceleration/impl/ResidualSumPreconditioner.cpp
```diff
@@ -43,15 +43,16 @@ void ResidualSumPreconditioner::_update_(bool                   timestepComplete
       sum += norms[k];
       offset += _subVectorSizes[k];
       norms[k] = std::sqrt(norms[k]);
-      PRECICE_CHECK(not math::equals(norms[k], 0.0), "A residual sub-vector in the residual-sum preconditioner became numerically zero. "
-                                                     "Thus, the preconditioner is no longer stable. Please try the value preconditioner instead.");
     }
     sum = std::sqrt(sum);
     PRECICE_ASSERT(sum > 0);
+    PRECICE_CHECK(not math::equals(sum, 0.0), "All residual sub-vectors in the residual-sum preconditioner are numerically zero."
```
MakisH, line 49:
```suggestion
    PRECICE_CHECK(not math::equals(sum, 0.0), "All residual sub-vectors in the residual-sum preconditioner are numerically zero. "
```

PR #826, opened 2020-07-19
Add data name to message

uekerman:
@BenjaminRueth Could you please give it a try?

### DIFF
--- a/src/cplscheme/config/CouplingSchemeConfiguration.cpp
+++ b/src/cplscheme/config/CouplingSchemeConfiguration.cpp
@@ -1068,7 +1068,7 @@ void CouplingSchemeConfiguration::checkIfDataIsExchanged(
                               << "acceleration. Please check the <exchange ... /> and "
                               << "<...-convergence-measure ... /> tags in the "
                               << "<coupling-scheme:... /> of your precice-config.xml.");
-} // namespace cplscheme
+}
 
 } // namespace cplscheme
 } // namespace precice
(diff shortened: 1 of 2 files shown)


src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -1052,8 +1051,19 @@ void CouplingSchemeConfiguration::checkIfDataIsExchanged(
...
+
+  std::string dataName = "";
+  for (mesh::PtrMesh mesh : _meshConfig->meshes()) {
+    for (mesh::PtrData data : mesh->data()) {
+      if (data->getID() == dataID) {
+        dataName = data->getName();
+        break;
+      }
+    }
+  }
```
fsimonis, line 1063: Can you refactor this into a little helper function?
Maybe `meah::PtrData CouplingSchemeConfiguration::findDataByID(int ID);`
Return `nullptr` if there is no match. Then this becomes:

```
auto dataptr = findDataByName(dataName);
if (dataptr) {
  std::string dataName = dataptr->getName();
}
```

uekerman: But so far we only need this here. Isn't that a bit over-engineered?

fsimonis: It's all about mental overhead at the call side. _(the inner loop should probably be a `findDataByID`in `Mesh`)_

BenjaminRueth: Done. See https://github.com/precice/precice/pull/826/commits/5413f41808606c87dc4e192ab30e63240e82284c.

BenjaminRueth:
Can be merged as soon as Travis is done.

PR #830, opened 2020-07-20
Rename LU decomposition to QR in configuration

uekerman:
It is actually a QR decomposition, not an LU.

### DIFF
--- a/src/mapping/config/MappingConfiguration.cpp
+++ b/src/mapping/config/MappingConfiguration.cpp
@@ -52,7 +52,7 @@ MappingConfiguration::MappingConfiguration(
                                .setDocumentation("Sets kind of preallocation for PETSc RBF implementation")
                                .setOptions({"estimate", "compute", "off", "save", "tree"});
   auto attrUseLU = makeXMLAttribute(ATTR_USE_QR, false)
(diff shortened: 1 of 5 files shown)


src/mapping/config/MappingConfiguration.cpp
```diff
@@ -51,8 +51,8 @@ MappingConfiguration::MappingConfiguration(
   auto attrPreallocation = makeXMLAttribute("preallocation", "tree")
                                .setDocumentation("Sets kind of preallocation for PETSc RBF implementation")
                                .setOptions({"estimate", "compute", "off", "save", "tree"});
-  auto attrUseLU = makeXMLAttribute(ATTR_USE_LU, false)
-                       .setDocumentation("If set to true, LU decomposition is used to solve the RBF system (only supported in serial)");
+  auto attrUseLU = makeXMLAttribute(ATTR_USE_QR, false)
+                       .setDocumentation("If set to true, QR decomposition is used to solve the RBF system (only supported in serial)");
```
uekerman, line 55:
```suggestion
                       .setDocumentation("If set to true, QR decomposition is used to solve the RBF system");
```

PR #832, opened 2020-07-20
Update information of PETSc in CMake

oguzziya:
This is a small update of description of PETSc in CMake since it is now optional for parallel RBFs

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -97,8 +97,8 @@ add_feature_info(MPICommunication PRECICE_MPICommunication
 add_feature_info(PETScMapping PRECICE_PETScMapping
   "Enables the PETSc-powered radial basic function mappings.
 
-   This enables the PETSc variant of radial basis function mappings which is highly recommended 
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -97,7 +97,8 @@ add_feature_info(MPICommunication PRECICE_MPICommunication
 add_feature_info(PETScMapping PRECICE_PETScMapping
   "Enables the PETSc-powered radial basic function mappings.
 
-   The radial basis function mappings require MPI and PETSc to work in parallel.
+   This enables the PETSc variant of radial basis function mappings which is highly recommended 
+   for large cases running in parallel.
```
MakisH, line 101: Maybe this could be more specific? Something in the lines of:

```suggestion
   This enables the PETSc-based variant of radial basis function mappings which can run in parallel,
   also across different compute nodes. This is highly recommended for large cases running in parallel.
```

Otherwise, it is not really clear "why?".

Issue #837, opened 2020-07-22
precice 2.1.0 and mingw64


3rav:
Hi, 

For 2.1.0 I have this error:

```
[ 70%] Linking CXX executable binprecice.exe
[ 70%] Building CXX object CMakeFiles/testprecice.dir/src/cplscheme/tests/AbsoluteConvergenceMeasureTest.cpp.obj
In file included from C:/msys64/usr/local/precice-2.1.0/src/acceleration/test/QRFactorizationTest.cpp:8:
C:/msys64/usr/local/precice-2.1.0/src/testing/TestContext.hpp:85:20: error: 'precice::testing::Participant precice::testing::operator""_on(const char*, long unsigned int)' has invalid argument list
   85 | inline Participant operator""_on(const char *name, long unsigned int)
      |                    ^~~~~~~~
C:/msys64/usr/local/precice-2.1.0/src/testing/TestContext.hpp:90:52: error: unable to find string literal operator 'operator""_on' with 'const char [1]', 'long long unsigned int' arguments
   90 | static_assert(std::is_same<Participant &, decltype(""_on(1_rank))>::value, "");
      |                                                    ^~~~~
C:/msys64/usr/local/precice-2.1.0/src/testing/TestContext.hpp:90:52: error: unable to find string literal operator 'operator""_on' with 'const char [1]', 'long long unsigned int' arguments
C:/msys64/usr/local/precice-2.1.0/src/testing/TestContext.hpp:90:66: error: template argument 2 is invalid
   90 | static_assert(std::is_same<Participant &, decltype(""_on(1_rank))>::value, "");
```

MakisH:
Good catch, 3rav! We are not testing on Windows yet, so such issues are very helpful insight.

This old discussion on the [GCC Bugzilla](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=50976) seems to be very interesting: the problem seems to be the . What version of mingw64 are you using?

3rav:
Hi,

GNU bash, version 4.4.23(2)-release (x86_64-pc-msys)

mingw-w64-x86_64-gcc 10.1.0-3
mingw-w64-x86_64-boost 1.73.0-4
mingw-w64-x86_64-cmake 3.17.3-1
mingw-w64-x86_64-eigen3 3.3.7-1
mingw-w64-x86_64-libxml2 2.9.10-4
mingw-w64-x86_64-yaml-cpp 0.6.3-3

fsimonis:
[...]

PR #839, opened 2020-07-22
Fix operator"" parameter type

fsimonis:
Fix for #837


fsimonis:
I guess it was the usual case of being "overwhelmed by the reference".
Looks quite clear to me now.

PR #843, opened 2020-07-27
Add GitHub Actions CI

fsimonis:

Adds a workflow which builds and tests preCICE on various platforms.

**Additional Information**
This partially implements #713

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -1,0 +1,135 @@
+name: Build and Test
+on:
+  pull_request:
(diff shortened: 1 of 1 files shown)


uekerman:
Is it expected that this does not run on this very PR? 
If yes I would say let's just merge it. 
Not a show stopper, just to understand better: What of #713 is not yet included here?

fsimonis:
@uekerman the Action runs on code changes only, so it won't run in this PR.

This PR currently does not include CentOS7 and Spack.

PR #849, opened 2020-08-18
Allow multiple mappings to and from the same mesh

### LINKED ISSUE #840: Allow for multiple mappings from or to the same mesh

This originates from https://precice.discourse.group/t/two-couplings-to-one-interface-reloaded/260 .
Consider the following situation

```xml
    <mesh name="MeshB">
      <use-data name="Pressure" />
      <use-data name="DisplacementTop" />
      <use-data name="DisplacementBottom" />
      <use-data name="DisplacementSum" />
    </mesh>

    <mesh name="MeshATop">
      <use-data name="Pressure" />
      <use-data name="DisplacementTop" />
    </mesh>
    
    <mesh name="MeshABottom">
      <use-data name="Pressure" />
      <use-data name="DisplacementBottom" />
    </mesh>

    <participant name="A">
      <use-mesh name="MeshATop" provide="yes"/>
      <use-mesh name="MeshABottom" provide="yes"/>
      <use-mesh name="MeshB" from="B"/>
            
      <action:summation timing="write-mapping-post" mesh="MeshB">
        <source-data name="DisplacementTop"/>
        <source-data name="DisplacementBottom"/>
        <target-data name="DisplacementSum"/>
      </action:summation>
      
      <write-data name="DisplacementTop" mesh="MeshATop"/>
      <write-data name="DisplacementBottom" mesh="MeshABottom"/>
      <read-data name="Pressure" mesh="MeshATop"/>
      <read-data name="Pressure" mesh="MeshABottom"/>

      <mapping:nearest-neighbor direction="read" from="MeshB" to="MeshATop" constraint="consistent" />
      <mapping:nearest-neighbor direction="read" from="MeshB" to="MeshABottom" constraint="consistent" />
      <mapping:nearest-neighbor direction="write" from="MeshATop" to="MeshB" constraint="consistent" />
      <mapping:nearest-neighbor direction="write" from="MeshABottom" to="MeshB" constraint="consistent" />
    </participant>
```

Such a configuration is currently not allowed. We do not allow for multiple mappings from or to the same mesh:

https://github.com/precice/precice/blob/42abc38c9d82a719db0f6982548b6d979931ccf6/src/mapping/config/MappingConfiguration.cpp#L391-L403

In general, for duplicated "to meshes", there is a good reason for this: it's not well defined. Should data values be overwritten or summed up or any other operation? More discussion on this topic: https://precice.discourse.group/t/two-couplings-to-one-interface-mixed-dimensional-coupling/68
However, for the case above it is indeed well-defined as the two write data fields `DisplacementTop` and `DisplacementBottom` are both **only** associated to one "from mesh", `MeshATop` and `MeshABottom`, respectively. 
For duplicated "from meshes" everything is well defined.
At the same time, we currently have many technical restrictions which do not allow for multiple mappings. One important example is the re-partitioning in `ReceivedPartition`:

https://github.com/precice/precice/blob/42abc38c9d82a719db0f6982548b6d979931ccf6/src/partition/Partition.hpp#L66-L68

Extend all mapping data structures such that we can handle multiple mappings from and to the same mesh.
At the same time, we still need to restrict the ill-defined duplicated "to meshes".

The forum posts discuss some workarounds.

uekerman:
Closes #840 
Ready for review.

* [x] Add integration tests that separately use multiple mappings from and to the same mesh, respectively
* [x] Allow multiple mappings from same mesh
* [x] Allow multiple mappings to same mesh if there are no overlaps
* [x] Extend unit tests of `ReceivedPartition`

### DIFF
--- a/src/partition/Partition.hpp
+++ b/src/partition/Partition.hpp
@@ -45,14 +45,14 @@ public:
   /// The partition is computed, i.e. the mesh re-partitioned if required and all data structures are set up.
   virtual void compute() = 0;
 
-  void setFromMapping(mapping::PtrMapping fromMapping)
+  void addFromMapping(mapping::PtrMapping fromMapping)
   {
-    _fromMapping = fromMapping;
+    _fromMappings.push_back(fromMapping);
   }
 
-  void setToMapping(mapping::PtrMapping toMapping)
+  void addToMapping(mapping::PtrMapping toMapping)
   {
-    _toMapping = toMapping;
+    _toMappings.push_back(toMapping);
   }
 
   void addM2N(m2n::PtrM2N m2n)
@@ -63,9 +63,9 @@ public:
 protected:
   mesh::PtrMesh _mesh;
 
-  mapping::PtrMapping _fromMapping;
+  std::vector<mapping::PtrMapping> _fromMappings;
 
-  mapping::PtrMapping _toMapping;
+  std::vector<mapping::PtrMapping> _toMappings;
 
   /// m2n connection to each connected participant
   std::vector<m2n::PtrM2N> _m2ns;
--- a/src/partition/Partition.hpp
+++ b/src/partition/Partition.hpp
@@ -47,7 +47,7 @@ public:
 
   void addFromMapping(mapping::PtrMapping fromMapping)
   {
-    _fromMappings.push_back(fromMapping);
+    _fromMappings.push_back(std::move(fromMapping));
   }
 
   void addToMapping(mapping::PtrMapping toMapping)
--- a/src/partition/Partition.hpp
+++ b/src/partition/Partition.hpp
@@ -52,7 +52,7 @@ public:
 
   void addToMapping(mapping::PtrMapping toMapping)
   {
-    _toMappings.push_back(toMapping);
+    _toMappings.push_back(std::move(toMapping));
   }
 
   void addM2N(m2n::PtrM2N m2n)
--- a/src/partition/ReceivedPartition.cpp
+++ b/src/partition/ReceivedPartition.cpp
@@ -104,7 +104,7 @@ void ReceivedPartition::compute()
 
   // check to prevent false configuration
   if (not utils::MasterSlave::isSlave()) {
-    PRECICE_CHECK(_fromMapping || _toMapping,
+    PRECICE_CHECK(not(_fromMappings.empty() && _toMappings.empty()),
                   "The received mesh " << _mesh->getName()
                                        << " needs a mapping, either from it, to it, or both. Maybe you don't want to receive this mesh at all?")
   }
@@ -119,10 +119,12 @@ void ReceivedPartition::compute()
   PRECICE_DEBUG("Tag vertices for filtering: 1st round.");
   _mesh->computeState(); // normals need to be ready for NP mapping
   // go to both meshes, vertex is tagged if already one mesh tags him
-  if (_fromMapping)
-    _fromMapping->tagMeshFirstRound();
-  if (_toMapping)
-    _toMapping->tagMeshFirstRound();
+  for (mapping::PtrMapping fromMapping : _fromMappings) {
+    fromMapping->tagMeshFirstRound();
+  }
+  for (mapping::PtrMapping toMapping : _toMappings) {
+    toMapping->tagMeshFirstRound();
+  }
 
   // (3) Define which vertices are owned by this rank
   PRECICE_DEBUG("Create owner information.");
@@ -130,10 +132,12 @@ void ReceivedPartition::compute()
 
   // (4) Tag vertices 2nd round (what should be filtered out)
   PRECICE_DEBUG("Tag vertices for filtering: 2nd round.");
-  if (_fromMapping)
-    _fromMapping->tagMeshSecondRound();
-  if (_toMapping)
-    _toMapping->tagMeshSecondRound();
+  for (mapping::PtrMapping fromMapping : _fromMappings) {
+    fromMapping->tagMeshSecondRound();
+  }
+  for (mapping::PtrMapping toMapping : _toMappings) {
+    toMapping->tagMeshSecondRound();
+  }
 
   // (5) Filter mesh according to tag
   PRECICE_INFO("Filter mesh " << _mesh->getName() << " by mappings");
@@ -281,7 +285,7 @@ void ReceivedPartition::filterByBoundingBox()
       PRECICE_DEBUG("Receive filtered mesh");
       com::CommunicateMesh(utils::MasterSlave::_communication).receiveMesh(*_mesh, 0);
 
-      if (areProvidedMeshesEmpty()) {
+      if (isAnyProvidedMeshNonEmpty()) {
         PRECICE_CHECK(not _mesh->vertices().empty(), errorMeshFilteredOut(_mesh->getName()));
       }
 
@@ -310,7 +314,7 @@ void ReceivedPartition::filterByBoundingBox()
       _mesh->clear();
       _mesh->addMesh(filteredMesh);
 
-      if (areProvidedMeshesEmpty()) {
+      if (isAnyProvidedMeshNonEmpty()) {
         PRECICE_CHECK(not _mesh->vertices().empty(), errorMeshFilteredOut(_mesh->getName()));
       }
     }
@@ -334,7 +338,7 @@ void ReceivedPartition::filterByBoundingBox()
       mesh::Mesh filteredMesh("FilteredMesh", _dimensions, _mesh->isFlipNormals(), mesh::Mesh::MESH_ID_UNDEFINED);
       mesh::filterMesh(filteredMesh, *_mesh, [&](const mesh::Vertex &v) { return _bb.contains(v); });
 
-      if (areProvidedMeshesEmpty()) {
+      if (isAnyProvidedMeshNonEmpty()) {
         PRECICE_CHECK(not _mesh->vertices().empty(), errorMeshFilteredOut(_mesh->getName()));
       }
 
@@ -450,15 +454,15 @@ void ReceivedPartition::prepareBoundingBox()
 
   PRECICE_DEBUG("Merge bounding boxes and increase by safety factor");
 
-  // Create BB around both "other" meshes
-  if (_fromMapping) {
-    auto other_bb = _fromMapping->getOutputMesh()->getBoundingBox();
+  // Create BB around all "other" meshes
+  for (mapping::PtrMapping fromMapping : _fromMappings) {
+    auto other_bb = fromMapping->getOutputMesh()->getBoundingBox();
     _bb.expandBy(other_bb);
     _bb.scaleBy(_safetyFactor);
     _boundingBoxPrepared = true;
   }
-  if (_toMapping) {
-    auto other_bb = _toMapping->getInputMesh()->getBoundingBox();
+  for (mapping::PtrMapping toMapping : _toMappings) {
+    auto other_bb = toMapping->getInputMesh()->getBoundingBox();
     _bb.expandBy(other_bb);
     _bb.scaleBy(_safetyFactor);
     _boundingBoxPrepared = true;
@@ -610,10 +614,19 @@ void ReceivedPartition::createOwnerInformation()
   }
 }
 
-bool ReceivedPartition::areProvidedMeshesEmpty() const
+bool ReceivedPartition::isAnyProvidedMeshNonEmpty() const
 {
-  return (_fromMapping && not _fromMapping->getOutputMesh()->vertices().empty()) ||
-         (_toMapping && not _toMapping->getInputMesh()->vertices().empty());
+  for (mapping::PtrMapping fromMapping : _fromMappings) {
+    if (not fromMapping->getOutputMesh()->vertices().empty()) {
+      return true;
+    }
+  }
+  for (mapping::PtrMapping toMapping : _toMappings) {
+    if (not toMapping->getInputMesh()->vertices().empty()) {
+      return true;
+    }
+  }
+  return false;
 }
 
 void ReceivedPartition::setOwnerInformation(const std::vector<int> &ownerVec)
--- a/src/partition/ReceivedPartition.cpp
+++ b/src/partition/ReceivedPartition.cpp
@@ -461,7 +461,7 @@ void ReceivedPartition::prepareBoundingBox()
     _bb.scaleBy(_safetyFactor);
     _boundingBoxPrepared = true;
   }
-  for (mapping::PtrMapping toMapping : _toMappings) {
+  for (mapping::PtrMapping& toMapping : _toMappings) {
     auto other_bb = toMapping->getInputMesh()->getBoundingBox();
     _bb.expandBy(other_bb);
     _bb.scaleBy(_safetyFactor);
--- a/src/partition/ReceivedPartition.cpp
+++ b/src/partition/ReceivedPartition.cpp
@@ -616,7 +616,7 @@ void ReceivedPartition::createOwnerInformation()
 
 bool ReceivedPartition::isAnyProvidedMeshNonEmpty() const
 {
-  for (mapping::PtrMapping fromMapping : _fromMappings) {
+  for (const auto& fromMapping : _fromMappings) {
     if (not fromMapping->getOutputMesh()->vertices().empty()) {
       return true;
     }
--- a/src/partition/ReceivedPartition.cpp
+++ b/src/partition/ReceivedPartition.cpp
@@ -621,7 +621,7 @@ bool ReceivedPartition::isAnyProvidedMeshNonEmpty() const
       return true;
     }
   }
-  for (mapping::PtrMapping toMapping : _toMappings) {
+  for (const auto& toMapping : _toMappings) {
     if (not toMapping->getInputMesh()->vertices().empty()) {
       return true;
     }
--- a/src/partition/ReceivedPartition.cpp
+++ b/src/partition/ReceivedPartition.cpp
@@ -104,7 +104,7 @@ void ReceivedPartition::compute()
 
   // check to prevent false configuration
   if (not utils::MasterSlave::isSlave()) {
-    PRECICE_CHECK(not(_fromMappings.empty() && _toMappings.empty()),
+    PRECICE_CHECK(hasAnyMapping(),
                   "The received mesh " << _mesh->getName()
                                        << " needs a mapping, either from it, to it, or both. Maybe you don't want to receive this mesh at all?")
   }
@@ -119,12 +119,7 @@ void ReceivedPartition::compute()
   PRECICE_DEBUG("Tag vertices for filtering: 1st round.");
   _mesh->computeState(); // normals need to be ready for NP mapping
   // go to both meshes, vertex is tagged if already one mesh tags him
-  for (mapping::PtrMapping fromMapping : _fromMappings) {
-    fromMapping->tagMeshFirstRound();
-  }
-  for (mapping::PtrMapping toMapping : _toMappings) {
-    toMapping->tagMeshFirstRound();
-  }
+  tagMeshFirstRound();
 
   // (3) Define which vertices are owned by this rank
   PRECICE_DEBUG("Create owner information.");
@@ -132,12 +127,7 @@ void ReceivedPartition::compute()
 
   // (4) Tag vertices 2nd round (what should be filtered out)
   PRECICE_DEBUG("Tag vertices for filtering: 2nd round.");
-  for (mapping::PtrMapping fromMapping : _fromMappings) {
-    fromMapping->tagMeshSecondRound();
-  }
-  for (mapping::PtrMapping toMapping : _toMappings) {
-    toMapping->tagMeshSecondRound();
-  }
+  tagMeshSecondRound();
 
   // (5) Filter mesh according to tag
   PRECICE_INFO("Filter mesh " << _mesh->getName() << " by mappings");
@@ -455,13 +445,13 @@ void ReceivedPartition::prepareBoundingBox()
   PRECICE_DEBUG("Merge bounding boxes and increase by safety factor");
 
   // Create BB around all "other" meshes
-  for (mapping::PtrMapping& fromMapping : _fromMappings) {
+  for (mapping::PtrMapping &fromMapping : _fromMappings) {
     auto other_bb = fromMapping->getOutputMesh()->getBoundingBox();
     _bb.expandBy(other_bb);
     _bb.scaleBy(_safetyFactor);
     _boundingBoxPrepared = true;
   }
-  for (mapping::PtrMapping& toMapping : _toMappings) {
+  for (mapping::PtrMapping &toMapping : _toMappings) {
     auto other_bb = toMapping->getInputMesh()->getBoundingBox();
     _bb.expandBy(other_bb);
     _bb.scaleBy(_safetyFactor);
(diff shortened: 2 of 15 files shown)


uekerman:
@ajaust Could you please give this a try? Have a look at the two added test configurations to see what now should be possible.

ajaust:
I have been testing it a bit and run into the following error message:
```
[config:MappingConfiguration]:406 in checkDuplicates: ERROR: There cannot be two mappings to mesh "HDFlowMesh" if the meshes from which is mapped contain duplicated data fields. Here, both from meshes contain data "Pressure". The mapping is not well defined. Which data "Pressure" should be mapped to mesh "HDFlowMesh"?
```

Are there, by any chance, still some extra guards in charge that need to be deactivated?

I worked on it step by step (OK... I tried to implement everything at once and got the error message above so I started with smaller steps):

1. Create a case in which `DisplacementTop` and `DisplacementBottom` from two meshes mapped onto one mesh and summed up via the Python action before the data is sent to the other solver based on the included test case. `Pressure` has been removed from coupling. Seems to **work** 
[serial-explicit-displacement-from-multiple-meshes.xml.txt](https://github.com/precice/precice/files/5525091/serial-explicit-displacement-from-multiple-meshes.xml.txt)

2. Create a case in which `Pressure` is mapped from one to two meshes based on the included test case. The displacements have been  removed from coupling. Seems to **work** 
[serial-explicit-pressure-to-multiple-meshes.xml.txt](https://github.com/precice/precice/files/5525090/serial-explicit-pressure-to-multiple-meshes.xml.txt)

3. Starting from the actual case with the pressure and displacements in the coupling I wanted to incorporate the summation of displacements (see 1.) while the pressure mapping is still active. **Fails** with error message above. 
[serial-explicit-displacement-from-multiple-meshes-with-pressure.xml.txt](https://github.com/precice/precice/files/5525088/serial-explicit-displacement-from-multiple-meshes-with-pressure.xml.txt)

In order to keep things simple I set it up as explicit coupling. 

Am I doing something stupid? Step 3 should eventually merge steps 1 and 2, but it fails halfway.

uekerman:
Thanks for the detailed report! Yes, the check it still too strict. Should ignore the pressure as the pressure is read and not write. Will fix this asap.

uekerman:
@ajaust I relaxed the check. Could you please give it another try?

ajaust:
@uekerman I tested it and the first results look promising.  Thank you so much!  I have used the following configuration: [serial-implicit-displacement-from-multiple-meshes-with-pressure.xml.txt](https://github.com/precice/precice/files/5542434/serial-implicit-displacement-from-multiple-meshes-with-pressure.xml.txt)

I see some lack of symmetry, but this is most likely caused by my coarse meshes (which are not perfectly symmetric itself) that I have used for testing.

I was wondering how I could do the parallel simulations now. In this case I would have to use two `read-consistent` mappings. When I think about it, I get a bit confused. Would this be possible to write the configuration in a way such that I sum up the Displacements on `BiotSolver` before it is send to `HDFlowSolver`?

uekerman:
> @uekerman I tested it and the first results look promising. Thank you so much! I have used the following configuration: [serial-implicit-displacement-from-multiple-meshes-with-pressure.xml.txt](https://github.com/precice/precice/files/5542434/serial-implicit-displacement-from-multiple-meshes-with-pressure.xml.txt)

Great :tada: 

> I was wondering how I could do the parallel simulations now. In this case I would have to use two read-consistent mappings. When I think about it, I get a bit confused. Would this be possible to write the configuration in a way such that I sum up the Displacements on BiotSolver before it is send to HDFlowSolver?

ouch :disappointed: 
No, that sounds like a show stopper. 
What we would need for that is sth like a `acceleration-prior` action timing here in line 72:
https://github.com/precice/precice/blob/641e5057cb5d71d39ed43ae1dbc6af597b60bf52/src/cplscheme/ParallelCouplingScheme.cpp#L68-L77
(and similarly for other coupling schemes)

Then your config could look like this: 
[serial-implicit-displacement-from-multiple-meshes-with-pressure.xml.txt](https://github.com/precice/precice/files/5547071/serial-implicit-displacement-from-multiple-meshes-with-pressure.xml.txt)

Such an additional timing would require some refactoring as now the action cannot be applied directly `SolverInterfaceImpl`.

But there is hope :eyes: (?)
From your forum posts it seems like you use matching meshes? If yes then simply make your displacement mappings conservative.

ajaust:
I am relieved that I am not too stupid. I had similar thoughts and the preCICE configuration that I have set up looks similar to your "what it would like" config. :)

I will go for the conservative mapping for now. I have just started a simulation in parallel with the conservative mapping and it looks fine so far. 

You interpreted correctly that we use matching meshes at the moment. For now that will suffice. Having the option to use non-matching meshes could be nice, I think. However, I must also admit that the engineers are not certain whether mixing of different meshes would to much sense to them (I don't know the details).

ajaust:
When running the simulation on parallel I ran into the following error message:

```
ERROR: The received mesh FractureMeshBottom needs a mapping, either from it, to it, or both. Maybe you don't want to receive this mesh at all?
```

I am not sure yet why that happens. I am still investigating if I do something wrong or not.

uekerman:
For which config? That sounds like a fixable problem.

ajaust:
It was for the configuration you have posted. I am a bit puzzled. It works on my local machine. 

I am trying to run my code on another machine. At the moment I am unsure whether I do something wrong or whether my installation on the other machine is broken. I have installed this PR via Spack on the other machine (or at least I thougt so). Besides the error message above I do not get the hash of the last commit as version when starting the code:

```
(0) 16:30:31 [impl::SolverInterfaceImpl]:162 in configure: This is preCICE version 2.1.1                                                    
(0) 16:30:31 [impl::SolverInterfaceImpl]:163 in configure: Revision info: no-info [Git failed/Not a repository]  
```

So at the moment I assume that my installation did something wrong or the hash just disappeared due to Spack. However, manually compiling preCICE turned out to fail (so far) :(

ajaust:
Nevermind. I messed something up when setting up the configuration file on the remote machine.I have a larger number of simulations running now. I will report back when they are finished and I have an idea of what is going on there.

I just quickly checked some logfiles and they looked very promising.

ajaust:
I might have run into some issues when the participants run in parallel. I will try to narrow it down a bit and come back with some example code.

ajaust:
I think the problem I observe for my application is actually my own fault. I am debugging this a bit more.

However, while debugging I wrote the following dummy solvers for testing: [dummy_multi-mesh.tar.gz](https://github.com/precice/precice/files/5826801/dummy_multi-mesh.tar.gz)

They behave a bit unexpected. If I run both solver with 1 thread, it works as expected. As soon as I run the solvers with

```
mpirun -n 2 ./dummysolver_high_dim.py precice-config-mixed-dim.xml 
```
and
```
mpirun -n 1 ./dummysolver_low_dim.py precice-config-mixed-dim.xml
```
the script `dummysolver_low_dim.py` obtains too large data in one of its data points. 

What do the dummy solvers do?
`dummysolver_high_dim` creates two times the same mesh for each process. The coordinates of the mesh start at point (rank, 0) until (rank-dx, 0) and then writes data to two meshes. The meshes resemble the two surfaces of a fracture

`dummysolver_low_dim` does pretty much the same as the other dummy solver, but only has one mesh as it resembles the lower dimensional mesh approximation.

uekerman:
@ajaust I can run and reproduce your results. Looking into it now.

src/partition/Partition.hpp
```diff
@@ -45,14 +45,14 @@ class Partition {
   /// The partition is computed, i.e. the mesh re-partitioned if required and all data structures are set up.
   virtual void compute() = 0;
 
-  void setFromMapping(mapping::PtrMapping fromMapping)
+  void addFromMapping(mapping::PtrMapping fromMapping)
   {
-    _fromMapping = fromMapping;
+    _fromMappings.push_back(fromMapping);
```
fsimonis, line 50: We can move the argument to prevent yet another unnecessary copy of the shared ptr.

```suggestion
    _fromMappings.push_back(std::move(fromMapping));
```

src/partition/Partition.hpp
```diff
@@ -45,14 +45,14 @@ class Partition {
...
   }
 
-  void setToMapping(mapping::PtrMapping toMapping)
+  void addToMapping(mapping::PtrMapping toMapping)
   {
-    _toMapping = toMapping;
+    _toMappings.push_back(toMapping);
```
fsimonis, line 55:
```suggestion
    _toMappings.push_back(std::move(toMapping));
```

src/partition/ReceivedPartition.cpp
```diff
@@ -104,7 +104,7 @@ void ReceivedPartition::compute()
 
   // check to prevent false configuration
   if (not utils::MasterSlave::isSlave()) {
-    PRECICE_CHECK(_fromMapping || _toMapping,
+    PRECICE_CHECK(not(_fromMappings.empty() && _toMappings.empty()),
```
fsimonis, line 107: We could bundle this in `anyMapping()`, `hasNoMapping()` or `hasAnyMapping()`.
This would simplify things a bit here.

src/partition/ReceivedPartition.cpp
```diff
@@ -119,21 +119,25 @@ void ReceivedPartition::compute()
...
-  if (_fromMapping)
-    _fromMapping->tagMeshFirstRound();
-  if (_toMapping)
-    _toMapping->tagMeshFirstRound();
+  for (mapping::PtrMapping fromMapping : _fromMappings) {
+    fromMapping->tagMeshFirstRound();
+  }
+  for (mapping::PtrMapping toMapping : _toMappings) {
+    toMapping->tagMeshFirstRound();
+  }
```
fsimonis, line 127: We could refactor this into a separate function.
This would make this long function a bit more readable.

```suggestion
  tagMeshFirstRounds();
```

src/partition/ReceivedPartition.cpp
```diff
@@ -119,21 +119,25 @@ void ReceivedPartition::compute()
...
-  if (_fromMapping)
-    _fromMapping->tagMeshSecondRound();
-  if (_toMapping)
-    _toMapping->tagMeshSecondRound();
+  for (mapping::PtrMapping fromMapping : _fromMappings) {
+    fromMapping->tagMeshSecondRound();
+  }
+  for (mapping::PtrMapping toMapping : _toMappings) {
+    toMapping->tagMeshSecondRound();
+  }
```
fsimonis, line 140: We could refactor this into a separate function.
This would make this long function a bit more readable.

```suggestion
  tagMeshSecondRounds();
```

src/partition/ReceivedPartition.cpp
```diff
@@ -450,15 +454,15 @@ void ReceivedPartition::prepareBoundingBox()
 
   PRECICE_DEBUG("Merge bounding boxes and increase by safety factor");
 
-  // Create BB around both "other" meshes
-  if (_fromMapping) {
-    auto other_bb = _fromMapping->getOutputMesh()->getBoundingBox();
+  // Create BB around all "other" meshes
+  for (mapping::PtrMapping fromMapping : _fromMappings) {
```
fsimonis, line 458: We should take the shared pointer by reference to prevent copies.
```suggestion
  for (mapping::PtrMapping& fromMapping : _fromMappings) {
```

src/partition/ReceivedPartition.cpp
```diff
@@ -450,15 +454,15 @@ void ReceivedPartition::prepareBoundingBox()
...
+    auto other_bb = fromMapping->getOutputMesh()->getBoundingBox();
     _bb.expandBy(other_bb);
     _bb.scaleBy(_safetyFactor);
     _boundingBoxPrepared = true;
   }
-  if (_toMapping) {
-    auto other_bb = _toMapping->getInputMesh()->getBoundingBox();
+  for (mapping::PtrMapping toMapping : _toMappings) {
```
fsimonis, line 464: Same as above

```suggestion
  for (mapping::PtrMapping& toMapping : _toMappings) {
```

src/partition/ReceivedPartition.cpp
```diff
@@ -610,10 +614,19 @@ void ReceivedPartition::createOwnerInformation()
   }
 }
 
-bool ReceivedPartition::areProvidedMeshesEmpty() const
+bool ReceivedPartition::isAnyProvidedMeshNonEmpty() const
 {
-  return (_fromMapping && not _fromMapping->getOutputMesh()->vertices().empty()) ||
-         (_toMapping && not _toMapping->getInputMesh()->vertices().empty());
+  for (mapping::PtrMapping fromMapping : _fromMappings) {
```
fsimonis, line 619: Same as above, but also don't need write access.
Thus we can use a const reference.
```suggestion
  for (const auto& fromMapping : _fromMappings) {
```

src/partition/ReceivedPartition.cpp
```diff
@@ -610,10 +614,19 @@ void ReceivedPartition::createOwnerInformation()
...
+    if (not fromMapping->getOutputMesh()->vertices().empty()) {
+      return true;
+    }
+  }
+  for (mapping::PtrMapping toMapping : _toMappings) {
```
fsimonis, line 624: Same as above
```suggestion
  for (const auto& toMapping : _toMappings) {
```

src/partition/ReceivedPartition.cpp
```diff
@@ -610,10 +614,19 @@ void ReceivedPartition::createOwnerInformation()
...
+    if (not toMapping->getInputMesh()->vertices().empty()) {
+      return true;
```
fsimonis, line 626: Do we need more tests here?

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -559,5 +561,52 @@ void ParticipantConfiguration::finishParticipantConfiguration(
...
+  PRECICE_TRACE();
+  using ConfMapping = mapping::MappingConfiguration::ConfiguredMapping;
+
+  for (const ConfMapping &configuredMapping : _mappingConfig->mappings()) {
+    bool sameToMesh   = mapping.toMesh->getName() == configuredMapping.toMesh->getName();
+    bool sameFromMesh = mapping.fromMesh->getName() == configuredMapping.fromMesh->getName();
+    if (sameToMesh && sameFromMesh) {
+      // It's really the same mapping, duplicated mappings are already checked for in MappingConfiguration
+      return;
+    }
```
fsimonis, line 577: Shouldn't this be an assertion?

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -559,5 +561,52 @@ void ParticipantConfiguration::finishParticipantConfiguration(
...
+
+    if (sameToMesh) {
+      for (const mesh::PtrData data : mapping.fromMesh->data()) {
```
fsimonis, line 580:
```suggestion
      for (const mesh::PtrData& data : mapping.fromMesh->data()) {
```

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -559,5 +561,52 @@ void ParticipantConfiguration::finishParticipantConfiguration(
...
+        for (const mesh::PtrData configuredData : configuredMapping.fromMesh->data()) {
```
fsimonis, line 581:
```suggestion
        for (const mesh::PtrData& configuredData : configuredMapping.fromMesh->data()) {
```

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -559,5 +561,52 @@ void ParticipantConfiguration::finishParticipantConfiguration(
...
+          bool sameFromData = data->getName() == configuredData->getName();
+
+          if (sameFromData) {
```
fsimonis, line 584: Using a guard here may be more readable:
```suggestion
          bool sameFromData = data->getName() == configuredData->getName();
          if (not sameFromData) {
              continue;
          }
```

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -559,5 +561,52 @@ void ParticipantConfiguration::finishParticipantConfiguration(
...
+            bool sameDirection = false;
+
+            if (mapping.direction == mapping::MappingConfiguration::WRITE) {
+              for (impl::DataContext &dataContext : participant->writeDataContexts()) {
```
fsimonis, line 588: We only need read access here
```suggestion
              for (const impl::DataContext &dataContext : participant->writeDataContexts()) {
```

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -559,5 +561,52 @@ void ParticipantConfiguration::finishParticipantConfiguration(
...
+                sameDirection |= data->getName() == dataContext.getName();
+              }
+            }
+            if (mapping.direction == mapping::MappingConfiguration::READ) {
+              for (impl::DataContext &dataContext : participant->readDataContexts()) {
```
fsimonis, line 593: Same as above
```suggestion
              for (const impl::DataContext &dataContext : participant->readDataContexts()) {
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -976,33 +976,37 @@ void SolverInterfaceImpl::mapWriteDataFrom(
 {
   PRECICE_TRACE(fromMeshID);
   PRECICE_VALIDATE_MESH_ID(fromMeshID);
-  impl::MeshContext &   context        = _accessor->meshContext(fromMeshID);
-  impl::MappingContext &mappingContext = context.fromMappingContext;
-  if (not mappingContext.mapping) {
+  impl::MeshContext &context = _accessor->meshContext(fromMeshID);
+
+  if (context.fromMappingContexts.empty()) {
     PRECICE_ERROR("You attempt to \"mapWriteDataFrom\" mesh "
```
fsimonis, line 982: This should be a  `PRECICE_CHECK`

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -976,33 +976,37 @@ void SolverInterfaceImpl::mapWriteDataFrom(
...
-      PRECICE_ASSERT(mappingContext.mapping == context.mappingContext.mapping);
-      mappingContext.mapping->map(inDataID, outDataID);
+
+  for (impl::MappingContext &mappingContext : context.fromMappingContexts) {
+    if (not mappingContext.mapping->hasComputedMapping()) {
+      PRECICE_DEBUG("Compute mapping from mesh \"" << context.mesh->getName() << "\"");
+      mappingContext.mapping->computeMapping();
     }
+    for (impl::DataContext &context : _accessor->writeDataContexts()) {
+      if (context.mesh->getID() == fromMeshID) {
```
fsimonis, line 998: We could use a guard here to prevent nesting.
```suggestion
      if (context.mesh->getID() != fromMeshID) {
          continue;
      }
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -976,33 +976,37 @@ void SolverInterfaceImpl::mapWriteDataFrom(
...
+        int inDataID             = context.fromData->getID();
+        int outDataID            = context.toData->getID();
```
fsimonis, line 1000: This could be merged with the bottom call to `->map()` to reduce the verbosity.
If readability is more important, moving it directly before the above call make things even clearer.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1011,34 +1015,38 @@ void SolverInterfaceImpl::mapReadDataTo(
 {
   PRECICE_TRACE(toMeshID);
   PRECICE_VALIDATE_MESH_ID(toMeshID);
-  impl::MeshContext &   context        = _accessor->meshContext(toMeshID);
-  impl::MappingContext &mappingContext = context.toMappingContext;
-  if (mappingContext.mapping.use_count() == 0) {
+  impl::MeshContext &context = _accessor->meshContext(toMeshID);
+
+  if (context.toMappingContexts.empty()) {
     PRECICE_ERROR("You attempt to \"mapReadDataTo\" mesh "
```
fsimonis, line 1021: This should be a `PRECICE_CHECK`.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1011,34 +1015,38 @@ void SolverInterfaceImpl::mapReadDataTo(
...
-      mappingContext.mapping->map(inDataID, outDataID);
-      PRECICE_DEBUG("Mapped values = " << utils::previewRange(3, context.toData->values()));
+
+  for (impl::MappingContext &mappingContext : context.toMappingContexts) {
+    if (not mappingContext.mapping->hasComputedMapping()) {
+      PRECICE_DEBUG("Compute mapping from mesh \"" << context.mesh->getName() << "\"");
+      mappingContext.mapping->computeMapping();
+    }
+    for (impl::DataContext &context : _accessor->readDataContexts()) {
+      if (context.mesh->getID() == toMeshID) {
```
fsimonis, line 1037: You could use a guard similar to the comment above.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1011,34 +1015,38 @@ void SolverInterfaceImpl::mapReadDataTo(
...
+        int inDataID             = context.fromData->getID();
+        int outDataID            = context.toData->getID();
```
fsimonis, line 1039: Same applies as above

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1372,8 +1380,12 @@ void SolverInterfaceImpl::configurePartitions(
       m2n::PtrM2N m2n = m2nConfig->getM2N(receiver, provider);
       m2n->createDistributedCommunication(context->mesh);
       context->partition->addM2N(m2n);
-      context->partition->setFromMapping(context->fromMappingContext.mapping);
-      context->partition->setToMapping(context->toMappingContext.mapping);
+      for (MappingContext &mappingContext : context->fromMappingContexts) {
```
fsimonis, line 1383: This should be read access only

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1372,8 +1380,12 @@ void SolverInterfaceImpl::configurePartitions(
...
+        context->partition->addFromMapping(mappingContext.mapping);
+      }
+      for (MappingContext &mappingContext : context->toMappingContexts) {
```
fsimonis, line 1386: Same as above

uekerman:
@ajaust  Your mesh definition seems a bit odd. The actual extent of the mesh depends on the number of ranks. That's why you get different results in serial and parallel. 

You probably want sth like:
```python
coordinates = np.linspace(solverProcessIndex / solverProcessSize , (solverProcessIndex + 1) / solverProcessSize, dimension * N / solverProcessSize)
```

ajaust:
> @ajaust Your mesh definition seems a bit odd. The actual extent of the mesh depends on the number of ranks. That's why you get different results in serial and parallel.

I agree that it is odd, but that was done (partially) for testing. I still found the coupling results weird as all data of rank 1 is added up to the most right-hand point of the lower dimensional mesh. This means data of rank 1 is sent to its nearest neighbor. The more I think about it, the more sense it makes. 

This mixed-dimensional coupling is a bit more challenging than I expected. :-)

src/precice/config/ParticipantConfiguration.cpp
uekerman: No, I altered the comment to make this easier to understand.
"It's really the same mapping, not a duplicated one. Those are already checked for in MappingConfiguration."
Clear?

src/partition/ReceivedPartition.cpp
uekerman: Not sure what you mean :see_no_evil:

fsimonis: In the diff view, codecov claims that this line is not covered.

src/precice/config/ParticipantConfiguration.cpp
fsimonis: Got it! Yes its clear

PR #852, opened 2020-09-01
Fix: preCICE erroneously expects cyclic communicator for IMVJ with restart

### LINKED ISSUE #851: preCICE erroneously expects cyclic communicator for IMVJ with restart
I ran into a problem when using preCICE's IMVJ acceleration in parallel together with FEniCS (Dolfin). It seems it tries to set up a communicator that is cyclic even though it does not want/need one?!

I open the issue here as it seems related to the preCICE core and not to the preCICE Python bindings.

**Describe your setup**

- Ubuntu 18.04 LTS
- preCICE v2.1.0 installed together with the Python bindings version v2.1.0.1. 
- The solver the I invoke is written in Python using FEniCS 2019.1.0. 
- All packages have been installed via Spack with gcc@9.3.0.

You can build the environment via spack. Some packages are not in the official Spack repository, but can be found in the GitHub repository mentioned in the attachements.

**Describe the problem**

When I run preCICE from Python in conjunction with FEniCS I run into a problem, iff I use IMVJ acceleration and the participant computing the acceleration is running in parallel. The simulation either hangs when being finalized or it crashes on initialization.

**Step To Reproduce**

I have setup a minimal example in the repository below. 

1. Install preCICE and FEniCS via Spack given the configuration in the attachement.
2. Load the environment. I generate a module file with `spack env loads -r` and source the resulting module file in a fresh shell.
3. Start the two solvers 
```
mpirun -n 2 ./dummydolfinsolver.py precice-config-IMVJ.xml SolverA MeshA
mpirun -n 2 ./dummydolfinsolver.py precice-config-IMVJ.xml SolverB MeshB
```
4. The simulation will fail:
    1. If you run a release build of preCICE, `SolverB` will hang in `interface.finalize()`. `SolverA` finishes normally.
    2. If you run a debug build of preCICE, `SolverB` crashes while `SolverA` might crash and/or hang. All ranks of `SolverB` show the following error message:
```
ASSERTION FAILED
  Location:          void precice::acceleration::impl::ParallelMatrixOperations::establishCircularCommunication()
  File:              /tmp/jaustar/spack-stage/spack-stage-precice-2.1.0-hhsl5io7wrmayaqdc53b4bu57uo3o67y/spack-src/src/acceleration/impl/ParallelMatrixOperations.cpp:33
  Rank:              0
  Failed expression: _needCyclicComm
```
I could reproduce the problem on two different machines (Laptop + Server).

**Expected behaviour**

The simulation finishes nicely. 

**Additional context**

- The problem **only** occurs if `SolverB` runs in parallel. If `SolverB` is run with `mpirun -n 1` it finishes nicely. I guess that is related with the acceleration being computed on `SolverB`.
- The number of threads for `SolverA` seems to be irrelevant.
- If ILS acceleration is used, everything is fine. A corresponding preCICE configuration is in the repository.

**Attachements**

- Repository for reproducing the error: https://github.com/ajaust/precice-dolfin-imvj-parallel-bug

ajaust:
Fixes #851 

I have tested in on my laptop (Ubuntu 18.04) and it worked fine there. I would test it on another machine tomorrow as well.

### DIFF
--- a/src/acceleration/impl/ParallelMatrixOperations.cpp
+++ b/src/acceleration/impl/ParallelMatrixOperations.cpp
@@ -11,7 +11,7 @@ void ParallelMatrixOperations::initialize(bool needCyclicComm)
 {
   PRECICE_TRACE();
 
-  if (_needCyclicComm && (utils::MasterSlave::isMaster() || utils::MasterSlave::isSlave())) {
+  if (needCyclicComm && (utils::MasterSlave::isMaster() || utils::MasterSlave::isSlave())) {
     _needCyclicComm = needCyclicComm;
     establishCircularCommunication();
   } else {
--- a/src/acceleration/impl/ParallelMatrixOperations.cpp
+++ b/src/acceleration/impl/ParallelMatrixOperations.cpp
@@ -11,7 +11,7 @@ void ParallelMatrixOperations::initialize(bool needCyclicComm)
 {
   PRECICE_TRACE();
 
-  if (utils::MasterSlave::isMaster() || utils::MasterSlave::isSlave()) {
+  if (_needCyclicComm && (utils::MasterSlave::isMaster() || utils::MasterSlave::isSlave())) {
     _needCyclicComm = needCyclicComm;
     establishCircularCommunication();
   } else {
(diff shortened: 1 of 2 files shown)


ajaust:
The build worked for me on the server. The tests here seems to fail as Eigen seems to have left Bitbucket.

fsimonis:
@ajaust Thanks for the hint about Eigen. I fixed it on develop.

ajaust:
Great! I guess we can delete/close #854 in this case.

fsimonis:
@ajaust Yes, sorry I noticed the PR too late. Thanks for the fix though!

Issue #200, opened 2018-11-21
Compiling precice on Windows 10 with MSVC 2017 Community


fsimonis:
I am currently trying to build precice natively on Windows 10 using Visual Studio 2017 Community Edition.
I shall document this adventure and eventually produce a Wiki Page with a guide.

# Guide (WIP)

## 0) Visual Studio 2017
1) Download: https://visualstudio.microsoft.com/vs/
2) Install the C++ Language Features
3) Install the CMake support
4) _optional_ install the git and github modules

### 1) Boost
1) Download Windows Binaries: https://sourceforge.net/projects/boost/files/boost-binaries/1.68.0/boost_1_68_0-msvc-14.1-64.exe
2) Install to `some/dir`
3) Set the Environment Variables 
  *  `BOOST_INCLUDEDIR` to `some/dir/`
  *  `BOOST_LIBRARYDIR` to `some/dir/TODO` **TODO: CHECK**

### 2) Eigen3
1) Download the source: https://bitbucket.org/eigen/eigen/downloads/?tab=tags
2) Extract to `some/dir`
3) Set Environment Variable:
  `EIGEN3_ROOT_DIR` to `some/dir`
  _Note: The directory must contain a folder called "Eigen"_

### 3) Microsoft MPI SDK
1) Download **both** installers from https://www.microsoft.com/en-us/download/details.aspx?id=55494
  _Note: The first contains the library and the second the binary_
2) Install to `some/dir`
3) Verify that the following envrioment Variables were set during the installation:
  
    * `MSMPI_BIN`
    * `MSMPI_INC`
    * `MSMPI_LIB32`
    * `MSMPI_LIB64`
 
   _Note: you can use `set MSMPI` to list the variables_

### 4) LibXML2
1) Download the windows binaries: https://www.zlatkovic.com/pub/libxml/64bit/
2) Install to `some/dir`
3) Remember the location _this is not a joke_

BenjaminRueth:
We already have some instructions on building preCICE on windows with mingw in [our wiki](https://github.com/precice/precice/wiki/mingw-(Windows-version)).

S0mmerD:
Yes, but i think what @fsimonis tries is to build preCICE on Windows natively. This might have some advantages over minGW, like full Visual Studio support.

fsimonis:
The current status is that Visual Studio 15 2017 is **unable** to build preCICE.
It has some _severe_ issues regarding templates and crashes while compiling this line:
[...]

PR #855, opened 2020-09-04
Add BUILD_TESTING option (default ON) to in/exclude building of tests

mennodeij:
Good morning!

As per the title, I have added a BUILD_TESTING option that is by default ON (i.e. no change to default behaviour) to exclude building the tests if set to OFF.

What happened is that I'm compiling this with Intel compilers (v2020 update 1) on our cluster and there are compiler errors when compiling tests. It looks like template partial specialization that the Intel compiler does not understand. Building with GCC works flawlessly. 

Upon investigation I came across #713 where it is stated that Intel compilers are difficult to _integrate into testing_ and are therefore ignored, while it is correctly stated that the Intel compilers are relevant on HPC infrastructure. 

I would be happy to contribute further on Intel compiler support in the future.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ set(preCICE_SOVERSION ${preCICE_VERSION_MAJOR})
 
 #
 # Overview of this configuration
-# 
+#
 # PREAMBLE
 # Setup Options
 # Find Mandatory Dependencies
@@ -59,6 +59,7 @@ option(PRECICE_PythonActions "Python support" ON)
 option(PRECICE_Packages "Configure package generation." ON)
 option(PRECICE_InstallTest "Add test binary and necessary files to install target." OFF)
 option(BUILD_SHARED_LIBS "Build shared libraries by default" OFF)
+option(BUILD_TESTING "Build tests" ON)
 option(PRECICE_ALWAYS_VALIDATE_LIBS "Validate libraries even after the validatation succeeded." OFF)
 option(PRECICE_ENABLE_C "Enable the native C bindings" ON)
 option(PRECICE_ENABLE_FORTRAN "Enable the native Fortran bindings" ON)
@@ -270,7 +271,7 @@ if(CMAKE_VERSION VERSION_LESS "3.11")
   endif()
 endif()
 
-# Add precice as an empty target 
+# Add precice as an empty target
 add_library(precice ${preCICE_DUMMY})
 set_target_properties(precice PROPERTIES
   # precice is a C++11 project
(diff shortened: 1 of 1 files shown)


fsimonis:
@mennodeij Thank you very much for adding support for `BUILD_TESTING`!

I rebased the PR on develop, which fixes our CI. This will be merged once all checks pass.

fsimonis:
> What happened is that I'm compiling this with Intel compilers (v2020 update 1) on our cluster and there are compiler errors when compiling tests. It looks like template partial specialization that the Intel compiler does not understand. Building with GCC works flawlessly.

We can only investigate and solve problems we are aware of. So you are very welcome to fill out a bug report. 

The vast majority of intel related problem we encountered so far are related to the fact that intel uses a standard library based on the version of gcc detected at runtime. Meaning, in clusters one has to load both modules for gcc (5+) and the intel compiler.
 
> Upon investigation I came across #713 where it is stated that Intel compilers are difficult to integrate into testing and are therefore ignored, while it is correctly stated that the Intel compilers are relevant on HPC infrastructure.
>
> I would be happy to contribute further on Intel compiler support in the future.

The general issue here is that we have to use licensed software in a public CI.
We currently do not have the manpower to develop a proper solution for this, but it is definitely a long term goal.
So we would be very grateful if you could help with this. :rocket:

mennodeij:
I investigated further into the problem, and found that it was related to the `BOOST_TEST_INFO` macro when applied to a `map<int, vector<int>>` type object called vertexDistribution (src/partition/tests/ProvidedPartitionTest.cpp:201). When this line is removed, or when the variable name is quoted to be a string, the Intel compiler works. I don't have enough knowledge to understand if that is a viable solution to the problem.

I'm using this on a cluster that runs CentOS8, which has GCC 8.3.1 as default, so there is no need to load additional gcc modules to get the correct runtime for the Intel compiler.

And for running the Intel compiler on CI, there seem to be possibilities to do that without a license, see https://github.com/google/safeside/issues/106, but I understand that you need to have the manpower to set it up _and_ maintain it.

fsimonis:
> found that it was related to the BOOST_TEST_INFO macro when applied to a map<int, vector<int>> type object called vertexDistribution

This is likely related to the prettyprinter library. The template trickery also confused Microsoft's compiler when I tested windows support #200. The full compiler log of the error would be very helpful to find the exact issue. Would you mind new bug report based on it?

@mennodeij  **Thank you** for pointing out Intels oneAPI Docker Images! This looks super interesting!
I am a bit unsure what the conditions are though.

Travis is being awkward again, so I'll merge your PR. Thanks for the contribution!

Issue #801, opened 2020-07-11
Build is unsupported by default


fsimonis:
**Please describe the problem you are trying to solve.**
When configuring preCICE, the resulting library will be static by default.
We currently do not support using preCICE as a static library as both the CMake config and the pkg-config files do not support the static library. ( The current CMake config is missing depending targets)

Thus our build system is **unsupported by default**.

**Describe the solution you propose.**

There are many solutions to this problem (ranked by workload):

1) Always build a shared library and forget about static libraries altogether.   
  _This obviously prevents to use preCICE as a static library, which is unreasonable._
4) Set `BUILD_SHARED_LIBS` to `ON` if not explicitly specified. PR: #796 
  _This will change a CMake default and may surprise users. This could be regarded as a breaking change._
2) Always build both shared and static libraries and only directly support the shared case.
  Requires to build preCICE internally as an object library first to prevent recompilation. This is required for hidden symbol visibility, so we should do it at some point anyhow.
3) Extend the CMake script to fully support static libraries.
  This requires to export some internal dependencies and requires to propagate the dependencies to the dependant.

**Additional context**
* The solverdummy tests currently fail when building static. Due to the same issue as #716 .
* Static pkg-config #313 
* User running into problems with static CMake #716 
* The default for xSDK is to build as a shared library. However, this requirement will be shifted from CMake to Spack in the future.

MakisH:
>     2. Set `BUILD_SHARED_LIBS` to `ON` if not explicitly specified. PR: #796
>        _This will change a CMake default and may surprise users. This could be regarded as a breaking change._
Until now we had the the definition that _"breaking == breaking the API or configuration"_. We also had the rule that _"if nobody uses a hidden feature, killing it (if possible with some deprecation placeholder) is ok"_. We have also changed already a few times things in the build system and location of files during 1.x. This was, of course, terribly ugly, but I think in this case it doesn't make sense to think of it as breaking. Breaking the broken default?

>     3. Always build both shared and static libraries and only directly support the shared case.
>        Requires to build preCICE internally as an object library first to prevent recompilation. This is required for hidden symbol visibility, so we should do it at some point anyhow.

I don't like this idea, for two reasons:
- When both the static and the shared library are present, the linker often prefers the static, which can cause difficult-to-debug issues. Please correct me if I am wrong.
- Building both by default would increase the building time a lot, making us prefer again a non-default setting.

uekerman:
[...]

Issue #677, opened 2020-02-18
Problems linking to installed preCICE static library via CMake


j8asic:
I'm trying to link to static preCICE via `target_link_libraries(myTarget PRIVATE precice::precice)`, but it does not work for me. Instead, I must manually import all the preCICE’s dependencies:
```
find_package(precice REQUIRED CONFIG)
set(Boost_USE_STATIC_LIBS ON)
find_package (Boost REQUIRED COMPONENTS system program_options filesystem log log_setup thread unit_test_framework)
find_package (Eigen3 REQUIRED NO_MODULE)
```
And then:
```
target_link_libraries (${PROJECT_NAME}
    PRIVATE
    precice::precice
    Eigen3::Eigen
    Boost::boost
    Boost::filesystem
    Boost::program_options
    Boost::system
    Boost::log
    Boost::log_setup
    Boost::thread
    Boost::unit_test_framework
)
```
And then still it wont work, because JSON and prettyprint are not found…
```
/usr/bin/ld: cannot find -lJSON
/usr/bin/ld: cannot find -lprettyprint
```
If this is needed, not renaming `nlohmann_json` to `JSON` could help, so people can use their system wide lib. And/or copy externally used json/printing libraries when `make install`.

System: Arch OS using newest CMake, preCICE, Eigen3, Boost, etc.

fsimonis:
We currently don't recommend nor support building precice as a static library due to some of the problems you experienced.

Building a library as a static library exposes its privately linked dependencies. We use the header-only libraries eigen, json, numpy and prettyprint as private dependencies. They will be required by every target 
 linking to precice via the link interface despite them being pure inferface libraries.

The real problem is petsc though, as finding it requires the user to use the FindPETSc script as well as its dependencies, which we would have to install alongside preCICE.

We could try to remove the header only dependencies from the link interface, but this requires a lot of tedious work which is better invested in other areas.

That said, I am happy to help you incorporate a coltributed solution to the aforementioned problem into the project.

j8asic:
I see, thanks for the info. I don't have a problem using preCICE as shared library, rather was curious. 

I have to disagree with removing header-only dependencies from linking as tedious. Tedious is at the moment - creating CMake subdir for each header-only dependency. You could just include those folders in the main project.

fsimonis:
@j8asic I would like to keep this issue open as this it still is a problem that we need to address at some point.

fsimonis:
We decided to support only shared libraries in #2299

PR #857, opened 2020-09-08
Improve ctest solverdummy build

mennodeij:
I found that when I build preCICE (with GCC and OpenMPI from module) that the tests that run the solver dummies do not build.

That has now been fixed in this PR:

- use CXX as second language when building C and Fortran dummies
- find required libraries based on preCICE library configuration
- transfer the preCICE library configration from main build to solver dummy builds with extra flags in `add_precice_test_build_solverdummy` function

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ set(preCICE_SOVERSION ${preCICE_VERSION_MAJOR})
 
 #
 # Overview of this configuration
-# 
+#
 # PREAMBLE
 # Setup Options
 # Find Mandatory Dependencies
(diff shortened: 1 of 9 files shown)


fsimonis:
Hi, this is actually a common issue if you build preCICE as a static library.
Your PR adds the preCICE-internal dependencies to the solverdummies. However, this should be solved inside the CMake files that come with preCICE itself.

You can find a collection of problems and possible ways forward in this issue https://github.com/precice/precice/issues/801.
A similar issue to this one is https://github.com/precice/precice/issues/677.

I am closing this PR as we need a preCICE-native solution to this problem.

Issue #856, opened 2020-09-08
Build error with Intel compilers


mennodeij:
**Describe your setup**

Operating system (e.g. Linux distribution and version): _CentOS 8_
preCICE Version: origin/develop

**Describe the problem**

Compiling the file src/partition/tests/ProvidedPartitionTest.cpp fails with Intel Compiler v2020 update 1.
(see error message below)

**Step To Reproduce**
1. module load intel/compiler intel/mpi cmake
2. export CC=icc CXX=icpc FC=ifort
3. configure with cmake and build as normal

**Expected behaviour**
Build of precice including tests to complete without error.

**Additional context**

The problem can be fixed by removing line src/partition/tests/ProvidedPartitionTest.cpp:201 with use of the `BOOST_TEST_INFO` macro, or by quoting the variable name instead of trying to pretty-print it.

error message:
```
/home/mdeij/src/precice/thirdparty/prettyprint/include/prettyprint/prettyprint.hpp(152): error: invalid partial specialization -- class "pretty_print::print_container_helper<T, TChar, TCharTraits, TDelimiters>::printer<std::pair<const int, std::vector<int, std::allocator<int>>>> [with T=std::vector<int, std::allocator<int>>, TChar=char, TCharTraits=std::char_traits<char>, TDelimiters=pretty_print::delimiters<std::vector<int, std::allocator<int>>, char>]" is already fully specialized
  struct print_container_helper<T, TChar, TCharTraits, TDelimiters>::printer<std::pair<T1, T2>> {
                                                                     ^
          detected during:
            instantiation of class "pretty_print::print_container_helper<T, TChar, TCharTraits, TDelimiters> [with T=std::vector<int, std::allocator<int>>, TChar=char, TCharTraits=std::char_traits<char>, TDelimiters=pretty_print::delimiters<std::vector<int, std::allocator<int>>, char>]" at line 475
            instantiation of "std::enable_if<pretty_print::is_container<T>::value, std::basic_ostream<TChar, TCharTraits> &>::type std::operator<<(std::basic_ostream<TChar, TCharTraits> &, const T &) [with T=std::vector<int, std::allocator<int>>, TChar=char, TCharTraits=std::char_traits<char>]" at line 160
[...]

PR #858, opened 2020-09-08
Fix intel compiler issues

mennodeij:
Two small things that gave problems when compiling with Intel compilers:

1. fix the usage of the BOOST_TEST_INFO macro when directly applied to a variable of type `map<int,vector<int>>`
2. usage of `REAL(8)` instead of `REAL` when used with a `double` variable in C++

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -4,7 +4,7 @@ set(preCICE_SOVERSION ${preCICE_VERSION_MAJOR})
 
 #
 # Overview of this configuration
-# 
+#
 # PREAMBLE
 # Setup Options
 # Find Mandatory Dependencies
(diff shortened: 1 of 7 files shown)


fsimonis:
Thanks for the hints!

> fix the usage of the BOOST_TEST_INFO macro when directly applied to a variable of type map<int,vector<int>>

We need a solution to the problem instead of a quick fix. I will investigate this a bit further. Can you please provide the full error message emitted by the intel compiler?

> usage of REAL(8) instead of REAL when used with a double variable in C++

Good point! For the sake of consistency and readability, we should use `DOUBLE PRECISION` instead of `REAL(8)` though.
@MakisH could this be an issue in further Fortran codes?

mennodeij:
The full error message is found in #856

fsimonis:
You are quick! Just altered my answer. Thanks again!

MakisH:
> Good point! For the sake of consistency and readability, we should use `DOUBLE PRECISION` instead of `REAL(8)` though.
> @MakisH could this be an issue in further Fortran codes?

`DOUBLE PRECISION` would indeed be clearer, more consistent, and maybe safer than specifying the number of bytes. Fortran also provides [`kind`s](http://fortranwiki.org/fortran/show/Real+precision), but this is another topic.

In the Fortran module [we also currently use `real(8)`](https://github.com/precice/fortran-module/blob/9826f277e3302cc1aef50741530538bd9d8d23c7/examples/solverdummy/solverdummy.f03#L11).

fsimonis:
Let's split this PR into 2 parts.

I committed your Fortran fix in 10a958233cc372f41c0df8c5025183b552edb081 and annotated you as the author to keep your contribution in the system.
The intel issue is more delicate and I am currently attempting to figure out how to solve the issue.

Thank you for your contribution!

PR #861, opened 2020-09-17
Fix and Format XML

fsimonis:
The aim of this PR is to consistently format our XML files, which revealed some small issues.

This PR:
* adds formatting support based on xmllint (Debian package name: `libxml-utils`).
* formats our xml files using the aforementioned tool. This fixed the incorrectly escaped sign '<' in our log filter.
* adds decoding support for escaped characters (e.g `&lt;`)

These changes are backwards compatible, so incorrectly escaped xml files will still work.

### DIFF
--- a/examples/solverdummies/precice-config.xml
+++ b/examples/solverdummies/precice-config.xml
@@ -1,52 +1,41 @@
 <?xml version="1.0"?>
-
 <precice-configuration>
-
   <log>
-    <sink type="stream" output="stdout"  filter= "%Severity% > debug" format="preCICE:%ColorizedSeverity% %Message%" enabled="true" />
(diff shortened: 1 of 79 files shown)


MakisH:
This would also be useful in the tutorials repository.

fsimonis:
So, I had a chat with @uekerman about the changes.
It turns out that most formatters do not allow to specify style options. A especially big problem is the vertical spacing. Neither `xmllint` not `tidy` not `xmlstarlet` allow to keep those.

So our best option is to use [prettier](https://prettier.io/), which tidies up xml files instead of completely reformatting them.

So I will have to change a few things in this PR.

fsimonis:
Update on prettier: the plugin `prettier-xml` has issues with indenting attributes when preserving whitespaces.
Using the python module `lxml` also requires additional work.

Issue #850, opened 2020-08-29
Check to prevent duplicated entries in acceleration config is too strict


ajaust:
Hi,

I ran into the following problem and was wondering whether your assertions are too strict or whether I did something wrong the whole time.

**Describe your setup**

- Ubuntu 18.04 LTS
- preCICE v2.1.0 installed together with the Python bindings version v2.1.0.1. 
- The solver the I invoke is written in Python using FEniCS 2019.1.0. 
- All packages have been installed via Spack with gcc@9.3.0.

**Describe the problem**

When starting the simulation it complaints about the duplicate occurence of the data tag `Displacement`. 

```
<class '__main__.InputParameters'>
Configuring preCICE
(0) 12:31:40 [acceleration::AccelerationConfiguration]:154 in xmlTagCallback: ERROR: You have provided a subtag <data name="Displacement" ... /> more than once in your <acceleration:.../>. Please remove the duplicated entry.
```

This configuration has been valid in preCICE v2.0.X and (with some changes regarding keywords) in v1.6.X.

**Step To Reproduce**

Run the simulation with the attached configuration or with some keyword that appears twice. 

This is related to my coupling described on discourse:
- https://precice.discourse.group/t/two-couplings-to-one-interface-mixed-dimensional-coupling/68/5
- https://precice.discourse.group/t/two-couplings-to-one-interface-reloaded/260/10

**Expected behaviour**
I expect the simulation to run as normally with duplicate data name tags as before since the mappings are all between different meshes.

**Additional context**

Possibly related to https://github.com/precice/precice/issues/698

**Attachements**

- [precice-config.xml.txt](https://github.com/precice/precice/files/5145089/precice-config.xml.txt)

ajaust:
Just for completenes: Changing the configuration to for the Displacements to have different identifiers `DisplacementTop` and `DisplacementBottom` fixes the problem. It is still unclear to me whether the first configuration that I have used before is actually illegal.

**Attachements**
- [precice-config-modified.xml.txt](https://github.com/precice/precice/files/5145152/precice-config-modified.xml.txt)

uekerman:
The error we introduced in https://github.com/precice/precice/commit/7b9fb630f5929ecf04b84593e72ea73bb6a819df is too strict. Identical data names on different meshes should be allowed.
@ajaust: As far as I see your config is legal.

BenjaminRueth:
[...]

PR #863, opened 2020-09-21
Fix error message for unique acceleration subtags

BenjaminRueth:
Check for unique combinations of mesh AND data name. Not only data name.

Closes https://github.com/precice/precice/issues/850.

I verified the correct behaviour with the two attached configuration files and the cpp solverdummy:

* [precice-config-error.xml](https://github.com/precice/precice/files/5255221/precice-config-error.txt)
* [precice-config-ok.xml](https://github.com/precice/precice/files/5255222/precice-config-ok.txt)

### DIFF
--- a/src/acceleration/config/AccelerationConfiguration.cpp
+++ b/src/acceleration/config/AccelerationConfiguration.cpp
@@ -149,10 +149,11 @@ void AccelerationConfiguration::xmlTagCallback(
     _config.relaxationFactor = callingTag.getDoubleAttributeValue(ATTR_VALUE);
   } else if (callingTag.getName() == TAG_DATA) {
     std::string dataName = callingTag.getStringAttributeValue(ATTR_NAME);
-    auto        success  = _uniqueDataNames.insert(dataName);
+    std::string meshName = callingTag.getStringAttributeValue(ATTR_MESH);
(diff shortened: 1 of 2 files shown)


src/acceleration/config/AccelerationConfiguration.cpp
```diff
@@ -149,10 +149,11 @@ void AccelerationConfiguration::xmlTagCallback(
     _config.relaxationFactor = callingTag.getDoubleAttributeValue(ATTR_VALUE);
   } else if (callingTag.getName() == TAG_DATA) {
     std::string dataName = callingTag.getStringAttributeValue(ATTR_NAME);
-    auto        success  = _uniqueDataNames.insert(dataName);
+    std::string meshName = callingTag.getStringAttributeValue(ATTR_MESH);
+    auto        success  = _uniqueDataAndMeshNames.insert(std::pair<std::string, std::string>(dataName, meshName));
```
fsimonis, line 153: You could use `. emplace(dataName, meshName) ` here to remove the type boilerplate.

BenjaminRueth: Do you mean something like this?

```suggestion
    auto            ret = _uniqueDataAndMeshNames.emplace(dataName, meshName);
    auto        success = ret.second;
```

BenjaminRueth: And, of course, thanks for the hint!

PR #865, opened 2020-09-23
Refactor CouplingData

BenjaminRueth:
The current design of `CouplingData` duplicates data and makes direct access to `mesh::Data::_values` possible via a pointer stored in `CouplingData`.

This PR tries to improve the class design, removes duplicate data (as far as possible) and introduces a clear hierarchy, where `mesh::Data::_values` can only accessed using `mesh::Data::values()`.

*Note:* I will remove the `deprecated`attributes before merging, since this is C++14.


src/acceleration/test/AccelerationMasterSlaveTest.cpp
```diff
@@ -416,70 +417,70 @@ BOOST_AUTO_TEST_CASE(testVIQNIMVJpp)
...
-    BOOST_TEST(testing::equals((*data.at(1)->values)(1), 1.99000000000000010214e-01), (*data.at(1)->values)(1));
+    BOOST_TEST(testing::equals(data.at(0)->values()(0), 1.00000000000000000000e+00), data.at(0)->values()(0));
+    BOOST_TEST(testing::equals(data.at(0)->values()(1), 1.01000000000000000888e+00), data.at(0)->values()(1));
+    BOOST_TEST(testing::equals(data.at(1)->values()(0), 1.99000000000000010214e-01), data.at(1)->values()(0));
+    BOOST_TEST(testing::equals(data.at(1)->values()(1), 1.99000000000000010214e-01), data.at(1)->values()(1));
     utils::append(newdvalues, 10.0);
     utils::append(newdvalues, 10.0);
   }
 
-  data.begin()->second->values = &newdvalues;
```
BenjaminRueth, line 455: This use-case is removed by this PR. One cannot anymore just provide new data and replace the pointer. Instead, one can only overwrite the data `CouplingData` is pointing to or one has to provide a new `PtrData`.

From my point of view it is good to remove this use-case, since one can otherwise easily introduce data that is not consistent with the other members of `CouplingData` (dimension...).

src/action/ComputeCurvatureAction.cpp
```diff
@@ -33,7 +33,7 @@ void ComputeCurvatureAction::performAction(
 
   if (getMesh()->getDimensions() == 2) {
     dataValues = Eigen::VectorXd::Zero(dataValues.size());
-    //assign(dataValues) = 0.0;
+    //assign(values) = 0.0;
```
uekerman, line 36: :thinking: Why that?

uekerman (approved):
Makes sense :+1:

src/action/ComputeCurvatureAction.cpp
BenjaminRueth: I think this got automatically renamed, but we can also just kill the comment, if you want? To me it does not really look like important informaiton.

uekerman: Good practice would be to never leave commented out code. But we are not living in a perfect world :see_no_evil:

BenjaminRueth: It's gone now 👍 One step closer to a perfect world 🚀

src/cplscheme/CouplingData.hpp
```diff
@@ -11,19 +11,33 @@ namespace cplscheme {
 struct CouplingData {
   using DataMatrix = Eigen::MatrixXd;
 
-  /// Data values of current iteration.
-  Eigen::VectorXd *values;
+  /// Returns a reference to the data values.
+  Eigen::VectorXd &values()
+  {
+    return data->values();
```
fsimonis, line 17:
```suggestion
    PRECICE_ASSERT(data != nullptr);
    return data->values();
```

src/cplscheme/CouplingData.hpp
```diff
@@ -11,19 +11,33 @@ namespace cplscheme {
...
+  }
+
+  /// Returns a const reference to the data values.
+  const Eigen::VectorXd &values() const
+  {
+    return data->values();
```
fsimonis, line 23:
```suggestion
    PRECICE_ASSERT(data != nullptr);
    return data->values();
```

src/cplscheme/CouplingData.hpp
```diff
@@ -36,16 +50,15 @@ struct CouplingData {
...
+      bool          requiresInitialization)
+      : data(data),
         mesh(mesh),
-        requiresInitialization(requiresInitialization),
-        dimension(dimension)
+        requiresInitialization(requiresInitialization)
   {
-    PRECICE_ASSERT(values != NULL);
+    PRECICE_ASSERT(data != nullptr);
+    PRECICE_ASSERT(mesh != nullptr);
```
fsimonis, line 61: This introduces an invariance to this `struct`, which cannot be maintained due to the public member access.
From a design standpoint, this should now be a `class`, which protects `data` and `mesh`. Let's postpone this change to another time though.

src/cplscheme/CouplingData.hpp
```diff
@@ -11,19 +11,33 @@ namespace cplscheme {
 struct CouplingData {
```
BenjaminRueth, line 11:
```suggestion
struct CouplingData {  // @todo: should be a class from a design standpoint. See https://github.com/precice/precice/pull/865#discussion_r495825098
```

src/cplscheme/CouplingData.hpp
BenjaminRueth: Meaning: Don't change it to a `class` now, but keep it on the radar? I would suggest to at least add a `@todo` in the code. See https://github.com/precice/precice/pull/865/files/3fd1911ecfe116124b84ddbc2f876d1b12fb99d1#r495835946.

BenjaminRueth: Resolved by ca62d12

PR #866, opened 2020-09-24
Add XML formatter

fsimonis:

* Add a custom xml-formatter based on the python library `lxml`
* Add support to `format-all`
* Format all xml files in the code-base

**Motivation**

Our XML files follow no style and there is no formatter available that produce human-readable xml files.

**Additional Information**

Related to #862

### DIFF
--- a/src/cplscheme/tests/serial-explicit-coupling-datainit.xml
+++ b/src/cplscheme/tests/serial-explicit-coupling-datainit.xml
@@ -1,21 +1,28 @@
-<?xml version="1.0"?>
-
+<?xml version="1.0" encoding="UTF-8" ?>
 <configuration>
-   <data:scalar name="Data0"/>
-   <data:vector name="Data1"/>
-   <data:scalar name="Data2"/>
-   <mesh name="Mesh">
-      <use-data name="Data0"/>
-      <use-data name="Data1"/>
-      <use-data name="Data2"/>
-   </mesh>
-   <m2n:sockets from="Participant0" to="Participant1"/>
-   <coupling-scheme:serial-explicit>
-      <participants first="Participant0" second="Participant1"/>
-      <time-window-size value="0.1" method="fixed"/>
-      <max-time-windows value="1"/>
-      <exchange data="Data0" mesh="Mesh" from="Participant1" to="Participant0"/>
-      <exchange data="Data1" mesh="Mesh" from="Participant1" to="Participant0" initialize="on"/>
-      <exchange data="Data2" mesh="Mesh" from="Participant0" to="Participant1"/>
-   </coupling-scheme:serial-explicit>
+  <data:scalar name="Data0" />
+
+  <data:vector name="Data1" />
+
+  <data:scalar name="Data2" />
+
+  <mesh name="Mesh">
+    <use-data name="Data0" />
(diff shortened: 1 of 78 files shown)


src/cplscheme/tests/serial-explicit-coupling-datainit.xml
```diff
@@ -1,21 +1,28 @@
...
-      <max-time-windows value="1"/>
-      <exchange data="Data0" mesh="Mesh" from="Participant1" to="Participant0"/>
-      <exchange data="Data1" mesh="Mesh" from="Participant1" to="Participant0" initialize="on"/>
-      <exchange data="Data2" mesh="Mesh" from="Participant0" to="Participant1"/>
-   </coupling-scheme:serial-explicit>
+  <data:scalar name="Data0" />
+
+  <data:vector name="Data1" />
+
+  <data:scalar name="Data2" />
```
uekerman, line 7: Could we get rid of the empty lines here? So, no new line between `data` tags?

src/cplscheme/tests/serial-explicit-coupling-datainit.xml
```diff
@@ -1,21 +1,28 @@
...
+    <participants first="Participant0" second="Participant1" />
+
+    <time-window-size value="0.1" method="fixed" />
+
+    <max-time-windows value="1" />
+
+    <exchange data="Data0" mesh="Mesh" from="Participant1" to="Participant0" />
+    <exchange data="Data1" mesh="Mesh" from="Participant1" to="Participant0" initialize="on" />
+    <exchange data="Data2" mesh="Mesh" from="Participant0" to="Participant1" />
+  </coupling-scheme:serial-explicit>
```
uekerman, line 27: Could we avoid empty lines within a coupling-scheme tag?

uekerman (commented):
The formatting of proper preCICE configs looks good.
For partial configs, which we use in some unit tests, however, the formatting now gives some new lines, which could be avoided.
Could we modify the formatter to also handle these cases?
If not straight forward let's not worry about this.

uekerman:
Hope this is not a silly question: why `config-formatter` and not `config-formatter.py`? :thinking:

src/cplscheme/tests/serial-explicit-coupling-datainit.xml
fsimonis: Good point. It currently groups on the full name of the tags, hence the 3 groups.

src/cplscheme/tests/serial-explicit-coupling-datainit.xml
fsimonis: This looks like a bug. Grouping should be disabled after the second level.

fsimonis: I looked into this.
This mentioned configuration is only a partial configuration as it does not contain the `<precice-configuration>` and `<solver-interface>` tags.
We could add an exception, however this would add more complexity to the formatter.

I would suggest to ignore formatting for these edge-cases.

fsimonis:
It's not a silly question, but I'll give you a silly answer. Using `config-formatter` feels more command-like without an extension.

uekerman (approved):

PR #871, opened 2020-10-03
Update removes build/last and adds mpirun -np 4

BenjaminRueth:
Documentation update

* `tools/compileAndTest.py` was removed. We now use `ctest`
* `build/last` was removed, when migrating to cmake. `testprecice` is now directly in `build`
* `mpirun -np 4` is required. If running `testprecice` without `mpirun -np 4` I get the error:

```
~/precice$ ./build/testprecice 
ERROR: The tests require at least 4 MPI processes.
```

### DIFF
--- a/docs/documents/050_tests.dox
+++ b/docs/documents/050_tests.dox
@@ -6,11 +6,10 @@ Running and writing tests
 
 Running
 ------
-The ```tools/compileAndTest.py``` script simplifies running all tsts.
+Use ```ctest``` or ```make test``` within the build folder to run all tests.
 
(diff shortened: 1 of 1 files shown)


MakisH (approved):
The `compileAndTest.py` actually still exists: https://github.com/precice/precice/blob/develop/tools/building/compileAndTest.py

But I agree that we should not refer to this as the main way to run the tests. This is also not the only documentation for `compileAndTest.py`, it is also described in the [tools README](https://github.com/precice/precice/blob/develop/tools/building/README.md).

The rest looks good to me.

We should maybe also add a link from the [wiki](https://github.com/precice/precice/wiki/Tests).

fsimonis (approved):

Issue #133, opened 2018-07-11
First Steps Waveform Relaxation / Subcycling


BenjaminRodenberg:
Below you find a sketch on a feature idea. The sketch is based on the [elastictube1d example](https://github.com/precice/elastictube1d), where pressure (p) and cross section area (a) are coupled to simulate an elastic tube with fluid flow. On long term we want to allow high order time integration with subcycling inside preCICE using waveform relaxation. 

This issue is a first step in this direction. We want to perform a linear interpolation inside preCICE. This helps, if provided data by one participant is not sufficient (for example due to subcycling).

A first experimental implementation going into this direction can be found in https://github.com/precice/elastictube1d/blob/convergenceStudies/PythonTube/FluidSolver.py lines 131-134.

![grafik](https://user-images.githubusercontent.com/5740604/42563903-af978808-84ff-11e8-9c29-ca624721221d.png)

BenjaminRueth:
somehow related: https://github.com/precice/precice/issues/149

For waveform relaxation we have to be able to exchange multiple (arbitrarily many) pieces of data per interface grid node per timestep. This is not supported by the currently existing API functions. Currently we can only exchange one double per grid node using `Send/WriteBlockScalarData` (scalar quantities like temperature) or `dim` doubles per grid node using `Send/WriteBlockVectorData` (vector valued quantities like velocity).

Sending more than `dim` values using `Send/WriteBlockVectorData` might be a possibility, but to my understanding this is not the correct use of the API and this function should be reserved for vector values (physical) quantities. Another reason, for not using an array to send multiple data values per interface grid node: For vector data index computation becomes vulnerable to error.

uekerman:
We collected four ways how to tackle waveform relaxation in preCICE:

(1) **Adapter Variant**: Construct higher-order polynomials in the adapter by misusing preCICE and exchanging multiple data values. Not use `VectorData` as above, but simply using many data values. This is no problem as we know a-priori how many we need if the solvers do not use adaptive timestep sizes. All this is hacky, but this is only a first step to have more testcases. @BenjaminRueth is looking into that. Of course, no changes to the API are necessary.

(2) **Black-box Variant**: preCICE constructs higher-order polynomials from pure black-box data. For the solvers everything seems like before, just as subcycling, but with better data. No changes to the API are necessary. We do not know yet how to do this numerically in a general way, but linear polynomials are not complicated to do. I will implement a first and document in this issue. 

[...]

PR #876, opened 2020-10-06
Refactor extrapolateData

BenjaminRueth:
* `storeData()` is now always called in `BaseCouplingScheme::accelerate()`
* update corresponding tests for extrapolation
* removes storing of data from `BaseCouplingScheme::extrapolateData(DataMap &data)`

Related to #133


src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -265,7 +265,6 @@ void BaseCouplingScheme::extrapolateData(DataMap &data)
       PRECICE_DEBUG("Extrapolate data: " << pair.first);
       PRECICE_ASSERT(pair.second->oldValues.cols() > 1);
       Eigen::VectorXd &values       = pair.second->values();
-      pair.second->oldValues.col(0) = values;  // = x^t
       values *= 2.0;                           // = 2*x^t
       values -= pair.second->oldValues.col(1); // = 2*x^t - x^(t-1)
       utils::shiftSetFirst(pair.second->oldValues, values);
```
uekerman, line 270: In principle you also store data in `shiftSetFirst`, right? Question is whether you want to change this in this PR or later?

uekerman (commented):
It's a bit bad practice to change a function from private to public just to allow usage in tests. Should not be overly complicated to work with friend declarations here, right? (compare how `extrapolateData` is called)

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRueth: I would prefer changing this later. https://github.com/BenjaminRueth/precice/tree/package_time overhauls this part of the extrapolation. I think changing it in this PR is not really necessary.

BenjaminRueth:
> It's a bit bad practice to change a function from private to public just to allow usage in tests. Should not be overly complicated to work with friend declarations here, right? (compare how `extrapolateData` is called)

Encapsulation should be fixed now. See d3036ae and 8238bea.

uekerman (approved):

Issue #870, opened 2020-10-02
MacOS Test Failure: Socket Communication


oguzziya:
**Describe the Setup**
Operating System: MacOS Catalina 10.15.5
preCICE Version: develop
Boost Version: 1.73.0

**Describe the problem**
While compiling with Boost 1.73.0, all the tests throw the error:

```
Test setup error: boost::runtime::arg_type_mismatch: Access with invalid type for argument corresponding to parameter log_level
```

which is related to the change in Boost 1.73.0. This problem can be solved using by defining `logLevel` in `tests/main.cpp` to `log_level_success`.

After changing, preCICE compiles successfully. Then, the following tests fail
```

 3 - precice.com (Failed)
          4 - precice.cplscheme (Failed)
          6 - precice.m2n (Failed)
         11 - precice.partition (Failed)
         13 - precice.serial (Failed)
         14 - precice.parallel (Failed)
         16 - precice.testing (Failed)
         19 - precice.solverdummy.build.cpp (Failed)
         22 - precice.solverdummy.run.cpp-cpp (Not Run)
         23 - precice.solverdummy.run.c-c (Timeout)
         24 - precice.solverdummy.run.fortran-fortran (Timeout)
         25 - precice.solverdummy.run.cpp-c (Failed)
         26 - precice.solverdummy.run.cpp-fortran (Failed)
         27 - precice.solverdummy.run.c-fortran (Timeout)
```

with error:

```
libc++abi.dylib: terminating with uncaught exception of type boost::wrapexcept<boost::system::system_error>: shutdown: Socket is not connected
```

**Step To Reproduce**
1. Set `logLevel = log_level_success` in `tests/main.cpp`
2. `cmake -DBUILD_SHARED_LIBS=ON -DCMAKE_CXX_FLAGS="-D_GNU_SOURCE" ..` (refer https://github.com/boostorg/stacktrace/issues/88 and https://github.com/boostorg/stacktrace/issues/94)
3. `make` 
4. `make test_base`

**Expected behaviour**
preCICE compiles without need for a code change and tests pass successfully.

PR #877, opened 2020-10-06
Fix MacOS Compilation and Tests

oguzziya:
This PR applies following changes to be able to compile and run on MacOS

- Conditional compile definition `_GNU_SOURCE` is added. (Refer to https://github.com/boostorg/stacktrace/issues/94 and https://github.com/boostorg/stacktrace/issues/88) 

- With Boost 1.73.0, a new `log_level` is introduced, which breaks the preCICE logging and causing all tests to fail on MacOS. (Refer to https://github.com/precice/precice/issues/870 and https://precice.discourse.group/t/testing-precice-installation-on-macos/296). This is solved by setting `log_level` to debug manually.

- During the socket shutdown, there was an uncaught exception which causes most of the tests to fail. The main reason behind this is that the socket is tried to be shut down while it is already shut down. It is solved by catching the exception.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -295,7 +295,7 @@ target_link_libraries(precice PRIVATE
   Boost::unit_test_framework
   )
 if(UNIX OR APPLE OR MINGW)
-  target_compile_definitions(precice PRIVATE _GNU_SOURCE BOOST_ALL_DYN_LINK BOOST_ASIO_ENABLE_OLD_SERVICES BOOST_GEOMETRY_DISABLE_DEPRECATED_03_WARNING)
+  target_compile_definitions(precice PRIVATE _GNU_SOURCE)
   target_link_libraries(precice PRIVATE ${CMAKE_DL_LIBS})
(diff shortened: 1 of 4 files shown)


CMakeLists.txt
```diff
@@ -295,6 +295,7 @@ target_link_libraries(precice PRIVATE
   Boost::unit_test_framework
   )
 if(UNIX OR APPLE OR MINGW)
+  target_compile_definitions(precice PRIVATE _GNU_SOURCE BOOST_ALL_DYN_LINK BOOST_ASIO_ENABLE_OLD_SERVICES BOOST_GEOMETRY_DISABLE_DEPRECATED_03_WARNING)
```
fsimonis, line 298: The remaining definitions are already added above.
```suggestion
  target_compile_definitions(precice PRIVATE _GNU_SOURCE)
```

fsimonis (approved):

fsimonis:
I'll merge it after adding the changelog

PR #880, opened 2020-10-16
Improve Test Log

fsimonis:

- Change testing log to a more informative format. The new format is `|`-based.
- Add an additional log file stream for tests
- Add fractional seconds to the test timestamp
- Remove the colorized output when writing the test log file

**Motivation**

The test refactoring allowed to set the rank information correctly. This is good for sanity but bad for readability as our default log format does not include the name of the participant.
The format is also difficult to understand as the log off all 4 threads are interleaved. Changing the log format to a machine-readable format has the upside that the log can be imported to other tools.

Using `|` as separators does not collide with the log content and is importable by LibreOffice or others.
The LibreOffice autofilter makes understanding the logs way easier.

To simplify user support and importing such logs, I added an additional sink that writes the log to `test.log`.


MakisH (approved):
I like this change. The new `test.log` is stored in `<build>/TestOutput/test/` and opening it in a CSV processing tool such as Calc is a great idea: one can e.g. filter rows that only come from one participant etc.

Would it make sense to provide some runtime swtich to also enable debug output? For example, using a [`log.conf`](https://github.com/precice/precice/wiki/Logging-Configuration#logconf). If this is possible already, we should document it.

fsimonis:
@MakisH the logging of the tests can be controlled using the following:
1. a `log.conf`file in the directory
2. the log level specified with `BOOST_TEST_LOG_LEVEL`
3. the log level passed to the `testprecice` executable (This is not compatible with boost versions though)

I added some of this information during the rework of the docs.

fsimonis:
One major problem with this approach is that the messages of the test framework are not captured in the log files.

fsimonis:
Multiple ranks write to the same file which results in broken output.
We need to create one file per rank and merge them at the end of the test run.
As ranks can fail, we have to implement the merge in ctest.

fsimonis:
I noticed that using `|` as separator does collide with the coupling info in advance. Shall we change that? Is that a breaking change? 

https://github.com/precice/precice/blob/442becf3f7d2fe6900f76ee5b8b39b1c49a8cf7f/src/cplscheme/BaseCouplingScheme.cpp#L402-L411

uekerman:
> Is that a breaking change?

In my opinion, no.

MakisH:
> I noticed that using `|` as separator does collide with the coupling info in advance. Shall we change that? Is that a breaking change?

Why would it be? For changing the logging?

fsimonis:
Right, I forgot that we also output the information as a separate file.

fsimonis:
Here are 2 examples from our tests. How shall we reformat them?

```
advance: it 1 | dt# 3 of 10 | t 2 | dt 1 | max dt 1 | ongoing yes | dt complete yes |
advance: it 7 of 100 | dt# 1 of 1 | t 0 | dt 1 | max dt 1 | ongoing yes | dt complete no | read-iteration-checkpoint|
```

Using a comma instead looks a bit crazy:
```
advance: it 1, dt# 3 of 10, t 2, dt 1, max dt 1, ongoing yes, dt complete yes,
advance: it 7 of 100, dt# 1 of 1, t 0, dt 1, max dt 1, ongoing yes, dt complete no, read-iteration-checkpoint, 
```

So we could also use a `key: value` format similar to the mapping state:
```
advance: iteration: 1, time-window: 3 of 10, time: 2, time-window-size: 1, max-timestep-length: 1, ongoing: yes, dt-complete: yes,
advance: iteration: 7 of 100, time-window: 1 of 1, time: 0, time-window-size: 1, max-timestep-length: 1, ongoing: yes time-window-complete: no, read-iteration-checkpoint, 
```

uekerman:
I like the `key: value` formatting.

BenjaminRodenberg:
> I like the `key: value` formatting.

I also like it. I only see the problem that `advance:` conflicts with this. Maybe 

```
[advance] iteration: 1, time-window: 3 of 10, time: 2, time-window-size: 1, max-timestep-length: 1, ongoing: yes, dt-complete: yes,
```

fsimonis:
>      I also like it. I only see the problem that advance: conflicts with this. Maybe

@BenjaminRodenberg The log format will change, so the actual output is the following:
```
17:04:26.511082|SolverTwo|0|cplscheme::BaseCouplingScheme|l83|sendData|Number of sent data sets = 1
17:04:26.511142|SolverTwo|0|cplscheme::BaseCouplingScheme|l228|advance|Time window completed
17:04:26.511323|SolverTwo|0|impl::SolverInterfaceImpl|l431|advance|iteration: 1, time-window: 2 of 10, time: 1, time-window-size: 1, max-timestep-length: 1, ongoing: yes, time-window-complete: yes, 
17:04:26.511373|SolverTwo|0|impl::SolverInterfaceImpl|l433|advance|Handle exports
```

I also fixed some linebreak in the output due to Eigen.
The output of `test.log` can now be fully read by a csv application. :tada:

PR #889, opened 2020-11-02
Fix double-usage of i-parameter in EventUtils.cpp

mennodeij:
Minor issue resulting in a compiler warning.


fsimonis:
Thank you for the contribution!

PR #892, opened 2020-11-04
Fix boost log_level for macOS

oguzziya:
[Previous fix for macOS](https://github.com/precice/precice/pull/877) was only for boost version 1.73.0. Homebrew updated boost to 1.74.0 and error still persist, independent of boost version. Therefore, the previous implementation is extended for macOS systems in general.


fsimonis (approved):

fsimonis:
This PR is a direct counterpart to the following test:
https://github.com/boostorg/test/blob/develop/example/unit_test_example_16.cpp#L18

Issue #847, opened 2020-08-06
Partition tests fail on CentOS / Fedora


MakisH:
**Describe your setup**

- Operating system (e.g. Linux distribution and version): CentOS 8 (on a 2-cores VM)
- preCICE Version: v2.1.0
- OpenMPI 4.0.2
- GCC 8.3.1

**Describe the problem**

After a successful build, the `precice.partition` tests fail (reported as "timeout", but this seems to be a failure):

![Screenshot from 2020-08-06 22-12-16](https://user-images.githubusercontent.com/4943683/89578438-6b658080-d832-11ea-83b5-30723039be10.png)

This is the main error:

https://github.com/precice/precice/blob/v2.1.0/src/partition/ReceivedPartition.cpp#L247-L255

Very verbose output: 
[partition.log](https://github.com/precice/precice/files/5037945/partition.log)

**Step To Reproduce**
1. Install CentOS 8
2. Get all the [dependencies](https://github.com/precice/precice/wiki/Dependencies#centos-8)
3. Configure and build preCICE with shared libs and without PETSc
4. Run `make test` or `ctest -V`. The verbose output presented here uses [`BOOST_TEST_LOG_LEVEL="all"`](https://github.com/precice/precice/wiki/Tests#getting-more-output)

**Expected behaviour**
The tests pass in a few seconds.

**Additional context**
I have not seen this test fail in any recent Ubuntu or Debian system. I have not tried CentOS 7, maybe @IshaanDesai knows more.

MakisH:
I get the exact same issue on Fedora 32 with OpenMPI v4.0.4, GCC 10.2.1, and Boost 1.69.

The solver dummies seem to run (in serial), but this is probably not helpful.

uekerman:
The problem appears to be in the test 

https://github.com/precice/precice/blob/54fcdb83f9bef91c2cb22125eaa5d9f179ae2b67/src/partition/tests/ReceivedPartitionTest.cpp#L737

The bounding box from slave 2 is sent correctly from slave to master, but the filtered mesh is lost while sending back. 

So, sth goes wrong here:

https://github.com/precice/precice/blob/54fcdb83f9bef91c2cb22125eaa5d9f179ae2b67/src/com/CommunicateMesh.cpp#L27

I suggest more debugging in this function.

MakisH:
[...]

Issue #879, opened 2020-10-15
MacOS Test Failure


oguzziya:
**Describe your setup**

Operating system: MacOS Catalina 10.15.7
preCICE Version: develop

**Describe the problem**
Partition tests and solverdummies are failing during testing.

```bash
         11 - precice.partition (Failed)
         19 - precice.solverdummy.build.cpp (Failed)
         20 - precice.solverdummy.build.c (Failed)
         21 - precice.solverdummy.build.fortran (Failed)
         22 - precice.solverdummy.run.cpp-cpp (Not Run)
         23 - precice.solverdummy.run.c-c (Not Run)
         24 - precice.solverdummy.run.fortran-fortran (Not Run)
         25 - precice.solverdummy.run.cpp-c (Not Run)
         26 - precice.solverdummy.run.cpp-fortran (Not Run)
         27 - precice.solverdummy.run.c-fortran (Not Run)
```

Partition tests failure is due to only this:
```bash
PartitionTests/ReceivedPartitionTests/RePartitionRBFLocal2D2": check pSolidzMesh->vertices()[5].isOwner() == true has failed [false != true]
Failure occurred in a following context:
    Mesh "SolidzMesh", dimensionality = 2:
GEOMETRYCOLLECTION(
POINT (0 1.95), POINT (0 2.1), POINT (0 4.5), POINT (0 5.95), POINT (0 6.1),
LINESTRING (0 1.95, 0 2.1), LINESTRING (0 2.1, 0 4.5), LINESTRING (0 4.5, 0 5.95), LINESTRING (0 5.95, 0 6.1)
```
I have no idea why this happens and its relation to the operating system. 

For the solverdummies, `build/Solverdummies/c` and `build/Solverdummies/cpp` directories are empty. When I run the solverdummies in `examples/solverdummies` everything seems to be fine. I suspect that there is a bash script difference between MacOS and Linux. When I run `ctest --output-on-failure`, encoding in terminal changes to cryptic language just after it enters solvedummy tests.

**Step To Reproduce**
1. cmake -DBUILD_SHARED_LIBS=ON ..
2. make test_base

**Expected behaviour**
All the tests should pass without need for any interaction.

uekerman:
I had a look into the `partition` failure and I am still quite puzzled.
Could you please only run the failing test `RePartitionRBFLocal2D2` and output debug logging?
```bash
mpiexec -np 4 ../build/debug/testprecice --run_test="*/*/RePartitionRBFLocal2D2"
```
[...]

PR #894, opened 2020-11-05
Fix out-of-range bug in local RBF ReceivedPartitionTest

uekerman:
A stupid bug in a `ReceivedPartitionTest`. Not defined behavior probably causes part of the problem of #879, could also be the problem of #847.

@oguzziya Please check #879 
@MakisH Please check #847


oguzziya (approved):

oguzziya:
Works on #879, all the tests pass without a problem on macOS 👍

MakisH (commented):
Good catch! But unfortunately, it does not seem to resolve #847.

fsimonis:
We should definitely convert the vector and map access from `c[i]` to `c.at(i)`.
This would turn🦄 in our tests to 💣 .

uekerman:
> We should definitely convert the vector and map access from c[i] to c.at(i).
This would turn :unicorn: in our tests to :bomb: .

I don't understand the :unicorn: reference, but sounds good :smile: 
Let's do that in a separate PR though.

PR #900, opened 2020-11-14
Fix vector access patterns

oguzziya:
This PR changes vector (or array) access patterns in tests to have more robust testing. Changes include
- All the standard library accesses are changed from `[ ]` to `.at()`. Only exception is `std::map` only while constructing the map
- There were some C-style heap-allocated arrays, these are transformed into `std::vector`
- Access for `Eigen`  containers are changed from `[ ]` to `( )` because of two reasons: First, It makes it easier to detect problematic `[ ]` usages; Second, only `Eigen::Vector` supports `[ ]` and in the case of `Eigen::Array` usage becomes consistent. This part is open to discussion actually.

### DIFF
--- a/examples/solverdummies/cpp/solverdummy.cpp
+++ b/examples/solverdummies/cpp/solverdummy.cpp
@@ -54,9 +54,9 @@ int main(int argc, char **argv)
 
   for (int i = 0; i < numberOfVertices; i++) {
     for (int j = 0; j < dimensions; j++) {
-      vertices[j + numberOfVertices * i]  = i;
-      readData[j + numberOfVertices * i]  = i;
-      writeData[j + numberOfVertices * i] = i;
+      vertices.at(j + numberOfVertices * i)  = i;
+      readData.at(j + numberOfVertices * i)  = i;
+      writeData.at(j + numberOfVertices * i) = i;
     }
   }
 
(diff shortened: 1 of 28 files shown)


fsimonis (approved):
There are too many changes to review all of them, but the tests pass and most of the changes look fine.

I am in favour of the eigen3 access change. This makes things more clear and is in line with the matrix access too.

All in all. Good work!

This PR shows that we are doing too much work manually.
We need to gradually migrate parts to [`testing::equals`](https://www.precice.org/doxygen/develop/namespaceprecice_1_1testing.html#a6a116337529290bbb1bf2545b5fb315a) for eigen (and`std::vector<double>`) and `BOOST_CHECK_EQUAL_COLLECTIONS`.

oguzziya:
I skipped `ParallelTests` since there were too many C-like arrays, I think it is related to the ease of use for the interface. Should I change them to `std` containers?

uekerman:
I unassigned me as reviewer. 4 eyes are enough here I guess :grin:

fsimonis:
> I skipped `ParallelTests` since there were too many C-like arrays, I think it is related to the ease of use for the interface. Should I change them to `std` containers?

Let's do this in a separate PR. We should definitely modernize our tests.

Issue #901, opened 2020-11-17
Assertion in QR Factorization triggered in a parallel setup


IshaanDesai:
**Describe your setup**

Operating system (e.g. Linux distribution and version): _Ubuntu 18.04_
preCICE Version: 2.1.1

FEniCS-Adapter commit https://github.com/precice/fenics-adapter/commit/61166ffd27ddb3d9808bc442c4e8d5f02cce75ea is used to reproduce this issue.

The IQN-ILS acceleration scheme triggers the assertion. The coupling scheme configuration is as follows:
```
      <coupling-scheme:serial-implicit>
         <participants first="HeatDirichlet" second="HeatNeumann"/>
         <max-time value="1"/>
         <time-window-size value=".1" valid-digits="8"/>
         <max-iterations value="100"/>
         <exchange data="Flux"        mesh="DirichletNodes" from="HeatDirichlet" to="HeatNeumann" />
         <exchange data="Temperature" mesh="NeumannNodes" from="HeatNeumann"   to="HeatDirichlet" initialize="true"/>
         <relative-convergence-measure data="Flux"        mesh="DirichletNodes" limit="1e-4"/>
         <relative-convergence-measure data="Temperature" mesh="NeumannNodes" limit="1e-4"/>
	 <acceleration:IQN-ILS> 
            <data name="Temperature" mesh="NeumannNodes"/>
            <initial-relaxation value="0.01"/>
            <max-used-iterations value="3"/>
            <time-windows-reused value="1"/>
            <filter type="QR2" limit="1e-3"/>
	 </acceleration:IQN-ILS>
	<acceleration:aitken>
            <data name="Temperature" mesh="NeumannNodes"/>
	    <initial-relaxation value="0.1"/>
	</acceleration:aitken>
      </coupling-scheme:serial-implicit>
```
The assertion is not triggered when Aitken acceleration is used:
```
<acceleration:aitken>
   <data name="Temperature" mesh="NeumannNodes"/>
   <initial-relaxation value="0.1"/>
</acceleration:aitken>
```

**Describe the problem**
In https://github.com/precice/fenics-adapter/commit/61166ffd27ddb3d9808bc442c4e8d5f02cce75ea there is a folder `prototyping_parallel/` which runs the `HT/paritioned-heat/fenics-fenics` tutorial in a parallel manner. When both participants (Dirichlet and Neumann) are run with two processors, the following assertion is triggered in the first time window:
```
ASSERTION FAILED
[...]

PR #905, opened 2020-11-19
Fix QR factorization assertions

oguzziya:
Fix for #901 
- Assertions are removed
- Tests to ParallelTests are added: One solver has all the mesh on the master, while other has all the mesh on the slave.


uekerman (approved):
Looks good. Thanks for adding this :+1: 

It's small, but still worth a changelog entry I guess. In the end it was simply a bug.

docs/changelog/905.md
```diff
@@ -0,0 +1 @@
+Fixed the assertion in QR factorization which does not allow empty partitions.
```
uekerman, line 1:
```suggestion
Fixed two wrong assertions in QR factorization, which did not allow meshes with only a single partition. 
```

Issue #770, opened 2020-06-10
Extend consistent mappings by scaling option


uekerman:
A consistent data mapping can be made "conservative" by a scaling sweep on the output mesh after the mapping. Please see [Bernhard's thesis](https://www5.in.tum.de/pub/Gatzhammer2014_preCICE.pdf), section 4.2.3.

Such an option might be important for conservation of flow rate in fluid-fluid coupling. But there are surely more use cases.

This is issue is a placeholder for the moment. A more detailed description on what this means exactly for NN, NP, and RBF, and existing / missing connectivity is necessary.

fsimonis:
Could we use an action with the `on-exchange-post` timing to achieve this?
Something like `ScaleByValue` would be simple to implement.

oguzziya:
As far as I understood from the thesis, the approach @fsimonis suggested seems applicable to me. In addition, such an action would be useful for different cases too. Is there a limitation from numerical point of view with this approach @uekerman ?

uekerman:
In principle, a solution with a stand-alone action sounds interesting. The technical limitation though is that we do not know the value to scale with in the action. To compute this one we need the source mesh. So, we would need a solution for that.
  
My first thought was to introduce a third mapping type besides "consistent" and "conservative". Or a further mapping attribute "correct-values" or similar. Both suggestions are not perfect either.

oguzziya:
Okay then. I will start working on the idea "third mapping type". Later we can decide to modify it based on the usability I think.

oguzziya:
Discussed with @uekerman. Bernhard's approach includes the construction of a continuous data field using shape functions for both sides of coupling. It allows us to compute the integrals easier even for meshes with no-connectivity. However, implementing the shape function would be too involved. Currently, the plan is:
- Use only mesh with connectivity, since we need the area to calculate the integral value
- Do consistent mapping
- Calculate integral values for input and output mesh, calculate scaling factor
- Correct the mapped values

oguzziya:
In #906, I implemented the core functionality and some tests for this feature but I have several questions:
- What about the distributed meshes? As far as I can see, if the scaling is calculated locally, it is not sure that the integral is conserved for the total mesh. A gather-scatter approach might be too expensive for this feature.
[...]

Issue #542, opened 2019-10-29
Geometric Safety Factor for Bounding Box Filters


fsimonis:
# Problem
After a receiving participant fully broadcast the mesh in the partitioning stage, meshes can be (pre) filtered using bounding boxes to reduce the mesh size used in the mapping-based filtering. These bounding boxes, however, are based on on vertices only.
Filtering out vertices vertices may be relevant for the mapping, thus we enlarge the bounding-boxes by a safety factor. This is useful for vertex-based mappings such as NN and RBF.

Especially for connectivity-based mappings, pure vertex-based bounding-box filtering lead to problems.
Depending on the safety factor, the filter may remove vertices, which in turn remove entire edges and triangles from the mesh.

# Proposed solution
When filtering a mesh using a bounding-box:
1. Select every vertex inside the bounding box
2. :new: Select every vertex connected to the above vertices
3. Build a mesh which consists of all above vertices and all available connectivity information.

# Example
Given the blue bounding box, only the internal nodes would be kept and thus all the connectivity information between the inner and outer vertices are lost.
When extending the inner vertices using the proposed solution, the resulting mesh would contain the vertices covered by the green and blue boxes, thus providing maximum connectivity information.
![issue542](https://user-images.githubusercontent.com/13552216/67786805-c8483500-fa6f-11e9-8274-b48a4f8b8b1c.png)

# Note
As the used mappings dictate the mapping requirements of a mesh, which influences weather there will be no connectivity information or not.

atotoun:
Except those points that BU explained, the text sounds good to me as well.

fsimonis:
I corrected the issue. Shall we remove these comments to avoid confusion in the future?

uekerman:
Yes, I'll delete them.

PR #906, opened 2020-11-20
Scaled consistent mapping implementation

oguzziya:
Implementation according to #770 

Current state: 
- Instead of a new mapping scheme, It seems easier to add an optional configuration option.


oguzziya:
I think this feature is ready to review. Only small point:
- Some distributed mesh tests on PETSc RBF does not pass although I have used the same mesh as Eigen RBF. I could not find an error but more experienced eyes would be better if I miss something or not.

src/mapping/Mapping.cpp
```diff
@@ -78,6 +79,106 @@ int Mapping::getDimensions() const
...
+
+  const auto &inputValues  = input()->data(inputDataID)->values();
+  auto &      outputValues = output()->data(outputDataID)->values();
+
+  int valueDimensions = input()->data(inputDataID)->getDimensions();
+  int meshDimensions  = input()->getDimensions();
+
+  // Integral is calculated on each direction separately
+  std::vector<double> integralInput(valueDimensions);
+  std::vector<double> integralOutput(valueDimensions);
```
fsimonis, line 102: There is a lot of code duplication here.
We could refactor the integration into a separate helper function and move it into `src/mesh/utils.hpp`. This also allows to test this functionality in isolation and reuse it in the future.

Ideally, these lines would look as follows:

```suggestion
  std::vector<double> integralInput = mesh::integrate(inputMesh, inputValues);
  std::vector<double> integralOutput = mesh::integrateOwners(outputMesh, outputValues);
```

src/mapping/Mapping.cpp
```diff
@@ -78,6 +79,106 @@ int Mapping::getDimensions() const
...
+  // Scale in each dimension
+  size_t const outSize = output()->vertices().size();
+  for (int dim = 0; dim < valueDimensions; ++dim) {
+    PRECICE_ASSERT(std::fabs(integralOutput.at(dim)) > 1e-9, "Surface integral is calculated as zero in consistent mapping scaling.");
+    PRECICE_ASSERT(std::fabs(integralInput.at(dim)) > 1e-9, "Surface integral is calculated as zero in consistent mapping scaling.");
+    double scalingFactor = integralInput.at(dim) / integralOutput.at(dim);
+    // Scaling factor is different in each direction, cannot directly multiply
+    for (size_t i = 0; i < outSize; i++) {
+      outputValues((i * valueDimensions) + dim) *= scalingFactor;
+    }
```
fsimonis, line 178: Isn't there a Eigen3-native version of this call?

src/mapping/Mapping.hpp
```diff
@@ -106,6 +109,15 @@ class Mapping {
   /// Method used by partition. Tags vertices that can be filtered out.
   virtual void tagMeshSecondRound() = 0;
 
+  /**
+   * @brief Scales the consistently mapped output data such that the surface integral
+   * of the values on input mesh and output mesh are equal
+   *
+   * Pre-conditions:
+   * - Input and output mesh should have full connectivity information.
```
fsimonis, line 117:
```suggestion
   * @pre Input and output mesh need full connectivity information.
```

src/mesh/Utils.hpp
```diff
@@ -132,5 +153,67 @@ std::array<Eigen::VectorXd, n> coordsFor(const std::array<Vertex *, n> &vertexPt
...
+      int vertex2 = face.vertex(1).getID() * valueDimensions;
+      int vertex3 = face.vertex(2).getID() * valueDimensions;
+      if (face.vertex(0).isOwner() and face.vertex(1).isOwner() and face.vertex(2).isOwner()) {
+        for (int dim = 0; dim < valueDimensions; ++dim) {
+          integral(dim) += (triangleArea(face) / 3.0) * (values(vertex1 + dim) + values(vertex2 + dim) + values(vertex3 + dim));
+        }
+      }
+    }
+  }
+  return std::move(integral);
```
fsimonis, line 215:
```suggestion
  return integral;
```

This is a common antipattern which works against the optimizer.

src/mesh/Utils.hpp
```diff
@@ -132,5 +153,67 @@ std::array<Eigen::VectorXd, n> coordsFor(const std::array<Vertex *, n> &vertexPt
...
+      for (int dim = 0; dim < valueDimensions; ++dim) {
+        integral(dim) += (triangleArea(face) / 3.0) * (values(vertex1 + dim) + values(vertex2 + dim) + values(vertex3 + dim));
+      }
+    }
+  }
+  return std::move(integral);
+}
+
+/// Given the data and the mesh, this function returns the surface integral
+inline Eigen::VectorXd integrateOverlap(PtrMesh mesh, PtrData data)
```
fsimonis, line 186: The name is a bit misleading.
A user can create an overlap by adding a triangle including vertices and edges multiple times.

src/mesh/Utils.hpp
```diff
@@ -44,6 +44,27 @@ inline double edgeLength(const Edge &e)
...
+  // Compute cross-product of vector A and vector B
+  auto area = 0.5 * vectorA.cross(vectorB).norm();
+  // Check if the vectors are co-linear (Pseudo 3D case)
+  if (area < 1e-9) {
+    // Area is the longest edge.
+    std::vector<double> edgeLengths{edgeLength(t.edge(0)), edgeLength(t.edge(1)), edgeLength(t.edge(2))};
+    area = *std::max_element(edgeLengths.begin(), edgeLengths.end());
+  }
+  return area;
+}
```
fsimonis, line 66: I would prefer a `double Triangle::area() const` instead of a non-member function as it is easier to visually digest.

fsimonis:
@oguzziya Can you please add some parallel integration tests to make sure that our assumptions are correct for parallel participants? Also 2 serial tests may be interesting. You can use the nearest-projection mapping for those.

A Ranks | B Ranks | Mapping runs on | Purpose
--- | --- | --- | ---
1 | 1 | A | Mesh is received from B, data from A
1 | 1 | B | Mesh and data are received from A
1 | 2 | B | Correct integral for filtered mesh
2 | 1 | B | Correct integral for joined mesh 

The 2 latter need full connectivity information. Thus an edge needs to be defined by each rank.

src/mapping/NearestNeighborMapping.cpp
```diff
@@ -46,7 +51,7 @@ void NearestNeighborMapping::computeMapping()
   const std::string     baseEvent = "map.nn.computeMapping.From" + input()->getName() + "To" + output()->getName();
   precice::utils::Event e(baseEvent, precice::syncMode);
 
-  if (getConstraint() == CONSISTENT) {
+  if (getConstraint() == CONSISTENT or getConstraint() == SCALEDCONSISTENT) {
```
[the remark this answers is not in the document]
oguzziya: I think only flipping the if and else would be enough.

src/mapping/tests/NearestProjectionMappingTest.cpp
```diff
@@ -213,6 +214,126 @@ BOOST_AUTO_TEST_CASE(ConsistentNonIncremental2D)
...
+
+  outMesh->allocateDataValues();
+  outValues = Eigen::VectorXd::Constant(outData->values().size(), 0.0);
+
+  // Compute and perform mapping
+  mapping.computeMapping();
+  mapping.map(inDataID, outDataID);
+
+  auto   outputIntegral = mesh::integrate(outMesh, outData);
+  double scaleFactor    = outValues(1) / inValues(0);
```
uekerman, line 264: Same as above: please check value if possible.

src/mapping/tests/NearestProjectionMappingTest.cpp
```diff
@@ -213,6 +214,126 @@ BOOST_AUTO_TEST_CASE(ConsistentNonIncremental2D)
...
+
+  outMesh->allocateDataValues();
+
+  //assign(outData->values()) = 0.0;
+  outValues = Eigen::VectorXd::Constant(outData->values().size(), 0.0);
+  mapping.computeMapping();
+  mapping.map(inDataID, outDataID);
+
+  auto   outputIntegral = mesh::integrate(outMesh, outData);
+  double scaleFactor    = outValues(0) / inValues(0);
```
uekerman, line 325: Same as above: please check value if possible.

src/mapping/tests/PetRadialBasisFctMappingTest.cpp
```diff
@@ -997,15 +1006,300 @@ BOOST_AUTO_TEST_CASE(testDistributedConservative2DV5Vector)
...
+                                     {0, -1, {0, 1}, {0, 0}},
+                                     {1, -1, {1, 0}, {0, 0}},
+                                     {1, -1, {1, 1}, {0, 0}},
+                                     {2, -1, {2, 0}, {0, 0}},
+                                     {2, -1, {2, 1}, {0, 0}},
+                                     {3, -1, {4.1, 0}, {0, 0}},
+                                     {3, -1, {4.2, 1}, {0, 0}}},
+                                    {{{0, 1}, -1}, {{0, 1}, -1}, {{0, 1}, -1}, {{0, 1}, -1}})
+                                    );
+}*/
```
uekerman, line 1067: Does this test still fail?

src/mapping/tests/PetRadialBasisFctMappingTest.cpp
```diff
@@ -997,15 +1006,300 @@ BOOST_AUTO_TEST_CASE(testDistributedConservative2DV5Vector)
...
+                                     {0, -1, {0, 1}, {0, 0}},
+                                     {0, -1, {1, 0}, {0, 0}},
+                                     {2, -1, {1, 1}, {0, 0}},
+                                     {2, -1, {2, 0}, {0, 0}},
+                                     {2, -1, {2, 1}, {0, 0}},
+                                     {3, -1, {3, 0}, {0, 0}},
+                                     {3, -1, {3, 1}, {0, 0}}},
+                                    {{{0, 1}, 0}, {{0, 2}, 0}, {{0, 1}, 2}, {{1, 2}, 2}, {{0, 1}, 3}}));
+}
+*/
```
uekerman, line 1190: Do these tests still fail?

src/mesh/Utils.cpp
```diff
@@ -0,0 +1,78 @@
...
+  const int       meshDimensions  = mesh->getDimensions();
+  const auto &    values          = data->values();
+  Eigen::VectorXd integral        = Eigen::VectorXd::Zero(valueDimensions);
+
+  if (meshDimensions == 2) {
+    for (const auto &edge : mesh->edges()) {
+      int vertex1 = edge.vertex(0).getID() * valueDimensions;
+      int vertex2 = edge.vertex(1).getID() * valueDimensions;
+
+      if (edge.vertex(0).isOwner() and edge.vertex(1).isOwner()) {
```
uekerman, line 52: I am afraid this way we lose edges where both vertex owners are different. What we could do is to only count half of the value for only one owner.

src/mesh/Utils.cpp
```diff
@@ -0,0 +1,78 @@
...
+      }
+    }
+  } else {
+    for (const auto &face : mesh->triangles()) {
+      int vertex1 = face.vertex(0).getID() * valueDimensions;
+      int vertex2 = face.vertex(1).getID() * valueDimensions;
+      int vertex3 = face.vertex(2).getID() * valueDimensions;
+
+      // If the rank owns at least two of the vertices of the triangle, calculate the integral
+      if ((face.vertex(0).isOwner() + face.vertex(1).isOwner() + face.vertex(2).isOwner()) > 1) {
```
uekerman, line 65: All three vertices could have different owners. Similar as above we could take 1/3 per owned vertex.

src/precice/tests/ParallelTests.cpp
```diff
@@ -1104,6 +1105,936 @@ BOOST_AUTO_TEST_CASE(UserDefinedMPICommunicatorPetRBF)
...
+    BOOST_TEST(expectedIntegral == calculatedIntegral);
+
+    cplInterface.advance(maxDt);
+    BOOST_TEST(!cplInterface.isCouplingOngoing(), "Sending participant should have to advance once!");
+    cplInterface.finalize();
+  }
+}
+
+// Simple 2 triangles. Read mapping on parallel participant B.
+// This test does not pass. When the from-mesh is repartitioned, ranks does not own all the vertices in the corresponding triangles.
```
uekerman, line 1225: But the CI passes?

src/precice/tests/ParallelTests.cpp
```diff
@@ -1104,6 +1105,936 @@ BOOST_AUTO_TEST_CASE(UserDefinedMPICommunicatorPetRBF)
...
+
+      cplInterface.advance(maxDt);
+      BOOST_TEST(!cplInterface.isCouplingOngoing(), "Receiving participant should have to advance once!");
+      cplInterface.finalize();
+    }
+    BOOST_TEST(calculatedIntegral == expectedIntegral);
+  }
+}
+
+// Diamond shped 2 triangles. Read mapping on serial participant B
```
uekerman, line 1347:
```suggestion
// Diamond shaped 2 triangles. Read mapping on serial participant B
```

src/precice/tests/ParallelTests.cpp
```diff
@@ -1104,6 +1105,936 @@ BOOST_AUTO_TEST_CASE(UserDefinedMPICommunicatorPetRBF)
...
+      utils::MasterSlave::allreduceSum(localCalculatedIntegral, calculatedIntegral, 1);
+
+      cplInterface.advance(maxDt);
+      BOOST_TEST(!cplInterface.isCouplingOngoing(), "Receiving participant should have to advance once!");
+      cplInterface.finalize();
+    }
+    BOOST_TEST(calculatedIntegral == expectedIntegral);
+  }
+}
+*/
```
uekerman, line 2036: What's the state of this test?

uekerman (changes requested):
Sry for my really slow reaction.
Implementation / coding style looks really good, and thanks for the already extensive tests! :+1:

There are, unfortunately, still some conceptual problems:
* `integrateOwner`: see below
* Most tests currently rely on the fact that `integrate` gives the correct result. It's used in the tested implementation and in the test. So, we cannot capture problems in `integrate` itself. I guess it would be good to add (simple) unit tests for `mesh/Utils` where we compute the expected results on paper.
* We need to find out why the PetRBF tests fail.

src/mesh/Utils.cpp
oguzziya: I give a little bit of thought to this idea but I think the problem would still persist. Lets' have an example (it is taken from one of the tests), the values in parenthesis are the ownership information.

Rank 0 defines a triangle with vertices: 0 1 2 (1, 1, 0)
Rank 1 defines a triangle with vertices: 0 1 2 (0, 0, 1)
Rank 1 defines a triangle with vertices: 3 0 2 (1, 0, 1)

If we use a ratio based on the ownership of the vertices, it will calculate the integral correctly for the (0 1 2) triangle, but not for the (3 0 2) triangle, since we multiply it with 2/3 which is wrong. I think the fundamental reason why this is complicated is that we not know how many ranks define a triangle. Maybe the ownership information for edge and triangle would also be useful, I have no idea whether it can be implemented or not though. What do you think about this?

src/mapping/tests/NearestNeighborMappingTest.cpp
```diff
@@ -166,5 +167,73 @@ BOOST_AUTO_TEST_CASE(ConservativeNonIncremental)
...
+
+  outMesh->allocateDataValues();
+
+  // Setup mapping with mapping coordinates and geometry used
+  precice::mapping::NearestNeighborMapping mapping(mapping::Mapping::SCALEDCONSISTENT, dimensions);
+
+  mapping.setMeshes(inMesh, outMesh);
+  BOOST_TEST(mapping.hasComputedMapping() == false);
+
+  // Map data with coinciding vertices, has to result in equal values.
```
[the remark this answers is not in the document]
oguzziya: Yes, a copy-paste mistake 😄

src/precice/tests/ParallelTests.cpp
oguzziya: The description was not updated, sorry about that.

src/precice/tests/ParallelTests.cpp
oguzziya: Unfortunately, it hangs and results in a timeout since there is missing connectivity information for the received partition. I am not sure whether my setup is correct: Each participant runs on 2 ranks and the mesh is partitioned. Mapping is read mapping.

src/mapping/tests/PetRadialBasisFctMappingTest.cpp
oguzziya: Unfortunately, yes. I cannot figure out why they fail. Setup is the same as the Eigen RBF one.

src/mesh/Utils.cpp
uekerman: Yes, you are completely right. Would have been too easy to be true. Yes, I agree that for a proper solution we would need sth like ownership for edges and triangles. That's not impossible to do. I guess a gather-scatter solution in the following function should not be too complicated to do:
https://github.com/precice/precice/blob/00589fe951f9715174eafbd66657294e0d8eb8e4/src/partition/ReceivedPartition.cpp#L468

src/precice/tests/ParallelTests.cpp
fsimonis: I'll have a look at the test

fsimonis: This is the test output when running in isolation: (tidied up)

```
(0) 11:44:13 [partition::ReceivedPartition]:609 in createOwnerInformation:
WARNING: 1 of 8 vertices of mesh MeshOne have been filtered out since they have no influence on the mapping.

(1) 11:44:13 [mapping::NearestProjectionMapping]:160 in computeMapping:
WARNING: 3D Mesh "MeshOne" does not contain triangles. Nearest projection mapping will map to primitives of lower dimension.

(1) 11:44:13 [mapping::Mapping]:89 in scaleConsistentMapping:
ERROR: Connectivity information is missing for the mesh MeshOne. Scaled consistent mapping requires connectivity information.
```

fsimonis: The issue seems to be that when running this in parallel, the SolverTwo rank 1 has full overlap with the single triangle of SolverOne rank 1. However, it only shares a single edge with SolverOne rank 0.

During the repartitioning, SolverTwo rank 1 has to compute the mapping for the mesh, received by SolverOne rank 0, which does not contain connectivity information (as expected). This leads to the error.

uekerman: So, too much is filtered out, correct? Sounds like we would need to adjust `tagMeshFirstRound` or `tagMeshSecondRound` of PetRBF.

oguzziya: I naively tried to implement extending filtering function such that the connected primitive information won't get lost according to https://github.com/precice/precice/issues/542, however, I did not work on this too much. I can implement this too but I think this PR getting bigger and bigger. Maybe I can create a separate PRs for extending Bounding Box and handling of global indexing of Triangles and Edges?

fsimonis: To avoid further confusion, the mentioned test uses nearest-neighbour, not PetRBF.

fsimonis: Maybe the duplicate edge leads to problems.
If the nearest-projection mapping chooses to project on a vertex of (SolverOne 0) instead of (SolverOne 1) due to the vertex placement, then it may select as an example 1 vertices from  (SolverOne 0), 2 vertex from (SolverOne 1).
In that case, the triangle is removed.

fsimonis:
@uekerman and I came to the conclusion that there are too many open questions regarding the parallel case.
Our current aggressive filtering approach will eventually lead to missing triangles and thus incomplete surfaces.
This could be solved with a more conservative tagging approach (e.g. tag complete BB), which in turn will make duplicate surfaces a nightmare.

This PR is functional and useful for the serial case (read mapping on serial solver).
To prevent this PR from diverging too much from develop, we propose the following:
* Let's reduce this PR to the serial implementation and tests. We further need to add a check and a descriptive error message for the unsupported case. (similar to the general read-consistent write-conservative constraints for parallel cases)
* Then move the parallel implementation and tests to a separate branch and open a PR and take some time to find a solution to the problem

@oguzziya let's get your work into develop!

oguzziya:
All the parallel implementation details and tests are removed. Right now it is fully functional for serial cases.

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -413,6 +413,8 @@ void ParticipantConfiguration::finishParticipantConfiguration(
         PRECICE_ERROR(
             "For a parallel participant, only the mapping"
             << " combinations read-consistent and write-conservative are allowed");
+      } else if (confMapping.mapping->getConstraint() == mapping::Mapping::SCALEDCONSISTENT) {
+        PRECICE_ERROR("Scaled consistent mapping is not supported for a parallel participant. Instead use write-conservative for a conservative mapping");
```
uekerman, line 417:
```suggestion
        PRECICE_ERROR("Scaled consistent mapping is not yet supported for a parallel participant. You could run in serial or use a plain (read-)consistent mapping instead.");
```

uekerman (approved):
I have not checked all details, but looks good. 
Only some final tweaks to the error message below. In my opinion a conservative mapping is no real alternative to the scaled-consistent one as it operates on different data. One would need to convert e.g. from stress to force to go from scaled-consistent to conservative. Just going to plain consistent is the easier alternative, despite its limitation concerning conservation of values.

oguzziya:
@fsimonis currently pr is ready to merge 👍

PR #910, opened 2020-12-01
Add docker release

fsimonis:
This PR adds a docker release workflow.

Once a release is published, this action will run and build a dockerimage installing the uploaded debian package of the release.
It will then push the result to `precice/precice:vX.Y.Z` and `precice/precice:latest`.

### DIFF
--- a/.github/workflows/release-docker.yml
+++ b/.github/workflows/release-docker.yml
@@ -1,20 +1,26 @@
 name: "Release Dockerfile"
 on:
   release:
-    types: [published, released]
+    types: released
 jobs:
   latest:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2
-      - uses: docker/build-push-action@v1
-        env:
-          version: ${{ github.events.release.tag_name }}
+      - name: Set up Docker Buildx
+        uses: docker/setup-buildx-action@v1
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
         with:
-          username: ${{ secrets.DOCKERHUB_USER }}
-          password: ${{ secrets.DOCKERHUB_PASS }}
-          dockerfile: "./tools/releasing/packaging/docker/Dockerfile"
-          repository: precice/precice
-          tags: ${{ env.version }},latest
+          username: precice
+          password: ${{ secrets.DOCKERHUB_TOKEN }}
+      - name: Build and push Dockerfile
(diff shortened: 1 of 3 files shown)


BenjaminRueth:
I know that I'm a bit late for the party, but some questions:

* Why do we not use `secrets.DOCKERHUB_USERNAME` ?
* Why not push a `develop` version to dockerhub for every new commit on develop? Might be handy for testing. See https://github.com/BenjaminRueth/precice/blob/develop/.github/workflows/release-docker.yml

fsimonis:
@BenjaminRueth 
> Why do we not use secrets.DOCKERHUB_USERNAME

I attempted this, but the GitHub actions when censored `precice` everywhere in the pipeline, making the output essentially unusable.

> Why not push a develop version to dockerhub for every new commit on develop?

Good idea. This needs to be implemented differently though as the action needs to somehow fetch the debian package of the build. This could possibly be done directly in the build-and-test pipeline.

BenjaminRueth:
> @BenjaminRueth
> 
> > Why do we not use secrets.DOCKERHUB_USERNAME
> 
> I attempted this, but the GitHub actions when censored `precice` everywhere in the pipeline, making the output essentially unusable.

I understand. The problem I encountered when reviewing the pipeline on my fork was that I had to manually change the correct occurences of `precice` to my username (or `DOCKERHUB_USERNAME`).

Is there maybe a way to use an alias name or some other way to authenticate on docker? Or can we avoid censoring for certain secrets that are not really secrets? Maybe a github secret is the right approach for the password, but not for the username?

BenjaminRueth:
> Is there maybe a way to use an alias name or some other way to authenticate on docker? Or can we avoid censoring for certain secrets that are not really secrets? Maybe a github secret is the right approach for the password, but not for the username?

I did some research and found out that one can get the [`github.repository_owner`](https://docs.github.com/en/free-pro-team@latest/actions/reference/context-and-expression-syntax-for-github-actions#github-context). From that we can define the environment variable `DOCKERHUB_USERNAME` in the workflow. Then it will not be censored and we still have a reasonable workflow for users that use the same name on dockerhub and github (I think this is a reasonable assumption). If one uses different names, one can easily customize how `DOCKERHUB_USERNAME` is set within the workflow.

fsimonis:
Answer in https://github.com/precice/precice/pull/924#issuecomment-759414312

Issue #280, opened 2019-02-14
Exception Handling of preCICE errors


fsimonis:
# Problem

In case of a User-error, preCICE calls `ERROR` which calls `std::exit`:
https://github.com/precice/precice/blob/aeae01676fccff9faeb192f35dba410df0cf8d2b/src/logging/LogMacros.hpp#L33-L38

Due to this, we currently have no way of testing user `ERROR`.

# Proposed Solution

It would be cleaner to throw an exception, which we could check via `BOOST_CHECK_EXCEPTION` [here](https://www.boost.org/doc/libs/1_69_0/libs/test/doc/html/boost_test/utf_reference/testing_tool_ref/assertion_boost_level_exception.html).

We could add a hierarchy for preCICE:
```cpp
class preciceError : std::runtime_error {};

class configurationError : preciceError {};
class communicationError : preciceError {};
class mappingError : preciceError {};
class interfaceError : preciceError {};
```

Depending on user settings (xSDK R3), we could transform those to error codes on the solver interface.

# References
* [Cpp Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#S-errors)
* [STL Error Handling](https://en.cppreference.com/w/cpp/error)
* [Boost Guidelines](https://www.boost.org/community/error_handling.html)

uekerman:
I would vote for
```c++
class configurationError : preciceError {};
class communicationError : preciceError {};
class numericalError : preciceError {}; // e.g. when RBF does not converge or QN has rank 0
class apiError : preciceError {}; // e.g. wrong usage of the API, e.g. calling advance before initialize
class geometryError : preciceError {}; // e.g. interfaces meshes do not match geometrically
```

floli:
Using exceptions generally is a good idea, but should be done with caution. The C++ exception system tends to behave unexpected.

* Use the term `exception` not `error`
* Derive from `std::exception` or `std::runtime_error`
* Use `precice::Error` not `preciceError`. That's what namespaces are for.
* Throwing an exception is dangerous when another exception is already in flight.

fsimonis:
@floli 
> Use the term `exception` not `error`

True, in our case they really are exceptions, not an errors.

> Derive from `std::exception` or `std::runtime_error`

✔️ 

> Use `precice::Error` not `preciceError`. That's what namespaces are for.

[...]

PR #916, opened 2020-12-08
Add check for advance timestep size

### LINKED ISSUE #915: Nice-to-have: error when the timestep size is zero
In `advance()` we are providing an argument for the timestep size:

https://github.com/precice/precice/blob/908cf7ae1a7e8d4a5bfa9e0088d8cf8f3ca3fd36/src/precice/SolverInterface.hpp#L145

One can by mistake provide a zero timestep (for multiple timesteps). In that case, it would be good if preCICE throws an error, detecting that the timestep size is smaller than our precision.

fsimonis:

This PR adds checks for the timestep size to enforce the following:
* It cannot be approximately zero.
* It cannot be negative.

**Additional Information**

Closes #915

### DIFF
--- a/src/math/differences.hpp
+++ b/src/math/differences.hpp
@@ -17,12 +17,6 @@ constexpr bool equals(const Eigen::MatrixBase<DerivedA> &A,
   return A.isApprox(B, tolerance);
 }
 
-/// Compares two doubles for equality up to tolerance
-constexpr inline bool equals(double a, double b, double tolerance = NUMERICAL_ZERO_DIFFERENCE)
-{
-  return std::abs(a - b) < tolerance;
-}
-
 /// Compares two scalar (arithmetic) types
 template <class Scalar>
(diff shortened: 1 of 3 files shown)


MakisH (commented):
Do we maybe also need a test here?

Other than that, the code looks nice and if the tests pass, the functionality should not change (I have not tried to run it, but it should be doing what I want).

Thank you for the very fast support - a happy customer. :smile:

src/math/differences.hpp
```diff
@@ -17,6 +17,12 @@ constexpr bool equals(const Eigen::MatrixBase<DerivedA> &A,
   return A.isApprox(B, tolerance);
 }
 
+/// Compares two doubles for equality up to tolerance
+constexpr inline bool equals(double a, double b, double tolerance = NUMERICAL_ZERO_DIFFERENCE)
+{
+  return std::abs(a - b) < tolerance;
+}
+
```
uekerman, line 25: Why do we need this additional function? Doesn't the function just below here do the job?

fsimonis: You are right! Thanks for pointing that out!

fsimonis:
@MakisH This leads us to the usual problem for not being able to test error cases. #280

uekerman (approved):

PR #917, opened 2020-12-09
Disable KSP_DIVERGED_DTOL for very small RHS in PETSc KSPLSQR

### LINKED ISSUE #904: GMRES cannot deal with consistent RBF mapping using constant input data
**Describe your setup**

Only happens sometimes, for example in the docker container provided here: https://github.com/dealii/code-gallery/pull/84#issuecomment-723416669

**Describe the problem**

For constant input data, a consistent RBF mapping with separated polynomial can lead to an apparently ill-conditioned (?) system, GMRES cannot cope with. 

A * x = b

with b very small, around machine precision, but not 0 yet. 

<details>
  <summary>PETSc output</summary>
  
  ```bash
PETSc GMRES:

KSP Object: Coefficient Solver (solverC_) 1 MPI processes
  type: gmres
    restart=30, using Classical (unmodified) Gram-Schmidt Orthogonalization with no iterative refinement
    happy breakdown tolerance 1e-30
  maximum iterations=10000, nonzero initial guess
  tolerances:  relative=1e-09, absolute=1e-50, divergence=10000.
  left preconditioning
  using PRECONDITIONED norm type for convergence test
PC Object: (solverC_) 1 MPI processes
  type: icc
    out-of-place factorization
    0 levels of fill
    tolerance for zero pivot 2.22045e-14
    using Manteuffel shift [POSITIVE_DEFINITE]
    matrix ordering: natural
    factor fill ratio given 1., needed 1.
      Factored matrix follows:
        Mat Object: 1 MPI processes
          type: seqsbaij
          rows=6, cols=6
          package used to perform factorization: petsc
          total: nonzeros=21, allocated nonzeros=21
          total number of mallocs used during MatSetValues calls=0
              block size is 1
  linear system matrix = precond matrix:
  Mat Object: C 1 MPI processes
    type: seqsbaij
    rows=6, cols=6
    total: nonzeros=21, allocated nonzeros=21
    total number of mallocs used during MatSetValues calls=0
        block size is 1
---[precice] ERROR:  The linear system of the RBF mapping from mesh boundary-mesh to mesh dealii-mesh has not converged. This means most probably that the mapping problem is not well-posed. Please check if your coupling meshes are correct. Maybe you need to fix axis-aligned mapping setups by marking perpendicular axes as dead?
  ```
</details>

**Solution**

Unclear yet. We need to better understand the problem first. Is it the preconditioner? Playing around with the PETSc settings could shed some light on this.

DavidSCN:
Resolves #904.

The problem and a workaround are described [on the PETSc mailing list](https://lists.mcs.anl.gov/pipermail/petsc-users/2017-March/032000.html). Using this workaround resolves the (original) problem and all tests pass. Since the maximum iteration option is still active, we cannot really do anything wrong when we change this option. However, I would like to have a look at the divergence residual for our original problem case before merging this. 

A general remark: applying the clang-format to the modified file changes some lines in my case and removes some whitespace errors.

### DIFF
--- a/src/mapping/PetRadialBasisFctMapping.hpp
+++ b/src/mapping/PetRadialBasisFctMapping.hpp
@@ -561,10 +561,11 @@ void PetRadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::computeMapping()
   // considered as diverged and the computation is aborted. The PETSC_DEFAULT value is 1e9. However, the
   // divergence residual is scaled using the RHS b of the system. In case the RHS is (still nonzero) very small
   // and we have a (bad) nonzero initial guess x as defined below, the divergence tolerance might be exceeded
-  // in the first iteration, although the system could be solved in the usual way. Therefore, we select a very
-  // high value (1e30) in order to disable the divergence check. In practice, the check is very rarely needed
-  // with Krylov methods. According to the PETSc people, even the rare use cases aim for a bad preconditioner,
-  // which is not used in our configuration. Hence, we can disable the divergence check without concerns.
+  // in the first iteration, although the system could be solved in the usual way. This behavior is essentially
+  // a glitch in the divergence check. Therefore, we select a very high value (1e30) in order to disable the
+  // divergence check. In practice, the check is very rarely needed with Krylov methods. According to the PETSc
+  // people, the rare use cases aim for a bad preconditioner, which is not even used in our configuration.
+  // Hence, we can disable the divergence check without concerns.
   KSPSetTolerances(_solver, _solverRtol, PETSC_DEFAULT, 1e30, PETSC_DEFAULT);
   KSPSetInitialGuessNonzero(_solver, PETSC_TRUE);
   CHKERRV(ierr);                            // Reuse the results from the last iteration, held in the out vector.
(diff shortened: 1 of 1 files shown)


DavidSCN:
Running the original (code-gallery) test case with the `ksp_monitor_true_residual` flag/ this code snippet 
```c++
  PetscInt it = 0;
  PetscReal norm=0;
  PetscViewerAndFormat *vf;
  ierr = PetscViewerAndFormatCreate(PETSC_VIEWER_STDOUT_WORLD, PETSC_VIEWER_DEFAULT, &vf);
  CHKERRQ(ierr);
  KSPSolve(ksp, b, x);
  KSPMonitorTrueResidualNorm(ksp, it, norm, vf);
  CHKERRQ(ierr);
```
prints 
```bash
  0 KSP preconditioned resid norm 0.000000000000e+00 true resid norm 6.110100926608e-01 ||r(i)||/||b|| 5.497994403102e+14
```
Hence, the mailing list problem is exactly reproduced.

uekerman:
Could you please add a brief comment in the code explaining what the argument is about and why choose it like this?

The code gallery case now works with this change? (and the original mapping setting)

DavidSCN:
> Could you please add a brief comment in the code explaining what the argument is about and why choose it like this?

I decided for a more comprehensive explanation. Otherwise, probably no one will understand without looking deeper into this topic.

> The code gallery case now works with this change? (and the original mapping setting)

Yes.

PR #920, opened 2020-12-14
Bump C++ version to c++14

fsimonis:

This PR bumps the required version of C++ from 11 to 14.

**Motivation**

Our dependency `Boost.Geomerty` requires C++14 since Boost `1.75`.


MakisH (approved):
This is reasonable, especially since all platforms we are targetting support C++14. We have also decided this since a while, but not implemented it since we did not have a clear reason to (as this one).

I have not tried to configure, but the code looks ok.

Issue #633, opened 2020-01-28
Enable two-level initialization by default


uekerman:
There are still a few details that need to be sorted out before we can enable two-level intitialization by default.

* [ ] `ReceivedPartition::createOwnerInformation` still applies a gather-scatter approach and is therefore a potential bottleneck
* [ ] Move exchange of `communicationMap` out of `Partition::compute()`. Probably to `m2n::completeSlavesConnection()`. Otherwise, we might get blocking sends as `MeshContexts` are in wrong order when calling `partition::compute`.
* [ ] Merge `accept/requestSlaves` and `completeSlavesConnection` to a single entry point
* [ ] Extend two-level initialization to multiple m2ns in `ProvidedPartition`
* [ ] Support coupling mode (i.e. a serial participant) with two-level initialization or automatically switch
* [ ] Support gather-scatter m2n with two-level initialization or automatically switch
* [ ] Get rid of vertex distribution and use only local information (`connectedRanks` or `communicationMaping`) in `m2n`. Reduce code duplication in `PointToPointCommunication`. 

Additionally, we need a better test coverage.

BenjaminRodenberg:
What's the status here? I guess it would be nice to use two-level initialization as the default option and delete the alternative implementation with the breaking release, because it should simplify the code. But I have no idea whether this is actually realistic. The todo list above looks like it is not.

uekerman:
No, that's unfortunately unrealistic. I guess that most of the points above are still open.
Independent of that, I would not delete the old gather-scatter mode as it is helpful for debugging.
I see this issue also not as breaking. It will only be faster :)

PR #921, opened 2020-12-14
Parallelize createOwnerInformation

atotoun:
`ReceivedPartition::createOwnerInformation`: the function is parallelized for bounding-box initialization according to [the issue 633](https://github.com/precice/precice/issues/633)


src/partition/ReceivedPartition.cpp
```diff
@@ -124,6 +126,14 @@ void ReceivedPartition::compute()
   if (_toMapping)
     _toMapping->tagMeshFirstRound();
 
+  // if(utils::MasterSlave::getRank()==0)
+  // {
+  //   for (auto & vtx : _mesh->vertices())
+  //   {
+  //     std::cout<< vtx.getGlobalIndex() << std::endl;
+  //   }
+  // }
```
uekerman, line 135: Let's not push this to develop. If you think this is helpful in the long run you could make it a logging output and potentially guard with `#ifndef NDEBUG`.

src/partition/ReceivedPartition.cpp
```diff
@@ -161,6 +171,14 @@ void ReceivedPartition::compute()
     // A vertex belongs to a specific connected rank if its global vertex ID lies within the ranks min and max.
     std::map<int, std::vector<int>> remoteCommunicationMap;
 
+    // if(utils::MasterSlave::getRank()==0)
+    // {
+    //   for (auto & vtx : _mesh->vertices())
+    //   {
+    //     std::cout<< vtx.getGlobalIndex() << std::endl;
+    //   }
+    // }
```
uekerman, line 180: Let's not push this to develop. If you think this is helpful in the long run you could make it a logging output and potentially guard with `#ifndef NDEBUG`.

src/partition/ReceivedPartition.cpp
```diff
@@ -470,143 +488,312 @@ void ReceivedPartition::createOwnerInformation()
...
+    1- receive local bb map from master
+    2- filter bb map to keep the connected ranks
+    3- own the vertices that only fits into the ranks bb
+    4- send number of owned vertices and shared vertices list to neighbors
+    5- for the remaining vertices: check if we have less vertexes -> own it!
+    */
+
+    // #1: receive local bb map from master
+    // Define and initialize localBBMap to save local bbs
+    prepareBoundingBox();
```
uekerman, line 502: Shouldn't that already happen in `compareBoundingBoxes` ?

src/partition/ReceivedPartition.cpp
```diff
@@ -470,143 +488,312 @@ void ReceivedPartition::createOwnerInformation()
...
+    
+    // store possible shared vertices in a map to communicate with neighbors, map: rank -> vertex_global_id
+    std::map<int, std::vector<double>> sharedVerticesSendMap;
+
+    // receive list of possible shared vertices from neighboring ranks
+    std::map<int, std::vector<double>> sharedVerticesReceiveMap;
+
+    if (utils::MasterSlave::isMaster()) {
+
+      // master receives local bb from each slave ranke
```
uekerman, line 527:
```suggestion
      // master receives local bb from each slave rank
```

src/partition/ReceivedPartition.cpp
```diff
@@ -470,143 +488,312 @@ void ReceivedPartition::createOwnerInformation()
...
+      for (int rankSlave = 1; rankSlave < utils::MasterSlave::getSize(); rankSlave++) {
+        com::CommunicateBoundingBox(utils::MasterSlave::_communication).receiveBoundingBox(localBBMap.at(rankSlave), rankSlave);
+      }
+
+      // master broadcast localBBMap to all slaves 
+      for (int rankSlave = 1; rankSlave < utils::MasterSlave::getSize(); rankSlave++) {
+        com::CommunicateBoundingBox(utils::MasterSlave::_communication).sendBoundingBoxMap(localBBMap, rankSlave);
+      }
```
uekerman, line 535: It would performance-wise be better to use a proper broadcast here.

src/partition/ReceivedPartition.cpp
```diff
@@ -470,143 +488,312 @@ void ReceivedPartition::createOwnerInformation()
...
 
-  if (utils::MasterSlave::isSlave()) {
+  if (m2n().usesTwoLevelInitialization()) {
+    /*
+    1- receive local bb map from master
+    2- filter bb map to keep the connected ranks
+    3- own the vertices that only fits into the ranks bb
+    4- send number of owned vertices and shared vertices list to neighbors
+    5- for the remaining vertices: check if we have less vertexes -> own it!
+    */
```
uekerman, line 498: Could you please additionally add a few lines of big-picture explanation here. What is done and why.
The steps you have are already very helpful, but still require the reader to understand how you got there.

src/partition/ReceivedPartition.cpp
```diff
@@ -470,143 +488,312 @@ void ReceivedPartition::createOwnerInformation()
...
+    }
+    else if (utils::MasterSlave::isSlave()) {
+      // slaves send local bb to master 
+      com::CommunicateBoundingBox(utils::MasterSlave::_communication).sendBoundingBox(_bb, 0);
+      // slaves receive localBBMap from master
+      com::CommunicateBoundingBox(utils::MasterSlave::_communication).receiveBoundingBoxMap(localBBMap, 0);
+    }    
+
+    // #2: filter bb map to keep the connected ranks
+    // remove the current bb from the map
```
uekerman, line 545:
```suggestion
    // remove the own bb from the map
```
please add an explanation why we do this

src/partition/ReceivedPartition.cpp
```diff
@@ -470,143 +488,312 @@ void ReceivedPartition::createOwnerInformation()
...
-        slaveTags[0][i]   = 1;
-      } else {
-        slaveTags[0][i] = 0;
-      }
+    // #4: Exchange number of already owned vertices with the neighbors
+    std::map<int, int> neighborRanksVertexCount; // for load balancing
+    int receivedCounter = 0;
+    for (auto & neighborRank: localConnectedBBMap)
+    {
+      utils::MasterSlave::_communication->aSend(counter, neighborRank.first);
```
uekerman, line 594: I don't like the `aSend` here a lot. I guess it could still lead to a deadlock for very large cases depending on the MPI implementation.
Better could be:
* first `aReceive`
* then `send`
* last, check that all receives did happen

src/partition/ReceivedPartition.cpp
```diff
@@ -470,143 +488,312 @@ void ReceivedPartition::createOwnerInformation()
...
+    // remove the current bb from the map
+    localBBMap.erase(utils::MasterSlave::getRank());
+    // find and store local connected ranks
+    for (auto &localBB : localBBMap) {
+      if (_bb.overlapping(localBB.second)) {
+        localConnectedBBMap.emplace(localBB.first, localBB.second);
+      }
+    }
+
+    // #3: filter vertices and keep those only fit into the current rank's bb
```
uekerman, line 554: Let's not use the word filter here. In our context, filter means that we delete vertices, which we do not do here.

src/partition/ReceivedPartition.cpp
```diff
@@ -470,143 +488,312 @@ void ReceivedPartition::createOwnerInformation()
...
+      {
+        std::vector<double> vec = sharingRank.second;
+        if( std::find(vec.begin(), vec.end(), sharedVerticesList[i]) != vec.end() )
+        {
+          if((counter > neighborRanksVertexCount[sharingRank.first]) ||
+             (counter == neighborRanksVertexCount[sharingRank.first] && utils::MasterSlave::getRank() > sharingRank.first) )
+          {
+            owned = false;
+          }
         }
```
uekerman, line 644: Could you please add some documentation on how you do the load balancing?
Why the `utils::MasterSlave::getRank() > sharingRank.first`?

src/partition/ReceivedPartition.cpp
```diff
@@ -470,143 +488,312 @@ void ReceivedPartition::createOwnerInformation()
...
-    for (int rank = 0; rank < utils::MasterSlave::getSize(); rank++) {
-      for (size_t i = 0; i < slaveOwnerVecs[rank].size(); i++) {
-        if (globalOwnerVec[slaveGlobalIDs[rank][i]] == 0 && slaveTags[rank][i] == 1) {
-          slaveOwnerVecs[rank][i]                 = 1;
-          globalOwnerVec[slaveGlobalIDs[rank][i]] = rank + 1;
+    setOwnerInformation(tags);
+    // end of bb section 
+  } else
+  {
+    std::cout<<"we use normal filtering ..." << std::endl;
```
uekerman, line 659: Please use Debug log

src/partition/ReceivedPartition.cpp
```diff
@@ -470,143 +488,312 @@ void ReceivedPartition::createOwnerInformation()
...
-    if (filteredVertices)
-      PRECICE_WARN(filteredVertices << " of " << _mesh->getGlobalNumberOfVertices()
-                                    << " vertices of mesh " << _mesh->getName() << " have been filtered out "
-                                    << "since they have no influence on the mapping.");
+      auto filteredVertices = std::count(globalOwnerVec.begin(), globalOwnerVec.end(), 0);
+      if (filteredVertices)
+        PRECICE_WARN(filteredVertices << " of " << _mesh->getGlobalNumberOfVertices()
+                     << " vertices of mesh " << _mesh->getName() << " have been filtered out "
+                     << "since they have no influence on the mapping.");
+    }
```
uekerman, line 796: I guess those could be helpful for the two-level variant as well.

uekerman (commented):
I think this goes in the right direction :+1: Maybe some things could still be simplified.
I have not yet checked all details and have not yet looked at the tests.

src/partition/ReceivedPartition.cpp
atotoun: That's right.

src/partition/ReceivedPartition.cpp
atotoun: True, I replaced it.

src/partition/ReceivedPartition.cpp
atotoun: Explanation is added. Please check for correctness and completeness.

src/partition/ReceivedPartition.cpp
atotoun: I don't understand how this could result in a deadlock. Maybe I can check after aSend? 

I think what you suggest won't work, since here we first need to exchange size and if not zero communicate the vertex list. 
aReceive for size will always result in 0 size and no aReceive for vertex list will be called.

src/partition/ReceivedPartition.cpp
atotoun: Documentation is added. 
This is for the case where a vertex can be shared between two ranks and they have same vertex count: the lower rank will own the vertex.

src/partition/ReceivedPartition.cpp
atotoun: Agreed and added.

atotoun:
@uekerman  I applied all the comments except one which I do not completely understand. Could you please have a look and see if you agree with my reply?

src/partition/tests/ReceivedPartitionTest.cpp
```diff
@@ -1087,6 +1087,147 @@ BOOST_AUTO_TEST_CASE(TestCompareBoundingBoxes3D)
...
+
+    part.communicate();
+    part.compute();
+
+    if (context.isMaster()) {
+      BOOST_TEST(mesh->getCommunicationMap()[0][0] == 0);
+      BOOST_TEST(mesh->getCommunicationMap()[0][1] == 1);
+      BOOST_TEST(mesh->getCommunicationMap()[0][2] == 2);
+
+      // std::cout<<"master sends vertices: " << mesh->getCommunicationMap()[0] << " to rank 0 and verticees " << mesh->getCommunicationMap()[1] << " to rank 1 " << std::endl;
```
uekerman, line 1198: Let's not merge these.

src/partition/tests/ReceivedPartitionTest.cpp
```diff
@@ -1087,6 +1087,147 @@ BOOST_AUTO_TEST_CASE(TestCompareBoundingBoxes3D)
...
+
+    part.addFromMapping(boundingFromMapping);
+    part.addToMapping(boundingToMapping);
+
+    part.compareBoundingBoxes();
+
+    m2n->requestSlavesPreConnection("FluidSlaves", "SolidSlaves");
+
+    part.communicate();
+    part.compute();
```
uekerman, line 1227: Shouldn't we test whether the actual owner information is as expected?

uekerman (changes requested):
I had a more thorough look. In principle the algorithmic idea is nice, but I am not yet sure if it does always work. We rely on the fact that all ranks always make the same decision on who is the owner, right? At the same time the decision depends on "counter", which changes dynamically. A particularly complicated case is if three ranks could own a vertex.
We definitely need to test more edge cases.

src/partition/ReceivedPartition.cpp
```diff
@@ -464,145 +466,330 @@ void ReceivedPartition::createOwnerInformation()
...
+
+    1- receive local bb map from master
+    2- filter bb map to keep the connected ranks
+    3- own the vertices that only fits into the ranks bb
+    4- send number of owned vertices and shared vertices list to neighbors
+    5- for the remaining vertices: check if we have less vertexes -> own it!
+    */
+
+    // #1: receive local bb map from master
+    // Define and initialize localBBMap to save local bbs
```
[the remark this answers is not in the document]
atotoun: Since we also communicate the remote BBs in other sections, keeping local would help to avoid confusion. I would say let's keep it.

src/partition/ReceivedPartition.cpp
```diff
@@ -464,145 +466,330 @@ void ReceivedPartition::createOwnerInformation()
...
+      for (auto & sharingRank : sharedVerticesReceiveMap)
+      {
+        std::vector<double> vec = sharingRank.second;
+        if( std::find(vec.begin(), vec.end(), sharedVerticesList[i]) != vec.end() )
+        {
+          if((counter > neighborRanksVertexCount[sharingRank.first]) ||
+             (counter == neighborRanksVertexCount[sharingRank.first] && utils::MasterSlave::getRank() > sharingRank.first) )
+          {
+            owned = false;
+          }
```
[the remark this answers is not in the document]
atotoun: Yes, that's right. 
If a vertex is owned by n ranks, this means the vertex is located in the BB of all of them -> all n ranks are neighbors and have intersecting BBs -> each rank compares the counter against all other n-1 -> this results in a unique decision. 

I agree with the edge testing, I all come up with some thing.

src/partition/ReceivedPartition.cpp
```diff
@@ -464,145 +466,330 @@ void ReceivedPartition::createOwnerInformation()
...
-          ranksAtInterface++;
+    for (auto & neighborRank: sharedVerticesSendMap)
+    {
+      int receiveSize = 0;
+      utils::MasterSlave::_communication->receive(receiveSize, neighborRank.first);
+      if (receiveSize != 0)
+      {
+        std::vector<double> receivedSharedVertices;
+        utils::MasterSlave::_communication->receive(receivedSharedVertices, neighborRank.first);
+        sharedVerticesReceiveMap.insert(std::make_pair(neighborRank.first, receivedSharedVertices));
```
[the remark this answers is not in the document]
atotoun: I still do not completely understand why aSend/receive can result in a deadlock but send/aReceive won't. 

The problem with aReceive/send is, that we have to separate the communication of number of vertices and vertices list. This is because, we first need to check if there is any vertex to communicate. If we do them together, the list size would always be 0, since it takes time to aReceive gets it,  and therefore, it will not enter to the if where it receives the actual list. 

If we separate the size and list communication, we would need to define another map to save the message size for each rank separately. This would be ugly in my opinion, and I would avoid it unless we make sure that the deadlock threat is real ;)  

Regarding the check after aSend, I think it is not necessary, since the aSend is followed by a receive. This is an implicit barrier, right?

src/partition/ReceivedPartition.cpp
atotoun: I replaced aSend/receive with a aReceive/send. 

For the rest, let's continue the discussion below.

src/partition/tests/ReceivedPartitionTest.cpp
atotoun: Right, test is modified.

src/partition/ReceivedPartition.cpp
```diff
@@ -464,145 +466,330 @@ void ReceivedPartition::createOwnerInformation()
...
+    5- for the remaining vertices: check if we have less vertexes -> own it!
+    */
+
+    // #1: receive local bb map from master
+    // Define and initialize localBBMap to save local bbs
+ 
+    mesh::Mesh::BoundingBoxMap localBBMap;
+    for (int rank = 0; rank < utils::MasterSlave::getSize(); rank++) {
+      localBBMap.emplace(rank, _bb);
+    }
```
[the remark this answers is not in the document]
atotoun: changed.

atotoun:
All comments are considered, either applied or replied. @uekerman ready for check.

src/partition/ReceivedPartition.cpp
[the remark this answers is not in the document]
uekerman: aSend uses `MPI_Isend` under the hood. There is no guarantee that this function is asynchronous. For large messages (such as the list of shared vertices at some point?) it might block. More information [here](https://www.mcs.anl.gov/research/projects/mpi/sendmode.html) for example.
Whenever we use an `aSend` or an `aReceive`, we should always check the request. Otherwise this is prone to fail if somebody else further develops this code part. You cannot be sure that the implicit barrier will always be there. Also, MPI does not give any guarantee that this implicit barrier is enough.
If we don't need the request, we could also just use a normal `send`.

src/partition/ReceivedPartition.cpp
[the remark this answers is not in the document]
uekerman: Yes, I think this can work. Before I thought that `counter` gets increased during this "2nd round", but it is actually frozen, which I guess is OK. Should not result in too much load imbalance.

uekerman:
Looks good in principle. 
* I still think it would be good to wait till the asynchronous communication succeeds.
* Moreover, I have a bit a bad feeling about the test coverage. We could actually do much simpler tests here. No need to have a "other participant". You could just create a `ReceivedPartition` on one participant and really unit test `createOwnerInformation()`. By creating a general test hull (similar to [PetRBF](https://github.com/precice/precice/blob/f1cb45f92cd5148174461d6123c69b2f2c92d08b/src/mapping/tests/PetRadialBasisFctMappingTest.cpp#L126) for example) and then probing it with different data sets. Understandable what I mean?

src/partition/tests/ReceivedPartitionTest.cpp
```diff
@@ -1071,414 +1074,330 @@ BOOST_AUTO_TEST_CASE(TestParallelSetOwnerInformation2D)
...
    */
-  PRECICE_TEST("Solid"_on(1_rank), "Fluid"_on(3_ranks).setupMasterSlaves(), Require::Events);
+  PRECICE_TEST(""_on(4_ranks).setupMasterSlaves(), Require::Events);
   //mesh creation
   int           dimensions   = 2;
   bool          flipNormals  = true;
   double        safetyFactor = 0;
   mesh::PtrMesh mesh(new mesh::Mesh("mesh", dimensions, testing::nextMeshID()));
-  mesh::PtrMesh receivedMesh(new mesh::Mesh("mesh", dimensions, testing::nextMeshID()));
+  // mesh::PtrMesh receivedMesh(new mesh::Mesh("mesh", dimensions, testing::nextMeshID()));
```
uekerman, line 1086:
```suggestion
```

uekerman (commented):
The tests still only cover very simple cases: 

* If shared vertices, all ranks hold them
* Then always rank 0 is owner
* No cases where the number of already owned vertices is used as criterion
* No cases with empty partitions

 With a general test hull, we could increase the coverage significantly without duplicating much code.

atotoun:
In fact, the first two cases (lower rank and rank with already lower vertices) had already been covered. I refactored the tests with the general hull and added a test for the case with empty partition.

src/partition/tests/ReceivedPartitionTest.cpp
```diff
@@ -1066,6 +1069,415 @@ BOOST_AUTO_TEST_CASE(TestCompareBoundingBoxes3D)
...
+    position << -2.0, 0.0, 1.0;
+    mesh->createVertex(position);
+    position << -0.1, 1.0, 1.0;
+    mesh->createVertex(position);
+    position << -1.0, 1.0, 1.0;
+    mesh->createVertex(position);
+    position << -2.0, 1.0, 1.0;
+    mesh->createVertex(position);
+  } else if (context.isRank(2)) {
+    Eigen::VectorXd position(dimensions);
```
uekerman, line 1419:
```suggestion
```

src/partition/tests/ReceivedPartitionTest.cpp
```diff
@@ -1066,6 +1069,415 @@ BOOST_AUTO_TEST_CASE(TestCompareBoundingBoxes3D)
...
+    }
+    BOOST_TEST(includeVertex == true);
+  }
+}
+
+BOOST_AUTO_TEST_CASE(parallelSetOwnerInformationEmptyPartition)
+{
+  /*
+    This test examines an edge case for parallel setOwnerinformation function in receivedpartition.cpp
+    for 2LI. The provided mesh includes a vertices at point (0, 0, 0) and (0, 0, 1). The rank 2 has an
```
uekerman, line 1356:
```suggestion
    for 2LI. The provided mesh includes vertices at points (0, 0, 0) and (0, 0, 1). Rank 2 has an 
```

uekerman (approved):
Test coverage looks better now :+1:
I guess, we need the templating  of `mesh` for the friend decleration of `testParallelSetOwnerInformation`?

@fsimonis What's the best practice here for white box testing? Can we avoid the templating?

atotoun:
@uekerman Yes, the compiler does not recognize the mesh at this point. I guess one way would be forward declaration of mesh class before this, but I am not sure how this would affect the rest of the things. Using Template seemed the easiest and cleanest to me. Open to better suggestions @fsimonis.

fsimonis:
Hi, I think the easiest would be to use something analogue to the WhiteBoxAccessor and the `impl()` function.
* Add a new struct
* Make it a friend of whatever class X it needs to access.
* Then add a function that given an &X, returns what you want to inspect.
* Then create a Boost test using a the accessor as Fixture.

https://github.com/precice/precice/blob/f71e555ca03c6b0f11443088a3b651b6e747033d/src/testing/Testing.hpp#L47-L62

https://github.com/precice/precice/blob/f71e555ca03c6b0f11443088a3b651b6e747033d/src/precice/SolverInterface.hpp#L787-L788

https://github.com/precice/precice/blob/f71e555ca03c6b0f11443088a3b651b6e747033d/src/precice/SolverInterface.hpp#L15-L17

Docs for using [Boost.Test with fixtures](https://www.boost.org/doc/libs/1_60_0/libs/test/doc/html/boost_test/tests_organization/fixtures/case.html). You need to use the accessor here, then you can directly call its member functions in the test.

src/partition/tests/ReceivedPartitionTest.cpp
```diff
@@ -1066,6 +1069,414 @@ BOOST_AUTO_TEST_CASE(TestCompareBoundingBoxes3D)
...
+
+  testing::ConnectionOptions options;
+  options.useOnlyMasterCom = false;
+  options.useTwoLevelInit  = true;
+  options.type             = testing::ConnectionType::PointToPoint;
+
+  auto                                      participantCom = com::PtrCommunication(new com::SocketCommunication());
+  m2n::DistributedComFactory::SharedPointer distrFactory;
+
+  auto m2n = m2n::PtrM2N(new m2n::M2N(participantCom, distrFactory, options.useOnlyMasterCom, options.useTwoLevelInit));
```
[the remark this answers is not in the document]
atotoun: Since there is only one participant, this does not work. It fails when connecting masters, since there is only one master.

src/partition/tests/ReceivedPartitionTest.cpp
```diff
@@ -1066,6 +1069,414 @@ BOOST_AUTO_TEST_CASE(TestCompareBoundingBoxes3D)
   tearDownParallelEnvironment();
 }
 
+template <typename T>
+void testParallelSetOwnerInformation(T &mesh, int dimensions)
```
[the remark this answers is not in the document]
atotoun: Doing: 
```
void testParallelSetOwnerInformation(mesh::PtrMesh &mesh, int dimensions)
```

atotoun:
> Hi, I think the easiest would be to use something analogue to the WhiteBoxAccessor and the `impl()` function.
> 
> * Add a new struct
> * Make it a friend of whatever class X it needs to access.
> * Then add a function that given an &X, returns what you want to inspect.
> * Then create a Boost test using a the accessor as Fixture.
> 
> https://github.com/precice/precice/blob/f71e555ca03c6b0f11443088a3b651b6e747033d/src/testing/Testing.hpp#L47-L62
> 
> https://github.com/precice/precice/blob/f71e555ca03c6b0f11443088a3b651b6e747033d/src/precice/SolverInterface.hpp#L787-L788
> 
> https://github.com/precice/precice/blob/f71e555ca03c6b0f11443088a3b651b6e747033d/src/precice/SolverInterface.hpp#L15-L17
> 
> Docs for using [Boost.Test with fixtures](https://www.boost.org/doc/libs/1_60_0/libs/test/doc/html/boost_test/tests_organization/fixtures/case.html). You need to use the accessor here, then you can directly call its member functions in the test.

I agree that adding a whightboxaccessor would help for further testing, as we would have this as a friend. However, in this test we really need to make the test friend of the class, as we want to call some private functions of this class. We already have access to what we want to inspect (mesh object), as we pass mesh by reference.

src/partition/tests/ReceivedPartitionTest.cpp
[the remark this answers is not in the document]
fsimonis: Ah you are right! Sorry about that

src/partition/tests/ReceivedPartitionTest.cpp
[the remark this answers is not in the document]
fsimonis: Yes, you need to change the friend declaration too.

src/partition/tests/ReceivedPartitionTest.cpp
```diff
@@ -1066,6 +1069,414 @@ BOOST_AUTO_TEST_CASE(TestCompareBoundingBoxes3D)
...
+  boundingFromMapping->setMeshes(mesh, mesh);
+  boundingToMapping->setMeshes(mesh, mesh);
+
+  ReceivedPartition part(mesh, ReceivedPartition::ON_SLAVES, safetyFactor);
+  part.addM2N(m2n);
+
+  part.addFromMapping(boundingFromMapping);
+  part.addToMapping(boundingToMapping);
+
+  part._mesh->computeBoundingBox();
```
fsimonis, line 1099: Can't you call this directly on the mesh instead going through the partition?

fsimonis:
> I agree that adding a whightboxaccessor would help for further testing, as we would have this as a friend. However, in this test we really need to make the test friend of the class, as we want to directly call some private functions of this class. We already have access to what we want to inspect (mesh object), as we pass mesh by reference.

If you need to call private functions from the test, then you can also use a friend-fxture. If you want to call P.functionA(a,b,c). Then you can create add a member function to the fixture which takes the partition and the arguments and calls the private function like this:
 ```cpp
class P {
private:
 void functionA(a,b,c);
 struct PFixture;
 friend PFixture;
};

struct PFixture {
  void functionA(P& p, A& a, B& b, C& c) {
    p.functionA(a, b, c)
  }
};
```

File: src/partition/tests/ReceivedPartitionTest.cpp (hunk 24 above)
Comment (atotoun) [reply src/partition/tests/ReceivedPartitionTest.cpp:1099]: yes, I can
done!

File: src/partition/tests/ReceivedPartitionTest.cpp (hunk 23 above)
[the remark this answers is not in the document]
Comment (atotoun) [reply src/partition/tests/ReceivedPartitionTest.cpp]: That way compilation fails. Compiler does not recognize mesh at this point. 
No need anymore, I implemented the fixture, no test friendship needed!

**Comment by atotoun** [comment]
> > I agree that adding a whightboxaccessor would help for further testing, as we would have this as a friend. However, in this test we really need to make the test friend of the class, as we want to directly call some private functions of this class. We already have access to what we want to inspect (mesh object), as we pass mesh by reference.
> 
> If you need to call private functions from the test, then you can also use a friend-fxture. If you want to call P.functionA(a,b,c). Then you can create add a member function to the fixture which takes the partition and the arguments and calls the private function like this:
> 
> ```c++
> class P {
> private:
>  void functionA(a,b,c);
>  struct PFixture;
>  friend PFixture;
> };
> 
> struct PFixture {
>  void functionA(P& p, A& a, B& b, C& c) {
>    p.functionA(a, b, c)
>  }
> };
> ```

Done.

File: src/partition/ReceivedPartition.hpp [hunk 25]
```diff
@@ -104,8 +99,28 @@ class ReceivedPartition : public Partition {
...
+    part.createOwnerInformation();
+  }
+  void tagMeshFirstRound(ReceivedPartition &part)
+  {
+    part.tagMeshFirstRound();
+  }
+  void prepareBoundingBox(ReceivedPartition &part)
+  {
+    part.prepareBoundingBox();
+  }
```
fsimonis, line 123: Moving this into a dedicated testing header would make things cleaner. Maybe `src/partition/tests/fixtures.hpp`?

src/partition/ReceivedPartition.hpp
```diff
@@ -104,8 +99,28 @@ class ReceivedPartition : public Partition {
...
+  friend struct receivedPartitionFixture;
+};
+
+/*
+ * @brief A fixture that is used to access private functions of the receivedPartition class.
+ *
+ * The fixture can be used to call private functions for individual testing. 
+ */
+struct receivedPartitionFixture {
+  void createOwnerInformation(ReceivedPartition &part)
```
fsimonis, line 112: Making these functions static will allow to call the functions without creating an instance of the fixture first.

```suggestion
  static void createOwnerInformation(ReceivedPartition &part)
```

src/partition/tests/ReceivedPartitionTest.cpp
```diff
@@ -1096,11 +1095,12 @@ void testParallelSetOwnerInformation(T &mesh, int dimensions)
...
-  part._mesh->computeBoundingBox();
-  part.prepareBoundingBox();
+  mesh->computeBoundingBox();
 
-  part.tagMeshFirstRound();
-  part.createOwnerInformation();
+  receivedPartitionFixture fixture;
+  fixture.prepareBoundingBox(part);
+  fixture.tagMeshFirstRound(part);
+  fixture.createOwnerInformation(part);
```
fsimonis, line 1103: If the functions are static one can do this:

```suggestion
  using Access = receivedPartitionFixture;
  Access::prepareBoundingBox(part);
  Access::tagMeshFirstRound(part);
  Access::createOwnerInformation(part);
```

src/partition/ReceivedPartition.cpp
```diff
@@ -469,144 +470,345 @@ void ReceivedPartition::createOwnerInformation()
...
+    std::vector<com::PtrRequest> vertexListRequests;
+
+    for (auto &receivingRank : sharedVerticesSendMap) {
+      int  sendSize = receivingRank.second.size();
+      auto request  = utils::MasterSlave::_communication->aSend(sendSize, receivingRank.first);
+      vertexListRequests.push_back(request);
+      if (sendSize != 0) {
+        auto request = utils::MasterSlave::_communication->aSend(receivingRank.second, receivingRank.first);
+        vertexListRequests.push_back(request);
       }
```
fsimonis, line 622: This logic can be implemented in `com::Communication` (the base class).
We could use an additional `out` iterator to save the requests:

Something along the lines of
```cpp
template<typename OutIter>
void Communication::aSend(const std::vector<int>& v , int rank, OutIter out) {
  //  ...
  *(out++) = sSend(v.size(), receivingRank.first);
  // ...
  *(out++) = aSend(v.data(), v.size(), receivingRank.first);
  // ---
}


utils::MasterSlave::_communication->aSend(receivingRank.second, receivingRank.first, std::back_inserter(vertexListRequests));
```

src/partition/ReceivedPartition.cpp
```diff
@@ -469,144 +470,345 @@ void ReceivedPartition::createOwnerInformation()
...
-          slaveOwnerVecs[rank][i]                 = 1; // Now rank is owner
-          globalOwnerVec[slaveGlobalIDs[rank][i]] = 1; // Vertex now has owner
-          counter++;
-          if (counter == localGuess)
-            break;
-        }
+    for (auto &neighborRank : sharedVerticesSendMap) {
+      int receiveSize = 0;
+      utils::MasterSlave::_communication->receive(receiveSize, neighborRank.first);
+      if (receiveSize != 0) {
```
fsimonis, line 628: Isn't this already covered by the vector overload of receive?

src/partition/ReceivedPartition.cpp
```diff
@@ -469,144 +470,345 @@ void ReceivedPartition::createOwnerInformation()
...
     }
 
-    // Second round: distribute all other vertices in a greedy way
-    PRECICE_DEBUG("Decide owners, second round in greedy way");
-    for (int rank : utils::MasterSlave::allRanks()) {
-      for (size_t i = 0; i < slaveOwnerVecs[rank].size(); i++) {
-        if (globalOwnerVec[slaveGlobalIDs[rank][i]] == 0 && slaveTags[rank][i] == 1) {
-          slaveOwnerVecs[rank][i]                 = 1;
-          globalOwnerVec[slaveGlobalIDs[rank][i]] = rank + 1;
+    // wait until aSend is complete.
```
fsimonis, line 635:
```suggestion
    // wait until all aSend are complete.
```

src/partition/ReceivedPartition.cpp
```diff
@@ -469,144 +470,345 @@ void ReceivedPartition::createOwnerInformation()
...
+    lowest number of vertices gets ownership to keep the load as balanced as 
+    possible. 
+    
+    Following steps are taken:
+
+    1- receive local bb map from master
+    2- filter bb map to keep the connected ranks
+    3- own the vertices that only fit into this rank's bb
+    4- send number of owned vertices and the list of shared vertices to neighbors
+    5- for the remaining vertices: check if we have less vertices -> own it!
```
fsimonis, line 494: 6. is missing here.

src/partition/ReceivedPartition.cpp
```diff
@@ -469,144 +470,345 @@ void ReceivedPartition::createOwnerInformation()
...
+              (counter == neighborRanksVertexCount[sharingRank.first] && utils::MasterSlave::getRank() > sharingRank.first)) {
+            owned = false;
+          }
         }
       }
+      if (owned) {
+        tags[sharedVerticesLocalIDs[i]] = 1;
+      } else {
+        tags[sharedVerticesLocalIDs[i]] = 0;
+      }
```
fsimonis, line 663: Can be written a bit more compact.

```suggestion
      tags[sharedVerticesLocalIDs[i]] = owned ? 1 : 0;
```

src/partition/ReceivedPartition.cpp
```diff
@@ -469,144 +470,345 @@ void ReceivedPartition::createOwnerInformation()
...
+    /* In case that a vertex can be shared between two ranks, the rank with lower
+       vertex count will own the vertex. 
+       If both ranks have same vertex count, the lower rank will own the vertex.
+    */
+
+    for (int i = 0; i < sharedVerticesGlobalIDs.size(); i++) {
+      bool owned = true;
+
+      for (auto &sharingRank : sharedVerticesReceiveMap) {
+        std::vector<double> vec = sharingRank.second;
```
fsimonis, line 651: Shouldn't this be `int` instead of `double`?

src/partition/ReceivedPartition.cpp
```diff
@@ -469,144 +470,345 @@ void ReceivedPartition::createOwnerInformation()
...
+        if (std::find(vec.begin(), vec.end(), sharedVerticesGlobalIDs[i]) != vec.end()) {
+          if ((counter > neighborRanksVertexCount[sharingRank.first]) ||
+              (counter == neighborRanksVertexCount[sharingRank.first] && utils::MasterSlave::getRank() > sharingRank.first)) {
+            owned = false;
```
fsimonis, line 655: At this moment `owned` won't change anymore.
So you could use this loop body as a predicate and  use [`std::any_of`](https://en.cppreference.com/w/cpp/algorithm/all_any_none_of) on  `sharedVerticesReceiveMap` or simply `break` to leave the for loop.

src/partition/ReceivedPartition.hpp
atotoun: when defined static, it is not able to call class members and it throws error.

src/partition/ReceivedPartition.hpp
atotoun: Done.

src/partition/ReceivedPartition.cpp
```diff
@@ -469,144 +470,345 @@ void ReceivedPartition::createOwnerInformation()
...
+    5- for the remaining vertices: check if we have less vertices -> own it!
+    */
+
+    // #1: receive local bb map from master
+    // Define and initialize localBBMap to save local bbs
+
+    mesh::Mesh::BoundingBoxMap localBBMap;
+    mesh::BoundingBox          dummyBB(_dimensions);
+    for (int rank = 0; rank < utils::MasterSlave::getSize(); rank++) {
+      localBBMap.emplace(rank, dummyBB);
```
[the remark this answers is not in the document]
atotoun: done

src/partition/ReceivedPartition.cpp
```diff
@@ -469,144 +470,345 @@ void ReceivedPartition::createOwnerInformation()
...
+    // Global IDs to be saved in:
+    std::vector<int> sharedVerticesGlobalIDs;
+    // Local IDs to be saved in:
+    std::vector<int> sharedVerticesLocalIDs;
+
+    // store possible shared vertices in a map to communicate with neighbors, map: rank -> vertex_global_id
+    std::map<int, std::vector<double>> sharedVerticesSendMap;
+
+    // receive list of possible shared vertices from neighboring ranks
+    std::map<int, std::vector<double>> sharedVerticesReceiveMap;
```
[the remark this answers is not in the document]
atotoun: That's right, changed. I also added necessary asend functions for vector<int> communication in com class.

src/partition/ReceivedPartition.cpp
```diff
@@ -469,144 +470,345 @@ void ReceivedPartition::createOwnerInformation()
...
+    }
+
+    // #3: check vertices and keep only those that fit into the current rank's bb
     int numberOfVertices = _mesh->vertices().size();
-    utils::MasterSlave::_communication->send(numberOfVertices, 0);
+    PRECICE_DEBUG("Tag vertices, number of vertices {}", numberOfVertices);
+    std::vector<int> tags(numberOfVertices, -1);
+    std::vector<int> globalIDs(numberOfVertices, -1);
+    bool             atInterface = false;
+    int              counter     = 0; // number of vertices owned by this rank
```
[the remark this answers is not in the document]
atotoun: changes counter to ownedVerticesCount. 

I need the loop index i inside the loop, thus I avoided using const auto ...

src/partition/ReceivedPartition.cpp
atotoun: I agree in general, but let's do it after pull request.

src/partition/ReceivedPartition.cpp
atotoun: You mean the message size? I think not, we need to check to avoid zero-sized messages.

src/partition/ReceivedPartition.cpp
atotoun: revised.

src/partition/ReceivedPartition.cpp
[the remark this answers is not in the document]
atotoun: done

src/partition/ReceivedPartition.cpp
atotoun: right, changes.

src/partition/ReceivedPartition.cpp
atotoun: done.

src/partition/ReceivedPartition.hpp
atotoun: done

src/partition/tests/ReceivedPartitionTest.cpp
atotoun: done

src/partition/ReceivedPartition.hpp
```diff
@@ -95,6 +98,9 @@ class ReceivedPartition : public Partition {
 
   /// Min global vertex IDs of remote connected ranks
   std::vector<int> _remoteMinGlobalVertexIDs;
+
+  /// Make the fixture friend of this class
+  friend struct receivedPartitionFixture;
```
BenjaminRodenberg, line 103: I'm just using the same pattern in #1050. I think the `friend` declaration should go before the access modifiers. Then you also directly see the friends. This looks more reasonable to me, since a private or public friend is a bit odd (even though I believe it defaults to private?). I would suggest to follow this structure:
```
class A {
friend struct friendOfA;
public:
    ...
private:
    ...
};
```

I found a similar code snipped here: https://docs.microsoft.com/en-us/cpp/cpp/friend-cpp?view=msvc-160.

However, there is also the structure you are using shown in cppreference. See https://en.cppreference.com/w/cpp/language/friend. In the end it seems to be a matter of taste. Any opinions?

atotoun: Yes, it seems more natural to me as well. I changed the declaration location as you suggest. 
One more thing: I separate the declaration and implementation of the fixture. I use a separate hpp file for that, which I think helps readability.

src/partition/ReceivedPartition.hpp
```diff
@@ -23,6 +23,9 @@ namespace partition {
  * Afterwards necessary distribution data structures are set up.
  */
 class ReceivedPartition : public Partition {
+  /// Make the fixture friend of this class
+  friend struct receivedPartitionFixture;
```
BenjaminRodenberg, line 27:
```suggestion
  friend struct ReceivedPartitionFixture;
```

src/partition/tests/ReceivedPartitionTest.cpp
```diff
@@ -1068,6 +1072,414 @@ BOOST_AUTO_TEST_CASE(TestCompareBoundingBoxes3D)
...
+
+  ReceivedPartition part(mesh, ReceivedPartition::ON_SLAVES, safetyFactor);
+  part.addM2N(m2n);
+
+  part.addFromMapping(boundingFromMapping);
+  part.addToMapping(boundingToMapping);
+
+  mesh->computeBoundingBox();
+
+  using Access = receivedPartitionFixture;
```
BenjaminRodenberg, line 1103:
```suggestion
  using Access = ReceivedPartitionFixture;
```

src/partition/tests/fixtures.hpp
```diff
@@ -0,0 +1,27 @@
...
+
+namespace precice {
+namespace partition {
+
+/*
+ * @brief A fixture that is used to access private functions of the receivedPartition class.
+ *
+ * The fixture can be used to call private functions for individual testing. 
+ */
+struct receivedPartitionFixture {
```
BenjaminRodenberg, line 11:
```suggestion
struct ReceivedPartitionFixture {
```

BenjaminRodenberg (changes requested):
* structs should be capitalized.
* `src/partition/tests/fixtures.hpp` could also go into `src/testing/ReceivePartitionFixture.hpp`. See https://github.com/precice/precice/pull/1050/files#r689567561

src/partition/tests/ReceivedPartitionTest.cpp
atotoun: done

src/partition/ReceivedPartition.hpp
atotoun: done

src/partition/tests/fixtures.hpp
atotoun: done

PR #922, opened 2020-12-16
Add nearest-neighbor-gradient mapping

mennodeij:
In many codes, the fields to be mapped also have a gradient. When mapping values, it can help convergence (both iterative and grid convergence) if instead of a first order interpolation scheme like nearest-neighbor, a second order interpolation scheme like nearest-neighbor-gradient is used. 

Therefore, this pull request adds this functionality to preCICE, with the aim to be minimally intrusive in the existing codebase, i.e. if a user does not have any gradients, everything works as normal.

Briefly, the configuration is done with XML as given below, where new tags `<gradient:scalar>` and `<gradient:vector>` are introduced. 
```
<configuration>
  <data:scalar name="Field"/>
  <gradient:scalar name="Field"/>

  <data:vector name="Field3"/>
  <gradient:scalar name="Field3"/>
  ...
</configuration>
```

An initial test showing functionality is added in `src/mesh/tests/NearestNeighborGradientMappingTest.cpp`.

I'm welcoming discussion on this pull request and hope to contribute to preCICE in this way.


mennodeij:
Let me start by saying: good that you have format checks, but from the feedback above I don't know what is wrong with the files I committed. I don't have `clang-format` to test myself.

mennodeij:
My apologies, this is a premature pull request. It seems I have missed quite a bit of functionality to get it working in the solver interface. Will report back when finished.

uekerman:
First of all, thank you very much for the potential contribution. They are very welcome!

After quickly browsing through your code, I think a sketch, formula or similar of the method you propose would help a lot. Either here or in an issue.
If I got it right what you propose builds on the fact that vertices can change:
```c++
inVertex0.setCoords(outVertex0.getCoords() + Eigen::Vector2d::Constant(0.1));
```
Right?
This is sth that currently is not possible, but sth that @fsimonis is working on.

An alternative to defining the gradient in the config would be to handle the data structure internally in the mesh. In your method, the user of preCICE would not read or write the gradient. Correct or did I understand this the wrong way?

mennodeij:
What the method entails is to improve the mapping. Currently - apart from more costly RBF method - the only method available is to find the closest vertex and use its field value:

```
F(x) = F(x_closest)
```
To improve on that, I propose to write a gradient from the solver to the exchange mesh and calculate the mapped field value as

```
F(x) = F(x_closest) + G(x_closest)*(x-x_closest)
```

But, as I said above, I will make a more complete pull request in the future.

uekerman:
OK, I see. Thanks for the explanation!
So, the gradient is indeed sth that the user would write, correct?

Of course, there is also the [nearest-project mapping](https://github.com/precice/precice/wiki/Mapping-Configuration), which has the same spatial convergence order as the gradient mapping you propose. But that one needs mesh connectivity, which your method does not. Did I get this part right?

mennodeij:
Exactly: the user writes the gradient in addition to the data to the exchange mesh. And it would indeed not require any mesh connectivity, and the user has the "freedom" to calculate their gradients any way they see fit.

uekerman:
An alternative to 
```xml
<configuration>
  <data:scalar name="Field"/>
  <gradient:scalar name="Field"/>

  <data:vector name="Field3"/>
  <gradient:scalar name="Field3"/>
  ...
</configuration>
```
could be additional API functions such as 
```c++
writeGradientScalarData(int           dataID,
                                        int           valueIndex,
                                        const double *gradientValues);)
```

fsimonis:
> Let me start by saying: good that you have format checks, but from the feedback above I don't know what is wrong with the files I committed. I don't have `clang-format` to test myself.

If you cannot get clang-format 8, then you can also use the docker image we use to test the formatting in the CI:

```
cd /path/to/precice
docker run --rm -v `pwd`:/repo -w /repo -it precice/ci-formatting:latest tools/formatting/format-all
```
This will start a new docker container based on our ci image, mount the current working directory to the `/repo` inside the container and run the formatting script inside it. The container will then stop and be removed.
Your repository should then be correctly formatted.

PR #924, opened 2020-12-18
Use repository owner

BenjaminRodenberg:
Use `github.repository_owner` to define docker username. See https://github.com/precice/precice/pull/910#issuecomment-740787797

### DIFF
--- a/.github/workflows/release-docker.yml
+++ b/.github/workflows/release-docker.yml
@@ -6,7 +6,7 @@ jobs:
   latest:
     runs-on: ubuntu-latest
     env:
-        docker_username: benjaminrueth
+        docker_username: precice
     steps:
       - uses: actions/checkout@v2
       - name: Set up Docker Buildx
--- a/.github/workflows/release-docker.yml
+++ b/.github/workflows/release-docker.yml
@@ -5,6 +5,8 @@ on:
 jobs:
   latest:
     runs-on: ubuntu-latest
+    env:
+        docker_username: benjaminrueth
     steps:
       - uses: actions/checkout@v2
       - name: Set up Docker Buildx
@@ -12,7 +14,7 @@ jobs:
       - name: Login to DockerHub
         uses: docker/login-action@v1
         with:
-          username: precice
+          username: ${{ env.docker_username }}
           password: ${{ secrets.DOCKERHUB_TOKEN }}
       - name: Build and push Dockerfile
         uses: docker/build-push-action@v2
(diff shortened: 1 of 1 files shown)


BenjaminRueth:
@fsimonis Do we have a nice way for testing this? I created the image `on: [push]` before 6b30da9  and login worked (see https://github.com/BenjaminRueth/precice/runs/1578797594?check_suite_focus=true). However, building failed, since `latest` is not an allowed version. I'm quite confident that the new implementation works, but in case you have any nice way for reviewing the changes I would be interested in how.

fsimonis:
I see the motivation behind this PR, but this introduces GitHub specific logic to save us from hard-coding a single name.
The hard-coded version is robust and the name most likely won't change anyhow.

So, I would prefer to KISS and close this PR. Arguments welcome.

BenjaminRodenberg:
Hard-coding `precice` is from my point of view not ideal, since it's easy to mix um one `precice` (username) with the other (image name). How about defining an environment variable `docker_username` in `Build and push Dockerfile` and use it? Then one can easily replace the hardcoded value `precice` when going to a fork.

BenjaminRodenberg:
> Hard-coding `precice` is from my point of view not ideal, since it's easy to mix um one `precice` (username) with the other (image name). How about defining an environment variable `docker_username` in `Build and push Dockerfile` and use it? Then one can easily replace the hardcoded value `precice` when going to a fork.

I implemented an example for this in 1c91f86. @fsimonis do you think this is a good compromise?

fsimonis:
> @fsimonis do you think this is a good compromise?

No. Hard-coding it only once is actually an improvement. :wink:
The env makes it also a bit more readable. 

Can you squash commits in this PR?

BenjaminRodenberg:
> > @fsimonis do you think this is a good compromise?
> 
> No. Hard-coding it only once is actually an improvement. wink
> The env makes it also a bit more readable.
> 
> Can you squash commits in this PR?

I usually do the squash and merge in one single step over the web interface. I now squashed all commits from this PR into a single one and force pushed. But I would still do the usual "Squash and merge" over the web interface.

.github/workflows/release-docker.yml
```diff
@@ -5,14 +5,16 @@ on:
 jobs:
   latest:
     runs-on: ubuntu-latest
+    env:
+        docker_username: benjaminrueth
```
fsimonis, line 9:
```suggestion
        docker_username: precice
```

PR #925, opened 2020-12-28
Fix compiler warnings

oguzziya:
This is a fix for the unused variable warnings for `WatchIntegral` tests. In addition, one copy elusion warning and one range-based for loop copy warning are also fixed.


fsimonis:
Thanks for the fixes!

PR #927, opened 2021-01-07
Delete apt/lists/* after apt-get update

BenjaminRodenberg:
Following recommended best practices. See https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run

I stumbled across this issue here: https://github.com/precice/python-bindings/pull/67/files/1dc2a389e036d5c50b1f0671819406686c8b25dc#r553293590

This makes `precice/precice` consistent with our other Dockerfiles. E.g. https://github.com/precice/systemtests and https://github.com/precice/ci-images.

Please note that this is a breaking change for users that currently use `precice/precice` and do not call `apt-get update` in their respective workflows!


fsimonis:
> Please note that this is a breaking change for users that currently use precice/precice and do not call apt-get update in their respective workflows!

This is not an issue, as things will eventually break for the user anyhow.  :wink:

PR #931, opened 2021-01-11
Use correct function interface in header

mennodeij:
Otherwise the function does not get de-mangled in the `extern "C" {..}` block and is not available to Fortran calls.


fsimonis:
@mennodeij Thank you very much for the fix!

Issue #929, opened 2021-01-08
Assertion hit instead of error message in writeBlockVectorData


IshaanDesai:
**Describe your setup**

Operating system: _Ubuntu 18.04_
preCICE Version: 2.1.1 (built in DEBUG mode)

preCICE commit: https://github.com/precice/precice/commit/fc9e02ce92d88132d36681cd40b98f46c2d86cf9

**Describe the problem**
The following assertion is hit:
```
ASSERTION FAILED
  Location:          void precice::impl::SolverInterfaceImpl::writeBlockVectorData(int, int, const int*, const double*)
  File:              /home/ishaan/precice/src/precice/impl/SolverInterfaceImpl.cpp:1078
  Rank:              0
  Failed expression: offset + dim < valuesInternal.size()
  Argument 0: 126
  Argument 1: 126
```
It is difficult to understand which argument to `writeBlockVectorData` is problematic and what is the problem.

**Step To Reproduce**
* Running the perpendicular flap tutorial case with SU2 and CalculiX here: https://github.com/precice/tutorials/tree/formatSU2-CCX-tutorial with commit: https://github.com/precice/tutorials/commit/43721e2d9f32c0b6da27c72664385f715b5f154e
* CalculiX-Adapter commit: https://github.com/precice/calculix-adapter/commit/914643c18e6249b59de4946f8a9d8631c34e64ea
* SU2-Adapter commit: https://github.com/precice/su2-adapter/commit/9761fd8e22d785592e912ccb6aad5d22e0dcaf29

**Expected behaviour**
An error message which explains what is the problem.

fsimonis:
One of the vertex indices is incorrect.
Given the message and the config, it looks like you registered 63 vertices and tried to access vertexid `63`.
`63 * 2 + 0 = 126`, for id `63`, 2d vertex data and the component `0`.

Do you somehow process the indices before passing them back to preCICE?

This really shouldn't be an assertion, but a user error.
The interface code is a bit messy to be honest.
Furthermore, in data access methods we look for a DataContext and then totally mix up `toData` and `fromData`.
I'll attempt to clean that up.

fsimonis:
@IshaanDesai could you try the tutorial with the version in #932? You should now run into an error instead of an assertion.

IshaanDesai:
> Do you somehow process the indices before passing them back to preCICE?

[...]

PR #932, opened 2021-01-12
Cleanup Interface Data Access

fsimonis:

- Unify and correct the data context access
  Accessor functions now extract and check the DataContext as well as its to/fromData and finally provide a `mesh::Data& data`.
  This allows to keep the remaining code short, clean and consistent.
- Refactor checks conditions to create a uniform look. This should simplify the refactoring of the solverinterface in the future.
- Reword index-related checks in the solverinterface from  
  `Error: Value index out of range. Please check that the size of dataOne is correct.` to  
  `Error: Cannot read data "dataOne" from invalid Vertex ID (42). Please make sure you only use the results from calls to setMeshVertex/Vertices().`

**Motivation**

The implementations now have a uniform look and there is no confusion of which data to use.

**Additional Information**

Related to #929

### DIFF
--- a/docs/changelog/932.md
+++ b/docs/changelog/932.md
@@ -1,0 +1,2 @@
+- Fixed input checks for data access functions.
+- Clarified the wording of errors messages in data access functions. They now refer to vertex IDs instead of indices.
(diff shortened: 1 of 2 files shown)


fsimonis:
@uekerman Can you please check that my use of the `DataContext` and the assumption of `to|fromData` is correct?

uekerman (approved):
Looks good

PR #933, opened 2021-01-13
fix indexing bug in solverdummy.cpp

IssaSaba:
the i index should be multiplied by "dimentions"  since j goes from 0 to "dimentions"

### DIFF
--- a/examples/solverdummies/c/solverdummy.c
+++ b/examples/solverdummies/c/solverdummy.c
@@ -60,9 +60,9 @@ int main(int argc, char **argv)
 
   for (int i = 0; i < numberOfVertices; i++) {
     for (int j = 0; j < dimensions; j++) {
-      vertices[j + numberOfVertices * i]  = i;
-      readData[j + numberOfVertices * i]  = i;
-      writeData[j + numberOfVertices * i] = i;
+      vertices[j + dimensions * i]  = i;
+      readData[j + dimensions * i]  = i;
+      writeData[j + dimensions * i] = i;
     }
   }
 
--- a/examples/solverdummies/cpp/solverdummy.cpp
+++ b/examples/solverdummies/cpp/solverdummy.cpp
@@ -54,9 +54,9 @@ int main(int argc, char **argv)
 
   for (int i = 0; i < numberOfVertices; i++) {
     for (int j = 0; j < dimensions; j++) {
-      vertices.at(j + numberOfVertices * i)  = i;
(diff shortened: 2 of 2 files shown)


MakisH (approved):
Thank you for the contribution, @IssaSaba! This looks correct to me and I wonder how this was working before. We don't visualize the results here, so we wouldn't have noticed it.

Maybe we also need to update the [reference output for the system tests](https://github.com/precice/systemtests/tree/develop/tests/TestCompose_1dtube_cxx/referenceOutput).

fsimonis:
Aren't we lucky that we are setting 3 vertices on a 3D mesh :see_no_evil:

IssaSaba:
You're welcome. 
Yes exactly. it works only for NumberofVertices=dimensions which is the case in the code, and any other value gives a segmentation fault. so maybe the default NumberofVertices can be changed to a more realistic value too.

uekerman:
We have the same mistake in the other solver dummies as well.

BenjaminRodenberg:
> We have the same mistake in the other solver dummies as well.

Just checked [python](https://github.com/precice/python-bindings/blob/ad14d4471363a1debf6046b393b3c35ae91bdebe/solverdummy/solverdummy.py#L45-L49) and [MATLAB](https://github.com/precice/matlab-bindings/blob/d4bc8307435a1c4c6c0967c080c0d7a5de3fd828/solverdummy/solverdummy.m#L39-L45). They look fine.

BenjaminRodenberg:
Fixed the C solverdummy in 8ad6bcc.

Fortran looks good (if I don't completely misunderstand Fortran):

https://github.com/precice/precice/blob/442becf3f7d2fe6900f76ee5b8b39b1c49a8cf7f/examples/solverdummies/fortran/solverdummy.f90#L49-L56

[Fortran-module](https://github.com/precice/fortran-module/blob/9826f277e3302cc1aef50741530538bd9d8d23c7/examples/solverdummy/solverdummy.f03#L54-L61) is also ok.

MakisH:
> Fortran looks good (if I don't completely misunderstand Fortran):
> 
> https://github.com/precice/precice/blob/442becf3f7d2fe6900f76ee5b8b39b1c49a8cf7f/examples/solverdummies/fortran/solverdummy.f90#L49-L56

Yes, this looks good. You are probably worried about the "row-first vs column-first", but this should not be an issue here. The only important thing is that we set and use the elements with the same convention.

Issue #934, opened 2021-01-14
ScaleByDtAction with SCALING_BY_DT is misleading


BenjaminRodenberg:
In preCICE the term "dt" has multiple meanings:

* time window size (what we provide in the config, time between two couplings)
* time step size (actual time step size a participant does, what goes into `advance(dt)`)

While working on #639 I observed that the `_scaling` option `SCALE_BY_DT` of `ScaleByDtAction` is very confusing:

The scaling is performed based on `fullDt` (i.e. the time window size):

https://github.com/precice/precice/blob/3195955bbe35b5c52d81fb1c1543eb4b4d9d718c/src/action/ScaleByDtAction.cpp#L45-L50

Documentation talks about the (computed) time step, which sounds for me like the actual time step:

https://github.com/precice/precice/blob/3195955bbe35b5c52d81fb1c1543eb4b4d9d718c/src/action/ScaleByDtAction.hpp#L16-L17

Finally in configuration it also sounds more like the time step:

https://github.com/precice/precice/blob/3195955bbe35b5c52d81fb1c1543eb4b4d9d718c/src/action/config/ActionConfiguration.cpp#L95-L102

### My Suggestion:

Since the implementation uses the time window size, I would suggest to follow the implementation and fix the documentation correspondingly (I'm already preparing a bigger PR for #639). However, we should update the name of the actual actions to something less confusing. 

https://github.com/precice/precice/blob/3195955bbe35b5c52d81fb1c1543eb4b4d9d718c/src/action/config/ActionConfiguration.cpp#L28-L30

are no good names. I propose:

```
NAME_SCALING_BY_TIME_STEP_TO_TIME_WINDOW_RATIO("scale-by-time-step-to-time-window-ratio"), 
NAME_SCALING_BY_COMPUTED_TIME_WINDOW_PART_RATIO("scale-by-computed-time-window-part-ratio"), 
NAME_SCALING_BY_TIME_WINDOW_SIZE("scale-by-time-window-size"), 
NAME_SCALING_BY_TIME_STEP_SIZE("scale-by-time-step-size"), 
```

If we want to make this a non-breaking change, we will need some aliases, as well. Please note that I introduced a fourth option for the time step scaling, as well.

For #639 I'm already modifying the names in the code, but keep the actual tags in the configuration the same, since I don't want to break existing code (i.e. `NAME_SCALING_BY_TIME_WINDOW_SIZE("scale-by-dt")`).

BenjaminRodenberg:
[...]

PR #935, opened 2021-01-14
Clarify time step vs. time window vs. dt

### LINKED ISSUE #639: Rename timestep to time window

BenjaminRodenberg:
Closes #639


BenjaminRodenberg:
@fsimonis no immediate need to review now, but since this PR includes some of the issues in `src/action` we just discussed it might be good taking a look at the corresponding pieces of code already now. If you think it makes sense (I'm not sure...), I can also move some of these changes (those related to #934) into another PR.

BenjaminRodenberg:
@uekerman @fsimonis: Since this PR renames many variable a second (or third) pair of eyes would be good here. With some changes in the python actions I'm not 100% sure that they are correct. Please take a closer look here.

docs/changelog/935.md
```diff
@@ -0,0 +1 @@
+- Improved documenation of [`action:scale-by-computed-dt-ratio`](https://www.precice.org/configuration-xml-reference.html#actionscale-by-computed-dt-ratio), [`action:scale-by-computed-dt-part-ratio`](https://www.precice.org/configuration-xml-reference.html#actionscale-by-computed-dt-part-ratio) and [`action:scale-by-dt`](https://www.precice.org/configuration-xml-reference.html#actionscale-by-dt) with respect to time step size vs. time window. size.
```
uekerman, line 1:
```suggestion
- Improved documentation of [`action:scale-by-computed-dt-ratio`](https://www.precice.org/configuration-xml-reference.html#actionscale-by-computed-dt-ratio), [`action:scale-by-computed-dt-part-ratio`](https://www.precice.org/configuration-xml-reference.html#actionscale-by-computed-dt-part-ratio) and [`action:scale-by-dt`](https://www.precice.org/configuration-xml-reference.html#actionscale-by-dt) with respect to time step size vs. time window size.
```

How does this PR improve the documentation? It only renames variables in preCICE.

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -33,15 +33,15 @@ BaseQNAcceleration::BaseQNAcceleration(
     double                  initialRelaxation,
     bool                    forceInitialRelaxation,
     int                     maxIterationsUsed,
-    int                     timestepsReused,
+    int                     pastTimeWindowsReused,
```
uekerman, line 36:
```suggestion
    int                     timeWindowsReused,
```
Isn't the "past" redundant as already covered by "reused"?

src/action/Action.hpp
```diff
@@ -17,7 +17,7 @@ class Action {
 public:
   /// Defines the time and place of application of the action.
   enum Timing {
-    ON_TIME_WINDOW_COMPLETE_POST, // On advancing to next dt, after adv. cpl scheme
+    ON_TIME_WINDOW_COMPLETE_POST, // On advancing to next time step, after adv. cpl scheme
```
uekerman, line 20: Are you sure here? Meaning that it is not "time window"? Does the action get called after each timestep or only after the complete time window?

src/action/PythonAction.py
```diff
@@ -1,9 +1,9 @@
 mySourceData = 0
 myTargetData = 0
 
-def performAction(time, dt, sourceData, targetData):
+def performAction(time, timeWindowSize, sourceData, targetData):
```
uekerman, line 4: Do we really want the time window size here and not the timestep size?

src/action/config/ActionConfiguration.cpp
```diff
@@ -25,9 +25,9 @@ ActionConfiguration::ActionConfiguration(
     const mesh::PtrMeshConfiguration &meshConfig)
     : NAME_DIVIDE_BY_AREA("divide-by-area"),
       NAME_MULTIPLY_BY_AREA("multiply-by-area"),
-      NAME_SCALE_BY_COMPUTED_DT_RATIO("scale-by-computed-dt-ratio"),
-      NAME_SCALE_BY_COMPUTED_DT_PART_RATIO("scale-by-computed-dt-part-ratio"),
-      NAME_SCALE_BY_DT("scale-by-dt"),
+      NAME_SCALING_BY_TIME_STEP_TO_TIME_WINDOW_RATIO("scale-by-computed-dt-ratio"),       //@todo rename this, breaking change!
+      NAME_SCALING_BY_COMPUTED_TIME_WINDOW_PART_RATIO("scale-by-computed-dt-part-ratio"), //@todo rename this, breaking change!
+      NAME_SCALING_BY_TIME_WINDOW_SIZE("scale-by-dt"),                                    //@todo rename this, breaking change!, currently misleading. See https://github.com/precice/precice/issues/934
```
uekerman, line 30: Could you please open an issue for milestone 3?
In this issue we should also discuss which actions to keep and which to drop completely.

uekerman (commented):
Looking good in general. A few difficult places, where I am also not sure completely below. If you say you had a thorough look, I'll trust you :grin:

src/acceleration/BaseQNAcceleration.cpp
BenjaminRodenberg: I agree. Will change this.

BenjaminRodenberg: Done. https://github.com/precice/precice/pull/1001

src/action/config/ActionConfiguration.cpp
BenjaminRodenberg: Done. https://github.com/precice/precice/issues/1009

docs/changelog/935.md
BenjaminRodenberg: I think my rationale was that before the meaning of `dt` was often unclear (window or step?). This PR tries to make this clear.

src/action/PythonAction.py
BenjaminRodenberg: I think this is not really defined yet. I just decided to use the `timeWindowSize`.

src/action/Action.hpp
BenjaminRodenberg:
```suggestion
    ON_TIME_WINDOW_COMPLETE_POST, // On advancing to next time window, after adv. cpl scheme
```
From the name `ON_TIME_WINDOW_COMPLETE_POST` the comment should also refer to the time window. Yes.

BenjaminRodenberg: @fsimonis Does this make sense?

src/action/PythonAction.py
BenjaminRodenberg: @fsimonis any opinion? From my point of view we just have to decide on A or B here, but I might be missing something. I decided for `timeWindowSize`.

docs/changelog/935.md
BenjaminRodenberg: Probably it's the best solution to just not add anything to the changelog.

src/action/Action.hpp
fsimonis: It is called when `_couplingScheme->isTimeWindowComplete())`, so that makes sense.

src/action/PythonAction.py
BenjaminRodenberg: In https://github.com/precice/precice/pull/935/files/0ee5d5af9f5a1393c49812dd6869808a45ce6aeb#diff-b613b1ccaeb22bcb08750b2aa4fa92acb0fa8a1bf151cab3043dd2b2f77319e0L6 it says `(global) timestep size`. I translated this to `timeWindowSize`. So everything is fine.

fsimonis: My gut feeling tells me that what we want depends on when the action is called.
I guess that the timestep size make sense for actions based on mappings and that the timewindow size makes sense for an action based on TIME_WINDOW_COMPLETE.

Not changing the behaviour is probably the best here.
We could also pass both by adding another callback in another PR.

src/action/Action.hpp
fsimonis: To clarify, `time window` makes sense :wink:

PR #395, opened 2019-05-20
cleanup Constants

BenjaminRueth:
working on #365 here.

We drop the following functions from the API:

```
const std::string& nameConfiguration();
const std::string& dataDisplacements();
const std::string& dataForces();
const std::string& dataVelocities();
```

closes #365 

- [x] ~~check whether removal of `if ( context.triggerSolverPlot ) {...}` is possible. See https://github.com/precice/precice/pull/395/commits/103c8de46c500e4c2c43192f0f05915dc7aece38~~ #407 
- [x] remove functions from python language bindings
- [x] remove functions from c language bindings
- [x] remove functions from fortran language bindings

PR #939, opened 2021-01-16
Remove obsolete trigger solver plot option

### LINKED ISSUE #407: Remove feature: triggerSolverPlot
https://github.com/precice/precice/blob/103c8de46c500e4c2c43192f0f05915dc7aece38/src/precice/impl/SolverInterfaceImpl.cpp#L320-L322

uekerman:
* closes #407 
* was a left-over TODO from #395, was still configurable, but not reachable through the `isActionRequired` interface (if you didn't guess the correct name by accident)

### DIFF
--- a/docs/changelog/939.md
+++ b/docs/changelog/939.md
@@ -1,0 +1,1 @@
+- Removed obsolete `trigger-solver-plot` config option.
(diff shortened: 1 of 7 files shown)


fsimonis:
So, this action gets forwarded to the CouplingScheme. Isn't there some dead code in there now?

uekerman:
> So, this action gets forwarded to the CouplingScheme. Isn't there some dead code in there now?

Not that I am aware of. CouplingScheme just takes care of the bookkeeping (which action is required, which one is fulfilled already, ...).

PR #940, opened 2021-01-17
Deprecate API method getMeshIDs

### LINKED ISSUE #821: Deprecate API method getMeshIDs
As discussed in https://github.com/precice/precice/pull/813, there seems to be no clear use case for `getMeshIDs`. Moreover, this currently returns an `std::set<int>`, which is a bit tricky to map to other languages.

Unless there is a clear use case for this, we should deprecate it and later remove it, to avoid additional complexity.

MakisH:
This closes #821 by:
- Adding the C++14 `[[deprecated]]` attribute in the declaration of `getMeshIDs()`
- Marking the method deprecated for Doxygen

```c++
  /**
   * @brief Returns a id-set of all used meshes by this participant.
   *
   * @deprecated Unclear use case and difficult to port to other languages.
   *             Prefer calling getMeshID for each meshName.
   * 
   * @returns the set of ids.
   */
  [[deprecated("Use getMeshID() for each meshName instead.")]]
  std::set<int> getMeshIDs() const;
```

If used at a solver, compilation will stop with the following error (g++ 10.2, assuming `-Wdeprecated-declarations`):
```
solverdummy.cpp:30:59: warning: ‘std::set<int> precice::SolverInterface::getMeshIDs() const’ is deprecated: Use getMeshID() for each meshName instead. [-Wdeprecated-declarations]
   30 |   std::set<int>         meshID     = interface.getMeshIDs();
      |                                                           ^
```

~~I am not sure if we should mark this as a breaking change: if a user enables `-Wdeprecated-declarations` or all warnings, it will complain. But the feature is still there, it just needs a "yes, I want to use deprecated things" action.~~ This should only be a warning, so everything is fine.

To reviewers:
- [x] @fsimonis looking at the code, do we need to update anything else?
- [x] @fsimonis check the (coming in a minute) changelog entry.
- [ ] @MakisH squash before merging.

### DIFF
--- a/src/precice/SolverInterface.hpp
+++ b/src/precice/SolverInterface.hpp
@@ -347,8 +347,7 @@ public:
    * 
    * @returns the set of ids.
    */
-  [[deprecated("Use getMeshID() for each meshName instead.")]]
-  std::set<int> getMeshIDs() const;
+  [[deprecated("Use getMeshID() for each meshName instead.")]] std::set<int> getMeshIDs() const;
 
   /**
    * @brief Creates a mesh vertex
--- a/src/precice/SolverInterface.hpp
+++ b/src/precice/SolverInterface.hpp
@@ -342,8 +342,12 @@ public:
   /**
    * @brief Returns a id-set of all used meshes by this participant.
    *
+   * @deprecated Unclear use case and difficult to port to other languages.
+   *             Prefer calling getMeshID for each meshName.
+   * 
    * @returns the set of ids.
    */
+  [[deprecated("Use getMeshID() for each meshName instead.")]]
   std::set<int> getMeshIDs() const;
 
   /**
--- a/src/precice/SolverInterface.hpp
+++ b/src/precice/SolverInterface.hpp
@@ -343,11 +343,11 @@ public:
    * @brief Returns a id-set of all used meshes by this participant.
    *
    * @deprecated Unclear use case and difficult to port to other languages.
-   *             Prefer calling getMeshID for each meshName.
+   *             Prefer calling getMeshID for specific mesh names.
    * 
    * @returns the set of ids.
    */
-  [[deprecated("Use getMeshID() for each meshName instead.")]] std::set<int> getMeshIDs() const;
+  [[deprecated("Use getMeshID() for specific mesh names instead.")]] std::set<int> getMeshIDs() const;
 
   /**
    * @brief Creates a mesh vertex
--- a/src/precice/SolverInterface.hpp
+++ b/src/precice/SolverInterface.hpp
@@ -257,7 +257,7 @@ public:
    *
    * @see hasToEvaluateFineModel()
    */
-  [[deprecated("The manifold mapping feature is currently not maintained.")]] bool hasToEvaluateSurrogateModel() const;
+  [[deprecated("The manifold mapping feature is no longer supported.")]] bool hasToEvaluateSurrogateModel() const;
 
   /**
    * @brief Checks if the solver has to evaluate the fine model representation.
--- a/src/precice/SolverInterface.hpp
+++ b/src/precice/SolverInterface.hpp
@@ -272,7 +272,7 @@ public:
    *
    * @see hasToEvaluateSurrogateModel()
    */
-  [[deprecated("The manifold mapping feature is currently not maintained.")]] bool hasToEvaluateFineModel() const;
+  [[deprecated("The manifold mapping feature is no longer supported.")]] bool hasToEvaluateFineModel() const;
 
   ///@}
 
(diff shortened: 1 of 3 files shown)


uekerman:
So far, we have always used preCICE warnings for deprecation. I like the C++ solution above as well, but that would change how we currently do things, just fyi.

uekerman:
I am not sure if the "use getMeshID() for each meshName instead" is what we want to say here. That somehow indicates that we know the use case :laughing: which I at least don't.

fsimonis:
`getMeshIDs()` does not make sense as there is no possibility to extract information from that id. So deprecating this is perfectly fine.

fsimonis (approved):
`getMeshIDs()` does not make sense as there is no possibility to extract information from that id. So deprecating this is perfectly fine.
Changelog and code looks good.

MakisH:
I changed the suggestion to `Use getMeshID() for specific mesh names instead.`. This should be clear enough.

MakisH:
> So far, we have always used preCICE warnings for deprecation. I like the C++ solution above as well, but that would change how we currently do things, just fyi.

This is a good point, but I only wanted to address the specific issue. As an overview, I find `@deprecated` in the following cases:
- `bool isTimeWindowComplete() const;`: We actually only deprecate it in Doxygen, there is no log output or anything else. We could again use `[[deprecated]]` and `@deprecated` here.
- `bool hasToEvaluateSurrogateModel() const;`: Same here.
- This one.

We also have in the Fortran bindings a few methods that only have a comment "Deprecated" and silently forward to their renamed equivalents. Here we could use `[[deprecated]]` and `@deprecated`.

In `src/action/config/ActionConfiguration..cpp` we use the logging macro `PRECICE_WARN` to say **at runtime** that we automatically select another action timing. This could not be replaced by `[[deprecated]]`, as this would be at compile-time.

@uekerman should I open a new issue or already make the compile-time deprecations consistent?

uekerman:
> I changed the suggestion to Use getMeshID() for specific mesh names instead.. This should be clear enough.

:+1:

> bool isTimeWindowComplete() const;: We actually only deprecate it in Doxygen, there is no log output or anything else. We could again use [[deprecated]] and @deprecated here.

Sounds like a leftover thing. `isTimeWindowComplete` should not be deprecated. `isTimestepComplete` was at some point.

> In src/action/config/ActionConfiguration..cpp we use the logging macro PRECICE_WARN to say at runtime that we automatically select another action timing. This could not be replaced by [[deprecated]], as this would be at compile-time.

True!

> @uekerman should I open a new issue or already make the compile-time deprecations consistent?

Seems only to be one further change. So feel free to do it directly here.

MakisH:
> > bool isTimeWindowComplete() const;: We actually only deprecate it in Doxygen, there is no log output or anything else. We could again use [[deprecated]] and @deprecated here.
> 
> Sounds like a leftover thing. `isTimeWindowComplete` should not be deprecated. `isTimestepComplete` was at some point.

Indeed, I was looking at the wrong method. The actual methods are `hasToEvaluateSurrogateModel` and `hasToEvaluateFineModel`.

> > @uekerman should I open a new issue or already make the compile-time deprecations consistent?
> 
> Seems only to be one further change. So feel free to do it directly here.

Done here, @uekerman please review again, in particular the deprecation messages for the surrogate/fine model API methods.

src/precice/SolverInterface.hpp
```diff
@@ -257,7 +257,7 @@ class SolverInterface {
    *
    * @see hasToEvaluateFineModel()
    */
-  bool hasToEvaluateSurrogateModel() const;
+  [[deprecated("The manifold mapping feature is currently not maintained.")]] bool hasToEvaluateSurrogateModel() const;
```
uekerman, line 260:
```suggestion
  [[deprecated("The manifold mapping feature is no longer supported.")]] bool hasToEvaluateSurrogateModel() const;
```

src/precice/SolverInterface.hpp
```diff
@@ -272,7 +272,7 @@ class SolverInterface {
    *
    * @see hasToEvaluateSurrogateModel()
    */
-  bool hasToEvaluateFineModel() const;
+  [[deprecated("The manifold mapping feature is currently not maintained.")]] bool hasToEvaluateFineModel() const;
```
uekerman, line 275:
```suggestion
  [[deprecated("The manifold mapping feature is no longer supported.")]] bool hasToEvaluateFineModel() const;
```

uekerman (approved):
Small text improvements. Otherwise, ready to merge.

src/precice/SolverInterface.hpp
uekerman: Otherwise, it sounds like we would plan to add it again.

PR #941, opened 2021-01-17
Clarify error message for checkIfDataIsExchanged

### LINKED ISSUE #842: Misleading error message "Data X is currently not exchanged"
Example situation:

```xml
<exchange data="Pressure" mesh="Fluid1-Mesh" from="Fluid1" to="Fluid2"/>
<exchange data="Velocity" mesh="Fluid2-Mesh" from="Fluid2" to="Fluid1"/>

<relative-convergence-measure limit="1.0e-3" data="Pressure" mesh="Fluid2-Mesh"/>
<relative-convergence-measure limit="1.0e-3" data="Velocity" mesh="Fluid2-Mesh"/>
```

This will lead to the following error message:
```
ERROR:  You need to exchange every data that you use for convergence measures and/or the iteration acceleration. Data "Pressure" is currently not exchanged, but used for convergence measures and/or iteration acceleration. Please check the <exchange ... /> and <...-convergence-measure ... /> tags in the <coupling-scheme:... /> of your precice-config.xml.
```

A user here would get confused, as the the data is actually exchanged, but not over the respective mesh.

We should include the name of the mesh here (which is currently not directly available in the scope):
https://github.com/precice/precice/blob/42abc38c9d82a719db0f6982548b6d979931ccf6/src/cplscheme/config/CouplingSchemeConfiguration.cpp#L1091-L1096

MakisH:
Closes #842.

The error message is now much more helpful, as the user gets the hint to check which mesh the data is being exchanged through. I tried also exporting the exact names, but this information was not easy to extract (see issue) and decided to go with the simplest solution, which is already good enough.

Reviewers:
- [x] @uekerman Check if the meaning of the error message is clear and agrees with what the method is actually checking.
- [x] @uekerman Check if my (upcoming) changelog entry is fine.
- [ ] @MakisH squash before merging.

### DIFF
--- a/docs/941.md
+++ b/docs/941.md
@@ -1,0 +1,1 @@
+- Improved the error message for not exchanging data over the same mesh used for convergence measures.
--- a/src/cplscheme/config/CouplingSchemeConfiguration.cpp
+++ b/src/cplscheme/config/CouplingSchemeConfiguration.cpp
@@ -757,13 +757,13 @@ mesh::PtrData CouplingSchemeConfiguration::getData(
                           << meshName << "\" is not configured.");
 }
 
-mesh::PtrData CouplingSchemeConfiguration::findDataByID(
+std::pair<mesh::PtrData, mesh::PtrMesh> CouplingSchemeConfiguration::findDataByID(
     int ID) const
 {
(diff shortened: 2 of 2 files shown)


uekerman (approved):
Definitely an improvement.

Issue #882, opened 2020-10-23
Extending error sub-vector of residual-sum preconditioner "numerically zero"


ajaust:
**Please describe the problem you are trying to solve.**

The error message leaves out other possible solution and could be more informative.

Current message:
```
[acceleration::ResidualSumPreconditioner]:53 in _update_: ERROR: A sub-vector in the residual-sum preconditioner became numerically zero. Thus, the preconditioner is no longer stable. Please try the value preconditioner instead.
```

**Describe the solution you propose.**

I would like to suggest the following extensions:
1. At the moment the suggested solution to a near zero `residual-sum` preconditioner is the `value` preconditioner.  An alternative would be to stick to the `residual-sum` preconditioner and to use the `freeze-after` option.
2. Maybe print the the current value of the current `residual-sum` that is considered "numerically zero" and the limit you use. At least that would be nice to know
    1. Trusting your choice of what "numerically zero" is, i.e. is your limit `1e-14` or `1e-6`?
    1. I (personally) want as much information as possible to potentially track down which part of the residual goes to zero too quickly. 

The new error message could look like that:

```
[acceleration::ResidualSumPreconditioner]:53 in _update_: ERROR: A sub-vector in the residual-sum preconditioner became numerically zero (1.42425e-22 < 1e-14). Thus, the preconditioner is no longer stable. Please try the freeze-after option or the value preconditioner instead.
```

**Describe alternatives you've considered**
None

**Additional context**
None

ajaust:
I looked into the Benjamin Uekermann's PhD thesis since I ran into this problem of a sub-vector becoming zero again.

Now I have some further questions/remarks:

- From the formulation of the error message I do not really understand which sub-vector parts becomes zero. Is it the `\tilde{x}^j - x^j` part or rather the numerator of the fraction (see p. 51 in the thesis)?
- Why is it a problem, if a sub-vector becomes zero? Am I not happy that some part of the solution is converged? 
- If a sub-vector is zero, could we decide to neglect this part when computing the residual sum. This parts does not need nor introduce anything to the scaling so I guess we are happy? Miriam suggested something like neglecting such parts in a recent meeting.
- The thesis mentions the work by Marks and Luke who use the square root of the sum as scaling factor to avoid overscaling. Is it an option in preCICE to use the square-root or is it not implemented? I am not completely sure about the implications yet, but I would be curious about it. The thesis mentions that it did not yield any benefit for FSI simulations, but it might be beneficial for other simulations.

[...]

PR #946, opened 2021-01-18
Improve error message and improve default setting for zero subvector

KyleDavisSA:
Closes https://github.com/precice/precice/issues/882
This PR improves the error message when a zero sub-vector occurs for the residual-sum preconditioner. A new default method starts using a weight of 1 for each sub-vector, and does not change the weights if a sub-vector is found. In this case, only a warning is provided instead of an error.

### DIFF
--- a/src/acceleration/impl/ResidualSumPreconditioner.cpp
+++ b/src/acceleration/impl/ResidualSumPreconditioner.cpp
@@ -46,21 +46,19 @@ void ResidualSumPreconditioner::_update_(bool                   timestepComplete
     }
     sum = std::sqrt(sum);
     if (math::equals(sum, 0.0)) {
-      PRECICE_WARN("All residual sub-vectors in the residual-sum preconditioner are numerically zero ( sum = " << sum << 
-                    "). This indicates that the data values exchanged between two succesive iterations did not change."
-                    " The simulation may be unstable, e.g. produces NAN values. Please check the data values exchanged"
-                    " between the solvers is not identical between iterations. The preconditioner scaling factors were"
-                    " not updated in this iteration and the scaling factors determined in the previous iteration were used.");
+      PRECICE_WARN("All residual sub-vectors in the residual-sum preconditioner are numerically zero ( sum = " << sum << "). This indicates that the data values exchanged between two succesive iterations did not change."
+                                                                                                                         " The simulation may be unstable, e.g. produces NAN values. Please check the data values exchanged"
+                                                                                                                         " between the solvers is not identical between iterations. The preconditioner scaling factors were"
+                                                                                                                         " not updated in this iteration and the scaling factors determined in the previous iteration were used.");
     }
 
     for (size_t k = 0; k < _subVectorSizes.size(); k++) {
       _residualSum[k] += norms[k] / sum;
       if (math::equals(_residualSum[k], 0.0)) {
-        PRECICE_WARN("A sub-vector in the residual-sum preconditioner became numerically zero ( sub-vector = " << _residualSum[k] << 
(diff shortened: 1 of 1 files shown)


MakisH:
@KyleDavisSA you need to [run `tools/formatting/format-all`](https://www.precice.org/dev-docs-dev-tooling.html) from the root directory to format your code first.

src/acceleration/impl/ResidualSumPreconditioner.cpp
```diff
@@ -45,20 +45,31 @@ void ResidualSumPreconditioner::_update_(bool                   timestepComplete
       norms[k] = std::sqrt(norms[k]);
     }
     sum = std::sqrt(sum);
-    PRECICE_CHECK(not math::equals(sum, 0.0), "All residual sub-vectors in the residual-sum preconditioner are numerically zero. "
-                                              "Your simulation probably got unstable, e.g. produces NAN values.");
+    if (math::equals(sum, 0.0)) {
+      PRECICE_WARN("All residual sub-vectors in the residual-sum preconditioner are numerically zero ( sum = " << sum << "). This indicates that the data values exchanged between two succesive iterations did not change."
+                                                                                                                         " The simulation may be unstable, e.g. produces NAN values. Please check the data values exchanged"
+                                                                                                                         " between the solvers is not identical between iterations.");
```
uekerman, line 51: As this is now a warning and not an error, we need to tell the user what is happening (i.e. no re-scaling etc).
Btw, you can improve readability of the formater with an additional line break.

src/acceleration/impl/ResidualSumPreconditioner.cpp
```diff
@@ -45,20 +45,31 @@ void ResidualSumPreconditioner::_update_(bool                   timestepComplete
...
-                                                            "Thus, the preconditioner is no longer stable. Please try the value preconditioner instead.");
+      if (math::equals(_residualSum[k], 0.0)) {
+        PRECICE_WARN("A sub-vector in the residual-sum preconditioner became numerically zero ( residualSum = " << _residualSum[k] << ") . If this occured directly after the second iteration, check that the initial-relaxation"
+                                                                                                                                      " factor is not equal to 1.0 if the coupling values of one solver"
+                                                                                                                                      " is zero in the first iteration."
+                                                                                                                                      " Please try freezing the preconditioner weights when using residual-sum"
+                                                                                                                                      " or use the value preconditioner instead.");
```
uekerman, line 61: Same as above, we need to tell the user what is happening. And we should weaken the statements, since we already have a solution now. Especially the freezing makes no sense then.
Please don't use variable names such as `residualSum`. Those are not understandable to users not reading the source code.

src/acceleration/impl/ResidualSumPreconditioner.cpp
```diff
@@ -45,20 +45,31 @@ void ResidualSumPreconditioner::_update_(bool                   timestepComplete
...
+      }
     }
 
     offset = 0;
     for (size_t k = 0; k < _subVectorSizes.size(); k++) {
-      for (size_t i = 0; i < _subVectorSizes[k]; i++) {
-        _weights[i + offset]    = 1 / _residualSum[k];
-        _invWeights[i + offset] = _residualSum[k];
+      if (!math::equals(_residualSum[k], 0.0)) {
```
uekerman, line 67:
```suggestion
      if (not math::equals(_residualSum[k], 0.0)) {
```

uekerman (changes requested):
Solution is good, but we need to make it more explicit to the user what we now do.

uekerman (approved):

PR #949, opened 2021-01-18
Release v2.2.0

fsimonis:
## How to work with this template

* [x] assign a release manager, who takes care of the process
* [x] assign each point below to a responsible person, before you continue. Use `@member`.

Only the release manager should update this post (even tickboxes, due to race conditions in editing). Everybody else should comment on the PR with the progress.

## Step by step guide

* [x] Look over [`CHANGELOG.md`](https://github.com/precice/precice/blob/develop/CHANGELOG.md) (all)
   * Add things, if necessary
   * Extract summary
   * Fix wording and tense
   * Sort the entries lexicographically
* [x] Merge master to develop (No commits after the release on master)
* [x] Check code base w.r.t code formatting (run [`precice/tools/formatting/check-format`](https://github.com/precice/precice/blob/develop/tools/formatting/check-format)) and reformat if required (run [`precice/tools/formatting/format-all`](https://github.com/precice/precice/blob/develop/tools/formatting/format-all))
* [x] Create branch `release-N` from develop. If needed, `git rebase develop`.
* [x] Open PR from `release-N` to `master` (use [this template](https://github.com/precice/precice/blob/add_PR_template/.github/PULL_REQUEST_TEMPLATE/release_pull_request_template.md))
* [ ] Do regression tests using the release branch (specific revision) _list below :arrow_down:_ (all)
* [ ] Fix potential problems on release branch (all)
* [ ] Look over the [Roadmap](https://www.precice.org/fundamentals-roadmap.html) and update entries.
* [x] Run `tools/releasing/bumpversion.sh MAJOR.MINOR.PATCH` to bump the version
* [x] Verify the version changes in:
   * [x] [CHANGELOG](https://github.com/precice/precice/blob/develop/CHANGELOG.md)
   * [x] [CMakeLists.txt](https://github.com/precice/precice/blob/develop/CMakeLists.txt)
   * [x] [debian changelog](https://github.com/precice/precice/blob/develop/tools/releasing/packaging/debian/changelog)
   * [x] (prepare independent release) [Python bindings](https://github.com/precice/python-bindings) @BenjaminRodenberg 
   * [x] (prepare independent release, if necessary!) [MATLAB bindings](https://github.com/precice/matlab-bindings) @BenjaminRodenberg 
* [x] Commit the version bump
* [ ] Draft message to mailing list @fsimonis 
* [ ] Write a draft "blog post" on [Discourse](https://precice.discourse.group/)
* [ ] Approve the PR with at least two reviews (all)
[...]

PR #951, opened 2021-01-19
Fix bug in configuration of Aitken acceleration

uekerman:
## Main changes of this PR

* Fixes a bug observed while testing the v2.2 release. See https://github.com/precice/precice/pull/949#issuecomment-762928195
* With the introduced fix the OpenFOAM-Nutils works.
* Also introduces an integration test to prevent running into this problem again.
* Downside of this fix is that we now have again the unnecessary attribute `enforce` for Aitken underrelexation.

## Author's checklist

## Reviewers' checklist


fsimonis:
Instead of adding the pointless tag again we could also adjust the parsing step here:
https://github.com/precice/precice/blob/38107651ef139927aad63b5be9e2e590f7e30003/src/acceleration/config/AccelerationConfiguration.cpp#L190-L192

```cpp
  } else if (callingTag.getName() == TAG_INIT_RELAX) {
    _config.relaxationFactor       = callingTag.getDoubleAttributeValue(ATTR_VALUE);
    if (callingTag.hasAttribute(ATTR_ENFORCE)) {
      _config.forceInitialRelaxation = callingTag.getBooleanAttributeValue(ATTR_ENFORCE);
    } else {
      _config.forceInitialRelaxation = false;
    }

```

fsimonis:
Another interesting observation: 
* the test does not fail when running with the release version of the release candidate.
* the test does not fail when running with the debug version of the release candidate.

fsimonis (approved):

Issue #735, opened 2020-04-20
Refactor RTree into query Package


fsimonis:
The implementation of the spacial index trees has become a major part of the Mesh and Mapping packages. However, the use of the `boost.geometry` rtree is an implementation detail and should not spread through our code-base. Emitted errors are often highly complicated and reading the code requires knowledge of some advances concepts.

To improve readability and the overall structure of the code, we should:
1. refactor the rtree from the `mesh` into the `query` package.
2. isolate it's use-cases and define clear interfaces to simplify the mapping codes. This is especially important for the cascaded probing in the nearest-projection mapping.
3. replace the old query functionality with an rtree-based implementation ( related to #243 )

fsimonis:
Closed by #957

Issue #243, opened 2019-02-02
Use RTree in WatchPoint


fsimonis:
We can use the _possibly cached_ RTrees of a Mesh to calculate the nearest-projection of a watchpoint.

This requires the same functionality as the following:
 https://github.com/precice/precice/blob/3204f47af46a53750d3840ef14fc9fd65d7fc3ba/src/mapping/NearestProjectionMapping.cpp#L46

PR #954, opened 2021-01-26
Refactor RTree into Query Package

oguzziya:
## Main changes of this PR
Changes according to #735 and #243

- Refactor the query functions and create an additional abstraction layer above `boost::geometry` with clear and limited use cases.
- Completly remove direct accesses to RTree from the interface
- Remove old and unused functionality from the query package

`boost::geometry` spread into the library, which increases compilation times and error complexity significantly. In addition, it reduces code readability. Removing RTree from the interface and using it on the `.cpp` file prevents the heavy compilation burden. In addition, clear and restricted usage of the `boost::geometry` functions make the code more maintainable.

The nearest projection mapping and watchpoint use similar interpolation ideas. Uniting this functionality into a single interface decreases code duplication and increases maintainability. Similarly, watchpoint benefits from the already cached RTree, which would increase the performance. 

## Author's checklist

* [x] Remove getters of RTree
* [ ] Use new interface in the tests
* [x] Remove `boost::geometry::rtree` from the interface
* [x] Fix `WatchPoint` tests

## Reviewers' checklist

* [ ] Functionality is based on the static functions. Is it a good design?


src/query/RTree.hpp
```diff
@@ -0,0 +1,50 @@
...
+    return distance < other.distance;
+  };
+};
+
+class rtree {
+public:
+  /// Get the closest vertex/edge/triangle to a vertex, return indices and distance
+  static std::vector<MatchType> getClosestVertex(const mesh::Vertex &source, const mesh::PtrMesh &targetMesh, int n = 1);
+  static std::vector<MatchType> getClosestEdge(const mesh::Vertex &source, const mesh::PtrMesh &targetMesh, int n = 1);
+  static std::vector<MatchType> getClosestTriangle(const mesh::Vertex &source, const mesh::PtrMesh &targetMesh, int n = 1);
```
fsimonis, line 33: We could use 3 different MatchTypes here.
A simple solution could be to tag the different types:

```cpp
template<typename Tag>
struct MatchType { /*  ... */ };

using VertexMatch = MatchType<struct VertexMatchTag>;
using EdgeMatch = MatchType<struct EdgeMatchTag>;
//...
static std::vector<VertexMatch> getClosestVertex(const mesh::Vertex &source, const mesh::PtrMesh &targetMesh, int n = 1);
```

src/query/RTree.hpp
```diff
@@ -0,0 +1,50 @@
...
+  MatchType(double d, int i)
+      : distance(d), index(i){};
+  constexpr bool operator<(MatchType const &other) const
+  {
+    return distance < other.distance;
+  };
+};
+
+class rtree {
+public:
```
fsimonis, line 29: This pseudo-class should disappear from the interface.

oguzziya:
I identified the problem related to `WatchPoint`. Only failing tests is when we change the mesh (add an edge for example) reinitialized the `WatchPoint`. Since the `RTree` is not updated when this change occurred, it uses the previous `RTree`. There are several options to fix this problem, but I am not sure which is the best.
* Add a flag for re-initialization and rebuild the `RTree` if the `WatchPoint` is reinitialized: Lazy and expensive approach, I do not think it is a good solution.
* Send a signal (`boost::signal`?) when we add `Vertex`, `Edge` or a `Triangle`. If the respective `RTree` cache is not empty, insert the new primitive to `RTree`: I am in favor of this approach, however, I am not sure if there is a side effect or not.
* Completely remove the support for re-initialization for `WatchPoint`: *just kidding...*

PR #959, opened 2021-02-03
Improve RTree and Query Interface

oguzziya:
## Main changes of this PR
Subpackage of #956. This PR aims to improve the `query` interface and reduce the `boost::geometry` spread in the code base.

Refer #956

## Author's checklist
* [x] Add event timings for the RTree generation
* [x] Implement missing functions
* [x] Implement local caching
* [x] Replace the functions in the code base
* [x] Implement tests

## Reviewers' checklist


src/mapping/NearestProjectionMapping.cpp
```diff
@@ -134,16 +109,10 @@ void NearestProjectionMapping::computeMapping()
...
-        }
         // Search for the origin inside the destination meshes vertices
-        indexVertices->query(bg::index::nearest(coords, 1),
-                             boost::make_function_output_iterator([&](int match) {
-                               _weights[i] = query::generateInterpolationElements(fVertices[i], tVertices[match]);
-                               distanceStatistics(bg::distance(fVertices[i], tVertices[match]));
-                             }));
+        auto matchedVertices = indexTree.getClosestVertex(fVertices[i]);
+        _weights[i]          = query::generateInterpolationElements(fVertices[i], tVertices[matchedVertices.at(0).index]);
+        distanceStatistics(matchedVertices.at(0).distance);
```
fsimonis, line 115: This could use `getClosestVertex()` (non-n version)

src/mapping/PetRadialBasisFctMapping.hpp
```diff
@@ -1436,11 +1414,8 @@ PetRadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::bgPreallocationMatrixA(mesh::
     }
 
     // -- PREALLOCATE THE COEFFICIENTS --
-    results.clear();
-    auto search_box = query::getEnclosingBox(oVertex, supportRadius);
-
-    tree->query(bg::index::intersects(search_box) and bg::index::satisfies([&](size_t const i) { return bg::distance(oVertex, inMesh->vertices()[i]) <= supportRadius; }),
-                std::back_inserter(results));
+    query::Index indexTree(inMesh);
+    auto         results = indexTree.getVerticesInsideBox(oVertex, supportRadius);
```
fsimonis, line 1418: It would be interesting to benchmark the inner-loop vs outer-loop version. Not too important here though.

src/query/Index.hpp
```diff
@@ -0,0 +1,79 @@
...
+  ~Index();
+
+  /// Get n number of closest vertices to the given vertex
+  std::vector<VertexMatch> getClosestVertex(const mesh::Vertex &sourceVertex, int n = 1);
+
+  /// Get n number of closest edges to the given vertex
+  std::vector<EdgeMatch> getClosestEdge(const mesh::Vertex &sourcesVertex, int n = 1);
+
+  /// Get n number of closest triangles to the given vertex
+  std::vector<TriangleMatch> getClosestTriangle(const mesh::Vertex &sourceVertex, int n = 1);
```
fsimonis, line 54: Let's make the n explicit here by using plural versions with an explicit n parameter.
The only singular version used is to get the vertices. Edges and Triangles don't need one yet.
```cpp
VertexMatch getClosestVertex(const mesh::Vertex &sourceVertex);`.
std::vector<VertexMatch> getClosestVertices(const mesh::Vertex &sourceVertex, int n);
```

src/query/Index.hpp
```diff
@@ -0,0 +1,79 @@
...
+  std::vector<size_t> getVerticesInsideBox(const mesh::BoundingBox &bb);
+
+  /// Clear all the cache
+  static void clearCache();
+
+  /// Clear the cache of given mesh
+  static void clearCache(int meshID);
+
+  /// Clear the cache of given mesh
+  static void clearCache(mesh::Mesh &mesh);
```
fsimonis, line 69: These don't have anything to do with the index class. Let's move them to the namespace or try to get rid of them altogether.

src/query/impl/RTreeWrapper.hpp
```diff
@@ -0,0 +1,70 @@
...
+using TriangleTraits = impl::RTreeTraits<mesh::Triangle>;
+
+struct MeshIndices {
+  VertexTraits::Ptr   vertexRTree;
+  EdgeTraits::Ptr     edgeRTree;
+  TriangleTraits::Ptr triangleRTree;
+};
+
+/// Class to encapsulate boost::geometry implementations
+class RTreeWrapper {
```
fsimonis, line 28: The `RTreeWrapper` is a stateless proxyclass to the underlying `static map<>`. The user has to instantiate this, but there is really no object to speak of.

I see 2 possible ways to proceed:
1) we turn the class into a namespace, or
2) we  implement the `RTreeWrapper` as a singleton holding the map as a member and forcing access via some `static RTreeWrapper instance();`.

I would prefer the latter, as this will make things easier to understand:  
There is one instance of `<something>`. This instance is responsible for indexing meshes if required and caching them to prevent recomputations.
Given the above description, a more suitable name for the singleton would be `Indexer`.

src/query/impl/RTreeWrapper.hpp
```diff
@@ -0,0 +1,70 @@
...
+  VertexTraits::Ptr getVertexRTree(const mesh::PtrMesh &mesh);
+
+  /// Return edge index tree from cache, if cache is empty, create the tree
+  EdgeTraits::Ptr getEdgeRTree(const mesh::PtrMesh &mesh);
+
+  /// Return triangle index tree from cache, if cache is empty, create the tree
+  TriangleTraits::Ptr getTriangleRTree(const mesh::PtrMesh &mesh);
+
+  /// Return boost::geometry version of enclosing box around a point
+  Box3d getEnclosingBox(const mesh::Vertex &middlePoint, double sphereRadius);
```
fsimonis, line 40: This doesn't belong here.
Maybe this could go into a utility file of the mesh?

src/query/Index.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+private:
+  std::unique_ptr<impl::RTreeWrapper> _rtreeWrapper;
+  std::unique_ptr<impl::MeshIndices>  _cache;
```
fsimonis, line 73: This still leaks information about the implementation.
Let's use a [classical data-only `pimpl` idiom](https://www.fluentcpp.com/2017/09/22/make-pimpl-using-unique_ptr/) to hide them.

Also, you can use `impl::MeshIndices` as a member as it contains pointers that can represent the `I don't exist yet` state.

fsimonis (changes requested):
This is already making some of the code clearer to read.
So it's definitely going into the right direction.
There are still some issues though.

src/query/impl/RTreeWrapper.hpp
oguzziya: If I move this i need to include `boost::geometry` there too. Does it worth it?

src/query/Index.hpp
oguzziya: I think moving them to a namespace is a better option since we need a layer to access these functions.

src/query/Index.hpp
oguzziya: I implemented singleton pattern, therefore there won't be a pointer to `RTreeWrapper` anymore. So, should I apply pimpl idiom to `MeshIndices`, it seemed a little bit unnecessary actually since we do not need a layer for it. In addition, I did not fully understand your second comment. In the `Index.hpp` compiler has no idea about `MeshIndices`although it contains only pointers. Therefore without a pointer, it cannot compile.

src/query/Index.hpp
oguzziya: I think moving them to a namespace is a better option since we still need to access these functions without introducing requiring any `boost::geometry` related interfaces.

src/query/impl/RTreeWrapper.hpp
fsimonis: Just realised that you made this method obsolete by implementing `Index::getVerticesInsideBox()` :wink:

src/query/Index.hpp
fsimonis: > In the Index.hpp compiler has no idea about MeshIndices although it contains only pointers

This is exactly why a pointer to implementation is useful. You can then `include "query/Indexer.hpp"` in the `Index.cpp` and use the MeshIndices in the impl. This may seem overkill, but it makes the header cleaner.

src/query/Index.hpp
```diff
@@ -0,0 +1,80 @@
...
+
+private:
+  struct IndexImpl;
+  std::unique_ptr<IndexImpl> _pimpl;
+
+  const mesh::PtrMesh             _mesh;
+  static precice::logging::Logger _log;
+};
+
+namespace rtree {
```
fsimonis, line 68: You don't need this namespace here. IMHO `precice::query::clearCache(mymesh);` is clear enough.

src/query/Index.cpp
```diff
@@ -0,0 +1,152 @@
...
+
+Index::Index(const mesh::PtrMesh &mesh)
+    : _mesh(mesh)
+{
+  _pimpl = std::make_unique<IndexImpl>(IndexImpl{});
+}
+
+Index::~Index()
+{
+}
```
fsimonis, line 28: You can also tell the compiler to provide the default.

```suggestion
Index::~Index() = default;
```

fsimonis:
The last step would be to check if the produced events are still useful.
Can you run a tutorial or maybe even integration test with this PR and check the resulting event timings?
More information on this topic [in our docs](https://www.precice.org/tooling-performance-analysis.html).

We need to measure the time it took to generate RTrees during the mappings.

src/query/Index.cpp
oguzziya: Unfortunately, I cant, since I have an incomplete type for pimpl, the compiler complains.

oguzziya:
Screen Shot 2021-02-09 at 12 45 24

The first call for `getClosestVertex` takes 2ms and the remaining ones take 1ms. Is this enough or should I do more detailed analysis?

fsimonis:
The time it takes to fetch the RTree is the interesting part. Profiling the individual queries is way to verbose (and probably way too expensive too).

What we really need to know is the timing of this line:
https://github.com/precice/precice/blob/d9c7cb96b980b1967aac379e82ae70e44c06fc6a/src/query/Index.cpp#L108

As an example, this would result in 1 Event per call to `nearestneightbor::computeMapping()` and at most 3 (Vertex, Edge, Triange) for `nearestprojection::computeMapping()`.

oguzziya:
I modified the locations of the events. We have one tree fetch per compute mapping call as you mentioned. I also attach the trace file.

[event_trace.log](https://github.com/precice/precice/files/5957255/event_trace.log)


Screen Shot 2021-02-10 at 10 52 10

PR #962, opened 2021-02-10
Parallel participant support for scaled-consistent mappings

oguzziya:
## Main changes of this PR
In #906 scaled-consistent mapping is introduced only for serial participants. This PR expands it for parallel participants too. There are a few things to accomplish first.

## Motivation and additional information
Currently, scaled-consistent mapping is only supported for serial participants.

## Author's checklist

* [ ] Add ownership information for `Edge` and `Triangle`
* [ ] Introduce more aggressive repartioning to consider connectivity information
* [ ] I added a changelog file with this PR number in `docs/changelog/` if there are noteworthy changes.
* [ ] I ran `tools/formatting/check-format` and everything is formatted correctly.
* [ ] I sticked to C++14 features.
* [ ] I sticked to CMake version 3.10.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?
* [ ] (more questions/tasks)

PR #963, opened 2021-02-10
Implement ownership and global indexing for edges and triangles

oguzziya:
## Main changes of this PR
Similar to the vertex ownership and global indexing, this PR implements ownership and global indexing for other two primitives.
* Define setters and getters
* Extend communication for global indices of edges and triangles
* Extend partitioning to account for these changes

This PR is part of #962 . For connectivity based mappings, there is no way to know exactly which connectivity element is defined on which rank and which are defined twice/overlapping for a received partition.

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/com/CommunicateMesh.cpp
+++ b/src/com/CommunicateMesh.cpp
@@ -58,11 +58,14 @@ void CommunicateMesh::sendMesh(
     _communication->send(vertexIDs, rankReceiver);
 
     std::vector<int> edgeIDs(numberOfEdges * 2);
+    std::vector<int> globalEdgeIDs(numberOfEdges);
     for (int i = 0; i < numberOfEdges; i++) {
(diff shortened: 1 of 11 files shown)


oguzziya:
Currently, global indexing and ownership work as intended. Unlike the vertex owner information creation, edge and triangle information take place after mesh filtering. Why? Since mesh filtering is based on tagging and edge-triangle does not implement this algorithm, there might be ownerless edges and triangles hanging around if done before filtering.

I tried out communicating this information only needed, but it ended being way messier than I imagined therefore, the conditional implementation only exists in the repartitioning stage. In my opinion, even that can be removed.

fsimonis:
I unassigned myself as this is too out of date. Let's keep this PR as inspiration on how to solve this.

Issue #468, opened 2019-07-24
Extend data mapping to volume coupling


fsimonis:
# Volume Coupling

## Current Situation

preCICE was designed primarily with surface coupling in mind.
This means that the nearest-projection mapping make assumptions which are wrong for volume coupling. Using 2D Meshes, the mapping will only map to edges, not triangles.
The current workaround is to use a nearest-neighbor mapping with vertex information only.
Furthermore, preCICE does not support 3 dimensional primitives such as tetrahedra or hexahedra.

## The Goal

preCICE and its mapping algorithms should handle volume coupling in all cases correctly.
It should also be possible to at least tetrahedra via the interface.

## Work Packages

* [ ] Extend `MappingRequirement` and fix the requirement for the various mappings.
* [x] Extend the configuration adding a switch between surface and volume coupling.
* [x] Extend SolverInterface to support adding Tetrahedra (`setMeshTertrahedron()`)
* [x] Extend exporters to support writing Tetrahedra (e.g. VTK)
* [x] Add support to interpolating in Tetrahedra as well as generating interpolation elements
* [x] Add necessary adapters to allow usage and indexing of tetrahedra in `Boost.Geometry`.

MakisH:
Just to be clear, preCICE currently is successfully being used for volume coupling (see, e.g. [this talk](https://www.youtube.com/watch?v=TrZD9SnG2Ts)), but not all the features are currently supported / make sense (mainly: nearest-projection mapping).

uekerman:
I adjusted the title because this issue is mainly a collection for what is needed to extend the data mapping to volume coupling.
There are other themes as well, e.g. how to do realize and optimize for in-memory communication, but that's better tackled in a separate issue later.

fsimonis:
As soon as we document that the user should prefer to triangulate quads and tetrahedrize hexahedra, we can close this issue.

[Defining mesh connectivity](https://precice.org/couple-your-code-defining-mesh-connectivity)

PR #965, opened 2021-02-14
Use possibly cached index trees in watch-point

oguzziya:
## Main changes of this PR
- Change closest find functions to use already cached index trees
- Use same functionality as nearest-projection mapping for interpolation

For the calculation of the watch point, we use closest point/edge/triangle finding. For nearest-projection and nearest-neighbor mappings index trees are possibly built up therefore we can use them to efficiently get the corresponding primitives.

Additionally, nearest-projection mapping and watch-point calculation involves similar operations, both using interpolation to determine the value. Therefore, using consistent functions across the same operation helps to improve maintainability.

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/precice/impl/WatchPoint.cpp
+++ b/src/precice/impl/WatchPoint.cpp
@@ -13,9 +13,7 @@
 #include "mesh/Mesh.hpp"
 #include "mesh/Triangle.hpp"
 #include "mesh/Vertex.hpp"
-#include "query/FindClosestEdge.hpp"
-#include "query/FindClosestTriangle.hpp"
-#include "query/FindClosestVertex.hpp"
+#include "query/Index.hpp"
 #include "utils/MasterSlave.hpp"
 #include "utils/assertion.hpp"
 
@@ -58,16 +56,13 @@ void WatchPoint::initialize()
 {
   PRECICE_TRACE();
   // Clear to allow reinitialization
-  _weights.clear();
-  _vertices.clear();
+  mesh::Vertex watchVertex(_point, 1);
+  query::Index indexTree(_mesh);
 
-  // Find closest vertex
   if (_mesh->vertices().size() > 0) {
-    query::FindClosestVertex findVertex(_point);
-    findVertex(*_mesh);
-    _vertices.push_back(&findVertex.getClosestVertex());
-    _shortestDistance = findVertex.getEuclidianDistance();
-    _weights.push_back(1.0);
+    auto projectionElement = indexTree.findNearestProjection(watchVertex, 4);
+    _interpolationElements = std::move(projectionElement.first);
+    _shortestDistance      = projectionElement.second;
   }
 
   if (utils::MasterSlave::isSlave()) {
@@ -92,43 +87,11 @@ void WatchPoint::initialize()
     }
   }
 
+  if (not _isClosest) {
+    _interpolationElements.clear();
+  }
+
   PRECICE_DEBUG("Rank: " << utils::MasterSlave::getRank() << ", isClosest: " << _isClosest);
-
-  if (!_isClosest) {
-    return;
-  }
-
-  // Find closest edge
-  query::FindClosestEdge findEdge(_point);
-  if (findEdge(*_mesh)) {
-    if (findEdge.getEuclidianDistance() < _shortestDistance) {
-      //         _closestEdge = & findEdge.getClosestEdge ();
-      _vertices.clear();
-      _vertices.push_back(&findEdge.getClosestEdge().vertex(0));
-      _vertices.push_back(&findEdge.getClosestEdge().vertex(1));
-      _shortestDistance = findEdge.getEuclidianDistance();
-      _weights.clear();
-      _weights.push_back(findEdge.getProjectionPointParameter(0));
-      _weights.push_back(findEdge.getProjectionPointParameter(1));
-    }
-  }
-  if (_mesh->getDimensions() == 3) {
-    // Find closest triangle
-    query::FindClosestTriangle findTriangle(_point);
-    if (findTriangle(*_mesh)) {
-      if (findTriangle.getEuclidianDistance() < _shortestDistance) {
-        _vertices.clear();
-        _vertices.push_back(&findTriangle.getClosestTriangle().vertex(0));
-        _vertices.push_back(&findTriangle.getClosestTriangle().vertex(1));
-        _vertices.push_back(&findTriangle.getClosestTriangle().vertex(2));
-        _shortestDistance = findTriangle.getEuclidianDistance();
-        _weights.clear();
-        _weights.push_back(findTriangle.getProjectionPointParameter(0));
-        _weights.push_back(findTriangle.getProjectionPointParameter(1));
-        _weights.push_back(findTriangle.getProjectionPointParameter(2));
-      }
-    }
-  }
 }
 
 void WatchPoint::exportPointData(
@@ -138,12 +101,12 @@ void WatchPoint::exportPointData(
     return;
   }
 
-  PRECICE_ASSERT(_vertices.size() == _weights.size());
+  //PRECICE_ASSERT(_vertices.size() == _weights.size());
   _txtWriter.writeData("Time", time);
   // Export watch point coordinates
   Eigen::VectorXd coords = Eigen::VectorXd::Constant(_mesh->getDimensions(), 0.0);
-  for (size_t i = 0; i < _vertices.size(); i++) {
-    coords += _weights[i] * _vertices[i]->getCoords();
+  for (const auto elem : _interpolationElements) {
+    coords += elem.weight * elem.element->getCoords();
   }
   if (coords.size() == 2) {
     _txtWriter.writeData("Coordinate", Eigen::Vector2d(coords));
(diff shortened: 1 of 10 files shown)


src/precice/impl/WatchPoint.cpp
```diff
@@ -61,13 +59,15 @@ void WatchPoint::initialize()
   _weights.clear();
   _vertices.clear();
 
+  mesh::Vertex watchVertex(_point, 0);
```
fsimonis, line 62: Having to create a dummy point is a strong indicator for the interface of `query::Index` being incomplete.
Looking at it, there is really no reason why the `Index::getClosestX` functions require a full Vertex as argument (aside from convenience).

My suggestion for now:
* We add overloads that take `const Eigen::VectorXd&` instead of `const mesh::Vertex&`.
* We can then implement the Vertex version then using the Eigen version.

src/precice/impl/WatchPoint.cpp
```diff
@@ -99,33 +99,33 @@ void WatchPoint::initialize()
   }
 
   // Find closest edge
-  query::FindClosestEdge findEdge(_point);
-  if (findEdge(*_mesh)) {
-    if (findEdge.getEuclidianDistance() < _shortestDistance) {
-      //         _closestEdge = & findEdge.getClosestEdge ();
+  auto closestEdges = indexTree.getClosestEdges(watchVertex, 1);
```
fsimonis, line 102: The edge/triangle with the shortest distance may not be an interpolating projection.
Have a look at the implementation in the nearest projection mapping.

fsimonis (changes requested):

oguzziya:
There are a few critical changes that need suggestion:
- For the time being, 3 new member functions are added to the `Index` class to handle the projection element computation: `findVertexProjectionElements()`, `findEdgeProjectionElements()`, `findTriangleProjectionElements()`. If edge cannot find appropriate projection element, fall backs to vertex and returns corresponding vertex match and interpolation element. Similarly, if an appropriate triangle projection element cannot be found, the first fall backs to the edge, then the vertex if necessary.
- This brings the problem of return type since we identify each match type with different tags. For the time being, I added the `GenericMatch` tag in order to handle this. This part is open to suggestions.
- Although having these functions in the `Index` class is extremely handy, I think this is not the best place.

src/query/Index.cpp
```diff
@@ -91,14 +75,84 @@ std::vector<TriangleMatch> Index::getClosestTriangles(const mesh::Vertex &source
...
+  _pimpl->indices.triangleRTree->query(bgi::nearest(sourceCoord, n),
                                        boost::make_function_output_iterator([&](impl::TriangleTraits::IndexType const &match) {
-                                         matches.emplace_back(bg::distance(sourceVertex, _mesh->triangles()[match.second]), match.second);
+                                         matches.emplace_back(bg::distance(sourceCoord, _mesh->triangles()[match.second]), match.second);
                                        }));
   std::sort(matches.begin(), matches.end());
   return matches;
 }
 
+VertexMatch Index::getClosestVertex(const mesh::Vertex &sourceVertex)
```
fsimonis, line 86: Please implement using the Eigen version.

src/query/Index.cpp
```diff
@@ -91,14 +75,84 @@ std::vector<TriangleMatch> Index::getClosestTriangles(const mesh::Vertex &source
...
+  // calling .back() on an empty vector is undefined.
+  if (not matches.empty()) {
+    return matches.back();
+  } else {
+    // return a default match with empty index
+    return VertexMatch(-1, std::numeric_limits<double>::max());
+  }
+}
+
+VertexMatch Index::getClosestVertex(const Eigen::VectorXd &sourceCoord)
```
fsimonis, line 99: There is a more straight-forward implementation for this exact case, as the nearest neighbour is always the nearest projection for vertices.

Have a look at the old implementation:

https://github.com/precice/precice/blob/6fdeb97c20a92ecc47f55bdb00a8f95cc60b77af/src/mapping/NearestProjectionMapping.cpp#L141-L146

src/query/Index.cpp
```diff
@@ -91,14 +75,84 @@ std::vector<TriangleMatch> Index::getClosestTriangles(const mesh::Vertex &source
...
+{
+  PRECICE_TRACE();
+  auto matches = getClosestVertices(sourceCoord, 1);
+  // calling .back() on an empty vector is undefined.
+  if (not matches.empty()) {
+    return matches.back();
+  } else {
+    // return a default match with empty index
+    return VertexMatch(-1, std::numeric_limits<double>::max());
```
fsimonis, line 108: Let's add an assertion here. The caller should check this.

src/query/Index.cpp
```diff
@@ -91,14 +75,84 @@ std::vector<TriangleMatch> Index::getClosestTriangles(const mesh::Vertex &source
...
+  }
+}
+
+EdgeMatch Index::getClosestEdge(const Eigen::VectorXd &sourceCoord)
```
fsimonis, line 112: The singular version is not required. [YAGNI](https://en.wikipedia.org/wiki/You_aren't_gonna_need_it)

src/query/Index.cpp
```diff
@@ -91,14 +75,84 @@ std::vector<TriangleMatch> Index::getClosestTriangles(const mesh::Vertex &source
...
+  // calling .back() on an empty vector is undefined.
+  if (not matches.empty()) {
+    return matches.back();
+  } else {
+    // return a default match with -1 index and distance
+    return EdgeMatch(-1, std::numeric_limits<double>::max());
+  }
+}
+
+TriangleMatch Index::getClosestTriangle(const Eigen::VectorXd &sourceCoord)
```
fsimonis, line 125: Also not needed

src/query/Index.hpp
```diff
@@ -40,23 +41,32 @@ class Index {
...
   /// Return all the vertices inside the box formed by vertex and radius
   std::vector<size_t> getVerticesInsideBox(const mesh::Vertex &centerVertex, double radius);
 
   /// Return all the vertices inside a bounding box
   std::vector<size_t> getVerticesInsideBox(const mesh::BoundingBox &bb);
 
+  /// Temporary before other refactor, return Interpolation elements and the distance
+  std::pair<InterpolationElements, GenericMatch> findVertexProjectionElements(const mesh::Vertex &sourceVertex, int n);
+  std::pair<InterpolationElements, GenericMatch> findEdgeProjectionElements(const mesh::Vertex &sourceVertex, int n);
+  std::pair<InterpolationElements, GenericMatch> findTriangleProjectionElements(const mesh::Vertex &sourceVertex, int n);
```
fsimonis, line 68: You are right regarding the `GenericMatch` it really is pointless.
Lets use only the distance to the matched primitive `std::pair<InterpolationElements, double>` for now. If we ever need the type of the matched primitive, we should add an enum to the GenericMatch which identifies the primitive it matched against.

fsimonis (changes requested):
This looks really good now!
Just a few changes and we are ready to merge.

oguzziya:
I made the following changes:
- Removed unnecessary closest finding functions. Actually, with the refactored projection element finding functions, explicit calls on closest finding functions are not necessary and can be implemented as private functions. However, it will cause some changes in tests, therefore I plan it to do in another PR.
- I take one step back from the projection element finding functions. Previously, falling back to one lower dimension was done inside functions. However, this behaviour is a little bit hideous and might cause errors in the future. In addition, watch point follows different approach, and not finding edge or triangle projection elements would cause one unnecessary calculation of vertex projection element. Therefore, falling back to lower dimensions is done in the neares projection mapping. 
- In watchpoint, vertices and weights were stored seperately. However, with the implementation of the interpolations elements, it is not necessary and it is refactored accordingly.

fsimonis:
> I take one step back from the projection element finding functions. Previously, falling back to one lower dimension was done inside functions. However, this behavior is a little bit hideous and might cause errors in the future.

The issue here is that this implementation requires a sentinel value, which requires to be checked by the caller, which in turn is dangerous. The resulting code is also less readable in both the function and the call-site.
The nearest-projection has the nice property that there is always a nearest projection, so lets use it.
This will get a bit more complicated when we eventually implement projection to volumes #468.

Anyhow, a good function name and documentation is the solution here.

> In addition, watchpoint follows a different approach, and not finding an edge or triangle projection elements would cause one unnecessary calculation of vertex projection element. Therefore, falling back to lower dimensions is done in the nearest projection mapping. If you do not like it I can switch it back to the implicit one.

Let's not worry about the runtime of the watchpoint initialization. Its implementation could be simplified anyhow (Find nearest projection, determine owner, done), but that is for another time.

oguzziya:
- I changed the function name that is going to be used in the library to `findNearestProjection()`. The individual functions are still in the interface since otherwise, the implementation gets dirty and there is code duplication. In addition, in the case of future volume coupling, it is easier to extend the functions without having to add so many if conditions.
- With this change, I also simplified the `WatchPoint` initialization, we directly find the projection element, then the owner.

src/precice/impl/WatchPoint.cpp
```diff
@@ -93,35 +88,6 @@ void WatchPoint::initialize()
   }
 
   PRECICE_DEBUG("Rank: " << utils::MasterSlave::getRank() << ", isClosest: " << _isClosest);
```
fsimonis, line 90: We could clear the `_interpolationElements` when `isClosest == false` this may prevent some mistakes in the future.

src/precice/impl/WatchPoint.cpp
```diff
@@ -131,12 +97,12 @@ void WatchPoint::exportPointData(
...
   }
 
-  PRECICE_ASSERT(_vertices.size() == _weights.size());
+  //PRECICE_ASSERT(_vertices.size() == _weights.size());
   _txtWriter.writeData("Time", time);
   // Export watch point coordinates
   Eigen::VectorXd coords = Eigen::VectorXd::Constant(_mesh->getDimensions(), 0.0);
-  for (size_t i = 0; i < _vertices.size(); i++) {
-    coords += _weights[i].weight * _vertices[i]->getCoords();
+  for (const auto elem : _interpolationElements) {
```
fsimonis, line 104:
```suggestion
  for (const auto& elem : _interpolationElements) {
```
No need to copy every element. Same below.

fsimonis (approved):
This refactoring really improves readability by __a lot__.
Really nicely done!

Issue #663, opened 2020-02-05
Introduce a Tweakable Assertion Policy


fsimonis:
# Problem description

Currently, assertions are disabled in non-debug builds.

Using assertions in release builds is a trade-off between:
* higher security and trust in the results at the cost of additional work
* less work and checks at the cost of security.

What a user/developer prefers depends on the use-case and we should not enforce a choice.

# Potential Solution

Provide the additional CMake Variable `PRECICE_ASSERTION_POLICY`:

 Choice | Comment
--- | --- 
`ON` | Always enable assertions
`DEBUG` | Enable assertions on debug builds only **default**
`OFF` |  Never enable assertions

uekerman:
Mmh, I thought we have `CMAKE_BUILD_TYPE` exactly for this. Doesn't that variable then become useless? Isn't that a bit over-engineered?

MakisH:
Isn't it standard and expected that Release builds (of any software) do not have assertions? Don't they increase the computations? Is there any example out there that does the same?

fsimonis:
Enabling assertions can make release builds easier to debug.
Furthermore, the presence of assertions has side-effects that can influence the optimizer. Thus testing release+assertions vs release-assertions can produce different results. A tweakable assertion policy can expose such problems.

> Don't they increase the computations?

Yes. The choice is weather or not it is reasonable to invest computations into additional security.

> Isn't it standard and expected that Release builds (of any software) do not have assertions?

Yes, but for safety-critical applications like satellites, mars rovers, etc..
Disabling assertions requires confidence that your software is bug-free.
Enabling assertions allows to ensure that one does not waste resources once the program hits an unexpected internal state.

PR #984, opened 2021-03-24
Fix missing return type for non-void functions

DavidSCN:
gcc complains currently in two places in the following way:

``` bash
precice/src/cplscheme/config/CouplingSchemeConfiguration.cpp:981:1: warning: control reaches end of non-void function [-Wreturn-type]
```
i just added a dummy return type in order to fix it.
## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)


DavidSCN:
The failing test is unrelated.

DavidSCN:
If you are confused because you don't see this warning: it only appears in `Release` builds because the `PRECICE_ASSERT` there is empty. I guess the compiler is smart enough to deduce the unreachable return from `assert(false)` if debug messages are enabled.

fsimonis:
Hmm this has been bugging me for a long time too, but there doesn't seem to be a "good" expressive solution to this.

I once stumbled across `BOOST_UNREACHABLE_RETURN`, maybe that could be an expressive solution to this problem.
Could you check if this is what we need? As we already use boost, we could just use it.

DavidSCN:
> Hmm this has been bugging me for a long time too, but there doesn't seem to be a "good" expressive solution to this.

Looking at the cases here, I think all we can do is adding a comment and a nonsense return. Release mode builds are meant to be in production runs where these kind of errors shouldn't occur. Even if they occur, the user can still switch back to debug messages. In any case, I would prefer this solution over compiler warnings.

> I once stumbled across BOOST_UNREACHABLE_RETURN, maybe that could be an expressive solution to this problem.
Could you check if this is what we need? As we already use boost, we could just use it.

I saw it in the code as well and tried with the following result

```bash
~/precice/src/mesh/config/DataConfiguration.cpp:97:10: error: void value not ignored as it ought to be
   97 |   return BOOST_UNREACHABLE_RETURN(-1);
      |          ^~~~~~~~~~~~~~~~~~~~~~~~
```
Maybe I need to take a closer look, let me check again.

DavidSCN:
> I saw it in the code as well and tried with the following result

Well.. this happens if you don't read before using the macro. It works now with `BOOST_UNREACHABLE`

fsimonis:
Also interesting may be this answer https://stackoverflow.com/a/2201827

DavidSCN:
Ok, so what's the consensus now? Do you want me to change it again? I think `BOOST_UNREACHABLE` will result in something very similar in the end. Of course `std::abort` will remove the include header in these files.

fsimonis:
I don't think that there is a consensus yet. Let's dig a little bit deeper first. 

What happens if a `BOOST_UNREACHABLE_RETURN` is reached?
Will it just do nothing or will it abort?

Think of a future dev (maybe you) adding a new mapping for example. What happens the dev overlooks one switch/if and the tag is not handled in one case. Then this could reach such a statement.

A hard abort or error message would be great in such a case.

DavidSCN:
> What happens if a BOOST_UNREACHABLE_RETURN is reached?
Will it just do nothing or will it abort?

`BOOST_UNREACHABLE_RETURN` is defined as `__builtin_unreachable()` which is (I guess) compiler specific. It leads to a segfault and the program is aborted. I guess reaching the `__builtin_unreachable()` is just undefined behavior. There is obviously no error message.

DavidSCN:
The seg fault for the boost version here is rather odd. In my opinion we have here two options:

1. Return something ('dummy') in order to avoid the compiler warning
2. Add an Assert/abort with a message which is not removed in `Release mode`

fsimonis:
Let's just abort in this case.

In `assertion.hpp` independent  of `NDEBUG`:
```
/** Displays an error message and aborts the program.
 * Use to mark unreachable statements under switch or if blocks.
 */
#define PRECICE_UNREACHABLE(mess) { std::cerr << mess << std::endl; std::abort(); }
```

Abort is marked as `[[noreturn]]`so there should be no further messages popping up.

DavidSCN:
Ok, last question: should we make it directly a bit more general? Or go only for unreachable?
```cpp
/** Displays an error message and aborts the program.
 * Use to mark unreachable statements under switch or if blocks.
 */
#define PRECICE_ASSERT_ALWAYS(cond, mess) { if(!cond){ std::cerr << mess << std::endl; std::abort();} }
```

fsimonis:
Let's go for the unreachable first.

We should have a look at #663 to solve the "when do assertions work"-issue.

DavidSCN:
We should actually (?) use `fmtlib` instead of `cerr`, but that's another story.

fsimonis:
Sure, you can copy the invocation of fmtlib from the `PRECICE_ERROR` macro.
We still need the `std::err` and the `std::endl` (for the flush) though.

DavidSCN:
> Sure, you can copy the invocation of fmtlib from the `PRECICE_ERROR` macro.
> We still need the `std::err` and the `std::endl` (for the flush) though.

I would like to go without fmtlib here the at the moment and maybe introduce it later in in the `PRECICE_ASSERT` and `PRECICE_UNREACHABLE` macro. Ready for merge/review.

PR #985, opened 2021-03-26
Fix macOS runner tag

oguzziya:
## Main changes of this PR
Apparently, I have used the wrong tag to identify the runner. Probably it should be `OSX`. It would be good if it can be confirmed since I cannot see the runner tags.

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/.github/workflows/build-and-test-mac.yml
+++ b/.github/workflows/build-and-test-mac.yml
@@ -4,12 +4,13 @@ on:
     branches:
       - master
       - develop
+      - fix-macos-ci
   workflow_dispatch:
 
 jobs:
   build:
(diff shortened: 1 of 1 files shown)


oguzziya:
I think there is a connection problem in the runner. I cannot ping `github.com` which is [a requirement for the self-hosted runner.](https://docs.github.com/en/actions/hosting-your-own-runners/about-self-hosted-runners#communication-between-self-hosted-runners-and-github) In addition when I run `./config.sh --check`, internet connection tests fail.

fsimonis:
So this PR is not a solution.

PR #986, opened 2021-04-03
Extend mesh filtering considering connectivity elements

oguzziya:
## Main changes of this PR
Mesh filtering does not consider the connectivity elements. With this extension, vertices chosen for filtering extended by elements they are connected, therefore connectivity information is not lost.

This is a useful feature for connectivity-based mappings, and a must for scaled consistently mapping (other alternative is no filtering)

## Author's checklist
* [ ] Add tests

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/mesh/Filter.cpp
+++ b/src/mesh/Filter.cpp
@@ -1,0 +1,46 @@
+#pragma once
+
+#include "mesh/Filter.hpp"
+
+namespace precice {
+namespace mesh {
+
+void addVertexToMesh(const Vertex &vertex, boost::container::flat_map<int, Vertex *> &vertexMap, Mesh &destination)
+{
+  Vertex &v = destination.createVertex(vertex.getCoords());
+  v.setGlobalIndex(vertex.getGlobalIndex());
+  if (vertex.isTagged()) {
+    v.tag();
+  }
+  v.setOwner(vertex.isOwner());
+  vertexMap[vertex.getID()] = &v;
+}
+
+void addEdgeToMesh(const Edge &edge, boost::container::flat_map<int, Edge *> &edgeMap, boost::container::flat_map<int, Vertex *> &vertexMap, Mesh &destination)
+{
+  if (vertexMap.count(edge.vertex(0).getID()) == 0) {
+    addVertexToMesh(edge.vertex(0), vertexMap, destination);
+  }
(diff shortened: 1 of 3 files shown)


oguzziya:
Right now I think this works as intended and as described in #542. Connectivity aware filtering extension is only activated if at least one of the mappings is a scaled consistent mapping. It can be changed of course. Second test that I put is intentionally a `NearestNeighbor` mapping, it was easier to track filtering in general.

src/partition/ReceivedPartition.cpp
```diff
@@ -135,10 +135,15 @@ void ReceivedPartition::compute()
                   _mesh->getName());
   }
 
+  // If any of the mappings is a scaled-scaled consistent mapping, extend filtering considering connectivity
+  _filterWithConnectivity = std::any_of(_fromMappings.begin(), _fromMappings.end(), [](const mapping::PtrMapping m) { return m->hasConstraint(mapping::Mapping::SCALEDCONSISTENT); }) or
+                            std::any_of(_toMappings.begin(), _toMappings.end(), [](const mapping::PtrMapping m) { return m->hasConstraint(mapping::Mapping::SCALEDCONSISTENT); });
```
fsimonis, line 140: @uekerman Is it a good idea to always extend include connectivity in these cases?

src/partition/tests/ReceivedPartitionTest.cpp
```diff
@@ -514,7 +514,7 @@ BOOST_AUTO_TEST_CASE(RePartitionRBFLocal2D1)
     createNastinMesh2D(pNastinMesh, context.rank);
 
     double            safetyFactor = 20.0;
-    ReceivedPartition part(pSolidzMesh, ReceivedPartition::NO_FILTER, safetyFactor);
+    ReceivedPartition part(pSolidzMesh, ReceivedPartition::ON_MASTER, safetyFactor);
```
fsimonis, line 517: Why did you need to change this?

src/partition/ReceivedPartition.cpp
uekerman: What do you mean by "always" here?

fsimonis: When using scaled consistent mappings, the filter will always extend the selection using connectivity information.

Issue #623, opened 2020-01-27
Change default symbol visibility to hidden


fsimonis:
To reduce the symbol visibility to the actual API, we should change the default visibility to `hidden`.

This has some upsides:
1. We adapt the default on Windows platforms
2. We reduce binary size
3. We reduce shared object load time

Downsides are:
1. We have to explicitly mark API functions (CMake can generate export headers)
2. We cannot use the shared library for the unit tests anymore.  
   We need to add an object library target that compiles the sources.

References:
* [GCC wiki on visibility](https://gcc.gnu.org/wiki/Visibility)
* [CMake generate export header](https://cmake.org/cmake/help/v3.10/module/GenerateExportHeader.html)
* [CMake visibilty preset](https://cmake.org/cmake/help/v3.10/prop_tgt/LANG_VISIBILITY_PRESET.html)
* [CMake inline visibility preset](https://cmake.org/cmake/help/v3.10/prop_tgt/VISIBILITY_INLINES_HIDDEN.html)
* [CMake add_library(OBJECT)](https://cmake.org/cmake/help/v3.10/command/add_library.html?highlight=add_library#object-libraries)

Related to #200

PR #990, opened 2021-04-10
Add fmtlib for log formatting

fsimonis:
## Main changes of this PR

Add [fmtlib](fmt.dev) to simplify formatting of error messages.
Changed log macros to format all arguments directly to `fmt::format()`.
Ported all log calls to the new syntax.

```cpp
PRECICE_XXX(message, [arg0], [arg1])
PRECICE_CHECK(cond, message, [arg0], [arg1])

PRECICE_INFO("Hello");
PRECICE_DEBUG("The polynomial linear system of the RBF mapping from mesh {} to mesh {} converged. {}",
              input()->getName(), output()->getName(), _QRsolver.summaryFor(tau));

PRECICE_CHECK(fromMesh.get() != nullptr,
              "Mesh \"{0}\" was not found while creating a mapping. "
              "Please correct the from=\"{0}\" attribute.",
              fromMeshName);
```

Short term goals:
* Easier to format error messages in code.
* More readable error messages in code.
* Prevent duplicate arguments in the formatting using positional arguments `{0} x {0}`

Long term goal:
* Drop `prettyprint` dependency.  
  (Still required by assertions and some additional places)
* Simplify formatting in exporters
* Replace by `std::format` in the future

* Formatting error messages is currently a tedious process as messages have to be interleaved with `<<`.
* This results in messages that are difficult to format and difficult to read.
* `clang-format` attempts to align `<<` which makes formatting even more tedious. 

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)


CMakeLists.txt
```diff
@@ -188,6 +188,8 @@ else()
 endif()
 precice_validate_prettyprint()
 
+# libfmt
+add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/thirdparty/fmt)
```
DavidSCN, line 192: That's only for the libraray. You need to add an include directory for the binary and tests as well in order to compile
```cmake 
  include_directories(${CMAKE_CURRENT_SOURCE_DIR}/thirdparty/fmt/include)
```

DavidSCN (commented):
Very nice, this improves the readability of the error messages in the code and writing error messages itself by a lot. Still, my long-term goal is to introduce a column limit at some time. 
Just one more comment: wouldn't it be better to use a git submodule or something similar instead of a hard copy? The `fmt` code is frequently updated and this way we could simply pull the latest changes into our code base.

CMakeLists.txt
fsimonis: This processes the [`CMakeLists.txt`](https://github.com/precice/precice/blob/b4aa52aca2c086782d02badc1c60955ba5593cf4/thirdparty/fmt/CMakeLists.txt) in the `thirdparty/fmt` directory.

fsimonis:
@DavidSCN The boost logger should take care of the linebreak at the end of log statements.
Do you have an example for such an issue?

CMakeLists.txt
DavidSCN: Still, it didn't compile on system without adding `include_directories`. Have you enabled testing and the binary building?

DavidSCN:
There is a line break, but I was thinking of an additional empty line. It changed now due to the restructuring, but an OpenFOAM error looked from the user perspective like this:
```
---[precice] ERROR:  The given mesh name "Fluid-Meshs" is unknown to preCICE. Please check the mesh definitions in the configuration.
Looking for any time directories without results (e.g. stray functionObjectProperties files, see openfoam-adapter issue #26 on GitHub)...
```
so there is no clear separation. But thinking about it again, maybe we should keep it like this. (1) The restructuring prevents such occurrences and (2) an empty line might be confusing if nothing follows.

fsimonis:
> But thinking about it again, maybe we should keep it like this. (1) The restructuring prevents such occurrences and (2) an empty line might be confusing if nothing follows.

I agree, however you can experiment with adding `\n` in the  [message format](https://www.precice.org/configuration-logging.html#examples).

CMakeLists.txt
fsimonis: I tested it what is the error message?

The include directory is propagated by the target.

DavidSCN: Tested it now as well in a fresh build and it works for me as well. I previously rebuild an old build, but still resourced the root CMakeLists.txt in order to incorporate the new include directory. I have no idea why. Sorry for the mess.

MakisH (approved):
- I like this proposal: it can make complex formatting easier, but mostly in the long term.
- Header-only, so it should not change anything for the user.
- I understand the changes, but I have not checked every file.
- The license (see `thirdparty/fmt/include/fmt/format.h` or `thirdparty/fmt/LICENSE.rst`) seems to be permissive, I think BSD, so it should be fine with our LGPL).
- I have not tried it, I woud just trust the CI.

TODO: Add a changelog entry.

Questions:
- Is the library C++14? We need to compile it with the same compiler.
- Is there anything to consider regarding xSDK and symbols we export when using header-only libraries?

Let me know if you would like me to look deeper into anything.

fsimonis:
@MakisH 
> Is the library C++14? We need to compile it with the same compiler.

It is actually C++11. So we could have started to use this a long time ago :wink: 

> Is there anything to consider regarding xSDK and symbols we export when using header-only libraries?

Not AFAI, however, we already use Eigen in a header-only way. So there won't be new The solution to this is #623 btw.

Issue #489, opened 2019-08-21
Add Contiguous Mapping


fsimonis:
## Scenario

Two solvers share a mesh that is not only identical (same vertex order and positions), but the participants also own contiguous sections of it.
This is the case when solvers partition their mesh uniformly.

## Problem

Using nearest neighbour mapping is the best we can do, but it is not aware of the contiguity. It thus has to generate an index tree, probe it and build a mapping.
This is costly during initialization and adds an overhead at each time-step.

## Proposal

We add a `contiguous` mapping, which makes the following assumptions:
1. Vertices are in identical order on all instances (participant masters and slaves).
2. Master and slaves own a section of this total order.
3. The position of vertices are identical for both meshes.

preCICE stores the data as a contiguous block of memory in the vertex order.
We can thus reduce `computeMap` to the problem of finding the overlap of two ranges.
Furthermore for consistent mappings, `map` becomes the identity and simply has to copy a block of memory. Conservative mappings have to add up blocks of contiguous memory, making vectorizers very happy.

![contiguous-mapping](https://user-images.githubusercontent.com/13552216/63504382-baa38b80-c4d1-11e9-9aba-23f88dd86e56.png)

BenjaminRueth:
The main use case for the contiguous mapping is if there are two identical meshes (like in your drawing). There should be a check to make sure that the data provided to the contiguous mapping is as expected and point 3 in your proposal is satisfied: The vertices that map to each other actually have the same coordinates (I will call these pairs of vertices siblings). 

If we want to avoid mapping cost, we cannot find the sibling of a node automatically. Therefore we rely on the user to provide the nodes in contiguous order. In this case finding the siblings is possible without mapping. During initialization we should perform a check whether the siblings actually have the same coordinates to avoid wrong user input to the contiguous mapping:

```
for vertex_id in range(n_vertices):
    assert(coordinates_mesh_A[vertex_id_mesh_A] == coordinates_mesh_B[vertex_id])
```

uekerman:
I guess that the suggestion comes out of @Dominanz 's application, right? There, however, we would not only have matching meshes and matching global order of vertices, but also matching partitions, right? Is there a _real_ use case, where would have matching meshes, matching global order, but not matching partitions (like the situation you sketched above)?  
I don't see ASTE as a _real_ case here :wink: 

If we have matching partitions we can also get around the complete repartitioning, which is the real expensive part. Already a warning here: implementation-wise this sounds easier than it is. 

[...]

PR #991, opened 2021-04-12
Remove output zeroing before data mapping

DavidSCN:
We currently zero out the data vector on the output mesh before performing any mapping. I don't get the rationale behind this. I guess it is a secure action with no further purpose, but maybe I am overlooking something. However, we would like to be able to directly access data on the output mesh without performing a mapping in preCICE (solver mapping) and this doesn't work if we zero out the data vector before performing the mapping and sending the data (in case of the 'solver mapping' the `map` function would of course be a 'do nothing' mapping in preCICE, but we need the function call nevertheless for compatibility reasons).
All tests pass locally.

### DIFF
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -1004,7 +1004,6 @@ void SolverInterfaceImpl::mapWriteDataFrom(
         continue;
       }
 
-      context.toData->values() = Eigen::VectorXd::Zero(context.toData->values().size());
       PRECICE_DEBUG("Map data \"" << context.fromData->getName()
                                   << "\" from mesh \"" << context.mesh->getName() << "\"");
       PRECICE_ASSERT(mappingContext.mapping == context.mappingContext.mapping);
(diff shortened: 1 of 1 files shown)


BenjaminRodenberg:
> However, we would like to be able to directly access data on the output mesh without performing a mapping in preCICE

Is this related to #489? The "without performing a mapping" just triggered my attention.

DavidSCN:
> > However, we would like to be able to directly access data on the output mesh without performing a mapping in preCICE
> 
> Is this related to #489? The "without performing a mapping" just triggered my attention.

Could be an application case, I guess, but the described mapping in #489 is a different one and we will (probably) not completely remove the overhead since we don't want to assume any contiguous indices.

uekerman:
Some data mapping methods add up values, in particular conservative schemes (for NP also the consistent variant).
Sth similar happens in the communication:

https://github.com/precice/precice/blob/6d27ad5db42997bbc06ff97b39eb4999196295b7/src/m2n/PointToPointCommunication.cpp#L627

There is also:

https://github.com/precice/precice/blob/6d27ad5db42997bbc06ff97b39eb4999196295b7/src/precice/impl/SolverInterfaceImpl.cpp#L444

I would not trust the "tests pass" too much. We should try to write a test that really pushes this to the edge.

fsimonis:
@DavidSCN see #760 for a very similar discussion between me and @uekerman  :wink:

DavidSCN:
> Some data mapping methods add up values, in particular conservative schemes (for NP also the consistent variant).

In my opinion we should then zero out the values in each individual place before summing them up. It took my quite some time to find this LOCs.

> Sth similar happens in the communication:

How is it related to the actual mapping?

DavidSCN:
I still believe that setting this up properly is worth the effort. However, after discussing with @uekerman it won't be merged right now, since @BenjaminRodenberg's WR will probably refactor these places anyway.

PR #992, opened 2021-04-12
Merge read/write function calls in SolverInterfaceImpl

DavidSCN:
## Main changes of this PR
This PR merges the implementation of our `read(Block)Scalar/VectorData` and `write(Block)Scalar/VectorData` into a single implementation for each task. 

We have quite some duplication in all these functions and changing the mapping might require changes in these functions. Changing it in all places is tedious. This PR would centralize the implementation.

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -1052,35 +1052,86 @@ void SolverInterfaceImpl::mapReadDataTo(
   performDataActions({action::Action::READ_MAPPING_POST}, time, 0, 0, 0);
 }
 
-void SolverInterfaceImpl::writeBlockVectorData(
-    int           dataID,
-    int           size,
-    const int *   valueIndices,
-    const double *values)
+template <int dataDimension>
+void SolverInterfaceImpl::readBlockVectorDataImpl(const int  dataID,
+                                                  const int  size,
+                                                  const int *valueIndices,
+                                                  double *   values) const
+
 {
   PRECICE_TRACE(dataID, size);
-  PRECICE_CHECK(_state != State::Finalized, "writeBlockVectorData(...) cannot be called after finalize().");
-  PRECICE_VALIDATE_DATA_ID(dataID);
+
+  // Return if nothing to read
   if (size == 0)
     return;
+
+  // Validate user input
+  PRECICE_CHECK(_state != State::Finalized, "readBlockVectorData(...) cannot be called after finalize().");
+  PRECICE_VALIDATE_DATA_ID(dataID);
+  PRECICE_REQUIRE_DATA_READ(dataID);
+
+  DataContext &context = _accessor->dataContext(dataID);
+  PRECICE_ASSERT(context.toData != nullptr);
+  mesh::Data &data = *context.toData;
+  PRECICE_CHECK(data.getDimensions() == dataDimension,
+                "Your specified data type dimension of the data \"" << data.getName() << "\" does not match the data dimension you used for reading data in your code."
+                                                                                         "This means you either use 'readVectorData' on a scalar data set or 'readScalarData' on a vector-valued data set.");
   PRECICE_ASSERT(valueIndices != nullptr);
   PRECICE_ASSERT(values != nullptr);
+
+  auto &     valuesInternal = data.values();
+  const auto vertexCount    = valuesInternal.size() / dataDimension;
+  for (int i = 0; i < size; i++) {
+
+    const auto valueIndex     = valueIndices[i];
+    const int  offsetInternal = valueIndex * dataDimension;
+    const int  offset         = i * dataDimension;
+
+    PRECICE_CHECK(0 <= valueIndex && valueIndex < vertexCount, "Cannot read data \"" << data.getName() << "\" to invalid Vertex ID (" << valueIndex << "). Please make sure you only use the results from calls to setMeshVertex/Vertices().");
+    for (int dim = 0; dim < dataDimension; dim++) {
+      values[offset + dim] = valuesInternal[offsetInternal + dim];
+    }
+  }
+}
+
+template <int dataDimension>
+void SolverInterfaceImpl::writeBlockVectorDataImpl(const int     dataID,
+                                                   const int     size,
+                                                   const int *   valueIndices,
+                                                   const double *values)
+{
+  PRECICE_TRACE(dataID, size);
+
+  // Return if nothing to write
+  if (size == 0)
+    return;
+
+  // Validate user input
+  PRECICE_CHECK(_state != State::Finalized, "writeData(...) cannot be called after finalize().");
+  PRECICE_VALIDATE_DATA_ID(dataID);
   PRECICE_REQUIRE_DATA_WRITE(dataID);
+
   DataContext &context = _accessor->dataContext(dataID);
   PRECICE_ASSERT(context.fromData != nullptr);
   mesh::Data &data = *context.fromData;
-  PRECICE_CHECK(data.getDimensions() == _dimensions,
-                "You cannot call writeBlockVectorData on the scalar data type \"" << data.getName()
-                                                                                  << "\". Use writeBlockScalarData or change the data type for \""
-                                                                                  << data.getName() << "\" to vector.");
+  PRECICE_CHECK(data.getDimensions() == dataDimension,
+                "Your specified data type dimension of the data \"" << data.getName() << "\" does not match the data dimension you used for writing data in your code."
+                                                                                         "This means you either use 'writeVectorData' on a scalar data set or 'writeScalarData' on a vector-valued data set.");
+  // Assert input pointer
+  PRECICE_ASSERT(valueIndices != nullptr);
+  PRECICE_ASSERT(values != nullptr);
+
   auto &     valuesInternal = data.values();
-  const auto vertexCount    = valuesInternal.size() / data.getDimensions();
+  const auto vertexCount    = valuesInternal.size() / dataDimension;
+  // Loop over provided size and assign values
   for (int i = 0; i < size; i++) {
-    const auto valueIndex = valueIndices[i];
+
+    const auto valueIndex     = valueIndices[i];
+    const int  offsetInternal = valueIndex * dataDimension;
+    const int  offset         = i * dataDimension;
+
     PRECICE_CHECK(0 <= valueIndex && valueIndex < vertexCount, "Cannot write data \"" << data.getName() << "\" to invalid Vertex ID (" << valueIndex << "). Please make sure you only use the results from calls to setMeshVertex/Vertices().");
-    int offsetInternal = valueIndex * _dimensions;
-    int offset         = i * _dimensions;
-    for (int dim = 0; dim < _dimensions; dim++) {
+    for (int dim = 0; dim < dataDimension; dim++) {
       PRECICE_ASSERT(offset + dim < valuesInternal.size(),
                      offset + dim, valuesInternal.size());
       valuesInternal[offsetInternal + dim] = values[offset + dim];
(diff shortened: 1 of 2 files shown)


src/precice/impl/SolverInterfaceImpl.hpp
```diff
@@ -634,6 +634,40 @@ class SolverInterfaceImpl {
   /// Computes, performs, and resets all suitable read mappings.
   void mapReadData();
 
+  /**
+   * @brief Generich implementation in order to write data
```
uekerman, line 638:
```suggestion
   * @brief Generic implementation in order to write data
```

src/precice/impl/SolverInterfaceImpl.hpp
```diff
@@ -634,6 +634,40 @@ class SolverInterfaceImpl {
...
+   * @param[in] values pointer to the data to write
+   */
+  template <int dataDimension>
+  void writeBlockVectorDataImpl(const int     dataID,
+                                const int     size,
+                                const int *   valueIndices,
+                                const double *values);
+
+  /**
+   * @brief Generich implementation in order to read data
```
uekerman, line 655:
```suggestion
   * @brief Generic implementation in order to read data
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1238,54 +1212,15 @@ void SolverInterfaceImpl::readBlockScalarData(
...
-                "You cannot call readBlockScalarData on the vector data type \"" << data.getName()
-                                                                                 << "\". Use readBlockVectorData or change the data type for \"" << data.getName() << "\" to scalar.");
-  auto &     valuesInternal = data.values();
-  const auto vertexCount    = valuesInternal.size();
-  for (int i = 0; i < size; i++) {
-    const auto valueIndex = valueIndices[i];
-    PRECICE_CHECK(0 <= valueIndex && valueIndex < vertexCount, "Cannot read data \"" << data.getName() << "\" to invalid Vertex ID (" << valueIndex << "). Please make sure you only use the results from calls to setMeshVertex/Vertices().");
-    values[i] = valuesInternal[valueIndex];
-  }
+  readBlockVectorDataImpl<1>(dataID, size, valueIndices, values);
```
uekerman, line 1215: Why not simplifying the name to `readBlockDataImpl`?

uekerman (commented):
I like the idea. Not sure whether we really need the templating here? @fsimonis opinion?

src/precice/impl/SolverInterfaceImpl.cpp
fsimonis: We don't really need dimensions here. What's important is weather the user called the scalar or vector variant.

fsimonis:
Good idea. :+1:
There are two issues with this though one major and one minor.

Error messages do not display the interface function and the original function call cannot be deduced any more:
* The user calls `SolverInterface::readBlockScalarData`
* **Old behaviour:** Error message location in the log is `SolverInterfaceImpl::readBlockScalarData`
 **New behaviour:** Error message location in the log is `SolverInterfaceImpl::readBlock[Vector]DataImpl`, which will be the same for all callers.

Thus we have to add the name of the interface function to each error message, meaning we have to somehow generate these error messages on the fly.

The minor issue is that additional information required in the `readBlock[Vector]DataImpl` is weather the user called the vector or scalar version.
There is no benefit in lifting this argument into the type system by making it a template argument and passing this at runtime suffices.
So, an `enum struct DataAccessKind : bool { Vector, Scalar };` could be passed as an additional argument instead:

```cpp
void SolverInterfaceImpl::readBlock[Vector]DataImpl(const int  dataID,
                                                  const int  size,
                                                  const int *valueIndices,
                                                  double *   values,
                                                  DataAccessKind access
                                                  ) const;
```

DavidSCN:
>     * **Old behaviour:** Error message location in the log is `SolverInterfaceImpl::readBlockScalarData`
>       **New behaviour:** Error message location in the log is `SolverInterfaceImpl::readBlock[Vector]DataImpl`, which will be the same for all callers.

in the current version we could add an Assertion entry in the `readBlock..` function which would reveal the actual location. Still, I guess we can provide a descriptive enough error message in the `...Impl` function. I would not see it as a major issue right now.

> There is no benefit in lifting this argument into the type system by making it a template argument and passing this at runtime suffices.
So, an enum struct DataAccessKind : bool { Vector, Scalar }; could be passed as an additional argument instead:

_mhm_ The purpose of this PR was to unify both functions. If you don't like templating here (also comment of @uekerman ) we could simply gather the dimension information from the accessor without any `enum` . The main idea behind the template was to let the compiler remove the dimension loop in case of a scalar data types (should result in a very similar machine code as in the current implementation). Nice side-effect would be an unrolled loop for the vector-valued data type.

fsimonis (changes requested):
> in the current version we could add an Assertion entry in the readBlock.. function which would reveal the actual location. Still, I guess we can provide a descriptive enough error message in the ...Impl function. I would not see it as a major issue right now.

Which API call results in an error is vital information for both developers and users to debug and help with issues. This information used to be present in both the error location and the error message.
Given the proposed changes, we loose both these hints and there is no easy way to find out which API call leads to the error.

I see two ways forward:
1. We move the input sanitization back to the main functions and refactor out the processing of the input.
2. Add all missing hints on what API was called to the error messages in the refactored function.  
  Dynamically assembling these messages may however turn out less readable than having the full messages directly at the interface.

I tend towards 1. Let's keep user sanitization as close as possible to the API.

> The main idea behind the template was to let the compiler remove the dimension loop in case of a scalar data types (should result in a very similar machine code as in the current implementation). Nice side-effect would be an unrolled loop for the vector-valued data type.

Claiming performance improvements without measurements is dangerous game. Note that this does not simply copy memory. We enforce no order on the passed indices, hence this may as well be random write access to the memory. If your aim is to decrease the runtime cost of these functions, then we need suitable test setups, measurements of the baselines, how significant the runtime is in relation to other calls and a comparison to the modified version.

DavidSCN:
> Which API call results in an error is vital information for both developers and users to debug and help with issues. This information used to be present in both the error location and the error message.

Let me add here, that the actual API function call is still visible in the stacktrace. 

> Given the proposed changes, we loose both these hints and there is no easy way to find out which API call leads to the error.
> 
> I see two ways forward:
> 
>     1. We move the input sanitization back to the main functions and refactor out the processing of the input.
> 
>     2. Add all missing hints on what API was called to the error messages in the refactored function.
>        Dynamically assembling these messages may however turn out less readable than having the full messages directly at the interface.
> 
> 
> I tend towards 1. Let's keep user sanitization as close as possible to the API.
> 

The downside here is again the vast duplication of the input arguments, even it is only about Asserts.


> Claiming performance improvements without measurements is dangerous game. Note that this does not simply copy memory. We enforce no order on the passed indices, hence this may as well be random write access to the memory. If your aim is to decrease the runtime cost of these functions, then we need suitable test setups, measurements of the baselines, how significant the runtime is in relation to other calls and a comparison to the modified version.

I agree. Let me try to underline this by an actual measurement. Do you have any test setup in this direction? I guess I need to set up something anyway..

fsimonis:
@DavidSCN Regarding the profiling: You don't even need a second participant for this.
You can just use `setMeshVertices` to create a mesh and then write data to the mesh without calling `initialize`.

DavidSCN:
That's a great idea :+1:

DavidSCN:
We have a speed-up of 2-3 for the `Block` functions, and around 10% for the non-Block functions on a data set consisting of 10k points.

```
Performed: 
10000 iterations using templates.

+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |      4.36s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| writeBlockScalarData            |     10000 |     0.116s |       2.7% |
| writeScalarData                 |     10000 |      4.15s |        95% |
+---------------------------------+-----------+------------+------------+
+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |       5.3s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| writeBlockVectorData            |     10000 |     0.189s |       3.6% |
| writeVectorData                 |     10000 |      5.02s |        95% |
+---------------------------------+-----------+------------+------------+


Performed: 
10000 iterations without templates.

+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |      4.97s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| writeBlockScalarData            |     10000 |     0.311s |       6.3% |
| writeScalarData                 |     10000 |      4.56s |        92% |
+---------------------------------+-----------+------------+------------+

+---------------------------------------------+------------+------------+
| Total wallclock time elapsed since start    |      5.09s |            |
|                                             |            |            |
| Section                         | no. calls |  wall time | % of total |
+---------------------------------+-----------+------------+------------+
| writeBlockVectorData            |     10000 |     0.337s |       6.6% |
| writeVectorData                 |     10000 |      4.66s |        91% |
+---------------------------------+-----------+------------+------------+


```

src/precice/impl/SolverInterfaceImpl.cpp
DavidSCN: I fixed it

DavidSCN:
Ok, after various discussions with @fsimonis I have no measured using `googlebench`. For convenience, I attached the [test.zip](https://github.com/precice/precice/files/6365987/test.zip)

We have now four test cases. Baseline is always the templated version. Only drawback is now case one. Please decide now for any version.

```
1. writeVectorData, type ordered access
Benchmark                        Time             CPU      Time Old      Time New       CPU Old       CPU New
-------------------------------------------------------------------------------------------------------------
BM_writeVector                -0.0682         -0.0681      33811511      31506782      33785372      31483785

2. writeBlockVectorData, type ordered access
Benchmark                        Time             CPU      Time Old      Time New       CPU Old       CPU New
-------------------------------------------------------------------------------------------------------------
BM_writeVector                +0.1246         +0.1244       3220585       3621931       3217952       3618223

3. writeVectorData, type random access
Benchmark                        Time             CPU      Time Old      Time New       CPU Old       CPU New
-------------------------------------------------------------------------------------------------------------
BM_writeVector                +0.4829         +0.4834      68329482     101328092      68250278     101239263

4. writeBlockVectorData, type random access
Benchmark                        Time             CPU      Time Old      Time New       CPU Old       CPU New
-------------------------------------------------------------------------------------------------------------
BM_writeVector                +0.0704         +0.0706      13563452      14517687      13547831      14503842
```

fsimonis:
The profiling is a bit incomplete. To gez the full picture, we need measurements for the baseline (develop) and your two proposed functions (templated and non-templated) for scalar and both 2d and 3d vectorial data versions for sequential and random vertex indices.
This however becomes too much for this PR.

My suggestion to continue with this:
* Solve the logging ambiguity issues of the PR. It needs to be absolutely clear where the call comes from in the logs.
* Implement this as a simple function with runtime parameters only. This allows us to implement the interfaces functions as one-liners. 

This resolves the code duplication.

Once this is done, we can revisit performance aspects of the implementation in a different PR.
These functions get called at least once per time window, hence it is worth looking into them.

DavidSCN:
> The profiling is a bit incomplete. To get the full picture, we need measurements for the baseline (develop) and your two proposed functions (templated and non-templated) for scalar and both 2d and 3d vectorial data versions for sequential and random vertex indices.

What is the purpose of comparing it against our current implementation? This PR is not about speed-up, but code duplication. The current profiling took already quite some time, since creating multiple instances of our solverinterface is tedious in the benchmark. Considering that the user input checks are about half of the code in each implementation

https://github.com/precice/precice/blob/34688da733169cdbea87fd9c66dd74362496f9ad/src/precice/impl/SolverInterfaceImpl.cpp#L1190-L1204

the duplication reduction is anyway very little (see discussion above). The effort for the suggested changes goes just too far, then let's keep the duplication as it is.

fsimonis:
> What is the purpose of comparing it against our current implementation?

To make sure that there is no performance regression introduced by the refactoring.

> the duplication reduction is anyway very little (see discussion above).

The duplication in the SolverinterfaceImpl is a real issue. We have 4 write functions and 4 read functions that essentially all run the same checks.
Unifying some of these will definitely result in a maintainability benefit. The big question is how to accomplish this without loss of information.
As mentioned before, the checks can be in both the interface and the generalised function. All that matters is that messages contain all the information to identify the caller.

I know that working on the interface can be frustrating and annoying, but the investment is worth it in the long run. Improving the "User Experience" directly benefits us too, as preCICE developers are also users.

PR #1001, opened 2021-04-23
Add Keys for Version and Revision

BenjaminRodenberg:

This changes the string returned by https://github.com/precice/precice/blob/34688da733169cdbea87fd9c66dd74362496f9ad/src/precice/SolverInterface.hpp#L800

from 

```
'2.2.0;v2.2.0-dirty;PRECICE_MPICommunication=Y;PRECICE_PETScMapping=Y;PRECICE_PythonActions=N;PRECICE_ENABLE_C=Y;PRECICE_ENABLE_FORTRAN=Y;CXX=GNU;CXXFLAGS= -g;LDFAGS='
```

to

```
'PRECICE_version=2.2.0;PRECICE_revision=v2.2.0-dirty;PRECICE_MPICommunication=Y;PRECICE_PETScMapping=Y;PRECICE_PythonActions=N;PRECICE_ENABLE_C=Y;PRECICE_ENABLE_FORTRAN=Y;CXX=GNU;CXXFLAGS= -g;LDFAGS='
```

This makes it easier to reformat the string in a key-value fashion. Here is a short example in python:

```
python3 -c "import precice; print({item.split('=')[0]:item.split('=')[1] for item in str(precice.get_version_information()).split(';')})"
```

`item.split('=')[1]` currently breaks, since there is no `=` in the first two items of the list separated by `;`.

## Author's checklist

## Reviewers' checklist

* [x] (more questions/tasks)

### DIFF
--- a/docs/changelog/1001.md
+++ b/docs/changelog/1001.md
@@ -1,0 +1,1 @@
+- Changed output of `SolverInterface::getVersionInformation()` to return a list of key-value pairs separated by a `;`.
--- a/src/precice/impl/versions.cpp.in
+++ b/src/precice/impl/versions.cpp.in
@@ -1,4 +1,4 @@
 #include "precice/impl/versions.hpp"
 
 const char * const precice::preciceRevision = "@preCICE_REVISION@";
-const char * const precice::versionInformation = "@preCICE_VERSION@;@preCICE_REVISION@;@preCICE_VERSION_INFORMATION@";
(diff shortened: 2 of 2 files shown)


BenjaminRodenberg:
Since this is a breaking change and also a low-gain feature, I and @fsimonis decided to close this PR. We can still reopen it, if at some point in time it becomes relevant again.

A good workaround for the python use case is, for example:

```
python3 -c "import precice; print({item.split('=')[0]:item.split('=')[1] for item in str(precice.get_version_information()).split(';')[2:]})"
```

In this case we loose the version, but we can just call `str(precice.get_version_information()).split(';')[0]`, if we are interested in the version and the statement from above in other cases.

PR #1004, opened 2021-04-26
Package time

BenjaminRodenberg:
## Main changes of this PR

This PR will move all functionality related to time windows and time steps into an independent package. This can be seen as a follow-up PR to #876.

This PR belongs to the bigger goal of implementing waveform relaxation in preCICE. See #133 

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)


src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -495,7 +490,7 @@ void BaseQNAcceleration::splitCouplingData(
     int   size       = cplData[id]->values().size();
     auto &valuesPart = cplData[id]->values();
     //Eigen::VectorXd& oldValuesPart = cplData[id]->oldValues.col(0);
-    cplData[id]->oldValues.col(0) = _oldValues.segment(offset, size); /// @todo: check if this is correct
+    cplData[id]->writeLastIteration(_oldValues.segment(offset, size)); /// @todo: check if this is correct
```
BenjaminRodenberg, line 493: @uekerman This is now actually the only place, where `CouplingData::writeLastIteration` is called. Everywhere else we only use `CouplingData::storeIteration` to shift the old data to the buffer.

src/cplscheme/BiCouplingScheme.hpp
```diff
@@ -96,6 +97,9 @@ class BiCouplingScheme : public BaseCouplingScheme {
   /// Returns all data to be received with data ID as given.
   CouplingData *getReceiveData(int dataID);
 
+  /// Returns waveform from given data ID.
+  time::Waveform *getWaveform(int dataID);
```
BenjaminRodenberg: I would like to get rid of this function.

src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp
```diff
@@ -450,69 +450,64 @@ BOOST_AUTO_TEST_CASE(testExtrapolateData)
...
+  scheme.setupDataMatrices();
   CouplingData *cplData = scheme.getSendData(dataID);
   BOOST_CHECK(cplData); // no nullptr
   BOOST_TEST(cplData->values().size() == 1);
-  BOOST_TEST(cplData->oldValues.cols() == 2);
-  BOOST_TEST(cplData->oldValues.rows() == 1);
+  BOOST_TEST(cplData->lastIteration().size() == 1);
   BOOST_TEST(testing::equals(cplData->values()(0), 0.0));
-  BOOST_TEST(testing::equals(cplData->oldValues(0, 0), 0.0));
-  BOOST_TEST(testing::equals(cplData->oldValues(0, 1), 0.0));
```
BenjaminRodenberg, line 461: We now test the `oldValues` in the corresponding unit test for the waveform and just look at the result of the extrapolation that the cplscheme has to care about.

BenjaminRodenberg (commented):
This PR still includes some open points. But from the general structure I think it's ready for a first review.

src/acceleration/BaseQNAcceleration.cpp
BenjaminRodenberg: We removed this part, since it is not needed.

src/cplscheme/BiCouplingScheme.hpp
BenjaminRodenberg: Done in 0ad83918a781c2e1e17f90eec46ed4e073218add.

src/cplscheme/BiCouplingScheme.cpp
```diff
@@ -120,5 +120,16 @@ CouplingData *BiCouplingScheme::getReceiveData(
...
 
+time::Waveform *BiCouplingScheme::getWaveform(
+    int dataID)
+{
+  PRECICE_TRACE(dataID);
+  WaveformMap::iterator iter = _waveforms.find(dataID);
+  if (iter != _waveforms.end()) {
+    return &(*(iter->second));
+  }
+  return nullptr;
```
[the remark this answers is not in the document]
BenjaminRodenberg: This function was removed in 0ad83918a781c2e1e17f90eec46ed4e073218add, but question is still valid for other functions, such as [`getSendData`or `getReceiveData`](https://github.com/BenjaminRodenberg/precice/blob/1eb4facfa8985dcf3d6892d4e2313bdcc573c0a7/src/cplscheme/BiCouplingScheme.cpp#L101-L121).

src/acceleration/AitkenAcceleration.cpp
```diff
@@ -47,14 +47,9 @@ void AitkenAcceleration::initialize(DataMap &cplData)
...
-  for (DataMap::value_type &pair : cplData) {
-    int cols = pair.second->oldValues.cols();
-    if (cols < 1) {
-      PRECICE_ASSERT(pair.second->values().size() > 0, pair.first);
-      utils::append(pair.second->oldValues,
-                    (Eigen::VectorXd) Eigen::VectorXd::Zero(pair.second->values().size()));
-    }
+  for (DataMap::value_type &pair : cplData) { // @todo: seems to be unnecessary, since this is again just a loop over CouplingScheme::getAccelerationData
+    PRECICE_ASSERT(pair.second->values().size() > 0, pair.first);
+    pair.second->storeIteration();
```
uekerman, line 52: Coupling data and acceleration data can be different, see https://precice.org/configuration-acceleration.html
If it is really not needed, let's remove it. Will only get more complicated if we don't solve it now.

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -145,7 +145,7 @@ void BaseQNAcceleration::initialize(
     for (size_t i = 0; i < _dimOffsets.size() - 1; i++) {
       int accumulatedNumberOfUnknowns = 0;
       for (auto &elem : _dataIDs) {
-        auto &offsets = cplData[elem]->mesh->getVertexOffsets();
+        const auto &offsets = cplData[elem]->getVertexOffsets();
```
uekerman, line 148: How is this change (the mesh, not the const) related to the PR?

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -174,13 +174,8 @@ void BaseQNAcceleration::initialize(
...
   }
 
-  // Append old value columns, if not done outside of acceleration already
-  for (DataMap::value_type &pair : cplData) {
-    int cols = pair.second->oldValues.cols();
-    if (cols < 1) { // Add only, if not already done
-      //PRECICE_ASSERT(pair.second->values().size() > 0, pair.first);
-      utils::append(pair.second->oldValues, (Eigen::VectorXd) Eigen::VectorXd::Zero(pair.second->values().size()));
-    }
+  for (DataMap::value_type &pair : cplData) { // @todo: seems to be unnecessary, since this is again just a loop over CouplingScheme::getAccelerationData
```
uekerman, line 177: see above

src/acceleration/ConstantRelaxationAcceleration.cpp
```diff
@@ -30,17 +30,8 @@ void ConstantRelaxationAcceleration::initialize(DataMap &cplData)
...
-  for (auto &elem : _dataIDs) {
-    entries += cplData[elem]->values().size();
-  }
-  for (DataMap::value_type &pair : cplData) {
-    int cols = pair.second->oldValues.cols();
-    if (cols < 1) {
-      PRECICE_ASSERT(pair.second->values().size() > 0, pair.first);
-      utils::append(pair.second->oldValues, (Eigen::VectorXd) Eigen::VectorXd::Zero(pair.second->values().size()));
-    }
+  for (DataMap::value_type &pair : cplData) { // @todo: seems to be unnecessary, since this is again just a loop over CouplingScheme::getAccelerationData
```
uekerman, line 33: see above

src/acceleration/IQNILSAcceleration.cpp
```diff
@@ -121,9 +121,10 @@ void IQNILSAcceleration::computeUnderrelaxationSecondaryData(
     Eigen::VectorXd &values = data->values();
     values *= _initialRelaxation; // new * omg
     Eigen::VectorXd &secResiduals = _secondaryResiduals[id];
-    secResiduals                  = data->oldValues.col(0); // old
-    secResiduals *= 1.0 - _initialRelaxation;               // (1-omg) * old
-    values += secResiduals;                                 // (1-omg) * old + new * omg
+    const auto       oldValues    = data->lastIteration();
+    secResiduals                  = oldValues; // old
+    secResiduals *= 1.0 - _initialRelaxation;  // (1-omg) * old
+    values += secResiduals;                    // (1-omg) * old + new * omg
```
uekerman, line 127: The new style seems a bit unnecessarily complex, right?

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -76,7 +77,7 @@ void BaseCouplingScheme::sendData(m2n::PtrM2N m2n, DataMap sendData)
   for (const DataMap::value_type &pair : sendData) {
     int size = pair.second->values().size();
     if (size > 0) {
-      m2n->send(pair.second->values().data(), size, pair.second->mesh->getID(), pair.second->getDimensions());
+      m2n->send(pair.second->values().data(), size, pair.second->getMeshID(), pair.second->getDimensions());
```
uekerman, line 80: How did this end up here?

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -92,22 +93,13 @@ void BaseCouplingScheme::receiveData(m2n::PtrM2N m2n, DataMap receiveData)
   for (DataMap::value_type &pair : receiveData) {
     int size = pair.second->values().size();
     if (size > 0) {
-      m2n->receive(pair.second->values().data(), size, pair.second->mesh->getID(), pair.second->getDimensions());
+      m2n->receive(pair.second->values().data(), size, pair.second->getMeshID(), pair.second->getDimensions());
```
uekerman, line 96: How did this end up here?

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -536,9 +500,8 @@ bool BaseCouplingScheme::measureConvergence()
 
     PRECICE_ASSERT(convMeasure.couplingData != nullptr);
     PRECICE_ASSERT(convMeasure.measure.get() != nullptr);
-    const auto &oldValues = convMeasure.couplingData->oldValues.col(0);
 
-    convMeasure.measure->measure(oldValues, convMeasure.couplingData->values());
+    convMeasure.measure->measure(convMeasure.couplingData->lastIteration(), convMeasure.couplingData->values());
```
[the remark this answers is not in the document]
uekerman: Don't like this too much. Would push complexity into ConvergenceMeasure and make those classes harder to read. BaseCouplingScheme is the better place to handle this complexity in my opinion.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -674,14 +637,15 @@ bool BaseCouplingScheme::accelerate()
     getAcceleration()->performAcceleration(getAccelerationData());
   }
 
-  // Store data for conv. measurement, acceleration, or extrapolation
-  storeData();
-
   // extrapolate new input data for the solver evaluation in time.
   if (convergence && (_extrapolationOrder > 0)) {
-    extrapolateData(getAccelerationData());
+    storeWindowData();
```
uekerman, line 642: Why `storeWindowData` only when extrapolation is used? We will need it for more later and should already test it now moving forward.

src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -254,18 +254,18 @@ class BaseCouplingScheme : public CouplingScheme {
   /// Map that links DataID to CouplingData
   typedef std::map<int, PtrCouplingData> DataMap;
 
+  /// Map that links DataID to Waveform
+  typedef std::map<int, time::PtrWaveform> WaveformMap;
+
+  /// All waveforms this coupling scheme needs data as a map "data ID -> waveform"
+  WaveformMap _waveforms;
```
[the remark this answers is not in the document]
uekerman: OK for me

src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -403,11 +397,27 @@ class BaseCouplingScheme : public CouplingScheme {
...
    * @brief Extrapolate coupling data from values of previous time windows
-   * @param data Data fields to extrapolate
    */
-  void extrapolateData(DataMap &data);
+  void extrapolateData();
+
+  /**
+   * @brief used for storing send/receive data at end of acceleration, if not converged.
+   */
+  void storeLastIteration() // @todo: Try to make this private. This method is currently, only used for sake of testing.
```
uekerman, line 413: Why only for testing? You call it in `accelerate`.

src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -403,11 +397,27 @@ class BaseCouplingScheme : public CouplingScheme {
...
+  {
+    PRECICE_ASSERT(isImplicitCouplingScheme());
+    for (DataMap::value_type &pair : getAccelerationData()) {
+      // PRECICE_ASSERT(pair.second->lastIteration.size() > 0);  // @todo: Why is this assertion failing?
```
uekerman, line 417: TODO

src/cplscheme/CouplingData.hpp
```diff
@@ -25,44 +46,52 @@ struct CouplingData { // @todo: should be a class from a design standpoint. See
...
+  const Eigen::VectorXd lastIteration() const
+  {
+    return _lastIteration;
+  }
 
-  mesh::PtrMesh mesh;
+  [[deprecated("lastIteration should be read only!")]] void writeLastIteration(Eigen::VectorXd value)
+  {
+    _lastIteration = value;
+  }
```
uekerman, line 62: That's not meant to be merged, right?

src/cplscheme/CouplingData.hpp
```diff
@@ -25,44 +46,52 @@ struct CouplingData { // @todo: should be a class from a design standpoint. See
...
 
-  int getDimensions()
+  std::vector<int> getVertexOffsets()
   {
-    PRECICE_ASSERT(data != nullptr);
-    return data->getDimensions();
+    return mesh->getVertexOffsets();
   }
 
+  ///  True, if the data values if this CouplingData requires to be initialized by a participant.
```
uekerman, line 74:
```suggestion
  ///  True, if the data values of this CouplingData require to be initialized by a participant.
```
?
"a" participant?

src/cplscheme/CouplingData.hpp
```diff
@@ -25,44 +46,52 @@ struct CouplingData { // @todo: should be a class from a design standpoint. See
...
-      : data(data),
-        mesh(mesh),
-        requiresInitialization(requiresInitialization)
-  {
-    PRECICE_ASSERT(data != nullptr);
-    PRECICE_ASSERT(mesh != nullptr);
-    PRECICE_ASSERT(mesh.use_count() > 0);
-  }
+  /// Data values of previous iteration.
+  Eigen::VectorXd _lastIteration;
```
uekerman, line 90: I like `_previousIteration` actually better.

src/cplscheme/SharedPointer.hpp
```diff
@@ -2,12 +2,21 @@
...
 #include <memory>
 
+namespace precice {
+namespace time {
+
+class Waveform;
+using PtrWaveform = std::shared_ptr<Waveform>;
+
+} // namespace time
+} // namespace precice
```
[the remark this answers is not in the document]
uekerman: Should normally go into `time/SharedPointer`.

src/time/Waveform.hpp
```diff
@@ -0,0 +1,53 @@
+#pragma once
+
+#include <Eigen/Core>
+#include "utils/EigenHelperFunctions.hpp"
+
+namespace precice {
+namespace time {
+
+class Waveform {
```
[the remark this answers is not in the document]
uekerman: OK for the moment. In the long run, I guess `Time` could be better. I can imagine that you want to give the name "waveform" to sth else in future extensions.
Also, `Time` would be most consistent to how we treat other packages, say `Mapping`, `M2N`, `Communication`, ...

src/time/Waveform.hpp
```diff
@@ -0,0 +1,53 @@
...
+public:
+  void addNewWindowData(Eigen::VectorXd data)
```
uekerman, line 11: I don't like this name too much. What is the new window? Do we only add data after each window?

uekerman (commented):
There are still some things we need to figure out. Now is the best time to understand the current code, I guess :grin: 
I guess it would be good to add some documentation to the time package already now. This could help to better define what we want to do exactly.

src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -403,11 +397,27 @@ class BaseCouplingScheme : public CouplingScheme {
    */
   bool accelerate();
```
BenjaminRodenberg, line 398: This name is confusing. We should rename this to `iterate()` or `doImplicitStep()`, since currently we either `performAcceleration` or `extrapolateData` here. Acceleration is only a part of what is happening in this function.

src/cplscheme/BaseCouplingScheme.cpp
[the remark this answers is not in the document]
BenjaminRodenberg: Then let's keep it as it is.

src/cplscheme/BaseCouplingScheme.hpp
BenjaminRodenberg: Done in a694b88

src/acceleration/BaseQNAcceleration.cpp
BenjaminRodenberg: `CouplingData::mesh` is private since https://github.com/BenjaminRodenberg/precice/commit/f5c60edb2589544934fc8881c90d3dc1a550ee9e. Not 100% belonging here, but since `mesh` is read-only it does not make sense to expose the complete mesh. We can also cherry pick this change, if you think it should not be part of this PR.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: Also related to https://github.com/BenjaminRodenberg/precice/commit/f5c60edb2589544934fc8881c90d3dc1a550ee9e.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: Also related to https://github.com/BenjaminRodenberg/precice/commit/f5c60edb2589544934fc8881c90d3dc1a550ee9e.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: Agree. We will have to change a few things in the initialization of the Waveform (by default initialize with `_extrapolationOrder = 1`), since otherwise there will not be any place to store to. But this is not performance critical and generalizes the code.

src/cplscheme/BaseCouplingScheme.hpp
BenjaminRodenberg: This assertion does not make sense anymore. `CouplingData::_lastIteration` (renamed to `_previousIteration`) is uninitialized at the beginning and only gets initialized with `CouplingData::values()` when calling `CouplingData::storeIteration()` the first time. Removed it in dd75e064837846d3da99deed1f8ed12ea6e841db.

src/cplscheme/CouplingData.hpp
BenjaminRodenberg: Yes. Removed it in dd75e064837846d3da99deed1f8ed12ea6e841db.

src/cplscheme/CouplingData.hpp
BenjaminRodenberg: I agree. Renamed it in dd75e064837846d3da99deed1f8ed12ea6e841db.

src/cplscheme/BaseCouplingScheme.hpp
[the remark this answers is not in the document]
BenjaminRodenberg: Then let's keep it as it is.

src/cplscheme/BaseCouplingScheme.hpp
BenjaminRodenberg: This todo is outdated. I removed it in c6e1deb407f3bbb7cdd0b84c9f71cbe156c3c958 and also renamed `storeLastIteration` to `storeIteration`.

src/cplscheme/CouplingData.hpp
BenjaminRodenberg: Should be "this participant". Changed it in e26437c63eaf5e7df470df7cfd5a902dc7440d2c.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: I implemented this change here: 6e516d7. We can even move `storeWindowData` further up, but then we will probably leave `BaseCouplingScheme::doImplicitStep`. I think it also makes sense to always support extrapolation / waveform relaxation (not only for implicit coupling schemes). We can also do an extrapolation with an explicit coupling scheme as long as the data is there.

BenjaminRodenberg: I just realized we cannot move it up too far, since we can only store the data after the convergence check: 4ea5d13

uekerman: We could store and potentially overwrite it in every iteration. I would argue that we should not worry about performance already now.

src/acceleration/BaseQNAcceleration.cpp
BenjaminRodenberg: Removing this part was tricky. See 7ab0220a248cd87138f6f8235d3e03f006101083. I think it's good to not initialize the CouplingData in the Acceleration. This should be done before. I will close this discussion, since it's resolved by 7ab0220a248cd87138f6f8235d3e03f006101083, but will comment on the new problem I see in `BaseCouplingScheme::setupDataMatrices` somewhere else.

src/acceleration/AitkenAcceleration.cpp
BenjaminRodenberg: Solved in 4046df9b6a92019d634b952f4ff8e1bc8523940a.

src/acceleration/ConstantRelaxationAcceleration.cpp
BenjaminRodenberg: Solved in 4046df9b6a92019d634b952f4ff8e1bc8523940a.

src/acceleration/BaseQNAcceleration.cpp
BenjaminRodenberg: This is cleaned up now via 03585b1. Especially 173dbdf.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: This is cleaned up now via 03585b1. Especially 173dbdf.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: This is cleaned up now via 03585b1. Especially 173dbdf.

src/time/Waveform.hpp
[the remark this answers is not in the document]
BenjaminRodenberg: Cleaned this up in aaa536b

src/acceleration/IQNILSAcceleration.cpp
BenjaminRodenberg: Yes. Was just unnecessarily complex. Fixed in 95b529ee07c3e90f03326adfb3a42d4eddee9a11.

src/time/Waveform.hpp
BenjaminRodenberg: Currently yes. But this is also related to the discussion [here](https://github.com/precice/precice/pull/1004#discussion_r627166583).

src/time/Waveform.hpp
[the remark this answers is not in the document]
BenjaminRodenberg: I agree. Let's create a `Time.hpp` with the corresponding class as soon as we know how the interface should look like. Right now the `Waveform` is pretty specialized and has only a single use-case. Creating an abstract class already now seems a bit like overengineering. We can rename `Waveform` and introduce an abstract class `Time` as soon as there is a reason to derive specializations from `Time` for different use-cases. Similarly to `Acceleration <- BaseQNAcceleration <- ...`.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: 44f05b2 is a quite big change in this direction. Idea: Store data of current window in 0th column and old data in additional columns. When a window is converged, we call a special method that moves the data down one column and does extrapolation. Zero order extrapolation is the same like no extrapolation and just uses the last value as an initial guess. Higher order extrapolation is also possible.

I guess we could move `updateThisWindow()` further up, if we want. What we now definitely need is more tests for the coupling scheme.

src/time/Waveform.hpp
BenjaminRodenberg: We now have two methods `updateThisWindow` and `moveToNextWindow`. Maybe this is better?

BenjaminRodenberg (commented):
This self-review is meant to help with the actual review.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -461,27 +428,21 @@ void BaseCouplingScheme::checkCompletenessRequiredActions()
...
-    for (DataMap::value_type &pair : data) {
-      int cols = pair.second->oldValues.cols();
-      PRECICE_DEBUG("Add cols: {}, cols: {}", pair.first, cols);
-      PRECICE_ASSERT(cols <= 1, cols);
-      utils::append(pair.second->oldValues,
-                    (Eigen::MatrixXd) Eigen::MatrixXd::Zero(pair.second->values().size(), _extrapolationOrder + 1 - cols));
-    }
+  // Reserve storage for data values
+  for (DataMap::value_type &pair : getAccelerationData()) {
+    time::PtrWaveform       ptrWaveform(new time::Waveform(pair.second->values().size(), _extrapolationOrder + 2));
```
uekerman, line 442: Why "extrapolationOrder + 2"? Why not just "extrapolationOrder"? Otherwise, this becomes hard to understand.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -461,27 +428,21 @@ void BaseCouplingScheme::checkCompletenessRequiredActions()
...
+    WaveformMap::value_type waveformPair = std::make_pair(pair.first, ptrWaveform);
+    _waveforms.insert(waveformPair);
+    pair.second->storeIteration(); // @ todo remove this duplicate of line above! But removing this line causes failure for mpirun --oversubscribe -np 4 ./testprecice -t PreciceTests/Serial/MultiCoupling
```
uekerman, line 445: TODO

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -665,22 +627,20 @@ bool BaseCouplingScheme::accelerate()
 
   // coupling iteration converged for current time window. Advance in time.
   if (convergence) {
+    moveToNextWindow();
     if (getAcceleration()) {
       getAcceleration()->iterationsConverged(getAccelerationData());
     }
```
uekerman, line 633: Is this order correct? Doesn't this mean that we now use the extrapolated data in the acceleration (e.g. to update W and V of the quasi-Newton methods)?

src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -396,18 +391,33 @@ class BaseCouplingScheme : public CouplingScheme {
...
-   * @brief Extrapolate coupling data from values of previous time windows
-   * @param data Data fields to extrapolate
+   * @brief finalizes this window's data and initializes data for next window.
    */
-  void extrapolateData(DataMap &data);
+  void moveToNextWindow();
+
+  /**
+   * @brief used for storing send/receive data at end of acceleration, if not converged.
+   */
```
uekerman, line 413: "if not converged" ? Currently you do it always

src/cplscheme/tests/ParallelImplicitCouplingSchemeTest.cpp
```diff
@@ -100,46 +100,32 @@ BOOST_AUTO_TEST_CASE(testMVQNPP)
...
+  forces->values() << 0.2, 0.2, 0.2, 0.2;
 
+  PtrCouplingData dpcd(new CouplingData(displacements, dummyMesh, false));
   PtrCouplingData fpcd(new CouplingData(forces, dummyMesh, false));
 
   DataMap data;
   data.insert(std::pair<int, PtrCouplingData>(0, dpcd));
   data.insert(std::pair<int, PtrCouplingData>(1, fpcd));
+  dpcd->storeIteration();
+  fpcd->storeIteration();
```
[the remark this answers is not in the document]
uekerman: Yes, good point.

src/cplscheme/tests/ParallelImplicitCouplingSchemeTest.cpp
```diff
@@ -198,46 +178,30 @@ BOOST_AUTO_TEST_CASE(testVIQNPP)
...
-  utils::append(fcol1, 0.2);
-  utils::append(fcol1, 0.2);
-  utils::append(fcol1, 0.2);
+  forces->values().resize(4);
+  forces->values() << 0.2, 0.2, 0.2, 0.2;
 
+  PtrCouplingData dpcd(new CouplingData(displacements, dummyMesh, false));
   PtrCouplingData fpcd(new CouplingData(forces, dummyMesh, false));
+  dpcd->storeIteration();
+  fpcd->storeIteration();
```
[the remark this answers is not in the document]
uekerman: Yes, good point.

src/time/Waveform.hpp
```diff
@@ -0,0 +1,70 @@
...
+  Eigen::MatrixXd _timeWindows;
+
+  Eigen::VectorXd extrapolateData(int order, int timeWindows)
+  {
+    Eigen::VectorXd extrapolatedValue;
+    if ((order == 0) || (timeWindows < 2 && order > 0)) {
+      PRECICE_ASSERT(this->_timeWindows.cols() > 0);
+      extrapolatedValue = this->_timeWindows.col(0);
+    } else if ((order == 1) || (timeWindows < 3 && order > 1)) { //timesteps is increased before extrapolate is called
+      // PRECICE_INFO("Performing first order extrapolation");
```
uekerman, line 52:
```suggestion
      PRECICE_DEBUG("Performing first order extrapolation");
```

src/time/Waveform.hpp
```diff
@@ -0,0 +1,70 @@
...
+      PRECICE_ASSERT(this->_timeWindows.cols() > 1);
+      extrapolatedValue = this->_timeWindows.col(0) * 2.0; // = 2*x^t
+      extrapolatedValue -= this->_timeWindows.col(1);      // = 2*x^t - x^(t-1)
+    } else if (order == 2) {
+      // PRECICE_INFO("Performing second order extrapolation");
```
uekerman, line 57:
```suggestion
      PRECICE_DEBUG("Performing second order extrapolation");
```

src/time/tests/WaveformTest.cpp
```diff
@@ -0,0 +1,81 @@
...
+  value(0) = 1.0;
+  waveform2.updateThisWindow(value);
+  BOOST_TEST(testing::equals(waveform2.lastTimeWindows()(0, 0), 1.0));
+  BOOST_TEST(testing::equals(waveform2.lastTimeWindows()(0, 1), 0.0));
+  BOOST_TEST(testing::equals(waveform2.lastTimeWindows()(0, 2), 0.0));
+  timeWindowCounter++;
+  waveform2.moveToNextWindow(timeWindowCounter, extrapolationOrder);
+  BOOST_TEST(testing::equals(waveform2.lastTimeWindows()(0, 0), 2.0));
+  BOOST_TEST(testing::equals(waveform2.lastTimeWindows()(0, 1), 1.0));
+  BOOST_TEST(testing::equals(waveform2.lastTimeWindows()(0, 2), 0.0));
```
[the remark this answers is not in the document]
uekerman: Wouldn't it be better to write such a comment as a comment in the code :grin: ?

uekerman (commented):
Looking good.
In later work we still need to improve the names "updateThisWindow" and "moveToNextWindow", but OK for now.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -461,27 +427,21 @@ void BaseCouplingScheme::checkCompletenessRequiredActions()
...
-      PRECICE_ASSERT(cols <= 1, cols);
-      utils::append(pair.second->oldValues,
-                    (Eigen::MatrixXd) Eigen::MatrixXd::Zero(pair.second->values().size(), _extrapolationOrder + 1 - cols));
-    }
+  // Reserve storage for data values
+  for (DataMap::value_type &pair : getAccelerationData()) {
+    time::PtrWaveform       ptrWaveform(new time::Waveform(pair.second->values().size(), _extrapolationOrder + 1));
+    WaveformMap::value_type waveformPair = std::make_pair(pair.first, ptrWaveform);
+    _waveforms.insert(waveformPair);
+    pair.second->storeIteration(); // @ todo remove this duplicate of line above! But removing this line causes failure for mpirun --oversubscribe -np 4 ./testprecice -t PreciceTests/Serial/MultiCoupling
```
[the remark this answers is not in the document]
uekerman: Is this outdated and can be resolved?

src/cplscheme/BiCouplingScheme.cpp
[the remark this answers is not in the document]
uekerman: Only checked quickly, but agree. Maybe the underlying data structure was updated at some point without improving these functions.

src/cplscheme/BaseCouplingScheme.cpp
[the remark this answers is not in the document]
BenjaminRodenberg: No. This is related to [this todo](https://github.com/precice/precice/pull/1004#discussion_r628058494) below. I will close it here and copy paste the important stuff to the comment below.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: From https://github.com/precice/precice/pull/1004#discussion_r627345604:

> Is it correct to call `storeIteration()` for `convMeasure.couplingData` and `pair.second` (from `getAccelerationData()`?). This means that there are convergence measures without an acceleration and there are acceleration schemes without a convergence measure. We might end up calling `storeIteration()` twice on the same piece of data (which is no problem, but a bit confusing.). If this use-case is intended, we can also remove the todo above.

As soon as we figure this out I can solve or remove the todo.

src/cplscheme/BiCouplingScheme.cpp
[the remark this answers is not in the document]
BenjaminRodenberg: Then I will try to take care of this directly on `develop`, since it's not related to this PR.

src/cplscheme/tests/ParallelImplicitCouplingSchemeTest.cpp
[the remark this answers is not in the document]
BenjaminRodenberg: Then I will move the test in `develop` and do some cleaning-up work here after the merge.

src/cplscheme/tests/ParallelImplicitCouplingSchemeTest.cpp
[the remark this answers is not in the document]
BenjaminRodenberg: Then I will move the test in `develop` and do some cleaning-up work here after the merge.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: 1. It should be a `extrapolationOrder + 1`, since we do the extrapolation in-place.
2. For zeroth order extrapolation you need one column (to read from and write to), for first order two, for second order three... Currently, we can also handle this inside of the `Waveform`. The parameter currently defines how much "memory" the waveform needs. 

Btw: zeroth order extrapolation is a special case. If we want to do linear interpolation inside the window in the future, we will also need two columns (beginning and end of window). But for the actual extrapolation one one column at the end of the window is needed.

BenjaminRodenberg: Fixed this in 24e7d63e9aebdb19484b11f95e2ab7373a1cff84

src/cplscheme/tests/ParallelImplicitCouplingSchemeTest.cpp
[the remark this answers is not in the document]
BenjaminRodenberg: Done in 7fac2f8.

src/cplscheme/tests/ParallelImplicitCouplingSchemeTest.cpp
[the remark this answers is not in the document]
BenjaminRodenberg: Done in 7fac2f8.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: After writing this example, I think the current order is incorrect. `doImplicitStep()` is quite complicated, so I will give an example to make sure I get it right:

### Some basics

* `getAccelerationData()[id].second->values()` holds the value that is exchanged. Let's say we have a Force of 5N here.
* `getAccelerationData()[id].second->previousIteration()` is a buffer. We use `storeIteration()` to write to this buffer.
* `_waveforms[id]->lastTimeWindows()` holds the value of the current window, but also of previous windows. Let's say [5N, 3N]. 3N is the Force that was computed in the last time window. We only store two samples, since we only support first order extrapolation in this example.

### Example

1. Let's assume we have just completed the last iteration in our current time window and reached convergence (just got Force 5N). So [`convergence`](https://github.com/BenjaminRodenberg/precice/blob/99ce4d134fae66a6d4aea920d5dcbe5e572e72d1/src/cplscheme/BaseCouplingScheme.cpp#L623) is true and we enter [this branch](https://github.com/BenjaminRodenberg/precice/blob/99ce4d134fae66a6d4aea920d5dcbe5e572e72d1/src/cplscheme/BaseCouplingScheme.cpp#L629-L634).
2. We now call `moveToNextWindow` and this triggers extrapolation. We use the values in `_waveforms[id]->lastTimeWindows()` ([5N, 3N]) and compute 7N through extrapolation. `moveToNextWindow` does two things:
    2.1. Set `getAccelerationData()[id].second->values()` to the newly computed value (7N)
    2.2. Update `_waveforms[id]->lastTimeWindows()` to [7N, 5N], i.e. throwing away the old value of 3N and storing the 7N.
3. After that we exit the branch and call [`storeIteration`](https://github.com/BenjaminRodenberg/precice/blob/99ce4d134fae66a6d4aea920d5dcbe5e572e72d1/src/cplscheme/BaseCouplingScheme.cpp#L643). This will save `getAccelerationData()[id].second->values()` (7N) into `getAccelerationData()[id].second->previousIteration()` (**This will be used for the convergence check in the next iteration!**).
4. We now exit `doImplicitStep()`.
5. After the participant has finished it's iteration (compute a new Force...) and called `advance()` we enter `doImplicitStep()` again. `getAccelerationData()[id].second->values()` was overwritten during the last iteration and now has the value (6N). `getAccelerationData()[id].second->previousIteration()` still holds the old value (7N).
6. We measure convergence and find out that we have not converged. Therefore, entering [this branch](https://github.com/BenjaminRodenberg/precice/blob/99ce4d134fae66a6d4aea920d5dcbe5e572e72d1/src/cplscheme/BaseCouplingScheme.cpp#L635-L640) and perform the acceleration on `getAccelerationData()`.
7. We exit the branch and continue at step 3. above.

Looking at [the version on `develop`](https://github.com/precice/precice/blob/8c2d78d03f1c9330fec41d1a8acbb33e0bcac375/src/cplscheme/BaseCouplingScheme.cpp#L658-L686) I think there is a difference: `storeData` is called **before** `extrapolateData`. Meaning: The extrapolated value is not stored and therefore also not used for convergence checks.

I will fix this now.

src/cplscheme/BaseCouplingScheme.cpp
uekerman: I like the change. Simplifies things.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -665,24 +618,35 @@ bool BaseCouplingScheme::accelerate()
...
   return convergence;
 }
+
+void BaseCouplingScheme::assignDataToConvergenceMeasure(ConvergenceMeasureContext *convergenceMeasure, int dataID)
+{
+  PRECICE_TRACE(dataID);
+  DataMap::iterator iter = _allData.find(dataID);
+  PRECICE_ASSERT(iter != _allData.end(), "Given data ID does not exist in _allData!");
+  convergenceMeasure->couplingData = &(*(iter->second));
+}
```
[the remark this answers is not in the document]
BenjaminRodenberg: I'm not sure whether we should do this. A private method does not harm in general and having this part as a method helps readability in the calling method.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -655,34 +601,47 @@ bool BaseCouplingScheme::anyDataRequiresInitialization(BaseCouplingScheme::DataM
...
   // Stop, when maximal iteration count (given in config) is reached
-  if (maxIterationsReached())
+  if (_iterations == _maxIterations)
     convergence = true;
 
   // coupling iteration converged for current time window. Advance in time.
   if (convergence) {
-    if (getAcceleration()) {
-      getAcceleration()->iterationsConverged(getAccelerationData());
+    // moveToNextWindow();  // @ todo: Wrong position for moveToNextWindow(). We should write a test that catches this! See correct position below.
```
BenjaminRodenberg, line 616: Todo: Write a test here.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -655,34 +601,47 @@ bool BaseCouplingScheme::anyDataRequiresInitialization(BaseCouplingScheme::DataM
...
-  // Store data for conv. measurement, acceleration, or extrapolation
-  storeData();
+  // Store data for conv. measurement, acceleration
+  storeIteration();
 
-  // extrapolate new input data for the solver evaluation in time.
-  if (convergence && (_extrapolationOrder > 0)) {
-    extrapolateData(getAccelerationData());
+  if (convergence) {
+    moveToNextWindow(); // @ todo: Correct position for moveToNextWindow().
```
BenjaminRodenberg, line 632: See above.

src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -519,12 +490,13 @@ class BaseCouplingScheme : public CouplingScheme {
...
 
-  /// Functions needed for initialize()
+  /// Map that links DataID to Waveform
+  typedef std::map<int, time::PtrWaveform> WaveformMap;
 
-  /**
-   * @brief merges all send and receive data into a single data structure.
-   */
-  virtual void mergeData() = 0;
+  /// All waveforms this coupling scheme needs data as a map "data ID -> waveform"
```
BenjaminRodenberg, line 496:
```suggestion
  /// All waveforms this coupling scheme needs data from as a map "data ID -> waveform"
```

src/acceleration/BroydenAcceleration.cpp
```diff
@@ -52,9 +52,10 @@ void BroydenAcceleration::computeUnderrelaxationSecondaryData(
     Eigen::VectorXd &          values = data->values();
     values *= _initialRelaxation; // new * omg
     Eigen::VectorXd &secResiduals = _secondaryResiduals[id];
-    secResiduals                  = data->oldValues.col(0); // old
-    secResiduals *= 1.0 - _initialRelaxation;               // (1-omg) * old
-    values += secResiduals;                                 // (1-omg) * old + new * omg
+    const auto &     oldValues    = data->previousIteration();
+    secResiduals                  = oldValues;
```
BenjaminRodenberg, line 56:
```suggestion
    secResiduals                  = data->previousIteration();
```
Unnecessarily complicated.

src/acceleration/MVQNAcceleration.cpp
```diff
@@ -120,9 +120,10 @@ void MVQNAcceleration::computeUnderrelaxationSecondaryData(
     Eigen::VectorXd &values = data->values();
     values *= _initialRelaxation; // new * omg
     Eigen::VectorXd &secResiduals = _secondaryResiduals[id];
-    secResiduals                  = data->oldValues.col(0); // old
-    secResiduals *= 1.0 - _initialRelaxation;               // (1-omg) * old
-    values += secResiduals;                                 // (1-omg) * old + new * omg
+    const auto       oldValues    = data->previousIteration();
+    secResiduals                  = oldValues; // old
```
BenjaminRodenberg, line 124:
```suggestion
    secResiduals                  = data->previousIteration();
```
Unnecessarily complicated.

src/time/SharedPointer.hpp
```diff
@@ -0,0 +1,12 @@
...
+#include <memory>
+
+namespace precice {
+namespace time {
+
+class Waveform;
+using PtrWaveform = std::shared_ptr<Waveform>;
+
+} // namespace time
+} // namespace precice
```
BenjaminRodenberg, line 12:
```suggestion
} // namespace precice

```
Add newline.

src/time/tests/WaveformTest.cpp
```diff
@@ -0,0 +1,79 @@
...
+  timeWindowCounter++;
+  waveform2.moveToNextWindow(timeWindowCounter, extrapolationOrder); // applies second order extrapolation
+  BOOST_TEST(testing::equals(waveform2.lastTimeWindows()(0, 0), 8.0));
+  BOOST_TEST(testing::equals(waveform2.lastTimeWindows()(0, 1), 4.0));
+  BOOST_TEST(testing::equals(waveform2.lastTimeWindows()(0, 2), 1.0));
+}
+
+BOOST_AUTO_TEST_SUITE_END()
+BOOST_AUTO_TEST_SUITE_END()
```
BenjaminRodenberg, line 79:
```suggestion
BOOST_AUTO_TEST_SUITE_END()

```
add newline

BenjaminRodenberg (commented):
I went through the PR and identified open todos and minor fixes that have still to be done.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: The new test takes care of this. See c6ebb77.

BenjaminRodenberg:
@uekerman Ready for the next review. We have a test now, which ensures that acceleration and extrapolation are applied correctly. There are a few minor todos from my perspective. Rest is ready to merge, as far as I see it.

# Open Todos:

* Some more documentation
* Splitting `Waveform.hpp` into an `hpp` and `cpp` file.

src/time/Waveform.hpp
```diff
@@ -0,0 +1,56 @@
...
+   * @brief Updates entry in _timeWindows corresponding to this window with given data
+   * @param data new sample for this time window
+   */
+  void updateThisWindow(Eigen::VectorXd data);
+
+  /**
+   * @brief Called, when moving to the next time window. All entries in _timeWindows are shifted. The new entry is initialized as the value from the last window (= constant extrapolation)
+   * @param timeWindows number of samples that are valid and may be used for extrapolation. Usually number of past time windows.
+   */
+  void moveToNextWindow(int timeWindows, int order = 0);
```
BenjaminRodenberg, line 29:
```suggestion
  void moveToNextWindow();
```
Removing the two arguments would simplify the interface easier to use, but the class would also become more complex:

* (extrapolation) `order` is defined when the object is created
* Keep track of `timeWindows` inside object (to some degree a duplication of what the coupling scheme also does.)

@uekerman Any opinion? We could also shift this discussion to the next PR, where interpolation will be introduced.

uekerman: This complexity will anyway be moved to the new package. I don't see a problem doing this now.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -463,29 +428,16 @@ void BaseCouplingScheme::checkCompletenessRequiredActions()
...
   PRECICE_TRACE();
-  PRECICE_DEBUG("Data size: {}", data.size());
-  // Reserve storage for convergence measurement of send and receive data values
-  for (ConvergenceMeasureContext &convMeasure : _convergenceMeasures) {
-    PRECICE_ASSERT(convMeasure.couplingData != nullptr);
-    if (convMeasure.couplingData->oldValues.cols() < 1) {
-      utils::append(convMeasure.couplingData->oldValues,
-                    (Eigen::MatrixXd) Eigen::MatrixXd::Zero(convMeasure.couplingData->values().size(), 1));
-    }
+  // Reserve storage for all data (acceleration and convergence measurement)
```
uekerman, line 434:
```suggestion
  // Reserve storage for all data
```
We do it for all data now, right? Not only such used in acc or cm.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -657,34 +603,46 @@ bool BaseCouplingScheme::anyDataRequiresInitialization(BaseCouplingScheme::DataM
   return false;
 }
 
-bool BaseCouplingScheme::accelerate()
+bool BaseCouplingScheme::doImplicitStep()
 {
+  updateThisWindow();
```
uekerman, line 608: I am still a bit confused by this name. Wouldn't "updateWaveforms" be better?
Or "storeDataInWaveforms"?

uekerman (approved):
Minor things, otherwise good to go. Please merge when done.
And don't forget the changelog.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: Fixed in 880edd9dd447723d8bcc9e9881435debfba7399a.

PR #1006, opened 2021-04-27
Modernize pass by value

DavidSCN:
## Main changes of this PR
One of the more debatable changes. I think the code is not really better readable and most compiler would optimize it probably anyway. I would rather disable the `modernize-pass-by-value` option, but let's see. Related to #999. 

## Author's checklist

* <del> I added a changelog file with this PR number in `docs/changelog/` if there are noteworthy changes.</del>
* [x] squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] (more questions/tasks)


MakisH:
I also thought that compilers optimize this anyway. I got the impression that over-using `std::move` can be confusing, so I guess there should be a clear argument for this change. But I am not a fitting reviewer for this.

uekerman:
In my non-C++-guru opinion, this reduces readability. Even if the compiler does not optimize it anyway, it's still debatable in non-performance-critical code parts.

fsimonis:
> I also thought that compilers optimize this anyway. I got the impression that over-using `std::move` can be confusing, so I guess there should be a clear argument for this change. But I am not a fitting reviewer for this.

@MakisH 
Passing by `T const &` requires the ctor to call the copy ctor of T.
Passing by value requires the ctor to call the copy ctor of T (as the argument is an lvalue).
Passing by value and `std::move` allows to call the move ctor (given it exists).

Move elision should allow for the copy to be directly constructed "in-place". However, this at least requires the ctor definition to be visible to the compiler at the call site of the ctor. In practise this means that the constructor definition has to be in the header. I have no idea how well this optimization is supported though.

> In my non-C++-guru opinion, this reduces readability.

@uekerman In terms of readability, the signature of the ctor actually becomes less busy. as the `const &` disappears.
Also the signature tells the caller "I need a copy, so give me a copy", which is actually more expressive.

The constructor definition definitely becomes more busy though.

DavidSCN:
As far as I understand this option is partially contradicting to [`performance-unnecessary-value-param`](https://clang.llvm.org/extra/clang-tidy/checks/performance-unnecessary-value-param.html). How could we come to a decision here?

fsimonis:
@DavidSCN the mentioned check is used to prevent passing by value when the function uses the argument in a read only fashion. An example would be invoking only const members on the argument.

This would trigger here:
```cpp
std:set<int> makeSet(std::vector<int> v) {
  return std:set<int>(v.cbegin(), v.cend());
} 
```

fsimonis:
As there are no counter arguments, I feel take the liberty to merge this PR.

Issue #483, opened 2019-08-12
Overweight Mesh Primitives


fsimonis:
Our mesh primitives are remarkably big. We should definitely consider to reduce their size to reduce the overall memory consumption of preCICE.

Primitive | Size [B] | Add. dynamically allocated
--- | --- | ---
`Triangle` | 144 | vertexMap
`Edge` | 128 |
`Vertex` | 136 |

Their common base class `Propertycontainer` is quite heavy weight at `88 Bytes`  as it stores:
* A _private_ logger
* vector of parents
* map of properties

Note that those all dynamically allocate even more memory on top of the numbers in the table above.

floli:
I agree that these classes are too fat.

The logger can be removed, as far I as I can tell by a quick scan-through. The classes do not make use of them, with the exception of `PropertyMeshContainer.hpp:145`.

Furthermore, I think that the entire system of attaching properties to vertex is not in use.

fsimonis:
I just ran preCICE with a NP testcase using ASTE. I ran the example first using 3772b9b98a455adb7d62a4479b26778e08001803 and then just deleted the logger in the Propertycontainer. This is the "default" testcase of the config in ASTE.
I run using 2x2 ranks and use meshes:
* A 100x100 fully connected with data  (x+y+z)
* B 200x200 fully connected

## Memory Consumption
_for the master rank of B (most work)_

### Before

![image](https://user-images.githubusercontent.com/13552216/62880250-462c5800-bd2d-11e9-8eed-3fa1bd988292.png)

* The 3 bottom slates _alone_ are `boost::log` related

### After

![image](https://user-images.githubusercontent.com/13552216/62880287-5ba18200-bd2d-11e9-935d-578dd1521905.png)

* The bottom orange slate is the `rtree`
* The next slate is 'mesh::createEdge()`

### Conclusion

* The logger in the `Propertycontainer` is responsible for a majority of the memory consumption.
* Removing this logger resulted in a peak memory reduction of __over 50%__, for this little testcase alone!

fsimonis:
We decided to make the logger static instead of completely removing it in bf6f4b62. This essentially has the same effect as removing it but gives us the possibility to log in the base class.
This is just one step in reducing the impact of the PropertyContainer.

floli:
Could you briefly comment in the source code why we made an exempt from the rule and use a static logger there?

fsimonis:
After the refactoring of the mesh namespace in #585, the situation looks as follows:


[...]

PR #1008, opened 2021-04-28
Improve vertex coordinate representation

fsimonis:
## Main changes of this PR

In a nutshell, this PR changes the coordinate representation of vertices from an `Eigen::VectorXd` to an `std::array<double,3>` plus a `short dim`.

The primary motivation is to reduce the memory footprint as already discussed in #483.
What used to be a 16B object plus 16B/24B dynamic allocation for 2D/3D is now a 24B array plus ~8B for the dimension.
In theory, this leads to no change for 2D and an 8B saving for 3D; in practise, the compiler packs the dimension with the other members resulting in a gain even for ~~3D~~ 2D.
As a bonus, this change reduces dynamic allocations by `Eigen::VectorXd` and improves data locality.

The `mesh::Vertex` now offers two ways to access the coordinate:
```cpp
// Direct access to the coordinate buffer.
// Use getDimensions() to get the dimensionality
const std::array<double, 3> & Vertex::rawCoords() const;

// Constructs and returns Eigen::VectorXd.
Eigen::VectorXd Vertex::getCoords() const;
```

The memory overhead of meshes unnecessarily big!
See #483 for more details.

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/com/CommunicateMesh.cpp
+++ b/src/com/CommunicateMesh.cpp
@@ -29,31 +29,30 @@ void CommunicateMesh::sendMesh(
     int               rankReceiver)
 {
   PRECICE_TRACE(mesh.getName(), rankReceiver);
-  int dim = mesh.getDimensions();
+  const int dim = mesh.getDimensions();
 
-  int numberOfVertices = mesh.vertices().size();
+  const auto &meshVertices     = mesh.vertices();
+  const int   numberOfVertices = meshVertices.size();
   _communication->send(numberOfVertices, rankReceiver);
   if (not mesh.vertices().empty()) {
     std::vector<double> coords(static_cast<size_t>(numberOfVertices) * dim);
     std::vector<int>    globalIDs(numberOfVertices);
     for (int i = 0; i < numberOfVertices; i++) {
-      for (int d = 0; d < dim; d++) {
-        coords[i * dim + d] = mesh.vertices()[i].getCoords()[d];
-      }
-      globalIDs[i] = mesh.vertices()[i].getGlobalIndex();
+      std::copy_n(meshVertices[i].rawCoords().begin(), dim, &coords[i * dim]);
+      globalIDs[i] = meshVertices[i].getGlobalIndex();
     }
     _communication->send(coords, rankReceiver);
     _communication->send(globalIDs, rankReceiver);
   }
 
-  int numberOfEdges = mesh.edges().size();
+  const int numberOfEdges = mesh.edges().size();
   _communication->send(numberOfEdges, rankReceiver);
   if (not mesh.edges().empty()) {
     //we need to send the vertexIDs first such that the right edges can be created later
     //contrary to the normal sendMesh, this variant must also work for adding delta meshes
     std::vector<int> vertexIDs(numberOfVertices);
     for (int i = 0; i < numberOfVertices; i++) {
-      vertexIDs[i] = mesh.vertices()[i].getID();
+      vertexIDs[i] = meshVertices[i].getID();
     }
     _communication->send(vertexIDs, rankReceiver);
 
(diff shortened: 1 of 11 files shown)


src/com/CommunicateMesh.cpp
```diff
@@ -201,14 +199,15 @@ void CommunicateMesh::broadcastSendMesh(const mesh::Mesh &mesh)
...
     }
     _communication->broadcast(vertexIDs);
 
     std::vector<int> edgeIDs(numberOfEdges * 2);
+    const auto &     meshEdges = mesh.edges();
     for (int i = 0; i < numberOfEdges; i++) {
-      edgeIDs[i * 2]     = mesh.edges()[i].vertex(0).getID();
-      edgeIDs[i * 2 + 1] = mesh.edges()[i].vertex(1).getID();
+      edgeIDs[i * 2]     = meshEdges[i].vertex(0).getID();
+      edgeIDs[i * 2 + 1] = meshEdges[i].vertex(1).getID();
```
DavidSCN, line 210: I know this is not your crime, but we should maybe choose a named variable for 2 (I guess `vertexIdsPerEdge` or so). `clang` complains about this 'magic' numbers and it would make the code for newcomer better readable.

src/com/CommunicateMesh.cpp
```diff
@@ -226,10 +225,11 @@ void CommunicateMesh::broadcastSendMesh(const mesh::Mesh &mesh)
...
 
       std::vector<int> triangleIDs(numberOfTriangles * 3);
+      const auto &     meshTriangles = mesh.triangles();
       for (int i = 0; i < numberOfTriangles; i++) {
-        triangleIDs[i * 3]     = mesh.triangles()[i].edge(0).getID();
-        triangleIDs[i * 3 + 1] = mesh.triangles()[i].edge(1).getID();
-        triangleIDs[i * 3 + 2] = mesh.triangles()[i].edge(2).getID();
+        triangleIDs[i * 3]     = meshTriangles[i].edge(0).getID();
+        triangleIDs[i * 3 + 1] = meshTriangles[i].edge(1).getID();
+        triangleIDs[i * 3 + 2] = meshTriangles[i].edge(2).getID();
```
DavidSCN, line 232: Same as above for `3`.

DavidSCN (commented):
> In theory, this leads to no change for 2D and an 8B saving for 3D; in practise, the compiler packs the dimension with the other members resulting in a gain even for 3D.

I guess you mean a gain even for _2D_ here?!

I like the change in general. One point ;) to ask: Why not using `std::array<double, dim>` and templating the dimension through the class?! I think the vertex and data classes are really the central place to do so. The question goes already a bit in the future direction: having a dead dangling coordinate for 2D volume coupling is probably too expensive.

src/com/CommunicateMesh.cpp
fsimonis: Do you mind opening an issue regarding this? That will make it easier to discuss alternatives.

src/com/CommunicateMesh.cpp
fsimonis: See https://github.com/precice/precice/pull/1008#discussion_r624929570

fsimonis:
> I guess you mean a gain even for 2D here?!

True, thanks for the hint.

> Why not using std::array<double, dim> and templating the dimension through the class?!

Implementing the dimension as a template argument requires to either duplicate all functionality based on vertices or requires to templatise everything that uses vertices (e.g. Mesh, Exporter, Mapping).
This also requires to move most of the code into the headers which will make compilation times skyrocket.

> The question goes already a bit in the future direction: having a dead dangling coordinate for 2D volume coupling is probably too expensive.

The spacial index trees work in 3D and having a 3rd coordinate that is set to 0 in a 2D case _significantly_ simplifies the interaction with boost geometry. With the `Eigen::VectorXd` coordinates, we needed to enlarge 2D coordinates to 3D coordinates when the 3rd coordinate was accessed by boost, which was ugly and error prone. Have a look at the old `RTreeAdapter.hpp` for more info on that.

Refactoring the dimensionality into the type system may be useful in the future, but this will be a major work-package.

PR #1010, opened 2021-04-28
Fix read and write validation macros

### LINKED ISSUE #989: Writing to a data field without a 'write-data' tag is not Asserted
**Describe the problem**
Writing data via one of the 'write' functionalities, where no mapping is defined leads to a failure in the following lines of code

https://github.com/precice/precice/blob/6d27ad5db42997bbc06ff97b39eb4999196295b7/src/precice/impl/Participant.cpp#L156-L162

since the `dataContext` is `nullptr`. However, this is either a misconfiguration of the user or of the adapter developer and we should capture it before and provide a more descriptive error message.

**Step To Reproduce**
Configure your data:
```
const int dataID = couplingInterface.getDataID("myData", meshID);
```
and write with this data
```
        couplingInterface.writeBlockScalarData(dataID, writeData.size(), ids.data(), writeData.data());
```

without specifying a mapping for "myData" in the config file.

DavidSCN:
## Main changes of this PR
Fixes #989.

The validation macros don't work with the current implementation. The reason is pretty simple by looking at the following three code snippets:
The check:

https://github.com/precice/precice/blob/813de05a87270ae9194bb4a499895a9fab4caef9/src/precice/impl/ValidationMacros.hpp#L117-L123

The assertion we hit
https://github.com/precice/precice/blob/813de05a87270ae9194bb4a499895a9fab4caef9/src/precice/impl/Participant.cpp#L155-L160

Definition of `is_used` 
https://github.com/precice/precice/blob/813de05a87270ae9194bb4a499895a9fab4caef9/src/precice/impl/Participant.cpp#L207-L212

Hence, we try to access the `dataContext` in order to print an error message although the `dataContext` might be `nullptr`, which is also the case we want to catch here. I don't see a method how we can access the data name in the validation macros in order to print a descriptive error message without the `dataAccessor`. The applied changes are more a dummy to illustrate the problem.
## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/precice/impl/ValidationMacros.hpp
+++ b/src/precice/impl/ValidationMacros.hpp
@@ -104,11 +104,10 @@
  */
 #define PRECICE_REQUIRE_DATA_READ_IMPL(id)                                                                                  \
   PRECICE_VALIDATE_DATA_ID_IMPL(id)                                                                                         \
-  DataContext &context = _accessor->dataContext(id);                                                                        \
   PRECICE_CHECK((_accessor->isDataUsed(id) && _accessor->isDataRead(id)),                                                   \
                 "This participant does not use Data \"{}\", but attempted to read it. "                                     \
                 "Please extend the configuarion of partiticpant \"{}\" by defining <read-data mesh=\"{}\" name=\"{}\" />.", \
-                context.getName(), _accessorName, context.mesh->getName(), context.getName());
+                "context.getName()", _accessorName, "context.mesh->getName()", "context.getName()");
 
 /** Implementation of PRECICE_REQUIRE_DATA_WRITE()
  *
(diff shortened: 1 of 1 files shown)


fsimonis:
These macros validate the data ids first using the refactor-worthy:

https://github.com/precice/precice/blob/813de05a87270ae9194bb4a499895a9fab4caef9/src/precice/impl/ValidationMacros.hpp#L93-L99

This check will fail if the given dataid is not in a MeshContext used by the 
`_dataID` is of (pretty) type:
```cpp
std::map<MeshID, std::map<DataName, DataID>>
```

This is populated in `configure()`
https://github.com/precice/precice/blob/813de05a87270ae9194bb4a499895a9fab4caef9/src/precice/impl/SolverInterfaceImpl.cpp#L215-L226

Do you have an example where this does not work?

DavidSCN:
> Do you have an example where this does not work?

You can find the example in the linked issue. From what I understand: the data `Force` is actually used by the participant (which is also validated by the first macro), but it is not part of the `dataContext` as there is missing the `<write-data` tag on the second participant. Hence, the `dataContext[ForceID]` is actually `nullptr`.

fsimonis:
Nice! Thanks for unravelling this issue!

This PR may be a fix, but it is not a solution to the problem.
Let's refactor this logic out of the SolverInterface into the Participant and add appropriate calls to implement the use-cases in the SolverInterface.

Issue #248, opened 2019-02-06
Readable Types for IDs


fsimonis:
# Problem
In our current code-base we use various IDs:
*  MPI ranks
* Local, Global IDs for Primitives (Vertex, Edge, Triangle, Quad)
* Mesh IDs
* Data IDs
* solverProcess IDs

The type currently used for said IDs is `int`, which does not convey any information about what it actually represents. This leads to an instant loss of information when used and we are left with only the name:
```cpp
int mid = interface.getMeshID("name");
^   ^
|   |_ might be MeshID
|_ type is ... a number and can even be negative

if (Parallel::getProcessRank() == mid) // comparing MeshID to MPI Ranks ?
mid = -1; // Negative mesh id?
interface.setMeshEdge(mid, 1, 3); // BOOM
```

This lack of visual information leads to even more unreadable composite types:
https://github.com/precice/precice/blob/d601da73c78bf684fa986bb4e2d81861fc96b982/src/mesh/Mesh.hpp#L54
https://github.com/precice/precice/blob/d601da73c78bf684fa986bb4e2d81861fc96b982/src/mesh/Mesh.hpp#L300
https://github.com/precice/precice/blob/d601da73c78bf684fa986bb4e2d81861fc96b982/src/m2n/PointToPointCommunication.cpp#L225

# Proposal
## Step 1: Aliases - Readability
Using aliases improves the readability without any change of the code:

```cpp
using VertexDistribution = std::map<MPIRank, std::vector<VertexID>>; 
std::map<std::string, MeshID> _nameIDPairs; 
std::map<MPIRank, std::vector<MPIRank>> buildCommunicationMap(
```

## Step 2: Strong Types - Usability
[Strong types](https://www.fluentcpp.com/posts/#strong-types) lift the ids into the type-system and prevents invalid conversions between these types at compile time:
```cpp
MeshID mid = getMeshID("Forces");
MPIRank rank = Parallel::getProcessRank();
int x(2);

mid = x; // error: no viable overloaded '=' no known conversion from int to MeshID
mid = rank; // error: no viable overloaded '=' no known conversion from MPIRank to MeshID
if (rank == mid) // error: invalid operands to binary expression (MPIRank and MeshID)
```
We can then add functionality to those strong types if we need to:
```cpp
if (Parallel::getProcessRank().isMaster()) { /* ... */ } else { /* ... */ };
bool operator==(MeshID lhs, MeshID rhs) { return lhs.id() == rhs.id() }
[...]

PR #1015, opened 2021-05-06
Fix enforced gather scatter communication with empty master ranks

### LINKED ISSUE #1013: Simulation hangs when using GatherScatterCommunication and the master rank of one solver has empty data sets
**Describe your setup**

preCICE Version: 2.2.0
pyprecice Version: 2.2.0.2
OpenMPI Version: 1.8.3

**Describe the problem**
When enforcing gather-scatter-communication and a parallel solver has data sets that are not empty but that are empty for the master rank, the data sets will not be sent to preCICE. I think the issue is in the [BaseCouplingScheme](https://precice.org/doxygen/master/BaseCouplingScheme_8cpp_source.html#l00069), where the if-clause in line 78 should be deleted or changed. If we go into the source code of the [GatherScatterCommunication](https://precice.org/doxygen/develop/GatherScatterCommunication_8cpp_source.html#l00061), we can see that the master rank is required to enter this method for the data to be sent at line 106. 

**Step To Reproduce**
1. Create simulation that enforces gather-scatter-communication and has one solver that runs in parallel with OpenMPI.
2. The parallel solver should have data sets that are empty for the master rank but otherwise have data points.
3. Run the simulation.

**Expected behaviour**
The simulation should not be hanging.

**Additional context**
There is a relevant forum post for this issue which can be found [here](https://precice.discourse.group/t/simulation-stuck-at-advance-for-both-solvers/555/29).

DavidSCN:
## Main changes of this PR
I already added the tests for now. There are two different bugs ( I added two tests): 
- 1. the provided partition of the master rank is empty
- 2. the received partition of the master rank is empty

I'm a bit puzzled on how to fix it. Letting the master rank sending empty data leads to a deadlock in the communication as well (first test case). I haven't yet looked into the second bug, it seems to be a bit more involved. 

Closes #1013 

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)


uekerman:
I think we have the same problem when receiving:

https://github.com/precice/precice/blob/0951af57c3419400925d7d1f0acb036bdd5ce577/src/cplscheme/BaseCouplingScheme.cpp#L94-L96

Do we need the `size>0` anyway? If yes, what about moving it further down into `Point2PointCom` and `GatherScatterCom`?

DavidSCN:
There is a bug in the vertex match function:

https://github.com/precice/precice/blob/0951af57c3419400925d7d1f0acb036bdd5ce577/src/query/Index.cpp#L32-L47

In case there is no match, the function just returns some uninitialized rubbish.

DavidSCN:
Not yet there, but I would like to get some feedback on this: I guess we have a similar problem for the nearest-projection mapping (Edges..). Any idea why we get there? The failing test is the second one: the provided partition of the master is not empty, but the received one is empty. 
I We should probably add more assertions in the other functions as well.

DavidSCN:
I have now everything together: our nearest-neighbor (probably other local mappings as well) cannot deal with empty received partitions, which might happen due to a too strict filtering (actually every vertex is filtered out).  Workaround `<use-mesh...  geometric-filter="no-filter` would work here. Filtering `on-master` seems to fail as well (even hangs, not sure if it is tested). Computing a bounding-box based on a single vertex might end up in a problematic empty bounding box as well (not yet completely sure, if this is really happening in the code),
@uekerman should we address this issue here or address this in a separate or do you disagree completely?

DavidSCN:
Thanks for the feedback. You may want to look at the failing tests for `Debug` and `Release`, which are caused by the empty partition. Can you tell me where I can find the assertion. Running 
```bash
~/precice/src/query $ egrep -rn 'PRECICE_ASSERT' *
impl/RTreeAdapter.hpp:200:  PRECICE_ASSERT(size == 2 || size == 3, size);
impl/Indexer.cpp:23:  PRECICE_ASSERT(mesh);
impl/Indexer.cpp:43:  PRECICE_ASSERT(mesh);
impl/Indexer.cpp:63:  PRECICE_ASSERT(mesh);
```
and what I would add is
`Index.cpp:42:  PRECICE_ASSERT(_mesh->vertices().size() != 0, _mesh->vertices().size());`
but that would only be in order to catch and prevent such issues in the future. The filtered partition is still a problem.

oguzziya:
If the assertion for `Index` is not added yet, I can add it now. For the filtering out, I have an open PR on extending the filtering with connectivity elements https://github.com/precice/precice/pull/986 . Currently, it is implemented only for `scaled-consistent` type of constraint but we can extend it for other types as well

oguzziya:
Well, an assertion on the `Index` is not trivial as I thought. What about the intentional empty ranks? we either need to distinguish it or instead of an assertion, we need to take care of the returned object from the query functions.

DavidSCN:
> If the assertion for Index is not added yet, I can add it now.

Very much appreciated!

> For the filtering out, I have an open PR on extending the filtering with connectivity elements #986 . Currently, it is implemented only for scaled-consistent type of constraint but we can extend it for other types as well

Does it include a solution for an empty received partition (due to bounding box filtering) for non-empty provided partitions (as it is the case here). I don't set any connectivity here, so I'm not sure if it would help.

oguzziya:
Unfortunately, it only covers the cases with connectivity.

For the assertion index tree, I am not sure whether it helps or not. If we add an assertion, we need to make sure that none of the empty partitions calls the index tree constructor, which in turn leads to a very dirty and again error-prone code (if someone forgets to put 
```
if (not mesh.vertices().empty()){
   Index myTree(mesh);
}
```
it would probably trigger the assertion. What I would propose is, add a check in the `getClosestVertex` function and return `distance = -1, index=-1` pass the responsibility to the caller.

DavidSCN:
> For the assertion index tree, I am not sure whether it helps or not. If we add an assertion, we need to make sure that none of the empty partitions calls the index tree constructor, which in turn leads to a very dirty and again error-prone code (if someone forgets to put

Maybe we should do both and wrap it around a `try..catch ` block.

> What I would propose is, add a check in the getClosestVertex function and throw a warning if there is no match and return distance = -1, index=-1 pass the responsibility to the caller.

Yes, that's also what I have done, at least partially. I guess we should still add an Assertion/abort in `getClosestVertex` since a well defined, but erroneous index (e.g. -1) leads further down just to cryptic error when trying to access the respective position in `vertexIndex`.

fsimonis:
The `Index` isn't really the problem here.

The mapping is a noop when the output mesh is empty and invalid when the input mesh is empty. (both empty is a noop too)
These are cases that should be handled in the mapping.

src/precice/tests/ParallelTests.cpp
```diff
@@ -176,6 +176,105 @@ BOOST_AUTO_TEST_CASE(ConstructAndExplicitFinalize)
...
+    // Set mesh vertices
+    interface.setMeshVertices(meshID, size, coordinates.data(), ids.data());
+
+    // Initialize the solverinterface
+    double dt = interface.initialize();
+
+    // Create some dummy writeData
+    std::vector<double> writeData;
+    for (unsigned int i = 0; i < size; ++i)
+      writeData.emplace_back(i + 1);
```
fsimonis, line 208: We decided to always use `{` blocks `}` after after control structures.
There already have been too many painful errors because of that style.
Could you please apply that to the whole test?

src/precice/tests/ParallelTests.cpp
```diff
@@ -176,6 +176,105 @@ BOOST_AUTO_TEST_CASE(ConstructAndExplicitFinalize)
...
+    // Get mesh and data IDs
+    SolverInterface interface(context.name, configFilename, context.rank, context.size);
+    const int       meshID      = interface.getMeshID("ParallelMesh");
+    const int       writeDataID = interface.getDataID("MyData1", meshID);
+    const int       readDataID  = interface.getDataID("MyData2", meshID);
+    const int       dim         = interface.getDimensions();
+    BOOST_TEST(dim == 2);
+
+    // Set coordinates, master according to input argument
+    const std::vector<double> coordinates = context.isMaster() ? masterPartition : std::vector<double>({0.0, 0.5, 0.0, 3.5, 0.0, 5.0});
```
fsimonis, line 195: Let's use a plain old `if` here instead the ternary operator.
Sometimes readability is more important than compactness.

Another point is that you can simply use `std::vector<double>{...}`. This will invoke the initializer_list constructor if it exists.

If you _really_ want the vector to be `const`, consider to write a separate function, or use an immediately invoked lambda expression:
```cpp
const auto coordinates =  [&] -> std::vector<double> { if (context.isMaster()) { return masterPartition; } else { return {0.0, 0.5, 0.0, 3.5, 0.0, 5.0}; }}();
```

src/partition/ReceivedPartition.cpp
```diff
@@ -328,17 +332,16 @@ void ReceivedPartition::filterByBoundingBox()
...
       PRECICE_DEBUG("Bounding box filter, filtered from {} to {} vertices, {} to {} edges, and {} to {} triangles.",
                     _mesh->vertices().size(), filteredMesh.vertices().size(),
                     _mesh->edges().size(), filteredMesh.edges().size(),
                     _mesh->triangles().size(), filteredMesh.triangles().size());
 
       _mesh->clear();
       _mesh->addMesh(filteredMesh);
+      if (isAnyProvidedMeshNonEmpty()) {
+        PRECICE_CHECK(not _mesh->vertices().empty(), errorMeshFilteredOut(_mesh->getName(), utils::MasterSlave::getRank()));
+      }
```
fsimonis, line 344: Just for my understanding. This really was the bug, correct?

src/partition/ReceivedPartition.cpp
```diff
@@ -238,14 +238,18 @@ void ReceivedPartition::compute()
 }
 
 namespace {
-std::string errorMeshFilteredOut(const std::string &meshName)
+std::string errorMeshFilteredOut(const std::string &meshName, const int rank)
```
fsimonis, line 241: As you already touch this function, could you please port it to fmtlib?

This also allows you to remove all the `+` from the message and use consecutive string literals instead.

Please use `utils/fmt.hpp`.
```
#include <utils/fmt.hpp>
return fmt::format("...", meshName, rank);
```

src/precice/tests/ParallelTests.cpp
DavidSCN: Done

src/query/Index.hpp
```diff
@@ -15,6 +15,7 @@ namespace query {
 
 /// Struct to hold index and distance information of the closest primitive
 template <class Tag>
+// TODO: Initialize these values with defaults
```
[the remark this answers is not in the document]
DavidSCN: Maybe it would be good to have something like `static int constants::invalidIndex = std::numeric_limits<int>::min()` in a utilities function in order to use it for such places and consistently throughout the code. That's actually similar to what the deal people do.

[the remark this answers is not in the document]
DavidSCN: It would also enable to catch up invalid indices in the code.

src/partition/ReceivedPartition.cpp
DavidSCN: Done as well

src/precice/tests/ParallelTests.cpp
DavidSCN: > Let's use a plain old if here instead the ternary operator.
Sometimes readability is more important than compactness.

I totally agree on the second point, but I don't see, how 6 LOCs 
```cpp
std::vector<double> coordinates;
if(context.isMaster()){
coordinates=bla;
}else{
coordinates=blabla;
}
```
are better readable than the one line given above. The lambda version is much worse readable than both mentioned above.

src/query/Index.hpp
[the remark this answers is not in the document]
fsimonis: This is strongly related to #248.
My suggestion is to simply use `-1` now and come up with a solution later.
You can use the doxygen `/// @todo text` to create an entry in https://precice.org/doxygen/develop/todo.html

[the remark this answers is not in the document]
DavidSCN: Added it.

src/query/Index.hpp
```diff
@@ -14,11 +14,13 @@ namespace precice {
 namespace query {
 
 /// Struct to hold index and distance information of the closest primitive
+/// @todo Replace default magic number initialization with a dedicated variable
 template <class Tag>
 struct MatchType {
   double distance;
   int    index;
```
fsimonis, line 21: This struct is a good candidate for [aggregate initialization](https://en.cppreference.com/w/cpp/language/aggregate_initialization).
In this case, we would need to delete all constructors and then specify default member initializers.

```cpp
struct MatchType {
  double distance{-1};
  int    index{-1};
  constexpr bool operator<(MatchType const &other) const;
};
```

DavidSCN: I agree on that, but do not know enough about aggregate initialization. Are you sure we need to remove all constructors (and why can we)?
What I would do now would look the following way:
```cpp
template <class Tag>
struct MatchType {
  double distance{-1};
  int    index{-1};
  MatchType() = default;
  MatchType(double d, int i)
      : distance(d), index(i){};

  constexpr bool operator<(MatchType const &other) const
  {
    return distance < other.distance;
  };
};
````

**Comment by DavidSCN** [comment]
For the sake of completeness I just tried Alex parallel solverdummy cases and the actual issue is fixed with this PR.

File: src/query/Index.hpp (hunk 6 above)
Comment (fsimonis) [reply src/query/Index.hpp]: The compiler writes the constructors for you. Have a look at [this example](https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:2,endLineNumber:16,positionColumn:2,positionLineNumber:16,selectionStartColumn:2,selectionStartLineNumber:16,startColumn:2,startLineNumber:16),source:'%23include%3Ciostream%3E%0A%0Astruct+S+%7B%0A++++int+a%7B-1%7D%3B%0A++++double+b%7B-1%7D%3B%0A%7D%3B%0A%0Aint+main()+%7B%0A++++S+s1%3B%0A++++std::cout+%3C%3C+s1.a+%2B+s1.b+%3C%3C+!'%5Cn!'%3B%0A%0A++++S+s2%7B1,+0.3%7D%3B+//+aggregate+init%0A++++std::cout+%3C%3C+s2.a+%2B+s2.b+%3C%3C+!'%5Cn!'%3B%0A%0A++++return+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:35.77023498694517,l:'4',n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:clang_trunk,filters:(b:'0',binary:'1',commentOnly:'0',demangle:'0',directives:'0',execute:'0',intel:'0',libraryCode:'0',trim:'1'),fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!((name:eigen,ver:'339')),options:'-Wall+-Wextra+--std%3Dc%2B%2B14',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1),l:'5',n:'0',o:'x86-64+clang+(trunk)+(Editor+%231,+Compiler+%231)+C%2B%2B',t:'0')),k:50,l:'4',m:35.56581986143187,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compiler:1,editor:1,fontScale:14,fontUsePx:'0',wrap:'1'),l:'5',n:'0',o:'%231+with+x86-64+clang+(trunk)',t:'0')),header:(),l:'4',m:64.43418013856814,n:'0',o:'',s:0,t:'0')),k:64.22976501305483,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4).

Comment (DavidSCN) [reply src/query/Index.hpp]: I see, thanks. I will add it, although it will hinder people from using `MatchType(1,2,)` with round braces `()` meaning this only works for curly braces, but I'm fine with that.

Comment (DavidSCN) [reply src/query/Index.hpp]: I cannot yet apply this change in a nice way due to the following warning (I skipped the following one):
```bash
~/precice/src/query/Index.cpp:45:115: warning: narrowing conversion of ‘matchID’ from ‘size_t’ {aka ‘long unsigned int’} to ‘int’ [-Wnarrowing]
   45 |                                        match = VertexMatch{bg::distance(sourceCoord, _mesh->vertices()[matchID]), matchID};
``` 
This is again related to #248. I added now the aggregate initialization without actually using it, meaning I still have the constructors in the class. 

Since I promised to finish this today, I will merge here. If you don't like it or think it is worth changing the aggregate initialization stuff in `Index.cpp`, feel free to add it later on.

DavidSCN:
I have now only partially addressed the last review comment (see https://github.com/precice/precice/pull/1015#discussion_r631285464), but it is not related to the actual issue and I promised the merge today. Since I need to hand over to  @fsimonis for the releasing anyway, maybe he wants to address this down the line.

PR #1016, opened 2021-05-06
Refactor mesh and data logic

fsimonis:
## Main changes of this PR

* Refactors all logic regarding mesh and data into the participant.
  This results in a lot of new smaller methods in the Participant
* Reordered the functions and added documentation to the functions in Pariticipant

There is too much unrelated logic and duplicate information in the SolverInterface.
This should simplify to refactor contexts in the future

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/precice/impl/Participant.cpp
+++ b/src/precice/impl/Participant.cpp
@@ -41,9 +41,18 @@ Participant::~Participant()
   _writeMappingContexts.deleteElements();
 }
 
-const std::string &Participant::getName() const
+/// Configuration interface
+
+void Participant::addAction(action::PtrAction &&action)
 {
-  return _name;
+  auto &context = meshContext(action->getMesh()->getID());
+  context.require(action->getMeshRequirement());
+  _actions.push_back(std::move(action));
+}
+
+void Participant::setUseMaster(bool useMaster)
+{
+  _useMaster = useMaster;
 }
 
 void Participant::addWatchPoint(
@@ -56,16 +65,6 @@ void Participant::addWatchIntegral(
     const PtrWatchIntegral &watchIntegral)
 {
   _watchIntegrals.push_back(watchIntegral);
-}
-
-std::vector<PtrWatchPoint> &Participant::watchPoints()
-{
-  return _watchPoints;
-}
-
-std::vector<PtrWatchIntegral> &Participant::watchIntegrals()
-{
-  return _watchIntegrals;
 }
 
 void Participant::useMesh(
@@ -142,26 +141,16 @@ void Participant::addWriteMappingContext(
   _writeMappingContexts.push_back(mappingContext);
 }
 
-const utils::ptr_vector<MappingContext> &Participant::readMappingContexts() const
-{
-  return _readMappingContexts;
-}
+// Data queries
 
-const utils::ptr_vector<MappingContext> &Participant::writeMappingContexts() const
-{
-  return _writeMappingContexts;
-}
-
-const DataContext &Participant::dataContext(
-    int dataID) const
+const DataContext &Participant::dataContext(int dataID) const
 {
   PRECICE_ASSERT((dataID >= 0) && (dataID < (int) _dataContexts.size()));
   PRECICE_ASSERT(_dataContexts[dataID] != nullptr);
   return *_dataContexts[dataID];
 }
 
-DataContext &Participant::dataContext(
-    int dataID)
+DataContext &Participant::dataContext(int dataID)
 {
   PRECICE_TRACE(dataID, _dataContexts.size());
   PRECICE_ASSERT((dataID >= 0) && (dataID < (int) _dataContexts.size()));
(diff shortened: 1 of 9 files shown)


src/precice/impl/Participant.hpp
```diff
@@ -73,98 +72,178 @@ class Participant {
...
   bool isDataWrite(int dataID) const;
 
+  /// What is the dataID of the used data from a used mesh given the meshid and the data name?
+  int getUsedDataID(const std::string &dataName, int meshID) const;
+
+  /// What is the name of the given data id
+  std::string getDataName(int dataID) const;
+  /// @}
+
+  /// @name Mesh querries
```
DavidSCN, line 148:
```suggestion
  /// @name Mesh queries
```

src/precice/impl/Participant.hpp
```diff
@@ -73,98 +72,178 @@ class Participant {
...
+      const mesh::PtrMesh &                         mesh,
+      const Eigen::VectorXd &                       localOffset,
+      bool                                          remote,
+      const std::string &                           fromParticipant,
+      double                                        safetyFactor,
+      bool                                          provideMesh,
+      partition::ReceivedPartition::GeometricFilter geoFilter);
+  /// @}
+
+  /// @name Data querries
```
DavidSCN, line 106:
```suggestion
  /// @name Data queries
```

src/precice/impl/Participant.hpp
```diff
@@ -73,98 +72,178 @@ class Participant {
...
+  int getUsedDataID(const std::string &dataName, int meshID) const;
+
+  /// What is the name of the given data id
+  std::string getDataName(int dataID) const;
+  /// @}
+
+  /// @name Mesh querries
+  /// @{
+  /// Provides access to all \ref MeshContext where the index is meshID. Can contain nullptr!
   const MeshContext &meshContext(int meshID) const;
```
DavidSCN, line 151: Could you add some more documentation here? What does it mean when I get `nullptr` and how to interpret it. Is it a misuse?!

src/precice/impl/Participant.hpp
```diff
@@ -73,98 +72,178 @@ class Participant {
...
+  void addWatchIntegral(const PtrWatchIntegral &watchIntegral);
+
+  void setUseMaster(bool useMaster);
+
+  void setMeshIdManager(std::unique_ptr<utils::ManageUniqueIDs> &&idm)
+  {
+    _meshIdManager = std::move(idm);
+  }
+
+  void addAction(action::PtrAction &&action);
```
DavidSCN, line 90: Could you add a short description of what these function do, even when it seems obvious to you?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -212,25 +213,10 @@ void SolverInterfaceImpl::configure(
...
-      _dataIDs[meshID][data->getName()] = data->getID();
-    }
-    std::string                meshName   = mesh->getName();
-    mesh::PtrMeshConfiguration meshConfig = config.getMeshConfiguration();
-  }
   // Register all MeshIds to the lock, but unlock them straight away as
   // writing is allowed after configuration.
-  for (const auto &meshID : _meshIDs) {
-    _meshLock.add(meshID.second, false);
+  for (const MeshContext *meshContext : _accessor->usedMeshContexts()) {
```
DavidSCN, line 218: Why not using `const auto &meshContext` as it is done for other loops?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1064,12 +1049,11 @@ void SolverInterfaceImpl::writeBlockVectorData(
...
   PRECICE_CHECK(_state != State::Finalized, "writeBlockVectorData(...) cannot be called after finalize().");
-  PRECICE_VALIDATE_DATA_ID(dataID);
-  if (size == 0)
-    return;
+  PRECICE_REQUIRE_DATA_WRITE(dataID);
   PRECICE_ASSERT(valueIndices != nullptr);
   PRECICE_ASSERT(values != nullptr);
-  PRECICE_REQUIRE_DATA_WRITE(dataID);
+  if (size == 0)
+    return;
```
DavidSCN, line 1056: As far as I understand we should rather use `PRECICE_CHECK` here as these quantities are user defined and therefore user variables. We should also add a descriptive error message.
I'm not sure if moving the early return further down: in case the size is indeed zero, i.e., a rank doesn't need to write, `values` and `valueIndices` could be `nullptr` without doing something wrong.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1689,10 +1666,7 @@ void SolverInterfaceImpl::syncTimestep(double computedTimestepLength)
 const mesh::Mesh &SolverInterfaceImpl::mesh(const std::string &meshName) const
 {
   PRECICE_TRACE(meshName);
-  const MeshContext *context = _accessor->usedMeshContextByName(meshName);
-  PRECICE_ASSERT(context && context->mesh,
-                 "Participant \"" << _accessorName << "\" does not use mesh \"" << meshName << "\"!");
-  return *context->mesh;
+  return *_accessor->usedMeshContext(meshName).mesh;
```
DavidSCN, line 1669: Does it actually resolve #989 ?

src/precice/impl/ValidationMacros.hpp
```diff
@@ -90,37 +89,31 @@
...
  * @attention Do not use this macro directly!
  */
-#define PRECICE_REQUIRE_DATA_READ_IMPL(id)                                                                                  \
-  PRECICE_VALIDATE_DATA_ID_IMPL(id)                                                                                         \
-  DataContext &context = _accessor->dataContext(id);                                                                        \
-  PRECICE_CHECK((_accessor->isDataUsed(id) && _accessor->isDataRead(id)),                                                   \
-                "This participant does not use Data \"{}\", but attempted to read it. "                                     \
-                "Please extend the configuarion of partiticpant \"{}\" by defining <read-data mesh=\"{}\" name=\"{}\" />.", \
-                context.getName(), _accessorName, context.mesh->getName(), context.getName());
+#define PRECICE_REQUIRE_DATA_READ_IMPL(id)                                                                                      \
```
DavidSCN, line 100: May I ask you to provide a comment, maybe at the top of this file, why you split all the macros in an implementation part and a definition?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -601,19 +587,22 @@ bool SolverInterfaceImpl::hasMesh(
...
   PRECICE_TRACE(meshName);
-  const auto pos = _meshIDs.find(meshName);
-  PRECICE_CHECK(pos != _meshIDs.end(),
+  PRECICE_CHECK(_accessor->hasMesh(meshName),
                 "The given mesh name \"{}\" is unknown to preCICE. "
                 "Please check the mesh definitions in the configuration.",
                 meshName);
-  return pos->second;
+  PRECICE_CHECK(_accessor->isMeshUsed(meshName),
+                "The given mesh name \"{0}\" is not used by the preCICE. "
```
DavidSCN, line 602:
```suggestion
                "The given mesh name \"{0}\" is not used by the participant \"{1}\". "
```

src/precice/impl/Participant.hpp
```diff
@@ -73,98 +72,178 @@ class Participant {
...
 
-  std::vector<action::PtrAction> &actions();
+  /// @name Other querries
+  /// @{
+  /// Returns the name of the participant.
+  const std::string &getName() const;
 
-  const std::vector<action::PtrAction> &actions() const;
+  /// Returns true, if the participant uses a master process.
+  bool useMaster();
```
DavidSCN, line 221:
```suggestion
  bool useMaster() const;
```
? And is 'using' here equal to 'runs on'?

src/precice/impl/Participant.hpp
fsimonis: Thanks for the hint! The documentation was incorrect.

src/precice/impl/Participant.hpp
fsimonis: Documented what this actually mean. Note that I just reordered these declarations.

src/precice/impl/SolverInterfaceImpl.cpp
fsimonis: That's a good point.
Will make this consistent with the rest as I am already changing parts of these functions.

src/precice/impl/SolverInterfaceImpl.cpp
fsimonis: This function is used for tests only and not part of the SolverInterface.
#989 should be fixed with the changes in the ValidationMacros

src/precice/impl/Participant.cpp
```diff
@@ -189,54 +178,80 @@ utils::ptr_vector<DataContext> &Participant::readDataContexts()
...
+    if (mcptr) {
+      continue;
+    }
+    for (const auto &dptr : mcptr->mesh->data()) {
+      if (dptr->getID() == dataID) {
+        return dptr->getName();
+      }
+    }
+  }
+  PRECICE_UNREACHABLE("The dataID is invalid." << dataID);
```
DavidSCN, line 242: The `UNREACHABLE` macro uses `fmtlib`. Not completely sure here, but shouldn't you use the brackets `{}` in order to put the argument in the stream?

src/precice/impl/Participant.cpp
```diff
@@ -255,29 +270,117 @@ std::vector<MeshContext *> &Participant::usedMeshContexts()
...
+std::string Participant::getMeshNameFromData(int dataID) const
+{
+  for (const MeshContext *mcptr : _meshContexts) {
+    for (const auto &dptr : mcptr->mesh->data()) {
+      if (dptr->getID() == dataID) {
+        return mcptr->mesh->getName();
+      }
+    }
+  }
+  PRECICE_UNREACHABLE("The dataID is invalid." << dataID);
```
DavidSCN, line 371: Same as above

src/precice/impl/SolverInterfaceImpl.cpp
DavidSCN: Ok. I agree on making things explicit and I will disable `modernize-use-auto` for the tidy checks.

src/precice/impl/SolverInterfaceImpl.cpp
DavidSCN: I guess I was in the wrong line when adding my review comment here. The `DataContext` call is gone in the respective validation macro, so I guess the PR still resolves the issue in #989. Let me try it out once this has been merged.

DavidSCN (approved):
One last comment (see below). All in all, this PR greatly improves the code documentation, thanks!

src/precice/impl/Participant.hpp
DavidSCN: What about `const` ing the function? If I misunderstand it you can just resolve it.

DavidSCN: Ah got it, may bad

src/precice/impl/Participant.cpp
fsimonis: This branch was still based on the non-fmtlib version of the macro.

PR #1024, opened 2021-05-18
Simplify MasterSlave use-cases

fsimonis:
## Main changes of this PR

- Add range access for all ranks and slave ranks simplifing some loops
- Add isParallel which is equivalent to `isMaster() || isSlave()`.

We often need to implement functionality for the cases:
* Parallel
* Not parallel
* Master
* Slave

The frist two did not have an explicit query. Now one can use `utils::MasterSlave::isParallel()`.

Another error-prone part was the iteration over all ranks, or the ranks of all slaves.
This PR now proves integer ranges to simplify this.

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/acceleration/BaseQNAcceleration.cpp
+++ b/src/acceleration/BaseQNAcceleration.cpp
@@ -124,7 +124,7 @@ void BaseQNAcceleration::initialize(
    *  last entry _dimOffsets[MasterSlave::getSize()] holds the global dimension, global,n
    */
   std::stringstream ss;
-  if (utils::MasterSlave::isMaster() || utils::MasterSlave::isSlave()) {
+  if (utils::MasterSlave::isParallel()) {
     PRECICE_ASSERT(utils::MasterSlave::_communication.get() != nullptr);
     PRECICE_ASSERT(utils::MasterSlave::_communication->isConnected());
 
@@ -506,7 +506,7 @@ void BaseQNAcceleration::iterationsConverged(
 {
   PRECICE_TRACE();
 
-  if (utils::MasterSlave::isMaster() || (not utils::MasterSlave::isMaster() && not utils::MasterSlave::isSlave()))
+  if (utils::MasterSlave::isMaster() || !utils::MasterSlave::isParallel())
     _infostringstream << "# time window " << tWindows << " converged #\n iterations: " << its
                       << "\n used cols: " << getLSSystemCols() << "\n del cols: " << _nbDelCols << '\n';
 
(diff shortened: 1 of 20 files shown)


uekerman:
We should really rename master / slave to sth more appropriate. It's starting to be a bit embarrassing.

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -506,7 +506,7 @@ void BaseQNAcceleration::iterationsConverged(
 {
   PRECICE_TRACE();
 
-  if (utils::MasterSlave::isMaster() || (not utils::MasterSlave::isMaster() && not utils::MasterSlave::isSlave()))
+  if (utils::MasterSlave::isMaster() || !utils::MasterSlave::isParallel())
```
DavidSCN, line 509: Not sure if it is wise to change from `not` to `!`. I thought the `not` is our coding convention, although I like the `!` more. If you start changing this here you should probably also change it in other places.

src/partition/ReceivedPartition.cpp
```diff
@@ -585,7 +585,7 @@ void ReceivedPartition::createOwnerInformation()
     }
 
     // Send information back to slaves
-    for (int rank = 1; rank < utils::MasterSlave::getSize(); rank++) {
+    for (int rank : utils::MasterSlave::slaves()) {
```
DavidSCN, line 588: Maybe it would be better to find some consensus on how to name the iterator. It is not really intuitive that the loop above starts from zero whereas the loop here starts from one. Maybe we should even rename `::ranks()` into `::allRanks()` in order to emphasize that the master rank is included.

src/acceleration/BaseQNAcceleration.cpp
fsimonis: I try to change the `not` to `!` whenever I touch some code as it will simplify the [port to MSVC](https://stackoverflow.com/a/24414224) in the long run.

src/partition/ReceivedPartition.cpp
fsimonis: > It is not really intuitive that the loop above starts from zero whereas the loop here starts from one.

That is the reason I implemented `slaves()` and `ranks()`.

> Maybe we should even rename ::ranks() into ::allRanks() in order to emphasize that the master rank is included.

I was wondering about this too. `allRanks()` is good as it makes both the range  and the type explicit.
To keep things consistent I would then rename `slaves()` to `slaveRanks()`.

DavidSCN: > I was wondering about this too. allRanks() is good as it makes both the range and the type explicit.
To keep things consistent I would then rename slaves() to slaveRanks().

Agree :ok_hand:

Issue #980, opened 2021-03-18
Extend multi coupling to scenario without a centered participant


uekerman:
**Please describe the problem you are trying to solve.**

[(Fully implicit) multi coupling](https://www.precice.org/configuration-coupling-multi.html#fully-implicit-multi-coupling) is currently restricted to cases where one participant can act as the center. Consider the [multiple flap tutorial](https://github.com/precice/tutorials/tree/develop/multiple-perpendicular-flaps):
```xml
    <coupling-scheme:multi>
      <participant name="Fluid" control="yes" />
      <participant name="Solid1" />
      <participant name="Solid2" />
      <exchange data="Stress1" mesh="Fluid1-Mesh-Centers" from="Fluid" to="Solid1" />
      <exchange data="Stress2" mesh="Fluid2-Mesh-Centers" from="Fluid" to="Solid2" />
      <exchange data="Displacement1" mesh="Solid1-Mesh" from="Solid1" to="Fluid" />
      <exchange data="Displacement2" mesh="Solid2-Mesh" from="Solid2" to="Fluid" />
      ...
      <acceleration:IQN-ILS>
        <data name="Displacement1" mesh="Solid1-Mesh" />
        <data name="Displacement2" mesh="Solid2-Mesh" />
        <data name="Stress1" mesh="Fluid1-Mesh-Centers" />
        <data name="Stress2" mesh="Fluid2-Mesh-Centers" />
        ...
      </acceleration:IQN-ILS>
    </coupling-scheme:multi>
```
`Fluid` is the center. All exchanged data is sent to or from `Fluid`. That's why `Fluid` can compute the quasi-Newton acceleration with all data (`control="yes"`).
As soon as we add sth like ...
```xml
<exchange data="Temperature" mesh="Other-Solid-Mesh" from="Solid1" to="Solid2 />
```
... we get a problem as we have no distinct center any more. 

There are even more complicated case such as a chain of participants:
A <-> B <-> C <-> D

One participant needs to get all data to compute the quasi-Newton acceleration. Let that be C, for instance. But C now has not even a communication channel to A.

**Describe the solution you propose.**

It's part of the issue to find a solution :grin: 

An important decision we need to take is whether we rely on the user to add missing information in the configuration, such as the missing communication channel between A and C above or whether we create these for the user. Maybe starting with the first option is easier.

To start things, I recommend to implement a number of (failing) integration tests such that we fully agree on the variants we want to support. On this basis, we should then discuss different solutions.

Important implementation detail:
[...]

PR #1025, opened 2021-05-18
Multi-coupling extension

oguzziya:
## Main changes of this PR
This PR extends multi-coupling for arbitrary setups without any explicit controller.

Details are in #980 

## What is done until now
* Basic integration tests. Without any actual coupling, just configuration.
3 Solvers: A <=> B <=> C
4 Solvers: A <=> B <=> C <=> D

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/cplscheme/MultiCouplingScheme.cpp
+++ b/src/cplscheme/MultiCouplingScheme.cpp
@@ -27,20 +27,19 @@ MultiCouplingScheme::MultiCouplingScheme(
     double                        timeWindowSize,
     int                           validDigits,
     const std::string &           localParticipant,
-    std::vector<m2n::PtrM2N>      m2ns,
+    std::map<std::string, m2n::PtrM2N>      m2ns,
     constants::TimesteppingMethod dtMethod,
-    int                           maxIterations)
+    int                           maxIterations,
+    bool                          isController)
     : BaseCouplingScheme(maxTime, maxTimeWindows, timeWindowSize, validDigits, localParticipant, maxIterations, Implicit, dtMethod),
-      _m2ns(std::move(m2ns))
+      _m2ns(std::move(m2ns)), _isController(isController)
 {
   PRECICE_ASSERT(isImplicitCouplingScheme(), "MultiCouplingScheme is always Implicit.");
-  setDoesFirstStep(false); // MultiCouplingScheme never does the first step, because it is never the first participant
-  for (size_t i = 0; i < _m2ns.size(); ++i) {
-    DataMap receiveMap;
-    DataMap sendMap;
-    _receiveDataVector.push_back(receiveMap);
-    _sendDataVector.push_back(sendMap);
-  }
+   
+  // Controller participant never does the first step, because it is never the first participant
+  setDoesFirstStep(!_isController);
+
+  PRECICE_DEBUG("MultiCoupling scheme is created for {}.", localParticipant);
 }
 
 std::vector<std::string> MultiCouplingScheme::getCouplingPartners() const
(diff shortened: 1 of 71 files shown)


src/precice/tests/multi-coupling-three-solver.xml
```diff
@@ -0,0 +1,93 @@
...
+      <max-iterations value="2" />
+      <exchange data="DataBA" mesh="MeshA" from="SolverB" to="SolverA" />
+      <exchange data="DataAB" mesh="MeshA" from="SolverA" to="SolverB" />
+      <exchange data="DataCB" mesh="MeshC" from="SolverC" to="SolverB" />
+      <exchange data="DataBC" mesh="MeshC" from="SolverB" to="SolverC" />
+      <exchange data="DataCB" mesh="MeshC" from="SolverC" to="SolverA" />
+      <relative-convergence-measure data="DataCB" mesh="MeshC" limit="1e-4" />
+      <relative-convergence-measure data="DataBA" mesh="MeshA" limit="1e-4" />
+      <acceleration:IQN-ILS>
+        <data name="DataAB" mesh="MeshA" />
```
uekerman, line 85: We should add another testcase here, where all exchanged data is used in the acceleration (and maybe also a convergence measure exists for each of them).
... simply by iterating over configs in the test above (no additional code should be necessary).

uekerman (changes requested):
Looks very promising already. Was actually easier than I thought. I guess that the case with four participants should already work as well. Doesn't it?

What will not work is if you run the non-central controller in parallel. Could you please add a safeguard check for this? 
(You could already prepare a testcase on your side for this as well, we will need it later once https://github.com/DavidSCN/precice/pull/1 is merged to develop).

src/cplscheme/MultiCouplingScheme.cpp
```diff
@@ -88,64 +106,82 @@ bool MultiCouplingScheme::exchangeDataAndAccelerate()
...
+  m2n->receive(convergence);
+  return convergence;
+}
+
+void MultiCouplingScheme::sendConvergence(const m2n::PtrM2N &m2n, bool convergence)
+{
+  PRECICE_ASSERT((_isController), "For convergence information the sending participant is always the controller one.");
+  m2n->send(convergence);
+}
+
```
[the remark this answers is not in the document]
oguzziya: I think I overlooked it, you are right. However, `BaseCouplingScheme` only has `sendConvergence`. `BiCouplingScheme` defines `receiveConvergence` maybe I can carry it to `BaseCouplingScheme` too?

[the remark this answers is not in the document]
uekerman: I could imagine you can -- I did not check the details. Sounds like this was a previous MultiCouplingScheme artifact.

src/cplscheme/MultiCouplingScheme.cpp
```diff
@@ -35,7 +36,7 @@ MultiCouplingScheme::MultiCouplingScheme(
       _m2ns(std::move(m2ns)), _controller(controller), _isController(controller == localParticipant)
 {
   PRECICE_ASSERT(isImplicitCouplingScheme(), "MultiCouplingScheme is always Implicit.");
-
+  PRECICE_CHECK((!_isController and ((not utils::MasterSlave::isMaster()) and (not utils::MasterSlave::isSlave())) or _isController), "Non-controller participants in the MultiCouplingScheme cannot be run in parallel. Either run participant {} in serial or change the controller to another participant.", localParticipant);
```
uekerman, line 39: It's a bit more complicated than this. We cannot run those participants in parallel that need to receive a mesh that has no geometric association to any provided mesh. So the participant that has a `use-mesh from="OtherPart"` tag.
This could be the controller or any other participant.
Mmh, maybe check if there is already another error message that we would anyway run into.

oguzziya: We already hit an error in `ReceivedPartition` that the mesh needs a mapping. I am not sure how to check what you described actually. We can do it on the configuration level but I am not sure. Maybe for the time being we do not allow the parallel cases (for both controller and non-controller) and once the PR of David is done we can come back to this?

uekerman: > I am not sure how to check what you described actually.

Me neither :grin: 

> Maybe for the time being we do not allow the parallel cases (for both controller and non-controller)

I thought about this too. Problem is that this would forbid cases that were allowed previously (before this PR).

Maybe we can just remove this one here and just let the user run into the "needs a mapping" error for the moment.

uekerman (approved):
Nice job :+1:

PR #1030, opened 2021-05-31
Link to overview picture in docs/README changed to internal GitHub link

IshaanDesai:
Currently the preCICE README does not show the overview picture:

Screenshot from 2021-05-31 15-27-24

This PR manually copies the overview picture into the `precice/docs/` folder and then links to it. This way the image is displayed correctly.


IshaanDesai:
The originating branch needs to be changed from a fork to the original repository to modify the link to the image. But first the question: is this type of image addition reasonable?

MakisH (changes requested):
Why do we need to move the image here to solve the issue?

We definitely need to avoid duplication, as we are already using different outdated versions of this picture in different places (presentations, theses, etc). We currently offer this as a central resource: https://github.com/precice/precice.github.io/tree/master/material/overview

That also serves as a place where anybody could find common material to write about preCICE. I would not like to move the overview picture unless explicitly needed.

docs/README.md
```diff
@@ -40,7 +40,7 @@ preCICE stands for Precise Code Interaction Coupling Environment. Its main compo
 If you are new to preCICE, please have a look at our [documentation](https://www.precice.org/docs.html) and at [precice.org](https://www.precice.org). You may also prefer to [get and install a binary package for the latest release](https://github.com/precice/precice/releases/latest) ([master branch](https://github.com/precice/precice/tree/master)).
 
 <div align="center" style="margin-bottom:30px">
-<img src="https://www.precice.org/material/overview/precice-overview.png" alt="preCICE overview" style="max-height: 100%; max-width: 100%">
+<img src="https://github.com/IshaanDesai/precice/blob/overview-picture/docs/precice-overview.png" alt="preCICE overview" style="max-height: 100%; max-width: 100%">
```
MakisH, line 43: Reminder: this is a fork

IshaanDesai: Resolved by linking to: https://github.com/precice/precice.github.io/blob/master/material/overview/precice-overview.png

IshaanDesai:
> We definitely need to avoid duplication, as we are already using different outdated versions of this picture in different places (presentations, theses, etc). 

Duplication of the overview picture is avoided by linking to: https://github.com/precice/precice.github.io/blob/master/material/overview/precice-overview.png

It appears that linking via the preCICE website is not possible for GitHub anymore. Hence the link: https://precice.org/material/overview/precice-overview.png does not work but above mentioned link works.

MakisH (approved):
So, you are now linking directly to GitHub, instead of going through our domain name. This makes total sense!

PR #1033, opened 2021-06-22
Enable a direct access to received meshes

DavidSCN:
## Main changes of this PR
Add a feature to directly write and read from received meshes.

Depending on the discretization(s), you might be able to recover higher-order mappings in space.
Have a look at the following exemplary results [mapping_test.pdf](https://github.com/precice/precice/files/6691914/mapping_test.pdf).

A documentation on how to use this functionality is given in the [source code](https://github.com/DavidSCN/precice/blob/099781aec9aaa697a1596e8794073552de11e8ab/src/precice/SolverInterface.hpp#L775-L827).

## Open to discussion

* [ ] decide on API function names and xml configuration names

 
## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)


src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -261,13 +274,18 @@ void ParticipantConfiguration::xmlTagCallback(
...
     /// @todo offset currently not supported
     //offset = tag.getEigenVectorXdAttributeValue(ATTR_LOCAL_OFFSET, _dimensions);
-    const std::string &                           from         = tag.getStringAttributeValue(ATTR_FROM);
-    double                                        safetyFactor = tag.getDoubleAttributeValue(ATTR_SAFETY_FACTOR);
-    partition::ReceivedPartition::GeometricFilter geoFilter    = getGeoFilter(tag.getStringAttributeValue(ATTR_GEOMETRIC_FILTER));
+    std::string                                   from              = tag.getStringAttributeValue(ATTR_FROM);
+    double                                        safetyFactor      = tag.getDoubleAttributeValue(ATTR_SAFETY_FACTOR);
+    partition::ReceivedPartition::GeometricFilter geoFilter         = getGeoFilter(tag.getStringAttributeValue(ATTR_GEOMETRIC_FILTER));
+    const bool                                    allowDirectAccess = tag.getBooleanAttributeValue(ATTR_DIRECT_ACCESS);
+    ;
```
uekerman, line 281:
```suggestion
```
?

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -261,13 +274,18 @@ void ParticipantConfiguration::xmlTagCallback(
...
+    if (allowDirectAccess)
+      PRECICE_WARN("You configured the received mesh \"{}\" using the option access-direct=\"true\", which is experimental.", name);
```
uekerman, line 283:
```suggestion
      PRECICE_WARN("You configured the received mesh \"{}\" to use the option access-direct=\"true\", which is currently still experimental. Use with care.", name);
```

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -286,7 +304,15 @@ void ParticipantConfiguration::xmlTagCallback(
           "Please extend the use-mesh tag as follows: <use-mesh name=\"{}\" from=\"(other participant)\" />",
           _participants.back()->getName(), name, name);
     }
-    _participants.back()->useMesh(mesh, offset, false, from, safetyFactor, provide, geoFilter);
+
+    // TODO: Maybe merge later with the config check above
+    if ((allowDirectAccess == true) && from == "") {
```
uekerman, line 309:
```suggestion
    if ((allowDirectAccess == true) && (from == "")) {
```

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -286,7 +304,15 @@ void ParticipantConfiguration::xmlTagCallback(
...
+      PRECICE_ERROR("Participant \"{}\" uses mesh \"{}\", which is not received (no \"from\"), but has a direct access defined. "
+                    "Please extend the use-mesh tag as follows: <use-mesh name=\"{}\" from=\"(other participant)\" />",
+                    _participants.back()->getName(), name, name);
```
uekerman, line 312:
```suggestion
                    "This combination of options is not allowed. "
                    "Please extend the use-mesh tag as follows: <use-mesh name=\"{}\" from=\"(other participant)\" />"
                    " or remove the direct access option.",
                    _participants.back()->getName(), name, name);
```

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -286,7 +304,15 @@ void ParticipantConfiguration::xmlTagCallback(
           "Please extend the use-mesh tag as follows: <use-mesh name=\"{}\" from=\"(other participant)\" />",
           _participants.back()->getName(), name, name);
     }
-    _participants.back()->useMesh(mesh, offset, false, from, safetyFactor, provide, geoFilter);
+
+    // TODO: Maybe merge later with the config check above
```
uekerman, line 308:
```suggestion
```
I guess with our current xml implementation we cannot support such a configuration-time check. OK for me like this.

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -471,10 +497,11 @@ void ParticipantConfiguration::finishParticipantConfiguration(
...
     int fromMeshID = dataContext.mesh->getID();
-    PRECICE_CHECK(participant->isMeshProvided(fromMeshID),
-                  "Participant \"{}\" has to use and provide mesh \"{}\" to be able to write data to it. "
-                  "Please add a use-mesh node with name=\"{}\" and provide=\"true\".",
-                  participant->getName(), dataContext.mesh->getName(), dataContext.mesh->getName());
+
+    //    PRECICE_CHECK(participant->isMeshProvided(fromMeshID),
+    //                  "Participant \"{}\" has to use and provide mesh \"{}\" to be able to write data to it. "
+    //                  "Please add a use-mesh node with name=\"{}\" and provide=\"true\".",
+    //                  participant->getName(), dataContext.mesh->getName(), dataContext.mesh->getName());
```
uekerman, line 504: We need a solution here. 
Maybe 
```c++
participant->isMeshProvided(fromMeshID) || participant->isDirectAccesAllowed(fromMeshID)
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1258,6 +1258,7 @@ void SolverInterfaceImpl::readBlockScalarData(
                 data.getName());
   auto &     valuesInternal = data.values();
   const auto vertexCount    = valuesInternal.size();
+  // Open issue: how to handle filtered data sets? valuesInternal is set to zero initially
```
uekerman, line 1261: Let's not merge this like this. Will get forgotten. In my opinion the current solution is OK enough. Better open an issue if you want to revisit this in the future.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1297,6 +1298,74 @@ void SolverInterfaceImpl::readScalarData(
...
+    const double *boundingBox,
+    const int     nBoundingBoxes) const
+{
+  PRECICE_TRACE(meshID);
+  PRECICE_REQUIRE_MESH_USE(meshID);
+  if (nBoundingBoxes == 0)
+    return;
+  PRECICE_CHECK(boundingBox != nullptr, "The provided bounding boxes are empty.");
+  //  PRECICE_REQUIRE_MESH_MODIFY(meshID);
+  PRECICE_CHECK(nBoundingBoxes == 1, "Using more than one bounding box is not yet implemented.");
```
uekerman, line 1312:
```suggestion
  PRECICE_CHECK(nBoundingBoxes == 1, "Using more than one bounding box is not yet supported.");
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1297,6 +1298,74 @@ void SolverInterfaceImpl::readScalarData(
...
+    bounds[2 * d + 1] = boundingBox[2 * d + 1];
+  }
+  // Create a bounding box
+  mesh::BoundingBox providedBoundingBox(bounds);
+  // Expand the mesh associated bounding box
+  mesh::BoundingBox &bbox = mesh->getBoundingBox();
+  bbox                    = std::move(providedBoundingBox);
+
+  // Do not allocate data values
+  // mesh->allocateDataValues();
```
uekerman, line 1335: Should we need this?

src/precice/impl/SolverInterfaceImpl.hpp
```diff
@@ -469,6 +469,36 @@ class SolverInterfaceImpl {
       int     valueIndex,
       double &value) const;
 
+  /**
+   * @brief setBoundingBox Define a region of interest in order to filter a
+   *        received mesh for a certain mesh region
+   *
+   * @param[in] boundingBox (axis aligned) which has in 3D the format
+   *            [x_min, x_max, y_min, y_max, z_min, z_max]
```
uekerman, line 477: Inconsistent with definition in `SolverInterface` (here only one bounding box)

src/precice/tests/ParallelTests.cpp
```diff
@@ -867,6 +867,248 @@ BOOST_AUTO_TEST_CASE(TestBoundingBoxInitialization)
...
+  const std::vector<double> writeDataSlave        = std::vector<double>({4, 5, 6});
+  const std::vector<double> expectedReadDataSlave = std::vector<double>({3, 4, 5});
+  runTestAccessReceivedMesh(_pathToTests + "explicit-access-direct.xml", boundingBoxSlave, writeDataSlave, expectedPositionSlave, expectedReadDataSlave, 1);
+}
+
+// Test case for a direct mesh access on one participant to a mesh defined
+// by another participant (see above). In addition to the direct mesh access
+// and data writing in one direction, an additional mapping (NN) is defined
+// in the other direction.
+BOOST_AUTO_TEST_CASE(LAccessReceivedMeshAndMapping)
```
uekerman, line 1017:
```suggestion
BOOST_AUTO_TEST_CASE(AccessReceivedMeshAndMapping)
```
?

src/precice/tests/ParallelTests.cpp
```diff
@@ -867,6 +867,248 @@ BOOST_AUTO_TEST_CASE(TestBoundingBoxInitialization)
...
+// StartIndex is here the first index to be used for writing on the slave rank
+void runTestAccessReceivedMesh(const std::string         configName,
+                               const std::vector<double> boundingBoxSlave,
+                               const std::vector<double> writeDataSlave,
+                               const std::vector<double> expectedPositionSlave,
+                               const std::vector<double> expectedReadDataSlave,
+                               const int                 startIndex)
+{
+
+  PRECICE_TEST("SolverOne"_on(2_ranks), "SolverTwo"_on(2_ranks));
```
uekerman, line 879: Wouldn't we be able to test for edge case at once if we use a 3-1 distribution?

src/precice/tests/explicit-access-direct-mapping.xml
```diff
@@ -0,0 +1,44 @@
...
+    </mesh>
+
+    <mesh name="MeshTwo">
+      <use-data name="Velocities" />
+      <use-data name="Forces" />
+    </mesh>
+
+    <participant name="SolverOne">
+      <use-mesh name="MeshOne" provide="yes" />
+      <use-mesh name="MeshTwo" from="SolverTwo" safety-factor="0" direct-access="true" />
```
uekerman, line 18:
```suggestion
      <use-mesh name="MeshTwo" from="SolverTwo" direct-access="true" />
```
See code discussion above.

uekerman (changes requested):
We are not that far away from merging, but there are still some usability aspects that need proper discussion. 
I am not 100% sure about all comments below. Sometimes, I suggest changes just to make us rethink the current solution.
I have not yet looked carefully at the tests. However, I think it would be good to also add a parallel 2-2 test that uses the two-level initialization. We might interfere with things there as we also touch and use the bounding boxes.
As we touch the API quite a bit I would also like to get input from @fsimonis on this PR.

src/partition/ReceivedPartition.cpp
```diff
@@ -93,6 +94,28 @@ void ReceivedPartition::compute()
...
   if (!utils::MasterSlave::isParallel()) { //coupling mode
     PRECICE_DEBUG("Handle partition data structures for serial participant");
+
+    if (_allowDirectAccess) {
+      // Prepare the bounding boxes
+      prepareBoundingBox();
+      // Filter out vertices not laying in the bounding box
+      mesh::Mesh filteredMesh("FilteredMesh", _dimensions, mesh::Mesh::MESH_ID_UNDEFINED);
+      // To discuss: maybe check this somewhere in the SolverInterfaceImpl, as we have now a similar check for the parallel case
+      PRECICE_CHECK(!_bb.empty(), "You are running in serial mode and the bounding box on mesh \"{}\", is empty. Did you call setBoundingBox with valid data?", _mesh->getName());
```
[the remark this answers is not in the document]
DavidSCN: That is a conceptual question. Since I am filtering the mesh according to the local bounding box, continuation does not make any sense without a defined bounding box. So, aborting would be the way to go here. I would keep the required bounding box filtering and definition for consistency reasons. If someone really wants to use preCICE in serial mode only and does not want to compute or provide any bounding box he can set a 'big enough' bounding box.
However, I am open for discussions here: I can also remove the filter here completely.

src/partition/ReceivedPartition.cpp
```diff
@@ -93,6 +94,28 @@ void ReceivedPartition::compute()
...
+      unsigned int nFilteredVertices = 0;
+      mesh::filterMesh(filteredMesh, *_mesh, [&](const mesh::Vertex &v) { if(!_bb.contains(v))
+              ++nFilteredVertices;
+          return _bb.contains(v); });
+
+      if (nFilteredVertices > 0)
+        PRECICE_WARN("{} vertices have been filtered out "
+                     "in serial mode and will be filled with zero values.",
+                     nFilteredVertices);
```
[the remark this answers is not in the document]
DavidSCN: Not sure if 'partitioning' is the right word for serial mode. I added additional information.

src/partition/ReceivedPartition.cpp
```diff
@@ -462,6 +485,22 @@ void ReceivedPartition::prepareBoundingBox()
...
+    auto &other_bb = _mesh->getBoundingBox();
+    _bb.expandBy(other_bb);
+    // TODO: How to treat the scale by safety factor here?
+    // The default value of 0.5 is for the current mappings
+    // still too large in comparison to what we want to do here.
+    // On the other hand, ignoring it makes the xml configuration
+    // inconsistent and does not provide any configurable option for
+    // 'bad' matching geometries.
+    //    _bb.scaleBy(_safetyFactor);
+    PRECICE_WARN("Ignoring the safety factor for bounding box initialization");
```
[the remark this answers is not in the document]
DavidSCN: What do you do then if you have bad matching geometries and a safety factor is actually desired. It would still be inconsistent with the precice-config, right? The scaling is applied anyway for all mappings a few lines above.

src/partition/ReceivedPartition.cpp
```diff
@@ -604,9 +645,12 @@ void ReceivedPartition::createOwnerInformation()
     }
 #endif
     auto filteredVertices = std::count(globalOwnerVec.begin(), globalOwnerVec.end(), 0);
-    if (filteredVertices)
+    if (filteredVertices) {
       PRECICE_WARN("{} of {} vertices of mesh {} have been filtered out since they have no influence on the mapping.",
                    filteredVertices, _mesh->getGlobalNumberOfVertices(), _mesh->getName());
+      if (_allowDirectAccess)
+        PRECICE_WARN("Filtered vertices will be filled with the zero data.");
+    }
```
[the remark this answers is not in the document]
DavidSCN: The difference is that in case a mapping is defined we know that we have conservative data when it becomes relevant.

src/precice/SolverInterface.hpp
```diff
@@ -772,6 +772,83 @@ class SolverInterface {
...
+   *    double dt = precice.initialize();
+   *    // Get the size of the filtered mesh within the bounding box
+   *    // (provided by the coupling participant)
+   *    const int otherMeshSize = precice.getMeshVertexSize(writeDataID);
+   *
+   *    // Allocate a vector for the vertices
+   *    std::vector<double> otherSolverMesh(otherMeshSize * dim);
+   *    precice.getMeshVerticesWithIDs(otherMeshID, otherMeshSize, ids.data(), otherSolverMesh.data());
+   *    // continue with time loop and write data directly to writeDataID
+   * @endcode
```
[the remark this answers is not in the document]
DavidSCN: > Do we use this "@code" environment already?

That's a doxygen command. Doxygen will render this part as a code-block in our API documentation. Since this feature will be experimental in the first step, we could host it here as well?!

src/precice/SolverInterface.hpp
```diff
@@ -772,6 +772,83 @@ class SolverInterface {
...
+   *        Defining a bounding box for serial runs of the solver (not to
+   *        be confused with serial coupling mode) is valid. However, an
+   *        exception is thrown (TODO) in case vertices are filtered out
+   *        completely on the receiving side as long as this is not explicitly
+   *        enabled in the configuration file (TODO).
+   *        The mesh region is specified through a collection of axis-aligned
+   *        bounding boxes given by the lower and upper [min and max]
+   *        bounding-box limits in each space dimension [x, y, z].
+   *
+   * @param[in] dataID ID of the Data you want to access through the bounding box
```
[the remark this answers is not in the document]
DavidSCN: You are right

src/precice/SolverInterface.hpp
```diff
@@ -772,6 +772,83 @@ class SolverInterface {
...
+   * @param[in] boundingBoxes collection of (axis aligned) bounding boxes which
+   *            has in 3D the format
+   *            [x_min0, x_max0, y_min0, y_max0, z_min0, z_max0, x_min1 ... ]
+   * @param[in] nBoundingBoxes number of bounding boxes.
```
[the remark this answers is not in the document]
DavidSCN: I don't like it since we have so many sizes, but I will change it.

src/precice/tests/explicit-access-direct-mapping.xml
DavidSCN: The test would fail if I change this line. Why should I change it?

src/precice/SolverInterface.hpp
```diff
@@ -772,6 +772,83 @@ class SolverInterface {
...
+   *    // Allocate a vector for the vertices
+   *    std::vector<double> otherSolverMesh(otherMeshSize * dim);
+   *    precice.getMeshVerticesWithIDs(otherMeshID, otherMeshSize, ids.data(), otherSolverMesh.data());
+   *    // continue with time loop and write data directly to writeDataID
+   * @endcode
+   *        Defining a bounding box for serial runs of the solver (not to
+   *        be confused with serial coupling mode) is valid. However, an
+   *        exception is thrown (TODO) in case vertices are filtered out
+   *        completely on the receiving side as long as this is not explicitly
+   *        enabled in the configuration file (TODO).
```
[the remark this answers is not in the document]
DavidSCN: I adjusted the documentation here according to the current state. It might change again, tough, so let us keep this unresolved for now as a reminder.

src/mesh/Mesh.hpp
```diff
@@ -210,6 +209,8 @@ class Mesh {
    */
   const BoundingBox &getBoundingBox() const;
 
+  BoundingBox &getBoundingBox();
```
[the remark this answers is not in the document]
DavidSCN: I added now an `expandBoundingBox` function instead.

src/precice/impl/SolverInterfaceImpl.cpp
DavidSCN: No

src/precice/tests/ParallelTests.cpp
DavidSCN: I don't really understand which edge cases you have in mind.

DavidSCN:
> However, I think it would be good to also add a parallel 2-2 test that uses the two-level initialization. We might interfere with things there as we also touch and use the bounding boxes.

I could simply copy the current parallel tests and let them run with two-level-initialization=true. Is this what you have in mind or do you have something more specific in mind?

uekerman:
We should also add clear "this is experimental" statements in the API description. To avoid that users misunderstand the new calls.
I guess it would then be OK to still change these API functions without calling it a breaking change?

src/partition/ReceivedPartition.cpp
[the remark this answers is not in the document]
uekerman: I guess we can't say yet what would be better. Should we keep this open for the moment?
I have a tendency towards allowing no bounding box, but I have no good arguments yet :grin: 
Implementation-wise, the current solution is simpler, right?

src/partition/ReceivedPartition.cpp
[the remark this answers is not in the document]
uekerman: Looks good now.

src/partition/ReceivedPartition.cpp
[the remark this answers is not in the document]
uekerman: Ah, OK. I did not see the scaling above. Then, here, I would not scale at all.
In my opinion, this should be the job of the user / the adapter. You define the bounding box, such that you have everything that interests you (i.e. you expand the bounding box by your own ghost layer or similar, everything else goes to another rank). Independent of how the other mesh looks like.
I agree though that we need a better solution for the preCICE config in the long run, like #1040.

src/partition/ReceivedPartition.cpp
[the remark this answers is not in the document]
uekerman: > The difference is that in case a mapping is defined we know that we have conservative data when it becomes relevant.

True.

Looks good now.

src/precice/SolverInterface.hpp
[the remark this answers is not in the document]
uekerman: I would prefer to have it on the website. There, we already have all means to mark it experimental (like a big warning block).
And we can properly explain how everything is connected (API, configuration option). I will need such an explanation to not forget already :grin:  
Otherwise, I am bit afraid that we will end up with duplicated examples.

src/precice/tests/ParallelTests.cpp
uekerman: Nothing specific. We would just have more flexibility:
- One rank empty, the other two overlapping
- Three ranks overlapping

Maybe these cases are not so relevant.

src/precice/tests/explicit-access-direct-mapping.xml
uekerman: OK, the safety factor belongs to the mesh below.
Just to be sure: the direct access does not use the safety factor, right?

src/precice/SolverInterface.hpp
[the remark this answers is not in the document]
DavidSCN: I removed the code example, referred to the website and kept the 'conceptual' description in the `SolverInterface`

src/partition/ReceivedPartition.cpp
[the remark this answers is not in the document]
DavidSCN: Simpler in the sense of consistent between serial and parallel execution. 

> I have a tendency towards allowing no bounding box

We should not require the user to change the code path when the same code is executed in serial or in parallel. This makes things just complicated. 'Allowing no bounding box' in serial and 'require' one in parallel would be hard to understand and not very user-friendly.

src/partition/ReceivedPartition.cpp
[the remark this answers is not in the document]
DavidSCN: Ok,  should we then raise a warning here or should we silently skip the scaling?

src/precice/tests/explicit-access-direct-mapping.xml
DavidSCN: It belongs more or less to the NN mapping below. The direct access does not use the safety factor, correct.

DavidSCN:
I added now the two-level initialization tests and added another test case with an empty partition. I hope this is enough for now as I'm not very eager to rewrite the parallel tests. The documentation on the website is now also ready for review.

src/precice/SolverInterface.hpp
```diff
@@ -782,7 +782,9 @@ class SolverInterface {
   /**
    * @brief setBoundingBoxes Define a region of interest on a received mesh
    *        (<use-mesh ... from="otherParticipant />") in order to receive
-   *        only a certain mesh region.
+   *        only a certain mesh region. Have a look at the website under
+   *        [sidebar]->Couple your code -> Advanced topics -> Accessing
+   *        received meshes directly for a comprehensive documentation
```
uekerman, line 787: Why not give the url? We will probably not change it in the future.

src/partition/ReceivedPartition.cpp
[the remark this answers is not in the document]
uekerman: Let's keep it as it is for the moment.

src/partition/ReceivedPartition.cpp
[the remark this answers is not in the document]
uekerman: I would silently skip the scaling. Maybe we can make this explicit in the documentation of `safety-factor`?

src/precice/SolverInterface.hpp
DavidSCN: Added..

src/mesh/Mesh.cpp
```diff
@@ -209,7 +209,10 @@ void Mesh::allocateDataValues()
 void Mesh::computeBoundingBox()
 {
   PRECICE_TRACE(_name);
-  BoundingBox bb(_dimensions);
+
+  // Keep the bounding box if set via the API function.
+  BoundingBox bb = _boundingBox.empty() ? BoundingBox(_dimensions) : BoundingBox(_boundingBox);
```
fsimonis, line 214: If the user sets a BB, then this Mesh should be empty and expanding it is a noop, right?
In that case, you could simply exit the function `return;`

How does this impact tests and the partitioning where `computeBoundingBox()` may be called multiple times on the same mesh?
Recomputing the BB will then be interpreted as "user set it via API".

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1297,6 +1298,70 @@ void SolverInterfaceImpl::readScalarData(
...
+    const int     meshID,
+    const double *boundingBox,
+    const int     size) const
+{
+  PRECICE_TRACE(meshID);
+  PRECICE_REQUIRE_MESH_USE(meshID);
+  if (size == 0)
+    return;
+  PRECICE_CHECK(boundingBox != nullptr, "The provided bounding boxes are empty.");
+  //  PRECICE_REQUIRE_MESH_MODIFY(meshID);
```
fsimonis, line 1311: What are the requirements in this case?
It would be good to define them here and come up with a similar macro.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1297,6 +1298,70 @@ void SolverInterfaceImpl::readScalarData(
...
+  // Get the related mesh
+  MeshContext & context = _accessor->meshContext(meshID);
+  mesh::PtrMesh mesh(context.mesh);
+  PRECICE_DEBUG("Define bounding box");
+  // Transform bounds into a suitable format
+  int                 dim = mesh->getDimensions();
+  std::vector<double> bounds(dim * 2);
+
+  for (int d = 0; d < dim; ++d) {
+    // Assert that min is lower or equal to max
```
fsimonis, line 1323: It doesn't assert a valid program state, it checks the user input. :wink:

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1297,6 +1298,70 @@ void SolverInterfaceImpl::readScalarData(
   PRECICE_DEBUG("Read value = {}", value);
 }
 
+void SolverInterfaceImpl::setBoundingBoxes(
+    const int     meshID,
+    const double *boundingBox,
```
fsimonis, line 1303: The name differs from the header.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1297,6 +1298,70 @@ void SolverInterfaceImpl::readScalarData(
...
+  for (int d = 0; d < dim; ++d) {
+    // Assert that min is lower or equal to max
+    PRECICE_CHECK(boundingBox[2 * d] <= boundingBox[2 * d + 1], "Your bounding box is ill defined, i.e. it has a negative volume. The required format is [x_min, x_max...]");
+    bounds[2 * d]     = boundingBox[2 * d];
+    bounds[2 * d + 1] = boundingBox[2 * d + 1];
+  }
+  // Create a bounding box
+  mesh::BoundingBox providedBoundingBox(bounds);
+  // Expand the mesh associated bounding box
+  mesh->expandBoundingBox(providedBoundingBox);
```
fsimonis, line 1331: So calling this multiple times will expand the bounding box.
Ideally, this would keep a collection of BBs, right?

This can be implement quite easily using a hierarchical data structure. (This is very common in physics.)

src/partition/ReceivedPartition.cpp
```diff
@@ -93,6 +94,28 @@ void ReceivedPartition::compute()
...
+      // Filter out vertices not laying in the bounding box
+      mesh::Mesh filteredMesh("FilteredMesh", _dimensions, mesh::Mesh::MESH_ID_UNDEFINED);
+      // To discuss: maybe check this somewhere in the SolverInterfaceImpl, as we have now a similar check for the parallel case
+      PRECICE_CHECK(!_bb.empty(), "You are running this participant in serial mode and the bounding box on mesh \"{}\", is empty. Did you call setBoundingBoxes with valid data?", _mesh->getName());
+      unsigned int nFilteredVertices = 0;
+      mesh::filterMesh(filteredMesh, *_mesh, [&](const mesh::Vertex &v) { if(!_bb.contains(v))
+              ++nFilteredVertices;
+          return _bb.contains(v); });
+
+      if (nFilteredVertices > 0)
```
fsimonis, line 110: Please wrap the block `{}`

src/partition/ReceivedPartition.cpp
```diff
@@ -462,6 +485,23 @@ void ReceivedPartition::prepareBoundingBox()
...
+  if (_allowDirectAccess) {
+    auto &other_bb = _mesh->getBoundingBox();
+    _bb.expandBy(other_bb);
+    // TODO: How to treat the scale by safety factor here?
+    // The default value of 0.5 is for the current mappings
+    // still too large in comparison to what we want to do here.
+    // On the other hand, ignoring it makes the xml configuration
+    // inconsistent and does not provide any configurable option for
+    // 'bad' matching geometries.
+    //    _bb.scaleBy(_safetyFactor);
```
fsimonis, line 499: IMHO you have 2 options:
* use the safety factor here and warn the user when handling a `<use-mesh />` tag.

> Warning: You requested direct access to the mesh "X" without specifying a safety-factor. The default will be used to enlarge your BB.

> Info: You requested direct access to the mesh "X". The safety-factor will not be applied.

* ignore the safety factor here and throw an error if the user asks for direct access and specifies a safety factor when handling a <use-mesh />` tag.



> Error: You requested direct access to the mesh "X" and specified a "safety-factor". This is not allowed, please adjust your BB accordingly prior to calling setBoundingBoxes.

src/partition/ReceivedPartition.cpp
```diff
@@ -556,6 +596,9 @@ void ReceivedPartition::createOwnerInformation()
 
     // Decide upon owners,
     PRECICE_DEBUG("Decide owners, first round by rough load balancing");
+    // Provide a more descriptive error message if direct access was enabled
+    PRECICE_CHECK(!(ranksAtInterface == 0 && _allowDirectAccess), "After repartitioning of mesh \"{}\" all ranks are empty. Did you forget to call \"setBoundingBoxes\" with valid data?", _mesh->getName());
```
fsimonis, line 600: >  Did you forget to call \"setBoundingBoxes\" with valid data?

Made me laugh :joy: 

Suggestion:

> Please check the dimensions of the provided bounding box and verify that it covers vertices in the mesh.

src/partition/ReceivedPartition.cpp
```diff
@@ -604,9 +647,12 @@ void ReceivedPartition::createOwnerInformation()
     }
 #endif
     auto filteredVertices = std::count(globalOwnerVec.begin(), globalOwnerVec.end(), 0);
-    if (filteredVertices)
+    if (filteredVertices) {
       PRECICE_WARN("{} of {} vertices of mesh {} have been filtered out since they have no influence on the mapping.",
                    filteredVertices, _mesh->getGlobalNumberOfVertices(), _mesh->getName());
+      if (_allowDirectAccess)
+        PRECICE_WARN("Associated data values of the filtered vertices will be filled with zero values in order to provide valid data for other participants when reading data.");
```
fsimonis, line 654: You could integrate this warning in the previous message by adding a `{}` at the end.
The you can pass a empty string or this message.

fsimonis (changes requested):
Check everything but the tests.

src/precice/SolverInterface.hpp
```diff
@@ -774,6 +774,75 @@ class SolverInterface {
...
+   *
+   * @param[in]  meshID corresponding mesh ID
+   * @param[in]  size return value of getMeshSize
+   * @param[out] ids ids corresponding to the coordinates
+   * @param[out] coordinates associated to the values (dim * @p getLocallyRelevantMeshSize)
+   *
+   * @pre IDs and coordinates need to have the correct size, which can be requested by getMeshVertexSize)
+   * @pre bounding box has been defined using @p setBoundingBoxes
+   */
+  void getMeshVerticesAndIDs(
```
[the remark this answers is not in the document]
DavidSCN: > The name doesn't reflect the precondition that the BBs must be set.

That's not the purpose of the name. 

>  The precondition is also very strict. We could simply do nothing if BBs haven't been set yet.

Keeping it strict helps to detect misuse. 

> Also Vertices can be both the coords and the ids, so the AndIDs is not required here.

As you already noted, `vertices` is used for the coordinates in other API functions as well, so that the `AndIDs` makes sense in my opinion. 

> We already have getMeshVertices which returns the coords given ids.
Wouldn't it suffice to add a getMeshMarkedVertexIDs or similar?
Then the user can first get the ids in the bb of the mesh, and then their coordinates.

In my opinion, this makes things for the user more complex. Still open to change it.

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -167,6 +167,19 @@ ParticipantConfiguration::ParticipantConfiguration(
                            .setDefaultValue(VALUE_FILTER_ON_SLAVES);
   tagUseMesh.addAttribute(attrGeoFilter);
 
+  auto attrDirectAccess = makeXMLAttribute(ATTR_DIRECT_ACCESS, false)
+                              .setDocumentation(
```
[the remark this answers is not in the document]
DavidSCN: This is a safeguard in order to tell people that we have an experimental feature. In addition, configuration errors are in my opinion easier to fix and easier to find. People should actively think about these feature.

src/precice/impl/Participant.hpp
```diff
@@ -238,6 +238,10 @@ class Participant {
   /// Get the used mesh id of a mesh with this name.
   int getUsedMeshID(const std::string &meshName) const;
 
+  /// Returns whether we are allowed to access a received mesh direct
+  /// which requires the config tag <use-mesh ... direct-access="true"
+  bool isDirectAccessAllowed(const int meshID) const;
```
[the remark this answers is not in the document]
DavidSCN: It adds a value for the (code) reader as the variable is still `const` within the function. IIRC, we already had the discussion, you can read the lengthy answer here https://stackoverflow.com/questions/117293/use-of-const-for-function-parameters

src/precice/SolverInterface.hpp
```diff
@@ -774,6 +774,75 @@ class SolverInterface {
...
+   * bounding boxes given by the lower and upper [min and max]
+   * bounding-box limits in each space dimension [x, y, z].
+   *
+   * @param[in] meshID ID of the mesh you want to access through the bounding box
+   * @param[in] boundingBoxCollection collection of (axis aligned) bounding boxes which
+   *            has in 3D the format
+   *            [x_min0, x_max0, y_min0, y_max0, z_min0, z_max0, x_min1 ... ]
+   * @param[in] size number of bounding boxes.
+   *
+   * @warning This functions is currently only implemented for nBoundingBoxes = 1.
```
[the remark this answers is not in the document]
DavidSCN: I could do that, but then I need to change the API later again.

src/precice/SolverInterface.hpp
```diff
@@ -774,6 +774,75 @@ class SolverInterface {
...
+   *
+   * @pre 'initialize' has not yet been called.
```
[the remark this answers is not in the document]
uekerman: That's a good point. We have not yet defined this, but we should. I guess, currently, we override data structures. Maybe we can instead skip the "array of bounding boxes" and just allow multiple calls.

src/precice/SolverInterface.hpp
```diff
@@ -774,6 +774,75 @@ class SolverInterface {
...
+   */
+  void setBoundingBoxes(
```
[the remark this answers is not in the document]
uekerman: I don't see any confusion for users as well. In particular, as we mean the same kind of bounding boxes, we also use in the partitioning. We use these bounding boxes for the re-partitioning.

Still, I like some of the suggestions. `setMeshAccessRegion` sounds good (better than `setBoundingBox`) if we only allow a single call of the function. 

I don't like "mark" as it would indicate that sth is already there and we just tag it. But here nothing is there yet.

src/precice/SolverInterface.hpp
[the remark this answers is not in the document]
uekerman: Indeed, in other API functions we use "vertices" only to refer to the coordinates. So, I think the additional "AndIDs" is good.

About the precondition: do we actually need it? Better would be if this function works for all meshes (which I guess is possible, maybe it even does already, (we would need integration tests for this then though)). Then, there is no room for confusion and less complicated things to document. Do you agree?

src/precice/config/ParticipantConfiguration.cpp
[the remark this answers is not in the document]
uekerman: Yes, this is sth @DavidSCN and me discussed. It's a safeguard for the moment (there are still some variants that do not work properly the user does not know of). Maybe we can (or should) get rid of the config option in the long run.

src/precice/SolverInterface.hpp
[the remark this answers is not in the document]
uekerman: We agreed that changing the API later would not be a big issue. So, I am in favor of making things simpler here. What we need to decide (see above) is whether this function can be called multiple times or not.

src/m2n/PointToPointCommunication.cpp
```diff
@@ -612,8 +612,6 @@ void PointToPointCommunication::receive(double *itemsToReceive,
   std::fill(itemsToReceive, itemsToReceive + size, 0);
 
   for (auto &mapping : _mappings) {
-    // if (not utils::MasterSlave::isMaster())
-    //   std::cout<< "indices " << mapping.indices << std::endl;
```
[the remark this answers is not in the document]
uekerman: Yes. Still, I guess not need to change it now (?). More a remark for future PRs.

src/mesh/Mesh.cpp
uekerman: > If the user sets a BB, then this Mesh should be empty and expanding it is a noop, right?

Not. setting BBs and using local meshes can be combined. Example: You want to have a read mapping from the received to your local mesh, but you want to directly write into the received mesh. And you need a larger part of the received mesh for writing than the read mapping needs.

src/partition/ReceivedPartition.cpp
uekerman: When combining direct access with a mapping, it's unfortunately more complicated. We need to safety factor for the mapping, but we should not use it for the "direct access region". The best solution, in my opinion, is to get rid of the safety factor configuration in the long run, but compute it based on the mapping (beyond this PR).

uekerman:
Let's not forget to also add the new API functions to the C and Fortran bindings. If not in this PRs, please open an issue. In this issue you could then also set a reminder that we need to port it to the external bindings as well.

src/precice/SolverInterface.hpp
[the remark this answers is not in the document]
DavidSCN: Yes, the function can repeatedly be called (although I don't see any sense in doing so) that's not a problem as we just expand the bounding box of the current mesh:

```cpp
  mesh->expandBoundingBox(providedBoundingBox);
```

Arrays of boxes are not implemented anyway, but should not affect this here.

src/precice/SolverInterface.hpp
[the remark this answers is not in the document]
DavidSCN: Renamed it now to `setMeshAccessRegion`

src/precice/SolverInterface.hpp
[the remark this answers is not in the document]
DavidSCN: We will only allow a single call of this function.

src/precice/impl/SolverInterfaceImpl.cpp
DavidSCN: Discussing with @uekerman: we decided to drop the collection idea here for the first steps. An extension might come in the future.

src/precice/SolverInterface.hpp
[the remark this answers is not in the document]
DavidSCN: Done

src/precice/SolverInterface.hpp
```diff
@@ -774,6 +774,75 @@ class SolverInterface {
...
+   * single mesh (from the other participant) is now involved in this
+   * situation since an 'own' mesh defined by the participant itself
+   * is not required any more. In order to re-partition the received
+   * mesh, the participant needs to define the mesh region it wants
+   * read data from and write data to.
+   * Defining a bounding box for serial runs of the solver (not to
+   * be confused with serial coupling mode) is valid. However, a
+   * warning is raised in case vertices are filtered out completely
+   * on the receiving side, since the associated data values of the
+   * filtered vertices are filled with zero data.
```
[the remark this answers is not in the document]
DavidSCN: Done

src/precice/impl/SolverInterfaceImpl.cpp
DavidSCN: Fixed

src/partition/ReceivedPartition.cpp
DavidSCN: Done.

src/precice/SolverInterface.hpp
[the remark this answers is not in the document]
DavidSCN: Agree and I will remove the precondition. Until now, I only considered this API call for this use-case. There might be others.

DavidSCN:
TODO list:

- [ ] add integration tests for `getMeshVertices` independent of the direct access flag
- [ ] add API functions to language bindings

src/partition/ReceivedPartition.cpp
DavidSCN: Giving the user the information that it is actually ignored is a good idea though. I will try to raise a warning in this case (not an error) and write some comments in the source code.

src/precice/impl/SolverInterfaceImpl.cpp
DavidSCN: The requirements are calling it before `initialize`, calling it only once and the mesh needs to be used. All conditions are checked now, I hope that's sufficient.

DavidSCN:
I (hopefully) addressed now all points and everything is prepared for another review round.

src/precice/impl/Participant.hpp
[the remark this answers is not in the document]
fsimonis: It doesn't make sense in the declaration though. It's immaterial to the caller if this is `const` or not.
Feel free to keep it in the definition though.

src/mesh/Mesh.cpp
fsimonis: And how does it affect recomputing the boundingboxes?

extras/bindings/c/include/precice/SolverInterfaceC.h
```diff
@@ -450,6 +450,31 @@ const char *precicec_actionWriteIterationCheckpoint();
...
 
+///@}
+
+/** @name Experimental Data Access
+ * These API functions are \b experimental and may change in future versions.
+ */
+///@{
+
+/**
+ * @brief Have a look at the c++ API documentation
```
fsimonis, line 461: You can directly link to the C++ entry.
Please check if this generates the correct link though.

```suggestion
 * @see precice::SolverInterface::setMeshAccessRegion()
 ```

Same below

File: src/precice/SolverInterface.hpp [hunk 42]
```diff
@@ -808,6 +808,27 @@ class SolverInterface {
    * on the receiving side, since the associated data values of the
    * filtered vertices are filled with zero data.
    *
+   * @note This function can only be called once per participant and
+   * rank and an assert is thrown, if you try to call it more than once.
```
fsimonis, line 812: API usage should __never__ result in an assert.

src/precice/impl/Participant.hpp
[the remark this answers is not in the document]
DavidSCN: It makes just as many sense as using `const` anywhere else on `lvalues`.

extras/bindings/c/include/precice/SolverInterfaceC.h
DavidSCN: What I have seen in other doxygen code is also the directive `@copydoc`. Have you ever considered the usage here?

src/precice/SolverInterface.hpp
DavidSCN: Fixed the wording.

extras/bindings/c/include/precice/SolverInterfaceC.h
fsimonis: I wasn't aware of that!
You could use it, as long as the documented parameters are identical.

src/precice/impl/Participant.hpp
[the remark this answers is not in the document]
fsimonis: As you say, it makes sense in the definition, not the declaration.

```cpp
// In the header.hpp
// I need a meshID and I promise not to change the object  
bool isDirectAccessAllowed(int meshID) const;


// In the source.cpp
// I need a meshID, I promise not to change the object and I will not change the ID in the implementation
bool isDirectAccessAllowed(const int meshID) const;
```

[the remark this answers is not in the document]
DavidSCN: I was not aware that this is a valid declaration for a `const` input argument :flushed:
Not sure if it is good practice to have these inconsistent in terms of the `const`'ness

src/mesh/Mesh.cpp
DavidSCN: Do you have an example where we recompute bounding boxes?

src/precice/impl/Participant.hpp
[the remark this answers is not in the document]
fsimonis: Documenting constraints where they are required is generally a good idea.
So you could argue that it is good practice :wink:

src/mesh/Mesh.cpp
fsimonis: This is necessary for adaptive meshes aka my thesis.

src/precice/impl/Participant.hpp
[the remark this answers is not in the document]
DavidSCN: Perfect, then I will close this :)

src/mesh/Mesh.cpp
DavidSCN: Then I would say it depends how you implement them and how they are supposed to work together with a direct access. I guess one of the following two options should work: delete the previous bounding boxes and 1) recompute them or 2) require the user to provide a new one. Do you agree?

fsimonis: So, resetting the mesh will also reset the bounding box. I would say that this sounds like expected behaviour.

This means that I will document it once I enable `resetMesh()`.

fsimonis (approved):

Issue #1034, opened 2021-06-23
Write initial data does not set initial data


KyleDavisSA:
**Describe your setup**

Operating system: _Ubuntu 18.04_
preCICE Version: develop - https://github.com/precice/precice/commit/f1cb45f92cd5148174461d6123c69b2f2c92d08b

**Describe the problem**
From Issue https://github.com/precice/precice/issues/1032 , the case also fails for the master branch if write initial data is not used. Printing the initial data in `concatenateCouplingData` to a log file in 

https://github.com/precice/precice/blob/bd0e158aa0cc795ba0c7b227cf1c086c73cbc896/src/acceleration/BaseQNAcceleration.cpp#L457-L473

we can see that the initial data is zero after write initial data has been written. This works correctly for v2.2.1. 

**Step To Reproduce**

You can manually see these values by including the following output in `concatenateCouplingData`
```
PRECICE_DEBUG("Values: {}", values(i));
PRECICE_DEBUG("Old Values: {}", oldValues(i)); 
```
If the initial write data has been set, this will be displayed in the `Old Values`.

**Expected behaviour**

The values in `Old Values` should have been the values written in `writeInitialData`, but they were zero instead.

Issue #1032, opened 2021-06-12
Incorrect data / IQN error in elastic-tube-1d in develop


MakisH:
**Describe your setup**

Operating system (e.g. Linux distribution and version): _Ubuntu 21.04_
preCICE Version: v2.2.1-211-gf1cb45f9
Tutorials 45eed (essentially v202104.1.1)

**Describe the problem**

Our elastic-tube-1d tutorial reports the following error when using preCICE from develop (not when using v2.2.1 as released):

```
(0) 14:45:12 [acceleration::BaseQNAcceleration]:202 in updateDifferenceMatrices: WARNING: The coupling residual equals almost zero. There is maybe something wrong in your adapter. Maybe you always write the same data or you call advance without providing new data first or you do not use available read data. Or you just converge much further than actually necessary.
(0) 14:45:12 [acceleration::BaseQNAcceleration]:231 in updateDifferenceMatrices: ERROR: Attempting to add a zero vector to the quasi-Newton V matrix. This means that the residual in two consecutive iterations is identical. There is probably something wrong in your adapter. Maybe you always write the same (or only incremented) data or you call advance without providing  new data first.
```

Full log: [solid-cpp.log](https://github.com/precice/precice/files/6642214/solid-cpp.log)

Most probably, this affects more cases.

**Step To Reproduce**
1. Run [the elastic-tube-1d tutorial](https://precice.org/tutorials-elastic-tube-1d.html) with `solid-cpp` and `fluid-cpp`.

**Expected behaviour**

Instead of exiting with an error, the `solid-cpp` solver should just start its time loop.

KyleDavisSA:
I reproduce this problem. preCICE receives nothing from the solvers (not even a zero value). This is a piece of the output from the log file:

```
(0) 09:10:34 [acceleration::BaseQNAcceleration]:469 in concatenateCouplingData: [0m  Input value: 
(0) 09:10:34 [acceleration::BaseQNAcceleration]:469 in concatenateCouplingData: [0m  Input value: 
(0) 09:10:34 [acceleration::BaseQNAcceleration]:472 in concatenateCouplingData: [0m  Input data norm: 
```

Printing each value from the computation of `crossSectionLength` and `pressure` result in the correct values, so the issue is where the data is written to preCICE:

```
interface.writeBlockScalarData(crossSectionLengthID, chunkLength, vertexIDs.data(), crossSectionLength.data());
```
[...]

PR #1036, opened 2021-06-23
Adds storeIteration for initial write data

KyleDavisSA:
## Main changes of this PR

stores the `previousIteration` after calling `writeInitialData`. 

This PR closes https://github.com/precice/precice/issues/1034 and https://github.com/precice/precice/issues/1032

## Author's checklist

* [x] I added a changelog file with this PR number in `docs/changelog/` if there are noteworthy changes. -> Fixes a bug, but the buggy version was never released, so no need to add anything to the changelog.

## Reviewers' checklist


MakisH:
This does fix #1032 for me. I do not understand the change, though.

KyleDavisSA:
For the quasi-Newton scheme, the `x^0` variable is always zero unless the initial write data is specified. In PR https://github.com/precice/precice/pull/1004, the initial write data is read in, but is not stored in the variable `cplData->previousIteration` for the QN method. That means in the `concatenateCouplingData` function in `BaseQNAcceleration`, `x^0` is always zero and never updated with the initial write data.

https://github.com/precice/precice/blob/bd0e158aa0cc795ba0c7b227cf1c086c73cbc896/src/acceleration/BaseQNAcceleration.cpp#L466

The 1D elastic tube test case is sensitive to this initial write data. The function `storeIteration()` updates the `previousIteration` values in:

https://github.com/precice/precice/blob/bd0e158aa0cc795ba0c7b227cf1c086c73cbc896/src/cplscheme/CouplingData.cpp#L38-L41

Adding this during  `exchangeInitialData` solved this problem.

BenjaminRodenberg:
Good find and thanks for the explanation! This is something I oversaw in #1004. 

Any idea why some of the tests are failing now? Do you think we should write a specific test to avoid this kind of bug in the future? I'll look into it more closely tomorrow.

KyleDavisSA:
I think a test is necessary. I will look into a test, that writes initial data and checks that value against `x^0` values.

BenjaminRodenberg:
> I think a test is necessary. I will look into a test, that writes initial data and checks that value against `x^0` values.

Then I will wait until you trigger me again or something pops up here. Right?

BenjaminRodenberg (commented):
I guess we also have to add `storeIteration` at the right places in `MultiCouplingScheme`:

https://github.com/precice/precice/blob/54707b3902af7d79b28ac276f304c31c6fcf5433/src/cplscheme/MultiCouplingScheme.cpp#L67-L82

KyleDavisSA:
Is there a potential bug in the SerialTests? Below is an output from the test:

```
/__w/precice/precice/src/precice/tests/SerialTests.cpp(401): error: in "PreciceTests/Serial/testExplicitWithDataInitialization": check 2.0 == valueDataB has failed [2 != 0]. Absolute value exceeds tolerance [|2| > 1e-09]
```

SolverB writes the initial data of 2

https://github.com/precice/precice/blob/f71e555ca03c6b0f11443088a3b651b6e747033d/src/precice/tests/SerialTests.cpp#L418-L421

and SolverA reads in the initial data

https://github.com/precice/precice/blob/f71e555ca03c6b0f11443088a3b651b6e747033d/src/precice/tests/SerialTests.cpp#L398-L401

However the test still says the the solverB data is zero. SolverA should not be reading the values before SolverB sets them?

uekerman:
> However the test still says the the solverB data is zero. SolverA should not be reading the values before SolverB sets them?

Not sure I understand what you mean, but the test looks correct. Not the test is buggy, preCICE is buggy.

fsimonis:
Shouldn't the receiving solver skip the call to initializeData?

Sender:
* is initializeData required?
* write Data(2) 
* initializeData
* mark as done

Receiver
* initialize (receives here)
* read data = 2

Actually, initializeData should throw an error if it is called without being required to.

BenjaminRodenberg:
I did another quick test comparing 3ff0899092862ad6f081df033ca9b51d0041befa (pre #1004) and 54707b3 (current state of this PR): I ran the 1d-elastic-tube cpp version https://github.com/precice/tutorials/commit/46ce8abee8dea365005da40f6e016a12f4bd97fd and compared the `precice-{Fluid,Solid}-iterations.log`. Here nothing suspicous shows up. I interpret this as a good sign and will dig deeper.

BenjaminRodenberg:
With this version the tests work and also the elastic-tube is ok. However, I would like to add a test that catches the underlying problem. I'm also not yet sure whether actually all additions to the code are necessary.

# Todos

- [ ] Develop a test that fails for 40eca9d and succeeds for 8197993
- [ ] Refactor #1036, if necessary

BenjaminRodenberg:
This PR is from my perspective ready. I would still suggest to merge #1050 first.

KyleDavisSA:
This PR closes #1034 and #1032, but does it also work in the expected manner for the newly constructed tests in PR https://github.com/precice/precice/pull/1050 ?

All tests pass now with this PR.

BenjaminRodenberg:
> This PR closes #1034 and #1032, but does it also work in the expected manner for the newly constructed tests in PR #1050 ?
> 
> All tests pass now with this PR.

When I tried this out a while ago: yes. You can locally merge this branch into `i_1032_test` (#1050) an run the tests. Then all tests (especially the previously failing ones) should succeed.

fsimonis:
Can we merge the tests into this PR, see if all is well and finally merge the fix?

PR #1037, opened 2021-06-25
Refactor mapReadData and mapWrittenData

BenjaminRodenberg:
## Main changes of this PR

Reduces code duplication.

## Author's checklist

* [x] I added a changelog file with this PR number in `docs/changelog/` if there are noteworthy changes. --> No noteworthy changes.

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -1473,7 +1473,7 @@ void SolverInterfaceImpl::computePartitions()
   }
 }
 
-void SolverInterfaceImpl::computeMappings(utils::ptr_vector<MappingContext> contexts, std::string mappingType)
+void SolverInterfaceImpl::computeMappings(utils::ptr_vector<MappingContext> contexts, const std::string &mappingType)
 {
   PRECICE_TRACE();
   using namespace mapping;
@@ -1492,7 +1492,7 @@ void SolverInterfaceImpl::computeMappings(utils::ptr_vector<MappingContext> cont
   }
 }
 
-void SolverInterfaceImpl::mapData(utils::ptr_vector<DataContext> contexts, std::string mappingType)
+void SolverInterfaceImpl::mapData(utils::ptr_vector<DataContext> contexts, const std::string &mappingType)
 {
   PRECICE_TRACE();
   using namespace mapping;
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -1473,46 +1473,52 @@ void SolverInterfaceImpl::computePartitions()
   }
 }
 
-void SolverInterfaceImpl::mapWrittenData()
+void SolverInterfaceImpl::computeMappings(utils::ptr_vector<MappingContext> contexts, std::string mappingType)
 {
-  PRECICE_TRACE();
   using namespace mapping;
   MappingConfiguration::Timing timing;
-  // Compute mappings
-  for (impl::MappingContext &context : _accessor->writeMappingContexts()) {
-    timing         = context.timing;
-    bool rightTime = timing == MappingConfiguration::ON_ADVANCE;
-    rightTime |= timing == MappingConfiguration::INITIAL;
+  for (impl::MappingContext &context : contexts) {
+    timing      = context.timing;
+    bool mapNow = timing == mapping::MappingConfiguration::ON_ADVANCE;
+    mapNow |= timing == mapping::MappingConfiguration::INITIAL;
     bool hasComputed = context.mapping->hasComputedMapping();
-    if (rightTime && not hasComputed) {
-      PRECICE_INFO("Compute write mapping from mesh \"{}\" to mesh \"{}\".",
-                   _accessor->meshContext(context.fromMeshID).mesh->getName(), _accessor->meshContext(context.toMeshID).mesh->getName());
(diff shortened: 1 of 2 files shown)


src/precice/impl/SolverInterfaceImpl.hpp
```diff
@@ -625,6 +625,15 @@ class SolverInterfaceImpl {
   /// Communicate meshes and create partitions
   void computePartitions();
 
+  /// Helper for mapWrittenData and mapReadData
+  void computeMappings(utils::ptr_vector<MappingContext> contexts, std::string mappingType);
+
+  /// Helper for mapWrittenData and mapReadData
+  void mapData(utils::ptr_vector<DataContext> contexts, std::string mappingType);
```
DavidSCN, line 632: I guess calling `mappingType` by const reference here should be sufficient, right?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1473,46 +1473,52 @@ void SolverInterfaceImpl::computePartitions()
   }
 }
 
-void SolverInterfaceImpl::mapWrittenData()
+void SolverInterfaceImpl::computeMappings(utils::ptr_vector<MappingContext> contexts, std::string mappingType)
 {
-  PRECICE_TRACE();
```
DavidSCN, line 1478: Why do you delete the trace here?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1473,46 +1473,52 @@ void SolverInterfaceImpl::computePartitions()
...
   MappingConfiguration::Timing timing;
-  // Compute mappings
-  for (impl::MappingContext &context : _accessor->writeMappingContexts()) {
-    timing         = context.timing;
-    bool rightTime = timing == MappingConfiguration::ON_ADVANCE;
-    rightTime |= timing == MappingConfiguration::INITIAL;
+  for (impl::MappingContext &context : contexts) {
+    timing      = context.timing;
+    bool mapNow = timing == mapping::MappingConfiguration::ON_ADVANCE;
+    mapNow |= timing == mapping::MappingConfiguration::INITIAL;
```
DavidSCN, line 1483: Since you already import the namespace functions above you could drop the namespace prefix here. If you still want to keep it explicit like this then I would propose to remove the `namespace` above.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1473,46 +1473,52 @@ void SolverInterfaceImpl::computePartitions()
...
+  for (impl::DataContext &context : contexts) {
     timing          = context.mappingContext.timing;
     bool hasMapping = context.mappingContext.mapping.get() != nullptr;
-    bool rightTime  = timing == MappingConfiguration::ON_ADVANCE;
-    rightTime |= timing == MappingConfiguration::INITIAL;
-    bool hasMapped = context.mappingContext.hasMappedData;
-    if (hasMapping && rightTime && (not hasMapped)) {
+    bool hasMapped  = context.mappingContext.hasMappedData;
+    bool mapNow     = timing == mapping::MappingConfiguration::ON_ADVANCE;
+    mapNow |= timing == mapping::MappingConfiguration::INITIAL;
```
DavidSCN, line 1503: same as above regarding the namespace

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRodenberg: Added `PRECICE_TRACE` to all the new functions in 9ef9db27efbadeccb2fdf3670e54d667f5846f6f. I did not intentionally delete the trace here. I rather forgot to add it in the new functions.

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRodenberg: Fixed in d275cf93ebe63626e45d7394e975e34690cbec46.

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRodenberg: Fixed in d275cf93ebe63626e45d7394e975e34690cbec46.

src/precice/impl/SolverInterfaceImpl.hpp
BenjaminRodenberg: Yes. I'm always getting very confused, when passing strings. So thanks for giving me this hint :D Fixed in e359bae998115c471113d20a701dfea0feaac7a6.

src/precice/impl/SolverInterfaceImpl.cpp
DavidSCN: I must admit I don't know where it should ideally go and where not. No we have it in both: the callee and the caller function. Feel free to merge if you know are familiar :)

DavidSCN (approved):

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRodenberg: @DavidSCN: If I get the tracer right it can go at the beginning of each function to tell the logger which function is entered. See here:

https://github.com/precice/precice/blob/bd0e158aa0cc795ba0c7b227cf1c086c73cbc896/src/logging/LogMacros.hpp#L42-L45

Issue #859, opened 2020-09-10
Remove deprecated API functions mapWriteDataFrom and mapReadDataTo


uekerman:
... and also the associated mapping timing.

There is currently no use case for these functions.

BenjaminRodenberg:
Thought the same when working on #1038. When removing these functions: We should also revise `DataContext`.

BenjaminRodenberg:
I added the `[[deprecated]]` attribute to these functions. See commit above.

BenjaminRodenberg:
#1222 is now merged into the breaking branch `precice:develop-v3.0.0`.

PR #1038, opened 2021-06-25
Refactor DataContext

BenjaminRodenberg:
## Main changes of this PR

Turns `DataContext` into a class. Add corresponding getters, setters and restrict access. Add member `participantData` and only set `toData` and `fromData`, when a mapping exists.

 This allows us to control access on `toData` and `fromData`. We will be able to add further functionality to the `DataContext` that might be required for implementation of waveform relaxation (#1029).

## Author's checklist

* [x] I added a changelog file with this PR number in `docs/changelog/` if there are noteworthy changes. --> Not relevant, since refactoring.

## Reviewers' checklist

* [ ] (more questions/tasks)


src/precice/impl/DataContext.hpp
```diff
@@ -11,18 +11,49 @@ namespace impl {
  * @brief Stores one Data object with related context. If this dataContext is not associated with a mapping,
  * fromData and toData refer to the same data object.
```
uekerman, line 12: I guess, we now should add more documentation here. Especially when a DataContext is created and what it stands for exactly.

src/precice/impl/DataContext.cpp
```diff
@@ -1,13 +1,103 @@
...
+  PRECICE_ASSERT(toData);
+  _mappingContext = mappingContext;
+  _fromData       = fromData;
+  _toData         = toData;
+}
+
+bool DataContext::hasMapping() const
+{
+  return _toData && _fromData && (_toData != _fromData);
+}
```
uekerman, line 96: Wouldn't it be nicer to have a member `_hasMapping`, which is set in `setMapping` and which we could assert in `getFromData` etc?

src/precice/impl/Participant.hpp
```diff
@@ -21,7 +21,7 @@
 
 namespace precice {
 namespace impl {
-struct DataContext;
+class DataContext;
```
uekerman, line 24: Do we still need / want this forward declaration?

uekerman (changes requested):
Good effort to bring more structure into `DataContext`. 
I am, however, not sure whether the solution to have a third field `participantData` is the best way to go. Before, there were two data fields, which isn't ideal already. Now, we solve this by adding a third field :grin:  
If we stick with this solution (I see the advantages as well), we need to add more documentation and find a more intuitive name than "participantData".

src/precice/impl/DataContext.hpp
```diff
@@ -11,18 +11,49 @@ namespace impl {
...
-  MappingContext mappingContext;
+  mesh::PtrData fromData();
+
+  std::string getFromDataName() const;
+
+  int getFromDataID() const;
+
+  mesh::PtrData toData();
+
+  std::string getToDataName() const;
```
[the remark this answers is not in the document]
BenjaminRodenberg: I started with a basic refactoring here. The reason why I added `toDataName` and `fromDataName` was that previously sometimes `toData->getName()` and sometimes `fromData->getName()` was used. I added an assertion in 3ed6d071fb0417410269db2fd6cc46ce00207b16 to make the fact that all names are the same explicit and cleanup up a bit. Now there is only a single function `getDataName()` which is used in all places.

src/precice/impl/DataContext.cpp
BenjaminRodenberg: Currently there are two places that need to publicly access `hasMapping()`:

```
void SolverInterfaceImpl::resetWrittenData()
{
  PRECICE_TRACE();
  for (DataContext &context : _accessor->writeDataContexts()) {
    context.participantData()->toZero();
    if (context.hasMapping()) {
      PRECICE_ASSERT(context.fromData() == context.participantData());
      context.toData()->toZero();
    }
  }
}
```
see [here](https://github.com/BenjaminRodenberg/precice/blob/3ed6d071fb0417410269db2fd6cc46ce00207b16/src/precice/impl/SolverInterfaceImpl.cpp#L1603).

We could use something similar to [this code](https://github.com/BenjaminRodenberg/precice/blob/3ed6d071fb0417410269db2fd6cc46ce00207b16/src/precice/impl/SolverInterfaceImpl.cpp#L1502): `bool hasMapping = context.mappingContext().mapping.get() != nullptr;` as a workaround. But to me this does not look very explicit.

In 1ebca03f20d26cdd4a808cde0ae5ae1dc8be316c I cleaned up the rest and added some assertions as proposed. I actually replaced the code snipped from above with a call of `hasMapping()`.

src/precice/impl/DataContext.hpp
```diff
@@ -11,18 +11,49 @@ namespace impl {
...
-struct DataContext {
-  bool used = false;
+class DataContext {
 
-  std::string getName() const;
+public:
+  DataContext(mesh::PtrData data, mesh::PtrMesh mesh);
 
-  mesh::PtrData fromData;
+  mesh::PtrData participantData();
```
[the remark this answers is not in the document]
BenjaminRodenberg: I tried to explain this [here](https://github.com/precice/precice/pull/1038/files#diff-ec343ae5f8584689e53ce2e3df4ab22be58d5e1506fdbc8ccb115e6999edfd00R46): `mesh::PtrData _participantData; // data this participant will write to and read from`.

This allows us to always ask for `participantData()` in read and write functions and not `toData()` for read and `fromData:` for write functions. This restricts the calls to `toData()` and `fromData()` to places where mapping happens:

Screenshot from 2021-06-29 18-43-14

Screenshot from 2021-06-29 18-43-28

At all other places where data is accessed, but the mapping is not really relevant, we just access `participantData()`:

Screenshot from 2021-06-29 18-42-59

src/precice/impl/Participant.hpp
BenjaminRodenberg: I replaced the forward declaration with a normal `#include "precice/impl/DataContext.hpp"` in 5dae15a914526ff92a84bed194d278cbf2f71af3.

src/precice/impl/DataContext.hpp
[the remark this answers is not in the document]
uekerman: Looks better now :+1:

src/precice/impl/DataContext.cpp
uekerman: Looks good.
You could directly set
```
bool _hasMapping = false;
```
in `DataContext.hpp` instead of setting it in the constructor.

src/precice/impl/DataContext.hpp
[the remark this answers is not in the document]
BenjaminRodenberg: Some ideas for the name from discussion with @uekerman:

* just `data`
* or `providedData`, which would be consistent with a provided mesh.

src/precice/impl/DataContext.hpp
BenjaminRodenberg: Important points for documentation:

* `toData` and `fromData` are always different, if a mapping exists.
* `toData` and `fromData` don't exist if no mapping exists.
* either `toData` or `fromData` is identical with `participantData` (this depends on whether the mapping is read or write. **todo:** This is something we might want to assert, as well)

src/precice/impl/DataContext.cpp
BenjaminRodenberg: Done in 7b2a73760dce1004d8eac10b52ca22c9db314ca6

BenjaminRodenberg:
Should be merged first:

- [ ] #1056 
- [ ] #1051

src/precice/impl/DataContext.hpp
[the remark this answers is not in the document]
BenjaminRodenberg: Renamed `participantData` to `providedData` in 34c7acd.

src/precice/impl/DataContext.hpp
BenjaminRodenberg: From my point of view documentation is in good shape now. I have the feeling that as soon as we remove `mapWriteDataFrom` and `mapReadDataTo` (#859) the `DataContext` will become quite simple. I don't think that it's really worth the effort documenting functions that we will remove very soon anyway.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1013,14 +1012,13 @@ void SolverInterfaceImpl::mapReadDataTo(
...
     for (impl::DataContext &context : _accessor->readDataContexts()) {
-      if (context.mesh->getID() != toMeshID) {
+      if (context.getMeshID() != toMeshID) {
         continue;
       }
-      context.toData->values() = Eigen::VectorXd::Zero(context.toData->values().size());
-      PRECICE_DEBUG("Map data \"{}\" to mesh \"{}\"", context.fromData->getName(), context.mesh->getName());
-      PRECICE_ASSERT(mappingContext.mapping == context.mappingContext.mapping);
-      mappingContext.mapping->map(context.fromData->getID(), context.toData->getID());
-      PRECICE_DEBUG("Mapped values = {}", utils::previewRange(3, context.toData->values()));
```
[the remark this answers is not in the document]
BenjaminRodenberg: Looks like an accident. Will add it again.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1549,20 +1547,20 @@ void SolverInterfaceImpl::mapData(utils::ptr_vector<DataContext> contexts, const
...
-    mapNow |= timing == MappingConfiguration::INITIAL;
-    if (mapNow && hasMapping && (not hasMapped)) {
-      int inDataID  = context.fromData->getID();
-      int outDataID = context.toData->getID();
-      PRECICE_DEBUG("Map \"{}\" data \"{}\" from mesh \"{}\"",
-                    mappingType, context.fromData->getName(), context.mesh->getName());
-      context.toData->values() = Eigen::VectorXd::Zero(context.toData->values().size());
-      PRECICE_DEBUG("Map from dataID {} to dataID: {}", inDataID, outDataID);
-      context.mappingContext.mapping->map(inDataID, outDataID);
-      PRECICE_DEBUG("Mapped values = {}", utils::previewRange(3, context.toData->values()));
```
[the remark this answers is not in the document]
BenjaminRodenberg: Also looks like an accident.

src/precice/impl/SolverInterfaceImpl.cpp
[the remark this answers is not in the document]
BenjaminRodenberg: Added the debug message again in dfa4d356ae31ae0d83403bdcc7abe18bfda82863. I now remember why I removed it: Because it allowed me to not provide any access to `DataContext::_toData`. If we want this debug message, we have to provide access and this makes the interface more complicated. I added a `@todo` to point to this.

I would rather not touch this point here, since it requires to refactor the mapping class hierarchy or introduce lots of code duplication, if we want to move this debug message into the mapping (where accessing `_toData` should not be any problem).

src/precice/impl/SolverInterfaceImpl.cpp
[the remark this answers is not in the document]
BenjaminRodenberg: See above. Added it again in dfa4d356ae31ae0d83403bdcc7abe18bfda82863.

BenjaminRodenberg:
I added two tests for the mapping context. There are two minor todos that I will take a look at later, but the test is generally ready for review.

uekerman:
Tests look good

src/precice/impl/DataContext.hpp
```diff
@@ -2,27 +2,82 @@
...
+ *   One of the two must be equal to providedData. fromData and toData must be different.
+ * - If this dataContext is not associated with a mapping, fromData and toData will be unset.
  */
-struct DataContext {
-  bool used = false;
+class DataContext {
 
-  std::string getName() const;
+public:
+  DataContext(mesh::PtrData data, mesh::PtrMesh mesh); // @ todo use mesh context for constructor? Would be more consistent with configureForReadMapping and configureForWriteMapping
```
BenjaminRodenberg, line 21: I will check whether this is easily doable or otherwise remove the todo.

uekerman (approved):

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1013,14 +1012,14 @@ void SolverInterfaceImpl::mapReadDataTo(
...
-      context.toData->values() = Eigen::VectorXd::Zero(context.toData->values().size());
-      PRECICE_DEBUG("Map data \"{}\" to mesh \"{}\"", context.fromData->getName(), context.mesh->getName());
-      PRECICE_ASSERT(mappingContext.mapping == context.mappingContext.mapping);
-      mappingContext.mapping->map(context.fromData->getID(), context.toData->getID());
-      PRECICE_DEBUG("Mapped values = {}", utils::previewRange(3, context.toData->values()));
+      context.resetToData();
+      PRECICE_DEBUG("Map data \"{}\" to mesh \"{}\"", context.getDataName(), context.getMeshName());
+      PRECICE_ASSERT(mappingContext.mapping == context.mappingContext().mapping);
+      mappingContext.mapping->map(context.getFromDataID(), context.getToDataID());
+      PRECICE_DEBUG("Mapped values = {}", utils::previewRange(3, context.toData()->values())); // @todo might be better to move this debug message into Mapping::map and remove getter DataContext::toData()
```
BenjaminRodenberg, line 1022: I will open an issue and point to this todo. @uekerman and I agree that this debug message should go into the mapping.

src/precice/impl/DataContext.hpp
BenjaminRodenberg: I tried to use meshContext, but it is not easy to do, because the meshContext is not available in the participant configuration. Calling `Participant::meshContext(meshID)` gave me a failing assertion in some tests. Therefore, I will remove this todo and leave the rest as it is.

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRodenberg: Created the issue https://github.com/precice/precice/issues/1073.

PR #1042, opened 2021-07-09
Add a warning for Aitken relaxation and parallel coupling schemes

DavidSCN:
.. in order to inform user about potential instabilities.

### DIFF
--- a/src/cplscheme/config/CouplingSchemeConfiguration.cpp
+++ b/src/cplscheme/config/CouplingSchemeConfiguration.cpp
@@ -1114,9 +1114,9 @@ void CouplingSchemeConfiguration::setParallelAcceleration(
 
     if (dynamic_cast<acceleration::AitkenAcceleration *>(_accelerationConfig->getAcceleration().get()) != nullptr)
       PRECICE_WARN("You configured participant \"{}\" in a parallel-implicit coupling scheme with \"Aitken\" "
-                   "acceleration, which is known to perform bad in parallel coupling schemes. Consider switching to a serial-implicit coupling "
-                   "scheme or changing the acceleration method.",
-                   participant);
+                   "acceleration, which is known to perform bad in parallel coupling schemes. "
+                   "See https://precice.org/configuration-acceleration.html#dynamic-aitken-under-relaxation for details."
+                   "Consider switching to a serial-implicit coupling scheme or changing the acceleration method.", participant);
   }
 }
 
--- a/src/cplscheme/config/CouplingSchemeConfiguration.cpp
+++ b/src/cplscheme/config/CouplingSchemeConfiguration.cpp
@@ -1114,8 +1114,8 @@ void CouplingSchemeConfiguration::setParallelAcceleration(
 
     if (dynamic_cast<acceleration::AitkenAcceleration *>(_accelerationConfig->getAcceleration().get()) != nullptr)
       PRECICE_WARN("You configured participant \"{}\" in a parallel-implicit coupling scheme with \"Aitken\" "
-                   "acceleration, which is known to perform bad. Consider switching to a serial-implicit coupling "
-                   "scheme or change the acceleration method.",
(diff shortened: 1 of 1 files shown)


src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -1110,6 +1111,12 @@ void CouplingSchemeConfiguration::setParallelAcceleration(
       checkIfDataIsExchanged(dataID);
     }
     scheme->setAcceleration(_accelerationConfig->getAcceleration());
+
+    if (dynamic_cast<acceleration::AitkenAcceleration *>(_accelerationConfig->getAcceleration().get()) != nullptr)
+      PRECICE_WARN("You configured participant \"{}\" in a parallel-implicit coupling scheme with \"Aitken\" "
+                   "acceleration, which is known to perform bad. Consider switching to a serial-implicit coupling "
```
BenjaminRodenberg, line 1117:
```suggestion
                   "acceleration, which is known to perform bad in this setup. Consider switching to a serial-implicit coupling "
```
or `in this context`? Otherwise it sounds like Aitken is always bad.

DavidSCN: Then let me try to be even more explicit:
```suggestion
                   "acceleration, which is known to perform bad in parallel coupling schemes. Consider switching to a serial-implicit coupling "
```

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -1110,6 +1111,12 @@ void CouplingSchemeConfiguration::setParallelAcceleration(
...
+                   "scheme or change the acceleration method.",
```
DavidSCN, line 1118:
```suggestion
                   "scheme or changing the acceleration method.",
```

BenjaminRodenberg (changes requested):
This PR looks mostly fine. There are two open todos that I will take care of:

- [ ] add more detailed explanation to documentation (see https://github.com/precice/precice.github.io/pull/71)
- [ ] link to documentation in warning.

Let's merge this PR as soon as the todos above are addressed.

DavidSCN:
Since I am not aware of the reasons I cannot really provide more documentation/information. Do you want to refer to a specific paper or do you want to open up a new section on the website?

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -1110,6 +1111,12 @@ void CouplingSchemeConfiguration::setParallelAcceleration(
       checkIfDataIsExchanged(dataID);
     }
     scheme->setAcceleration(_accelerationConfig->getAcceleration());
+
+    if (dynamic_cast<acceleration::AitkenAcceleration *>(_accelerationConfig->getAcceleration().get()) != nullptr)
+      PRECICE_WARN("You configured participant \"{}\" in a parallel-implicit coupling scheme with \"Aitken\" "
+                   "acceleration, which is known to perform bad in parallel coupling schemes. Consider switching to a serial-implicit coupling "
+                   "scheme or changing the acceleration method.",
+                   participant);
```
BenjaminRodenberg, line 1119:
```suggestion
      PRECICE_WARN("You configured participant \"{}\" in a parallel-implicit coupling scheme with \"Aitken\" "
                   "acceleration, which is known to perform bad in parallel coupling schemes. "
                   "See https://precice.org/configuration-acceleration.html#dynamic-aitken-under-relaxation for details."
                   "Consider switching to a serial-implicit coupling scheme or changing the acceleration method.", participant);
```

I'm currently working on the documentation in https://github.com/precice/precice.github.io/pull/71. We could link to this place in the warning.

BenjaminRodenberg (approved):
We now have everything we need. As soon as the tests are through, I will do the merge.

PR #1044, opened 2021-07-21
Remove FASTEST fortran language bindings

DavidSCN:
## Main changes of this PR
Removes the FASTEST fortran solverinterface.

The FASTEST interface was historically introduced for the ExaFSA project only (to handle fluid and acoustic coupling with different time windows, so different `SolverInterfaces`). It is currently unused. In the FASTEST version of FNB, our normal Fortran bindings are used. There is no obvious reason to maintain and develop it further. As we want to extend the API now in #1033, it makes sense to get rid of this additional burden now. 

## Author's checklist

## Reviewers' checklist


fsimonis:
What is the motivation to remove them now?
_Just for the record. _

DavidSCN:
I added a motivation. @uekerman feel free to edit it if you have other aspects not mentioned.

uekerman (approved):

PR #1045, opened 2021-07-21
Introduce named types for preCICE own data structures

DavidSCN:
## Main changes of this PR
I would like to push the development of https://github.com/precice/precice/issues/248 a bit forward. By now, I just added the most obvious IDs and how I would declare them. Any feedback is very welcome. In the long run, we can potentially just switch from 32 to 64 bit indices. This PR does not touch anything of the code base yet.

See https://github.com/precice/precice/issues/248

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/utils/types.hpp
+++ b/src/utils/types.hpp
@@ -1,0 +1,38 @@
+#pragma once
+/**
+ * A namespace in which all relevant types used throughout the code base
+ * are decalred.
+ */
+namespace types {
+
+/**
+ * Type used for the IDs of vertices
+ */
+using vertexID = unsigned int;
+
+/**
+ * Type used for the IDs of edges
+ */
+using edgeID = unsigned int;
+
+/**
+ * Type used for the IDs of triangles
+ */
+using triangleID = unsigned int;
+
+/**
+ * Type used for the IDs of matching entities
+ */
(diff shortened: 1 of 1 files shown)


src/utils/types.hpp
```diff
@@ -0,0 +1,38 @@
...
+
+/**
+ * Type used for the IDs of triangles
+ */
+using triangleID = unsigned int;
+
+/**
+ * Type used for the IDs of matching entities
+ */
+using matchID = unsigned int;
```
uekerman, line 26: I don't understand this one. Could you sketch an example?

fsimonis:
There are a few issues here.

The namespace is not complete. It should at least start with precice. 


The location is obscure. We need a plan here.
We have types at the interface and internal times. The latter consists of package internal types and types used across packages.
Historically, we have `SharedPointer.hpp` files in some packages. We could rename these to `types.hpp` and declare local types there.

The interface types are more tricky though. I expect simple typedefs in the solverInterface to do the trick.
The question is how we should expose these in the C interface as the implementation will "see" both.

DavidSCN:
Closing this as the changes are already in #1048

Issue #887, opened 2020-10-29
Build with 64 bit type for id's, indices, ... not possible


TEFEdotCC:
Problem from preCICE/openfoam-adapter.

If openfoam is build with `WM_LABEL_SIZE=64`, the `FOAM::label` type is an 64 bit integer. That makes problems (possible overflow) to create an consistent adapter for openfoam.
Why? In that configuration openfoam can handle 64bit indizies and precice only 32bit.
Is it necessary? ... if the case does not need 64 storage --> no, but the switch should be possible without any conflicts. Future is comming.

Example on https://github.com/precice/openfoam-adapter/blob/master/Interface.C#L201-L212
The line in the for-loop `faceField.size()` is on gcc an `long int` 64 bit type, but the precice function `setMeshTriangleWithEdges` can only handle `int` 32 bit. So, it is not possible to set `int triVertIDs[faceField.size()*(triaPerQuad*nodesPerTria)];` to `Foam::label triVertIDs[faceField.size()*(triaPerQuad*nodesPerTria)];`
Also the return type of `faceField.size()` in the example is 64 bit integer.

fsimonis:
Hi, we had a similar discussion regarding the precision used for the data values #726.

Implementing a configurable index size will be a quite a massive task but should also be significantly simpler as it is not tied into the numerics.
The absence of 64bit indices could also be a problem for implementing volume coupling with particle codes #468. So there are definitely use-cases here.

Maybe this could be combined with #248.

fsimonis:
Postponing this and #405 due to high dev cost, more communication at no direct gain.
We can implement this in a backwards-compatible way by duplicating id-specific functions.
We run into other practical issues before we can take advantage of more than 2 147 million vertices.

PR #1048, opened 2021-07-27
Introduce named types

fsimonis:
## Main changes of this PR

Introduces a header providing named types for common internals.
The current use-case is to make types and locations more readable.

Related to #248 #1045

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/acceleration/impl/ParallelMatrixOperations.hpp
+++ b/src/acceleration/impl/ParallelMatrixOperations.hpp
@@ -7,12 +7,14 @@
 #include <stddef.h>
 #include <string>
 #include <vector>
+
 #include "com/Communication.hpp"
 #include "com/MPIPortsCommunication.hpp"
 #include "com/Request.hpp"
 #include "com/SharedPointer.hpp"
 #include "logging/LogMacros.hpp"
 #include "logging/Logger.hpp"
+#include "precice/types.hpp"
 #include "utils/MasterSlave.hpp"
 #include "utils/assertion.hpp"
 
@@ -226,7 +228,7 @@ private:
   {
     PRECICE_TRACE();
     for (int i = 0; i < leftMatrix.rows(); i++) {
-      int rank = 0;
+      Rank rank = 0;
       // find rank of processor that stores the result
       // the second while is necessary if processors with no vertices are present
       // Note: the >'=' here is crucial: In case some procs do not have any vertices,
@@ -291,7 +293,7 @@ private:
       // distribute blocks of summarizedBlocks (result of multiplication) to corresponding slaves
       result = summarizedBlocks.block(0, 0, offsets[1], r);
 
-      for (int rankSlave : utils::MasterSlave::allSlaves()) {
+      for (Rank rankSlave : utils::MasterSlave::allSlaves()) {
         int off       = offsets[rankSlave];
         int send_rows = offsets[rankSlave + 1] - offsets[rankSlave];
 
--- a/src/com/Communication.cpp
+++ b/src/com/Communication.cpp
@@ -1,9 +1,11 @@
-#include "Communication.hpp"
 #include <algorithm>
 #include <memory>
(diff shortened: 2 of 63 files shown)


DavidSCN:
Very nice, thanks! I will later go through the files more in detail. The most obvious question (and difference towards my suggestions): why not using `unsigned int` here or asking in a different way: what is the purpose of having -2e9 IDs available which we can never use? For meshIDs etc. this doesn't really matter, but for vertices, it should. The change would even be API compatible since we extend the current ID range.

In a follow up, I would like to introduce then some common constants referring to this types, e.g. `static const DataID invalidDataID = -1` or so in order to simplify initialization.

fsimonis:
> The most obvious question (and difference towards my suggestions): why not using unsigned int here or asking in a different way: what is the purpose of having -2e9 IDs available which we can never use?

This PR serves readability/maintainability only and doesn't change types.

For the interesting discussion regarding the types itself, please continue in #887.

src/precice/types.hpp
```diff
@@ -0,0 +1,37 @@
...
+ * Type used for the IDs of data
+ */
+using DataID = int;
+
+/**
+ * Type used for the IDs of meshes
+ */
+using MeshID = int;
+
+using Rank = int;
```
DavidSCN, line 35: Some documentation (though it seems obvious) might be helpful

src/precice/types.hpp
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Type used for the IDs of triangles
+ */
+using TriangleID = int;
+
+/**
+ * Type used for the IDs of matching entities
+ */
+using MatchID = int;
```
DavidSCN, line 23: In my PR, @uekerman requested some documentation clarification on this.

DavidSCN (commented):
Overall very nice. I'm not sure if we can already spot every occurrence of the readable datatypes. For some of the data types, I would expect that I can change the alias and the code still compiles (e.g. for VertexID). However, this is not the case at the moment and I'm not sure if that's the purpose of this PR (feel free to try it out). Maybe, we  can also clean this up in a follow-up PR.

src/com/Communication.cpp
```diff
@@ -53,7 +55,7 @@ void Communication::reduceSum(double const *itemsToSend, double *itemsToReceive,
   }
 }
 
-void Communication::reduceSum(double const *itemsToSend, double *itemsToReceive, int size, int rankMaster)
+void Communication::reduceSum(double const *itemsToSend, double *itemsToReceive, int size, Rank rankMaster)
```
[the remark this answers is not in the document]
fsimonis: Good point! The solution here would be to use a span. I had my eyes on a drop-in implementation for quite some time now.

That's for another PR though.

src/acceleration/impl/ParallelMatrixOperations.hpp
```diff
@@ -291,7 +293,7 @@ class ParallelMatrixOperations {
       // distribute blocks of summarizedBlocks (result of multiplication) to corresponding slaves
       result = summarizedBlocks.block(0, 0, offsets[1], r);
 
-      for (int rankSlave : utils::MasterSlave::allSlaves()) {
+      for (Rank rankSlave : utils::MasterSlave::allSlaves()) {
         int off       = offsets[rankSlave];
         int send_rows = offsets[rankSlave + 1] - offsets[rankSlave];
```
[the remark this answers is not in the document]
fsimonis: Offsets are rank related, but they aren't ranks. So this is where things get a bit tricky.
We could add RankOffset as an alias though.

src/precice/types.hpp
fsimonis: Used in nearest projection and could really go into the query package.

A nearest projection can match vertices, edge and triangles, so the type of the match is unclear.

PR #1050, opened 2021-07-29
Extend test for initializeData for coupling schemes

BenjaminRodenberg:
## Main changes of this PR

This PR extends the tests:

* `mpirun -np 4 ./testprecice -t CplSchemeTests/SerialImplicitCouplingSchemeTests/testInitializeData`
* `mpirun -np 4 ./testprecice -t CplSchemeTests/ParallelImplicitCouplingSchemeTests/testInitializeData`

The tests now verify that `previousIteration` is correctly set when `CouplingScheme::initializeData` is called.

This check is currently missing, which led to the bug described in #1032. At the current state the test is expected to fail. #1036 provides a fix for #1032 and this test is expected to succeed as soon as #1036 is merged.

## Author's checklist

* [x] I added a changelog file with this PR number in `docs/changelog/` if there are noteworthy changes. -> Not necessary

## Reviewers' checklist

* [ ] (more questions/tasks)


BenjaminRodenberg:
I think we should first merge this PR and then #1036. The we have a failing test on develop, which clearly documents the bug explained in #1032 and can then provide the corresponding fix #1036.

Note to the reviewers: There are a few todos in the tests, but I would rather solve these issues at a later point in time. The goal of this test is to avoid bugs such as #1032. The todos are mainly about the code design and how the interface of the coupling scheme is used.

src/cplscheme/ParallelCouplingScheme.hpp
```diff
@@ -8,6 +8,13 @@
 #include "m2n/SharedPointer.hpp"
 #include "utils/assertion.hpp"
 
+// Forward declaration to friend the boost test struct
+namespace CplSchemeTests {
+namespace ParallelImplicitCouplingSchemeTests {
+struct testInitializeData;
+}
+} // namespace CplSchemeTests
```
fsimonis, line 16: I think we should go for a reusable solution as  in https://github.com/precice/precice/pull/921#issuecomment-870590088

BenjaminRodenberg: Done in 906dd95cac721658597d15b0862ad04ae4488f80.

BenjaminRodenberg:
@fsimonis ready for review.

src/cplscheme/SerialCouplingScheme.hpp
```diff
@@ -84,5 +88,27 @@ class SerialCouplingScheme : public BiCouplingScheme {
...
+  {
+    return cplscheme.getReceiveData(dataID);
+  }
+
+  CouplingData *getSendData(SerialCouplingScheme &cplscheme, int dataID)
+  {
+    return cplscheme.getSendData(dataID);
+  }
+};
+
```
fsimonis, line 112: Could you move this into a separate header in `src/testing/` and into the `precice::testing` namespace?
Also we use capitalized names for types `struct SerialCouplingSchemeFixture`.

fsimonis:
I guess, reviewing this really only makes sense when the original issue is fixed, right?

For future reference: It would be better to have tests alongside the fix in the same PR.

BenjaminRodenberg:
> I guess, reviewing this really only makes sense when the original issue is fixed, right?
> 
> For future reference: It would be better to have tests alongside the fix in the same PR.

My main motivation for this PR is to create a test that helps us to see that there is actually an issue #1032. The issue fixed by this PR is that there is no test showing this bug. #1032 is fixed in #1036. We can merge all in one single PR, but I think it's a cleaner and more explicit approach to

1) reproduce #1032 via a test in #1050 and
2) fix #1032 in #1036.

Having a failing test on develop is not a problem, especially, if the fix is ready. Not having any test is actually a problem.

src/cplscheme/SerialCouplingScheme.hpp
BenjaminRodenberg: > Also we use capitalized names for types `struct SerialCouplingSchemeFixture`.

Then our format checker should also check this. I think the same issue is present in #921. Just added a quick review over there.

I will quickly take care of the rest.

BenjaminRodenberg: Done in d6342d3 and deae723

KyleDavisSA:
WIth this PR, the following tests fail as expected:

```
The following tests FAILED:
	  4 - precice.cplscheme (Failed)
	 20 - precice.solverdummy.build.cpp (Failed)
	 21 - precice.solverdummy.build.c (Failed)
	 22 - precice.solverdummy.build.fortran (Failed)
	 23 - precice.solverdummy.run.cpp-cpp (Not Run)
	 24 - precice.solverdummy.run.c-c (Not Run)
	 25 - precice.solverdummy.run.fortran-fortran (Not Run)
	 26 - precice.solverdummy.run.cpp-c (Not Run)
	 27 - precice.solverdummy.run.cpp-fortran (Not Run)
	 28 - precice.solverdummy.run.c-fortran (Not Run)

```

PR #1051, opened 2021-07-29
Code simplifications w.r.t mesh->data().at(ID)

BenjaminRodenberg:
## Main changes of this PR

Code simplifications.

Reduces code complexity.

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)


BenjaminRodenberg:
There seems to be a problem with some of the tests that I will have to investigate. @fsimonis: A thorough review does not make sense at this point in time, but do you agree with the overall direction this PR is heading into?

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -755,11 +755,7 @@ mesh::PtrData CouplingSchemeConfiguration::findDataByID(
     int ID) const
 {
   for (const mesh::PtrMesh &mesh : _meshConfig->meshes()) {
-    for (mesh::PtrData data : mesh->data()) {
-      if (data->getID() == ID) {
-        return data;
-      }
-    }
+    return mesh->data(ID);
```
fsimonis, line 758: This requires ID to exist.

fsimonis (changes requested):

src/cplscheme/config/CouplingSchemeConfiguration.cpp
BenjaminRodenberg: Good point. Will do this.

BenjaminRodenberg: Done in 5ec6d2a761aeb1a5f6c81a65fde85f072eabe237

fsimonis (approved):

PR #1052, opened 2021-07-30
Add bindings to CI workflow

DavidSCN:
## Main changes of this PR
Triggers our build and test CI pipeline in case there are changes in the native bindings (C/fortran).

Might be very rarely, but in case there are changes in the bindings only, the CI is not triggered. Related to #1044 

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -11,6 +11,7 @@ on:
       - 'src/**'
       - 'thirdparty/**'
       - 'CMakeLists.txt'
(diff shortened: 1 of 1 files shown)


BenjaminRodenberg:
This only includes the "native bindings", correct? So not: python, MATLAB. I think this goes into the right direction, since python and MATLAB bindings have their own tests (or should have their own tests). Could you make this more clear in the header of this PR?

DavidSCN:
Done @BenjaminRodenberg

PR #1053, opened 2021-07-30
Port communication backend to using span

fsimonis:
## Main changes of this PR

* Added the glorious span implementation from https://github.com/tcbrindle/span  
  This is a drop-in replacement for the C++20 `std::span`
* Ported the entire communication back-end from using `T* data, size_t size` to using `precice::span<T>`.  
  Using the namespace `precice` allows to simply rename it to `std::` once we upgrade to C++20.
* Ported all tests to the new span method.
* Cleaned up the interface considerably.

Using `T*` + `size` is error prone and verbose. Even worse is using a single size for multiple `T*`, something we frequently do in the communication back-end.
A span is a neat non-owning wrapper around this and provides convenience methods such as `begin(), end(), empty()`.
`std::span` is part of the C++20 standard, so there is no reason not to take advantage of its features right now.

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/acceleration/test/ParallelMatrixOperationsTest.cpp
+++ b/src/acceleration/test/ParallelMatrixOperationsTest.cpp
@@ -106,12 +106,12 @@ BOOST_AUTO_TEST_CASE(ParVectorOperations)
   int    iaa   = (int) a;
   int    ires1 = 0, ires2 = 0;
 
-  utils::MasterSlave::allreduceSum(a, res1, 1);
-  utils::MasterSlave::allreduceSum(iaa, ires2, 1);
-  utils::MasterSlave::allreduceSum(aa.data(), res2.data(), 2);
+  utils::MasterSlave::allreduceSum(a, res1);
+  utils::MasterSlave::allreduceSum(iaa, ires2);
+  utils::MasterSlave::allreduceSum(aa, res2);
 
-  utils::MasterSlave::reduceSum(aa.data(), res3.data(), 2);
-  utils::MasterSlave::reduceSum(iaa, ires1, 1);
+  utils::MasterSlave::reduceSum(aa, res3);
+  utils::MasterSlave::reduceSum(iaa, ires1);
 
   BOOST_TEST(testing::equals(res1, 10.));
   BOOST_TEST(testing::equals(ires2, 10));
--- a/src/com/Communication.cpp
+++ b/src/com/Communication.cpp
@@ -47,7 +47,7 @@ void Communication::reduceSum(precice::span<double const> itemsToSend, precice::
   std::vector<double> received(itemsToReceive.size());
   // receive local results from slaves
   for (size_t rank = 0; rank < getRemoteCommunicatorSize(); ++rank) {
-    auto request = aReceive({received}, rank + _rankOffset);
+    auto request = aReceive(received, rank + _rankOffset);
     request->wait();
     for (size_t i = 0; i < itemsToReceive.size(); i++) {
       itemsToReceive[i] += received[i];
@@ -96,7 +96,7 @@ void Communication::allreduceSum(precice::span<double const> itemsToSend, precic
 
   // send reduced result to all slaves
   std::vector<PtrRequest> requests;
-  requests.resize(getRemoteCommunicatorSize());
+  requests.reserve(getRemoteCommunicatorSize());
   for (size_t rank = 0; rank < getRemoteCommunicatorSize(); ++rank) {
     requests.push_back(aSend(itemsToReceive, rank + _rankOffset));
   }
--- a/src/com/Communication.cpp
+++ b/src/com/Communication.cpp
@@ -6,6 +6,7 @@
 #include "Request.hpp"
 #include "logging/LogMacros.hpp"
 #include "precice/types.hpp"
+#include "utils/assertion.hpp"
 
 namespace precice {
 namespace com {
@@ -41,6 +42,7 @@ void Communication::connectMasterSlaves(std::string const &participantName,
 void Communication::reduceSum(precice::span<double const> itemsToSend, precice::span<double> itemsToReceive)
 {
   PRECICE_TRACE(itemsToSend.size(), itemsToReceive.size());
+  PRECICE_ASSERT(itemsToSend.size() == itemsToReceive.size());
 
   std::copy(itemsToSend.begin(), itemsToSend.end(), itemsToReceive.begin());
 
@@ -58,6 +60,7 @@ void Communication::reduceSum(precice::span<double const> itemsToSend, precice::
 void Communication::reduceSum(precice::span<double const> itemsToSend, precice::span<double> itemsToReceive, Rank rankMaster)
 {
   PRECICE_TRACE(itemsToSend.size(), itemsToReceive.size());
+  PRECICE_ASSERT(itemsToSend.size() == itemsToReceive.size());
 
   auto request = aSend(itemsToSend, rankMaster);
   request->wait();
@@ -91,6 +94,7 @@ void Communication::reduceSum(int itemToSend, int &itemToReceive, Rank rankMaste
 void Communication::allreduceSum(precice::span<double const> itemsToSend, precice::span<double> itemsToReceive)
 {
   PRECICE_TRACE(itemsToSend.size(), itemsToReceive.size());
+  PRECICE_ASSERT(itemsToSend.size() == itemsToReceive.size());
 
   reduceSum(itemsToSend, itemsToReceive);
 
@@ -109,6 +113,7 @@ void Communication::allreduceSum(precice::span<double const> itemsToSend, precic
 void Communication::allreduceSum(precice::span<double const> itemsToSend, precice::span<double> itemsToReceive, Rank rankMaster)
 {
   PRECICE_TRACE(itemsToSend.size(), itemsToReceive.size());
+  PRECICE_ASSERT(itemsToSend.size() == itemsToReceive.size());
 
   reduceSum(itemsToSend, itemsToReceive, rankMaster);
   // receive reduced data from master
--- a/src/com/Communication.cpp
+++ b/src/com/Communication.cpp
@@ -38,28 +38,28 @@ void Communication::connectMasterSlaves(std::string const &participantName,
 /**
  * @attention This method modifies the input buffer.
  */
-void Communication::reduceSum(double const *itemsToSend, double *itemsToReceive, int size)
+void Communication::reduceSum(precice::span<double const> itemsToSend, precice::span<double> itemsToReceive)
 {
-  PRECICE_TRACE(size);
+  PRECICE_TRACE(itemsToSend.size(), itemsToReceive.size());
 
-  std::copy(itemsToSend, itemsToSend + size, itemsToReceive);
+  std::copy(itemsToSend.begin(), itemsToSend.end(), itemsToReceive.begin());
 
-  std::vector<double> received(size);
+  std::vector<double> received(itemsToReceive.size());
   // receive local results from slaves
   for (size_t rank = 0; rank < getRemoteCommunicatorSize(); ++rank) {
-    auto request = aReceive(received.data(), size, rank + _rankOffset);
+    auto request = aReceive({received}, rank + _rankOffset);
     request->wait();
-    for (int i = 0; i < size; i++) {
+    for (size_t i = 0; i < itemsToReceive.size(); i++) {
       itemsToReceive[i] += received[i];
     }
   }
 }
 
-void Communication::reduceSum(double const *itemsToSend, double *itemsToReceive, int size, Rank rankMaster)
+void Communication::reduceSum(precice::span<double const> itemsToSend, precice::span<double> itemsToReceive, Rank rankMaster)
 {
-  PRECICE_TRACE(size);
+  PRECICE_TRACE(itemsToSend.size(), itemsToReceive.size());
 
-  auto request = aSend(itemsToSend, size, rankMaster);
+  auto request = aSend(itemsToSend, rankMaster);
   request->wait();
 }
 
@@ -81,34 +81,24 @@ void Communication::reduceSum(int itemToSend, int &itemToReceive, Rank rankMaste
 {
   PRECICE_TRACE();
 
-  auto request = aSend(&itemToSend, 1, rankMaster);
+  auto request = aSend(itemToSend, rankMaster);
   request->wait();
 }
 
 /**
  * @attention This method modifies the input buffer.
  */
-void Communication::allreduceSum(double const *itemsToSend, double *itemsToReceive, int size)
+void Communication::allreduceSum(precice::span<double const> itemsToSend, precice::span<double> itemsToReceive)
 {
-  PRECICE_TRACE(size);
+  PRECICE_TRACE(itemsToSend.size(), itemsToReceive.size());
 
-  std::copy(itemsToSend, itemsToSend + size, itemsToReceive);
-
-  std::vector<double> received(size);
-  // receive local results from slaves
-  for (size_t rank = 0; rank < getRemoteCommunicatorSize(); ++rank) {
-    auto request = aReceive(received.data(), size, rank + _rankOffset);
-    request->wait();
-    for (int i = 0; i < size; i++) {
-      itemsToReceive[i] += received[i];
-    }
-  }
+  reduceSum(itemsToSend, itemsToReceive);
 
   // send reduced result to all slaves
-  std::vector<PtrRequest> requests(getRemoteCommunicatorSize());
+  std::vector<PtrRequest> requests;
+  requests.resize(getRemoteCommunicatorSize());
   for (size_t rank = 0; rank < getRemoteCommunicatorSize(); ++rank) {
-    auto request   = aSend(itemsToReceive, size, rank + _rankOffset);
-    requests[rank] = request;
+    requests.push_back(aSend(itemsToReceive, rank + _rankOffset));
   }
   Request::wait(requests);
 }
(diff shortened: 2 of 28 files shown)


src/com/MPICommunication.cpp
```diff
@@ -83,26 +83,26 @@ PtrRequest MPICommunication::aSend(const int *itemsToSend, int size, Rank rankRe
   return PtrRequest(new MPIRequest(request));
 }
 
-void MPICommunication::send(const double *itemsToSend, int size, Rank rankReceiver)
+void MPICommunication::send(precice::span<const double> itemsToSend, Rank rankReceiver)
```
BenjaminRodenberg, line 86: Do we need this code duplication of `void MPICommunication::send(precice::span<const int> itemsToSend, Rank rankReceiver)`?

src/com/MPIDirectCommunication.cpp
```diff
@@ -76,17 +77,19 @@ void MPIDirectCommunication::requestConnection(std::string const &acceptorName,
...
-  PRECICE_TRACE(size);
+  PRECICE_TRACE(itemsToSend.size());
+  PRECICE_ASSERT(itemsToSend.size() == itemsToReceive.size());
   Rank rank = _commState->rank();
-  MPI_Reduce(const_cast<double *>(itemsToSend), itemsToReceive, size, MPI_DOUBLE, MPI_SUM, rank, _commState->comm);
+  MPI_Reduce(const_cast<double *>(itemsToSend.data()), itemsToReceive.data(), itemsToSend.size(), MPI_DOUBLE, MPI_SUM, rank, _commState->comm);
 }
 
-void MPIDirectCommunication::reduceSum(double const *itemsToSend, double *itemsToReceive, int size, Rank rankMaster)
+void MPIDirectCommunication::reduceSum(precice::span<double const> itemsToSend, precice::span<double> itemsToReceive, Rank rankMaster)
```
BenjaminRodenberg, line 88: Code duplication of `void MPIDirectCommunication::reduceSum(precice::span<double const> itemsToSend, precice::span<double> itemsToReceive)`?

src/com/MPIDirectCommunication.cpp
```diff
@@ -102,16 +105,18 @@ void MPIDirectCommunication::reduceSum(int itemToSend, int &itemsToReceive, Rank
...
 {
-  PRECICE_TRACE(size);
-  MPI_Allreduce(const_cast<double *>(itemsToSend), itemsToReceive, size, MPI_DOUBLE, MPI_SUM, _commState->comm);
+  PRECICE_TRACE(itemsToSend.size());
+  PRECICE_ASSERT(itemsToSend.size() == itemsToReceive.size());
+  MPI_Allreduce(const_cast<double *>(itemsToSend.data()), itemsToReceive.data(), itemsToSend.size(), MPI_DOUBLE, MPI_SUM, _commState->comm);
 }
 
-void MPIDirectCommunication::allreduceSum(double const *itemsToSend, double *itemsToReceive, int size, Rank rankMaster)
+void MPIDirectCommunication::allreduceSum(precice::span<double const> itemsToSend, precice::span<double> itemsToReceive, Rank rankMaster)
```
BenjaminRodenberg, line 115: Code duplication of `void MPIDirectCommunication::allreduceSum(precice::span<double const> itemsToSend, precice::span<double> itemsToReceive)`?

src/com/MPIDirectCommunication.cpp
```diff
@@ -138,56 +143,52 @@ void MPIDirectCommunication::allreduceSum(int itemToSend, int &itemToReceive, Ra
...
-  MPI_Bcast(itemsToReceive, size, MPI_INT, rankBroadcaster, _commState->comm);
+  PRECICE_TRACE(itemsToReceive.size());
+  MPI_Bcast(itemsToReceive.data(), itemsToReceive.size(), MPI_INT, rankBroadcaster, _commState->comm);
 }
 
 void MPIDirectCommunication::broadcast(int itemToSend)
 {
   PRECICE_TRACE();
-  broadcast(&itemToSend, 1);
+  broadcast(precice::span<const int>{&itemToSend, 1});
```
BenjaminRodenberg, line 161: Why `1`? Could you replace this magic number with something more descriptive?

src/com/MPIDirectCommunication.cpp
```diff
@@ -138,56 +143,52 @@ void MPIDirectCommunication::allreduceSum(int itemToSend, int &itemToReceive, Ra
...
 }
 
 void MPIDirectCommunication::broadcast(int &itemToReceive, Rank rankBroadcaster)
 {
   PRECICE_TRACE();
-  broadcast(&itemToReceive, 1, rankBroadcaster);
+  broadcast(precice::span<int>{&itemToReceive, 1}, rankBroadcaster);
```
BenjaminRodenberg, line 167: See above.

src/com/MPIDirectCommunication.cpp
```diff
@@ -138,56 +143,52 @@ void MPIDirectCommunication::allreduceSum(int itemToSend, int &itemToReceive, Ra
...
 }
 
-void MPIDirectCommunication::broadcast(const double *itemsToSend, int size)
+void MPIDirectCommunication::broadcast(precice::span<const double> itemsToSend)
 {
-  PRECICE_TRACE(size);
-  MPI_Bcast(const_cast<double *>(itemsToSend), size, MPI_DOUBLE, 0, _commState->comm);
+  PRECICE_TRACE(itemsToSend.size());
+  MPI_Bcast(const_cast<double *>(itemsToSend.data()), itemsToSend.size(), MPI_DOUBLE, 0, _commState->comm);
```
BenjaminRodenberg, line 173: Why `0`? Could you replace this magic number with something more descriptive?

src/com/MPIDirectCommunication.cpp
```diff
@@ -138,56 +143,52 @@ void MPIDirectCommunication::allreduceSum(int itemToSend, int &itemToReceive, Ra
...
-  MPI_Bcast(itemsToReceive, size, MPI_DOUBLE, rankBroadcaster, _commState->comm);
+  PRECICE_TRACE(itemsToReceive.size());
+  MPI_Bcast(itemsToReceive.data(), itemsToReceive.size(), MPI_DOUBLE, rankBroadcaster, _commState->comm);
 }
 
 void MPIDirectCommunication::broadcast(double itemToSend)
 {
   PRECICE_TRACE();
-  broadcast(&itemToSend, 1);
+  broadcast(precice::span<const double>{&itemToSend, 1});
```
BenjaminRodenberg, line 185: Why `1`? Could you replace this magic number with something more descriptive?

src/com/MPIDirectCommunication.cpp
```diff
@@ -138,56 +143,52 @@ void MPIDirectCommunication::allreduceSum(int itemToSend, int &itemToReceive, Ra
...
 }
 
 void MPIDirectCommunication::broadcast(double &itemToReceive, Rank rankBroadcaster)
 {
   PRECICE_TRACE();
-  broadcast(&itemToReceive, 1, rankBroadcaster);
+  broadcast(precice::span<double>{&itemToReceive, 1}, rankBroadcaster);
```
BenjaminRodenberg, line 191: Why `1`? Could you replace this magic number with something more descriptive?

src/com/SocketCommunication.cpp
```diff
@@ -491,7 +491,7 @@ void SocketCommunication::send(double itemToSend, Rank rankReceiver)
 
 PtrRequest SocketCommunication::aSend(const double &itemToSend, Rank rankReceiver)
 {
-  return aSend(&itemToSend, 1, rankReceiver);
+  return aSend(precice::span<const double>{&itemToSend, 1}, rankReceiver);
```
BenjaminRodenberg, line 494: What do the curly braces do? Why `1`?

src/com/SocketCommunication.cpp
```diff
@@ -671,7 +670,7 @@ void SocketCommunication::receive(double &itemToReceive, Rank rankSender)
 
 PtrRequest SocketCommunication::aReceive(double &itemToReceive, Rank rankSender)
 {
-  return aReceive(&itemToReceive, 1, rankSender);
+  return aReceive(precice::span<double>{&itemToReceive, 1}, rankSender);
```
BenjaminRodenberg, line 673: See above.

src/utils/MasterSlave.cpp
```diff
@@ -161,16 +161,23 @@ void MasterSlave::reduceSum(double *sendData, double *rcvData, int size)
...
   }
 }
 
-void MasterSlave::reduceSum(int &sendData, int &rcvData, int size)
+void MasterSlave::reduceSum(const double &sendData, double &rcvData)
+{
+  PRECICE_TRACE();
+  reduceSum(precice::span<const double>{&sendData, 1},
+            precice::span<double>{&rcvData, 1});
+}
```
BenjaminRodenberg, line 178: Why is `void MasterSlave::reduceSum(const double &sendData, double &rcvData)` so short and `void MasterSlave::reduceSum(const int &sendData, int &rcvData)` so long? Shouldn't both functions look (more or less) the same?

src/utils/MasterSlave.hpp
```diff
@@ -58,21 +59,23 @@ class MasterSlave {
...
+  static void allreduceSum(double &sendData, double &rcvData);
+
+  static void allreduceSum(int &sendData, int &rcvData);
 
   static void broadcast(bool &value);
 
   static void broadcast(double &value);
 
-  static void broadcast(double *values, int size);
+  static void broadcast(precice::span<double> values);
```
BenjaminRodenberg, line 78: Move this function up for consistency:

```
  static void broadcast(precice::span<double> values);

  static void broadcast(bool &value);

  static void broadcast(double &value);
```

BenjaminRodenberg (approved):
As far as I understand the main point is to get rid of passing `size` as an argument by using `precice::span<...>`. This makes the code more understandable and less verbose. So nice :+1: 

I added some minor comments below:

* Code duplication (not sure if this is the point of this PR and maybe there is a reason, why you preferred to implement more or less the same function twice)
* Magic numbers (this is legacy and might also go into another PR at a later point in time)
* C++20 span: As soon as we move to C++20, we can use the implementation from C++20. Should we document this todo in an issue? Then we will know what we have to do as soon as we upgrade to C++20.

src/com/Communication.cpp
```diff
@@ -38,28 +39,30 @@ void Communication::connectMasterSlaves(std::string const &participantName,
...
+  PRECICE_ASSERT(itemsToSend.size() == itemsToReceive.size());
 
-  std::copy(itemsToSend, itemsToSend + size, itemsToReceive);
+  std::copy(itemsToSend.begin(), itemsToSend.end(), itemsToReceive.begin());
 
-  std::vector<double> received(size);
+  std::vector<double> received(itemsToReceive.size());
   // receive local results from slaves
   for (size_t rank = 0; rank < getRemoteCommunicatorSize(); ++rank) {
-    auto request = aReceive(received.data(), size, rank + _rankOffset);
```
DavidSCN, line 50: Shouldn't we be able t o select here the type `Rank` instead of `size_t`?

DavidSCN (commented):
I did not went through the span implementation itself as I assume it is fine. Are we licence-wise on the safe side here?
Other than that, looks good. I hope our tests coverage for the communication is sufficiently (although the compiler delivers probably quite some consistency here).

src/com/Communication.cpp
```diff
@@ -38,28 +39,30 @@ void Communication::connectMasterSlaves(std::string const &participantName,
...
+  PRECICE_TRACE(itemsToSend.size(), itemsToReceive.size());
+  PRECICE_ASSERT(itemsToSend.size() == itemsToReceive.size());
 
-  std::copy(itemsToSend, itemsToSend + size, itemsToReceive);
+  std::copy(itemsToSend.begin(), itemsToSend.end(), itemsToReceive.begin());
 
-  std::vector<double> received(size);
+  std::vector<double> received(itemsToReceive.size());
   // receive local results from slaves
   for (size_t rank = 0; rank < getRemoteCommunicatorSize(); ++rank) {
```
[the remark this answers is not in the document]
DavidSCN: IIRC I had already a discussion with @fsimonis and we agreed on using `++rank` instead of `++rank`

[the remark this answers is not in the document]
BenjaminRodenberg: @DavidSCN you wrote `++rank` twice. On what did you agree? As long as it's consistent I'm fine. Personally `rank++` is more natural to be, but I don't really mind.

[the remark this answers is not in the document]
DavidSCN: Oops, corrected it :D
Yes, more naturally, but has some disadvantage I don't remember.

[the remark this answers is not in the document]
fsimonis: `++thing` increments the thing and returns a reference to the thing.

`thing++` saves a copy, then increments the thing and then returns the copy. Using this without using the return value doesn't really make sense.

Compilers know how to get rid of the unnecessary work in most cases. Hence this doesn't matter for most cases. Most cases is not every case though. Imagine some fancy iterator that does some extra work on advance like evaluating a function or filtering something.

So, if `++thing` is what you want to do, you should use it. Using even in cases it doesn't matter simply is a matter of consistency.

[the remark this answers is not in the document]
BenjaminRodenberg: Sounds like a good reason for keeping `++thing`. I will try to get used to `++thing`, as well, and change my habits.

Interesting: In the CPP reference a for loop also uses `++thing` (https://en.cppreference.com/w/cpp/language/for). So I have no idea, why the `thing++` is more natural for me in this context.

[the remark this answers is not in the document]
MakisH: > So I have no idea, why the `thing++` is more natural for me in this context.

It is more natural because we see it more often (see... C++).

To demystify this, look at the signatures of the methods in the table at the top of this page: https://en.cppreference.com/w/cpp/language/operator_incdec The `thing++` takes an argument by value, returns a value. The `++thing` takes a reference and returns a reference. As @fsimonis wrote, in most cases it does not matter, but there can be specific cases where it does.

There seems to be no [core guideline](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md) for this. I would just try to use `++thing` by default for the same reason as we are using references by default: safety (no accidental slicing of objects) and efficiency (no unnecessary copying around).

MakisH:
I have not looked in detail in the changes, but on the high-level, this change is very good: it improves safety, does not need extra dependencies, and makes the code C++20-like, without having to increase the standard. Very nice! :+1:

src/acceleration/test/ParallelMatrixOperationsTest.cpp
```diff
@@ -106,12 +106,12 @@ BOOST_AUTO_TEST_CASE(ParVectorOperations)
...
-  utils::MasterSlave::allreduceSum(iaa, ires2, 1);
-  utils::MasterSlave::allreduceSum(aa.data(), res2.data(), 2);
+  utils::MasterSlave::allreduceSum(a, res1);
+  utils::MasterSlave::allreduceSum(iaa, ires2);
+  utils::MasterSlave::allreduceSum(aa, res2);
 
-  utils::MasterSlave::reduceSum(aa.data(), res3.data(), 2);
-  utils::MasterSlave::reduceSum(iaa, ires1, 1);
+  utils::MasterSlave::reduceSum(aa, res3);
+  utils::MasterSlave::reduceSum(iaa, ires1);
```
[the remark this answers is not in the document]
fsimonis: I agree, but is is off-topic :wink:

src/com/SocketCommunication.cpp
fsimonis: This constructs a span with a single element. I admit that this is ugly.
Maybe something like `precice::wrapInSpan<const double>(*itemToSend)` would be more readable.
Using braces calls the constructor.

src/com/Communication.cpp
```diff
@@ -81,49 +84,40 @@ void Communication::reduceSum(int itemToSend, int &itemToReceive, Rank rankMaste
...
-      itemsToReceive[i] += received[i];
-    }
-  }
+  reduceSum(itemsToSend, itemsToReceive);
 
   // send reduced result to all slaves
-  std::vector<PtrRequest> requests(getRemoteCommunicatorSize());
+  std::vector<PtrRequest> requests;
+  requests.reserve(getRemoteCommunicatorSize());
   for (size_t rank = 0; rank < getRemoteCommunicatorSize(); ++rank) {
```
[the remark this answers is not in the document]
fsimonis: Discussed before

src/com/Communication.cpp
```diff
@@ -231,26 +225,25 @@ void Communication::broadcast(int &itemToReceive, Rank rankBroadcaster)
   receive(itemToReceive, rankBroadcaster + _rankOffset);
 }
 
-void Communication::broadcast(const double *itemsToSend, int size)
+void Communication::broadcast(precice::span<const double> itemsToSend)
```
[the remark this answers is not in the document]
fsimonis: Possible, but this is a bit beyond the reach of this PR.

src/com/MPICommunication.cpp
fsimonis: We could avoid it, but the alternatives are more verbose and complex (templating, partial specialisation etc)

src/com/MPIDirectCommunication.cpp
fsimonis: See previous comment

src/com/MPIDirectCommunication.cpp
fsimonis: See previous comment

src/com/MPIDirectCommunication.cpp
fsimonis: Added a helper for this case.

src/com/MPIDirectCommunication.cpp
fsimonis: True, but a bit too off topic.

src/utils/MasterSlave.cpp
fsimonis: I reused the reduceSum for "multiple doubles" to implement "one double".
There is no "multiple ints" version, so I had to keep that.

fsimonis:
tcbrindle/span is licensed under the [Boost Software License 1.0](https://github.com/tcbrindle/span/blob/master/LICENSE_1_0.txt).
So there is no conflict here.

PR #1058, opened 2021-08-03
Add Rank as data in parallel vtk exports

fsimonis:
## Main changes of this PR

When exporting vtk in parallel participant, this PR adds the additional scalar data `Rank`, which contains the rank of the partition.

Open questions:
* Is the name `Rank` too general?
* Could the name collide with a user-data name? Shall we use something longer like `preCICERank`?
* Shall we reserve that name via the configuration?
* Should also the non-parallel vtk file include this data as a constant?

This simplifies debugging vtk exports from parallel solvers, as loading the master file now contains the partitioning information. This was not obvious in the absence of connectivity information. Now using a point gaussian with the rank colouring shows the partitions for such cases.

This also simplifies processing vtk files in external tools, such as ASTE.

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/docs/changelog/1058.md
+++ b/docs/changelog/1058.md
@@ -1,0 +1,1 @@
+- Added the `Rank` of partitions in parallel VTK exports
--- a/docs/changelog/1058.md
+++ b/docs/changelog/1058.md
@@ -1,1 +1,1 @@
-- Added the `Rank` of partitions in parallel VTK exports
+- Added the `Rank` of partitions in VTK exports
--- a/src/io/ExportVTK.cpp
+++ b/src/io/ExportVTK.cpp
@@ -97,6 +97,10 @@ void ExportVTK::exportData(std::ofstream &outFile, mesh::Mesh const &mesh)
 {
   outFile << "POINT_DATA " << mesh.vertices().size() << "\n\n";
 
+  outFile << "SCALARS Rank unsigned_int\n";
+  std::fill_n(std::ostream_iterator<char const *>(outFile), mesh.vertices().size(), "0 ");
+  outFile << "\n\n";
+
   for (const mesh::PtrData &data : mesh.data()) { // Plot vertex data
     Eigen::VectorXd &values = data->values();
     if (data->getDimensions() > 1) {
--- a/src/io/ExportVTKXML.cpp
+++ b/src/io/ExportVTKXML.cpp
@@ -240,7 +240,7 @@ void ExportVTKXML::exportData(
   outFile << "\">\n";
 
   // Export the current rank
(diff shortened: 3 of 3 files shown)


DavidSCN:
>Open questions:

In my opinion:

>   Is the name Rank too general?

No, that's just fine.

>    Could the name collide with a user-data name? Shall we use something longer like preCICERank?

It would not even be a problem. paraView can visualize both.

>    Shall we reserve that name via the configuration?

You mean adding a config option for the name?

>    Should also the non-parallel vtk file include this data as a constant?

Yes, please.

DavidSCN:
> I meant, disallowing the user to choose Rank as a data name, but as it is not necessary as it isn't an issue.

Ah ok, now I get it. This might be an issue, indeed. Maybe `preciceRank` is better.

MakisH:
If I understand correctly, this adds one scalar in each VTK file, describing the MPI rank the points come from. I assume that, at the moment, each rank writes a different VTK file.

This is definitely nice-to-have, and if you already found a use case, that's great! Often the MPI rank is only implied by the filename, but having it as proper VTK data is nice. I have never seen this approach before.

> * Is the name `Rank` too general?

This is clear.

> * Could the name collide with a user-data name? Shall we use something longer like `preCICERank`?

Definitely making it more specific would be fine (or even better), but I struggle to imagine a use case for defining a user-data `Rank` (especially to not mean "MPI rank"). What would happen in that case, since you mentioned it?

> * Shall we reserve that name via the configuration?

I am not sure if this is really necessary. I guess we are in low effort - low benefit territory, so just do whatever you find easier.

> * Should also the non-parallel vtk file include this data as a constant?

Yes, this would make the VTK files more consistent and predictable.

BenjaminRodenberg:
> ## Main changes of this PR
> 
> When exporting vtk in parallel participant, this PR adds the additional scalar data `Rank`, which contains the rank of the partition.

I also think this is very useful!

> Open questions:
> 
>     * Is the name `Rank` too general?

From my perspective this is fine.
 
>     * Could the name collide with a user-data name? Shall we use something longer like `preCICERank`?
>
>     * Shall we reserve that name via the configuration?

For now, I would suggest to not worry. As @MakisH already described this might be a high-effort/low-gain thing. As far as I understand we can still add the option to have a user-defined name for the rank in the exported vtk file at a later point in time - if this really becomes necessary.

fsimonis:
Ok, so let's merge this now.

If there are issues with double usage of `Rank` then we can still change this before the next release.

uekerman:
Nice addition!

> I have never seen this approach before

Alya uses this as well.

PR #1061, opened 2021-08-04
Remove legacy VTK exporter

fsimonis:
## Main changes of this PR

This PR removes the legacy VTK exporter and replaces it by the VTK XML exporter, which works for both serial and parallel participants.

We need to update the docs, as preCICE will export VTK files, which have the `.vtu` extension.
There will be no more `.vtk` files.

The XML VTK exporter was introduced to handle parallel vtk exports, but the legacy VTK export was never removed.

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)


src/io/ExportVTK.hpp
```diff
@@ -3,57 +3,88 @@
...
+class Edge;
+class Triangle;
+} // namespace mesh
 } // namespace precice
 
 namespace precice {
 namespace io {
 
-/// Writes polygonal, or triangle meshes to vtk files.
+/// Writes meshes to xml-vtk files. Only for parallel usage. Serial usage (coupling mode) should still use ExportVTK
```
DavidSCN, line 21: Could you explain this comment more in detail? Why distinguish these cases here? The class is actually called `ExportVTK`.

DavidSCN (commented):
Could you explain these changes and the motivation a bit more in detail? IIRC this PR does not only remove the former VTK exporter but also replaces it with some other functionality.

uekerman:
I am bit unsure whether we should actually regard this change as breaking.
Tutorial scrips depend on the vtk ending, e.g. https://github.com/precice/tutorials/blob/master/flow-over-heated-plate/plot-final-interface-temperature.py
Users might also have their own scripts already.
@MakisH what do you think?

src/io/ExportVTK.hpp
fsimonis: That's a copy-paste artefact. Thanks for spotting it!

fsimonis:
@DavidSCN 
Currently, we have 2 exporters for VTK: `ExportVTK` for legacy VTK and `ExportVTKXML` for parallel  xml-based VTK.

When defining a `<export:vtk />` tag, preCICE either selects the legacy VTK exporter for serial participants or the XML VTK exporter for parallel participants.
This is necessary as legacy VTK does not support multiple parallel files.

| | `ExportVTK` files  | `ExportVTKXML` files
--- | --- | ---
Serial | `.vtk` | `.vtu`
Parallel | :x: | `.vtu` + `.pvtu` 

The XML VTK exporter supports both serial and parallel participants though, so there is no real reason to keep the legacy VTK exporter around. In fact, removing the legacy exporter will remove the selection process, which makes some code simpler and more readable.

src/io/io.dox
```diff
@@ -10,7 +10,7 @@ currently implemented:
 Export:
 
 - Geometry VTK (geometry and data, for serial computations only)
-- Geometry VTKXML (geometry and data, for parallel computations only)
+- Geometry VTK (geometry and data, for parallel computations only)
```
MakisH, line 13: I guess this part would be simplified and be merged to one?

MakisH (commented):
Even though I like this simplification, I think that (as it is now) it should wait for v3.

I am also not sure we should change this in a feature release. We promise that users can painlessly upgrade preCICE and continue their work without breaking anything. What if I have already setup my simulation workflow to automatically post-process my `.vtk` files? I upgrade to preCICE v2.3.0 and suddenly there are no `.vtk` files anymore and my scripts break. I need to read the changelog and change my scripts.

In our tutorials, the only example of actually processing these files is in the [flow-over-heated-plate](https://github.com/precice/tutorials/blob/6d90ae3eb179113beb400e472c8832e29a48db9b/flow-over-heated-plate/plot-final-interface-temperature.py#L7). We do also export VTK files in [quickstart](https://github.com/precice/tutorials/blob/152fd79ce2a6556fc3fe9804c18ea5006ef9b665/quickstart/precice-config.xml), while we often encourage users to use this feature.

We recently defined "breaking changes" in the OpenFOAM adapter as changes requiring updates in the tutorials configuration (as we had trouble defining what the API of the adapter is). I think this is a perspective we can also use here.

**Alternatively,** is there any easy way to convert from `vtu` to `vtk`? Then we could simplify the selection already now and add (short) backwards-compatibility code that we can then remove in v3.

By the way, why have we concluded that `.vtk` does not support parallel cases at all? The legacy way is to simply add the rank into the name of the VTK file. and then open all the files together and group them, e.g., in ParaView. I agree that this is not intrinsic support by the format, but the format can be used in this way.

fsimonis:
I am surprised that this is considered such a massive change.

From a user perspective, only the filename changes. Both formats encode the same information and are both supported by the vtk library as well as paraview.
So, the required change is `s/vtk/vtu/`. Everything else should be handled completely transparently by the used libraries.

MakisH:
> I am surprised that this is considered such a massive change.

I do not consider it massive, I just consider it breaking, in the sense that it breaks workflows. Whether output files is in our extended definition of API to preserve, is a different story. During the v1.x era we had quite a few such changes and I think we confused users back then. We have already managed this much more nicely in v2.x so far.

Putting us in the same position, we will also need to adjust our tutorial scripts so that they work both with preCICE v2.3+ and v.2.0-2.2. And we will need to tell users to also update their tutorials version. Yes, for only one file, but still.

> From a user perspective, only the filename changes.

Yes, but this is something that breaks automation. In a very smalll, usually auxiliary, and easily fixable aspect, but we break the trust contract with the user.

I am not against this change, I just want to see stronger motivation for including it now, or some backwards compatibility measure (we can probably think of something here). If this is PR needed to develop something else, then sure. If this is a nice to have, then it should wait.

fsimonis:
Removing the VTK exporter is not a clean solution.
My planned approach to this issues is as follows:
* rename the `ExportVTKXML` exporter to `ExportVTU`.
* add a new exporter for VTU only `<export:vtu />`.
* display a warning when `<export:vtk />` is used with a parallel participant. After all, the result are `.pvtu` and `.vtu` files.

PR #1065, opened 2021-08-12
Replace c-style casts by static cast

DavidSCN:
## Main changes of this PR
Replace c-style casts by static cast as part #999 .

C-style casts are discouraged

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/acceleration/AitkenAcceleration.cpp
+++ b/src/acceleration/AitkenAcceleration.cpp
@@ -60,7 +60,7 @@ void AitkenAcceleration::performAcceleration(
   Eigen::VectorXd oldValues;
   for (int id : _dataIDs) {
     utils::append(values, cplData[id]->values());
-    utils::append(oldValues, (Eigen::VectorXd) cplData[id]->previousIteration());
+    utils::append(oldValues, Eigen::VectorXd(cplData[id]->previousIteration()));
   }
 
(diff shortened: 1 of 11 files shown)


fsimonis (commented):
I am confused about the casts from `Eigen::VectorXd::Zero()` to `Eigen::VectorXd`.
This shouldn't be necessary.

DavidSCN:
@fsimonis you mean the cast itself rather than the proposed change, right?

DavidSCN:
@fsimonis `append` is templated and the compiler cannot deduce the argument from the return type of `zero`.

fsimonis (approved):
We should have a look at this eigen append again in the future.

DavidSCN:
A pretty simple solution would be the following:
```diff
-template <typename Derived1>
+template <typename VectorType>
 void append(
-    Eigen::VectorXd &                       v,
-    const Eigen::PlainObjectBase<Derived1> &app)
+    Eigen::VectorXd & v,
+    const VectorType &app)
```
so that the compiler can easily deduce the argument.

fsimonis:
Given that appending zero is a common call to append, we could introduce `appendZeros(vec, count)`.
This makes things even less verbose.

PR #1069, opened 2021-08-23
Add equals function for comparing std::vector<double>

BenjaminRodenberg:
## Main changes of this PR

* Add `boost::test_tools::predicate_result equals<DerivedA, DerivedB>(std::vector<DerivedA> &A, std::vector<DerivedB> &B, double tolerance = precice::math::NUMERICAL_ZERO_DIFFERENCE)`
* Don't use `BOOST_TEST(vecA == vecB)`, but `BOOST_TEST(testing::equals(vecA, vecB))` with some floating point tolerance.

A `BOOST_TEST(vecA == vecB)` caused a lot of confusion in https://github.com/BenjaminRodenberg/precice/pull/1. We were lucky until now that floating point accuracy did not matter here, but it will as soon as interpolation (or something else) happens when data is exchanged.

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/m2n/tests/PointToPointCommunicationTest.cpp
+++ b/src/m2n/tests/PointToPointCommunicationTest.cpp
@@ -108,12 +108,12 @@ void runP2PComTest1(const TestContext &context, com::PtrCommunicationFactory cf)
     c.send(data);
     c.receive(data);
 
-    BOOST_TEST(data == expectedData);
+    BOOST_TEST(testing::equals(data, expectedData));
   } else {
     c.acceptConnection("B", "A");
 
     c.receive(data);
-    BOOST_TEST(data == expectedData);
+    BOOST_TEST(testing::equals(data, expectedData));
     process(data);
     c.send(data);
   }
@@ -183,12 +183,12 @@ void runP2PComTest2(const TestContext &context, com::PtrCommunicationFactory cf)
 
     c.send(data);
     c.receive(data);
-    BOOST_TEST(data == expectedData);
+    BOOST_TEST(testing::equals(data, expectedData));
   } else {
     c.acceptConnection("B", "A");
 
     c.receive(data);
-    BOOST_TEST(data == expectedData);
+    BOOST_TEST(testing::equals(data, expectedData));
     process(data);
     c.send(data);
   }
--- a/src/precice/tests/ParallelTests.cpp
+++ b/src/precice/tests/ParallelTests.cpp
@@ -865,7 +865,7 @@ void runTestAccessReceivedMesh(const std::string         configName,
 
     // Check the received vertex coordinates
     std::vector<double> expectedPositions = context.isMaster() ? std::vector<double>({0.0, 1.0, 0.0, 2.0, 0.0, 3.0}) : expectedPositionSlave;
-    BOOST_TEST(expectedPositions == coordinates);
+    BOOST_TEST(testing::equals(expectedPositions, coordinates));
 
     // Check the received vertex IDs (IDs are local?!)
     std::vector<int> expectedIDs;
@@ -915,7 +915,7 @@ void runTestAccessReceivedMesh(const std::string         configName,
       std::vector<double> ownCoordinates(ownMeshSize * dim);
       interface.getMeshVerticesAndIDs(meshID, ownMeshSize, ownIDs.data(), ownCoordinates.data());
       BOOST_TEST(ownIDs == ids);
-      BOOST_TEST(ownCoordinates == ownCoordinates);
+      BOOST_TEST(testing::equals(ownCoordinates, ownCoordinates));  // @todo looks wrong
     }
 
     // Initialize the solverinterface
@@ -1095,7 +1095,7 @@ BOOST_AUTO_TEST_CASE(AccessReceivedMeshAndMapping)
       // Expected data according to the writeData
       // Values are summed up
       std::vector<double> expectedData = context.isMaster() ? std::vector<double>({0, 1, 0}) : std::vector<double>({1, 2, 2});
-      BOOST_TEST(expectedData == readData);
+      BOOST_TEST(testing::equals(expectedData, readData));
     }
 
   } else {
@@ -1130,7 +1130,7 @@ BOOST_AUTO_TEST_CASE(AccessReceivedMeshAndMapping)
       // Expected data according to the writeData
       // Values are summed up
       std::vector<double> expectedData = context.isMaster() ? std::vector<double>({15, 16}) : std::vector<double>({22, 6, 7});
-      BOOST_TEST(expectedData == readData);
+      BOOST_TEST(testing::equals(expectedData, readData));
     }
   }
 }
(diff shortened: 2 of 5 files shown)


src/testing/Testing.hpp
```diff
@@ -82,6 +82,23 @@ boost::test_tools::predicate_result equals(const Eigen::MatrixBase<DerivedA> &A,
...
+template <typename TA, typename TB>
+boost::test_tools::predicate_result equals(const std::vector<TA> &A,
+                                           const std::vector<TB> &B,
+                                           double                 tolerance = math::NUMERICAL_ZERO_DIFFERENCE)
+{
+  Eigen::MatrixXd MatrixA(A.size(), 1);
+  for (int i = 0; i < A.size(); ++i) {
+    MatrixA(i, 0) = A[i];
+  }
+  Eigen::MatrixXd MatrixB(B.size(), 1);
```
DavidSCN, line 95: I think you could do something like `std::copy(A.begin(), A.end(), MatrixA.data())` or is it not possible?

src/testing/Testing.hpp
```diff
@@ -82,6 +82,23 @@ boost::test_tools::predicate_result equals(const Eigen::MatrixBase<DerivedA> &A,
   return true;
 }
 
+/// equals to be used in tests. Prints both operatorans on failure
+template <typename TA, typename TB>
```
DavidSCN, line 86: Two remarks: 

- this is only for test purpose, I know, but it would be nice to select more descriptive names for the template arguments. Here, the parameter replace the number type, so maybe `Number` or `NumerType` would be better.
- is it really intended to compare containers with different types? What happens if one container uses `double` precision and the other `float`? Are they still the same if the tolerance is satisfied? According to my understanding, all three input arguments should have the same basic type?

src/testing/Testing.hpp
```diff
@@ -82,6 +82,23 @@ boost::test_tools::predicate_result equals(const Eigen::MatrixBase<DerivedA> &A,
...
+boost::test_tools::predicate_result equals(const std::vector<TA> &A,
+                                           const std::vector<TB> &B,
+                                           double                 tolerance = math::NUMERICAL_ZERO_DIFFERENCE)
+{
+  Eigen::MatrixXd MatrixA(A.size(), 1);
```
DavidSCN, line 91: What happens if the two input container do not have the same size? I would propose to assert it here and exit immediately as they are not the same(?)

src/m2n/tests/PointToPointCommunicationTest.cpp
```diff
@@ -108,12 +108,12 @@ void runP2PComTest1(const TestContext &context, com::PtrCommunicationFactory cf)
     c.send(data);
     c.receive(data);
 
-    BOOST_TEST(data == expectedData);
```
DavidSCN, line 111: Could we prevent the utilization of the 'old* equal method?

DavidSCN (commented):
Only some general feedback comments. Feel free to merge.

src/m2n/tests/PointToPointCommunicationTest.cpp
BenjaminRodenberg: Not sure if this is possible or even reasonable. We still need the normal `==` at many places and for some datatypes it is also totally fine to use this operator. I would say it's the job of the programmer to make sure to not compare floating point values via `==`, since this can work, but is generally unsafe.

src/testing/Testing.hpp
BenjaminRodenberg: Seems to work. Nice :)

src/testing/Testing.hpp
BenjaminRodenberg: > Two remarks:
> 
> * this is only for test purpose, I know, but it would be nice to select more descriptive names for the template arguments. Here, the parameter replace the number type, so maybe `Number` or `NumerType` would be better.

Will do this.

> * is it really intended to compare containers with different types? What happens if one container uses `double` precision and the other `float`? Are they still the same if the tolerance is satisfied? According to my understanding, all three input arguments should have the same basic type?

I will try to change this function to only use a single template parameter.

src/testing/Testing.hpp
BenjaminRodenberg: Will do this.

src/precice/tests/ParallelTests.cpp
```diff
@@ -915,7 +915,7 @@ void runTestAccessReceivedMesh(const std::string         configName,
       std::vector<double> ownCoordinates(ownMeshSize * dim);
       interface.getMeshVerticesAndIDs(meshID, ownMeshSize, ownIDs.data(), ownCoordinates.data());
       BOOST_TEST(ownIDs == ids);
-      BOOST_TEST(ownCoordinates == ownCoordinates);
+      BOOST_TEST(testing::equals(ownCoordinates, ownCoordinates)); // @todo looks wrong
```
BenjaminRodenberg, line 918: @DavidSCN or @fsimonis can you comment on this line? I will merge the PR now, because this line is not related, but we should remove or fix it.

fsimonis: Yes,this looks wrong :wink:

BenjaminRodenberg: Ok. Any Idea how to make this less wrong? :smirk:

DavidSCN: My bad, I will fix it

DavidSCN: Done

fsimonis:
@BenjaminRodenberg I am not very keen on this addition for some reasons:
* Boost.Test already compares floating point numbers using [a tolerance](https://github.com/precice/precice/blob/cff1b7fff5b8c80e62a0e6c3404e4c045577cb4e/src/testing/main.cpp#L96)
* The predicate doesn't show where the ranges differ
* The predicate doesn't implement a ["real" predicate](https://www.boost.org/doc/libs/1_65_1/libs/test/doc/html/boost_test/testing_tools/custom_predicates.html)
* The predicate only needs one template parameter. Comparing `std::vector<double>` to `std::vector<int>` is a mistake in the test.
* There is already a [build-in tool](https://www.boost.org/doc/libs/1_65_1/libs/test/doc/html/boost_test/testing_tools/extended_comparison/collections.html) for this:  
   ```cpp
    BOOST_TEST(a == b, boost::test_tools::per_element() );
    ```

BenjaminRodenberg:
> @BenjaminRodenberg I am not very keen on this addition for some reasons:
> 
>     * Boost.Test already compares floating point numbers using [a tolerance](https://github.com/precice/precice/blob/cff1b7fff5b8c80e62a0e6c3404e4c045577cb4e/src/testing/main.cpp#L96)
> 
>     * The predicate doesn't show where the ranges differ
> 
>     * The predicate doesn't implement a ["real" predicate](https://www.boost.org/doc/libs/1_65_1/libs/test/doc/html/boost_test/testing_tools/custom_predicates.html)
> 
>     * The predicate only needs one template parameter. Comparing `std::vector<double>` to `std::vector<int>` is a mistake in the test.
> 
>     * There is already a [build-in tool](https://www.boost.org/doc/libs/1_65_1/libs/test/doc/html/boost_test/testing_tools/extended_comparison/collections.html) for this:
>       ```c++
>        BOOST_TEST(a == b, boost::test_tools::per_element() );
>       ```

@fsimonis can you open an issue? I fear otherwise these comments might get lost.

PR #1072, opened 2021-08-31
Add support for 3D meshes in ScaleByAreaAction

### LINKED ISSUE #752: Implement ScaleByArea for 3D Meshes
The ScaleByArea action is implemented only for the 2D case. It would be nice to have a working and tested implementation to handle the 3D case as well.

Implement handling of the 3D case in following function:

https://github.com/precice/precice/blob/47747abf9babb977a6b45edbe16ac69ec8d6948b/src/action/ScaleByAreaAction.cpp#L20-L24

_do nothing_

**Additional context**
The PR https://github.com/precice/precice/pull/727 already ran into issues when enforcing 2D-only on a configuration-basis.

mhmdkanj:
## Main changes of this PR

Extend the ScaleByAreaAction to apply to 3D meshes as well instead of 2D meshes only.

closes #752 

## Author's checklist

* [x] I added a changelog file with this PR number in `docs/changelog/` if there are noteworthy changes -> **Not so worthy to add to Changelog**

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/action/ScaleByAreaAction.cpp
+++ b/src/action/ScaleByAreaAction.cpp
@@ -31,26 +31,35 @@ void ScaleByAreaAction::performAction(
     double timeWindowSize)
 {
   PRECICE_TRACE();
-  PRECICE_ASSERT(getMesh()->getDimensions() == 2, "The ScaleByAreaAction requires a mesh of dimensionality 2.");
-  auto &          targetValues = _targetData->values();
-  Eigen::VectorXd areas        = Eigen::VectorXd::Zero(getMesh()->vertices().size());
-  for (mesh::Edge &edge : getMesh()->edges()) {
-    areas[edge.vertex(0).getID()] += edge.getEnclosingRadius();
-    areas[edge.vertex(1).getID()] += edge.getEnclosingRadius();
+  const int       meshDimensions  = getMesh()->getDimensions();
+  auto &          targetValues    = _targetData->values();
+  const int       valueDimensions = _targetData->getDimensions();
+  Eigen::VectorXd areas           = Eigen::VectorXd::Zero(getMesh()->vertices().size());
+  PRECICE_ASSERT(targetValues.size() / valueDimensions == areas.size());
+
+  if (meshDimensions == 2) {
+    for (mesh::Edge &edge : getMesh()->edges()) {
+      areas[edge.vertex(0).getID()] += edge.getEnclosingRadius();
+      areas[edge.vertex(1).getID()] += edge.getEnclosingRadius();
+    }
+  } else {
+    for (mesh::Triangle &face : getMesh()->triangles()) {
+      areas[face.vertex(0).getID()] += face.getArea() / 3.0;
+      areas[face.vertex(1).getID()] += face.getArea() / 3.0;
+      areas[face.vertex(2).getID()] += face.getArea() / 3.0;
+    }
   }
-  int dimensions = _targetData->getDimensions();
-  PRECICE_ASSERT(targetValues.size() / dimensions == areas.size());
   if (_scaling == SCALING_DIVIDE_BY_AREA) {
     for (int i = 0; i < areas.size(); i++) {
-      for (int dim = 0; dim < dimensions; dim++) {
-        int valueIndex = i * dimensions + dim;
+      for (int dim = 0; dim < valueDimensions; dim++) {
+        int valueIndex = i * valueDimensions + dim;
         targetValues[valueIndex] /= areas[i];
       }
     }
   } else if (_scaling == SCALING_MULTIPLY_BY_AREA) {
     for (int i = 0; i < areas.size(); i++) {
-      for (int dim = 0; dim < dimensions; dim++) {
-        int valueIndex = i * dimensions + dim;
+      for (int dim = 0; dim < valueDimensions; dim++) {
+        int valueIndex = i * valueDimensions + dim;
         targetValues[valueIndex] *= areas[i];
       }
     }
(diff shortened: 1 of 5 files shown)


uekerman (changes requested):
Looks good. Thanks for the contribution.

These checks also need to be removed to make it work.

https://github.com/precice/precice/blob/c186fd0ec8e57337c0ecaebb78efd55ea266fef3/src/action/config/ActionConfiguration.cpp#L244-L256

And I think it's worth a changelog entry.
"Added support for ..."

mhmdkanj:
@uekerman Thanks for your feedback!
I've already integrated those suggestions now.

uekerman (approved):

DavidSCN:
Do we catch (via error messages) the case where someone tries to use the feature without defining connectivity in 3D?

mhmdkanj:
> Do we catch (via error messages) the case where someone tries to use the feature without defining connectivity in 3D?

@DavidSCN Currently, calling this function/action assumes already that some connectivity is defined (creation of edges in 2D OR creation of triangles in 3D). So if such connectivity is NOT defined, then no error messages are thrown during the process.

Throwing an exception or doing checks in the action config in this case would be beneficial in the sense that calling the scaling action would prevent the nullification (in case of multiplication) or division by zero (in case of division) of the target values. Would that be a possible scenario when using this action while not defining connected vertices in 2D nor 3D?

DavidSCN:
> > Do we catch (via error messages) the case where someone tries to use the feature without defining connectivity in 3D?
> 
> @DavidSCN Currently, calling this function/action assumes already that some connectivity is defined (creation of edges in 2D OR creation of triangles in 3D). So if such connectivity is NOT defined, then no error messages are thrown during the process.
> 
> Throwing an exception or doing checks in the action config in this case would be beneficial in the sense that calling the scaling action would prevent the nullification (in case of multiplication) or division by zero (in case of division) of the target values. Would that be a possible scenario when using this action while not defining connected vertices in 2D nor 3D?

Yes, I assume that this might be a frequent issue since not even all of our official adapter support setting connectivity. If you want to implement this you could use something like `PRECICE_CHECK( cond, message)`, where the condition checks for the existence of the connectivity. If you don't want to implement this, please open up an issue so that we keep it in mind.

uekerman:
> Throwing an exception or doing checks in the action config

In the configuation of the action, we cannot already check for this since at this point meshes are not filled yet. Such a check would need to go into `ScaleByAreaAction::performAction`.

@mhmdkanj We can also fix this in a separate PR. If it's easy for you, you can do it here. If you are puzzled what to do, let's merge here and open a separate issue and/or PR.

mhmdkanj:
> In the configuation of the action, we cannot already check for this since at this point meshes are not filled yet. Such a check would need to go into `ScaleByAreaAction::performAction`.
> 
> @mhmdkanj We can also fix this in a separate PR. If it's easy for you, you can do it here. If you are puzzled what to do, let's merge here and open a separate issue and/or PR.

@uekerman Yup, I just realized that when executing the tests. Please check the second commit if these checks are appropriate. Thanks in advance!

src/action/ScaleByAreaAction.cpp
```diff
@@ -31,26 +31,39 @@ void ScaleByAreaAction::performAction(
...
-    areas[edge.vertex(1).getID()] += edge.getEnclosingRadius();
+  const int       meshDimensions  = getMesh()->getDimensions();
+  auto &          targetValues    = _targetData->values();
+  const int       valueDimensions = _targetData->getDimensions();
+  Eigen::VectorXd areas           = Eigen::VectorXd::Zero(getMesh()->vertices().size());
+  PRECICE_ASSERT(targetValues.size() / valueDimensions == areas.size());
+
+  if (meshDimensions == 2) {
+    PRECICE_CHECK(getMesh()->edges().size() != 0,
+                  "The scaling action expects a connected 2D mesh. Please ensure that the mesh contains at least one \"edge\".");
```
uekerman, line 42:
```suggestion
                  "The multiply/divide-by-area actions require meshes with connectivity information. In 2D, please ensure that the mesh {} contains edges.",  getMesh()->getName());
```

src/action/ScaleByAreaAction.cpp
```diff
@@ -31,26 +31,39 @@ void ScaleByAreaAction::performAction(
...
+    for (mesh::Edge &edge : getMesh()->edges()) {
+      areas[edge.vertex(0).getID()] += edge.getEnclosingRadius();
+      areas[edge.vertex(1).getID()] += edge.getEnclosingRadius();
+    }
+  } else {
+    PRECICE_CHECK(getMesh()->triangles().size() != 0,
+                  "The scaling action expects a connected 3D mesh. Please ensure that the mesh contains at least one \"triangle\".");
```
uekerman, line 49:
```suggestion
                   "The multiply/divide-by-area actions require meshes with connectivity information. In 3D, please ensure that the mesh {} contains triangles.",  getMesh()->getName());
```

uekerman (approved):
Minor tweaks to the text. Otherwise looks good :+1:

DavidSCN (approved):

uekerman:
@mhmdkanj Thanks for the contribution again and congrats on the first merge :tada: :grinning: 

Small technical remark: force pushes are sometimes good, but not always necessary. For such PRs, we often anyway squash & merge, so it's no problem if there are many commits. And for reviewing it's sometimes easier to also see the delta in between review cycles. Otherwise, great work.

PR #1074, opened 2021-09-03
Explicitly require python3 for config formatting

DavidSCN:
## Main changes of this PR
Sets the environment for our config formatter to `python3`

My system uses python2 by default, which leads to the following error message: 
```
  File "~/precice/tools/formatting/config-formatter", line 48
    def print(self, text=''):
            ^
SyntaxError: invalid syntax
```

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/tools/formatting/config-formatter
+++ b/tools/formatting/config-formatter
@@ -1,4 +1,4 @@
-#! /usr/bin/env python
+#! /usr/bin/env python3
 
 from lxml import etree
 import itertools


fsimonis (changes requested):
I did some digging on this some time ago.

There is __no reliable way__ of requesting a specific version of python for your script.

Some distros use python2 and python3, some use python for their main version, some use python for python 3 exclusively. Some don't provide python2 nor python3. Python version switching tools provide python and environments like venv provide python and pythonN.

Python 2 having reached EOL simplified things a bit. The consensus for user scripts is now is to use `python`. This makes sense as python development of bigger projects rely on venv anyhow.

If you are using Ubuntu 20.04 ,then you can install the [`python-is-python3` package](https://launchpad.net/ubuntu/focal/+package/python-is-python3) to change `python` in the user-space. You can also run it with `python3`

DavidSCN:
I'm aware of the issue, but I don't see a drawback of just asking for python3. python 2 is outdated anyway, I have other applications relying explicitly on python-is-python2 though.

fsimonis:
I vaguely recall that there were some issue with the CI and on my workstation, so I had a look at them and python3 seems to be present everywhere.
I'll merge this, if other problems arise, we can fix them on develop.

DavidSCN:
Thanks!

PR #1077, opened 2021-09-13
Build shared libraries by default

### LINKED ISSUE #801: Build is unsupported by default

fsimonis:
## Main changes of this PR

This PR changes the default build type of the preCICE library to `shared`.

Closes #801

Requires changes of the documentation.

As discussed in #801, our current build configuration is not supported by default.
Having to explicitly change the build type to shared is tedious and unnecessary.
Do to our push towards minimal symbol visibility, using shared libraries should become the way to go in the future.

Supporting static libraries via CMake is still a long term goal.

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,7 +58,7 @@ option(PRECICE_PETScMapping "Enable use of the PETSc linear algebra library." ON
 option(PRECICE_PythonActions "Python support" ON)
 option(PRECICE_Packages "Configure package generation." ON)
 option(PRECICE_InstallTest "Add test binary and necessary files to install target." OFF)
-option(BUILD_SHARED_LIBS "Build shared libraries by default" ON)
+option(BUILD_SHARED_LIBS "Build shared instead of static libraries" ON)
 option(BUILD_TESTING "Build tests" ON)
 option(PRECICE_ALWAYS_VALIDATE_LIBS "Validate libraries even after the validatation succeeded." OFF)
 option(PRECICE_ENABLE_C "Enable the native C bindings" ON)
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -58,7 +58,7 @@ option(PRECICE_PETScMapping "Enable use of the PETSc linear algebra library." ON
 option(PRECICE_PythonActions "Python support" ON)
 option(PRECICE_Packages "Configure package generation." ON)
 option(PRECICE_InstallTest "Add test binary and necessary files to install target." OFF)
-option(BUILD_SHARED_LIBS "Build shared libraries by default" OFF)
+option(BUILD_SHARED_LIBS "Build shared libraries by default" ON)
 option(BUILD_TESTING "Build tests" ON)
 option(PRECICE_ALWAYS_VALIDATE_LIBS "Validate libraries even after the validatation succeeded." OFF)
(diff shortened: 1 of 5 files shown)


CMakeLists.txt
```diff
@@ -58,7 +58,7 @@ option(PRECICE_PETScMapping "Enable use of the PETSc linear algebra library." ON
 option(PRECICE_PythonActions "Python support" ON)
 option(PRECICE_Packages "Configure package generation." ON)
 option(PRECICE_InstallTest "Add test binary and necessary files to install target." OFF)
-option(BUILD_SHARED_LIBS "Build shared libraries by default" OFF)
+option(BUILD_SHARED_LIBS "Build shared libraries by default" ON)
```
DavidSCN, line 61: I like the change. However, the `CMake` description is a bit weird as the last argument denotes just the default argument. Launching the cmake-gui renders just `Build shared libraries by default` as argument description, which is confusing. Maybe something like `Build preCICE as a shared library (default is ON)` would fit better if you want to inform the user about the defualt configuration.

fsimonis: I agree that this looks a bit odd.

What about simply `Build shared instead of static libraries`.
The default is what the user sees first, so it shouldn't be necessary to explicitly print it. This would also make it inconsistent with the other options.

DavidSCN: Sounds good :+1:

MakisH (approved):
This will definitely and finally allow people to just run `cmake .. && make`, without having to first read the documentation. So, this is a huge improvement!

I do not consider this breaking, as the default is anyway not working at the moment. Moreover, all scripts that currently explicitly enable shared objects, will keep working.

We still need to update the documentation on the website.

uekerman:
> We still need to update the documentation on the website.

Is somebody already looking into this? Is there an issue?

MakisH:
> > We still need to update the documentation on the website.
> 
> Is somebody already looking into this? Is there an issue?

I can do it. We have an issue now: https://github.com/precice/precice.github.io/issues/83

PR #1078, opened 2021-09-13
Fix DataName access via DataIDs

### LINKED ISSUE #989: Writing to a data field without a 'write-data' tag is not Asserted

DavidSCN:
## Main changes of this PR
Fixes #989, as already discussed in #1010 and https://github.com/precice/precice/issues/989#issuecomment-839653687

Have a look at the issues mentioned above. The main problem after refactoring this was the data name access and the erroneous `if` statement

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/precice/impl/Participant.cpp
+++ b/src/precice/impl/Participant.cpp
@@ -225,7 +225,7 @@ int Participant::getUsedDataID(const std::string &dataName, MeshID meshID) const
 std::string Participant::getDataName(DataID dataID) const
 {
   for (const MeshContext *mcptr : _meshContexts) {
-    if (mcptr) {
+    if (!mcptr) {
       continue;
     }
     for (const auto &dptr : mcptr->mesh->data()) {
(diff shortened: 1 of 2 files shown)


fsimonis:
Thank you for fixing my typos :sweat_smile:

PR #1080, opened 2021-09-13
Add Access to Mesh Requirements via API

### LINKED ISSUE #467: Add access to MeshRequirement to SolverInterface
## Problem
 
preCICE does not always require connectivity information for meshes.
In case a configuration uses only nearest-neighbor/RBF mappings for a given mesh, then it does not require connectivity information and will ignore calls to `setMeshEdge` and `setMeshTriangle`.
This reduces bookkeeping and memory consumption inside the preCICE library.

The adapter using preCICE does not know about this information and always has to generate the connectivity information and call `setMeshXXX`.

## Proposition

We should consider adding a function to the interface that allows to fetch this information.

This allows the adapter to conditionally skip the unnecessary generation of this information.

Example:
* `bool getMeshRequireEdges(int meshID) const`
* `bool getMeshRequireTriangles(int meshID) const`

fsimonis:
## Main changes of this PR

Adds access to the mesh requirements to the SolverInterface.

```cpp
bool SolverInterface::isMeshConnectivityRequired(int meshID) const;
```

Closes #467

Generating connectivity information and passing it to preCICE can be expensive, but preCICE does not always need full mesh connectivity for every mesh.
preCICE ignores API calls to define connectivity if this is not required, but there is no way for the solver to know this.
Thus, solvers/adapters have to be manually configured to skip the connectivity generation.

This PR proposes an API which allows a solver to get information for each mesh directly from preCICE.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/extras/bindings/c/src/SolverInterfaceC.cpp
+++ b/extras/bindings/c/src/SolverInterfaceC.cpp
@@ -189,6 +189,12 @@ int precicec_getDataID(const char *dataName, int meshID)
   return impl->getDataID(stringDataName, meshID);
 }
 
+int precicec_isMeshConnectivityRequired(int meshID)
+{
+  PRECICE_CHECK(impl != nullptr, errormsg);
+  return impl->isMeshConnectivityRequired(meshID);
+}
+
 int precicec_setMeshVertex(
     int           meshID,
     const double *position)
--- a/extras/bindings/c/src/SolverInterfaceC.cpp
+++ b/extras/bindings/c/src/SolverInterfaceC.cpp
@@ -192,7 +192,10 @@ int precicec_getDataID(const char *dataName, int meshID)
 int precicec_isMeshConnectivityRequired(int meshID)
 {
   PRECICE_CHECK(impl != nullptr, errormsg);
-  return impl->isMeshConnectivityRequired(meshID);
+  if (impl->isMeshConnectivityRequired(meshID)) {
+    return 1;
+  }
+  return 0;  
 }
 
 int precicec_setMeshVertex(
--- a/extras/bindings/c/src/SolverInterfaceC.cpp
+++ b/extras/bindings/c/src/SolverInterfaceC.cpp
@@ -195,7 +195,7 @@ int precicec_isMeshConnectivityRequired(int meshID)
   if (impl->isMeshConnectivityRequired(meshID)) {
     return 1;
   }
-  return 0;  
+  return 0;
 }
 
 int precicec_setMeshVertex(
--- a/extras/bindings/fortran/include/precice/SolverInterfaceFortran.hpp
+++ b/extras/bindings/fortran/include/precice/SolverInterfaceFortran.hpp
@@ -345,6 +345,21 @@ void precicef_get_data_id_(
 
 /**
  * Fortran syntax:
+ * precicef_has_data(
+ *   INTEGER   meshID
+ *   INTEGER   required)
+ *
+ * IN:  meshID
+ * OUT: required(1:true, 0:false)
+ *
+ * @copydoc precice::SolverInterface::isMeshConnectivityRequired()
+ */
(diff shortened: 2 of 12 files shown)


extras/bindings/fortran/include/precice/SolverInterfaceFortran.hpp
```diff
@@ -343,6 +343,21 @@ void precicef_get_data_id_(
...
+ * precicef_has_data(
+ *   INTEGER   meshID
+ *   INTEGER   required)
+ *
+ * IN:  meshID
+ * OUT: required(1:true, 0:false)
+ *
+ * @copydoc precice::SolverInterface::isMeshConnectivityRequired()
+ */
+void precicef_isMeshConnectivityRequired(
```
uekerman, line 357:
```suggestion
void precicef_is_mesh_connectivity_required_(
```

extras/bindings/fortran/src/SolverInterfaceFortran.cpp
```diff
@@ -274,6 +274,14 @@ void precicef_get_data_id_(
   *dataID = impl->getDataID(stringDataName, *meshID);
 }
 
+void precicef_isMeshConnectivityRequired(
```
uekerman, line 277:
```suggestion
void precicef_is_mesh_connectivity_required_(
```

extras/bindings/fortran/src/SolverInterfaceFortran.cpp
```diff
@@ -274,6 +274,14 @@ void precicef_get_data_id_(
...
+    const int *meshID,
+    int *      required)
+{
+  PRECICE_CHECK(impl != nullptr, errormsg);
+  *required = impl->isMeshConnectivityRequired(*meshID);
```
uekerman, line 282:
```suggestion
  if (impl->isMeshConnectivityRequired(*meshID)) {
    *required = 1;
  } else {
    *required = 0;
  }
```
At least, this is how we do it for similar API functions.

extras/bindings/c/src/SolverInterfaceC.cpp
```diff
@@ -189,6 +189,12 @@ int precicec_getDataID(const char *dataName, int meshID)
   return impl->getDataID(stringDataName, meshID);
 }
 
+int precicec_isMeshConnectivityRequired(int meshID)
+{
+  PRECICE_CHECK(impl != nullptr, errormsg);
+  return impl->isMeshConnectivityRequired(meshID);
```
uekerman, line 195:
```suggestion
  if (impl->isMeshConnectivityRequired(meshID)) {
    return 1;
  }
  return 0;  
```
At least, this is how we do it for similar API functions.

src/precice/SolverInterface.hpp
```diff
@@ -349,6 +349,18 @@ class SolverInterface {
    */
   [[deprecated("Use getMeshID() for specific mesh names instead.")]] std::set<int> getMeshIDs() const;
 
+  /**
+   * @brief Checks if the given mesh requires connectivity.
+   *
+   * preCICE may require connectivity information from the solver and will
+   * ignore any API calls regarding connectivity if it is not required.
```
uekerman, line 356:
```suggestion
   * preCICE may require connectivity information from the solver and
   * ignores any API calls regarding connectivity if it is not required.
```

uekerman (changes requested):
Implementation and tests look good :+1:
Only a few doubt about the C and Fortran bindings. It's really problematic that we don't have tests here (preSERVE :cry: ) ... what you could always do is to quickly modify a solverdummy and test "offline".

We should not forget to extend the documentation here: https://precice.org/couple-your-code-defining-mesh-connectivity.html

fsimonis:
I manually tested this using the Fortran and C solverdummies and it works fine.

uekerman (approved):

PR #1083, opened 2021-09-14
Track number of valid samples inside Waveform and simplify interface.

BenjaminRodenberg:
## Main changes of this PR

Tracks number of valid samples inside Waveform and allows to simplify corresponding interface.

This is only a small simplification which I factored out from #1029 to merge directly. This avoids ambiguity with respect to what "timeWindows" means and makes `Waveform` easier to use.

A newly constructed waveform now has only a single valid sample by default. New valid samples can be added to the waveform by moving to the next window. This will archive the current sample in `_timeWindows` and allows the user to add a new sample via `store()`.

## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -251,7 +251,7 @@ void BaseCouplingScheme::moveToNextWindow()
   PRECICE_TRACE(_timeWindows);
   for (DataMap::value_type &pair : _allData) {
     PRECICE_DEBUG("Store data: {}", pair.first);
-    _waveforms[pair.first]->moveToNextWindow(getTimeWindows(), _extrapolationOrder);
+    _waveforms[pair.first]->moveToNextWindow(_extrapolationOrder);
     pair.second->values() = _waveforms[pair.first]->lastTimeWindows().col(0);
   }
 }
--- a/src/time/Waveform.cpp
+++ b/src/time/Waveform.cpp
@@ -19,6 +19,7 @@ Waveform::Waveform(
      */
   int initializedNumberOfSamples = std::max({2, extrapolationOrder + 1});
   _timeWindows                   = Eigen::MatrixXd::Zero(initializedNumberOfData, initializedNumberOfSamples);
+  _numberOfValidSamples          = 1; // we assume that upon creation the first sample is always valid.
   PRECICE_ASSERT(numberOfSamples() == initializedNumberOfSamples);
   PRECICE_ASSERT(numberOfData() == initializedNumberOfData);
 }
@@ -31,15 +32,13 @@ void Waveform::store(const Eigen::VectorXd &data)
   this->_timeWindows.col(columnID) = data;
 }
 
-void Waveform::moveToNextWindow(int timeWindows, int order)
+void Waveform::moveToNextWindow(int order)
 {
-  auto initialGuess = extrapolateData(order, timeWindows);
+  if (_numberOfValidSamples < numberOfSamples()) {
+    _numberOfValidSamples++;
+  }
+  auto initialGuess = extrapolateData(order);
   utils::shiftSetFirst(this->_timeWindows, initialGuess);
-}
-
-const Eigen::MatrixXd &Waveform::lastTimeWindows()
-{
-  return _timeWindows;
 }
 
 int Waveform::numberOfSamples()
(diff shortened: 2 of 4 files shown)


src/time/Waveform.cpp
```diff
@@ -31,34 +32,42 @@ void Waveform::store(const Eigen::VectorXd &data)
...
 }
 
-void Waveform::moveToNextWindow(int timeWindows, int order)
+void Waveform::moveToNextWindow(int order)
 {
-  auto initialGuess = extrapolateData(order, timeWindows);
+  if (_numberOfValidSamples < numberOfSamples()) {
+    _numberOfValidSamples++;
+  }
+  auto initialGuess = extrapolateData(order);
```
BenjaminRodenberg, line 40:
```suggestion
  auto initialGuess = extrapolateData(order);
  if (_numberOfValidSamples < numberOfSamples()) {
    _numberOfValidSamples++;
  }
```
We first have to compute the initial guess and then increase `_numberOfValidSamples`. Otherwise an unitialized sample might be used for the computation of the initial guess.

docs/changelog/1083.md
```diff
@@ -0,0 +1 @@
+- Simplify interface of `time::Waveform` by storing and updating number of past time windows (and therefore valid samples) inside waveform. A newly created waveform only has a single valid sample, which is initialized as zero. `Waveform::store()` is used to write new data to the waveform. Calling `Waveform::moveToNextWindow()` will increase the counter for valid samples and archive old samples `_timeWindows`.
```
uekerman, line 1:
```suggestion
- Simplified interface of `time::Waveform` by storing and updating number of past time windows (and therefore valid samples) inside waveform. A newly created waveform only has a single valid sample, which is initialized as zero. `Waveform::store()` is used to write new data to the waveform. Calling `Waveform::moveToNextWindow()` increases the counter for valid samples and archive old samples `_timeWindows`.
```

src/time/Waveform.hpp
```diff
@@ -39,6 +38,11 @@ class Waveform {
    */
   int numberOfSamples();
 
+  /**
+   * @brief returns number of valid samples in time stored by this waveform
+   */
+  int numberOfValidSamples();
```
uekerman, line 44: Why is this function needed? Currently, it's only used in tests if I checked correctly.

src/time/Waveform.hpp
```diff
@@ -48,6 +52,9 @@ class Waveform {
   /// Data values of time windows.
   Eigen::MatrixXd _timeWindows;
 
+  /// number of valid samples in _timeWindows
+  int _numberOfValidSamples;
```
uekerman, line 56: I guess we could find a better name here. The variable is only needed for extrapolation, right? Or do you see a use case for real waveform stuff as well? If not, we could use sth like `maxPossibleExtrapolationOrder`?

uekerman, line 56: I guess we could find a better name here. The variable is only needed for extrapolation, right? Or do you see a use case for real waveform stuff as well? If not, we could use sth like `maxPossibleExtrapolationOrder`?

uekerman (changes requested):
Yes, simplifies the interface nicely :+1:
A few minor doubts below.

src/time/Waveform.hpp
BenjaminRodenberg: You are right. We can again hide this function and make the class a friend or use the correct fixture. I'll take care of this.

src/time/Waveform.hpp
BenjaminRodenberg: `MaxPossibleExtrapolationOrder` is not ideal, because this will also be relevant for interpolation later. What we are doing is counting the number of samples that have been "locked" by the coupling scheme. This means: As soon as a time window is complete the coupling scheme will trigger the waveform to store the latest sample. This will then copy the sample to a place in `_timeWindows` where it will not change anymore. Therefore, one more sample is available for interpolation/extrapolation, when we enter the next window.

BenjaminRodenberg: This is a duplicate

BenjaminRodenberg:
There are still some tests to update here. I also stumbled over #1089 while doing this. I will first try to solve #1089 and then continue here.

src/time/Waveform.hpp
uekerman: ok, I see. But how is this then different from "numberOfTimeWindows"?

BenjaminRodenberg: What do you mean? Can you give a link? In `Waveform.{cpp,hpp}` I cannot find this variable.

BenjaminRodenberg:
> There are still some tests to update here. I also stumbled over #1089 while doing this. I will first try to solve #1089 and then continue here.

#1089 seems to be more complicated. I will keep this PR non-breaking. Meaning: Sticking to the existing formula and not touch the formula for second order extrapolation.

src/time/Waveform.hpp
BenjaminRodenberg: I took care of this. Let's first merge https://github.com/precice/precice/pull/1094. Then this PR becomes a bit smaller, since the fixture is already there.

src/time/Waveform.cpp
BenjaminRodenberg: I tried out this buggy implementation on the basis of de659d928083fbde0b96b440ff67e0949264742c and the tests are able to detect this. So there is no need for an additional test.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -154,10 +154,19 @@ void BaseCouplingScheme::initialize(double startTime, int startTimeWindow)
...
 
+  if (not _sendsInitializedData && not _receivesInitializedData) {
+    if (isImplicitCouplingScheme()) {
+      if (not doesFirstStep()) {
+        storeDataInWaveforms();
+        moveToNextWindow();
+      }
+    }
+  }
+
```
BenjaminRodenberg, line 169: I think this is an interesting change: Since the state is now tracked internally in the Waveform, we also have to tell the Waveform "initialization is done and next thing is advance" before we enter advance the first time. 

* If we call `initializeData()` this is already done, since otherwise the initial data will be lost (see https://github.com/precice/precice/pull/1098/files#diff-ad728b4925806f3679256c80a1006598dd5e4f4ec5c22fe90121729abad234c2R175-R181).
* If we don't call `initializeData()`, this part is missing and, therefore, we have to take care of this in `initialize()` (the last thing happening before calling `advance()`.

It might be a good idea to factor this out into a new function `preAdvance()` that may only be called once and which must be called directly before `advance()`. Any opinion on this @uekerman?

uekerman: `preAdvance()` sounds complicated :see_no_evil: 
In principle, we know what happens during configuration already. You could there pass this information to `time` already.

BenjaminRodenberg: Problem I see during configuration: If there is data, we don't know it yet. Additionally, we also do not know the mesh / number of unknowns yet.

PR #1084, opened 2021-09-14
Clean-up direct access functionality

DavidSCN:
## Main changes of this PR
@IshaanDesai could you please test if your erroneous setup fails with this PR?
Related to #1033 
Closes https://github.com/precice/precice/issues/1085
## Author's checklist

## Reviewers' checklist

* [ ] (more questions/tasks)

### DIFF
--- a/src/precice/SolverInterface.hpp
+++ b/src/precice/SolverInterface.hpp
@@ -851,6 +851,11 @@ public:
    *             corresponding data values (dim * \p size)
    *
    * @pre IDs and coordinates need to have the correct size, which can be queried by @p getMeshVertexSize()
+   *
+   * @note This function can be called on received meshes as well as provided
+   * meshes. However, you need to call this function after @p initialize(),
+   * if the meshID corresponds to a received mesh, since the relevant mesh data
+   * is exchanged during the @p initialize() call.
    */
   void getMeshVerticesAndIDs(
       const int meshID,
--- a/src/precice/SolverInterface.hpp
+++ b/src/precice/SolverInterface.hpp
@@ -344,7 +344,7 @@ public:
    *
    * @deprecated Unclear use case and difficult to port to other languages.
    *             Prefer calling getMeshID for specific mesh names.
-   * 
+   *
    * @returns the set of ids.
    */
   [[deprecated("Use getMeshID() for specific mesh names instead.")]] std::set<int> getMeshIDs() const;
@@ -370,6 +370,11 @@ public:
    *
    * @param[in] meshID the id of the mesh
    * @returns the amount of the vertices of the mesh
+   *
+   * @pre This function can be called on received meshes as well as provided
+   * meshes. However, you need to call this function after @p initialize(),
+   * if the \p meshID corresponds to a received mesh, since the relevant mesh data
+   * is exchanged during the @p initialize() call.
    */
   int getMeshVertexSize(int meshID) const;
 
@@ -500,9 +505,9 @@ public:
    * @param[in] secondEdgeID ID of the second edge of the Quad
    * @param[in] thirdEdgeID ID of the third edge of the Quad
    * @param[in] fourthEdgeID ID of the forth edge of the Quad
-   * 
+   *
    * @pre edges with firstEdgeID, secondEdgeID, thirdEdgeID and fourthEdgeID were added to the mesh with the ID meshID.
-   * 
+   *
    */
   void setMeshQuad(
       int meshID,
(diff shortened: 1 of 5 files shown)


IshaanDesai:
I verified the new error check using [this example](https://github.com/IshaanDesai/coupled-heat-conduction/blob/8bdfb55996a664b2d91578134754a7fc1b1c402d/micro-manager.py) and when I try to get vertex ids and coordinates before `initialize()` I see the correct error message:

```bash
ERROR: initialize() has to be called before accessing  data of the received mesh "macro-mesh" on 
participant "Micro-manager".
```

IshaanDesai:
This also closes https://github.com/precice/precice/issues/1085

src/precice/SolverInterface.hpp
```diff
@@ -851,6 +851,11 @@ class SolverInterface {
    *             corresponding data values (dim * \p size)
    *
    * @pre IDs and coordinates need to have the correct size, which can be queried by @p getMeshVertexSize()
+   *
+   * @note This function can be called on received meshes as well as provided
+   * meshes. However, you need to call this function after @p initialize(),
+   * if the meshID corresponds to a received mesh, since the relevant mesh data
+   * is exchanged during the @p initialize() call.
```
fsimonis, line 858: You can simply state this as a precondition

```
 * @pre initialize() has been called
```

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -515,7 +515,7 @@ void ParticipantConfiguration::finishParticipantConfiguration(
 
   for (impl::DataContext &dataContext : participant->readDataContexts()) {
     int toMeshID = dataContext.getMeshID();
-    PRECICE_CHECK(participant->isMeshProvided(toMeshID),
+    PRECICE_CHECK(participant->isMeshProvided(toMeshID) || participant->isDirectAccessAllowed(toMeshID),
```
fsimonis, line 518: In which case is a mesh provided but cannot be accessed directly by the same participant?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -627,6 +627,12 @@ int SolverInterfaceImpl::getMeshVertexSize(
 {
   PRECICE_TRACE(meshID);
   PRECICE_REQUIRE_MESH_USE(meshID);
+  // In case we access received mesh data: check, if the requested mesh data has already been received.
+  // Otherwise, the function call doesn't make any sense
+  PRECICE_CHECK((_state == State::Initialized) || _accessor->isMeshProvided(meshID), "initialize() has to be called before accessing"
```
fsimonis, line 632: We should add this precondition to the doc of `getMeshVertexSize`.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1325,6 +1331,12 @@ void SolverInterfaceImpl::getMeshVerticesAndIDs(
   PRECICE_TRACE(meshID, size);
   PRECICE_REQUIRE_MESH_USE(meshID);
   PRECICE_DEBUG("Get {} mesh vertices with IDs", size);
+
+  // Check, if the requested mesh data has already been received. Otherwise, the function call doesn't make any sense
+  PRECICE_CHECK((_state == State::Initialized) || _accessor->isMeshProvided(meshID), "initialize() has to be called before accessing"
```
fsimonis, line 1336: We should add this precondition to the doc of `getMeshVerticesAndIDs`.

fsimonis (commented):
Did you forget to commit the test that belongs to the xml?

src/precice/impl/SolverInterfaceImpl.cpp
DavidSCN: Good point

src/precice/SolverInterface.hpp
DavidSCN: Yes, I guess it needs some more explanation. I like the more verbose variant. I changed the note to an `@pre` though.

src/precice/config/ParticipantConfiguration.cpp
DavidSCN: In case it was not configured. We require an explicit configuration of this feature at the moment. Otherwise people most probably do things the wrong way.

DavidSCN:
Ready to merge and or review.

src/precice/config/ParticipantConfiguration.cpp
fsimonis: Thanks, got it now. This is required by #1033.

Issue #976, opened 2021-03-11
Added zero-vector throws error in very last coupling iteration


ajaust:
**Describe your setup**

Operating system (e.g. Linux distribution and version): _Ubuntu 18.04_
preCICE Version: 2.2.0

**Describe the problem**

I have a test case with very tight convergence measure and very small values (0 up to 1e-12) that are exchanged. These small values are not uncommon for velocities in porous media. The difference matrix is updated with a (almost) zero vector which lets preCICE issue error that crashes the simulation. However, the convergence measures are fulfilled and this is the only coupling time-windows (steady-state simulation). However, I assume that this applies to the last time window in general. 

```
(0) 17:05:57 [cplscheme::BaseCouplingScheme]:559 in measureConvergence: relative convergence measure: relative two-norm diff = 6.7474e-08, limit = 1e-06, normalization = 3.65118e-09, conv = true
(0) 17:05:57 [cplscheme::BaseCouplingScheme]:559 in measureConvergence: relative convergence measure: relative two-norm diff = 3.64636e-07, limit = 1e-06, normalization = 6.19158e-12, conv = true
(0) 17:05:57 [cplscheme::BaseCouplingScheme]:563 in measureConvergence: All converged
(0) 17:05:57 [acceleration::BaseQNAcceleration]:201 in updateDifferenceMatrices: WARNING: The coupling residual equals almost zero. There is maybe something wrong in your adapter. Maybe you always write the same data or you call advance without providing new data first or you do not use available read data. Or you just converge much further than actually necessary.
(0) 17:05:57 [acceleration::BaseQNAcceleration]:230 in updateDifferenceMatrices: ERROR: Attempting to add a zero vector to the quasi-Newton V matrix. This means that the residual in two consecutive iterations is identical. There is probably something wrong in your adapter. Maybe you always write the same (or only incremented) data or you call advance without providing  new data first.
```

The overall simulation hangs since the other solver is not informed about this error.

**Step To Reproduce**

1. Set up a simulation with tight convergence measure such that you get (nearly) zero difference vectors
2. Taylor your convergence measure such that it is fullfilled when the simulation aborts due to zero-difference vectors

Example configuration: [precice-config-parallel-implicit.xml.txt](https://github.com/precice/precice/files/6124775/precice-config-parallel-implicit.xml.txt)

**Expected behaviour**

[...]

PR #1092, opened 2021-09-15
Updates difference matrices if residuals are not zero only

KyleDavisSA:
## Main changes of this PR

When updating the difference matrices, if the residuals are zero an error is thrown. This PR only updates the difference matrices if the residual is non-zero, otherwise only a warning is thrown.

Closes https://github.com/precice/precice/issues/976

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/acceleration/BaseQNAcceleration.cpp
+++ b/src/acceleration/BaseQNAcceleration.cpp
@@ -228,43 +228,45 @@ void BaseQNAcceleration::updateDifferenceMatrices(
       Eigen::VectorXd deltaXTilde = _values;
       deltaXTilde -= _oldXTilde;
 
-      PRECICE_CHECK(not math::equals(utils::MasterSlave::l2norm(deltaR), 0.0),
-                    "Attempting to add a zero vector to the quasi-Newton V matrix. This means that the residual "
-                    "in two consecutive iterations is identical. There is probably something wrong in your adapter. "
-                    "Maybe you always write the same (or only incremented) data or you call advance without "
-                    "providing  new data first.");
+      if (not math::equals(utils::MasterSlave::l2norm(deltaR), 0.0)) {
+        
+        bool columnLimitReached = getLSSystemCols() == _maxIterationsUsed;
+        bool overdetermined     = getLSSystemCols() <= getLSSystemRows();
+        if (not columnLimitReached && overdetermined) {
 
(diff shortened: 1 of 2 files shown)


src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -228,11 +228,17 @@ void BaseQNAcceleration::updateDifferenceMatrices(
...
       deltaXTilde -= _oldXTilde;
 
-      PRECICE_CHECK(not math::equals(utils::MasterSlave::l2norm(deltaR), 0.0),
-                    "Attempting to add a zero vector to the quasi-Newton V matrix. This means that the residual "
-                    "in two consecutive iterations is identical. There is probably something wrong in your adapter. "
-                    "Maybe you always write the same (or only incremented) data or you call advance without "
-                    "providing  new data first.");
+      double residualMagnitude = utils::MasterSlave::l2norm(deltaR);
+
+      if (utils::MasterSlave::l2norm(_values) < 1e-14) {
```
uekerman, line 233:
```suggestion
      if (not math::equals(utils::MasterSlave::l2norm(_values), 0.0)) {
```

uekerman (changes requested):

BenjaminRodenberg:
I'm not really aware of all the implications of this PR, but as far as I understand normalization only makes sense, if `utils::MasterSlave::l2norm(_values) < 1`. The scaling should only be applied to small values (as described in #976). In https://github.com/precice/precice/pull/1095#issuecomment-934212563 I observed this error as a false positive, because large `_values` (`utils::MasterSlave::l2norm(_values) > 1`) artificially decrease `residualMagnitude` which leads to triggering `PRECICE_CHECK(not math::equals(residualMagnitude, 0.0), ...)`, even though everything is fine.

I added a few print statements in the code:

```
residualMagnitude before: 3.79196e-14
MasterSlave::l2norm(_values): 12.9285
residualMagnitude after: 2.93301e-15
---[precice] ERROR:  Attempting to add a zero vector to the quasi-Newton V matrix. This means that the residual in two consecutive iterations is identical. There is probably something wrong in your adapter. Maybe you always write the same (or only incremented) data or you call advance without providing  new data first.
```

A possible fix is provided here:

https://github.com/BenjaminRodenberg/precice/commit/92898a30aaa1dcee4950a4d556eabb9fde669c12

PR #1094, opened 2021-09-15
Create WaveformFixture and use it in WaveformTest.

BenjaminRodenberg:
## Main changes of this PR

* Create `WaveformFixture.cpp` and use it in `WaveformTest.cpp`.
* Make more functions in `Waveform.hpp` private

Allows to make functions private that are exclusively used in testing.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/testing/WaveformFixture.cpp
+++ b/src/testing/WaveformFixture.cpp
@@ -1,0 +1,16 @@
+#include "testing/WaveformFixture.hpp"
+
+namespace precice {
+namespace testing {
+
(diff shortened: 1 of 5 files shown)


uekerman (approved):
Looks good to me.

Does not directly belong here: I find the name `numberOfData()` very hard to understand. Especially if you think of "data" as sth uncountable like water.

The documentation does not help much:
> returns number of data per sample in time stored by this waveform
   
Maybe "number of data sets"?
What is it actually? :D Shouldn't there be a waveform per data set?

PR #1095, opened 2021-09-15
Release v2.3.0

fsimonis:
## How to work with this template

Release Manager is @fsimonis 

Only the release manager should update this post (even tickboxes, due to race conditions in editing). Everybody else should comment on the PR with the progress.

## Preparation
* Open milestone PRs and [`changelog directory`](https://github.com/precice/precice/blob/develop/docs/changelog/). Then add missing changelogs (all)
* Pull changes from both `develop` and `master`.
* On develop, check code base w.r.t code formatting (run [`precice/tools/formatting/check-format`](https://github.com/precice/precice/blob/develop/tools/formatting/check-format)) and reformat if required (run [`precice/tools/formatting/format-all`](https://github.com/precice/precice/blob/develop/tools/formatting/format-all)). (This should never be the case!)

## Step by step guide

* [x] Merge master to develop (No commits after the release on master)
* [x] Create branch `release-vMAJOR.MINOR.PATCH` from develop. If needed, `git rebase develop`.
* [x] Run `tools/releasing/bumpversion.sh MAJOR.MINOR.PATCH` to bump the version
* [x] Look over [`CHANGELOG.md`](https://github.com/precice/precice/blob/develop/CHANGELOG.md) (all)
   * Add things, if necessary
   * Extract summary
   * Fix wording and tense
   * Sort the entries lexicographically
* [x] Verify the version changes in:
   * [x] [CHANGELOG](https://github.com/precice/precice/blob/develop/CHANGELOG.md)
   * [x] [CMakeLists.txt](https://github.com/precice/precice/blob/develop/CMakeLists.txt)
   * [x] [debian changelog](https://github.com/precice/precice/blob/develop/tools/releasing/packaging/debian/changelog)
* [x] Commit as `Bump version MAJOR.MINOR.PATCH`
* [x] Push the release branch upstream (must be in the precice repository)
* [x] Open PR from `release-N` to `master` (use [this template](https://github.com/precice/precice/blob/develop/tools/releasing/PULL_REQUEST_TEMPLATE/release_pull_request_template.md))
* [x] Do regression tests using the release branch (specific revision) _list below :arrow_down:_ (all)
* [x] Fix potential problems in develop (all)
* Asynchronously
   * [x] (prepare independent release) [Python bindings](https://github.com/precice/python-bindings)
   * [ ] (prepare independent release, if necessary!) [MATLAB bindings](https://github.com/precice/matlab-bindings)
   * [x] Draft message to mailing list
[...]

PR #1096, opened 2021-09-16
Calulix test case regression

BenjaminRodenberg:
## Author's checklist

## Reviewers' checklist


BenjaminRodenberg:
This PR is probably a dead-end, but let's keep it until we have found the bug which stops us from releasing #1095

fsimonis:
Shifting any discussion to #1098 and closing this PR.

BenjaminRodenberg:
@KyleDavisSA Could you [open an issue](https://github.com/precice/precice/issues) for the bug you spotted? Currently this closed PR is the only place, where the bug is implicitly mentioned and documented. Please quickly describe the rough setup. I think this helps us for future reference. Would also be helpful, if you could add your reference results from preCICE v2.2.1 in the issue.

PR #1097, opened 2021-09-16
Fix used order depending on window

BenjaminRodenberg:
## Main changes of this PR

Fixes condition for extrapolation:

* At the end of the first window we always have to use no extrapolation (constant extrapolation), because there is only a single sample available.
* At the end of the second window we are at maximum allowed to use first order extrapolation, because there are only two samples available.
* At the end of the third window we are allowed to do second order extrapolation, because there are three samples available

This is a regression compared to https://github.com/precice/precice/releases/tag/v2.2.1, which can be seen by using extrapolation of different order and comparing the results.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/1097.md
+++ b/docs/changelog/1097.md
@@ -1,0 +1,1 @@
+Fixed used extrapolation order depending on current time window.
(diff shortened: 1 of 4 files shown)


uekerman (approved):
Still needs a changelog entry
"Fixed used extrapolation order depending on current time window."

BenjaminRodenberg:
@fsimonis Feel free to merge this PR after review. Related PR is also #1098, but let's merge this one first and then continue with the next one.

fsimonis:
I'll omit the changelog entry, as this bug was introduced and fixed on develop.

uekerman:
> The last release was fine.

Ah, OK. I thought we had this bug since longer already and it was just refactored on develop.

fsimonis:
Actually, we should probably go further back in time to make sure.

PR #1098, opened 2021-09-16
Move storeIteration() after moveToNextWindow()

BenjaminRodenberg:
## Main changes of this PR

Swap order of `storeIteration()` and `moveToNextWindow()`. 
Also clarify `Waveform::extrapolateData()`.

Extrapolation (inside `moveToNextWindow()`) has to be done before `storeIteration()`, because otherwise the extrapolated value is not available to the following acceleration step (uses `CouplingData::_previousIteration`).

## Author's checklist

* [x] Merge #1097 first.

## Reviewers' checklist

### DIFF
--- a/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp
+++ b/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp
@@ -1593,7 +1593,7 @@ BOOST_AUTO_TEST_CASE(testInitializeData)
     BOOST_TEST(receiveCouplingData->values().size() == 1);
     BOOST_TEST(testing::equals(receiveCouplingData->values()(0), 4.0));
     BOOST_TEST(receiveCouplingData->previousIteration().size() == 1);
-    BOOST_TEST(testing::equals(receiveCouplingData->previousIteration()(0), 0.0));
+    BOOST_TEST(testing::equals(receiveCouplingData->previousIteration()(0), 4.0));
     BOOST_TEST(sendCouplingData->values().size() == 3);
     BOOST_TEST(testing::equals(sendCouplingData->values(), Eigen::Vector3d(1.0, 2.0, 3.0)));
     BOOST_TEST(sendCouplingData->previousIteration().size() == 3);
--- a/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp
+++ b/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp
@@ -745,8 +745,8 @@ BOOST_AUTO_TEST_CASE(testAccelerationWithLinearExtrapolation)
     // first order extrapolation
     BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 4); // = 2*3 - 2
   } else if (context.isNamed(second)) {
-    // extrapolation only applied to accelerated data. So data written by first participant.
-    BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 3);
+    // first order extrapolation
+    BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 5); // = 2*3 - 1
   }
 
   // reached end of simulation, ready to finalize
@@ -941,10 +941,10 @@ BOOST_AUTO_TEST_CASE(testLinearExtrapolationInit)
         BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 0); // = 2*2 - 4
       } else if (i == 1) {
         // accelerated data from second participant
-        BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 3.5); // = 0.5 * 4 + 0.5 * 3
+        BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 1.5); // = 0.5 * 0 + 0.5 * 3
       } else if (i == 2) {
         // accelerated data from second participant
-        BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 3.25); // = 0.5 * 3.5 + 0.5 * 3
+        BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 2.25); // = 0.5 * 1.5 + 0.5 * 3
       }
     } else if (context.isNamed(second)) {
       // extrapolation only applied to accelerated data. So data written by first participant.
@@ -980,8 +980,8 @@ BOOST_AUTO_TEST_CASE(testLinearExtrapolationInit)
     // first order extrapolation
     BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 4); // = 2*3 - 2
   } else if (context.isNamed(second)) {
-    // extrapolation only applied to accelerated data. So data written by first participant.
-    BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 3);
+    // first order extrapolation
+    BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 5); // = 2*3 - 1
   }
 
   // reached end of simulation, ready to finalize
@@ -1230,8 +1230,8 @@ BOOST_AUTO_TEST_CASE(testAccelerationWithQuadraticExtrapolation)
     // second order extrapolation
     BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 7.5); // = 2.5*5 - 2*3 + 0.5*2
   } else if (context.isNamed(second)) {
-    // extrapolation only applied to accelerated data. So data written by first participant.
-    BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 5);
+    // second order extrapolation
+    BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 7); // = 2.5*5 - 2*3 + 0.5*1
   }
 
   // reached end of simulation, ready to finalize
--- a/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp
+++ b/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp
@@ -465,25 +465,38 @@ BOOST_AUTO_TEST_CASE(testExtrapolateData)
   BOOST_TEST(testing::equals(cplData->values()(0), 0.0));
   BOOST_TEST(testing::equals(cplData->previousIteration()(0), 0.0));
 
-  cplData->values()(0) = 1.0;
+  // start first window
+  cplData->values()(0) = 1.0; // data provided at end of first window
   scheme.setTimeWindows(scheme.getTimeWindows() + 1);
   scheme.storeDataInWaveforms();
   BOOST_TEST(testing::equals(cplData->values()(0), 1.0));
-  scheme.moveToNextWindow();
+
+  // go to second window
+  scheme.moveToNextWindow(); // uses zeroth order extrapolation at end of first window
   BOOST_TEST(testing::equals(cplData->previousIteration()(0), 0.0));
-
   scheme.storeIteration();
-  BOOST_TEST(testing::equals(cplData->values()(0), 2.0));
-  BOOST_TEST(testing::equals(cplData->previousIteration()(0), 2.0));
-
-  cplData->values()(0) = 4.0;
+  BOOST_TEST(testing::equals(cplData->values()(0), 1.0));
+  BOOST_TEST(testing::equals(cplData->previousIteration()(0), 1.0));
+  cplData->values()(0) = 4.0; // data provided at end of second window
   scheme.setTimeWindows(scheme.getTimeWindows() + 1);
   scheme.storeDataInWaveforms();
-  scheme.moveToNextWindow();
-  BOOST_TEST(testing::equals(cplData->previousIteration()(0), 2.0));
+
+  // go to third window
+  scheme.moveToNextWindow(); // uses first order extrapolation (maximum allowd) at end of second window
+  BOOST_TEST(testing::equals(cplData->previousIteration()(0), 1.0));
   scheme.storeIteration();
   BOOST_TEST(testing::equals(cplData->values()(0), 7.0));
   BOOST_TEST(testing::equals(cplData->previousIteration()(0), 7.0));
+  cplData->values()(0) = 10.0; // data provided at end of third window
+  scheme.setTimeWindows(scheme.getTimeWindows() + 1);
+  scheme.storeDataInWaveforms();
+
+  // go to fourth window
+  scheme.moveToNextWindow(); // uses first order extrapolation (maximum allowd) at end of third window
+  BOOST_TEST(testing::equals(cplData->previousIteration()(0), 7.0));
+  scheme.storeIteration();
+  BOOST_TEST(testing::equals(cplData->values()(0), 16.0)); // = 2*10 - 4
+  BOOST_TEST(testing::equals(cplData->previousIteration()(0), 16.0));
 
   // Test second order extrapolation
   cplData->values() = Eigen::VectorXd::Zero(cplData->values().size());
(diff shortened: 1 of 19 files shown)


fsimonis:
There are some more differences to master:

Scenario: Serial implicit coupling scheme
Master: applies extrapolation only to send data
Develop: applies extrapolation to all data

Schenario: using 0th order or there being only available 1 sample
Master: stores only the written data in the history (BaseCouplingScheme::storeData, ::extrapolateData does nothing)
Develop: stores the written data (storeIteration) and the extrapolated data (duplicate) in the history (moveToNextWindow)

uekerman:
I can reproduce the extrapolation issue with the Alya Flap.
With this fix, results look nearly good. The only difference is still that I get significantly different results in the convergence file in the first iteration of the second time window.

uekerman:
I also found / fixed another "bug" in https://github.com/precice/precice/pull/1098/commits/8cf8cb807156e8af06976a362eabac582348d2a2
I guess this does not alter information since wrongly extrapolated data (data which is received in a serial coupling scheme) is later overwritten anyway. Still, it makes the code harder to understand if we extrapolate data where extrapolation makes no sense.

fsimonis:
Without extrapolation, I get the same iterations for the elastic-tube-1d tutorial for 8cf8cb8, v2.2.1 and  v2.1.1.
(It crashes with extrapolation enabled, so I cannot test that case)

uekerman:
> The only difference is still that I get significantly different results in the convergence file in the first iteration of the second time window.

I don't get this regression when switching extrapolation off. And maybe to clarify, the first difference appears in the first iteration of the second time window, afterwards everything is different.

uekerman:
I found another (the?) problem.

This here is actually wrong:

https://github.com/precice/precice/blob/8cf8cb807156e8af06976a362eabac582348d2a2/src/time/Waveform.cpp#L57-L74

`timeWindows < 3` should be `timeWindows < 2` instead etc.
We need to take the intial data into account for extrapolation (and zero values if not initialized).
This means that we also need to call `storeDataInWaveforms()` after/in `initializeData`.

Similar to previously done in `updateOldValues()` here:

https://github.com/precice/precice/blob/3ea47c25ba972e89f5f23b4e4d1707daf578a2a2/src/cplscheme/SerialCouplingScheme.cpp#L73-L94

BenjaminRodenberg:
Updated the `Waveform.cpp` and tests according to https://github.com/precice/precice/pull/1098#issuecomment-924093493.

@uekerman does this look better?

For future reference (in case we have to roll back):

* 8cf8cb8 ignores initial data (both initialized and uninitialized data)
* d130d68 uses initial data (both initialized and uninitialized data, potentially using zero)

BenjaminRodenberg:
I did a study for 1) the CalculiX example where we originally discovered the bug and 2) the partitioned heat equation. In both cases I looked at the iterations.

# CalculiX example

* There is still a slight difference in iterations between this PR (d130d68) and preCICE@master (3ea47c2). So I guess we have not yet found *the* bug.

# partitioned heat equation

* I used `<extrapolation-order value="2" />` and kept everything else as in https://github.com/precice/tutorials/commit/45eedc6281bf08958fcefe59756ed58961dcb147.
* I took a look at `precice-Neumann-iterations.log` for this PR (d130d68), preCICE@master (3ea47c2) and preCICE@develop (d130d68); Data below
* Some conclusions:
    * We can also see the bug for the relatively simple partitioned heat equation
    * For this PR the iterations are clearly better, but still not the same as for master
    * We only need a single iteration for many time windows. This is expected, because our analytical solution is only linear in time and therefore the extrapolation is ideal, if we use the correct data. I have to furtherly investigate and understand this.

### d130d68

```
TimeWindow  TotalIterations  Iterations  Convergence  QNColumns  DeletedQNColumns  DroppedQNColumns  
     1       9       9       1       8       0       0  
     2      14       5       1      10       0       2  
     3      17       3       1      10       2       0  
     4      18       1       1      10       2       0  
     5      19       1       1      10       2       0  
     6      20       1       1      10       2       0  
     7      21       1       1      10       2       0  
     8      24       3       1      10       2       0  
     9      27       3       1      10       2       0  
    10      28       1       1      10       2       0
```

### 28eeef8

```
TimeWindow  TotalIterations  Iterations  Convergence  QNColumns  DeletedQNColumns  DroppedQNColumns  
     1       9       9       1       8       0       0  
     2      15       6       1      10       3       0  
     3      21       6       1      10       5       0  
     4      27       6       1      10       5       0  
     5      29       2       1      10       1       0  
     6      31       2       1      10       1       0  
     7      33       2       1      10       1       0  
     8      37       4       1      10       3       0  
     9      42       5       1      10       4       0  
    10      47       5       1      10       4       0
```

### 3ea47c2
```
TimeWindow  TotalIterations  Iterations  Convergence  QNColumns  DeletedQNColumns  DroppedQNColumns  
1           9                9  1  8  0  0  
2           10               1  1  8  0  0  
3           11               1  1  8  0  0  
4           12               1  1  8  0  0  
5           15               3  1  10  0  0  
6           16               1  1  10  0  0  
7           19               3  1  10  2  0  
8           22               3  1  10  2  0  
9           23               1  1  10  2  0  
10          25               2  1  10  1  0
```

BenjaminRodenberg:
The most simple case I can come up with is by changing the partitoned heat tutorial the following way:

```
diff --git a/partitioned-heat-conduction/fenics/problem_setup.py b/partitioned-heat-conduction/fenics/problem_setup.py
index cb1b164..e8073db 100644
--- a/partitioned-heat-conduction/fenics/problem_setup.py
+++ b/partitioned-heat-conduction/fenics/problem_setup.py
@@ -35,7 +35,7 @@ class StraightBoundary(SubDomain):
 
 
 def get_geometry(domain_part):
-    nx = ny = 9
+    nx = ny = 3
     low_resolution = 5
     high_resolution = 5
     n_vertices = 20
diff --git a/partitioned-heat-conduction/precice-config.xml b/partitioned-heat-conduction/precice-config.xml
index a810f28..b0262fa 100644
--- a/partitioned-heat-conduction/precice-config.xml
+++ b/partitioned-heat-conduction/precice-config.xml
@@ -41,13 +41,14 @@
 
       <participants first="Dirichlet" second="Neumann"/>
-      <max-time value="1.0"/>
+      <max-time value="0.2"/>
       <time-window-size value="0.1"/>
       <max-iterations value="100"/>
       <exchange data="Flux" mesh="Dirichlet-Mesh" from="Dirichlet" to="Neumann" />
       <exchange data="Temperature" mesh="Neumann-Mesh" from="Neumann" to="Dirichlet" initialize="true"/>
       <relative-convergence-measure data="Flux" mesh="Dirichlet-Mesh" limit="1e-5"/>
       <relative-convergence-measure data="Temperature" mesh="Neumann-Mesh" limit="1e-5"/>
+      <extrapolation-order value="1" />
       <acceleration:IQN-ILS>
         <data name="Temperature" mesh="Neumann-Mesh"/>
         <initial-relaxation value="0.1"/>
```

* For our reference implementation (3ea47c2) we need 6 iterations in the first and only a single iteration in the second window.
* For this PR (d130d68) the simulation breaks with the following error
```
---[precice]  iteration: 6 of 100, time-window: 1, time: 0 of 0.2, time-window-size: 0.1, max-timestep-length: 0.1, ongoing: yes, time-window-complete: no, read-iteration-checkpoint 
Solving linear variational problem.
---[precice]  relative convergence measure: relative two-norm diff of data "Flux" = 4.42e-14, limit = 1.00e-05, normalization = 4.00e+00, conv = true
---[precice]  relative convergence measure: relative two-norm diff of data "Temperature" = 2.82e-15, limit = 1.00e-05, normalization = 6.99e+00, conv = true
---[precice]  All converged
---[precice] WARNING:  The number of columns in the least squares system exceeded half the number of unknowns at the interface. The system will probably become bad or ill-conditioned and the quasi-Newton acceleration may not converge. Maybe the number of allowed columns ("max-used-iterations") should be limited.
---[precice] ERROR:  Attempting to add a zero vector to the quasi-Newton V matrix. This means that the residual in two consecutive iterations is identical. There is probably something wrong in your adapter. Maybe you always write the same (or only incremented) data or you call advance without providing  new data first.
```

fsimonis:
Could this be affected by https://github.com/precice/precice/pull/1092 ?

src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp
```diff
@@ -1265,20 +1536,32 @@ BOOST_AUTO_TEST_CASE(testInitializeData)
...
     BOOST_TEST(fixture.isImplicitCouplingScheme(cplScheme));
     cplScheme.initializeData();
     BOOST_TEST(cplScheme.hasDataBeenReceived());
+    // ensure that initial data was read
+    BOOST_TEST(receiveCouplingData->values().size() == 3);
     BOOST_TEST(testing::equals(receiveCouplingData->values(), Eigen::Vector3d(1.0, 2.0, 3.0)));
     BOOST_TEST(receiveCouplingData->previousIteration().size() == 3);
-    BOOST_TEST(testing::equals(receiveCouplingData->previousIteration(), Eigen::Vector3d(0.0, 0.0, 0.0)));
-    BOOST_TEST(sendCouplingData->previousIteration().size() == 1);
+    BOOST_TEST(testing::equals(receiveCouplingData->previousIteration(), Eigen::Vector3d(1.0, 2.0, 3.0)));
```
BenjaminRodenberg, line 1557: I think this here is the important part: If the initial data should be considered for extrapolation, we also have to initialize the `previousIteration` with the initial data. Otherwise the data is lost and the extrapolation cannot use it. I will try to prepare a fix now. Maybe the fix will also end up somewhere else (not in `previousIteration`, but somewhere in `Waveform`)

src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp
```diff
@@ -673,17 +702,253 @@ BOOST_AUTO_TEST_CASE(testAccelerationWithLinearExtrapolation)
...
+  }
+
+  // second window
+  for (int i = 0; i < maxIterations; i++) {
+    // first, second and third iteration
+    BOOST_TEST(cplScheme.isCouplingOngoing());
+    if (context.isNamed(first)) {
+      if (i == 0) {
+        // first order extrapolation uses initial data and final value from last window.
+        BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 0); // = 2*2 - 4
```
BenjaminRodenberg, line 941: This part is currently failing, because the extrapolation uses `0` for initial data - even though `4` has been provided above.

BenjaminRodenberg:
d0958a1 changes functionality (extrapolation is now applied to both participants, even for serial coupling), but also seems to fix most of the issues. There is still a minor inconsistency that is probably related to convergence measurement. Current todos from my perspective:

- [ ] Fix convergence measurement
- [ ] Make sure that functionality does not change ( = only extrapolate accelerated data)

uekerman:
Works for the Alya Flap :tada: 
@DavidSCN @KyleDavisSA Could you please check your case again with this version?

DavidSCN:
Running 10 time steps FSI3 gives exactly the same results between master and this branch (extrapolation order 2)

BenjaminRodenberg:
From my perspective this PR is now ready for review & merging. @uekerman and @KyleDavisSA, please review. I will then squash and merge.

Short summary:

* Main problem was that extrapolation was not applied correctly. Initial data should be used for extrapolation, as well. Even if it is uninitialized ( = 0 ). Meaning: At the end of the first window, we use the result of the first window and the initial data for performing first order extrapolation, and so on.
* Our tests also check this behavior. I also added a new test for first order extrapolation with non-zero initial data.
* Our tests are now able to detect the regression that will be fixed by this PR. I ran `git checkout develop src/cplscheme/BaseCouplingScheme.cpp` to get to the buggy implementation (5344998) and ran the updated version of the tests from this PR. I now get several failing tests (see below).
* I did not add a Changelog entry, because the buggy implementation was not released. This is just an internal bugfix that does not have to go into the changelog.

### Failing tests for implementation from `develop` 

```
4: /home/benjamin/Programming/precice/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp(709): error: in "CplSchemeTests/SerialImplicitCouplingSchemeTests/testAccelerationWithLinearExtrapolation": check mesh->data(receiveDataIndex)->values()(0) == 3.5 has failed [2.5 != 3.5]. Relative difference exceeds tolerance [0.4 > 1e-09]
4: /home/benjamin/Programming/precice/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp(712): error: in "CplSchemeTests/SerialImplicitCouplingSchemeTests/testAccelerationWithLinearExtrapolation": check mesh->data(receiveDataIndex)->values()(0) == 3.25 has failed [2.75 != 3.25]. Relative difference exceeds tolerance [0.181818 > 1e-09]
4: /home/benjamin/Programming/precice/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp(749): error: in "CplSchemeTests/SerialImplicitCouplingSchemeTests/testAccelerationWithLinearExtrapolation": check mesh->data(receiveDataIndex)->values()(0) == 3 has failed [5 != 3]. Relative difference exceeds tolerance [0.666667 > 1e-09]
4: /home/benjamin/Programming/precice/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp(941): error: in "CplSchemeTests/SerialImplicitCouplingSchemeTests/testLinearExtrapolationInit": check mesh->data(receiveDataIndex)->values()(0) == 0 has failed [4 != 0]. Absolute value exceeds tolerance [|4| > 1e-09]
4: /home/benjamin/Programming/precice/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp(944): error: in "CplSchemeTests/SerialImplicitCouplingSchemeTests/testLinearExtrapolationInit": check mesh->data(receiveDataIndex)->values()(0) == 1.5 has failed [2.5 != 1.5]. Relative difference exceeds tolerance [0.666667 > 1e-09]
4: /home/benjamin/Programming/precice/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp(947): error: in "CplSchemeTests/SerialImplicitCouplingSchemeTests/testLinearExtrapolationInit": check mesh->data(receiveDataIndex)->values()(0) == 2.25 has failed [2.75 != 2.25]. Relative difference exceeds tolerance [0.222222 > 1e-09]
4: /home/benjamin/Programming/precice/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp(984): error: in "CplSchemeTests/SerialImplicitCouplingSchemeTests/testLinearExtrapolationInit": check mesh->data(receiveDataIndex)->values()(0) == 3 has failed [5 != 3]. Relative difference exceeds tolerance [0.666667 > 1e-09]
4: /home/benjamin/Programming/precice/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp(1150): error: in "CplSchemeTests/SerialImplicitCouplingSchemeTests/testAccelerationWithQuadraticExtrapolation": check mesh->data(receiveDataIndex)->values()(0) == 3.5 has failed [2.5 != 3.5]. Relative difference exceeds tolerance [0.4 > 1e-09]
4: /home/benjamin/Programming/precice/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp(1153): error: in "CplSchemeTests/SerialImplicitCouplingSchemeTests/testAccelerationWithQuadraticExtrapolation": check mesh->data(receiveDataIndex)->values()(0) == 3.25 has failed [2.75 != 3.25]. Relative difference exceeds tolerance [0.181818 > 1e-09]
4: /home/benjamin/Programming/precice/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp(1194): error: in "CplSchemeTests/SerialImplicitCouplingSchemeTests/testAccelerationWithQuadraticExtrapolation": check mesh->data(receiveDataIndex)->values()(0) == 4.25 has failed [4 != 4.25]. Relative difference exceeds tolerance [0.0625 > 1e-09]
4: /home/benjamin/Programming/precice/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp(1197): error: in "CplSchemeTests/SerialImplicitCouplingSchemeTests/testAccelerationWithQuadraticExtrapolation": check mesh->data(receiveDataIndex)->values()(0) == 4.625 has failed [4.5 != 4.625]. Relative difference exceeds tolerance [0.0277778 > 1e-09]
4: /home/benjamin/Programming/precice/src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp(1234): error: in "CplSchemeTests/SerialImplicitCouplingSchemeTests/testAccelerationWithQuadraticExtrapolation": check mesh->data(receiveDataIndex)->values()(0) == 5 has failed [7 != 5]. Relative difference exceeds tolerance [0.4 > 1e-09]
```

These failures are expected and show that at 5344998 extrapolation is performed incorrectly. These tests pass when using the implementation from this PR.

src/precice/tests/SerialTests.cpp
```diff
@@ -3255,20 +3255,23 @@ void multiCouplingThreeSolvers(const std::string configFile, const TestContext &
   }
 }
 
+// @todo this test causes problems, if run locally via make test_base
```
uekerman, line 3258: Where are these comments coming from? Why are there problems?

uekerman (approved):
Looking good. Only one doubt below.

src/precice/tests/SerialTests.cpp
BenjaminRodenberg: If I run `make test_base` the serial tests run into a timeout.

I'm not sure what's going on here. I guess it might be system specific, because I cannot reproduce this behavior with our CI. If I use `testprecice` everything is ok, as well. So I do not think that we really have to worry. I still would like to understand what's going on. But I think the priority here is not very high currently.

Did you run the tests? If yes, how? I you also see strange behavior for `make test_base`, I think we should worry. But I currently assume that just something is broken on my system.

fsimonis: The tests run on my up2date arch without issues.
Maybe you have a stale `precice-run` in `TestOutputs/serial`?

BenjaminRodenberg: I and @fsimonis did a clean rebuild of preCICE and the problem is gone. Probably some build artifacts causing trouble. Will remove the `@todo` and then the PR is ready for merging.

fsimonis: @BenjaminRodenberg and I traced the issue down to stale files.

PR #1099, opened 2021-09-20
Add experimental tag to unlock full API

fsimonis:
## Main changes of this PR

This PR adds an explicit way of enabling experimental features in preCICE.
Using experimental features without enabling this tag will result in: _reformatted_
```
ERROR:
You called the API function "setMeshAccessRegion", which is part of the experimental API.
You may unlock the full API by specifying <solver-interface experimental="true" /> in the configuration.
Please be aware that experimental features may change at any time.
```

Experimental features may be enabled by setting the `experimental` attribute in the `solver-interface` to `true`.
```xml
<solver-interface experimental="true" ... />
```

This approach
* prevents users to rely on experimental features by mistake, as they need to manually enable them.
* embeds experimental use-cases directly into the configuration file.
* simplifies support due to the above.
* allows to release experimental features to early adopters.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/config/SolverInterfaceConfiguration.hpp
+++ b/src/precice/config/SolverInterfaceConfiguration.hpp
@@ -46,7 +46,7 @@ public:
   int getDimensions() const;
 
   /// @brief Returns whether experimental features are allowed or not
-  int allowsExperimental() const
+  bool allowsExperimental() const
   {
     return _experimental;
   }
--- a/src/precice/config/SolverInterfaceConfiguration.hpp
+++ b/src/precice/config/SolverInterfaceConfiguration.hpp
@@ -45,6 +45,12 @@ public:
    */
   int getDimensions() const;
 
+  /// @brief Returns whether experimental features are allowed or not
+  int allowsExperimental() const
+  {
+    return _experimental;
+  }
+
   const mesh::PtrDataConfiguration getDataConfiguration() const
   {
     return _dataConfiguration;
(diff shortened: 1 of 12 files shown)


src/precice/config/SolverInterfaceConfiguration.hpp
```diff
@@ -45,6 +45,12 @@ class SolverInterfaceConfiguration : public xml::XMLTag::Listener {
    */
   int getDimensions() const;
 
+  /// @brief Returns whether experimental features are allowed or not
+  int allowsExperimental() const
```
uekerman, line 49:
```suggestion
  bool allowsExperimental() const
```
?

uekerman (approved):
Thanks for the really quick addition!
I like the feature.
Not sure, what to do with the "resetMesh" integration tests.
They still pass, right?
Anyway, please open an issue (if not already done) to not forget these.

src/precice/config/SolverInterfaceConfiguration.hpp
fsimonis: Bad typos

MakisH (commented):
Nice how simple this turned out! Even integration tests can use the same workflow.

Regarding documentation, shall we only document this together with each experimental feature page, or should we add it anywhere else (e.g. in the [overview](https://precice.org/configuration-overview.html), as "more features are available")?

uekerman:
We could also do a new subfolder under "couple your code", parallel to "advanced", called "experimental". From "configuration/overview" we could link there.

MakisH:
> We could also do a new subfolder under "couple your code", parallel to "advanced", called "experimental". From "configuration/overview" we could link there.

Issue opened: https://github.com/precice/precice.github.io/issues/88

PR #1101, opened 2021-09-23
Add DaRUS badge to README

uekerman:
Do we want to put such a badge here? Feel free to modify.
Will probably only work once published on DaRUS, hence the draft.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -16,13 +16,13 @@
 <a style="text-decoration: none" href="https://www.precice.org/" target="_blank">
     <img src="https://img.shields.io/website-up-down-green-red/https/precice.org.svg?label=website" alt="preCICE website status">
 </a>
-<a style="text-decoration: none" href="https://doi.org/10.18419/darus-2125" target="_blank">
-    <img src="https://img.shields.io/badge/doi-10.18419%2Fdarus--2125-d45815.svg" alt="preCICE distribution">
-</a>
 
 **Project Status**  
 <a style="text-decoration: none" href="https://github.com/precice/precice/releases/latest" target="_blank">
     <img src="https://img.shields.io/github/release/precice/precice.svg" alt="Release">
+</a>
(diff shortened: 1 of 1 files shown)


MakisH (commented):
We should definitely add a badge, yes!

However, shouldn't we better put it under "project status"?

Another question: how do we want to treat this DOI once we have our new reference paper published? What do we want people to cite? How should we guide them between the two options?

We could also make our own badge "preCICE distribution" that shows the version instead of the DOI, but it still links there.

uekerman:
> However, shouldn't we better put it under "project status"?

Yes, right.

> Another question: how do we want to treat this DOI once we have our new reference paper published? What do we want people to cite? How should we guide them between the two options?

That's an important question, which we should discuss in detail. I'll add it to the next telco.

uekerman:
I updated the badge to distribution v2202 and moved it to the project status section.

MakisH (approved):
That's already nice! Just to clarify: if I am an external and see a repository mentioning a DOI (as here), I will directly assume this is the citation. This may be a bit confusing for now, since we also have the "what to cite" at the bottom, but I guess we will tidy this up once we also have the v2 paper online.

By the way, do we want to push/wait to also have the v2 paper in v2.4?

uekerman:
> By the way, do we want to push/wait to also have the v2 paper in v2.4?

You mean in the README?

MakisH:
> > By the way, do we want to push/wait to also have the v2 paper in v2.4?
> 
> You mean in the README?

yes, since we have the 2016 paper at the bottom of the README.

uekerman:
> By the way, do we want to push/wait to also have the v2 paper in v2.4?

I guess this is a unnecessary dependency. In the end, it's not so important what is in the README of the release, but what we have on develop (meaning rendered on the default branch on GitHub).
Let's update this, once we do all the "new paper updates".

Merging here.

PR #1106, opened 2021-10-05
Set extrapolation order in CouplingScheme constructor.

BenjaminRodenberg:
## Main changes of this PR

Set extrapolation order in constructor. This allows us to remove the setter method and declare `const _extrapolationOrder`.

Simplifies class hierarchy and requires user to explicitly set extrapolation order for implicit coupling schemes (in tests).

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/cplscheme/MultiCouplingScheme.hpp
+++ b/src/cplscheme/MultiCouplingScheme.hpp
@@ -35,6 +35,7 @@ public:
  * @param[in] m2ns M2N communications to all other participants of coupling scheme.
  * @param[in] dtMethod Method used for determining the time window size, see https://www.precice.org/couple-your-code-timestep-sizes.html
  * @param[in] maxIterations maximum number of coupling sub-iterations allowed.
+ * @param[in] extrapolationOrder order used for extrapolation
  */
   MultiCouplingScheme(
       double                             maxTime,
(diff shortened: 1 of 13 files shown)


src/cplscheme/MultiCouplingScheme.hpp
```diff
@@ -45,7 +46,8 @@ class MultiCouplingScheme : public BaseCouplingScheme {
       std::map<std::string, m2n::PtrM2N> m2ns,
       constants::TimesteppingMethod      dtMethod,
       const std::string &                controller,
-      int                                maxIterations = -1);
+      int                                maxIterations      = -1,
+      int                                extrapolationOrder = -1);
```
uekerman, line 50: MultiCouplingScheme only exists in implicit mode, right?
One could argue that there should be no default value here then.

uekerman (approved):
Looks good. Does indeed simplify things a bit.
If you want to go the extra mile, you could convert the magic number "-1" into an `UNDEFINED` constant.
Please merge when ready.

src/cplscheme/MultiCouplingScheme.hpp
BenjaminRodenberg: Good point. Did this in 4f2a110715aa948b4240107445232c4d8c83b32b.

PR #1109, opened 2021-10-06
Refactor track time windows in coupling scheme

BenjaminRodenberg:
## Main changes of this PR

Changes the way how `_timeWindows` is incremented, if a coupling step converges

I think that incrementing `_timeWindows` first and then (if there is no convergence) decrementing it again is confusing. Note that we should be very careful here, since this PR might easily introduce a regression. We should carefully check output files for some test cases before merging.

## Author's checklist

* [x] merge #1083 
* [ ] change to merge into `develop`

## Reviewers' checklist


src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -228,11 +226,11 @@ void BaseCouplingScheme::advance()
         // coupling iteration.
         PRECICE_ASSERT(math::greater(_computedTimeWindowPart, 0.0));
         _time = _time - _computedTimeWindowPart;
-        _timeWindows -= 1;
       } else { // write output, prepare for next window
```
uekerman, line 229: Now, you treat the time window counter and time differently. I guess this could reduce readability again.

src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -373,6 +373,11 @@ class BaseCouplingScheme : public CouplingScheme {
    */
   void determineInitialReceive(DataMap &receiveData);
 
+  /**
+   * @brief returns true, if current iteration is in the last time window.
+   */
+  bool isLastTimeWindow() const;
```
uekerman, line 379: The name of this function is confusing. I would expect this function to be true when we reached the last time window of the simulation.

uekerman (changes requested):
The treatment of time windows is complicated, yes, but I am not fully convinced yet that this change will make it easier. IIRC we have touched this already a few times (some years back).
Please don't underestimate the side conditions this has on log files (watchpoint, export, iterations, convergence, ...) and logging. It would require thorough testing (serial vs parallel coupling, files written by first or second participant, subcycling, explicit vs implicit, ...)

In a way, the old realization is also closer to how things are typically handled in adapters.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: True. So we should probably also touch `_time` and make it consistent.

BenjaminRodenberg:
@uekerman This PR has diverged quite a lot and looking at the current situation I don't see a clear gain and at the same time the risk is high, because test coverage on some of the side effects you mentioned above is bad as far as I know. I'm closing this PR now. We can still go back to it and reopen it/redo the changes, if we feel the pain of the implementation of time window tracking again. Feel free to reopen, if you have the impression we are loosing something important here.

This PR is a v3.0.0 PR anyway, because I think applying these changes to v2 only makes things more complicated.

PR #1122, opened 2021-10-28
Add tooling API

### LINKED ISSUE #1117: Tooling API

`binprecice` is currently implemented using the xml and config internals. This is ugly and leads to unnecessary dependencies, which in turn makes the CMake configuration ugly and prevents #623.

The clean solution is to implement its functionality as part of the API as stand-alone functions, similar to `precice::getVersionInformation()`. Then `binprecice` could simply rely on the API of preCICE.

Implement the required functions as stand-alone functions in the API.
Currently, these are:
* Print the configuration reference as XML with annotations.
* Print the DTD file, which can be used to validate XML configuration files.
* Print the markdown version of the reference, which is used for the website.

**Open questions:**

How should we name these functions?
Should we implement them as separate functions or a single function using an enum.
```
// option A
void printConfigAsMD(std::ostream &out);
void printConfigAsDTS(std::ostream &out);
void printConfigAsXML(std::ostream &out);

// option B
enum struct ReferenceType { md, dtd, xml };
void printReference(std::ostream &out, ReferenceType reftype);
```

Should we place these functions in the header `precice/SolverInterface.hpp` or in a new header such as `precice/Tooling.hpp`?
The latter could simplify adding additional tools in the future.

Can we omit these functions from the bindings?
After all, we ship the `binprecice`, which is a convenient way of using these features.

fsimonis:
## Main changes of this PR

This PR introduces a tooling API in a separate header, which allows to implement `binprecice` without using preCICE internals.
This simplifies the source of the `binprecice` utility as well the necessary CMake.
I also extended the `binprecice` by a `version` command to print the version string.

Closes #1117 

* Clear API functions, no use of internals
* Cleaner CMake. `binprecice` now simply needs to link against the library.
* Simplifies #623 

## Author's checklist

## Reviewers' checklist

* [ ] Do you agree with the separate header `Tooling.hpp`?
* [ ] Do you agree with using separate functions instead of an enum for the output format?

### DIFF
--- a/src/drivers/main.cpp
+++ b/src/drivers/main.cpp
@@ -31,7 +31,7 @@ int main(int argc, char **argv)
     return 0;
   }
   if (action == "md" && args == 0) {
-    printConfigReference(std::cout, ConfigReferenceType::Markdown);
+    printConfigReference(std::cout, ConfigReferenceType::MD);
     return 0;
   }
   if (action == "xml" && args == 0) {
--- a/src/drivers/main.cpp
+++ b/src/drivers/main.cpp
@@ -12,6 +12,7 @@ void printUsage()
   std::cerr << "Print DTD for XML config :  binprecice dtd\n";
   std::cerr << "Print Markdown reference :  binprecice md\n";
   std::cerr << "Print preCICE version    :  binprecice version\n";
+  std::cerr << "                            binprecice --version\n";
 }
 
 int main(int argc, char **argv)
@@ -38,7 +39,7 @@ int main(int argc, char **argv)
     printConfigReference(std::cout, ConfigReferenceType::XML);
     return 0;
   }
-  if (action == "version" && args == 0) {
+  if ((action == "version" || action == "--version") && args == 0) {
     std::cout << precice::getVersionInformation() << '\n';
     return 0;
   }
--- a/src/drivers/main.cpp
+++ b/src/drivers/main.cpp
@@ -24,16 +24,18 @@ int main(int argc, char **argv)
   const std::string action(argv[1]);
   const int         args = argc - 2;
 
+  using namespace precice::tooling;
+
   if (action == "dtd" && args == 0) {
-    precice::tooling::printConfigAsXML(std::cout);
+    printConfigReference(std::cout, ConfigReferenceType::DTD);
     return 0;
   }
   if (action == "md" && args == 0) {
-    precice::tooling::printConfigAsMD(std::cout);
+    printConfigReference(std::cout, ConfigReferenceType::Markdown);
     return 0;
   }
   if (action == "xml" && args == 0) {
-    precice::tooling::printConfigAsDTD(std::cout);
+    printConfigReference(std::cout, ConfigReferenceType::XML);
     return 0;
   }
   if (action == "version" && args == 0) {
--- a/src/drivers/main.cpp
+++ b/src/drivers/main.cpp
@@ -1,9 +1,7 @@
+#include <cassert>
 #include <iostream>
+#include <precice/SolverInterface.hpp>
 #include <string>
-#include "precice/config/Configuration.hpp"
-#include "precice/impl/versions.hpp"
-#include "utils/assertion.hpp"
-#include "xml/Printer.hpp"
 
 void printUsage()
 {
@@ -43,17 +41,13 @@ int main(int argc, char **argv)
   }
 
   if (runHelp) {
-    precice::config::Configuration config;
-    precice::xml::toDocumentation(std::cout, config.getXMLTag());
+    precice::printConfigAsXML(std::cout);
   } else if (runDtd) {
-    precice::config::Configuration config;
-    precice::xml::toDTD(std::cout, config.getXMLTag());
+    precice::printConfigAsDTD(std::cout);
   } else if (runMD) {
-    precice::config::Configuration config;
-    std::cout << "<!-- generated with preCICE " PRECICE_VERSION " -->\n";
-    precice::xml::toMarkdown(std::cout, config.getXMLTag());
+    precice::printConfigAsMD(std::cout);
   } else {
-    PRECICE_ASSERT(false);
+    assert(false);
   }
   return 0;
 }
--- a/src/drivers/main.cpp
+++ b/src/drivers/main.cpp
@@ -1,6 +1,7 @@
 #include <cassert>
 #include <iostream>
-#include <precice/SolverInterface.hpp>
+#include <precice/Tooling.hpp>
+#include <stdexcept>
 #include <string>
 
 void printUsage()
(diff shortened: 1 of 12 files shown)


src/precice/Tooling.hpp
```diff
@@ -0,0 +1,21 @@
+#pragma once
+
+#include <iosfwd>
+#include <string>
+
+namespace precice {
+
+namespace tooling {
```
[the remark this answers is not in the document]
DavidSCN: Nice, thanks!

uekerman (changes requested):
Indeed simplifies things :+1:
I did not look into details here, more the big picture.

Shouldn't `getVersion` also be moved to the tooling header now? (We would probably consider this breaking, but we could move it with v3. If you agree please open an issue and/or do a redirection now already.)

Seems we have no (integration) tests for binprecice, right? They should be easy to implement. Just take a config and do a regression against the expected files.

src/precice/Tooling.hpp
```diff
@@ -0,0 +1,39 @@
...
+ *
+ * @see \ref precice::getVersionInformation which is exposed via the bindings
+ */
+namespace tooling {
+
+/** @brief Prints the configuration reference as Markdown.
+ *
+ * @param[inout] out the stream to write the result to
+ */
+void printConfigAsMD(std::ostream &out);
```
uekerman, line 23:
```suggestion
void printConfigReferenceAsMD(std::ostream &out);
```
It does not print a valid config. Thinking one step ahead we might also print actual configs in the future.

fsimonis:
> Shouldn't getVersion also be moved to the tooling header now? (We would probably consider this breaking, but we could move it with v3. If you agree please open an issue and/or do a redirection now already.)

Moving `getVersion` to the `Tooling.hpp` will be tricky because:
* This has to be part of the API to stay compliant with xSDK. Meaning we have to/should support it in the bindings.
* Moving things around in the API it is a breaking change.
* The bindings then cover only a fraction of the `Tooling.hpp`, which may be confusing.
* The bindings then need to provide additional headers and source files to stay consistent with the C++ version.


> Seems we have no (integration) tests for binprecice, right? They should be easy to implement. Just take a config and do a regression against the expected files

Not directly for binprecice. The functionality is covered in the [PrinterTests](https://github.com/precice/precice/blob/develop/src/xml/tests/PrinterTest.cpp) though.

We could add tests, that call `binprecice`. The question is if this is worth it.

src/precice/Tooling.hpp
fsimonis: True. The longer function name makes a single function now more appealing to me.

src/drivers/main.cpp
```diff
@@ -1,59 +1,47 @@
...
 {
-  std::cout << "Usage:\n\n";
-  std::cout << "Print XML reference      :  binprecice xml\n";
-  std::cout << "Print DTD for XML config :  binprecice dtd" << std::endl;
-  std::cout << "Print Markdown reference :  binprecice md" << std::endl;
+  std::cerr << "Usage:\n\n";
+  std::cerr << "Print XML reference      :  binprecice xml\n";
+  std::cerr << "Print DTD for XML config :  binprecice dtd\n";
+  std::cerr << "Print Markdown reference :  binprecice md\n";
+  std::cerr << "Print preCICE version    :  binprecice version\n";
```
MakisH, line 14: Why `binprecice version` and not `binprecice --version`? Or why not both? It is more complicated?

src/drivers/main.cpp
```diff
@@ -1,59 +1,47 @@
...
+  std::cerr << "Check configuration      :  binprecice check FILE [ PARTICIPANT [ COMMSIZE ] ]\n";
```
MakisH, line 15: `binprecice check` is not mentioned in the changelog.

MakisH (commented):
> Does the changelog entry make sense? Is it formatted correctly?

Yes and yes.

> Do you understand the code changes?

Yes. Nice that this also simplifies `CMakeLists.txt` and I like that you use a more "to the point" parsing of the command-line arguments.

> Do you agree with the separate header `Tooling.hpp`?

Yes, as this would "hide" these tools (meant only for `binprecice`) from the header that all the solvers and bindings are including. It also makes it clearer that we don't want these methods in the bindings.

> Do you agree with using separate functions instead of an enum for the output format?

With the long names, the separate functions look pretty nice! I agree.

src/drivers/main.cpp
fsimonis: This somehow sneaked in from another branch. Thanks for pointing this out!

src/drivers/main.cpp
fsimonis: `version` really is a command (_please give me the version info_) and not an option to an existing command (_please print the XML reference and additionally tell me the version info_).

I know that many tools implement the version as on option, however, many of these implementations are total non-sense. Docker for example prints the version and quits when passing `--version`, so it is a command disguised as an option. Also options normally come with a short form. For the version, this should be `-v`, which is also often used for short form of `--verbose`.

So, we could add the `--version` option, but it is a classical case of _high cost, low benefit_.

fsimonis:
I merged all config generation functions into a single function with a type argument and added tests for both the interface functions and the binprecice itself.

src/drivers/main.cpp
MakisH: But it should be an option to the application, right? An application does not need to have additional commands. I usually don't know anything about an application, but I know that I can type `application --version`.

MakisH:
> > With the long names, the separate functions look pretty nice! I agree.
> 
> @MakisH What's your opinion on the merged functions? `getConfigReferenceAsXML` starts getting a bit long IMHO, but I am still a bit unsure.

This enum way seems even nicer! In the issue, you mentioned that future reference types could need additional arguments. What/which direction did you have in mind?

src/precice/Tooling.hpp
```diff
@@ -0,0 +1,42 @@
...
+/** The type of reference to generate
+ * @see \ref precice::tooling::printConfigReference
+ */
+enum struct ConfigReferenceType {
+  /// XML with inlined help
+  XML = 0,
+  /// DTD to check an XML
+  DTD = 1,
+  /// Markdown version used for the website
+  Markdown = 2
```
uekerman, line 28:
```suggestion
  MD = 2
```
Wouldn't this be more consistent?

fsimonis: True, will fix in a separate commit

fsimonis:
> In the issue, you mentioned that future reference types could need additional arguments. What/which direction did you have in mind?

@MakisH It really was more of a worry instead of a concrete problem. I am not sure if we will ever run into this though.

src/precice/Tooling.hpp
fsimonis: Merged in 29fb82c

src/drivers/main.cpp
fsimonis: Well, not really. `application --version` means run the application (or default command) and additionally print the version. `application version` means, give me the version.
It is a weird confused convention (e.g [`git --version`](https://git-scm.com/docs/git#Documentation/git.txt---version) is implemented as an "alias" to `git version`, while `docker --version` means something different than `docker version`).

Anyhow, I'll add `--version` as an alias to move on.

uekerman (approved):
Tried it out. `binprecice` works like a charm. Tests look good.
Nice job :+1:

MakisH (approved):

PR #1128, opened 2021-11-05
Add VTK PolyData Export

### LINKED ISSUE #694: File export for Paraview Glance
**Please describe the problem you are trying to solve?**
I would like to export and visualize the interface meshes of a participant using [paraview glance](https://kitware.github.io/paraview-glance/).
This is currently not possible as the legacy vtk files are not supported.

Add an additional export format which can be imported into glance.

Conversion using an external tool, which is inconvenient.

kursatyurt:
## Main changes of this PR

Add VTK PolyData `vtp` support for serial cases.

(closes #694)
Parallel cases are not supported. Could it be possible or `xmlPolyData` is required? 

## Author's checklist

* [X] I ran `tools/formatting/check-format` and everything is formatted correctly.
* [X] I sticked to C++14 features.
* [X] I sticked to CMake version 3.10.
* [X] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does change in configuration correct?

### DIFF
--- a/src/io/Constants.cpp
+++ b/src/io/Constants.cpp
@@ -19,6 +19,11 @@ int exportVTKXML()
   return 3;
 }
 
+int exportVTP()
+{
+  return 4;
+}
+
 } // namespace constants
 } // namespace io
 } // namespace precice
--- a/src/io/Constants.hpp
+++ b/src/io/Constants.hpp
@@ -7,6 +7,7 @@ namespace constants {
 int exportVTK();
 int exportAll();
 int exportVTKXML();
+int exportVTP();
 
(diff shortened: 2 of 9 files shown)


fsimonis (changes requested):
Thanks for the contribution!
I really like the addition of the `vtp` exporter.

However, we should take the extra step and implement the XML format instead of the legacy format. This then also opens the doors for the parallel support.

It is actually extremely similar to the `VTKXMLExporter`, which supports both serial and parallel participants.

fsimonis:
Closed in favour of #1137

PR #1132, opened 2021-11-09
Configuration Checker

fsimonis:
## Main changes of this PR

This PR extends `binprecice` with the ability to check configuration files for correctness.
The syntax is as follows:
```
$ binprecice check FILE [ PARTICIPANT [ COMMSIZE ] ]
```

* `FILE` is the path to the configuration file
* `PARTICIPANT` is the _optional_ participant to check. This is necessary to detect in-depth configuration issues.
* `COMMSIZE` is the _optional_ size of the communicator. This is necessary to check custom `<master />` tags.

Running the entire solver to test a configuration is not practical.
This extension offers a quick and simple method to check a configuration file which ships directly with preCICE.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/man/man1/binprecice.1
+++ b/docs/man/man1/binprecice.1
@@ -1,29 +1,45 @@
-.TH binprecice 1  "January 18, 2019" "USER COMMANDS"
+.TH binprecice 1
 
 .SH NAME
-binprecice \- dumps the xml reference used for the configuration of libprecice
+binprecice \- command line utility for preCICE
 
 .SH SYNOPSIS
 .B binprecice xml
 |
 .B dtd
+|
+.B md
+|
+[\fB version \fR|\fB --version \fR]
+|
+.B check \fIfile\fR [\fIparticipant\fR [\fIsize\fR]]
 
 .SH OPTIONS
+
 .TP
 .B xml
-Prints the XML reference
+Prints the configuration reference as XML with inlined help to the standard output.
+
 .TP
 .B dtd
-Prints the DTD for the XML config
+Prints the DTD file to the standard output, which may be used to check XML configuration files.
 
-.SH RETURN VALUES
-binprecice returns a zero exist status if it succeeds.
+.TP
+.B md
+Prints the configuration reference as markdown to the standard output.
+
+.TP
+.B version \fR|\fB --version \fR
+Prints the preCICE version string ( \fIprecice::getVersionInformation()\fR ) to the standard output.
+
+.TP
+.B check \fIfile\fR [\fIparticipant\fR [\fIsize\fR]]
+Checks given preCICE configuration \fIfile\fR and prints any errors. Pass the name of a \fIparticipant\fR to fully check its configuration. The optional \fIsize\fR indicates how many ranks run in parallel, which checks custom \fI<master: ... />\fR tags.
+
+.SH EXIT STATUS
+binprecice returns a zero exist status if it succeeds to run the provided command.
 Non zero is returned in case of failure.
 
-.SH AUTHOR
-Frédéric Simonis (simonis (at) in.tum.de)
-
 .SH SEE ALSO
-testprecice(1)
 .PP
-www.precice.org \- the website of the project
+precice.org \- the website of the project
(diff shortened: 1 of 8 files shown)


ajaust:
I guess this is related to the discussion in [config visualizer repository](https://github.com/precice/config-visualizer). More precisely the discussion we had here https://github.com/precice/config-visualizer/issues/9.

docs/man/man1/binprecice.1
```diff
@@ -1,29 +1,45 @@
...
+Prints the preCICE version string ( \fIprecice::getVersionInformation()\fR ) to the standard output.
 
-.SH AUTHOR
-Frédéric Simonis (simonis (at) in.tum.de)
+.TP
+.B check \fIfile\fR [\fIparticipant\fR [\fIsize\fR]]
+Checks given preCICE configuration \fIfile\fR and prints any errors. Pass the name of a \fIparticipant\fR to fully check its configuration. The optional \fIsize\fR indicates how many ranks run in parallel, which checks custom \fI<master: ... />\fR tags.
+
+.SH EXIT STATUS
+binprecice returns a zero exist status if it succeeds to run the provided command.
```
MakisH, line 40:
```suggestion
binprecice returns a zero exit status if it succeeds to run the provided command.
```

src/drivers/main.cpp
```diff
@@ -13,6 +13,7 @@ void printUsage()
   std::cerr << "Print Markdown reference :  binprecice md\n";
   std::cerr << "Print preCICE version    :  binprecice version\n";
   std::cerr << "                            binprecice --version\n";
+  std::cerr << "Check configuration file :  binprecice check FILE [ PARTICIPANT [ COMMSIZE ] ]\n";
```
MakisH, line 16: I am not sure if this is intended, but the style for `FILE [ PARTICIPANT [ COMMSIZE ] ]` is a bit different in the manpage.

src/precice/tests/ToolingTests.cpp
```diff
@@ -48,5 +48,27 @@ BOOST_AUTO_TEST_CASE(DTDReference)
...
+          1));
+}
+
+BOOST_AUTO_TEST_CASE(Parallel)
+{
+  BOOST_REQUIRE_NO_THROW(
+      precice::tooling::checkConfiguration(
+          precice::testing::getPathToSources() + "/precice/tests/config-checker.xml",
+          "SolverTwo",
+          4));
```
MakisH, line 68: Mostly a question than a comment: Is it intended that this test and the one in `CTestConfig.cmake` use a different communicator size?

MakisH (commented):
Just a few minor comments/questions to consider.

I really like the idea, I have not yet tried out the tool.

Don't forget to add it in the [documentation](https://precice.org/tooling-overview.html).

src/precice/tests/ToolingTests.cpp
fsimonis: It is not intended, but also doesn't matter. From the preCICE perspective there are only 2 cases: 1 and >1

docs/man/man1/binprecice.1
```diff
@@ -1,29 +1,45 @@
-.TH binprecice 1  "January 18, 2019" "USER COMMANDS"
+.TH binprecice 1
```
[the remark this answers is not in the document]
fsimonis: We could use pandoc, but this then means that we need pandoc to provide manpages when the user installs precice from source. 
Is this reasonable?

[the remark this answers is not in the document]
boris-martin: Never thought of this, as in my case there was no such thing as "install from source" (just an executable produced), but you could include compiled documentation in the source code, or add an option to have no manpage if you don't have pandoc.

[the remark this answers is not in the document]
MakisH: > We could use pandoc, but this then means that we need pandoc to provide manpages when the user installs precice from source. Is this reasonable?

I did not think of this use case and indeed, let's not add an extra dependency just for this... :see_no_evil:

src/drivers/main.cpp
MakisH: I don't mind any, I mainly pointed at the inconsistency.

fsimonis: Let's leave it then as it is. Thanks for pointing it out though.

MakisH (approved):
I tried it with some arbitrary config errors and it works nicely. It of course shows only one error at the time, which we could maybe hint to the user (although they will quickly figure it out).

This tool would have been so nice to have when we were redesigning the error messages...

fsimonis:
> This tool would have been so nice to have when we were redesigning the error messages...

This is a very good point. Maybe it will come in handy in the future again.

PR #1133, opened 2021-11-10
Shift ownership of Waveform into configuration

BenjaminRodenberg:
## Main changes of this PR

Waveforms are now created during configuration. The resulting object is owned by the configuration class. Other classes only receive a pointer to this object.

This requires us to split creation and initialization of the Waveform object, because `valuesSize` is not known during configuration.

Main motivation is this observation: https://github.com/BenjaminRodenberg/precice/pull/1/files#r745679418

Collecting waveforms at a central place should help us to avoid duplicates and get a better overview.

## Author's checklist

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly? @uekerman 
* [ ] Do you understand the code changes? @uekerman


BenjaminRodenberg:
@uekerman we could already introduce a `WaveformConfiguration.cpp/TimeConfiguration.cpp` here. This would be mostly empty, but provides some structure for #1029. What's you opinion?

src/cplscheme/BiCouplingScheme.cpp
```diff
@@ -50,9 +50,10 @@ BiCouplingScheme::BiCouplingScheme(
 }
 
 void BiCouplingScheme::addDataToSend(
-    const mesh::PtrData &data,
-    mesh::PtrMesh        mesh,
-    bool                 requiresInitialization)
+    const mesh::PtrData &    data,
+    const time::PtrWaveform &ptrWaveform,
```
uekerman, line 54: Why do we have both data and waveform here? Shouldn't the one replace the other?

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -948,15 +949,19 @@ void CouplingSchemeConfiguration::addDataToBeExchanged(
 
     const bool requiresInitialization = exchange.requiresInitialization;
     if (from == accessor) {
-      scheme.addDataToSend(exchange.data, exchange.mesh, requiresInitialization);
+      time::PtrWaveform ptrWaveform(new time::Waveform(_config.extrapolationOrder));
+      // @todo store all waveforms in _waveforms? See acceleration config.
+      scheme.addDataToSend(exchange.data, ptrWaveform, exchange.mesh, requiresInitialization);
```
uekerman, line 954: Couldn't we create the waveforms closer to the time-stepping part of the config? At some point, we will also get configuration options there. Here, it feels a bit misplaced.

src/time/Waveform.hpp
```diff
@@ -16,12 +16,19 @@ class Waveform {
...
    */
-  Waveform(const int valuesSize,
-           const int extrapolationOrder);
+  Waveform(const int extrapolationOrder);
+
+  /**
+   * @brief Used to initialize _timeWindowsStorage according to required size.
+   * @param valuesSize defines how many values one sample in time consists of
+   */
+  void initializeData(const int valuesSize);
```
uekerman, line 31: The name indicates that data is actually initialized here. In preCICE words, this means that data is set to non-zero values. Why not just `initialize`? Or `initializeStorage`?

uekerman (commented):
Goes in the right direction :+1:

> We could already introduce a WaveformConfiguration.cpp/TimeConfiguration.cpp here. This would be mostly empty, but provides some structure for #1029. What's you opinion?

I would only add it once we know what we want to configure. If we realize that there is not so much to configure, we could also keep it in the coupling scheme configuration.

Let's find a better name for `setupDataMatrices`.

src/time/Waveform.hpp
BenjaminRodenberg: `initialize` is good. In case we have to initialize more things at a later point in time, we also don't have to change it again.

src/cplscheme/BiCouplingScheme.cpp
BenjaminRodenberg: I don't think that this is easily possible. The `Waveform` is currently just a buffer for values to be used for interpolation and extrapolation. The `Data` also stores the id and the name of the data - and of course the `_values` itself.

I guess we could replace `Data::_values` with a pointer to the data stored in the waveform. Then the `Waveform` will become our only store for values. But this is a big step, because `Data::_values` are used and updated almost everywhere (also outside of the time or cplscheme package). I am also not sure whether this is a good idea, because then all packages have to interact with the time package and the `Waveform`.

Currently we are copying `Data::_values` into the `Waveform` on demand. Only the cplscheme packages accesses the waveform. Other packages do not have to worry about the waveform at all, because it is wrapped by the cplscheme. Storing data happens here: 

https://github.com/precice/precice/blob/b784b3849142693c315a16698328bfa975b348e5/src/cplscheme/BaseCouplingScheme.cpp#L265

src/cplscheme/config/CouplingSchemeConfiguration.cpp
BenjaminRodenberg: I would rather move this at a later point in time, if necessary. Currently adding the waveform at a different point requires us to add more functions to the `CouplingScheme` (`addWaveformToReceive`, `addWaveformToSend`) and I don't see a clear benefit.

Right now adding everything with a single function enforces the rule "each data has a waveform". This rule might be removed at a later point in time, but currently we are following this rule (because also extrapolation is applied globally to all data).

BenjaminRodenberg:
> Let's find a better name for `setupDataMatrices`.

Renamed it in eb035ffb362490d9aa388dd5fec4e459b03190de to `initializeStorage`.

src/cplscheme/config/CouplingSchemeConfiguration.cpp
BenjaminRodenberg: I started drafting the shift of the ownership of the Waveform into `Data`. This is also a possible way to go, but still needs some refinement. See https://github.com/BenjaminRodenberg/precice/pull/4.

However, I think it automatically solves [this problem](https://github.com/BenjaminRodenberg/precice/pull/1/files#r745679418).

src/cplscheme/BiCouplingScheme.cpp
BenjaminRodenberg: https://github.com/BenjaminRodenberg/precice/pull/4 actually goes some steps into this direction, but entirely removing the `PtrData` is, as explained above, a big step.

BenjaminRodenberg:
I think we can close this PR when #1145 is merged:

* Ownership of Extrapolation is shifted to `CouplingData` in #1145.
* The Waveform is removed in #1145, because without extrapolation the Waveform has no functionality (yet). But #1029 will introduce the Waveform again, now with interpolation only.

Let's keep this PR open until #1145 is merged. Then I will check again and see whether this PR still contains something we want to keep.

BenjaminRodenberg:
Closing the PR, because it is outdated. Refer to #1145.

PR #1136, opened 2021-11-12
Add VTU exporter

fsimonis:
## Main changes of this PR

* Adds a configurable exporter to export to vtu/pvtu files. `<export:vtu ... />`
* Adds a warning when using `<export:vtk />` with parallel participants: 
  ```
  You use the VTK exporter in a parallel participant. Note that this will export as PVTU instead. For consistency, prefer to use "<export:vtu ... />" instead.
  ```

* The legacy vtk exporter is incapable of exporting meshes of parallel participants as the format lacks support for this. This is why the VTU format exists, which allows to export piece `.vtu` files which are then accessible as a whole using a single `.pvtu` file.
* The name `vtk` exporter is misleading as it will silently fallback to exporting VTU files for parallel participants.
* This PR adds the `vtu` exporter, which allows to explicitly request VTU exports for both serial and parallel participants.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/io/config/ExportConfiguration.cpp
+++ b/src/io/config/ExportConfiguration.cpp
@@ -14,7 +14,7 @@ ExportConfiguration::ExportConfiguration(xml::XMLTag &parent)
   XMLTag::Occurrence occ = XMLTag::OCCUR_ARBITRARY;
   {
     XMLTag tag(*this, VALUE_VTK, occ, TAG);
-    tag.setDocumentation("Exports meshes to VTK text files.");
+    tag.setDocumentation("Exports meshes to VTK legacy format files. Parallel participants will use the VTU exporter instead.");
     tags.push_back(tag);
   }
   {
--- a/src/io/config/ExportConfiguration.cpp
+++ b/src/io/config/ExportConfiguration.cpp
@@ -17,6 +17,11 @@ ExportConfiguration::ExportConfiguration(xml::XMLTag &parent)
     tag.setDocumentation("Exports meshes to VTK text files.");
     tags.push_back(tag);
   }
+  {
+    XMLTag tag(*this, VALUE_VTU, occ, TAG);
+    tag.setDocumentation("Exports meshes to VTU files in serial or PVTU files with VTU piece files in parallel.");
+    tags.push_back(tag);
+  }
 
   auto attrLocation = XMLAttribute<std::string>(ATTR_LOCATION, "")
                           .setDocumentation("Directory to export the files to.");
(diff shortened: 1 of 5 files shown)


MakisH (approved):
Nice consistency improvement! Slowly pushing awareness for the right formats.

- I understand the code changes. No objections anywhere, it is quite simple.
- The changelog looks good.
- I have not tried running it (let me know if you want me to).

Remember to document this on the website.

We could also update a few mentions in the tutorials documentation (and the config of the quickstart). But this is also not necessary.

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -549,10 +549,13 @@ void ParticipantConfiguration::finishParticipantConfiguration(
     io::PtrExport exporter;
     if (exportContext.type == VALUE_VTK) {
       if (context.size > 1) {
+        PRECICE_WARN("You use the VTK exporter in a parallel participant. Note that this will export as PVTU instead. For consistency, prefer to use \"<export:vtu ... />\" instead.");
```
MakisH, line 552: Minor language suggestions:

```suggestion
        PRECICE_WARN("You are using the VTK exporter in a parallel participant. Note that this will export as PVTU instead. For consistency, prefer \"<export:vtu ... />\" instead.");
```

uekerman (approved):
Please don't forget to open an issue to clean this up with v3.

src/io/config/ExportConfiguration.cpp
```diff
@@ -14,7 +14,12 @@ ExportConfiguration::ExportConfiguration(xml::XMLTag &parent)
   XMLTag::Occurrence occ = XMLTag::OCCUR_ARBITRARY;
   {
     XMLTag tag(*this, VALUE_VTK, occ, TAG);
-    tag.setDocumentation("Exports meshes to VTK text files.");
+    tag.setDocumentation("Exports meshes to VTK legacy format files. Parallel participants will use the VTU exporter instead.");
+    tags.push_back(tag);
+  }
+  {
+    XMLTag tag(*this, VALUE_VTU, occ, TAG);
+    tag.setDocumentation("Exports meshes to VTU files in serial or PVTU files with VTU piece files in parallel.");
```
DavidSCN, line 22: @fsimonis I just tried this functionality and would like to add the following:
Using the `vtu` exporter in serial does not create a `pvtu` file, but appends the rank number in the `vtu` file. Therefore, it is not possible to select the time-series in paraView, but only each file individually (the rank stays the same), which makes it somewhat useless. I would propose to 
1) either generate a `pvtu` file in serial runs as well 
2) drop the rank index in serial runs so that a time-series can still be visualized
3) use the serial vtk exporter even if a vtu exporter was configured.

fsimonis: Thanks for pointing that out! A fix is coming in #1137

Issue #1135, opened 2021-11-11
CI strategy for warnings


MakisH:
We currently have a few warnings on develop. For example:
- `BiCouplingScheme.cpp` triggers `-Wreorder`
- `SolverInterface.cpp` triggers `-Wsign-compare` and `-Wunused-but-set-variable`
- `ParallelTests.cpp` triggers `-Wunused-variable`

These warnings also appear in the GitHub Actions, but they don't make the pipeline fail. Therefore, relying on the succeding pipeline, we let these slip under the carpet.

At the same time, I also get non-preCICE related warnings locally, e.g., from Boost.

What if we enable `-Werror` in our CI? If this is too strict, we could make an extra, "strict" step with `-Werror` and allow it to fail.

fsimonis:
This has already been attempted in #1018.

Getting the release build to a warning-free state will take some time as there is quite some code which feeds into assertions.
The [maybe_unused](https://en.cppreference.com/w/cpp/language/attributes/maybe_unused) attribute would help here, but it is a C++17 feature.

To minimize the "cost" of this addition, we could add `-Werror` only to the arch debug CIs.
This way we get warnings
* from the newest compilers
* using the newest dependencies
* including the assertion code

MakisH:
I will dig deeper into this.

By the way, I forgot to mention that I was building a Debug build, which still has warnings.

fsimonis:
@MakisH I prototyped a version of the CI in #1138, which is part of the precice repo.
You can use it to _dig deeper_ and try some things out if you like.

> Debug build, which still has warnings.

Sadly this is the case.

MakisH:
Would wrapping external headers in a `#pragma GCC diagnostic` wrapper be an option? First time I hear about this, but suggested here: https://stackoverflow.com/a/46946857/2254346

Also, currently we only get one external warning from Boost about a deprecation. Are we using the latest Boost version? On the one hand, Arch should help with getting the latest versions of dependencies. On the other hand, it would also get the latest versions of compilers, which could generate new deprecation warnings.

fsimonis:
The deprecation is due to an unused header, which is easy to fix. :sweat_smile:

Using pragmas sounds a bit extreme as long as we don't run into real issues.

MakisH:
Then I can imagine that we either:
- make an "allowed to fail" workflow with warnings enabled (I understand that in GitHub Actions this is not currently supported)
[...]

PR #1138, opened 2021-11-19
Add selective Werror to CI

fsimonis:
## Main changes of this PR

This PR adds `-Werror` and enables more warnings in the archlinux CI.
This ensures we are aware of warnings of the newest compilers only.

This PR is primarily a sandbox until we figure out exactly what we need.

#1135


fsimonis:
Superseeded by #1170

PR #1139, opened 2021-11-20
Move constant UNDEFINED_EXTRAPOLATION_ORDER into time::Time

BenjaminRodenberg:
## Main changes of this PR

Creates a class `time::Time` for storing constants on extrapolation.

This class can be used for storing other constants (e.g. `UNDEFINED_INTERPOLATION_ORDER`) at a later point in time. It also helps to further split the packages `cplscheme` and `time`.

Alternatively, we could also create a class `time::Constants`, but I followed the rationale in `cplscheme::CouplingScheme`.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -18,6 +18,7 @@
 #include "mesh/Data.hpp"
 #include "mesh/Mesh.hpp"
 #include "precice/types.hpp"
+#include "time/Time.hpp"
 #include "time/Waveform.hpp"
 #include "utils/EigenHelperFunctions.hpp"
 #include "utils/MasterSlave.hpp"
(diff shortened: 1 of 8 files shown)


uekerman:
To be consistent with other packages in preCICE, `Time.hpp` should be reserved for the abstract interface class of the package. See e.g. [CouplingScheme.hpp](https://github.com/precice/precice/blob/develop/src/cplscheme/CouplingScheme.hpp), [Communication.hpp](https://github.com/precice/precice/blob/develop/src/com/Communication.hpp), or [Mapping.hpp](https://github.com/precice/precice/blob/develop/src/mapping/Mapping.hpp).
`Waveform` would then inherit from this class if one day we have different options how to treat data in time.

Other than this, there is no reason against the refactoring. But I also don't fully understand the motivation yet. Shouldn't this be part of a larger PR that does all the touching of extrapolation?

BenjaminRodenberg:
We can close this PR when #1145 is merged, because extrapolation does not belong to time anymore but to the coupling scheme. If necessary, time will provide a `UNDEFINED_INTERPOLATION_ORDER`, but this is not related to this PR and will be solved in #1029.

PR #1141, opened 2021-11-22
Simplification of constructors and some extra add documentation

ajaust:
## Main changes of this PR

This PR does the following:
- Make constructors of `SolverInterfaceImpl` and `SolverInterface` simpler by setting default parameter for communicator to avoid code duplication.
- Add documentation for `accessorProcessRank` and `accessorCommunicatorSize` in class `SolverInterfaceImpl`. However, this contains some information duplication of documenation of `SolverInterface`.

I stumbled upon some code duplication in the constructors of `SolverInterface` and `SolverInterfaceImpl`. Both classes provide an interface that accepts an explicitly set communicator and without the communicator specified. Instead of actually defining the constructors twice one can set a default parameter for the `communicator` variable/pointer.

The main advantages are:
- Less code and thus less code that can be wrong
- Easier to read classes as they get shorts
- Less code that has to be documented
- Less code and documentation duplication. For the constructor of `SolverInterfaceImpl` there was actually documentation missing that I tried to fix.

If you work on reducing code duplication anyway, please ignore this PR. I found it easy to change. I ran the tests locally and they all passed (Ubuntu 20.04).

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/SolverInterface.cpp
+++ b/src/precice/SolverInterface.cpp
@@ -9,15 +9,6 @@ SolverInterface::SolverInterface(
     const std::string &participantName,
     const std::string &configurationFileName,
     int                solverProcessIndex,
-    int                solverProcessSize)
-    : _impl(new impl::SolverInterfaceImpl(participantName, configurationFileName, solverProcessIndex, solverProcessSize))
-{
-}
-
-SolverInterface::SolverInterface(
-    const std::string &participantName,
-    const std::string &configurationFileName,
-    int                solverProcessIndex,
     int                solverProcessSize,
     void *             communicator)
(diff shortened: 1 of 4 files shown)


fsimonis:
Thanks for the contribution!

Hi, we added the second constructor to keep the binary interface compatible with previous versions of precice.
In retrospective we should have implemented your proposed constructor in version 2.0.0.

So, while I approve of this breaking change, we have to wait until 3.0.0.

fsimonis:
We decided to keep the MPI communicator a separate argument for the very niche use-case.
We simplified the SolverInterfaceImpl using C++17 `std::optional`.

PR #1142, opened 2021-11-22
Make reformatting options more verbose

ajaust:
## Main changes of this PR

- Mention `reformat-all` script in PR template
- `check-format` mentions `reformat-all` script if some files are formatted wrongly.
 
  ```bash
  $ tools/formatting/check-format
  Using binary: clang-format-8
  GNU parallel detected.
  Checking 388 C/C++ files
  100% 388:0=0s tools/testing/communication_dummies/mainB.cpp                                                                                 
  Checking 100 XML files
  100% 100:0=0s src/xml/tests/xmltest-unique-attribute-config.xml                                                                             
  The following files are not formatted correctly:
  src/precice/impl/SolverInterfaceImpl.cpp
  You can automatically reformat all files using the script tools/formatting/format-all
  ```

  Note, that the last line in the output is new.

## Motivation and additional information

I found author's checklist of limited help when it comes to format checking. I ran the `check-format` tool and my files were formatted wrongly, see

```bash
$ tools/formatting/check-format
Using binary: clang-format-8
GNU parallel detected.
Checking 388 C/C++ files
100% 388:0=0s tools/testing/communication_dummies/mainB.cpp                                                                                 
Checking 100 XML files
100% 100:0=0s src/xml/tests/xmltest-unique-attribute-config.xml                                                                             
The following files are not formatted correctly:
src/precice/SolverInterface.hpp
src/precice/impl/SolverInterfaceImpl.hpp
```

However, it did not give me insight regarding what is wrong, nor a patch to apply, nor any hint to the `format-all` script. Sadly, the `README.md` in `tools/formatting` does not help much either.

## Author's checklist

* [ ] I added a changelog file with this PR number in `docs/changelog/` if there are noteworthy changes.
* [x] I ran `tools/formatting/check-format` and everything is formatted correctly.
* [x] I sticked to C++14 features.
* [x] I sticked to CMake version 3.10.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

PR #1143, opened 2021-11-24
Add tools to CMake

fsimonis:
## Main changes of this PR

This PR adds targets for our tools to CMake, which allows to quickly run the most common tools from the build directory.
This is related to @ajaust 's proposed #1142.

Job | Manual | Addition
--- | --- | ---
Generate documentation | `doxygen` | `make doxygen`
Format the codebase | `tools/formatting/format-all` | `make format`
Updating sources | `tools/building/updateSourceFiles.py` | `make sourcesIndex`
Add changelog entry | `tools/building/createChangelog` | `make changelog`

Makes the tooling more straight forward to use and simplifies the author checklist in the PR template.

## Author's checklist

* [ ] I added a changelog entry using `make changelog` if there are noteworthy changes.

## Reviewers' checklist

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -9,8 +9,8 @@ Short rational why preCICE needs this change. If this is already described in an
 
 ## Author's checklist
 
-* [ ] I added a changelog file with this PR number in `docs/changelog/` if there are noteworthy changes.
-* [ ] I ran `tools/formatting/format-all` to ensure everything is formatted correctly.
+* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
+* [ ] I ran `make format` to ensure everything is formatted correctly.
 * [ ] I sticked to C++14 features.
 * [ ] I sticked to CMake version 3.10.
 * [ ] I squashed / am about to squash all commits that should be seen as one.
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -576,7 +576,7 @@ endif()
 #
 
 add_custom_target(
-  docs
+  doxygen
   COMMAND doxygen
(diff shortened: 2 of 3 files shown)


uekerman:
I like. I only have doubts concerning the naming:
* `make update` could mean anything
* `make docs` sounds too much like it would generate our user documentation. Why not `make doxygen`?

MakisH:
Great idea!



> I like. I only have doubts concerning the naming:
> 
>     * `make update` could mean anything
> 
>     * `make docs` sounds too much like it would generate our user documentation. Why not `make doxygen`?

I have the same concerns. Suggestions:

- `make cmakesources`? (I had to look up again what this does to understand the "update")
- `make doxygen` as well.

Would it help to have a two-level naming scheme, instead? Something like `make release:changelog`, `make release:sources`, etc. Probably `release` is not the right category here, but maybe it helps spark another idea.

fsimonis:
`make doxygen` :heavy_check_mark: 

`make update` and `make cmakesources` read like they "must by run".
Maybe something like `refreshSources`, `discoverSources`, `findSources`, `reindexSources`, or `reindex`.

MakisH:
> `make update` and `make cmakesources` read like they "must by run".

Sounds like we should find a consistent pattern: `make verb` or `make object`. I prefer the latter.

> Maybe something like `refreshSources`, `discoverSources`, `findSources`, `reindexSources`, or `reindex`.

Aren't these verbs? I like all of them, I think the "reindex" is the most accurate term, but maybe not the most intuitive / first thing that comes to mind.

For objects, how about `sourcesIndex`? Maybe others have better suggestions.

fsimonis:
I agree with using objects and `sourceIndex` is exactly what this is.

PR #1144, opened 2021-11-30
Add CSV exporter

fsimonis:
## Main changes of this PR

This PR adds a CSV exporter, which allows to export the vertex information to a semicolon-separated values format.

Intended purpose is:
* provide an accessible export format
* usable without Paraview
* easy to import into speadsheet applications (quick visualisation of data)
* easy to check (may be useful for integration tests)
* more natural fit for applications with pseudo-meshes

The exporter
* does not export connectivity information
* saves 1 file per rank
* uses the same naming scheme as the other exporters

To merge and process csv of a parallel participant, one can use pandas:
```py
import pandas
import glob

name = "A-ExporterTwo.dt1"
df = pandas.concat([pandas.read_csv(name, sep=";") for name in glob.glob(f"{name}_*.csv")], ignore_index=True)
```

To import timesteps as well
```py
import pandas
import glob
import re

l = [  (re.search("dt(\d+)_", s).group(1), s) for s in  glob.glob(f"{n}.dt*_*.csv")]
p = pandas.concat([pandas.read_csv(file, sep=";").assign(dt=dt) for dt, file in l], ignore_index=True)
```

## Author's checklist

* [x] I ran `tools/formatting/format-all` to ensure everything is formatted correctly.

## Reviewers' checklist

### DIFF
--- a/docs/changelog/1144.md
+++ b/docs/changelog/1144.md
@@ -1,0 +1,1 @@
+- Added export to CSV using `<export:csv />`.
--- a/src/io/ExportCSV.cpp
+++ b/src/io/ExportCSV.cpp
@@ -34,8 +34,10 @@ void ExportCSV::doExport(
 
   // Construct full filename
   std::string filename{name};
+  int         rank{0};
   if (utils::MasterSlave::isParallel()) {
-    filename.append("_").append(std::to_string(utils::MasterSlave::getRank()));
+    rank = utils::MasterSlave::getRank();
+    filename.append("_").append(std::to_string(rank));
   }
   filename.append(".csv");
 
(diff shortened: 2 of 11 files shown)


KyleDavisSA:
The CSV output works well, however if you set a directory using ` <export:csv directory="outputFolder" />`, and you have not created the folder, then the results are not saved anywhere. This is not true when using VTK's. Using  ` <export:vtk directory="outputFolder" />`, the folder is created automatically.

KyleDavisSA (commented):
CSV output looks good. Only comment is the output directory needs to be created before running preCICE.

fsimonis:
@KyleDavisSA Thanks for testing the feature. I fixed the location creation.
Can you pull, rebuild and check if this fixes the issue for you?

KyleDavisSA (approved):
Works as expected.

PR #1145, opened 2021-11-30
Refactor how extrapolation is performed

BenjaminRodenberg:
## Main changes of this PR

* Create Extrapolation class in `cplscheme/impl` for functionality.
* Move all current functionality of Waveform into Extrapolation. Waveform is reserved for interpolation. See #1029 
* Store Extrapolation in CouplingData.

We want to implement extrapolation and interpolation as described in #1029. This PR adds some changes from #1029 directly to develop and simplifies #1029.

## Author's checklist

* [ ] I ran `tools/formatting/format-all` to ensure everything is formatted correctly.

## Reviewers' checklist


src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -257,12 +253,12 @@ void BaseCouplingScheme::advance()
...
 }
 
-void BaseCouplingScheme::storeDataInWaveforms()
+void BaseCouplingScheme::storeExtrapolationData()
 {
   PRECICE_TRACE(_timeWindows);
   for (DataMap::value_type &pair : _allData) {
     PRECICE_DEBUG("Store data: {}", pair.first);
-    _waveforms[pair.first]->store(pair.second->values());
+    pair.second->storeDataInExtrapolation();
```
uekerman, line 261:
```suggestion
    pair.second->storeExtrapolationData();
```
?
If we really use two different names for the two different functions, the name should explain the difference.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -450,15 +445,16 @@ void BaseCouplingScheme::checkCompletenessRequiredActions()
   }
 }
 
-void BaseCouplingScheme::setupDataMatrices()
+void BaseCouplingScheme::initializeStorage()
```
uekerman, line 448:
```suggestion
void BaseCouplingScheme::initializeStorages()
```
?

src/cplscheme/CouplingData.hpp
```diff
@@ -45,14 +47,24 @@ class CouplingData {
...
 private:
   /**
    * @brief Default constructor, not to be used!
    *
    * Necessary when compiler creates template code for std::map::operator[].
    */
   CouplingData()
-      : requiresInitialization(false)
+      : requiresInitialization(false),
+        _extrapolation(0)
```
uekerman, line 67: Wouldn't it be better to use `UNDEFINED_EXTRAPOLATION_ORDER` here as well?

src/cplscheme/impl/Extrapolation.cpp
```diff
@@ -1,70 +1,73 @@
-#include "time/Waveform.hpp"
+#include "Extrapolation.hpp"
 #include <algorithm>
+#include "cplscheme/CouplingScheme.hpp"
 #include "logging/LogMacros.hpp"
 #include "utils/EigenHelperFunctions.hpp"
 
 namespace precice {
-namespace time {
+namespace cplscheme {
```
uekerman, line 8: IIRC we normally also use an `impl` namespace here.

uekerman (approved):
Looks good. Only minor things below. Feel free to directly merge if fixed and no more questions.

src/cplscheme/CouplingData.hpp
BenjaminRodenberg: Yes. I will implement this :+1:

src/cplscheme/impl/Extrapolation.cpp
BenjaminRodenberg: Yes. I will add this namespace.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: Ok.

PR #1147, opened 2021-12-08
Remove CPACK_PACKAGE_MAINTAINER property

ajaust:
The CPACK_PACKAGE_MAINTAINER property of CPack does not exist, see https://cmake.org/cmake/help/latest/module/CPack.html.

There exists a `CPACK_DEBIAN_PACKAGE_MAINTAINER` property (https://cmake.org/cmake/help/latest/cpack_gen/deb.html#variable:CPACK_DEBIAN_PACKAGE_MAINTAINER) which is set to the value of CPACK_PACKAGE_CONTACT by default.

## Author's checklist

* [ ] I ran `tools/formatting/format-all` to ensure everything is formatted correctly.

## Reviewers' checklist


fsimonis:
Thanks!

PR #1150, opened 2021-12-10
Add release pipeline

### LINKED ISSUE #319: Automate package creation
Now that we have Debian packages, it would be good to automate the process, so that e.g. [Travis could create them](https://travis.debian.net/).

Moreover/alternatively, we could use [Open Build Service](https://openbuildservice.org/help/), or something similar, to produce packages for other distributions as well.

fsimonis:
## Main changes of this PR

This PR adds a release pipeline for Ubuntu package creation and merges it with the dockerfile workflow.
It works as follows:

### Triggers
* Trigger is a pushed tag `vN.N.N`
* Can also be manually trigger

### Steps
1. Create a draft release with the same name as the tag, following our current naming scheme.
2. Build packages:
   1. Build the `Dockerfile` and push the tags `latest` as well as the git tag without the v (`v2.3.4 -> precice/precice:2.3.4`)
   2. Build Ubuntu packages (tar + deb) and upload them to the draft release 

Less manual work and fewer mysterious python scripts for package generation.

Closes #319 

Example pipleline: https://github.com/fsimonis/precice/actions/runs/1564205970


.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
...
+
+  package-ubuntu:
+    needs: release-draft
+    name: 'Package Ubuntu ${{ matrix.UBUNTU }}'
+    runs-on: ubuntu-latest
+    container: 'ubuntu:${{ matrix.UBUNTU }}'
+    strategy:
+      fail-fast: true
+      matrix:
+        UBUNTU: ["18.04", "20.04", "20.10", "21.04"]
```
MakisH, line 27: The 20.10 is not supported anymore and the 21.04 is going to reach End-Of-Life on January 20. This [table](https://en.wikipedia.org/wiki/Ubuntu_version_history#Table_of_versions) is very helpful.

General guideline:
- LTS releases (18.04, 20.04, 22.04, ...) are supported for five years
- point releases (20.10, 21.04, 21.10, 22.10, ...) are supported for nine months
- releases do overlap over time and we should support them if easy (e.g., even though the 21.10 is out, I am still on 21.04, as I still have enough time to upgrade)

```suggestion
        UBUNTU: ["18.04", "20.04", "21.04", "21.10"]
```

.github/workflows/release.yml
```diff
@@ -0,0 +1,101 @@
...
+          file: build/libprecice*.deb*
+          file_glob: true
+          tag: ${{ github.ref }}
+      - name: Upload tar package
+        uses: svenstaro/upload-release-action@v2
+        with:
+          repo_token: ${{ secrets.GITHUB_TOKEN }}
+          file: build/libprecice*.tar*
+          file_glob: true
+          tag: ${{ github.ref }}
```
MakisH, line 72: In the beginning, we added also the `.tar.gz` files as they can be useful to people that don't use Ubuntu, but something else with compatible dependencies. I have a hard time defining the exact target group and [who these few people can be](https://utilify.me/github-release-stats).

As the current list of packages is very long and can be confusing/daunting to the new user, what if we ommit the `.tar.gz` (and the respective hashes) in the next release, and find out if anyone complains?

The [example draft release](https://github.com/fsimonis/precice/releases/tag/v2.666.0) lists 18 different assets (including the code archives). Removing the `.tar.gz` would drop this to 10 files. Removing the 20.10, in addition, would drop this to 8 assets (three Debian packages with their hashes and the two standard code archives).

```suggestion
          cmake -DCMAKE_INSTALL_PREFIX=/usr -DCMAKE_BUILD_TYPE=Release -DCPACK_GENERATOR="DEB" -DPRECICE_PETScMapping=${{ env.PETSc }} -DBUILD_TESTING=OFF -DBUILD_SHARED_LIBS=ON ..
      - name: Compile
        working-directory: build
        run: make -j $(nproc)
      - name: Create packages
        working-directory: build
        run: cpack
      - name: Upload debian package
        uses: svenstaro/upload-release-action@v2
        with:
          repo_token: ${{ secrets.GITHUB_TOKEN }}
          file: build/libprecice*.deb*
          file_glob: true
          tag: ${{ github.ref }}
```

MakisH (commented):
Looks very nice and I am very happy that we will not have to do this manually anymore in mystical and cryptic ways. :bat:

What do we do when we would like to manually generate new packages for new Ubuntu versions? Manually trigger the pipeline? How would that work since there will already be a release? Do we maybe need a separate workflow for that?

.github/workflows/release.yml
fsimonis: Sure, let's try it out.

fsimonis:
@MakisH I added a workflow which can be used to request a package for any Ubuntu version.
Note that this will always install `petsc-dev`, so it isn't compatible with Ubuntu 18.04.

MakisH:
Wait... we can have inputs in GitHub Actions? :exploding_head: That sounds awesome! But how would one trigger it, then? We will probably need some documentation.

> Note that this will always install petsc-dev, so it isn't compatible with Ubuntu 18.04.

This is perfectly ok.

fsimonis:
@MakisH You can add inputs to `workflow_dispatch` events. These have to have to be triggered manually in the Actions tab.

MakisH (approved):
Thanks for adding documentation for every workflow! This looks great now!

PR #1160, opened 2022-01-01
Use Eventtimings submodule 

### LINKED ISSUE #419: Turn EventTimings into a CMake submodule
The [EventTimings project](https://github.com/precice/EventTimings) is a fully fledged CMake project.
We should consider refactoring it out of preCICE by placing it into a submodule.

mohamad-altaweel:
## Main changes of this PR

removed EventTimings files from utils and added EventTimings as a subdirectory. Resolves #419

The EventTimings project is a fully fledged CMake project. Hences, We could refactor it out of preCICE by placing it into a submodule.

## Author's checklist

* [X] I ran `make format` to ensure everything is formatted correctly.
* [X] I sticked to C++14 features.
* [X] I sticked to CMake version 3.10.

## Reviewers' checklist


fsimonis:
Hi and thanks for tackling this!

One big open point is how we deal with submodules in this repository.
This change currently breaks a lot of documentation, the CI and possibly even our Spack package.
Main issue is that source downloads on github don't include the submodules.

@MakisH Didn't we once discuss this thoroughly?

MakisH:
> One big open point is how we deal with submodules in this repository. This change currently breaks a lot of documentation, the CI and possibly even our Spack package. Main issue is that source downloads on github don't include the submodules.

I think this is a good change, but I understand that this GitHub limitation will actually hit us.(you can test it by downloading the [website repository archive](https://github.com/precice/precice.github.io/archive/refs/heads/master.zip), which does not include the tutorials or the OpenFOAM adapter archives).

In principle, we can fix all these. I see two options:

- Merge this in preCICE v3 (where users will bump anyway into breaking changes).
- Merge this now and at the same time commit the content of the imported modules (I think this is possible). This way, we can keep all the code here (as it already is), not break anything, and manually update the content whenever we want. But still, this would not be a clean solution, as the content and the actual module may quickly go out of synch. Also, if we don't actually use the module, we don't have a good reason to implement it.

> @MakisH Didn't we once discuss this thoroughly?

I vaguely remember some discussing something similar a long time ago. In any case, https://github.com/precice/precice/pull/990#issuecomment-817875256 is related.

MakisH:
I added this to the v3 milestone, but we should look at it in a larger context. I understand we mainly are not sure if https://github.com/precice/precice/issues/419 is something we want, so let's first decide there.

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -11,7 +11,7 @@
 #include "precice/types.hpp"
 #include "query/Index.hpp"
 #include "utils/EigenHelperFunctions.hpp"
-#include "utils/Event.hpp"
+#include "../EventTimings/include/EventTimings/Event.hpp"
```
MakisH, line 14: It feels strange that we need to set a specific path. We should rather add this to the include path.

(same in the `.cpp` file)

MakisH (commented):
Regarding the specific contribution, I understand that this essentially:
- adds https://github.com/precice/EventTimings as a submodule, essentially dropping the need to duplicate the files that define the `Event` class.
- replaces `precice::utils::Event` by `EventTimings::Event`
- sources these from `#include "EventTimings/Event.hpp"`, instead of `#include "utils/Event.hpp"`
- removes the files `src/utils/Event.[cpp|hpp]`, and `src/utils/EventUtils.[cpp|hpp]`.

fsimonis:
Closed in favour of the ongoing simplification #1410 #1419

PR #1163, opened 2022-01-13
Add the support-radius option for Gaussian RBF

### LINKED ISSUE #765: Calculation of shape-parameter Gaussian RBF from support-radius

DavidSCN:
## Main changes of this PR
Adds the support-radius as an option to the Gaussian RBF functions. In case both, the support-radius and shape-parameter are defined, the support radius is applied. I still need to do some testing, but would prefer some feedback in case the changes are refused.

Closes #765 

## Author's checklist

* [x] I tested the changes and wrote a test

## Reviewers' checklist


src/mapping/config/MappingConfiguration.cpp
```diff
@@ -347,6 +350,10 @@ MappingConfiguration::ConfiguredMapping MappingConfiguration::createMapping(
       configuredMapping.mapping = PtrMapping(
           new RadialBasisFctMapping<VolumeSplines>(constraintValue, dimensions, VolumeSplines(), xDead, yDead, zDead));
     } else if (type == VALUE_RBF_GAUSSIAN) {
+      if (!std::isinf(supportRadius)) {
+        // Compute shape parameter from the support radius
+        shapeParameter = std::sqrt(-std::log(Gaussian::cutoffThreshold)) / supportRadius;
+      }
```
DavidSCN, line 356: I could have shifted this over into the `Gaussian` class. However, I decided against this as the class itself can (in theory) still get two independent values (shape-parameter and support-radius).

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -81,8 +81,11 @@ MappingConfiguration::MappingConfiguration(
   }
   {
     XMLTag tag(*this, VALUE_RBF_GAUSSIAN, occ, TAG);
-    tag.setDocumentation("Local radial-basis-function mapping based on the Gaussian RBF with a cut-off threshold.");
-    tag.addAttribute(attrShapeParam);
+    tag.setDocumentation("Local radial-basis-function mapping based on the Gaussian RBF using a cut-off threshold.");
+    tag.addAttribute(makeXMLAttribute<double>(ATTR_SHAPE_PARAM, 0.)
+                         .setDocumentation("Specific shape parameter for RBF basis function."));
+    tag.addAttribute(makeXMLAttribute(ATTR_SUPPORT_RADIUS, std::numeric_limits<double>::infinity())
+                         .setDocumentation("Support radius of each RBF basis function (global choice)."));
```
uekerman, line 88: Why creating the attributes here? The other cases are handled differently.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -347,6 +350,10 @@ MappingConfiguration::ConfiguredMapping MappingConfiguration::createMapping(
       configuredMapping.mapping = PtrMapping(
           new RadialBasisFctMapping<VolumeSplines>(constraintValue, dimensions, VolumeSplines(), xDead, yDead, zDead));
     } else if (type == VALUE_RBF_GAUSSIAN) {
+      if (!std::isinf(supportRadius)) {
+        // Compute shape parameter from the support radius
+        shapeParameter = std::sqrt(-std::log(Gaussian::cutoffThreshold)) / supportRadius;
```
uekerman, line 355: Are you sure that there is not a "2" missing in front of the support radius?

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -388,6 +395,10 @@ MappingConfiguration::ConfiguredMapping MappingConfiguration::createMapping(
           new PetRadialBasisFctMapping<VolumeSplines>(constraintValue, dimensions, VolumeSplines(),
                                                       xDead, yDead, zDead, solverRtol, polynomial, preallocation));
     } else if (type == VALUE_RBF_GAUSSIAN) {
+      if (!std::isinf(supportRadius)) {
+        // Compute shape parameter from the support radius
+        shapeParameter = std::sqrt(-std::log(Gaussian::cutoffThreshold)) / supportRadius;
```
uekerman, line 400: Are you sure that there is not a "2" missing in front of the support radius?

src/mapping/config/MappingConfiguration.cpp
uekerman: I would also handle it in the configuration.

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -149,12 +149,12 @@ class Gaussian : public CompactSupportBase {
       return std::exp(-std::pow(_shape * radius, 2.0)) - _deltaY;
   }
 
+  /// Below that value the function is supposed to be zero. Defines the support radius if not explicitely given
+  static constexpr double cutoffThreshold = 1e-9;
+
```
uekerman, line 154: Mmh, static. Couldn't we treat it like `NUMERICAL_ZERO_DIFFERENCE` in `math/differences.hpp`?

uekerman (commented):
Thanks for pushing this :+1:

I could imagine that the way you use default values in the config is a bit error prone as they are no real default values, but "uninitialized" values.

>  In case both, the support-radius and shape-parameter are defined, the support radius is applied

I think, we should not allow this. Minimum a warning, but we could also throw an error.

src/mapping/config/MappingConfiguration.cpp
DavidSCN: We need to set default values, otheriwse, we break compatibility.

src/mapping/config/MappingConfiguration.cpp
DavidSCN: I copied the formula from the Gaussian class, so yes, pretty sure. Do you have a formula in mind?

src/mapping/impl/BasisFunctions.hpp
DavidSCN: Note that this is a `static` class member and not a static variable, which is something different. Also, the `cutoffThreshold` is something different then `NUMERICAL_ZERO_DIFFERENCE`. I still think that a member variable is the way to go here, but I'm open for discussions.

src/mapping/config/MappingConfiguration.cpp
uekerman: I looked at the documentation here: https://precice.org/tooling-rbf-shape.html
There it is also a bit ambiguous. What does "cover by the support radius" mean? A radius cannot cover sth, only the support :grin:
If you are sure, I am sure.

src/mapping/config/MappingConfiguration.cpp
uekerman: That's connected to what I wrote above:

> I could imagine that the way you use default values in the config is a bit error prone as they are no real default values, but "uninitialized" values.

We could instead check whether the user defined the attributes or not, right? And then throw errors if none or both are defined.

DavidSCN:
Conclusion: add selection option.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -81,8 +81,13 @@ MappingConfiguration::MappingConfiguration(
...
     XMLTag tag(*this, VALUE_RBF_GAUSSIAN, occ, TAG);
-    tag.setDocumentation("Local radial-basis-function mapping based on the Gaussian RBF with a cut-off threshold.");
-    tag.addAttribute(attrShapeParam);
+    tag.setDocumentation("Local radial-basis-function mapping based on the Gaussian RBF using a cut-off threshold.");
+    tag.addAttribute(makeXMLAttribute<double>(ATTR_SHAPE_PARAM, 0.)
+                         .setDocumentation("Specific shape parameter for RBF basis function."));
+    tag.addAttribute(makeXMLAttribute(ATTR_SUPPORT_RADIUS, 0.)
+                         .setDocumentation("Support radius of each RBF basis function (global choice)."));
+    tag.addAttribute(makeXMLAttribute<bool>(ATTR_USE_SHAPE_PARA, true)
+                         .setDocumentation("Select between the shape-parameter and the support-radius in order to determine the local cut-off of the Gaussian RBF."));
```
fsimonis, line 90: The mapping would be easier to use if the configuration requires one of the two to be passed instead of having to specify which one to use.

We could set both to `NaN` by default and enforce that only one of the two is specified.
This should be easy to check.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -160,15 +165,16 @@ void MappingConfiguration::xmlTagCallback(
...
-    double        solverRtol     = 1e-9;
+    std::string   dir                       = tag.getStringAttributeValue(ATTR_DIRECTION);
+    std::string   fromMesh                  = tag.getStringAttributeValue(ATTR_FROM);
+    std::string   toMesh                    = tag.getStringAttributeValue(ATTR_TO);
+    std::string   type                      = tag.getName();
+    std::string   constraint                = tag.getStringAttributeValue(ATTR_CONSTRAINT);
+    Timing        timing                    = getTiming(tag.getStringAttributeValue(ATTR_TIMING));
+    double        shapeParameter            = 0.0;
+    double        supportRadius             = 0.0;
+    bool          gaussianUseShapeParameter = true;
```
fsimonis, line 176: This will become easier to handle later on if you use a struct + enum to encode the RBF parameter;

```cpp
struct RBFParameter {

  enum struct Type {
    ShapeParameter,
    SupportRadius
  };

  Type type;
  double value;
}

// ... later in the RBF section

PREICICE_CHECK(isfinite(shapeParameter) ^ isfinite(supportRadius), "None set");
PREICICE_CHECK(isfinite(shapeParameter) && isfinite(supportRadius), "Both set");

RBFParameter param;
if (isfinite(shapeParameter)) {
  param.type = RBFParameter::Type::ShapeParameter;
  param.value = shapeParameter;
else {
  param.type = RBFParameter::Type::ShapeParameter;
  param.value = shapeParameter;
}

createMapping( ... param);

```

This will allow to simplify the createMapping in the future.

src/mapping/impl/BasisFunctions.hpp
fsimonis: Shouldn't this be configurable by the user?

fsimonis (changes requested):

src/mapping/impl/BasisFunctions.hpp
DavidSCN: The cutoff threshold is and was constant just as before. Making it configurable might be desirable, but not the purpose of this PR.

src/mapping/config/MappingConfiguration.cpp
DavidSCN: Alright, I changed according to this approach.

DavidSCN:
I changed now the default values to be `NaN` as proposed by @fsimonis and deduce automatically which one has been set by the user.

uekerman:
The suggestions from @fsimonis and mine went a bit in opposite directions. I am also OK with deducing the user's wish automatically. It would still be good to throw a warning if both are configured, however.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -324,6 +327,18 @@ MappingConfiguration::ConfiguredMapping MappingConfiguration::createMapping(
   bool    usePETSc        = false;
   RBFType rbfType         = RBFType::EIGEN;
 
+  RBFParameter param;
+  // Check valid combinations for the Gaussian RBF input
+  if (type == VALUE_RBF_GAUSSIAN) {
+    const bool exactlyOneSet = (std::isfinite(supportRadius) && !std::isfinite(shapeParameter)) ||
+                               (std::isfinite(shapeParameter) && !std::isfinite(supportRadius));
```
DavidSCN, line 334: > The suggestions from @fsimonis and mine went a bit in opposite directions. I am also OK with deducing the user's wish automatically. It would still be good to throw a warning if both are configured, however.

@uekerman We safeguard in a strict way against specifying both.

uekerman: :+1:

uekerman (commented):
Could we add a minimal integration test using the new option?
Other than this, no additional comments from my side.

DavidSCN:
Maybe open follow-up question: Would it make the `extras/rbfShape` script obsolete?

uekerman:
> Maybe open follow-up question: Would it make the extras/rbfShape script obsolete?

Yes, in principle it does IMO.

src/mapping/impl/BasisFunctions.hpp
fsimonis: I opened an issue https://github.com/precice/precice/issues/1265

src/mapping/config/MappingConfiguration.cpp
fsimonis: Looks like we had a misunderstanding.
The method I propose above encapsulates the kind (shape-parameter or support-radius) and the value into a single type.
This allows to replace both parameters `double shapeParameter` and `double supportRadius` by a single parameter of type `RBFParamter`. This then saves additional work in `createMapping`.

As it stands, you don't need the `struct RBFParamter` at all, you can simply replace it by the `enum struct Type`.
Also, you don't need the member `_rbfParameter` as you already deduce this information in the function where it is needed.

fsimonis:
> The suggestions from @fsimonis and mine went a bit in opposite directions. I am also OK with deducing the user's wish automatically. It would still be good to throw a warning if both are configured, however.

@uekerman Not sure what you mean? In what way do they differ?

src/mapping/config/MappingConfiguration.cpp
DavidSCN: So you propose to store this information for all rbf data mappings within this struct?

DavidSCN: I removed now the member variable `_rbfParameter` and the shape parameter as well as the support radius from the function arguments. They were now replaced by the `RBFParameter`. Is this now what you had in mind?

fsimonis: Exactly! This monstrosity of a function looks a bit cleaner now. :tada:

docs/changelog/1163.md
```diff
@@ -0,0 +1 @@
+- Added the 'support-radius' as an additional configuration option to the Gaussian RBF mapping configuraiton [#1163](https://github.com/precice/precice/pull/1163).
```
uekerman, line 1:
```suggestion
- Added the 'support-radius' as an additional configuration option to the Gaussian RBF mapping configuration [#1163](https://github.com/precice/precice/pull/1163).
```

tests/serial/mapping-rbf-gaussian/helpers.cpp
```diff
@@ -0,0 +1,89 @@
...
+  Vector3d coordOneC{1.0, 1.0, z};
+  Vector3d coordOneD{0.0, 1.0, z};
+  double   valOneA = 1.0;
+  double   valOneB = 3.0;
+  double   valOneC = 5.0;
+  double   valOneD = 7.0;
+
+  // MeshTwo
+  Vector3d coordTwoA{0.0, 0.0, z}; // Maps to vertex A
+  Vector3d coordTwoB{0.0, 0.5, z}; // Maps to edge AD
```
uekerman, line 28: Comments here sounds like this would be a NP mapping?
Are the expected values really 1.0 and 4.0 sharp?

uekerman: ah, ok, but then the test does not really test the support radius, right?

DavidSCN: No, it tests the configuration.

uekerman: But the configuration is not tested in the sense that we actually test that the correct thing is configured. Instead of a proper mapping one could also do white-box testing, but then this test should probably become a unit test instead.

docs/changelog/1163.md
```diff
@@ -0,0 +1 @@
+- Added the 'support-radius' as an additional configuration option to the Gaussian RBF mapping configuration [#1163](https://github.com/precice/precice/pull/1163).
```
fsimonis, line 1: Upon common request, the release script now adds the PR number to every changelog entry. So you don't need to do this manually.

```suggestion
- Added the 'support-radius' as an additional configuration option to the Gaussian RBF mapping configuration.
```

DavidSCN:
I added now more comprehensive tests. However, the test fails when using our eigen based implementation (no PETSc) on ubuntu 2004 (only) as the tolerance of the mapped values is exceeded by a lot. I have no idea what to do about it and I cannot reproduce this problem locally. Only reasonable explanation from my perspective would be the eigen version. I still want to point out that the deviation is the same for the support radius and the shape parameter, i.e., the configuration works as expected.

uekerman:
Let's open an issue on this and merge here.

> I still want to point out that the deviation is the same for the support radius and the shape parameter, i.e., the configuration works as expected.

:+1:

fsimonis:
[9ab7ddf](https://github.com/precice/precice/pull/1163/commits/9ab7ddf69431abb6ce579a994fc05bc73cb9879c) works _for me_ with and without PETSc enabled using eigen 3.4.0. Ubuntu 20.04 LTS uses [eigen 3.3.7](https://ubuntu.pkgs.org/20.04/ubuntu-universe-arm64/libeigen3-dev_3.3.7-2_all.deb.html).

fsimonis:
The tests `Integration/Serial/MappingRbfGaussian/{GaussianShapeParameter, GaussianSupportRadius}` fail with Eigen 3.3.7 due to tolerance issues.


```
check valueA == expectedValTwoA has failed [0.99999999999999645 != 1.0000000014191541].
Relative difference exceeds tolerance [1.41916e-09 > 1e-09]
```

So, I reduced the tolerance of the problematic check down to 1e-08.

DavidSCN:
> So, I reduced the tolerance of the problematic check down to 1e-08.

I mean it doesn't resolve or disables the test, but probably sufficient for now. Do we still want an issue on this phenomenon?

Issue #1162, opened 2022-01-12
Wrong polynomial configuration for Eigen QR-based RBF mapping


DavidSCN:
**Describe the problem**
For all RBF mappings, we have in preCICE the configuration option of the `polynomial` which can be `on`, `off` and `separate`. It is known, that the separate treatment of the polynomial delivers by far the best results (have a look at Florian's thesis or the reference paper). The integrated polynomial could even run into convergence issues in situations where the separated polynomial can easily handle the system. However, our pseudo-parallel eigen3 based QR-RBF implementation only implements the integrated polynomial version (`polynomial=on`). Therefore, setting the `polynomial=separate` (or relying on the default) has no effect and does not give an RBF mapping with a separate polynomial treatment.

There are two situations where one might run into this problem

1. preCICE was build without PETSc leading to the QR-RBF fallback
2. `use-qr-decomposition="true"` was configured

**Expected behaviour**
We should either throw an error if one of the described scenarios occurs or implement the separated polynomial for the QR-RBF. The simple solution of throwing an error comes with the problem of the default configuration being `polynomial=separate`, i.e., if PETSc is not installed the user needs to change the polynomial configuration in any RBF mapping (and needs to change it back if PETSc is installed later on).

uekerman:
I would say, let's implement the missing variants. Should not be too complicated.

PR #1166, opened 2022-01-22
Add separate polynomial handling for Eigen RBF 

DavidSCN:
## Main changes of this PR
... as described in #1162 

I still need to add some tests. A rather naive idea would be to reconfigure some of the existing tests to run with a separate polynomial, as we test all basis functions in the same way for the same system.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mapping/RadialBasisFctMapping.hpp
+++ b/src/mapping/RadialBasisFctMapping.hpp
@@ -81,11 +81,14 @@ private:
   /// QR decomposed system matrix. Evaluated basis function on the input mesh
   Eigen::ColPivHouseholderQR<Eigen::MatrixXd> _qrMatrixC;
 
+  /// QR decomposed system matrix. Evaluated basis function on the input mesh
+  Eigen::ColPivHouseholderQR<Eigen::MatrixXd> _qrMatrixQ;
+
   /// Vandermonde Matrix for linear polynomial, constructed from vertices of the input mesh
-  // Eigen::MatrixXd _matrixQ;
+  Eigen::MatrixXd _matrixQ;
 
   /// Coordinates of the output mesh to evaluate the separated polynomial
-  // Eigen::MatrixXd _matrixV;
+  Eigen::MatrixXd _matrixV;
 
   /// true if the mapping along some axis should be ignored
   std::vector<bool> _deadAxis;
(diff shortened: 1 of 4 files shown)


src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -123,7 +138,41 @@ RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::RadialBasisFctMapping(
     setInputRequirement(Mapping::MeshRequirement::VERTEX);
     setOutputRequirement(Mapping::MeshRequirement::VERTEX);
   }
+
+  PRECICE_CHECK(polynomial == Polynomial::ON || polynomial == Polynomial::SEPARATE, "RBF mappings without a polynomial ( polynomial = \"off\") are only implemented using PETSc. Please install preCICE with PETSc or remove the use-qr-decomposition=\"true\" tag in your configuration file.");
```
uekerman, line 142: Why this restriction? Shouldn't this case be very simple to add?

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -123,7 +138,41 @@ RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::RadialBasisFctMapping(
...
+      deadDimensions += 1;
+  }
+  polyparams    = (_polynomial == Polynomial::ON) ? 1 + dimensions - deadDimensions : 0;
+  sepPolyparams = (_polynomial == Polynomial::SEPARATE) ? 1 + dimensions - deadDimensions : 0;
+}
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::setDeadAxis(bool xDead, bool yDead, bool zDead)
+{
+  // TODO: Remove duplication here with PETSc based implementation
```
uekerman, line 158: This could be done by a common abstract basis class. Or do you have another solution in mind?
These "TODOs" in the code get always easily lost. Better to open an issue (here "good first issue") and remove the TODO here.

uekerman (commented):
Sorry for the very slow reply here.
I did not check every line, but the overall changes make sense.
As mentioned above, still needs tests. And a changelog entry.

DavidSCN:
Superseded by #1335

PR #1167, opened 2022-01-27
Enable boost unit test help messages

DavidSCN:
## Main changes of this PR

Enables user and developer interaction with the `testprecice` executable by making the exception conditional on the number of provided command line arguments. 

As just learned from @ajaust, the boost unit test framework has awesome stand-alone features. One of my favorite ones is probably `tesprecice --list_content` in order to get an overview of all available tests, because I always need to search for (nested) test names when I want to execute them selectively. You can also use `testprecice --help` in order to get a full picture of all available options of our test executable. The help list is additionally much more comprehensive than our developer documentation with respect to configurable options.
However, in our current test suite, we can't use these features, since we immediately exit if our testsuite is not executed on four ranks. The main purpose of this exception is to prevent (unknowing) users from running the tests in serial and running into failures. This behavior would still be caught by counting the input arguments, so that `./testprecice` (which doesn't make sense) results in an error, but `testprecice --help` works.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/testing/main.cpp
+++ b/src/testing/main.cpp
@@ -123,9 +123,9 @@ int main(int argc, char *argv[])
   const auto size = utils::Parallel::current()->size();
   logging::setMPIRank(rank);
 
-  if (size < 4) {
+  if (size < 4 && argc < 2) {
     if (rank == 0) {
-      std::cerr << "ERROR: The tests require at least 4 MPI processes.\n";
+      std::cerr << "ERROR: The tests require at least 4 MPI processes. Please use \"mpirun -np 4 ./tesprecice\" or \"ctest\" to run the full testsuite. \n";
     }
     return 2;
   }
--- a/src/testing/main.cpp
+++ b/src/testing/main.cpp
@@ -125,7 +125,7 @@ int main(int argc, char *argv[])
 
   if (size < 4 && argc < 2) {
     if (rank == 0) {
-      std::cerr << "ERROR: The tests require at least 4 MPI processes. Please use \"mpirun -np 4 ./tesprecice\" or \"ctest\" to run the full testsuite. \n";
+      std::cerr << "ERROR: The tests require at least 4 MPI processes. Please use \"mpirun -np 4 ./testprecice\" or \"ctest\" to run the full testsuite. \n";
     }
     return 2;
   }


BenjaminRodenberg:
That's great! Currently I'm always using `mpirun -np 4 ./testprecice --whatever` and then (somehow) deal with the totally messed up output. Merging this will make my life much easier :)

ajaust:
Most part of the honor goes to @uekerman since it is actually his lecture. I just used his material. :wink:

fsimonis:
I had a deeper look into this.

We could let rank 0 enter `init_unit_tests()` and exit from there. This will work for all flags apart from `--list_content` and `--list_labels` as they are handled after the initialization.

We could let rank 0 leave `init_unit_tests()` (like this PR does).
This will trigger the checks from tests that run on more then 1 rank. 

I opened an issue in the boost test repo regarding a hook https://github.com/boostorg/test/issues/324.
Even if such a feature gets implemented, this won't be available to us for quite some time.

fsimonis:
I found a viable alternative to letting rank 0 run into error messages.

We could conditionally attach a fixture to the master test suite which displays an error and exits.
This would result in:
```
$ ./testprecice
Running 529 test cases...
ERROR: The tests require at least 4 MPI processes. Please use "mpirun -np 4 ./tesprecice" or "ctest" to run the full testsuite.
```

I don't think the functionality is part of their public interface though.

DavidSCN:
I must admit I don't get the issue you describe. The proposed solution of this PR works for me just like a charm.

src/testing/main.cpp
```diff
@@ -123,9 +123,9 @@ int main(int argc, char *argv[])
   const auto size = utils::Parallel::current()->size();
   logging::setMPIRank(rank);
 
-  if (size < 4) {
+  if (size < 4 && argc < 2) {
     if (rank == 0) {
-      std::cerr << "ERROR: The tests require at least 4 MPI processes.\n";
+      std::cerr << "ERROR: The tests require at least 4 MPI processes. Please use \"mpirun -np 4 ./tesprecice\" or \"ctest\" to run the full testsuite. \n";
```
ajaust, line 128:
```suggestion
      std::cerr << "ERROR: The tests require at least 4 MPI processes. Please use \"mpirun -np 4 ./testprecice\" or \"ctest\" to run the full testsuite. \n";
```
Typo

fsimonis:
> I must admit I don't get the issue you describe. The proposed solution of this PR works for me just like a charm.

Hmm. I recompiled from scratch and the errors are gone. So, I assume that the issue I ran into was on my side.

Anyhow, the change is super useful, let's get it merged!

fsimonis (approved):

PR #1169, opened 2022-01-30
Added API Methods for writing the gradient data needed for Nearest Neighbor Gradient Mapping

ariguiba:
## Main changes of this PR

Part 2 of Nearest Neighbor Mapping method, issue #1239 . Added API Methods to add gradient data and the necessary changes to the communication and storage of data, as well as integration tests.

Added API methods are found in SolverInterfaceImpl and include: 
-``void writeScalarGradientData ( int dataID, int  valueIndex, const double *value );``
-``void writeBlockScalarGradientData ( int dataID, int size, int*  valueIndices, const double *value, rowMajor = false );``
-``void writeVectorGradientData ( int dataID, int  valueIndex, const double *value, rowMajor = false );``
-``void writeBlockVectorGradientData ( int dataID, int size, int*  valueIndices, const double *value, rowMajor = false );``
 
The values are passed as a block of continuous memory. 
By default, we use the following format for 2D-space (the same format applies for all methods for the 3D case end eventually for higher spatial dimensions) :

- ``ScalarGradientData : (v_dx, v_dy) `` corresponding to the scalar data ``v``
- ``VectorGradientData : (vx_dx, vx_dy, vy_dx, vy_dy) `` corresponding to the data vector ``v = (vx, vy)``
- ``BlockScalarGradientData : (v0_dx, v0_dy, v1_dx, v1_dy, ... ) `` corresponding to the data block ``v0, v1, v2, ...``
- ``BlockVectorGradientData : (v0x_dx, v0x_dy, v0y_dx, v0y_dy, v1x_dy, v1x_dy, v1y_dx, v1y_dy, ... ) `` corresponding to the data block ``v0 = (v0x, v0y), v1 = (v1x, v1y), ...``

The optional ``rowMajor`` parameter allows to enter the values derived in the spacial dimensions first. 
For example, for ``BlockVectorGradientData``, the format looks as follows: `` (v0x_dx, v0y_dx, v1x_dx, v1y_dx, ....
v0x_dy, v0y_dy, v1x_dy, v1y_dy, ... )``

The added integration tests include: 
- Serial tests for : ``write-consistent``, ``read-consistent`` mapping for both ``scalar`` and ``vector`` data types, as well as tests for ``Block``-Methods
- Parallel tests for :  ``read-consistent`` mapping for both ``scalar`` and ``vetor`` data types

## Author's checklist

## Reviewers' checklist


fsimonis:
This feature needed a public discussion regarding its design.
Reviewing 1800 lines of new code that touches about every aspect of the code-base without any preliminary information is quite a challenge.

Please open an issue and explain the point of the feature as well as the various implementation candidates both for documentation and communication.
Some questions I have right away:
* Does the gradient apply to a dataset, or to all vertices of a mesh? Does it make sense to have different gradient configured on the same mesh?
* Can we handle the gradient just like normal data to prevent the duplication in the API? Example `getDataGradientID()` together with `setVectorData()` as usual.

Note on the PR: `src/precice/impl/versions.hpp`, `src/precice/impl/versions.cpp` are autogenereted files and must not be comitted to the repo.  This means that you used an in-source build. This is highly discouraged. Please build out-of source in a separate build directory (also for your own sanity :smiley:  )

fsimonis:
@ariguiba I noticed that the code contains `Gradient` as a `MeshRequirement`, which we changed in #1154 . Could it be that these changes are still based on the initial state of the PR instead of the final version/current develop?

Also, one of the upsides of creating smaller commits is that it makes rebasing way easier as conflicts are smaller.


> There was a public discussion after the intermediate presentation explaining details of the API as well as of the theory.

Future developers and myself didn't attend this discussion, so we need some kind of documentation to understand what decisions were taken and why. Issues are ideal for this as they can be linked to proposed solutions.

> Also note, that most of the lines are test configuration files.

~600 lines of ~1800 are XML files.

> There is no need to merge and review this PR right now. In the future, there will be a thesis documenting this and I don't see a point in opening another issue here.

Fair enough, but we need a short and clear description of what is going on here. A thesis is not suitable for that.

uekerman:
Related PRs:
- https://github.com/precice/precice/pull/1154
- https://github.com/precice/precice/pull/922

src/m2n/M2N.cpp
```diff
@@ -222,7 +223,12 @@ void M2N::send(
       _masterCom->receive(ack, 0);
       _masterCom->send(ack, 0);
     }
-    Event e("m2n.sendData", precice::syncMode);
+
+    if (withGradient)
+      Event e("m2n.sendGradientData", precice::syncMode);
+    else
+      Event e("m2n.sendData", precice::syncMode);
```
DavidSCN, line 230: You won't be able to measure here anything. As stated in the [Eventtimings](https://github.com/precice/EventTimings#readme), the framework measures the time until destruction. In this case, the object is destroyed immediately after leaving the if branch. I guess you can simply use a ternary operator as a remedy.

DavidSCN: In case the event doesn't allow for it, you could also parametrize the event name ("send(Gradient)Data") and create it later using the parametrized name.

DavidSCN:
We definitely need an API function `isDataGradientRequired` similar to `isMeshConnectivityRequired`.

DavidSCN:
For documentation: We probably need to investigate how this method performs with read mappings and the acceleration schemes in preCICE, as the acceleration is computed based on the input data, which we modify later on. Using a serial coupling scheme and Aitken acceleration results in my little experiment in an iteration count five times higher than the 'usual' iteration count.

src/precice/impl/SolverInterfaceImpl.hpp
```diff
@@ -380,15 +406,33 @@ class SolverInterfaceImpl {
...
+   * @brief Write scalar data to the interface mesh
    *
    * The exact mapping and communication must be specified in XYZ.
    *
-   * @param[in] fromDataID ID of the data to be written, e.g. 1 = forces
-   * @param[in] dataPosition Position (coordinate, e.g.) of data to be written
-   * @param[in] dataValue Value of the data to be written
+   * @param fromDataID       [IN] ID of the data to be written (2 = temperature, e.g.)
+   * @param dataPosition [IN] Position (coordinate, e.g.) of data to be written
+   * @param dataValue    [IN] Value of the data to be written
```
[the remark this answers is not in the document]
DavidSCN: If you intend to rearrange the function list, please do not use all caps for `IN`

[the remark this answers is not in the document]
DavidSCN: If you intend to rearrange the function list, please do not use all caps for `IN`

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1172,6 +1179,247 @@ void SolverInterfaceImpl::writeScalarData(
...
+  PRECICE_REQUIRE_DATA_WRITE(dataID);
+
+  PRECICE_CHECK(value != nullptr, "writeVectorGradientData() was called with values == nullptr");
+
+  WriteDataContext &context = _accessor->writeDataContext(dataID);
+  PRECICE_ASSERT(context.providedData() != nullptr);
+  mesh::Data &data = *context.providedData();
+
+  // Check if Data object with ID dataID has been initialized with gradient data
+  PRECICE_CHECK(data.hasGradient(), "Data \"{}\" has no gradient values available! Please turn on the gradient attribute in the configuration file.", data.getName())
```
DavidSCN, line 1314:
```suggestion
  PRECICE_CHECK(data.hasGradient(), "Data \"{}\" has no gradient values available! Please turn on the gradient attribute in the configuration file.", data.getName())
```
```suggestion
  PRECICE_CHECK(data.hasGradient(), "Data \"{}\" has no gradient values available. Please turn on the gradient attribute in the configuration file.", data.getName())
```
Let's not use exclamation marks.

DavidSCN, line 1314:
```suggestion
  PRECICE_CHECK(data.hasGradient(), "Data \"{}\" has no gradient values available! Please turn on the gradient attribute in the configuration file.", data.getName())
```
```suggestion
  PRECICE_CHECK(data.hasGradient(), "Data \"{}\" has no gradient values available. Please turn on the gradient attribute in the configuration file.", data.getName())
```
Let's not use exclamation marks.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1172,6 +1179,247 @@ void SolverInterfaceImpl::writeScalarData(
...
+        // Values are entered derived in spatial dimensions first : gradient matrix read rowwise
+        const int offset                                           = dimData * _dimensions;
+        gradientValuesInternal(dimSpace, offsetInternal + dimData) = value[offset + dimSpace];
+      } else {
+        // Values are entered derived components first : gradient matrix read columnwise
+        const int offset                                           = dimSpace * _dimensions;
+        gradientValuesInternal(dimSpace, offsetInternal + dimData) = value[offset + dimData];
+      }
+
+      //TODO: Add asserts
```
DavidSCN, line 1350: TODO

DavidSCN, line 1350: TODO

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1172,6 +1179,247 @@ void SolverInterfaceImpl::writeScalarData(
...
+  PRECICE_CHECK(values != nullptr, "writeBlockVectorGradientData() was called with values == nullptr");
+
+  // Get the data
+  WriteDataContext &context = _accessor->writeDataContext(dataID);
+  PRECICE_ASSERT(context.providedData() != nullptr);
+
+  mesh::Data &data = *context.providedData();
+
+  // Check if the Data object with ID dataID has been initialized with gradient data
+  PRECICE_CHECK(data.hasGradient(), "Data \"{}\" has no gradient values available! Please turn on the gradient attribute in the configuration file.", data.getName())
```
DavidSCN, line 1379:
```suggestion
  PRECICE_CHECK(data.hasGradient(), "Data \"{}\" has no gradient values available! Please turn on the gradient attribute in the configuration file.", data.getName())
```
```suggestion
  PRECICE_CHECK(data.hasGradient(), "Data \"{}\" has no gradient values available. Please turn on the gradient attribute in the configuration file.", data.getName())
```
Same as above. Also, could you describe how to do that "turn on the gradient attribue"?

DavidSCN, line 1379:
```suggestion
  PRECICE_CHECK(data.hasGradient(), "Data \"{}\" has no gradient values available! Please turn on the gradient attribute in the configuration file.", data.getName())
```
```suggestion
  PRECICE_CHECK(data.hasGradient(), "Data \"{}\" has no gradient values available. Please turn on the gradient attribute in the configuration file.", data.getName())
```
Same as above.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1172,6 +1179,247 @@ void SolverInterfaceImpl::writeScalarData(
...
+          const int offsetIn                                 = i * _dimensions;
+          gradientValues(dimSpace, offsetInternal + dimData) = values[offsetOut + offsetIn + dimData];
+        } else {
+          // Values are entered derived components first : gradient matrices input one after the other (read columnwise)
+          const int offsetOut                                = i * _dimensions * _dimensions;
+          const int offsetIn                                 = dimData * _dimensions;
+          gradientValues(dimSpace, offsetInternal + dimData) = values[offsetOut + offsetIn + dimSpace];
+        }
+
+        // TODO: Add asserts
```
DavidSCN, line 1419: TODO

DavidSCN, line 1419: TODO

DavidSCN (commented):
Looks good overall. We need to adapt the doxygen description of the API methods a little bit (see comments). I didn't comment in all places in case of similar descriptions, so please expand the edits to all relevant places.

src/precice/SolverInterface.hpp
```diff
@@ -628,6 +641,60 @@ class SolverInterface {
...
+   *                                v0x_dz, v0y_dz, v0z_dz, v1x_dz, v1y_dz, v1z_dz, ... , vnx_dz, vny_dz, vnz_dz)
+   *
+   *
+   * @param[in] dataID ID to write to.
+   * @param[in] size Number n of vertices.
+   * @param[in] values Pointer to the gradient values read columnwise by default.
+   * @param[in] rowMajor Allows to input the data derived in spatial directions first
+   *
+   * @pre count of available elements at gradient values matches the configured dimension * size
+   * @pre count of available elements at valueIndices matches the given size
```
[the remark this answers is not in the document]
ariguiba: Since in the implementation we don't (or rather we can't) check if the pointers have the needed size, we assume that it is a pre conditition right ?

src/precice/impl/SolverInterfaceImpl.hpp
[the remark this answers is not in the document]
ariguiba: These weren't my changes, they got applied after i pulled the latest changes from the develop branch, since there were conflicts in this class. Should I revert them ?

tests/serial/mapping-nearest-neighbor-gradient/GradientTestBidirectionalReadVector.xml
```diff
@@ -0,0 +1,52 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<precice-configuration>
+  <solver-interface dimensions="3" experimental="on">
```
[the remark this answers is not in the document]
ariguiba: This was suggested during the review, since we might remove the flag "gradient = on". I wasn't sure if I should apply the attribute here or under the data attribute.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1172,6 +1179,247 @@ void SolverInterfaceImpl::writeScalarData(
...
+  PRECICE_CHECK(_state != State::Finalized, "writeScalarGradientData(...) cannot be called before finalize().")
+  PRECICE_REQUIRE_DATA_WRITE(dataID);
+  PRECICE_DEBUG("gradient value = {}", Eigen::Map<const Eigen::VectorXd>(value, _dimensions).format(utils::eigenio::debug()));
+
+  WriteDataContext &context = _accessor->writeDataContext(dataID);
+  PRECICE_ASSERT(context.providedData() != nullptr);
+  mesh::Data &data = *context.providedData();
+
+  //Check if data has been initialized to include gradient data
+  PRECICE_CHECK(data.hasGradient(), "Data \"{}\" has no gradient values available! Please turn on the gradient attribute in the configuration file.", data.getName())
```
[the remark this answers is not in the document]
ariguiba: I added the whole code inside an if-branch with ``isDataGradientRequired`` so the data is not added at all if this is false. And kept the check inside the branch. I'm not sure this is the best option though.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1172,6 +1179,247 @@ void SolverInterfaceImpl::writeScalarData(
...
+                _dimensions, context.getDataName());
+  PRECICE_VALIDATE_DATA(value, _dimensions);
+
+  // Gets the gradientvalues matrix corresponding to the dataID
+  auto &     gradientValues = data.gradientValues();
+  const auto vertexCount    = gradientValues.cols() / context.getDataDimensions();
+
+  //Check if the index and dimensions are valid
+  PRECICE_CHECK(valueIndex >= -1,
+                "Invalid value index ({}) when writing gradient scalar data. Value index must be >= 0. "
```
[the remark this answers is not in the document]
ariguiba: I implemented this consistently as in the ``writeData``-methods. But you are right it does make more sense. Should I change it for all the methods or just the gradient methods ?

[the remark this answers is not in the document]
ariguiba: I implemented this consistently with the ``writeData``-methods. I think the reason we don't call it ``vertexIndex`` is because it's not the absolute index of the vertex in the whole mesh but rather just the index in the current block set of data input.
I am not sure but I can change it nevertheless, if you think it's more fitting.

docs/changelog/1169.md
```diff
@@ -0,0 +1,6 @@
+
+- Added API methods for entering gradient data in the solver interface. Methods available are for scalar gradient data, vector gradient data, block scalar gradient data and block vector gradient data
+- Added integration tests for serial execution: to test read and write consistent mapping for scalar and vector data and test the different block methods
+- Added integration tests for parallel execution: to test read consistent mapping for scalar and vector data.
+- Changed the coupling scheme base class to additionally send gradient data 
+- Change the m2n communication to send gradient data
```
uekerman, line 6: Item 2 and 3 are currently not understandable without item 1. Typically, we should be able to sort such changelog entries arbitrarily (e.g. alphabetically). We could either add more information or create a hierarchy of information here.

```
- Added nearest-neighbor-gradient mapping.
  - Added integration tests ...
  - ...
```

src/precice/SolverInterface.hpp
```diff
@@ -361,6 +361,19 @@ class SolverInterface {
...
+   * We check if the data object has been intialized with the gradient flag.
+   *
+   * preCICE may require gradient data information from the solver and
+   * ignores any API calls regarding gradient data if it is not required.
+   * (When applying a nearest-neighbor-gradient mapping)
+   *
+   * @param[in] dataID the id of the data
+   * @returns whether gradient is required
+   */
+  bool isDataGradientRequired(int dataID) const;
```
uekerman, line 375:
```suggestion
  bool isGradientRequired(int dataID) const;
```
? Could be sufficient already as we pass a dataID?
No strong opinion, just a suggestion.

src/precice/SolverInterface.hpp
```diff
@@ -628,6 +641,60 @@ class SolverInterface {
...
+   *                                        v1x_dx, v1x_dy, v1x_dz, v1y_dx, v1y_dy, v1y_dz, v1z_dx, v1z_dy, v1z_dz,
+   *                                        ... ,
+   *                                        vnx_dx, vnx_dy, vnx_dz, vny_dx, vny_dy, vny_dz, vnz_dx, vnz_dy, vnz_dz)
+   *
+   * corresponding to the vector data v0 = (v0x, v0y, v0z) , v1 = (v1x, v1y, v1z), ... , vn = (vnx, vny, vnz) differentiated in spatial directions x,y and z.
+   *
+   * The optional \p rowsFirst attribute allows to enter the derivatives directions-wise:
+   *
+   * For the 3D-format as follows: (v0x_dx, v0y_dx, v1x_dx, v1y_dx, ... , vnx_dx, vny_dx,
+   *                                v0x_dy, v0y_dy, v1x_dy, v1y_dy, ... , vnx_dy, vny_dy)
```
uekerman, line 671:
```suggestion
   * For the 2D-format as follows: (v0x_dx, v0y_dx, v1x_dx, v1y_dx, ... , vnx_dx, vny_dx,
   *                                v0x_dy, v0y_dy, v1x_dy, v1y_dy, ... , vnx_dy, vny_dy)
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1172,6 +1179,280 @@ void SolverInterfaceImpl::writeScalarData(
                 "Please make sure you only use the results from calls to setMeshVertex/Vertices().",
                 context.getDataName(), valueIndex);
   values[valueIndex] = value;
+
+  PRECICE_DEBUG("Written Scalar Value = {}", value);
```
uekerman, line 1183:
```suggestion
  PRECICE_DEBUG("Written scalar value = {}", value);
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1172,6 +1179,280 @@ void SolverInterfaceImpl::writeScalarData(
...
+    int           valueIndex,
+    const double *gradientValue)
+{
+
+  PRECICE_TRACE(dataID, valueIndex);
+  PRECICE_CHECK(_state != State::Finalized, "writeScalarGradientData(...) cannot be called after finalize().")
+  PRECICE_REQUIRE_DATA_WRITE(dataID);
+
+  if (isDataGradientRequired(dataID)) {
+    PRECICE_DEBUG("gradient value = {}", Eigen::Map<const Eigen::VectorXd>(gradientValue, _dimensions).format(utils::eigenio::debug()));
```
uekerman, line 1197:
```suggestion
    PRECICE_DEBUG("Gradient value = {}", Eigen::Map<const Eigen::VectorXd>(gradientValue, _dimensions).format(utils::eigenio::debug()));
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1172,6 +1179,280 @@ void SolverInterfaceImpl::writeScalarData(
...
+    PRECICE_CHECK(gradientValue != nullptr, "writeScalarGradientData() was called with gradientValues == nullptr");
+
+    WriteDataContext &context = _accessor->writeDataContext(dataID);
+    PRECICE_ASSERT(context.providedData() != nullptr);
+    mesh::Data &data = *context.providedData();
+
+    //Check if data has been initialized to include gradient data
+    PRECICE_CHECK(data.hasGradient(), "Data \"{}\" has no gradient values available! Please set the gradient flag to true under the data attribute in the configuration file.", data.getName())
```
uekerman, line 1205:
```suggestion
    PRECICE_CHECK(data.hasGradient(), "Data \"{}\" has no gradient values available. Please set the gradient flag to true under the data attribute in the configuration file.", data.getName())
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1172,6 +1179,280 @@ void SolverInterfaceImpl::writeScalarData(
...
+
+    // Size of the gradient data input : must be spaceDimensions * dataDimensions -> here spaceDimensions (since for scalar: dataDimensions = 1)
+    PRECICE_CHECK(data.getSpacialDimensions() == _dimensions,
```
uekerman, line 1208: In most code parts, preCICE uses the spelling "spatial", no "spacial". Not critical, I just wanted to mention it.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1172,6 +1179,280 @@ void SolverInterfaceImpl::writeScalarData(
...
+
+    PRECICE_CHECK(data.hasGradient(), "Data \"{}\" has no gradient values available! Please set the gradient flag to true under the data attribute in the configuration file.", data.getName())
+
+    PRECICE_CHECK(data.getDimensions() == 1,
+                  "You cannot call writeBlockScalarGradientData on the vector data type \"{}\". Use writeBlockVectorGradientData or change the data type for \"{}\" to scalar.",
+                  data.getName(), data.getName());
+
+    PRECICE_CHECK(data.getSpacialDimensions() == _dimensions,
+                  "The spatial dimensions of the data object and of the mesh don't match. Please make sure to initialize {} space dimensions for the scalar data type \"{}\".",
+                  _dimensions, context.getDataName());
```
uekerman, line 1273: This one should be an assertion, right?
I don't see how a user could cause this.

src/precice/impl/SolverInterfaceImpl.hpp
```diff
@@ -362,6 +365,29 @@ class SolverInterfaceImpl {
   /// Computes and maps all read data mapped to mesh with given ID.
   void mapReadDataTo(int toMeshID);
 
+  /**
+   * @brief Write vectorial data to the interface mesh
+   *
+   * The exact mapping and communication must be specified in XYZ.
```
uekerman, line 371:
```suggestion
```
?

tests/parallel/mapping-nearest-neighbor-gradient/GradientTestParallelScalar.cpp
```diff
@@ -0,0 +1,89 @@
...
+    interface.setMeshVertices(meshID, 6, positions, vertexIDs);
+    interface.initialize();
+    int    dataID    = interface.getDataID("Data2", meshID);
+    double values[6] = {1.0, 2.0, 3.0, 4.0, 5.0, 6.0};
+
+    interface.writeBlockScalarData(dataID, 6, vertexIDs, values);
+
+    if (interface.isDataGradientRequired(dataID)) {
+
+      BOOST_TEST(interface.isDataGradientRequired(dataID) == true);
```
uekerman, line 76: The test should be outside of the if-clause.

```suggestion
BOOST_TEST(interface.isDataGradientRequired(dataID));
    if (interface.isDataGradientRequired(dataID)) {
    
```

tests/parallel/mapping-nearest-neighbor-gradient/GradientTestParallelScalar.xml
```diff
@@ -0,0 +1,51 @@
...
+      <use-data name="Data2" />
+    </mesh>
+
+    <mesh name="MeshTwo">
+      <use-data name="Data1" />
+      <use-data name="Data2" />
+    </mesh>
+
+    <participant name="SolverOne">
+      <master:mpi-single />
```
uekerman, line 18:
```suggestion
```
This should be default and not required.

src/m2n/M2N.hpp
```diff
@@ -147,9 +147,12 @@ class M2N {
   void createDistributedCommunication(const mesh::PtrMesh &mesh);
 
   /// Sends an array of double values from all slaves (different for each slave).
+  /// Setting the flag to true indicates that the data send is gradient data
```
uekerman, line 150:
```suggestion
  /// Setting the flag "withGradient" to true indicates that the data send is gradient data
```

src/m2n/M2N.hpp
```diff
@@ -147,9 +147,12 @@ class M2N {
...
+  /// The spatial directions have to be added manually as an input in valueDimension
```
uekerman, line 151: I don't understand this sentence.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -126,7 +126,7 @@ MappingConfiguration::MappingConfiguration(
   }
   {
     XMLTag tag(*this, VALUE_NEAREST_NEIGHBOR_GRADIENT, occ, TAG);
-    tag.setDocumentation("Nearest-neighbour mapping which uses a rstar-spacial index tree to index meshes and run nearest-neighbour queries with a gradient optimization.");
+    tag.setDocumentation("Nearest-neighbor-gradient mapping which uses a rstar-spacial index tree to index meshes and runs nearest-neighbor queries with an linear approximation using gradient data.");
```
uekerman, line 129: Sounds like the queries use the linear approximation.
The queries only give the nearest neighbor. Afterwards, for the mapping we use a linear approximation.

src/mesh/Mesh.cpp
```diff
@@ -186,7 +186,7 @@ const PtrData &Mesh::data(const std::string &dataName) const
 PtrData &Mesh::createDataWithGradient(
```
uekerman, line 186: Couldn't we make the "withGradient" information an argument of `createData`?
Could make the overall logic a bit easier to understand.

src/m2n/M2N.hpp
uekerman: Is this flag actually only needed for the name of the event? If yes, I would vote against adding it. Understandable for the test of the thesis, but not sth we might need in the long run, no?

src/precice/SolverInterface.hpp
```diff
@@ -651,13 +718,52 @@ class SolverInterface {
...
+   * @pre initialize() has been called
+   * @pre vertex with dataID exists and contains data
+   * @pre Data with dataID has attribute hasGradient = true
+   *
+   * @see SolverInterface::setMeshVertex()
+   */
+  void writeVectorGradientData(
+      int           dataID,
+      int           valueIndex,
+      const double *gradientValue,
```
uekerman, line 756:
```suggestion
      const double *gradientValues,
```

?

uekerman (changes requested):
Nice job, @ariguiba :+1:

I did not look at all details, but only add the code parts I was curious about when reading your thesis. And most of the time, I only commented on the first occasion of a change.

There are still a few things where some discussion and tweaks are necessary, but the overall concept looks good.

- Let's not forget to also add documentation on https://precice.org/configuration-mapping.html.

- General concern: I am not sure if we need the `rowFirst`, `rowMajor` arguments. At least for me, they confuse me more than I could imagine that they help. Some confusing aspects:
  - `writeBlockVectorGradientData`: I would expect a different behavior for `rowFirst=true`: transposing the inner Jacobi matrix, but keeping the vertex dimension the outer most one.
  - `writeBlockScalarGradientData`: if this function has an argument `rowsFirst`, we would also need to add one for `setMeshVertices`, `readBlockVectorData`, and `writeBlockVectorData`.

  I think it might lead to less bugs on user side if we clearly communicate which format we want (we do this already fairly well) and only allow one format. Users know how to transpose data structures in their codes. I don't think we could do a more efficient job in preCICE.

- Are we still missing an error message if a user configures a conservative NNG? 

@DavidSCN 

In the long run, we want to get rid of the `gradient` attribute in the configuration, right? (but let preCICE deduce this information from the configured mapping methods). If yes, could you please already open an issue?
Seeing this, we should probably indeed still make the nearest-neighbor-gradient mapping experimental. If we remove the attribute later, this should not become a breaking change.

uekerman:
Right now, gradient data is always communicated, right? Even if not needed.

tests/parallel/mapping-nearest-neighbor-gradient/GradientTestParallelScalar.cpp
```diff
@@ -0,0 +1,89 @@
...
+#include "precice/impl/SharedPointer.hpp"
+#include "precice/impl/SolverInterfaceImpl.hpp"
+#include "precice/types.hpp"
+#include "testing/TestContext.hpp"
+#include "testing/Testing.hpp"
+
+using namespace precice;
+using precice::testing::TestContext;
+
+BOOST_AUTO_TEST_SUITE(PreciceTests)
```
BenjaminRodenberg, line 35: The main test suite was renamed to `Integration` in https://github.com/precice/precice/commit/8b51a94a2ca69cbbd0ff0b4e2d4c165769aefe9f.

src/precice/impl/SolverInterfaceImpl.hpp
ariguiba: These are changes to the ``writeData``-methods, which were probably a result of the merge with the current develop version, but I am not responsible for these.

ariguiba:
> Right now, gradient data is always communicated, right? Even if not needed.

Yes. The idea behind is, is that if the user adds gradient data, the nearest-neighbor-gradient will definitely be used. I don't know if there is a use-case where we have the gradient data available but wouldn't want to communicate it.

src/mesh/Mesh.cpp
ariguiba: The problem with this option, is that we would have to set the spatial directions as well to a default value. An option would be to set the spatial dimensions by default to 1 for data when creating it (since this is irrelevant for data without gradient anyways) since we always use a vector to store all the values (even for vector data) but it not very "correct" in a sense that even data can be of vector values, but when creating the object this could be seen as a "wrong" input. I don't know in that case which is better. But this can be further discussed of course.

src/precice/impl/SolverInterfaceImpl.hpp
```diff
@@ -362,6 +365,29 @@ class SolverInterfaceImpl {
...
+   * @param[in] dataValue Value of the data to be written
+   */
+  void writeVectorData(
+      int           fromDataID,
+      int           valueIndex,
+      const double *value);
+
+  /**
+   * @copydoc precice::SolverInterface::writeVectorGradientData()
+   */
```
BenjaminRodenberg, line 384:
```suggestion
  /// @copydoc precice::SolverInterface::writeVectorGradientData
```
You can remove the brackets and use a single-line comment. This is shorter and simpler with the same effect. See https://github.com/precice/precice/pull/1249. (You only have to provide the arguments of the function, if it is not unique to avoid ambiguities)

You can check the resulting doxygen by running `make doxygen` and then opening `precice/docs/source-code-documentation/index.html`

ariguiba:
A little summary of all the changes that resulted from the reviews (and that came after the thesis):
- ``createDataWithGradient`` in the mesh class was removed merged with the ``createData`` method with an additional flag ``withGradient``, the method head looks like this now ``createData(string name, int dimension, DataID id, bool withGradient = false)``
- Removed the ``rowMajor`` flag in the ``writeBlockScalarGradientData`` method and changed the meaning in ``writeBlockVectorGradientData`` so only the inner matrix is read rowwise but the elements are read one after the other.
- Conservative NNG check is done in the configuration file, not in the mapping class anymore
- All the writeGradient-methods are exprimental for now 
- All the writeGradient-methods include an ``isGradientDataRequired`` check, otherwise an error is thrown and nothing happens in the method
- Removed the gradient-flag in the events when communicating data (since its only purpose was to give different names to the event)
- Some refactoring, mostly renaming of variables: 

   - ``isDataGradientRequired`` -> ``ìsGradientDataRequired``
   - spacial -> spatial
   - ``rowMajor`` -> ``rowsFirst`` for ``writeVector(Block)Gradient`` methods
   - All writeGradient-Methods have the argument ``gradientValues`` (instead of ``value(s)``)

DavidSCN (approved):
Thanks!

PR #1170, opened 2022-02-01
Enable compiler annotations in workflows

### LINKED ISSUE #1135: CI strategy for warnings

fsimonis:
## Main changes of this PR

Attempts to integrate compiler warnings as annotations in GitHub.
Currently, this only shows warnings of the newest compilers aka on Arch.

This PR is experimental.

Closes #1135
Replaces #1138

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -119,6 +119,9 @@ jobs:
         with:
           name: ${{ format('{0} {1} {2} {3}', matrix.IMAGE, matrix.CXX, matrix.CONFIG, matrix.TYPE) }} CompileCommands 
           path: build/compile_commands.json
+      - name: Enable compiler annotations
+        uses: electronjoe/gcc-problem-matcher@v1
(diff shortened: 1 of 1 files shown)


MakisH (approved):
Very nice! This is a good compromise between and I am happy we are not the only ones with this problem .

fsimonis:
I extended the warnings to `-Wextra -Wno-unused-arguments` and the results look good!
Merging now!

PR #1179, opened 2022-02-10
Refactor barycentric coordinate calculation functions

### LINKED ISSUE #1019: Simplify geometric operations on mesh primitives
The current implementation of the barycentric coordinates are overly complicated and explicitly require the normals of triangles or edges. There are simpler ways to implement this. 

* Simplify  `math/barycenter.[ch]pp`
* Remove `Edge::computeNormal()` as it is unnecessary.

_none_

**Additional context**
#1000
#179

qwach:
## Main changes of this PR

Vastly simplifies the functions for calculating barycentric coordinates relative to edges and triangles. The former is now a simple scalar projection and the latter implements an algorithm from a [paper](https://vccimaging.org/Publications/Heidrich2005CBP/Heidrich2005CBP.pdf) by Heidrich (2005). In both cases, the projection onto the primitive in Cartesian coordinates is no longer calculated, and hence not returned either, but it was not actually needed anyway, and if necessary can be calculated given the corner points and the barycentric coordinates. Performance does not seem to be impacted.

Closes #1019 

## Author's checklist

* [ ] ~~I added a changelog file with `make changelog` if there are user-observable changes since the last release.~~

## Reviewers' checklist

* [ ] ~~Does the changelog entry make sense? Is it formatted correctly?~~

### DIFF
--- a/src/math/barycenter.cpp
+++ b/src/math/barycenter.cpp
@@ -13,18 +13,18 @@ namespace barycenter {
 Eigen::Vector2d calcBarycentricCoordsForEdge(
     const Eigen::VectorXd &a,
     const Eigen::VectorXd &b,
-    const Eigen::VectorXd &p)
+    const Eigen::VectorXd &u)
 {
   using Eigen::Vector2d;
   using Eigen::VectorXd;
 
   const int dimensions = a.size();
   PRECICE_ASSERT(dimensions == b.size(), "A and B need to have the same dimensions.", dimensions, b.size());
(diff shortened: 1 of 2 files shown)


fsimonis (changes requested):
That's a very clean and readable contribution! Thank you very much!

Only nitpick I have is your naming choice for the location parameter.
You called it _p_ , which is conventionally used for projections. Could you rename it to _u_ ?

qwach:
Done, and thanks!

PR #1181, opened 2022-02-10
Add CMake variable to control when assertions are enabled

### LINKED ISSUE #663: Introduce a Tweakable Assertion Policy

qwach:
## Main changes of this PR
Adds CMake variable `PRECICE_ASSERTION_POLICY` that can be set to `ON`/always, `DEBUG`/debug builds only, or `OFF`/never to govern in which builds assertions are enabled, which is translated to preprocessor flags `ASSERT_ALWAYS`, `ASSERT_DEBUG`, and `ASSERT_NEVER`, respectively. The default is `DEBUG`.

Closes #663 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -73,6 +73,18 @@ if(PRECICE_PETScMapping AND NOT PRECICE_MPICommunication)
 endif()
 
 set(PRECICE_CTEST_MPI_FLAGS "" CACHE STRING "Add additional flags to mpiexec for running tests via CTest.")
+set(PRECICE_ASSERTION_POLICY DEBUG CACHE STRING "For which builds to enable assertions (always, debug only, never). Debug only is default.")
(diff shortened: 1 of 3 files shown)


fsimonis:
Thanks for your contribution. I already solved the issue in #1177 as it really started to become a priority.
I just noticed that I forgot to link to the original issue #633. Sorry about that :sweat_smile:

PR #1182, opened 2022-02-13
Change VTK/VTU Output Array Format

kursatyurt:
## Main changes of this PR

Change `vtk` export array type from `SCALAR` (1(defafult)-4 Component) and `VECTOR` (3 Component) to `FIELD` which allows arbitrary number of components.

Removes unnecessary pseudo-3D vector type from `vtu` export by removing additional component `0.0` and writing the correct number of components for 2D vectors. 

For ASTE replay mode we need to parse preCICE outputs. For 2D cases, the additional component causes errors while parsing data. This change writes the exact amount of components for n-D vectors.

## Author's checklist

* [] I sticked to C++14 features.

## Reviewers' checklist

### DIFF
--- a/src/io/ExportVTK.cpp
+++ b/src/io/ExportVTK.cpp
@@ -21,7 +21,7 @@ namespace io {
 void ExportVTK::doExport(
     const std::string &name,
     const std::string &location,
-    const mesh::Mesh  &mesh)
+    const mesh::Mesh & mesh)
 {
   PRECICE_TRACE(name, location, mesh.getName());
   PRECICE_ASSERT(name != std::string(""));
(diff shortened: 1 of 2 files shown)


kursatyurt:
Even if working with field data and reducing the dimension of the data helps to reduce memory/file footprint, it seems it is going to be error-prone for the data created outside of the preCICE (using ASTE tools or ParaView). Hence, like the point data for the vector data writing as 3D and while parsing ignoring the last component is seems to be the best solution that fits general output from any software.

Lessons learned : 
In VTK format even for 2D data one should expect 3D mesh points and 3D vectors and handle them by themself.

PR #1183, opened 2022-02-13
New API  method isScalar(dataID)

kursatyurt:
## Main changes of this PR

Implement a new method to solver interface to get whether a data is a scalar or not.

As discussed in https://github.com/precice/aste/issues/84 and #474  in order to avoid duplication in configuration files we need a method to get information about data is whether scalar or not. 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/SolverInterface.cpp
+++ b/src/precice/SolverInterface.cpp
@@ -19,7 +19,7 @@ SolverInterface::SolverInterface(
     const std::string &configurationFileName,
     int                solverProcessIndex,
     int                solverProcessSize,
-    void              *communicator)
+    void *             communicator)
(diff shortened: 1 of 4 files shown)


uekerman:
Random remarks:

* Please also implement a simple integration test for this
* When adding it to the C++ API, we also need to add it to all bindings :grin:

fsimonis:
@kursatyurt can you comment on why you closed this PR?

kursatyurt:
> @kursatyurt can you comment on why you closed this PR?

This new API method is supposed to be used in ASTE in order to reduce user input from vector data names and scalar data names separately to just data names. However, we decided to ask them separately as user input.

PR #1185, opened 2022-02-16
Remove mostly redundant initialization of _allData.

BenjaminRodenberg:
## Main changes of this PR

`_allData` is storing mostly redundant information. Initializing it correctly is prone to error. Therefore, I would like not actually store `_allData`, but compute it on-the-fly depending on the context it is used.

Question: Is there a nicer way to return a reference to `_allData`? I guess we have to store `_allData` somewhere, if we want to return a reference. Or is there a possibility to create a wrapper that looks like a reference from the perspective of client code asking for `_allData`?

Belongs to the bigger work package #1029, because I will have to change some of the underlying datastructures in the coupling scheme. If I don't have to worry about `_allData` anymore, this will make things easier.

I'm not sure how performance critical this is, but since we are only collecting and rearranging pointers, I don't expect a large impact.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -256,7 +256,7 @@ void BaseCouplingScheme::advance()
 void BaseCouplingScheme::storeExtrapolationData()
 {
   PRECICE_TRACE(_timeWindows);
-  for (DataMap::value_type &pair : _allData) {
+  for (DataMap::value_type &pair : getAllData()) {
     PRECICE_DEBUG("Store data: {}", pair.first);
     pair.second->storeExtrapolationData();
   }
@@ -449,7 +449,7 @@ void BaseCouplingScheme::initializeStorages()
 {
   PRECICE_TRACE();
   // Reserve storage for all data
-  for (DataMap::value_type &pair : _allData) {
+  for (DataMap::value_type &pair : getAllData()) {
     pair.second->initializeExtrapolation();
   }
   // Reserve storage for acceleration
@@ -482,8 +482,8 @@ void BaseCouplingScheme::addConvergenceMeasure(
     bool                        doesLogging)
 {
   ConvergenceMeasureContext convMeasure;
-  PRECICE_ASSERT(_allData.count(dataID) == 1, "Data with given data ID must exist!");
-  convMeasure.couplingData = &(*_allData[dataID]);
+  PRECICE_ASSERT(getAllData().count(dataID) == 1, "Data with given data ID must exist!");
+  convMeasure.couplingData = &(*getAllData()[dataID]);
   convMeasure.suffices     = suffices;
   convMeasure.strict       = strict;
   convMeasure.measure      = std::move(measure);
@@ -662,8 +662,8 @@ bool BaseCouplingScheme::doImplicitStep()
 void BaseCouplingScheme::assignDataToConvergenceMeasure(ConvergenceMeasureContext *convergenceMeasure, DataID dataID)
 {
   PRECICE_TRACE(dataID);
-  DataMap::iterator iter = _allData.find(dataID);
-  PRECICE_ASSERT(iter != _allData.end(), "Given data ID does not exist in _allData!");
+  DataMap::iterator iter = getAllData().find(dataID);
+  PRECICE_ASSERT(iter != getAllData().end(), "Given data ID does not exist in getAllData()!");
   convergenceMeasure->couplingData = &(*(iter->second));
 }
 
--- a/src/cplscheme/BaseCouplingScheme.hpp
+++ b/src/cplscheme/BaseCouplingScheme.hpp
@@ -251,6 +251,12 @@ protected:
   void receiveData(const m2n::PtrM2N &m2n, const DataMap &receiveData);
 
   /**
+   * @brief interface to provide all CouplingData, depending on coupling scheme being used
+   * @return DataMap containing all CouplingData
+   */
+  virtual DataMap &getAllData() = 0;
+
+  /**
    * @brief Function to determine whether coupling scheme is an explicit coupling scheme
    * @returns true, if coupling scheme is explicit
    */
@@ -354,7 +360,7 @@ protected:
   void storeIteration()
   {
     PRECICE_ASSERT(isImplicitCouplingScheme());
-    for (DataMap::value_type &pair : _allData) {
+    for (DataMap::value_type &pair : getAllData()) {
       pair.second->storeIteration();
     }
   }
--- a/src/cplscheme/BiCouplingScheme.hpp
+++ b/src/cplscheme/BiCouplingScheme.hpp
@@ -8,6 +8,7 @@
 #include "logging/Logger.hpp"
 #include "m2n/SharedPointer.hpp"
 #include "mesh/SharedPointer.hpp"
+#include "precice/impl/SharedPointer.hpp"
 #include "precice/types.hpp"
 #include "utils/assertion.hpp"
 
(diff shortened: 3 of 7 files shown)


src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -460,6 +450,22 @@ class BaseCouplingScheme : public CouplingScheme {
...
+   * time windows can be performed.
+   *
+   * The standard predictor is of order zero, i.e., simply the converged values
+   * of the last time windows are taken as initial guess for the coupling iterations.
+   * Currently, an order 1 predictor (linear extrapolation) and order 2 predictor
+   * (see https://doi.org/10.1016/j.compstruc.2008.11.013, p.796, Algorithm line 1 )
+   * is implement besides that.
+   */
+  const int _extrapolationOrder;
+
```
BenjaminRodenberg, line 468: Shifted this to avoid `-Wreorder`.

src/cplscheme/BiCouplingScheme.hpp
```diff
@@ -84,6 +85,23 @@ class BiCouplingScheme : public BaseCouplingScheme {
...
   }
 
+  /**
+   * @brief BiCouplingScheme has _sendData and _receiveData
+   * @returns DataMap with all data
+   */
+  DataMap getAllData() override
+  {
+    DataMap allData;
+    // @todo user C++17 std::map::merge
```
uekerman, line 95:
```suggestion
    // @todo use C++17 std::map::merge
```

uekerman (approved):
Didn't look into performance issues, but seems uncritical to me.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -477,8 +478,8 @@ void BaseCouplingScheme::addConvergenceMeasure(
     bool                        doesLogging)
 {
   ConvergenceMeasureContext convMeasure;
-  PRECICE_ASSERT(_allData.count(dataID) == 1, "Data with given data ID must exist!");
-  convMeasure.couplingData = _allData[dataID];
+  PRECICE_ASSERT(getAllData().count(dataID) == 1, "Data with given data ID must exist!");
+  convMeasure.couplingData = getAllData()[dataID];
```
fsimonis, line 482:
```suggestion
  auto allData = getAllData();
  PRECICE_ASSERT(allData.count(dataID) == 1, "Data with given data ID must exist!");
  convMeasure.couplingData = allData[dataID];
```

or if you really want to be fancy:

```suggestion
  convMeasure.couplingData = [&] {
    auto allData = getAllData();
    PRECICE_ASSERT(allData.count(dataID) == 1, "Data with given data ID must exist!");
    return std::move(allData[dataID]);
  }();
```

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -634,13 +635,15 @@ bool BaseCouplingScheme::doImplicitStep()
...
+      auto accelerationData = getAccelerationData();
+      _acceleration->iterationsConverged(accelerationData);
     }
     newConvergenceMeasurements();
   } else {
     // no convergence achieved for the coupling iteration within the current time window
     if (_acceleration) {
-      _acceleration->performAcceleration(getAccelerationData());
+      auto accelerationData = getAccelerationData();
+      _acceleration->performAcceleration(accelerationData);
```
fsimonis, line 646: This shouldn't be necessary if you adjust the interface of Acceleration. #1194

src/cplscheme/BiCouplingScheme.hpp
```diff
@@ -84,6 +85,23 @@ class BiCouplingScheme : public BaseCouplingScheme {
...
+  {
+    DataMap allData;
+    // @todo use C++17 std::map::merge
+    for (auto &pair : _sendData) {
+      allData.emplace(pair);
+    }
+    for (auto &pair : _receiveData) {
+      allData.emplace(pair);
+    }
+    return allData;
```
fsimonis, line 102: You can simply copy A and add B

```suggestion
    DataMap allData{_sendData};
    allData.insert(_receiveData.begin(), _receiveData.end());
    return allData;
```

src/cplscheme/MultiCouplingScheme.cpp
```diff
@@ -149,9 +146,6 @@ void MultiCouplingScheme::addDataToSend(
   PtrCouplingData     ptrCplData(new CouplingData(data, std::move(mesh), initialize, getExtrapolationOrder()));
   DataMap::value_type dataPair = std::make_pair(id, ptrCplData);
   _sendDataVector[to].insert(dataPair);
```
fsimonis, line 148: You could shorten this even more by directly calling `emplace(id, ptrCplData)`.

Same below

src/cplscheme/MultiCouplingScheme.hpp
```diff
@@ -92,19 +93,44 @@ class MultiCouplingScheme : public BaseCouplingScheme {
...
+  DataMap getAllData() override
+  {
+    DataMap allData;
+    // @todo user C++17 std::map::merge
+    for (auto &sendData : _sendDataVector) {
+      for (auto &dataPair : sendData.second) {
+        if (!utils::contained(dataPair.first, allData)) {
+          allData.emplace(dataPair);
+        }
+      }
```
fsimonis, line 109: You could simply use `allData.insert(sendData.second.begin(), sendData.second.end())`.

[Reference](https://en.cppreference.com/w/cpp/container/map/insert)

> Inserts element(s) into the container, if the container doesn't already contain an element with an equivalent key. 

Same below

src/cplscheme/SerialCouplingScheme.hpp
```diff
@@ -76,9 +76,14 @@ class SerialCouplingScheme : public BiCouplingScheme {
...
-  DataMap &getAccelerationData() override
+  DataMap getAccelerationData() override
   {
-    return getSendData();
+    // create a copy of send data for read-only access.
+    DataMap accelerationData;
+    for (auto &dataPair : getSendData()) {
+      accelerationData.emplace(dataPair);
+    }
+    return accelerationData;
```
fsimonis, line 86: Can't you simply return the send data?

```cpp
return getSendData();
```

fsimonis (commented):
All in all, nice and simple changes.

I was a bit more picky here to further reduce the size of this class.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: Exactly :+1: 039a23b6926fc7bbca46348b33ca73a59f3ba3c8

src/cplscheme/SerialCouplingScheme.hpp
BenjaminRodenberg: Yes. I think this was introduced by me on the fly and now it's not needed anymore. Removed it again in 701d134fbe1b53007856d615f85b48dcd0130c74.

src/cplscheme/MultiCouplingScheme.hpp
BenjaminRodenberg: Done. 854ae5fabe13b0709de5b1282301fa818d3af0c6

src/cplscheme/MultiCouplingScheme.cpp
BenjaminRodenberg: Done [21254c4](https://github.com/precice/precice/pull/1185/commits/21254c426aca3b4fd23c81d546f56b4cc7477843).

src/cplscheme/BiCouplingScheme.hpp
BenjaminRodenberg: Fixed in https://github.com/precice/precice/pull/1185/commits/854ae5fabe13b0709de5b1282301fa818d3af0c6.

PR #1186, opened 2022-02-17
Add tests for non-experimental API and subcycling.

BenjaminRodenberg:
## Main changes of this PR

Rearrange existing tests for subcycling and add new ones for different coupling schemes.

Adding these tests helps to ensure that #1029 does not introduce a regression or accidentally adds new features, if `experimental=false`.

## Author's checklist

* [x] I added a changelog file with `make changelog` if there are user-observable changes since the last release. (no user-side changes)

## Reviewers' checklist


tests/serial/time/explicit/parallel-coupling/ReadWriteScalarDataWithSubcycling.cpp
```diff
@@ -0,0 +1,134 @@
...
+    meshID        = precice.getMeshID("MeshTwo");
+    writeDataID   = precice.getDataID("DataTwo", meshID);
+    writeFunction = dataTwoFunction;
+    readDataID    = precice.getDataID("DataOne", meshID);
+    readFunction  = dataOneFunction;
+  }
+
+  int n_vertices = 1;
+
+  std::vector<VertexID> vertexIDs(n_vertices, 0);
```
uekerman, line 59: Why a vector here? It's only one value.

tests/serial/time/explicit/parallel-coupling/ReadWriteScalarDataWithSubcycling.cpp
```diff
@@ -0,0 +1,134 @@
...
+  } else {
+    BOOST_TEST(context.isNamed("SolverTwo"));
+    meshID        = precice.getMeshID("MeshTwo");
+    writeDataID   = precice.getDataID("DataTwo", meshID);
+    writeFunction = dataTwoFunction;
+    readDataID    = precice.getDataID("DataOne", meshID);
+    readFunction  = dataOneFunction;
+  }
+
+  int n_vertices = 1;
```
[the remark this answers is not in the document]
uekerman: Or maybe `nVertices` to be consistent with your names below. Or actually, you could drop the variable completely. As the value is 1, you do not really need it.

tests/serial/time/explicit/parallel-coupling/ReadWriteScalarDataWithSubcycling.cpp
```diff
@@ -0,0 +1,134 @@
...
+  vertexIDs[0] = precice.setMeshVertex(meshID, Eigen::Vector3d(0.0, 0.0, 0.0).data());
+
+  int    nSubsteps     = 4; // perform subcycling on solvers. 4 steps happen in each window.
+  int    nWindows      = 5; // perform 5 windows.
+  double maxDt         = precice.initialize();
+  double windowDt      = maxDt;
+  int    timestep      = 0;
+  int    timewindow    = 0;
+  double dt            = windowDt / (nSubsteps - 0.5); // Timestep length desired by solver. E.g. 4 steps with size 4/7. Fourth step will be restricted to 2/7 via preCICE steering to fit into the window.
+  double expectedDts[] = {4.0 / 7.0, 4.0 / 7.0, 4.0 / 7.0, 2.0 / 7.0};
```
uekerman, line 73: So `windowDt` above should be 2.0? Maybe add a `BOOST_TEST` to make this easier to understand.

uekerman: I think "desired" in the comment above is misleading.

tests/serial/time/explicit/parallel-coupling/ReadWriteScalarDataWithSubcycling.cpp
```diff
@@ -0,0 +1,134 @@
...
+  double currentDt     = dt; // Timestep length used by solver
+  double time          = timestep * dt;
+
+  if (precice.isActionRequired(precice::constants::actionWriteInitialData())) {
+    for (int i = 0; i < n_vertices; i++) {
```
uekerman, line 78: This loop is unnecessary (and makes the complete thing harder to skim).

uekerman: More occasions below.

tests/serial/time/explicit/parallel-coupling/ReadWriteScalarDataWithSubcycling.cpp
```diff
@@ -0,0 +1,134 @@
...
+  while (precice.isCouplingOngoing()) {
+    double readTime = timewindow * windowDt; // both solvers lag one window behind for parallel-explicit coupling.
+    BOOST_TEST(readData.size() == n_vertices);
+    if (timestep % nSubsteps == 0) {
+      BOOST_TEST(precice.isReadDataAvailable());
+    } else {
+      BOOST_TEST(!precice.isReadDataAvailable());
+    }
+    for (int i = 0; i < n_vertices; i++) {
+      oldReadData = readData[i];
```
uekerman, line 96: Why is this variable needed? Similar for `oldWriteData` below.

tests/serial/time/explicit/parallel-coupling/ReadWriteScalarDataWithSubcycling.cpp
uekerman: More occasions below.

tests/serial/time/explicit/parallel-coupling/ReadWriteScalarDataWithSubcycling.cpp
```diff
@@ -0,0 +1,134 @@
...
+      }
+      BOOST_TEST(readData[i] == readFunction(readTime, i));
+    }
+
+    // solve usually goes here. Dummy solve: Just sampling the writeFunction.
+    BOOST_TEST(currentDt == expectedDts[timestep % nSubsteps]);
+    time += currentDt;
+
+    if (precice.isWriteDataRequired(currentDt)) {
+      BOOST_TEST(writeData.size() == n_vertices);
```
uekerman, line 108: This seems to be an odd test. Don't test the test, but test preCICE. Again makes the complete things unnecessarily hard to skim.

uekerman (changes requested):
I only had a detailed look at one of the tests. Not outsourcing common parts of tests into generic functions, I guess is OK here. But this should not mean that we copy unnecessary code parts, see below. Goal should be that each individual test is as easy to understand as possible. Once it fails, we wan to understand what the test does :grin:

BenjaminRodenberg:
@uekerman Do you think there are other tests that we would need here to make sure we don't break the classical subcycling?  The only thing I can imagine is data initialization with subcycling. But this should already be covered by other tests.

 We could also shift some tests from other places over here. I just looked into `SerialTests.cpp` and `ParallelTests.cpp` for `subcycling` and found the following test.

https://github.com/precice/precice/blob/703fbe46165676bc16a4e7cfbad077b1ff8609b4/src/precice/tests/SerialTests.cpp#L1492-L1504

If the main purpose of this test is not subcycling, then we should leave it as it is (in this PR).

tests/serial/time/explicit/parallel-coupling/ReadWriteScalarDataWithSubcycling.cpp
```diff
@@ -0,0 +1,134 @@
...
+BOOST_AUTO_TEST_SUITE(PreciceTests)
+BOOST_AUTO_TEST_SUITE(Serial)
+BOOST_AUTO_TEST_SUITE(Time)
+BOOST_AUTO_TEST_SUITE(Explicit)
+BOOST_AUTO_TEST_SUITE(ParallelCoupling)
+
+/**
+ * @brief Test to run a simple coupling with subcycling.
+ *
+ * Ensures that each time step provides its own data, but preCICE will only exchange data at the end of the window.
```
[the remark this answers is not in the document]
BenjaminRodenberg: Fixed in fc56ed0d1ea8b411a02d5b7c2ed2c15c80d1161e.

tests/serial/time/explicit/parallel-coupling/ReadWriteScalarDataWithSubcycling.cpp
```diff
@@ -0,0 +1,134 @@
...
+ */
+BOOST_AUTO_TEST_CASE(ReadWriteScalarDataWithSubcycling)
+{
+  PRECICE_TEST("SolverOne"_on(1_rank), "SolverTwo"_on(1_rank));
+
+  SolverInterface precice(context.name, context.config(), 0, 1); // serial coupling, SolverOne first
```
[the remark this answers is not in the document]
BenjaminRodenberg: Yes. Removed it, because it's not really needed in 89769c6e23e2875692d03bd82b3a640e1438324b

tests/serial/time/explicit/parallel-coupling/ReadWriteScalarDataWithSubcycling.cpp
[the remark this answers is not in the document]
BenjaminRodenberg: Removed it in ad91a3a8d3bcef95dacf5658019536c22411a03a.

tests/serial/time/explicit/parallel-coupling/ReadWriteScalarDataWithSubcycling.cpp
BenjaminRodenberg: Cleaned it up in ad91a3a8d3bcef95dacf5658019536c22411a03a.

tests/serial/time/explicit/parallel-coupling/ReadWriteScalarDataWithSubcycling.cpp
BenjaminRodenberg: Changed this in ad91a3a8d3bcef95dacf5658019536c22411a03a.

tests/serial/time/explicit/parallel-coupling/ReadWriteScalarDataWithSubcycling.cpp
BenjaminRodenberg: Yes. Removed it in ad91a3a8d3bcef95dacf5658019536c22411a03a.

tests/serial/time/explicit/parallel-coupling/ReadWriteScalarDataWithSubcycling.cpp
BenjaminRodenberg: It's not needed. I think this becomes more interesting, when we use waveform relaxation. But let's not worry about it here. Cleaned it up in 1bb43353bf57c481beb06ce99eac9cd918263f7b.

tests/serial/time/implicit/parallel-coupling/ReadWriteScalarDataWithSubcycling.cpp
```diff
@@ -0,0 +1,142 @@
...
+    timestep++;
+    if (precice.isActionRequired(precice::constants::actionReadIterationCheckpoint())) { // at end of window and we have to repeat it.
+      iterations++;
+      timestep = windowStartStep;
+      time     = windowStartTime;
+      precice.markActionFulfilled(precice::constants::actionReadIterationCheckpoint()); // this test does not care about checkpointing, but we have to make the action
+    }
+    if (precice.isTimeWindowComplete()) {
+      timewindow++;
+      iterations = 0;
```
uekerman, line 128: You could also test the value of `iterations` here

uekerman (approved):
Much easier to understand now :+1:

> @uekerman Do you think there are other tests that we would need here to make sure we don't break the classical subcycling? The only thing I can imagine is data initialization with subcycling. But this should already be covered by other tests.

Maybe testing a `CompositionalCouplingScheme` case? Not much more comes to mind.

> Flite and Calculix testcase

That's still one from Bernhard. Not sure whether this one is about subcycling mainly. I guess you have a better overview.

What's the difference between `DoNothingWithSubcycling` and `ReadWriteScalarDataWIthSubcycling`? Why does the one only exist for serial coupling? Do we need both? If yes, maybe we could find better names.

BenjaminRodenberg:
> What's the difference between `DoNothingWithSubcycling` and `ReadWriteScalarDataWIthSubcycling`? Why does the one only exist for serial coupling? Do we need both? If yes, maybe we could find better names.

Removed this test in [381c7de](https://github.com/precice/precice/pull/1186/commits/381c7de1ca69fb50fef187ed624d0e182f69f408). I was hesitant about removing an existing test, but if you also don't really see a purpose for this test anymore, we can remove it. I think the test `ReadWriteScalarDataWIthSubcycling` tests the functionality of `DoNothingWithSubcycling` and even more.

BenjaminRodenberg:
> > Flite and Calculix testcase
> 
> That's still one from Bernhard. Not sure whether this one is about subcycling mainly. I guess you have a better overview.

Opened an issue. I think we should delete it, because I don't really see why we need it. https://github.com/precice/precice/issues/1214

tests/serial/time/implicit/parallel-coupling/ReadWriteScalarDataWithSubcycling.cpp
BenjaminRodenberg: Done in [c0550e4](https://github.com/precice/precice/pull/1186/commits/c0550e4ace598f11dab8d929dca8e4f90e34c832).

BenjaminRodenberg:
> Maybe testing a `CompositionalCouplingScheme` case? Not much more comes to mind.

Added a very simple for-sake-of-completeness test in [29c6ae2](https://github.com/precice/precice/pull/1186/commits/29c6ae29a1dcdc2ed65386ffe3524b6aed204cbe). Doing something more exciting would require rewriting most of the test and we are testing compositional coupling schemes somewhere else. I think to make sure that subcycling does not do something strange this should be enough (e.g. counting iterations twice or not at all, if a compositional scheme is used).

PR #1191, opened 2022-03-02
Allow SolverInterface reconstruction

fsimonis:
## Main changes of this PR

This PR changes the EventUtils to allow the reconstruction of a SolverInterface.
The EventRegistry now only calls MPI related functionality up to including `EventRegistry::finalize()` and allows to be reinitialized.

Currently, writing such tests requires to pass the MPI comm to preCICE using the SolverInterface constructor.

This allows to reconstruct SolverInterfaces:

```cpp
SolverInterface interface(...);
interface.initialize();
interface.finalize();

SolverInterface interface2(...);
...
```
or

```cpp
{
  SolverInterface interface(...);
  interface.initialize();
  ...
}
{
  SolverInterface interface(...);
  interface.initialize();
  ...
}
```

If preCICE was compiled with MPI, then you need to manage the MPI lifetime yourself:

```
MPI_Initialize(nullptr, nullptr);
MPI_Comm comm = MPI_COMM_WORLD;
SolverInterface interface(..., &comm);
interface.initialize();
interface.finalize();

SolverInterface interface2(...&comm);
interface2.inittialize();

MPI_Finalise();
```

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/utils/Parallel.hpp
+++ b/src/utils/Parallel.hpp
@@ -10,9 +10,7 @@
 
 #ifndef PRECICE_NO_MPI
 #include <mpi.h>
-#else // not PRECICE_NO_MPI
-#define MPI_COMM_NULL nullptr
-#endif // not PRECICE_NO_MPI
+#endif // no PRECICE_NO_MPI
 
 namespace precice {
 namespace logging {
@@ -52,7 +50,7 @@ public:
     std::vector<AccessorGroup> groups;
 
     /// The native communicator that represents this state
-    Communicator comm = MPI_COMM_NULL;
+    Communicator comm;
 
     /// A shared pointer to the parent CommState
     CommStatePtr parent = nullptr;


src/precice/tests/SerialTests.cpp
```diff
@@ -227,6 +227,31 @@ BOOST_AUTO_TEST_CASE(ConstructAndExplicitFinalize)
...
+BOOST_AUTO_TEST_CASE(ReconstructSameScope)
+{
+  PRECICE_TEST("SolverOne"_on(1_rank), "SolverTwo"_on(1_rank));
+  std::string config = _pathToTests + "lifecycle.xml";
+
+  SolverInterface interface1(context.name, config, context.rank, context.size, &context.comm);
+  interface1.finalize();
+
+  SolverInterface interface2(context.name, config, context.rank, context.size, &context.comm);
+  interface2.finalize();
```
uekerman, line 239: I guess the really critical variant is we also call `initialize()` (e.g. after defining some one-vertex mesh). Could we add such a test?

src/testing/TestContext.hpp
```diff
@@ -166,6 +166,8 @@ class TestContext {
   /// wheather this context is valid or not
   bool invalid = false;
 
+  precice::utils::Parallel::Communicator comm = MPI_COMM_NULL;
```
uekerman, line 169: Could you please add a comment here? One might wonder what comm this is. It's the overall communicator, right?

uekerman (changes requested):
Pushing towards allowing multiple `SolverInterface` instances is a very good thing :+1:

For a few things here, I need help to understand them:

* What does re-initializing the "events" actually mean. Will we overwrite the existing file? Will the new file contain data of both interfaces (if same participant) or only of the last run? Or do we create a new file increasing some counter? I think we should document these things somewhere (where? Maybe in the constructor of `Solverinterface`?) Same questions concerns all other log files as well (iterations, convergence).
* How do we port back the changes to the `EventUtils` class from here to https://github.com/precice/EventTimings ? Do we? 
* > If preCICE was compiled with MPI, then you need to manage the MPI lifetime yourself:
    ```cpp
    MPI_Initialize(nullptr, nullptr); 
    MPI_Comm comm = MPI_COMM_WORLD;
    SolverInterface interface(..., &comm);
    interface.initialize();
    interface.finalize();
    SolverInterface interface2(...&comm);
    interface2.inittialize();
    MPI_Finalise();
    ```
    I do not yet understand this part. Why do we have to pass the MPI communicator here? Normally, if a user initializes MPI, preCICE should just take this one and use `MPI_COMM_WORLD`.

src/precice/tests/SerialTests.cpp
BenjaminRodenberg: I think this test should go into `precice/tests/serial` into an individual file.

src/precice/tests/SerialTests.cpp
```diff
@@ -227,6 +227,31 @@ BOOST_AUTO_TEST_CASE(ConstructAndExplicitFinalize)
...
+}
+
+BOOST_AUTO_TEST_CASE(ReconstructSeparateScope)
```
BenjaminRodenberg, line 242: Same here: Test should go into `precice/tests/serial`.

fsimonis:
> What does re-initializing the "events" actually mean.

In this case: try not to crash.

> Will we overwrite the existing file?

Just implemented a counter that will appear on successive runs.

> Will the new file contain data of both interfaces (if same participant) or only of the last run? Or do we create a new file increasing some counter? I think we should document these things somewhere (where? Maybe in the constructor of Solverinterface?) Same questions concerns all other log files as well (iterations, convergence).

There will be 2 files. One for each run.

> How do we port back the changes to the EventUtils class from here to https://github.com/precice/EventTimings ? Do we?

We need to copy-back the changes at some point.

>  I do not yet understand this part. Why do we have to pass the MPI communicator here? Normally, if a user initializes MPI, preCICE should just take this one and use MPI_COMM_WORLD.

Ah, you are right! preCICE initializes but doesn't take ownership of an existing the MPI session.
This only confuses the tests.

This suffices:
```
MPI_Initialize(nullptr, nullptr); 
SolverInterface interface(...);
interface.initialize();
interface.finalize();
SolverInterface interface2(...);
interface2.inittialize();
MPI_Finalise();
```

src/testing/TestContext.hpp
fsimonis: Done. It is the communicator of the current participant.
So, the one with the requested amount of ranks.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -179,6 +180,19 @@ void SolverInterfaceImpl::configure(
...
+namespace {
+/// Keeps track of SolverInterface initalizations and returns a suffix for the event timings file
+std::string eventSuffix()
+{
+  static int run{0};
+  ++run;
+  if (run == 1) {
+    return "";
+  }
+  return std::to_string(run);
```
uekerman, line 192:
```suggestion
  return std::to_string("-" + run);
```
? No strong opinion.

fsimonis: EventUtils handle this kind of formatting.

fsimonis:
The tests are superseeded by #1874

Current behaviour is to replace events though. Good enough until it is an issue.

PR #1193, opened 2022-03-03
Hide non-API symbols

### LINKED ISSUE #623: Change default symbol visibility to hidden

fsimonis:
## Main changes of this PR

**This requires us to drop support for Ubuntu 18.04 LTS.**

This PR ~~upgrades the CMake version to 3.12 and~~ enables hidden symbols. _I rebased it on develop_

Hiding Non-API symbols consists out of 3 parts:

1. The CMake library
    * We use an object library `preciceInternal` to compile all preCICE sources.
    * We link these sources into a shared or static library `precice`
    * We link `precice-tools` to `precice`, so it needs to use the public API
    * We link `testprecice` to `preciceInternals`, which is not influenced by the symbol visibility
3. Hiding symbols by default
    * We set the symbol and inline visibility of `preciceInternal` to hidden.
5. Exporting only the API
    * We auto-generate an export header and add it to the public headers of preCICE.
    * We annotate all API function with `PRECICE_API` 

Closes #623

## Author's checklist

* [ ] I sticked to CMake version 3.16.

## Reviewers' checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -461,14 +461,25 @@ endif()
 # Setup General Install for:
 # precice - the library
 # precice-tools - the precice binary
+
+# preciceInternal needs to be added to the export set if we are building a static library
+set(precice_install_targets precice)
+if(NOT BUILD_SHARED_LIBS)
+  set(precice_install_targets ${precice_install_targets} preciceInternal)
+endif()
+if(PRECICE_BUILD_TOOLS)
+  set(precice_install_targets ${precice_install_targets} precice-tools)
+endif()
+
 include(GNUInstallDirs)
-install(TARGETS precicelib
+install(TARGETS ${precice_install_targets}
   EXPORT preciceTargets
   LIBRARY DESTINATION ${CMAKE_INSTALL_LIBDIR}
   ARCHIVE DESTINATION ${CMAKE_INSTALL_LIBDIR}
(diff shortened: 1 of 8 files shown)


fsimonis:
Rebased to the current develop.

fsimonis:
Superseeded by #1497 and #1498

PR #1198, opened 2022-03-05
Fixed errors in gradient mapping 

ariguiba:
## Main changes of this PR
- Fix error in Nearest Neighbor Gradient: Offsets are calculated as output - input vertex. 
- Removed conservative constraint for Nearest Neighbor Gradient, as this not applicable with gradient data (Mathematically inconsistent with the definition of conservative mapping, since we *don't* have a conservation of data over the interface)

## Author's checklist

### DIFF
--- a/src/mapping/NearestNeighborGradientMapping.cpp
+++ b/src/mapping/NearestNeighborGradientMapping.cpp
@@ -20,7 +20,7 @@ NearestNeighborGradientMapping::NearestNeighborGradientMapping(
     int        dimensions)
     : NearestNeighborBaseMapping(constraint, dimensions, true, "NearestNeighborGradientMapping", "nng")
 {
-  PRECICE_CHECK(!hasConstraint(CONSERVATIVE), "Nearest Neighbor Gradient is not defined for conservative mapping");
+  PRECICE_CHECK(!hasConstraint(CONSERVATIVE), "Nearest-neighbor-gradient mapping is not implemented using a \"conservative\" constraint. Please select constraint=\""consistent\" or a different mapping method.");
 
   if (hasConstraint(SCALEDCONSISTENT)) {
     setInputRequirement(Mapping::MeshRequirement::FULL);
--- a/src/mapping/NearestNeighborGradientMapping.cpp
+++ b/src/mapping/NearestNeighborGradientMapping.cpp
@@ -20,7 +20,7 @@ NearestNeighborGradientMapping::NearestNeighborGradientMapping(
     int        dimensions)
     : NearestNeighborBaseMapping(constraint, dimensions, true, "NearestNeighborGradientMapping", "nng")
 {
-  PRECICE_CHECK(!hasConstraint(CONSERVATIVE), "Nearest-neighbor-gradient mapping is not implemented using a \"conservative\" constraint. Please select constraint=\"" consistent\" or a different mapping method.");
+  PRECICE_CHECK(!hasConstraint(CONSERVATIVE), "Nearest-neighbor-gradient mapping is not implemented using a \"conservative\" constraint. Please select constraint=\" consistent\" or a different mapping method.");
 
   if (hasConstraint(SCALEDCONSISTENT)) {
     setInputRequirement(Mapping::MeshRequirement::FULL);
--- a/src/mapping/NearestNeighborGradientMapping.cpp
+++ b/src/mapping/NearestNeighborGradientMapping.cpp
@@ -22,7 +22,7 @@ NearestNeighborGradientMapping::NearestNeighborGradientMapping(
 {
   PRECICE_CHECK(!hasConstraint(CONSERVATIVE), "Nearest-neighbor-gradient mapping is not implemented using a \"conservative\" constraint. Please select constraint=\" consistent\" or a different mapping method.");
 
-  PRECICE_WARN(!hasConstraint(SCALEDCONSISTENT), "The scaled-consistent mapping hasn't been specifically tested with nearest-neighbor-gradient. Please avoid using it or chose another mapping method. ");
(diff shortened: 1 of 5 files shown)


src/mapping/NearestNeighborGradientMapping.cpp
```diff
@@ -20,6 +20,8 @@ NearestNeighborGradientMapping::NearestNeighborGradientMapping(
     int        dimensions)
     : NearestNeighborBaseMapping(constraint, dimensions, true, "NearestNeighborGradientMapping", "nng")
 {
+  PRECICE_CHECK(!hasConstraint(CONSERVATIVE), "Nearest Neighbor Gradient is not defined for conservative mapping");
```
DavidSCN, line 23:
```suggestion
  PRECICE_CHECK(!hasConstraint(CONSERVATIVE), "Nearest-neighbor-gradient mapping is not implemented using a \"conservative\" constraint. Please select constraint=\""consistent\" or a different mapping method.");
```
Is scaled consistent working? If it was just never tested we should also safeguard against it and explain that it requires some additional testing.

src/mapping/NearestNeighborGradientMapping.cpp
```diff
@@ -41,12 +43,8 @@ void NearestNeighborGradientMapping::onMappingComputed(mesh::PtrMesh origins, me
...
     const auto &sourceVertexCoords  = origins->vertices()[i].getCoords();
 
-    // We calculate the distances uniformly for both mapping constraints as the difference (input - output)
-    if (hasConstraint(CONSERVATIVE)) {
-      _offsetsMatched[i] = sourceVertexCoords - matchedVertexCoords;
-    } else {
-      _offsetsMatched[i] = matchedVertexCoords - sourceVertexCoords;
-    }
+    // We calculate the distances uniformly for consistent mapping constraint as the difference (output - input)
+    _offsetsMatched[i] = sourceVertexCoords - matchedVertexCoords;
```
DavidSCN, line 47: The comment claims output - input, but you write 'source - matched', is this correct?

DavidSCN (commented):
Looks good, thank you!

Only comment: you never tested the scaled-consistent NNG mapping and we don't have tests for the method at all, right? It might work already with the current implementation, but we should safeguard against its usage as long as we haven't tested it (which doesn't need to be your concern right now).

Please also add a changelog entry.

src/mapping/NearestNeighborGradientMapping.cpp
ariguiba: The scaled consistent mapping is defined in the parent class 'Mapping' so I assumed that it would work exactly the same for NN, NP and NNG for now. But I can add a comment to clearify this.

src/mapping/NearestNeighborGradientMapping.cpp
DavidSCN: You could also write your comment more explicit in this regard so that its clear to everyone.

DavidSCN:
What about the scaled-consistent case (see above)?

src/mapping/NearestNeighborGradientMapping.cpp
```diff
@@ -20,7 +20,7 @@ NearestNeighborGradientMapping::NearestNeighborGradientMapping(
     int        dimensions)
     : NearestNeighborBaseMapping(constraint, dimensions, true, "NearestNeighborGradientMapping", "nng")
 {
-  PRECICE_CHECK(!hasConstraint(CONSERVATIVE), "Nearest-neighbor-gradient mapping is not implemented using a \"conservative\" constraint. Please select constraint=\" consistent\" or a different mapping method.");
+  PRECICE_CHECK(!hasConstraint(CONSERVATIVE), "Nearest-neighbor-gradient mapping is not implemented using a \"conservative\" constraint. Please select constraint=\" consistent\" or a different mapping method. Note that the scaled-consistent method reverts to the method implemented for all mappings but hasn't been specifically tested with nearest-neighbor-gradient.");
```
DavidSCN, line 23: I understand what you want to say, but the user certainly won't, as he or she is not familiar with the implementation details. Also, this warning does not show up in case the user selects `constraint=scaled-consistent`. What we need is something like `PRECICE_WARN(!hasConstraint(SCALEDCONSISTENT), ...` or `PRECICE_CHECK(...` telling the user that the scaled-consistent variant of the NNG mapping has not been tested.

src/mapping/NearestNeighborGradientMapping.cpp
```diff
@@ -20,6 +20,10 @@ NearestNeighborGradientMapping::NearestNeighborGradientMapping(
     int        dimensions)
     : NearestNeighborBaseMapping(constraint, dimensions, true, "NearestNeighborGradientMapping", "nng")
 {
+  PRECICE_CHECK(!hasConstraint(CONSERVATIVE), "Nearest-neighbor-gradient mapping is not implemented using a \"conservative\" constraint. Please select constraint=\" consistent\" or a different mapping method.");
+
+  PRECICE_CHECK(!hasConstraint(SCALEDCONSISTENT), "The scaled-consistent mapping hasn't been specifically tested with nearest-neighbor-gradient. Please avoid using it or choose another mapping method. ");
```
DavidSCN, line 25: If you use `PRECICE_CHECK` there is no chance for the user to 'avoid using it', as preCICE will anyway crash all the time. Maybe better use `PRECICE_WARN` and just warn the user that it is not tested.

DavidSCN, line 25: If you use `PRECICE_CHECK` there is no chance for the user to 'avoid using it', as preCICE will anyway crash all the time. Maybe better use `PRECICE_WARN` and just warn the user that it is not tested.

DavidSCN:
Thanks!

PR #1199, opened 2022-03-07
Add script for getting all authors contributing to organization.

BenjaminRodenberg:
## Main changes of this PR

Adds a script to parse all repos of an orgnization for commit authors.

Must not end up in this repository and not necessarily in `tools`. @MakisH you probably have a better overview.

Helps us to create a list with names of all contributors.

## Reviewers' checklist

### DIFF
--- a/tools/community/list_collabs.py
+++ b/tools/community/list_collabs.py
@@ -1,57 +1,77 @@
-import argparse
+"""
+Usage: list_collabs.py [--user USER] [--token TOKEN]
+
+Create a list of all collaborators that commited to an organization. Provide user name and access token, if you need to
+send many requests.
+
+  -h --help         show this
+  -u --user USER    github user name
+  -t --token TOKEN  github token (see https://docs.github.com/en/rest/overview/other-authentication-methods#basic-authentication)
+"""
+from docopt import docopt
 import requests
 from requests.auth import HTTPBasicAuth
 
-# use authentication, if you want to get a lot of results. Github limits output for unauthenticated users. See https://docs.github.com/en/rest/overview/other-authentication-methods#basic-authentication
-parser = argparse.ArgumentParser()
-parser.add_argument('-u', '--user', type=str)
-parser.add_argument('-t', '--token', type=str)
-args = parser.parse_args()
+if __name__ == '__main__':
+    organization = "precice"
+    arguments = docopt(__doc__)
 
-organization = "precice"
+    commit_authors = set()
+    skipped_commits = []
+    number_commits = 0
 
-commit_authors = set()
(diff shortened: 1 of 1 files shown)


MakisH:
Thank you, this is useful! Can you also add a `README.md` on what it does and how to use it?

BenjaminRodenberg:
> Thank you, this is useful! Can you also add a `README.md` on what it does and how to use it?

I added documentation for the `--help` flag. Providing an individual `README.md` means additional maintainment and is not really needed just for a script in my opinion. Therefore, I would like to avoid it.

tools/community/list_collabs.py
```diff
@@ -0,0 +1,77 @@
...
+            # Traverse all commits on page
+            for commit in response.json():
+                number_commits += 1
+                try:
+                    # extract name of author
+                    author = commit['author']['login']
+                    commit_authors.add(author)
+                except TypeError:
+                    # need to skip many commits. Not sure why.
+                    skipped_commits.append(commit)
```
MakisH, line 73: Would be good to store these commits in a file or show them with some option. Running this I got 1949 skipped commits, a number that sounds a bit alarming and generates questions.

tools/community/list_collabs.py
```diff
@@ -0,0 +1,77 @@
...
+
+        # While page has entries, continue
+        while len(response.json()) > 0:
+            print("On page {} got {} hits.".format(pageno, len(response.json())))
+            pageno += 1
+            api_url = "https://api.github.com/repos/{}/{}/commits?per_page=100&page={}".format(organization, repo, pageno)
+            if '--user' in arguments.keys():
+                response = requests.get(api_url, auth=HTTPBasicAuth(arguments['--user'], arguments['--token']))
+            else:
+                response = requests.get(api_url)
```
MakisH, line 63: We should also check the response for being valid. When running without a user/password, I got stuck in a "On page (increasing) got 2 hits.", but no error. I had to guess that this is related to the access.

MakisH:
> > Thank you, this is useful! Can you also add a `README.md` on what it does and how to use it?
> 
> I added documentation for the `--help` flag. Providing an individual `README.md` means additional maintainment and is not really needed just for a script in my opinion. Therefore, I would like to avoid it.

But every other folder in `tools/` has a `README.md`, explaning what these scripts are.
I am not sure this is the right place, but I can also not think of any better place at the moment.

MakisH:
@BenjaminRodenberg I still believe that this is not really the right place for this script. Important is that it stays documented somewhere. My original plan was to make a different script that can give us a ratio of external/total contributions. But this had merely academic intention.

PR #1148, opened 2021-12-09
Provide a tool for generation of integration tests

BenjaminRodenberg:
## Main changes of this PR

Provides a developer tool to generate a skeleton for integration tests.

## Motivation and additional information

We want to restructure the integration tests and this tool will help us. See #1021.

This PR was extracted from #1140, for more details, please refer to #1140.

## Author's checklist

* [ ] I added a changelog file with this PR number in `docs/changelog/` if there are noteworthy changes.
* [ ] I ran `tools/formatting/format-all` to ensure everything is formatted correctly.
* [ ] I sticked to C++14 features.
* [ ] I sticked to CMake version 3.10.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

Issue #1149, opened 2021-12-10
Move all integration tests to new structure


BenjaminRodenberg:
**Please describe the problem you are trying to solve.**

As discussed in #1021, we want to move all integration tests in `precice/src/precice/tests` to the new structure proposed and implemented in #1148. This issue gives an overview over the progress. Please mention this issue in related PRs.

**Describe the solution you propose.**

See #1148 and the proposed structure for the integration tests. As a first step we only want to apply the new structure to all tests to make it easier to rearrange individual tests. See https://github.com/precice/precice/issues/1021#issuecomment-976698077.

PR #1202, opened 2022-03-10
Add integration test suite detection

fsimonis:
## Main changes of this PR

This PR adds the detection of integration test suites.

So all directories in `tests/` will be converted to CamelCase and available as ctests called `precice.integration.CamelCase`.

The createTest script now also reminds the user to update the source index.

Follow-up of #1148

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/cmake/CTestConfig.cmake
+++ b/cmake/CTestConfig.cmake
@@ -277,16 +277,6 @@ add_precice_test(
   TIMEOUT ${PRECICE_TEST_TIMEOUT_SHORT}
   )
 add_precice_test(
-  NAME serial
-  ARGUMENTS "--run_test=PreciceTests/Serial"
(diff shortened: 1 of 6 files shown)


BenjaminRodenberg:
@DavidSCN and I will use this directly, when working on #1149. This should suffice as a review.

DavidSCN (commented):
Tested it and works in principle. General question: we also have a `tools/testing` directory. Wouldn't it be better to place the test related scripts therein?

BenjaminRodenberg (approved):
We can merge this PR from my perspective. Moving the script is a good point, but not extremely critical. I will merge this PR latest today afternoon, because #1149 depends on it.

DavidSCN:
@BenjaminRodenberg do you think this change might be worth a changelog?

PR #1203, opened 2022-03-13
Finer classification of mesh connectivity requirements

boris-martin:
## Main changes of this PR

Instead of storing whether vertices are sufficient or if "full connectivity" is required, we define connectivity requirements through the "most complex primitive required": vertices, edges, surfaces, or (in the future with volume coupling), volume elements. There is a total ordering:  `VOLUME > SURFACE > EDGE > VERTEX > UNDEFINED`.

For instance, in the current state, if only "VERTEX" are required, using `setMeshEdges` and `setMeshTriangles` will do nothing. To have them do something, the requirement "FULL" is needed. Now, we can ask for only edges but no triangles (e.g. in surface coupling in 2D), which will disable `setMeshTriangles` but not `setMeshEdges`.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mapping/Mapping.cpp
+++ b/src/mapping/Mapping.cpp
@@ -130,14 +130,7 @@ bool Mapping::hasConstraint(const Constraint &constraint) const
 
 bool operator<(Mapping::MeshRequirement lhs, Mapping::MeshRequirement rhs)
 {
-  switch (lhs) {
-  case (Mapping::MeshRequirement::UNDEFINED):
-    return rhs != Mapping::MeshRequirement::UNDEFINED;
-  case (Mapping::MeshRequirement::VERTEX):
-    return rhs == Mapping::MeshRequirement::FULL;
-  case (Mapping::MeshRequirement::FULL):
-    return false;
-  };
+  return int(lhs) < int(rhs);
   BOOST_UNREACHABLE_RETURN(false);
 }
 
@@ -149,6 +142,15 @@ std::ostream &operator<<(std::ostream &out, Mapping::MeshRequirement val)
     break;
   case (Mapping::MeshRequirement::VERTEX):
     out << "VERTEX";
+    break;
+  case (Mapping::MeshRequirement::EDGE):
+    out << "EDGE";
+    break;
+  case (Mapping::MeshRequirement::SURFACE):
+    out << "SURFACE";
+    break;
+  case (Mapping::MeshRequirement::VOLUME):
+    out << "VOLUME";
     break;
   case (Mapping::MeshRequirement::FULL):
     out << "FULL";
--- a/src/mapping/Mapping.hpp
+++ b/src/mapping/Mapping.hpp
@@ -42,8 +42,14 @@ public:
     UNDEFINED = 0,
     /// Vertices only.
     VERTEX = 1,
+    // Edges also required (e.g. surface coupling in 2D nearest projection)
+    EDGE = 2,
+    // Surfaces required (e.g. surface coupling in 3D nearest projection, or volumetric NP in 2D)
+    SURFACE = 3,
+    // Volume data required (e.g. volumetric 3D coupling)
+    VOLUME = 4,
     /// Full mesh.
-    FULL = 2
+    FULL = 5
   };
 
   /// Constructor, takes mapping constraint.
--- a/src/mapping/NearestNeighborGradientMapping.cpp
+++ b/src/mapping/NearestNeighborGradientMapping.cpp
@@ -21,8 +21,14 @@ NearestNeighborGradientMapping::NearestNeighborGradientMapping(
     : NearestNeighborBaseMapping(constraint, dimensions, true, "NearestNeighborGradientMapping", "nng")
 {
   if (hasConstraint(SCALEDCONSISTENT)) {
-    setInputRequirement(Mapping::MeshRequirement::FULL);
-    setOutputRequirement(Mapping::MeshRequirement::FULL);
+    // Full requirements for surface integrals: edges in 2D, surfaces in 3D
+    if (getDimensions() == 2) {
+      setInputRequirement(Mapping::MeshRequirement::EDGE);
+      setOutputRequirement(Mapping::MeshRequirement::EDGE);
+    } else {
+      setInputRequirement(Mapping::MeshRequirement::SURFACE);
+      setOutputRequirement(Mapping::MeshRequirement::SURFACE);
+    }
   } else {
     setInputRequirement(Mapping::MeshRequirement::VERTEX);
     setOutputRequirement(Mapping::MeshRequirement::VERTEX);
--- a/src/mapping/NearestNeighborMapping.cpp
+++ b/src/mapping/NearestNeighborMapping.cpp
@@ -20,11 +20,14 @@ NearestNeighborMapping::NearestNeighborMapping(
     : NearestNeighborBaseMapping(constraint, dimensions, false, "NearestNeighborMapping", "nn")
 {
   if (hasConstraint(SCALEDCONSISTENT)) {
-    setInputRequirement(Mapping::MeshRequirement::FULL);
-    setOutputRequirement(Mapping::MeshRequirement::FULL);
-  } else {
-    setInputRequirement(Mapping::MeshRequirement::VERTEX);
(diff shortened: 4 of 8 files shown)


fsimonis:
The mesh requirement now mixes two concerns: the required primitive and the required kind of connectivity.

Think of it this way:
```
kind + mesh dimensionality = primitive
```

Kind | Dims | Primitive
--- | --- | ---
Points | - | Vertex
Surface | 2D | Edge
Surface | 3D | Triangle
Volume | 2D | Triangle
Volume | 3D | Tetrahedron

We should choose the one or the other:

**Using the kind** makes the code more expressive as a component can request what kind of connectivity it needs, independent of the dimensionality of the mesh. The kind only makes sense together with a dimensionality.
**Using the primitive** simplifies the code in API calls etc as the requirement directly contains the required primitive.

src/mapping/Mapping.cpp
```diff
@@ -130,14 +130,7 @@ bool Mapping::hasConstraint(const Constraint &constraint) const
...
 {
-  switch (lhs) {
-  case (Mapping::MeshRequirement::UNDEFINED):
-    return rhs != Mapping::MeshRequirement::UNDEFINED;
-  case (Mapping::MeshRequirement::VERTEX):
-    return rhs == Mapping::MeshRequirement::FULL;
-  case (Mapping::MeshRequirement::FULL):
-    return false;
-  };
+  return int(lhs) < int(rhs);
```
fsimonis, line 133: Let's use a `static_cast<int>` here to make this more explicit.

Also, you should now enforce the underlying type of the enum to be int: 
```cpp
enum class MeshRequirement : int {
```

boris-martin:
It's probably best to have both? An adapter could deduce the "kind" from dimension+primitive but it's probably much better to read the kind. Internally the kind->primitive conversion is easy to do for overriding primitives creation.

Overall I'd be in favor of the "kind" choice. I'm not sure how to integrate the scaled-consistent case though. Is it generalizable to volume coupling? Should it always request surface and throw an error when used in volume coupling? I don't fully understand this part.

On a side note, are "Surface" and "Volume" the best choice of words for the "kind"? Talking about something like "boundary" vs "full/whole domain" seems less confusing and more dimension-independent. But well we've been talking about "surface coupling" for a while.
Right now I call primitives requirements with edge/surface/volume instead of edge/triangle/tetrahedron to take in account other elements (e.g. quads). If we keep this changing the terminology for kind would be welcome.

fsimonis:
> It's probably best to have both? An adapter could deduce the "kind" from dimension+primitive but it's probably much better to read the kind. Internally the kind->primitive conversion is easy to do for overriding primitives creation.

The adapter (as in using the API) can only ask "preCICE, do you need connectivity from me?".
preCICE then decides based on the mesh requirement which primitives to ignore in `setMeshX` calls.

> Overall I'd be in favor of the "kind" choice. I'm not sure how to integrate the scaled-consistent case though. Is it generalizable to volume coupling? Should it always request surface and throw an error when used in volume coupling? I don't fully understand this part.

@uekerman Please correct me if I am wrong, but this should then be the same, but for volumes instead of surfaces.

@boris-martin anyhow, let's ignore the scaled consistent for now.
 
> On a side note, are "Surface" and "Volume" the best choice of words for the "kind"? Talking about something like "boundary" vs "full/whole domain" seems less confusing and more dimension-independent. But well we've been talking about "surface coupling" for a while. Right now I call primitives requirements with edge/surface/volume instead of edge/triangle/tetrahedron to take in account other elements (e.g. quads). If we keep this changing the terminology for kind would be welcome.

@uekerman I think this kind of wording is a matter for you.

uekerman:
> @uekerman Please correct me if I am wrong, but this should then be the same, but for volumes instead of surfaces.

Yes, but let's indeed ignore this one first and exclude it for the moment. No high priority.

> > On a side note, are "Surface" and "Volume" the best choice of words for the "kind"? Talking about something like "boundary" vs "full/whole domain" seems less confusing and more dimension-independent. But well we've been talking about "surface coupling" for a while. Right now I call primitives requirements with edge/surface/volume instead of edge/triangle/tetrahedron to take in account other elements (e.g. quads). If we keep this changing the terminology for kind would be welcome.
> @uekerman I think this kind of wording is a matter for you.

I must admit I am a bit confused by your discussion, but I just try to put it in my words.

We should distinguish between surface and volume coupling. This "kind" (? ... we need a better name) should be deduced from the used mapping (if applicable). And then we have everything, we need to distinguish the requirements, no?

boris-martin:
Ok for skipping the scaled-consistent for now.
@uekerman it is still a discussion whether to consider volume NP as an option of NP or a whole mapping. But in both cases, Mapping could have a get"Kind" easily. (Except that NN and RBF can be used for both kinds.)
Once you have kind and dimension, requirements are computable.

But from an adapter perspective, asking "should I configure the surface or the volume" makes more sense than "do I need to add some triangles"?


(Actually, for scaled consistent, we might need to add a kind flag to NN and RBF too, I think)

uekerman:
From adapter perspective, you should already know whether a mesh is a surface or a volume mesh. This information should not come from preCICE. Think of the CalculiX adapter. When iterating over the mesh, you already know what it is. This should be triggered from the adapter config. 
From preCICE, you only need to learn whether defined mappings require connectivity or not.

A tricky thing is probably still the question: could a mesh be used both in a surface and a volume mapping? I guess currently no (maybe with the work of @MakisH towards geometric multiscale mapping, at some point, yes). We should forbid this currently.

> @uekerman it is still a discussion whether to consider volume NP as an option of NP or a whole mapping

My tendency goes towards as a separate mapping, also with a name "without projection". But one could argue here.

uekerman:
From adapter perspective, you should already know whether a mesh is a surface or a volume mesh. This information should not come from preCICE. Think of the CalculiX adapter. When iterating over the mesh, you already know what it is. This should be triggered from the adapter config. 
From preCICE, you only need to learn whether defined mappings require connectivity or not.

A tricky thing is probably still the question: could a mesh be used both in a surface and a volume mapping? I guess currently no (maybe with the work of @MakisH towards geometric multiscale mapping, at some point, yes). We should forbid this currently.

> @uekerman it is still a discussion whether to consider volume NP as an option of NP or a whole mapping

My tendency goes towards as a separate mapping, also with a name "without projection". But one could argue here.

fsimonis:
@uekerman @boris-martin For clarification:

These mesh requirement are for internal use only. The only API call exposed to adapter is `isConnectivityRequired()`.

The mesh requirements are used to filter unneeded connectivity information directly at the API `setMeshC` calls. Needed connectivity is derived from the configuration of preCICE.

Example
```
nearest-neighbour mapping on 3d mesh -> the mesh requires only vertices
watch-integral on the same mesh -> the mesh now accepts triangles (and edges)
```

fsimonis:
My suggestion to move on with this:

1. we use the MeshRequirement closest to the original meaning: "what is the most specific primitive I need"
3. we add a `protected: void Mapping::deduceRequirements() = 0 virtual` which uses `Mapping::inputMesh()` and `Mapping::outputMesh()` to deduce the requirements and sets them using `Mapping::set(In|Output)Requirement`.
3. we call this after we set the meshes in `Mapping::setMeshes()`.
4. we implement `deduceRequirements` for the various Mappings and remove the calls to `set(In|Out)putRequirement` from the constructors of the mapping implementations.

MakisH:
> A tricky thing is probably still the question: could a mesh be used both in a surface and a volume mapping? I guess currently no (maybe with the work of @MakisH towards geometric multiscale mapping, at some point, yes). We should forbid this currently.

Indeed, coupling all the points of a 1D domain with the interface of a 3D domain (radial mapping) would mean that we couple a surface (boundary of the 3D) with a volume (complete domain of the 1D). We actually already have this example in the generator-propagator training code.

fsimonis:
@uekerman @MakisH

The combination of surface and volume coupling is immaterial to the MeshContext.
The only purpose of the MeshRequirement is to avoid unnecessary connectivity in the Meshes.

Let's end this discussion by implementing the MeshRequirement as the required primitives.
This also moves unnecessary logic out of the SolverInterface.

boris-martin:
Closing as this is way too old. Opened an issue to document it.

PR #1207, opened 2022-03-19
Made triangle barycentric coordinates possible in 2D 

boris-martin:
## Main changes of this PR

- `precice::math::barycenter::calcBarycentricCoordsForTriangle` now works with triangles in a 2D mesh.
- Added a unit test for correctness. (A copy of the existing one in 3D but with the Z-component removed, basically)
- Removed the assertion `dimensions == 3` in the Triangle constructor.

Prerequisite to volume coupling.

## Author's checklist

## Reviewers' checklist

@fsimonis is a changelog needed here? I guess not but I didn't check whether there's somewhere written that you can't add a triangle in 2D where it should be changed.
The algorithm for 3D seemed hard to adapt in 2D so I designed a new one with pen & paper. Not sure it is optimal but I think it is cheap enough.
However, I'm not sure whether an if-else in the function is a better design choice than two separate functions or not.


src/math/barycenter.cpp
```diff
@@ -41,35 +41,58 @@ Eigen::Vector2d calcBarycentricCoordsForEdge(
...
+  } else {
+    // Dimension == 2 => No cross product
+    Vector2d ab, ac, ub, uc, ua;
+    ab = b - a;
+    ac = c - a;
+    ub = b - u;
+    uc = c - u;
+    ua = a - u;
+
+    scaleFactor          = 1.0 / crossProduct2D(ab, ac);
```
fsimonis, line 91: This will crash for zero vectors. Can you assert this above?

boris-martin: It wasn't checked in 3D either, so I just added assertions for both cases.

boris-martin:
@fsimonis I just realized in (surface) NP there is a test case that uses a degenerate triangle, so my new assertion makes the test fail.
https://github.com/precice/precice/blob/574969099df887429ecceec495603ec35281b040/src/mapping/tests/NearestProjectionMappingTest.cpp#L343-L349

Is that intended?

fsimonis:
The API currently allows such triangles, so we should we support them here as well.

Maybe we could/should extend the checks of the input mesh.

fsimonis:
Ok, supporting degenerate triangles is nonsense here.

I propose to do the following in this PR:
* Assert that the triangles are "valid" in the projection logic. (as you did)
* Remove tests testing degenerate triangles.

In another PR:
* Check the triangles to be valid in the `SolverInterface` and raise a `PRECICE_ERROR`. I'll do this asap.

boris-martin:
I removed the degenerate test and all checks have passed :)

fsimonis:
Perfect. I'll merge this and prepare additional checks in the API.

PR #1212, opened 2022-03-21
Fixed outdated documentation

boris-martin:
Quick fix of a comment that no longer reflects implementation.

Implementation of `Triangle::operator==`:
https://github.com/precice/precice/blob/502d1aa6a278131908871ef9f54cbac2ba24cc86/src/mesh/Triangle.cpp#L109-L113

There is no comparison of the normal vector anymore.


fsimonis:
Thanks for the fix!

PR #1216, opened 2022-03-23
Simplify implementation of data mapping in SolverInterfaceImpl

BenjaminRodenberg:
## Main changes of this PR

Creates function `isMappingRequired` and uses `mapData` everywhere, where mapping is performed.

Branched off from #1187 

## Author's checklist

* [x] Move `isMappingRequired` into `DataContext`?
* [x] Move `mapData` into `DataContext`?
* [x] Move most of the implementation of `SolverInterfaceImpl::resetWrittenData` into `DataContext`?

## Reviewers' checklist

### DIFF
--- a/src/precice/impl/DataContext.cpp
+++ b/src/precice/impl/DataContext.cpp
@@ -98,10 +98,10 @@ bool DataContext::isMappingRequired()
     return false;
   }
 
-  auto timing    = _mappingContext.timing;
+  auto       timing    = _mappingContext.timing;
   const bool hasMapped = _mappingContext.hasMappedData;
-  const bool mapNow    = (timing == MappingConfiguration::ON_ADVANCE) |= 
-                         (timing == MappingConfiguration::INITIAL);
+  const bool mapNow    = (timing == MappingConfiguration::ON_ADVANCE) |=
+      (timing == MappingConfiguration::INITIAL);
 
   if ((not mapNow) || hasMapped) {
     return false;
--- a/src/precice/impl/DataContext.cpp
+++ b/src/precice/impl/DataContext.cpp
@@ -26,12 +26,6 @@ std::string DataContext::getDataName() const
   return _providedData->getName();
 }
 
-int DataContext::getProvidedDataID() const
-{
-  PRECICE_ASSERT(_providedData);
-  return _providedData->getID();
-}
-
 int DataContext::getFromDataID() const
 {
   PRECICE_TRACE();
@@ -42,23 +36,12 @@ int DataContext::getFromDataID() const
 
 void DataContext::resetData()
 {
-  resetProvidedData();
+  // See also https://github.com/precice/precice/issues/1156.
+  _providedData->toZero();
   if (hasMapping()) {
     PRECICE_ASSERT(hasWriteMapping());
-    resetToData();
+    _toData->toZero();
   }
-}
-
-void DataContext::resetProvidedData()
-{
-  PRECICE_TRACE();
-  _providedData->toZero();
-}
-
-void DataContext::resetToData()
-{
-  PRECICE_TRACE();
-  _toData->toZero();
 }
 
 int DataContext::getToDataID() const
@@ -114,8 +97,8 @@ bool DataContext::isMappingRequired()
     return false;
   }
 
-  auto timing    = mappingContext().timing;
-  bool hasMapped = mappingContext().hasMappedData;
+  auto timing    = _mappingContext.timing;
+  bool hasMapped = _mappingContext.hasMappedData;
   bool mapNow    = timing == MappingConfiguration::ON_ADVANCE;
   mapNow |= timing == MappingConfiguration::INITIAL;
 
@@ -128,10 +111,11 @@ bool DataContext::isMappingRequired()
 
 void DataContext::mapData()
 {
+  PRECICE_ASSERT(hasMapping());
   int fromDataID = getFromDataID();
   int toDataID   = getToDataID();
-  resetToData();
-  mappingContext().mapping->map(fromDataID, toDataID);
+  _toData->toZero();
+  _mappingContext.mapping->map(fromDataID, toDataID);
 }
 
 bool DataContext::hasReadMapping() const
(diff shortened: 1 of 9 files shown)


BenjaminRodenberg:
@uekerman I realized that we could also move `mapData` into `DataContext`. I think we already had this discussion previously, but it would allow us to delete some more functions (`mappingContext()`) of `DataContext` or make them private and avoid indirections (`context.mappingContext().mapping->map(fromDataID, toDataID)`).

I'm also fine with the state as it is, but I currently do not really see why moving `isMappingRequired` is fine and moving `mapData` is not.

src/precice/impl/DataContext.cpp
```diff
@@ -104,6 +104,27 @@ bool DataContext::hasMapping() const
...
+{
+  using namespace mapping;
+  MappingConfiguration::Timing timing;
+
+  if (not hasMapping()) {
+    return false;
+  }
+
+  timing         = mappingContext().timing;
+  bool hasMapped = mappingContext().hasMappedData;
```
fsimonis, line 117: Little simplification

```suggestion
  if (not hasMapping()) {
    return false;
  }

  auto timing    = mappingContext().timing;
  bool hasMapped = mappingContext().hasMappedData;
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1571,31 +1564,13 @@ void SolverInterfaceImpl::computeMappings(const utils::ptr_vector<MappingContext
...
-  int inDataID  = context.getFromDataID();
-  int outDataID = context.getToDataID();
+  int fromDataID = context.getFromDataID();
+  int toDataID   = context.getToDataID();
   PRECICE_DEBUG("Map \"{}\" data \"{}\" from mesh \"{}\"",
                 mappingType, context.getDataName(), context.getMeshName());
   context.resetToData();
-  PRECICE_DEBUG("Map from dataID {} to dataID: {}", inDataID, outDataID);
-  context.mappingContext().mapping->map(inDataID, outDataID);
-  PRECICE_DEBUG("Mapped values = {}", utils::previewRange(3, context.toData()->values())); // @todo might be better to move this debug message into Mapping::map and remove getter DataContext::toData()
```
DavidSCN, line 1598: Why do you remove this DEBUG print? I think it might be useful. I think it might be useful.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1040,11 +1037,7 @@ void SolverInterfaceImpl::mapReadDataTo(
       if (context.getMeshID() != toMeshID) {
         continue;
       }
-      context.resetToData();
-      PRECICE_DEBUG("Map data \"{}\" to mesh \"{}\"", context.getDataName(), context.getMeshName());
-      PRECICE_ASSERT(mappingContext.mapping == context.mappingContext().mapping);
```
DavidSCN, line 1045: Why removing the Assert here?

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRodenberg: Good that you saw this. I removed it because it's the only place where `DataContext::toData()` is actually used. Maintaining this public function just for a debug statement is in my opinion too much and therefore I removed the function and the debug statement. Also fixed this now in 39910cc237a889e9a6cef862dd4e967e9f742bce and cleaned it up.

If somebody really needs to debug here, this can also happen somewhere else where this data is actually used and accessed, for example inside the mapping.

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRodenberg: No specific reason. I added it again and I guess we can keep it. See fc50bcb81df7829d0aef598643de7de0cb6278fe. It might still be a good idea to remove it at a later point in time when more functionality is moved inside `DataContext`.

src/precice/impl/SolverInterfaceImpl.cpp
uekerman: IIRC we added these Debug statements here in a dedicated effort by @fsimonis 
These Debug statements are really important when developing an adapter. Let's please not remove them.

uekerman (changes requested):
Looks good expect the removed mapping preview.

uekerman:
> @uekerman I realized that we could also move mapData into DataContext. I think we already had this discussion previously, but it would allow us to delete some more functions (mappingContext()) of DataContext or make them private and avoid indirections (context.mappingContext().mapping->map(fromDataID, toDataID)).

Both are OK. In a previous version you hid the mapping completely in the `DataContext` class, meaning that there was nothing like "map data" in `SolverInterfaceImpl`. This, I didn't like because it should be easy to discover where the mapping happens when skimming `SolverInterfaceImpl`. This is no longer the case in https://github.com/precice/precice/pull/1217, so no concerns from my side.

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRodenberg: Moved the debug statement inside `DataContext` then we also don't need `DataContext::toData()`.

BenjaminRodenberg:
@DavidSCN can you take another look at this PR?

src/precice/impl/DataContext.cpp
```diff
@@ -104,6 +91,35 @@ bool DataContext::hasMapping() const
...
+{
+  using namespace mapping;
+  if (not hasMapping()) {
+    return false;
+  }
+
+  auto timing    = _mappingContext.timing;
+  bool hasMapped = _mappingContext.hasMappedData;
+  bool mapNow    = timing == MappingConfiguration::ON_ADVANCE;
+  mapNow |= timing == MappingConfiguration::INITIAL;
```
DavidSCN, line 104: I understand that you just copied this implementation from the original code snippet(s), but I already stumbled across this once: For me the code would be much better readable using 
```suggestion
  const bool hasMapped = _mappingContext.hasMappedData;
  const bool mapNow    = (timing == MappingConfiguration::ON_ADVANCE) |= 
                         (timing == MappingConfiguration::INITIAL);
```
because it reveals the dependencies of `mapNow` more obvious. However, opinions might be different and I just mention it here for the sake of completeness.

DavidSCN (commented):
Only a minor comment, otherwise feel free to merge.

uekerman (approved):

PR #1218, opened 2022-03-24
Add Quicktest for simpler integration tests

fsimonis:
## Main changes of this PR

This PR adds a helper class called QuickTest. It allows to quickly write simple integration tests using a single mesh per participant without having to handle vertex ids, storage etc.
`read()`s and `write()`s infer the sizes from the argument.
Reads will be saved in the `reads` vector for later inspection. You can also call `.last()` to inspect the latest read data, which is useful for checking data directly after `read()`s

Basic use-case:

1. Create a solverinterface as usual
   ```cpp
   SolverInterface interface{...};
   ```
2. Create a quicktest for a mesh
   ```cpp
   QuickTest qt{interface, "MeshOne"};
   ```
3. set your mesh and initialize
   ```cpp
   qt.setVertices({0.0, 0.0, 0.0,
                   1.0, 0.0, 0.0})
     .initialize();
   ```
4. read, write, advance as usual
   ```cpp
   qt.initializeData()
     .read("DataTwo"_vector)
     .write("DataOne"_scalar, {1, 2})
     .advance()
     .read("DataTwo"_vector)
     .finalize();
   ```
5. test read data
   ```cpp
   auto expected0 = {1, 1, 1, 4, 4, 4};
   auto expected1 = {0, 0, 0, 3, 3, 3};
   BOOST_TEST(qt.reads.at(0) == expected0, boost::test_tools::per_element());
   BOOST_TEST(qt.reads.at(1) == expected1, boost::test_tools::per_element());
   ```

You can chain most calls of `QuickTest` (except `finalize()` for obvious reasons.)
```cpp
QuickTest qt{interface, "MeshTwo"};
qt.setVertices({0.0, 0.0, 0.0,
                0.33, 0.0, 0.0,
                0.66, 0.0, 0.0,
                1.0, 0.0, 0.0})
    .initialize()
    .write("DataTwo"_vector, {1, 1, 1,
                              2, 2, 2,
                              3, 3, 3,
                              4, 4, 4})
    .initializeData()
    .read("DataOne"_scalar)
    .write("DataTwo"_vector, {0, 0, 0,
                              1, 1, 1,
                              2, 2, 2,
                              3, 3, 3})
    .advance()
    .finalize();

auto expected0 = {1, 1, 2, 2};
BOOST_TEST(qt.reads.at(0) == expected0, boost::test_tools::per_element());
```

Reduce boilerplate to quickly write integration tests.

I implemented this to quickly test various scenarios with reinitialization. 

The data arguments could be names only if we had #1183 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/testing/QuickTest.hpp
+++ b/src/testing/QuickTest.hpp
@@ -1,0 +1,136 @@
+#pragma once
+
+#include <boost/test/unit_test.hpp>
+#include <cmath>
+#include <precice/SolverInterface.hpp>
+#include <string>
+#include <vector>
+
+namespace precice {
+namespace testing {
+
+struct QuickTest {
+
+  /// Represents named data and its type
+  struct Data {
+    std::string name;
+    bool        vectorial;
+  };
+
+  /// Creates quicktest based on one mesh
+  QuickTest(SolverInterface &si, const std::string &mesh)
+      : interface(&si), dims(si.getDimensions()), meshID(si.getMeshID(mesh))
+  {
+  }
+
+  /** Sets mesh vertices based on the given coordinates while saving ids for later access.
+   *
+   * Will automatically deduce the amount of vertices to set.
+   *
+   * @see precice::SolverInterface::setMeshVertices()
+   */
+  QuickTest &setVertices(const std::vector<double> &pos)
+  {
+    int n = static_cast<int>(pos.size()) / dims;
+    BOOST_REQUIRE(n > 0);
+    vertexIDs.resize(n, -1);
+    interface->setMeshVertices(meshID, n, pos.data(), vertexIDs.data());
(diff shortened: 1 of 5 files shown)


DavidSCN (commented):
I understand where this is aiming and code-wise you can merge this from my side. However, I'm not sure in terms of usability, as this introduces an additional layer into the integration tests and this additional layer requires the developer to look-up which parts are actually done in the quicktest function (and how) and which are not. I guess a crucial step would be to carefully document this in the 'Quicktest API' functions. In any case, we should increase the visibility of the Quicktest if we want people to use it.

fsimonis:
Closing this PR as v3 and other recent testing enhancements reduce the overhead of writing tests a lot.

PR #1226, opened 2022-03-28
Improved edge barycentric coordinates computation

boris-martin:
## Main changes of this PR

To compute the projection P of a point U on the edge AB, we move around the AB direction (i.e. normalized AB vector) by a factor `dot(AB, AU) / |AB|`. Then the ratio between AP and AB is used to compute barycentric coordinates. That's a lot of division, and it can be simplified as "barycentricCoordinate(1) = dot(AB, AU)/dot(AB, AB)`. Previously, there were two divisions by `sqrt(dot(AB, AB))` which was not very efficient.

New code is shorter, simpler, and almost certainly  faster (one less division and one less square root).

I didn't do any profiling, but removing square roots when possible seems like a no-brainer if it doesn't harm readability of maintainability.
I don't think the old way was better in terms of numerical stability either.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/math/barycenter.cpp
+++ b/src/math/barycenter.cpp
@@ -25,17 +25,15 @@ Eigen::Vector2d calcBarycentricCoordsForEdge(
 
   Vector2d barycentricCoords;
   VectorXd ab, au;
-  double   lenAb, lenProjected;
 
-  // constant per edge
-  ab    = b - a;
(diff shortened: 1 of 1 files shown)


fsimonis:
Nice readability boost

PR #1227, opened 2022-03-29
Change macos runner to the one provided by GitHub

uekerman:
I am changing the macOS build and test action from our own runner (which was offline for a long time, so tests did not run), to the ones provided by GitHub. 
`macos-latest` points to the latest available runner. We could also hard-code `macos-11` or similar.

I am currently checking out the master branch, since develop gives a build problem.

```
[ 59%] Building CXX object CMakeFiles/testprecice.dir/src/testing/ParallelCouplingSchemeFixture.cpp.o
/Users/runner/work/precice/precice/src/testing/GlobalFixtures.cpp:40:10: error: expected expression
  return = bu::log_successful_tests;
```

@fsimonis Any idea what could be the problem here?

We should switch back to develop. I only checkout master here to show that it used to build.

Joint work with @ajaust

### DIFF
--- a/.github/workflows/build-and-test-mac.yml
+++ b/.github/workflows/build-and-test-mac.yml
@@ -8,7 +8,7 @@ on:
 jobs:
   build:
     name: ${{ format('{0} {1} {2}', matrix.CXX, matrix.CONFIG, matrix.TYPE) }}
-    runs-on: [self-hosted, macOS]
+    runs-on: macos-11
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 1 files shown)


.github/workflows/build-and-test-mac.yml
```diff
@@ -20,6 +30,12 @@ jobs:
             CXX: 'clang++'
             TYPE: Release
     steps:
+    - name: Check out preCICE
+      uses: actions/checkout@v3
+      with:
+        ref: master
```
uekerman, line 36: Here I am checking out the master branch. 

```suggestion
```

gives the develop branch.

PR #1231, opened 2022-03-30
Fix typo in return of unittest for MacOS 

kursatyurt:
This PR fixes error mentioned in #1227.

There was an assignment in the return statement due to a typo. It is removed.


fsimonis:
:see_no_evil:

PR #1238, opened 2022-03-31
Add pkg-config file for binary directory

### LINKED ISSUE #674: Generate a libprecice.pc for the binary tree
# Problem Description

We currently only generate a `libprecice.pc` file usable after installing preCICE.
This requires users to install preCICE in order to use it in `pkg-config`-based adapters etc.
It would be nice to provide an additional pkg-config to use preCICE from the binary directory.

# Potential Solution

Generate a pkg-config usable from the binary dir.

fsimonis:
## Main changes of this PR

This PR adds a pkg-config file for the binary directory in addition to the installation candidate in `libs/pkgconfig/`.

This should simplify using the preCICE develop version with adapters using pkg-config.

Closes #674

## Author's checklist

## Reviewers' checklist

* [ ] Is the pkg-config working on your system?


DavidSCN:
Adding now something like `export PKG_CONFIG_PATH="$HOME/precice/build/:$PKG_CONFIG_PATH"` works in principle just as good as exporting the libprecice.so from an install target package config file. I still run into a linker error when using the openfoam-adapter `Allwmake` script, but this is probably an insufficiency of the adapter script rather than preCICE.

DavidSCN (approved):

PR #1242, opened 2022-04-08
Move static Index into Mesh

### LINKED ISSUE #1192: Remove static query::Indexer
**Describe your setup**

preCICE 2.3.1

**Describe the problem**

The indexer of the query package internally caches Index trees to prevent them from being regenerated.
This is yet another static state, which we need to remove in order to robustly allow multiple SolverInterfaces.

**Proposed solutions**

One solution would be to attach the index directly to the Mesh.
This would simplify the access to such indexes, as mappings etc already have access to the Mesh.

**Additional context**

#1049

fsimonis:
## Main changes of this PR

This PR removes the `static` caching of the RTrees in the query package and adds the index directly to the `mesh::Mesh`.
This allowed me to remove the use of signals to reset caches etc.
All mappings and watchpoints now use the `mesh::Mesh::index()` instead of using some `static`ally cached entry.

Less static state and simpler lifetime management.
Using the geometric index is more intuitive, as it can directly be accessed via the `Mesh`.

Closes #1192

## Author's checklist

## Reviewers' checklist


src/mapping/NearestNeighborBaseMapping.cpp
```diff
@@ -53,11 +51,6 @@ void NearestNeighborBaseMapping::computeMapping()
     searchSpace = input();
   }
 
-  // For log outputs
-  precice::utils::Event e2(baseEvent + ".getIndexOnVertices", precice::syncMode);
-  query::Index          indexTree(searchSpace);
-  e2.stop();
```
boris-martin, line 59: Why remove the event-based stuff?

src/mesh/Mesh.cpp
```diff
@@ -29,18 +29,11 @@ Mesh::Mesh(
...
       _id(id),
-      _boundingBox(dimensions)
+      _boundingBox(dimensions),
+      _index(*this)
 {
   PRECICE_ASSERT((_dimensions == 2) || (_dimensions == 3), _dimensions);
   PRECICE_ASSERT(_name != std::string(""));
-
-  meshChanged.connect([](Mesh &m) { query::clearCache(m); });
-  meshDestroyed.connect([](Mesh &m) { query::clearCache(m); });
```
boris-martin, line 38: When does the clearing happen then? when we call index.clear()?

src/precice/impl/WatchPoint.cpp
```diff
@@ -97,7 +96,7 @@ void WatchPoint::exportPointData(
     return;
   }
 
-  //PRECICE_ASSERT(_vertices.size() == _weights.size());
+  // PRECICE_ASSERT(_vertices.size() == _weights.size());
```
boris-martin, line 99: Do we want these comments to stay?

src/query/Index.hpp
```diff
@@ -74,22 +75,27 @@ class Index {
...
+  void clear();
+
 private:
-  struct IndexImpl;
+  class IndexImpl;
   std::unique_ptr<IndexImpl> _pimpl;
 
-  const mesh::PtrMesh             _mesh;
+  /// The indexed Mesh.
+  mesh::Mesh *_mesh;
```
boris-martin, line 97: Why use raw pointer instead of smart pointers?

boris-martin (commented):
Much easier to read overall. I didn't check the details of `Index.cpp` as it looks quite complex, but the impacts on the rest of the code are really nice.

src/mapping/NearestNeighborBaseMapping.cpp
fsimonis: This event is now part of the `Index` itself:
https://github.com/precice/precice/pull/1242/files#diff-4c0c54bc6aa3d8fe12a773602b1c27237290e890cd134b796ca933f85785b012R50

src/mesh/Mesh.cpp
fsimonis: Exactly, this is done in `Mesh::clear()` and when necessary in mappings.

fsimonis:
@boris-martin Thanks for the review!

src/query/Index.hpp
boris-martin: Isn't that the purpose of weak pointers?

fsimonis: Weak pointers model a non-owning handle to a shared pointer.
To access the object, you tell the weak pointer that you require ownership, then it returns an owning shared pointer. Think of deferred ownership.

PR #1249, opened 2022-04-19
Remove duplicate documentation.

### LINKED ISSUE #1244: Duplication of documentation in SolverIntefaceImpl unclear.
Currently, there is a lot of duplicate and outdated documentation in `SolverInterfaceImpl` and `SolverInterface`. I would suggest to delete most of the documentation of `SolverInterfaceImpl`, because it takes a lot of effort to keep both files consistent and I do not see the purpose of duplication here. Documentation in `SolverInterface` is obviously useful for our users. 

Am I missing something here?

To give an example for inconsistent, duplicate and outdated documentation:

https://github.com/precice/precice/blob/d770c53ff59a7023c9e4933949d9e8a0797d5739/src/precice/impl/SolverInterfaceImpl.hpp#L423-L439

https://github.com/precice/precice/blob/d770c53ff59a7023c9e4933949d9e8a0797d5739/src/precice/SolverInterface.hpp#L696-L723

BenjaminRodenberg:
## Main changes of this PR

Removes duplicate documentation. Closes #1244 

Makes our lives easier.

## Author's checklist

## Reviewers' checklist


BenjaminRodenberg:
I could not build the doxygen documentation, because installing doxygen caused problems. @fsimonis could you help out here? I also realized that some of the functions that already use `@copydoc` don't link as expected (see https://precice.org/doxygen/develop/classprecice_1_1impl_1_1SolverInterfaceImpl.html#aff3674e008f4d491b4a59a578455e604). I think you need to provide the arguments to make the linking work, but I could not test it unfortunately.

src/precice/impl/SolverInterfaceImpl.hpp
```diff
@@ -45,41 +45,29 @@ class Mesh;
...
    *
    * A solver that wants to use the SolverInterfaceImpl must instantiate an object
    * of this class. The object has to be configured by one of the configure
    * methods before it has a reasonable state and can be used.
    *
-   * @param[in] configurationFileName Name (with path) of the xml config file.
-   * @param[in] participantName Name of the participant using the interface. Has to
-   *                            match the name given for a participant in the
-   *                            xml configuration file.
+   * @copydetails SolverInterface::SolverInterface(const std::string&, const std::string&, int, int)
```
BenjaminRodenberg, line 61: This looks a bit messy, but because the middle paragraph is explicitly mentioning `SolverInterfaceImpl` I did not just want to move it inside `SolverInterface`. Is this comment needed? If we explicitly mention the class, it makes copying the documentation more complicated or impossible. Could this be moved to the brief description?

BenjaminRodenberg: I decided to remove most of this part, because it is very outdated. I just kept the part `Use the parameter communicator to specify a custom global MPI communicator. Pass a null pointer to signal preCICE to use MPI_COMM_WORLD.` However, this only shows up in `SolverInterfaceImpl`, because this is only interesting for developers. Users should not pass a `nullptr`, but use the constructor without the `communicator` argument. Is this correct? @uekerman @fsimonis 

A sidenote: I would actually not allow to call `SolverInterface::SolverInterface(const std::string &participantName, const std::string &configurationFileName, int solverProcessIndex, int solverProcessSize, void *communicator)` with `communicator=nullptr`, because this is not the way the public interface is supposed to be used as far as I understand it. Am I missing some context here? We could also open an issue on this and change it with the next breaking release.

src/precice/SolverInterface.hpp
```diff
@@ -56,6 +58,8 @@ class SolverInterface {
       int                solverProcessSize);
 
   /**
+   * @brief Constructor with support for custom MPI_COMM_WORLD.
```
fsimonis, line 61:
```suggestion
   * @brief Constructor with support for custom MPI communicator.
```

src/precice/SolverInterface.hpp
```diff
@@ -82,6 +86,13 @@ class SolverInterface {
   /**
    * @brief Fully initializes preCICE
    *
+   * - Initiates MPI communication, if not done yet.
```
fsimonis, line 89:
```suggestion
```

The caller provides the communicator and thus already initialized.

src/precice/SolverInterface.hpp
```diff
@@ -124,6 +135,13 @@ class SolverInterface {
   /**
    * @brief Advances preCICE after the solver has computed one timestep.
    *
+   * - Sends and resets coupling data written by solver to coupling partners.
+   * - Receives coupling data read by solver.
+   * - Computes and applied data mappings.
```
fsimonis, line 140:
```suggestion
   * - Computes and applies data mappings.
```

src/precice/SolverInterface.hpp
```diff
@@ -124,6 +135,13 @@ class SolverInterface {
...
+   * - Computes acceleration of coupling data.
+   * - Exchanges and computes information regarding the state of the coupled
+   *   simulation.
```
fsimonis, line 143: The order is a bit mixed.
It should be
* write mapping
* send
* coupling info and timestep
* accelerate (not sure about this one)
* receive 
* read mapping

src/precice/SolverInterface.hpp
```diff
@@ -147,10 +165,28 @@ class SolverInterface {
...
    * @pre finalize() has not been called.
    *
    * @post Communication channels are closed.
    * @post Meshes and data are deallocated
+   * @post Finalized managed PETSc
+   * @post Finalized managed MPI
+   *
+   * @warning
+   * Finalize is not the inverse of initialize().
+   * Finalize has to be called after construction.
```
fsimonis, line 189: It will be called in the destructor if it wasn't called.

src/precice/impl/SolverInterfaceImpl.hpp
fsimonis: Yes, this should only be used by user who actually needs to provide a MPI_Comm

fsimonis (commented):
---

src/precice/SolverInterface.hpp
BenjaminRodenberg: Then I will remove these lines. I don't think we really need the warnings here.

BenjaminRodenberg: 6a42ccbe756f69cfdda319a034bf02951da11a55

src/precice/SolverInterface.hpp
BenjaminRodenberg: I don't read this as an ordered list. The order is also different depending on the coupling scheme and whether the first or second participant calls advance. From the outside perspective (`SolverInterface` and `SolverInterfaceImpl`) the order of the actions is an implementation detail. Therefore, I would leave it as it is.

src/precice/impl/SolverInterfaceImpl.hpp
BenjaminRodenberg: I will open an issue on this. We should not resolve it here, because this is not a documentation issue anymore.

BenjaminRodenberg: I opened an issue here: https://github.com/precice/precice/issues/1260

src/precice/SolverInterface.hpp
```diff
@@ -41,6 +41,8 @@ class SolverInterface {
   ///@{
 
   /**
+   * @brief Constructor
```
fsimonis, line 44:
```suggestion
   * @brief Constructs a SolverInterface for the given participant
```

src/precice/SolverInterface.hpp
```diff
@@ -56,6 +58,8 @@ class SolverInterface {
       int                solverProcessSize);
 
   /**
+   * @brief Constructor with support for custom MPI communicator.
```
fsimonis, line 61:
```suggestion
   * @brief Constructs a SolverInterface for the given participant and a custom MPI communicator.
```

fsimonis (approved):
2 minor changes. Then please check the formatting and merge.

PR #1257, opened 2022-04-22
Removes RS-LS truncation of stored columns

KyleDavisSA:
## Main changes of this PR
This PR removes the arbitrary truncation of all but the first `5` columns per time window for the least squares restart for IQ-IMVJ method. This negatively impacts the performance of the RS-LS method. The accumulation of columns is handled by the filtering the of the RS-LS columns during the restart step.

Tests have shown that this truncation ignores information helpful for convergence. The problem of too many columns used is covered by the filtering step in https://github.com/precice/precice/blob/cf1c06e2298baf0f1efe0f078aef9b298d286ef4/src/acceleration/MVQNAcceleration.cpp#L584-L603

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/1257.md
+++ b/docs/changelog/1257.md
@@ -1,1 +1,1 @@
-- Removed least squares truncation of stored columns per time window for IQN-IMVJ acceleration restart.
+- Removed restriction to only use 5 iterations in the `RS_LS` restart mode of the IQN-IMVJ acceleration method.
--- a/docs/changelog/1257.md
+++ b/docs/changelog/1257.md
@@ -1,0 +1,1 @@
+- Removed least squares truncation of stored columns per time window for IQN-IMVJ acceleration restart.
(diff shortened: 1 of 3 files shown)


uekerman:
Do we still need the member variable `_usedColumnsPerTimeWindow` then?
It's pretty suspicious that we can change sth like without any test failing. IMVJ is really not well tested.

KyleDavisSA:
No, we would not need the variable `_usedColumnsPerTimeWindow`. It is weird that it is there at all, and I found the results between the different restart were better by removing this.

uekerman:
We also need a changelog entry here.

docs/changelog/1257.md
```diff
@@ -0,0 +1 @@
+- Removed least squares truncation of stored columns per time window for IQN-IMVJ acceleration restart.
```
uekerman, line 1:
```suggestion
- Removed restriction to only use 5 iterations in the `RS_LS` restart mode of the IQN-IMVJ acceleration method.
```

uekerman (approved):
I assume that the original motivation to have this limit was efficiency of the restart, meaning that the restart itself does not become to expensive.

PR #1258, opened 2022-04-22
Added "Volume Cell Interpolation" mapping support in 2D

boris-martin:
## Main changes of this PR

- Added a new data mapping, "Volume Cell Interpolation", similar to NP but designed for volumetric coupling. Only supported in 2D for now. Usage: `volume-cell-interpolation` in the XML config and set up connectivitiy.
- Made necessary adjustments to make it work: allow the user to use `setMeshTriangle` in 2D without discarding triangles, modified communications to send triangles when communicating a mesh, etc.
- Added unit tests to check the correct interpolation, and an integration test to check that the API works. (Contains a consistent mapping with a triangle as input and a single point as output, and test both read and write mapping). **Should I add some parallel tests?**
- Some refactoring: common parts between NP and VCI adre shared in a new base class

Milestone of https://github.com/precice/precice/issues/468.

## Author's checklist

## Reviewers' checklist

@fsimonis : For now I added a fall-back on nearest neighbor when no matching triangle is found.
For some reason I can't seem to run make changelog. And I don't know exactly how to format it.


boris-martin:
As of now my computer doesn't support clang-format-8 so I'll wait for the merge of https://github.com/precice/precice/pull/1276  to format again

src/query/Index.hpp
```diff
@@ -86,6 +86,7 @@ class Index {
    */
   ProjectionMatch findNearestProjection(const Eigen::VectorXd &location, int n);
 
+  ProjectionMatch findNearestVolume(const Eigen::VectorXd &location, int n);
```
fsimonis, line 89: Maybe we can come up with a better name for this.
This will look for the cell including the point and return the nearest vertex as a fallback.

boris-martin:
Added a lot of tests (Conservative mapping & parallel integration tests), now it should be ready.
Designing these tests highlighted the need for https://github.com/precice/precice/pull/986, as I had problems of filtering in some cases. For instance, in the conservative test, if we apply a force on the edge of a triangle, the third vertex of the triangle is unused, which leads to the whole triangle being discarded and the mapping falling back to NN.

@fsimonis a quickfix to this would be to restore the "fall-back on edge if no triangle is found", what do you think ?

fsimonis:
@uekerman Wanted to have a look at this once it is ready to merge. I added him as a reviewer.

src/mapping/BarycentricBaseMapping.hpp
```diff
@@ -0,0 +1,57 @@
...
+   */
+  virtual void map(
+      int inputDataID,
+      int outputDataID) override;
+
+  virtual void tagMeshFirstRound()  = 0;
+  virtual void tagMeshSecondRound() = 0;
+
+protected:
+  logging::Logger _log{"mapping::BarycentricBaseMapping"};
```
fsimonis, line 49: Please make this logger `private` and define separate loggers in `NearestProjection` and `CellInterpolation`. Otherwise it will be impossible to differentiate these two in the logs.

src/mapping/VolumeCellInterpolation.cpp
```diff
@@ -0,0 +1,164 @@
...
+    : BarycentricBaseMapping(constraint, dimensions)
+{
+  if (constraint == CONSISTENT) {
+    setInputRequirement(Mapping::MeshRequirement::FULL);
+    setOutputRequirement(Mapping::MeshRequirement::VERTEX);
+  } else if (constraint == CONSERVATIVE) {
+    setInputRequirement(Mapping::MeshRequirement::VERTEX);
+    setOutputRequirement(Mapping::MeshRequirement::FULL);
+  } else {
+    PRECICE_ASSERT(constraint == SCALEDCONSISTENT, constraint);
```
fsimonis, line 41: We should also cover scaled-consistent mappings. Have a look at NP scaled-consistent tests.

src/mapping/VolumeCellInterpolation.cpp
```diff
@@ -0,0 +1,164 @@
...
+
+  if (getDimensions() == 2) {
+    if (!fVertices.empty() && searchSpace->triangles().empty()) {
+      PRECICE_WARN("2D Mesh \"{}\" does not contain triangles. "
+                   "Volume cell interpolation falls back to nearest neighbor mapping.",
+                   searchSpace->getName());
+    }
+  } else {
+    if (!fVertices.empty() && searchSpace->triangles().empty()) {
+      PRECICE_WARN("3D Mesh \"{}\" does not contain tetrahedrons. "
```
fsimonis, line 78:
```suggestion
      PRECICE_WARN("3D Mesh \"{}\" does not contain tetrahedra. "
```

src/mapping/NearestProjectionMapping.hpp
```diff
@@ -25,33 +25,8 @@ class NearestProjectionMapping : public Mapping {
...
-   */
-  virtual void map(
-      int inputDataID,
-      int outputDataID) override;
-
   virtual void tagMeshFirstRound() override;
   virtual void tagMeshSecondRound() override;
-
-private:
-  logging::Logger _log{"mapping::NearestProjectionMapping"};
```
fsimonis, line 50: This should remain

src/mapping/VolumeCellInterpolation.hpp
```diff
@@ -0,0 +1,33 @@
...
+  /// Destructor, empty.
+  virtual ~VolumeCellInterpolation() {}
+
+  /// Computes the projections and interpolation relations.
+  virtual void computeMapping() override;
+
+  virtual void tagMeshFirstRound() override;
+  virtual void tagMeshSecondRound() override;
+};
+
```
fsimonis, line 31: This should have a private logger.

fsimonis (commented):
While easy to read, I think it would be a good idea to extract some code into separate PRs:
* the dimensionality changes in the `Filter` and `CommunicateMesh`
* the Refactoring of the NearestProjection mapping into the `BarycentricBaseMapping`

The rest then should stay in this PR.

fsimonis:
@boris-martin Both split PRs are merged. You can tidy this PR up now.

boris-martin:
Seems I messed up the history where refactoring. New PR here: https://github.com/precice/precice/pull/1297

Issue #1201, opened 2022-03-10
Investigate upgrade to C++17


fsimonis:
**Please describe the problem you are trying to solve.**
Once we move our baseline versions from Ubuntu 18.04 LTS to 20.04 LTS, we should investigate moving from C++14 to C++17.

This would currently only impact cray.

https://en.cppreference.com/w/cpp/compiler_support

MakisH:
In that case, we can use/improve:
- `[[maybe_uninitialized]]`: https://github.com/precice/precice/pull/1285
- https://github.com/precice/precice/blob/6d27ad5db42997bbc06ff97b39eb4999196295b7/src/utils/traits.hpp#L8
- https://github.com/precice/precice/blob/6d91dc8a221d7e9faf713d8fc8cc494f74d9d263/src/utils/algorithm.hpp#L69

fsimonis:
Available on Ubuntu 20.04 LTS:

* g++ / libstdc++ version 10
* clang 12

So, C++17 isn't an issue from the compiler/STL side of things.

The Intel compiler version shipped with the oneAPI toolkits is 2022.1, so the newest.

fsimonis:
Also, we will be able to make good use of `std::variant` to clean-up #982

MakisH:
> Available on Ubuntu 20.04 LTS:
> 
>     * g++ / libstdc++ version 10

Note that the default is 9.3, though. Still with full C++17 support, but to look at the right lists.

PR #1261, opened 2022-04-24
Introduce private SolverInterfaceImpl constructor

### LINKED ISSUE #1260: SolverInterface constructor accepts communicator = nullptr
**Describe your setup**

preCICE Version: 20c4f71

**Describe the problem**

There are two constructors for `SolverInterface`. One constructor accepts an additional `communicator` argument to allow for custom MPI communicators. Currently it is possible to pass `nullptr` here, which basically maps to the behavior of the `SolverInterface` constructor without the `communicator` argument. This should not be possible from my perspective. If I don't want to use a custom MPI communicator, I should use the right constructor.

**Step To Reproduce**

1. Modify `solverdummy.cpp`:
```
-  SolverInterface interface(solverName, configFileName, commRank, commSize);
+  SolverInterface interface(solverName, configFileName, commRank, commSize, nullptr);
```
2. Running solver dummy is possible without any error.

**Expected behaviour**

If the `SolverInterface` constructor is called with `communicator=nullptr` and error should be given to the user.

**Additional context**

Stumbled across this issue while working on #1249. I expect that working on this issue will also make the documentation more clear.

BenjaminRodenberg:
## Main changes of this PR

Introduces a private constructor. Closes #1260.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -85,11 +85,16 @@ SolverInterfaceImpl::SolverInterfaceImpl(
     const std::string &configurationFileName,
     int                accessorProcessRank,
     int                accessorCommunicatorSize,
-    void *             communicator)
+    void *             communicator,
+    bool               allowNullptr)
     : _accessorName(std::move(participantName)),
       _accessorProcessRank(accessorProcessRank),
       _accessorCommunicatorSize(accessorCommunicatorSize)
 {
+  if (!allowNullptr) {
+    PRECICE_CHECK(communicator != nullptr,
+                  "Passing \"nullptr\" as \"communicator\" to SolverInterface constructor is not allowed. Please use the SolverInterface constructor without the \"communicator\" argument, if you don't want to pass an MPI communicator.");
+  }
   PRECICE_CHECK(!_accessorName.empty(),
                 "This participant's name is an empty string. "
                 "When constructing a preCICE interface you need to pass the name of the "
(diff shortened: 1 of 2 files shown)


BenjaminRodenberg:
I'm not sure whether this is actually breaking and must go into `3.0.0` or we could also declare it as a simple bugfix and put it into `2.x.x`.

fsimonis:
This is an ideal fit for an `std/boost::optional` as it allows differentiating between a user-provided nullptr and no user-provided ptr. This requires an upgrade to C++17 for the `std::` version #1201. 

There are 2 ways to implement this cleanly:

### 2. keep the pimple clean
 Have 2 constructors with the same interface in both SolverInterface and SolverInterfaceImpl, then forward the latter two to a single private constructor that takes a `::optional<void *>`.

```cpp
class SolverInterfaceImpl{
public:
  SolverInterfaceImpl(
      std::string        participantName,
      const std::string &configurationFileName,
      int                solverProcessIndex,
      int                solverProcessSize)
: SolverInterfaceImpl(participantName, configurationFileName, solverProcessIndex, solverProcessSize, nullopt) {}

  SolverInterfaceImpl(
      std::string        participantName,
      const std::string &configurationFileName,
      int                solverProcessIndex,
      int                solverProcessSize,
      void *             communicator)
: SolverInterfaceImpl(participantName, configurationFileName, solverProcessIndex, solverProcessSize, make_optional(communicator)) {}
private:
  SolverInterfaceImpl(
      std::string        participantName,
      const std::string &configurationFileName,
      int                solverProcessIndex,
      int                solverProcessSize,
      std::optional<void *> communicator)
}
```

### 2. keep the interface clean

Have 1 constructor in SolverInterfaceImpl that takes a `::optional<void *>`, then call it from the 2 constructors in SolverInterface.

```cpp
class SolverInterfaceImpl{
public:
  SolverInterfaceImpl(
      std::string        participantName,
      const std::string &configurationFileName,
      int                solverProcessIndex,
      int                solverProcessSize,
      std::optional<void *> communicator)
}
```

BenjaminRodenberg:
> This is an ideal fit for an `std/boost::optional` as it allows differentiating between a user-provided nullptr and no user-provided ptr. This requires an upgrade to C++17 for the `std::` version #1201.
> 

As long as we did not conclude on #1201 I'm a bit hesitant implementing it here, because this would make this PR unmergable until we upgrade to C++ 17. The solution sounds reasonable, but does (at the current state) not fit into preCICE (we can still keep it as an issue and improve it in the future).

fsimonis:
An optional is really only a slight refactoring of this solution. So, let's merge this once CI passes.

fsimonis (approved):

BenjaminRodenberg:
I just realized that this change is considered breaking. Therefore, we should not merge this PR before we are sure that our next release will be `v3.0.0`. We can still keep this PR open and consider it ready to merge.

fsimonis:
The case, which could be regarded as a breaking change, would already break MPI.
So, I don't see this as breaking.

BenjaminRodenberg:
This PR is ready to merge as soon as the tests are complete.

Issue #1107, opened 2021-10-05
Wrong error is raised, if relative-convergence-measure is set to a very small tolerance


BenjaminRodenberg:
**Describe your setup**

Operating system (e.g. Linux distribution and version): _Ubuntu 20.04_
preCICE Version: 30d2b21
tutorials Version: https://github.com/precice/tutorials/commit/90804544a51bdef40c013ca56b4dbeff8bb1d10e

**Describe the problem**
When running the partitioned-heat-conduction tutorial with a very low tolerance a misleading error is raised.

I'm using the case from the tutorials with a small modification in `precice-config.xml`:
```
      <relative-convergence-measure data="Flux" mesh="Dirichlet-Mesh" limit="2e-14"/>
      <relative-convergence-measure data="Temperature" mesh="Neumann-Mesh" limit="2e-14"/>
```
Then I run the FEniCS version of the tutorial as described in the tutorials documentation. After a few iterations, I get the following error:
```
---[precice] ERROR:  Attempting to add a zero vector to the quasi-Newton V matrix. This means that the residual in two consecutive iterations is identical. There is probably something wrong in your adapter. Maybe you always write the same (or only incremented) data or you call advance without providing  new data first.
```

**Expected behaviour**
The error message is misleading, because the adapter works correctly. I guess the reason for the problem is that the convergence tolerance is very low and therefore the `residualMagnitude` seems to be zero.

Some possible solutions:

* Only raise a warning and add some more context, if the tolerance is low.
* Raise an error, if the `residualMagnitude` is really zero and try to catch the case I'm describing above elsewhere.

**Additional context**

See https://github.com/precice/precice/pull/1095#issuecomment-934500260, https://github.com/precice/precice/pull/1092#issuecomment-934472790 and https://github.com/precice/precice/issues/976.

uekerman:
Even better would be a "warning" and then simply not add the column to the quasi Newton matrices.

KyleDavisSA:
This can be done, but reaching convergence might be a problem in this case. If the QN method cannot add a new column as it is so close to zero, perhaps the coupling might not be able to reach the convergence threshold? The solvers will continue iterating until a max number of iterations per time window is reached.

KyleDavisSA:
I think an issue might be here:
https://github.com/precice/precice/blob/e3a6f88dbebb29a0b92a98f30c6865406cb5dbd1/src/acceleration/BaseQNAcceleration.cpp#L210-L212

[...]

PR #1266, opened 2022-04-25
Adds a warning for small relative convergence thresholds

KyleDavisSA:
## Main changes of this PR
Adds a warning if the relative convergence threshold is close to the machine tolerance.

closes https://github.com/precice/precice/issues/1107

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/acceleration/BaseQNAcceleration.cpp
+++ b/src/acceleration/BaseQNAcceleration.cpp
@@ -212,10 +212,9 @@ void BaseQNAcceleration::updateDifferenceMatrices(
       }
 
       PRECICE_CHECK(not math::equals(residualMagnitude, 0.0),
-                    "Attempting to add a zero vector to the quasi-Newton V matrix. This means that the residual "
-                    "in two consecutive iterations is identical. There is probably something wrong in your adapter. "
-                    "Maybe you always write the same (or only incremented) data or you call advance without "
-                    "providing  new data first.");
+                    "Attempting to add a zero vector to the quasi-Newton V matrix. This means that the residuals "
+                    "in two consecutive iterations are identical. If a relative convergence limit was selected, "
+                    "consider increasing the convergence threshold.");
 
       bool columnLimitReached = getLSSystemCols() == _maxIterationsUsed;
       bool overdetermined     = getLSSystemCols() <= getLSSystemRows();
--- a/src/cplscheme/config/CouplingSchemeConfiguration.cpp
+++ b/src/cplscheme/config/CouplingSchemeConfiguration.cpp
@@ -697,6 +697,10 @@ void CouplingSchemeConfiguration::addRelativeConvergenceMeasure(
                 "Please check the <relative-convergence-measure limit=\"{}\" data=\"{}\" mesh=\"{}\" /> subtag "
                 "in your <coupling-scheme ... /> in the preCICE configuration file.",
                 limit, dataName, meshName);
+  if (math::smallerEquals(limit/10, 0.0)){
+    PRECICE_WARN("Relative convergence limit is too small. This may cause a conflict with "
+                 "the quasi-Newton acceleration scheme if selected. ");
+  }
 
   impl::PtrConvergenceMeasure measure(new impl::RelativeConvergenceMeasure(limit));
   ConvergenceMeasureDefintion convMeasureDef;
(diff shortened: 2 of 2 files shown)


IshaanDesai:
I tried the `partitioned-heat-conduction` case with a relative convergence measure limit of `2e-14` and I see the Warning which is added in this PR :+1:

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -212,10 +212,9 @@ void BaseQNAcceleration::updateDifferenceMatrices(
       }
 
       PRECICE_CHECK(not math::equals(residualMagnitude, 0.0),
-                    "Attempting to add a zero vector to the quasi-Newton V matrix. This means that the residual "
-                    "in two consecutive iterations is identical. There is probably something wrong in your adapter. "
-                    "Maybe you always write the same (or only incremented) data or you call advance without "
-                    "providing  new data first.");
+                    "Attempting to add a zero vector to the quasi-Newton V matrix. This means that the residuals "
+                    "in two consecutive iterations are identical. If a relative convergence limit was selected, "
+                    "consider increasing the convergence threshold.");
```
uekerman, line 217: Why removing the previous hints? They are still helpful, no?
On the other hand, I would not mention the relative convergence limit here. We already have the warning below. Could only confuse here more than it helps?
If a user runs into this error message, I guess it is still most of the times due to

> There is probably something wrong in your adapter. Maybe you always write the same (or only incremented) data or you call advance without providing  new data first.

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -697,6 +697,10 @@ void CouplingSchemeConfiguration::addRelativeConvergenceMeasure(
                 "Please check the <relative-convergence-measure limit=\"{}\" data=\"{}\" mesh=\"{}\" /> subtag "
                 "in your <coupling-scheme ... /> in the preCICE configuration file.",
                 limit, dataName, meshName);
+  if (math::smallerEquals(limit / 10, 0.0)) {
+    PRECICE_WARN("Relative convergence limit is too small. This may cause a conflict with "
+                 "the quasi-Newton acceleration scheme if selected. ");
```
uekerman, line 702: "You chose a relative convergence limit of X. This is close to the hard-coded numerical resolution of preCICE, Y, which might lead to numerical stability issues."
... or similar.
It's not only the quasi-Newton acc, we could also get problems in the mapping etc. And we need the numbers here to prevent the user from starting a trial-and-error testing.

Using `smallerEquals` is problematic here as soon as the check above is removed. Let's better check the absolute value of the limit.

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -722,6 +726,11 @@ void CouplingSchemeConfiguration::addResidualRelativeConvergenceMeasure(
                 "Please check the <residul-relative-convergence-measure limit=\"{}\" data=\"{}\" mesh=\"{}\" /> subtag "
                 "in your <coupling-scheme ... /> in the preCICE configuration file.",
                 limit, dataName, meshName);
+  if (math::smallerEquals(limit / 10, 0.0)) {
+    PRECICE_WARN("Residual Relative convergence limit is too small. This may cause a conflict with "
+                 "the quasi-Newton acceleration scheme if selected. ");
```
uekerman, line 731: See above.

uekerman (changes requested):
Location of the warning is good, but we should still tweak the text a bit.

src/acceleration/BaseQNAcceleration.cpp
KyleDavisSA: If you write the same data, the first error should be seen:
https://github.com/precice/precice/blob/d1800d5801256861c7672907020a11d9bf4a2cbd/src/acceleration/BaseQNAcceleration.cpp#L178-L183

But if the change in residual is becoming very low, then you are nearing convergence.

uekerman: > If you write the same data, the first error should be seen

True.

KyleDavisSA: Should we leave this part as is now?

src/cplscheme/config/CouplingSchemeConfiguration.cpp
KyleDavisSA: Are we planning on removing the check here?

KyleDavisSA: Something like 
```
 if (limit > 10*math::NUMERICAL_ZERO_DIFFERENCE){
    PRECICE_WARN("The relative convergence limit=\"{}\" is close to the hard-coded numerical resolution of preCICE. "
                 "This may lead to instabilities. Increase the relative convergence limit. ", limit);
  }
  ```

fsimonis: Why not simply check `math::equals(limit, 0)`?
If this is true, then the limit is too small.

src/acceleration/BaseQNAcceleration.cpp
uekerman: > On the other hand, I would not mention the relative convergence limit here. We already have the warning below. Could only confuse here more than it helps?

This concern still holds. Do you agree?

src/cplscheme/config/CouplingSchemeConfiguration.cpp
uekerman: We should also give the user the value of `math::NUMERICAL_ZERO_DIFFERENCE` to avoid trial-and-error testing.

src/acceleration/BaseQNAcceleration.cpp
uekerman: Well above, you only get a warning

> The coupling residual equals almost zero ...

One would still run into this error.
But in the end, not soo important.

fsimonis:
We are still missing a changelog entry

PR #1267, opened 2022-04-25
Fix wrong `DataContext` for multiple mappings

### LINKED ISSUE #1204: Multiple read mappings to same mesh assign wrong mappingContext to dataContext
**Describe your setup**

Current preCICE develop version

**Describe the problem**

When we allowed multiple read mappings to the same mesh in #849, we forgot to update the following assignment (which still assumes that there is only one read mapping to a certain mesh). 

https://github.com/precice/precice/blob/ebce55ab5c563295909b0179f6610b6e4d4fa529/src/precice/config/ParticipantConfiguration.cpp#L517-L534

A user ran into this problem: 

https://precice.discourse.group/t/participant-reads-zeros-for-mapped-data/994/9

(simplified) preCICE config:

```xml
    <mesh name="Fluid1-Centroids">
      <use-data name="Density-f2" />
      <use-data name="Velocity-f2" />
      <use-data name="Pressure-f2" />
    </mesh>

    <mesh name="Fluid2-Centroids">
      <use-data name="Density-f2" />
      <use-data name="Velocity-f2" />
    </mesh>

    <mesh name="Fluid2-Nodes">
      <use-data name="Pressure-f2" />
    </mesh>

    <participant name="Fluid1">
      <use-mesh name="Fluid1-Centroids" provide="yes" />
      <use-mesh name="Fluid2-Centroids" from="Fluid2" />
      <use-mesh name="Fluid2-Nodes" from="Fluid2" />

      <read-data name="Density-f2" mesh="Fluid1-Centroids" />
      <read-data name="Velocity-f2" mesh="Fluid1-Centroids" />
      <read-data name="Pressure-f2" mesh="Fluid1-Centroids" /> 

      <mapping:nearest-neighbor
        direction="read"
        from="Fluid2-Centroids"
        to="Fluid1-Centroids"
        constraint="consistent" />
      <mapping:nearest-projection
        direction="read"
        from="Fluid2-Nodes"
        to="Fluid1-Centroids"
        constraint="consistent" />
    </participant>
[...]
```

Error message

```
(0) 10:09:48 [config::ParticipantConfiguration]:529 in finishParticipantConfiguration: ERROR: Mesh “Fluid2-Nodes” needs to use data “Density-f2” to allow a read mapping to it. Please add a use-data node with name=“Density-f2” to this mesh.
```

Obviously, this error message is wrong, but more concerning is that also the assignment

https://github.com/precice/precice/blob/ebce55ab5c563295909b0179f6610b6e4d4fa529/src/precice/config/ParticipantConfiguration.cpp#L531

is wrong.

**Step To Reproduce**

Let's try to reproduce the problem with an integration test. A copy of this one should be a good starting point:

https://github.com/precice/precice/blob/ebce55ab5c563295909b0179f6610b6e4d4fa529/src/precice/tests/SerialTests.cpp#L2926

**Additional context**

Of course, we have the same problem for multiple **write** mappings **from** the same mesh.

DavidSCN:
## Main changes of this PR
Resolves #1204 and related to #849.

In case multiple read mappings target the same mesh 'multiple read to mappings' or multiple write mappings originate from the same mesh 'multiple write from mappings', we raised an erroneous error message and assigned the wrong data context. The logic was now reworked. In the configuration, we iterate now first over all mappings and look for corresponding read data and write data tags. The error is now transformed into an if condition and an error is raised in case no proper tags can be found.

## Author's checklist

## Reviewers' checklist


tests/serial/multiple-mappings/MultipleWriteFromMappings.xml
```diff
@@ -0,0 +1,52 @@
...
+        direction="write"
+        from="MeshB"
+        to="MeshATop"
+        constraint="consistent" />
+      <mapping:nearest-neighbor
+        direction="write"
+        from="MeshB"
+        to="MeshABottom"
+        constraint="consistent" />
+    </participant>
```
DavidSCN, line 40: @uekerman I'm not sure if this test is supposed to work as well. Currently (as before) it fails. I guess the main reason is that we call 
https://github.com/precice/precice/blob/7618982a33a340a151c6964430799048314d3bdd/src/precice/config/ParticipantConfiguration.cpp#L535
twice using the same data context (although we do the same for the read case, not completely sure about this).

DavidSCN: > I guess the main reason is that we call

Yes, that's definitely the case.

> although we do the same for the read case where is works, not completely sure about this

No, in this case we configure another data context (namely the read data context, which is different for each mesh and data). The config here simply doesn't work with the `dataContext`

uekerman: Such a case should work, yes. Good example. We need to extend the implementation here.

uekerman (changes requested):
Good extension of test cases and I like the comments in the configuration part :+1:
The tricky part, however, still remains :grin:

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -517,43 +517,62 @@ void ParticipantConfiguration::finishParticipantConfiguration(
...
+                        participant->getName(), dataContext.getMeshName(), dataContext.getMeshName());
+          dataContext.configureMapping(mappingContext, meshContext);
+          dataFound = true;
+        }
       }
     }
+    PRECICE_CHECK(dataFound,
+                  "Participant \"{}\" defines a write mapping from mesh \"{}\" to mesh \"{}\", "
+                  "but there are no corresponding read data and write data tags.",
+                  participant->getName(), mappingContext.mapping->getInputMesh()->getName(), mappingContext.mapping->getOutputMesh()->getName());
```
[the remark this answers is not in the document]
DavidSCN: I don't think this is the case. We first iterate over the mappings and look for corresponding data contexts. The data contexts are created according to the read data and write data tag, and not according to the use data tag

src/precice/tests/DataContextTest.cpp
```diff
@@ -55,23 +55,76 @@ BOOST_AUTO_TEST_CASE(testDataContextWriteMapping)
...
+  // Test dedicated content of the first mapping configuration
+  BOOST_TEST(fixture.getFromDataID(dataContext, 0) == ptrFromData->getID());
+  BOOST_TEST(fixture.getToDataID(dataContext, 0) == ptrToData->getID());
+  BOOST_TEST(fixture.mappingContext(dataContext)[0].fromMeshID == mappingContext.fromMeshID);
+  BOOST_TEST(fixture.mappingContext(dataContext)[0].toMeshID == mappingContext.toMeshID);
+  BOOST_TEST(fixture.mappingContext(dataContext)[0].hasMappedData == mappingContext.hasMappedData);
+  BOOST_TEST(fixture.mappingContext(dataContext)[0].mapping == mappingContext.mapping);
+  BOOST_TEST(fixture.mappingContext(dataContext)[0].timing == mappingContext.timing);
+
+  // Add a second mapping targeting a different to mesh
```
BenjaminRodenberg, line 78: Can you put this into an individual test? This keeps the existing test simple. Each unit test should also have a single purpose.

src/precice/tests/DataContextTest.cpp
```diff
@@ -55,23 +55,76 @@ BOOST_AUTO_TEST_CASE(testDataContextWriteMapping)
...
+  BOOST_TEST(fixture.getFromDataID(dataContext, 1) == ptrFromData->getID());
+  BOOST_TEST(fixture.getToDataID(dataContext, 1) == ptrToData2->getID());
+  BOOST_TEST(fixture.mappingContext(dataContext)[1].fromMeshID == mappingContext2.fromMeshID);
+  BOOST_TEST(fixture.mappingContext(dataContext)[1].toMeshID == mappingContext2.toMeshID);
+  BOOST_TEST(fixture.mappingContext(dataContext)[1].hasMappedData == mappingContext2.hasMappedData);
+  BOOST_TEST(fixture.mappingContext(dataContext)[1].mapping == mappingContext2.mapping);
+  BOOST_TEST(fixture.mappingContext(dataContext)[1].timing == mappingContext2.timing);
 }
 
 BOOST_AUTO_TEST_CASE(testDataContextReadMapping)
```
BenjaminRodenberg, line 130: Do we also need a test for multiple read mappings?

src/precice/impl/DataContext.hpp
```diff
@@ -79,7 +79,7 @@ class DataContext {
    * @param[in] mappingContext Context of the mapping
    * @param[in] meshContext Context of mesh this mapping is mapping from or to
    */
-  virtual void configureMapping(const MappingContext &mappingContext, const MeshContext &meshContext) = 0;
+  virtual void addMappingConfiguration(const MappingContext &mappingContext, const MeshContext &meshContext) = 0;
```
BenjaminRodenberg, line 82: Please also update the class documentation above that a data context can have no, one or multiple mappings.

src/precice/impl/DataContext.cpp
```diff
@@ -84,40 +85,39 @@ bool DataContext::hasMapping() const
...
 void DataContext::mapData()
 {
   PRECICE_ASSERT(hasMapping());
-  int fromDataID = getFromDataID();
-  int toDataID   = getToDataID();
-  _toData->toZero();
-  _mappingContext.mapping->map(fromDataID, toDataID);
-  PRECICE_DEBUG("Mapped values = {}", utils::previewRange(3, _toData->values()));
+  // Reset the toData
+  std::for_each(_toData.begin(), _toData.end(), [](auto &data) { data->toZero(); });
```
BenjaminRodenberg, line 102: `std::for_each` is nice, but since you already have a loop below, I would directly do the call there.

BenjaminRodenberg: Alternatively, you could also remove the loop and put everything into your lambda function like above with the `std::any_of` in `isMappingRequired`.

src/precice/impl/DataContext.cpp
```diff
@@ -57,23 +59,22 @@ std::string DataContext::getMeshName() const
...
+MeshID DataContext::getMeshID() const
 {
   PRECICE_ASSERT(_mesh);
   return _mesh->getID();
 }
 
-void DataContext::setMapping(MappingContext mappingContext, mesh::PtrData fromData, mesh::PtrData toData)
+void DataContext::addMapping(MappingContext mappingContext, mesh::PtrData fromData, mesh::PtrData toData)
 {
-  PRECICE_ASSERT(!hasMapping());
```
BenjaminRodenberg, line 68: This avoids by construction that a mapping in the `DataContext` is just overwritten with another mapping accidentally. I'm missing this security mechanism a bit in the new implementation. Is it, for example, allowed that the same mapping is added twice? What happens, if you do this? A corresponding unit test might also be a good idea here.

src/precice/impl/DataContext.hpp
```diff
@@ -91,25 +91,25 @@ class DataContext {
   DataContext(mesh::PtrData data, mesh::PtrMesh mesh);
 
   /// Defines the mapping associated to this DataContext. A DataContext may also exist without a mapping.
-  MappingContext _mappingContext;
+  std::vector<MappingContext> _mappingContext;
```
BenjaminRodenberg, line 94: Please update the documentation and name. It should read `_mappingContexts`.

src/precice/impl/DataContext.hpp
```diff
@@ -91,25 +91,25 @@ class DataContext {
...
-  mesh::PtrData _fromData;
+  std::vector<mesh::PtrData> _fromData;
 
   /// If a mapping exists, mesh::PtrData the mapping maps from.
-  mesh::PtrData _toData;
+  std::vector<mesh::PtrData> _toData;
 
   /**
-   * @brief Helper to set _mappingContext, _fromData and _toData.
+   * @brief Helper to add _mappingContext, _fromData and _toData.
```
BenjaminRodenberg, line 106: This comment is misleading. You don't add a `_mappingContext`, but you add a `mappingContext` to the `_mappingContexts` container, if I understand this correctly.

src/precice/impl/DataContext.hpp
```diff
@@ -91,25 +91,25 @@ class DataContext {
...
    *
    * @param mappingContext MappingContext this DataContext will be associated to.
```
BenjaminRodenberg, line 108: This also needs an update.

src/precice/impl/DataContext.hpp
```diff
@@ -91,25 +91,25 @@ class DataContext {
...
    * @param fromData Data the mapping maps from.
    * @param toData Data the mapping maps to.
    */
-  void setMapping(MappingContext mappingContext, mesh::PtrData fromData, mesh::PtrData toData);
+  void addMapping(MappingContext mappingContext, mesh::PtrData fromData, mesh::PtrData toData);
```
BenjaminRodenberg, line 112: I slightly prefer `appendMapping`, because this tells me that there is something listlike which contains mappings and I'm appending my mapping to this thing.

src/precice/impl/DataContext.hpp
```diff
@@ -126,24 +126,28 @@ class DataContext {
...
 
 private:
-  /// Mesh associated with _providedData.
+  /// Unique mesh associated with _providedData.
   mesh::PtrMesh _mesh;
 
   static logging::Logger _log;
 
   /**
    * @brief Get the ID of _fromData. Used for performing the mapping outside of this class.
```
BenjaminRodenberg, line 135: Please update this also to reflect that we get the ID from a vector.

src/precice/impl/DataContext.hpp
```diff
@@ -126,24 +126,28 @@ class DataContext {
...
-   * @return int ID of _fromData.
+   * @param[in] dataVectorIndex Index of the 'fromData' vector this data context holds
+   *
+   * @return DataID ID of _fromData.
    */
-  int getFromDataID() const;
+  DataID getFromDataID(int dataVectorIndex) const;
 
   /**
    * @brief Get the ID of _toData. Used for performing the mapping outside of this class.
```
BenjaminRodenberg, line 144: See above.

src/precice/impl/DataContext.cpp
DavidSCN: You cannot override things in this function as `emplace_back` never overrides any content. If you add the same mapping twice, you would also execute it twice. We could safeguard against this if that's what you have in mind.

src/precice/tests/DataContextTest.cpp
DavidSCN: No, this would be ambiguous from the user perspective and it is also ensured via an assertion.

DavidSCN:
> I don't really like it that the complexity of the configuration is increased, but I think this is just necessary. Or do you by chance have an idea how this could be shifted?

You mean the configuration in ParticipantConfiguration? This was the actual bug which lead downstream to further code changes. The increase in complexity is to some extend the price for the capability of preCICE to handle such cases. For now, I would go with the current version until we have tests covering everything properly. Maybe one can find simplifications later on.

docs/changelog/1267.md
```diff
@@ -0,0 +1 @@
+- Fixed the wrong data assignment for multiple read mappings to the same mesh or multiple write mappings from the same mesh [#1267](https://github.com/precice/precice/pull/1267).
```
fsimonis, line 1: No PR number needed.

```suggestion
- Fixed the wrong data assignment for multiple read mappings to the same mesh or multiple write mappings from the same mesh.
```

src/precice/impl/DataContext.cpp
BenjaminRodenberg: I guess it does not make a lot of sense to add the same mapping twice, does it? If you can add a safeguard here, this would be good.

src/precice/impl/DataContext.cpp
DavidSCN: I went for the first, as I need the index in the loop.

DavidSCN:
@BenjaminRodenberg I tried to fix all the documentation related things. Please comment if there are still unclear.

src/precice/config/ParticipantConfiguration.cpp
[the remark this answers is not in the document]
uekerman:
```suggestion
                  "Participant \"{}\" defines a write mapping from mesh \"{}\" to mesh \"{}\", "
                  "but there is either no corresponding write-data tag or the meshes used " 
                  "by this participant lack the necessary use-data tags.",
                  participant->getName(), mappingContext.mapping->getInputMesh()->getName(), mappingContext.mapping->getOutputMesh()->getName());
```

BenjaminRodenberg (approved):
Looks good. I found a few more things, but they are not critical and from my Waveform/DataContext perspective I think you can just merge this PR, when you have taken a look at my last remarks.

src/precice/impl/DataContext.hpp
DavidSCN: I don't have a strong opinion on this, so I just changed to the append version.

src/precice/impl/DataContext.cpp
```diff
@@ -84,40 +91,40 @@ bool DataContext::hasMapping() const
...
 
-  if ((not mapNow) || hasMapped) {
-    return false;
-  }
-
-  return true;
+  return std::any_of(_mappingContexts.begin(), _mappingContexts.end(), [](const auto &context) {
+    const auto timing = context.timing;
+    const bool mapNow = (timing == mapping::MappingConfiguration::ON_ADVANCE) || (timing == mapping::MappingConfiguration::INITIAL);
+    return (mapNow && !context.hasMappedData); });
```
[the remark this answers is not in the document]
DavidSCN: I really like the concise coding style of the version above. It helps me to understand faster what's going on. Now that you leave it to me, I will stick to it.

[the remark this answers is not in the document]
BenjaminRodenberg: Sounds fair ;)

Issue #926, opened 2020-12-30
precice 2.1.1 pkgbuild fails on Arch


floli:
Hey There!

On a recent Arch system (`boost 1.75.0-2`) I try to build precice using the attached [PKGBUILD.log](https://github.com/precice/precice/files/5755324/PKGBUILD.log) (remove the .log suffix), relevant build commands are:
```
cmake . \
          -DCMAKE_BUILD_TYPE=None \
          -DCMAKE_INSTALL_PREFIX=/usr \
          -DBUILD_SHARED_LIBS=ON \
          -DPRECICE_MPICommunication=ON \
          -DPRECICE_PETScMapping=OFF \
          -DPRECICE_PythonActions=ON \
          -Wno-dev
    make
```
however the build fails with an exploding cascade of build errors, starting with
```
In Datei, eingebunden von /usr/include/boost/geometry/core/tag.hpp:23,
                 von /usr/include/boost/geometry/core/ring_type.hpp:28,
                 von /usr/include/boost/geometry/core/closure.hpp:24,
                 von /usr/include/boost/geometry/geometry.hpp:34,
                 von /usr/include/boost/geometry.hpp:17,
                 von /home/florian/software/AUR/precice/src/precice-2.1.1/src/mesh/RTree.hpp:4,
                 von /home/florian/software/AUR/precice/src/precice-2.1.1/src/mapping/NearestNeighborMapping.cpp:10:
/usr/include/boost/geometry/util/type_traits_std.hpp:54:25: Fehler: expected template-name before »<« token
   54 |     : std::conditional_t<Trait::value, conjunction<Traits...>, Trait>
      |                         ^
/usr/include/boost/geometry/util/type_traits_std.hpp:54:25: Fehler: expected »{« before »<« token
/usr/include/boost/geometry/util/type_traits_std.hpp:68:25: Fehler: expected template-name before »<« token
   68 |     : std::conditional_t<Trait::value, Trait, disjunction<Traits...>>
      |                         ^
```
[precice-build.log](https://github.com/precice/precice/files/5755322/precice-build.log)

Best Regards!

fsimonis:
Nice to hear from you @floli!

Boost geometry moved to C++14, which thankfully required us to upgrade as well.
We already bumped the version in develop.

This means that preCICE 2.1.1 won't build on Arch Until the next release. _Unless patched of course_

fsimonis:
This will be resolved with the new release.

floli:
[...]

Issue #1155, opened 2021-12-21
Code formatting ignores line length in documentation / comments


BenjaminRodenberg:
Currently our code formatter seems to ignore the line length in documentation / comments. This makes comments hard to read in github, when reviewing (see https://github.com/precice/precice/pull/1029#discussion_r772205012). Our pipeline that checks for correct formatting, for example, does not complain about this line:

https://github.com/precice/precice/blob/9b3f6b7fa7e78122931db98cf34abb76ac656206/src/time/Waveform.hpp#L46-L49

DavidSCN:
This is actually not specific to commented code, but to all our code (line-breaks in error messages were mostly manually added). The reason is the following line in our formatting settings

https://github.com/precice/precice/blob/6da260d3385d1bb81a9ffbf37f4e04c86510da2f/.clang-format#L47

I would (again) strongly support that we add a limit, since the current setting also allows some ambiguity (e.g. adding line-breaks in function declarations or else-where).

fsimonis:
I am in favour of automatic as much as possible in regards of styling.

We need to agree on a version of clang-format to keep formatting consistent.
Then we can agree on a configuration and apply it to the code base.

DavidSCN:
AFAIK we agreed on `clang-format-8` in our code-base? We could increase the version, of course.

fsimonis:
Ubuntu 20.04 LTS provides clang-format 10.0 this is also available for Alpine 3.12 (CI image)

So, we could increase the version to clang-format 10 without issues.

DavidSCN:
For the sake of completeness: on ubuntu 20.04 you can get all clang-format version between 6 and 12. Upgrading is a good idea, though.

The more interesting question is probably the actual `ColumnLimit`. I would suggest a small limit of around `100`, because it would allow a window splitting and reading complete code side-by-side.

MakisH:
FYI, on GitHub on the desktop, the limit seems to be 160 characters. So, anything below that should be fine. I agree with the tighter limit for the same reasons.

fsimonis:
Decision: upgrade to version 12 and try out various line lengths (100, 120, 160)
When: directly before the release to prevent conflicts

DavidSCN:
Here a summary (note that there are already quite a lot of changes removing the current ambiguity, i.e., the code does not exceed the column limit)

- limit: 160: 329 files changed, 4127 insertions(+), 8930 deletions(-)
[...]

Issue #1026, opened 2021-05-19
Rename MasterSlave to something more appropriate


fsimonis:
**Please describe the problem you are trying to solve.**
In preCICE we use the terms master and slave to distinguish  between multiple ranks running in parallel.
These terms are archaic and have some baggage.
Alternatives have been discussed at about every other project meeting and it is finally time to come to a conclusion and move on.

Moving away from these wide-spread terms will break a widespread convention, but if one always follows conventions, one can never improve.

There are 3 main execution paths in preCICE:
* a Participant runs on a single process (size = 1)
* a Participant runs on multiple parallel processes (size > 1) and
  * the current rank is the "main rank" (aka master) who sometimes takes the leading role in coordination (rank = 0)
  * the current ranks is a rank who sometimes needs to be coordinated (aka slave) the aforementioned rank (rank > 0)

The direct interface to this information is the [`MasterSlave`](https://precice.org/doxygen/develop/classprecice_1_1utils_1_1MasterSlave.html) utility class, which provides this information based on its configured `size` and `rank`.

This terminology is also used for the communication between participants.
preCICE first connects master ranks of participants and then establishes a connection between slaves. 

**Alternative name for the class `MasterSlave`**

We use communication in 2 cases:
* inside a participant (MasterSlave) and
* between participants (M2N).

I propose to use another very common terminology from networking:
* use intra-communication (as in IntraNet) for the communication between ranks of a single participant
* use inter-communication  (as in InterNet) for the communication between participants

This allows us to rename `utils::MasterSlave` to `utils::IntraComm` or `utils::IntraCommunication`.

**Alternative names for the states master and slave**

The following is a table of alternatives I came across while researching this and that I picked up in discussions.
I numbered them to simplify the discussion.

Nr | master | slave | Opinion
--- | --- | --- | ---
1 | master | follower | :-1: :-1: :-1:
2 | leader | follower | :-1: :-1: :-1:
3 | primary | secondary | :+1: :+1: :+1: 
4 | main | secondary | :+1: :-1:
5 | agency | agent | :-1: :-1:
6 | coordinator | coordinated | :-1: :-1: 
7 | coordinator | node | :-1:
8 | controller | worker | :+1: :-1: :-1: 
9 | supervisor | worker | :+1: :-1: :-1:
10 | parent | child | :+1: :-1: :-1:

My current favourite is (3) as it does not impose additional relations.
However, it still isn't really fitting as for a non-parallel case, `isPimary() = false`.

[...]

PR #1271, opened 2022-04-26
Remove vector overloads in communication

fsimonis:
## Main changes of this PR

This PR removes the `std::vector` overloads of the `Communication` class and replaces them with `sendRange` and `receiveRange` variants.
These variants send the size + content and are purely implemented using the `Communication` interface.
The current `receiveRange(rank, AsVectorTag<T>)` returns a `std::vector<T>`, which simplifies many call sites.

I used the `[[deprecate]]` attribute to ensure that I checked every single call of these overloads.

The primary goal of this PR is to unify the behaviour of the overload sets of `send`, `receive`, `aSend`, `aReceive` of the `Communication` class.

The bug-prone issue:
* `std::vector<T>` overloads would send/receive the size, then the content
* `T` and `span<T>` overloads would only send/receive the content

Mixing these overloads was very easy to do and very difficult to spot as shown in #926.

Now, the `send` calls send a range or a single itema and `receive` calls receive a range or a single item.
Use `sendRange` to send size + content and `receiveRange` to receive a size + content.

## Author's checklist

## Reviewers' checklist


fsimonis:
> Question: wouldn't it be better if we got completely rid of one of the two variants? What would be an argument against this?

I initially tried to remove communicating (size + content) completely. This becomes extremely verbose though.
The partitioning code is already rather complex and introducing even more communication code makes this a total nightmare to follow.
This is why I did a full U-turn and implemented a method that simply returns the range as a vector. This managed to reduce some boilerplate in the partitioning.

fsimonis:
We should merge this after #1026 and before #1155

PR #1276, opened 2022-05-02
Update code base formatting

### LINKED ISSUE #1155: Code formatting ignores line length in documentation / comments

davidscn:
## Main changes of this PR
Upgrades the `clang-format` version to `clang-format-12`, introduces a column limit of 120 as discussed in #1155 and applies corresponding changes to the code base. Resolves #1155. In case anyone has formatting objections, now would be a good time to mention them. 

## Author's checklist

## Reviewers' checklist


extras/bindings/c/include/precice/SolverInterfaceC.h
```diff
@@ -356,19 +311,12 @@ void precicec_writeBlockVectorData(
...
 
 /**
  * @brief See precice::SolverInterface::writeBlockScalarData().
  */
-void precicec_writeBlockScalarData(
-    int           dataID,
-    int           size,
-    const int *   valueIndices,
-    const double *values);
+void precicec_writeBlockScalarData(int dataID, int size, const int *valueIndices, const double *values);
```
DavidSCN, line 319: With this changes, `clang-format` resolves the current formatting ambiguity of our code base (e.g. both code versions would have been accepted in our current formatting settings). However, having multiple function arguments on multiple lines might be desirable. Do we want to enforce something line breaks in function arguments or go with the current version?

fsimonis: I prefer one item per line. I find that it makes the signature easier to digest.

DavidSCN: Yes I agree, these arguments
```
BinPackArguments: false
BinPackParameters: false
```
would set the arguments on separate lines if they don't fit into the 120 characters. Or do you want to enforce the separate line for all functions?

fsimonis:
Let's move the reformatting behind the 2.4.0 release to avoid breaking _at least_ #1279 #1258 #1271.

PR #1279, opened 2022-05-04
Refactor RBF mappings to have a common base class

DavidSCN:
## Main changes of this PR
Moves common content of the PETSc RBF mappings and the eigen RBF mappings to a common base class.

There are several functions duplicated in both methods.

## Author's checklist

## Reviewers' checklist


fsimonis:
Great effort!

I always wondered how much logic could be shifted into an rbf base class, which doesn't even depend on the actual basis function itself.
This could reduce compilation time by quite a lot.

DavidSCN:
> which doesn't even depend on the actual basis function itself.

I also thought about this, but it requires some additional effort. In order to keep a reasonable PR size, I would go with the current solution.

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -51,101 +51,58 @@ class RadialBasisFctMapping : public Mapping {
...
-  setDeadAxis(xDead, yDead, zDead);
 }
 
 template <typename RADIAL_BASIS_FUNCTION_T>
 void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::computeMapping()
 {
   PRECICE_TRACE();
 
-  precice::utils::Event e("map.rbf.computeMapping.From" + input()->getName() + "To" + output()->getName(), precice::syncMode);
+  precice::utils::Event e("map.rbf.computeMapping.From" + this->input()->getName() + "To" + this->output()->getName(), precice::syncMode);
```
boris-martin, line 90: What is the added value of adding "this" ?

boris-martin:
Looks great to me! I didn't check the whole implementation of each subclass but the subdivision makes sense to me. Just curious about all the added `this` who seem redundant to me.

src/mapping/RadialBasisFctMapping.hpp
DavidSCN: The compiler forces me to explicitly refer to the base class member variables here, as the base class is now a class template with a template parameter depending on the instantiation of the derived class. If not put here, the compiler will not find the variable in the first place.

PR #1274, opened 2022-04-29
Renaming class MasterSlave to IntraComm and corresponding references

IshaanDesai:
## Main changes of this PR

This PR is a follow-up of https://github.com/precice/precice/pull/1253 and completes the renaming process. In particular the renaming of the class `MasterSlave` to `IntraComm` is done, along with the cascading name changes that are required.

## Author's checklist

* [x] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [x] I ran `make format` to ensure everything is formatted correctly.
* [x] I sticked to C++14 features.
* [x] I sticked to CMake version 3.16.3.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

PR #1281, opened 2022-05-05
Change default configuration of filtering to "on-secondary-ranks"

DavidSCN:
## Main changes of this PR
As part of #1274, we forgot to update the default configuration of the participant. In the current version, the user gets a warning about deprecation without having set the corresponding deprecated tag. 

## Author's checklist

## Reviewers' checklist


fsimonis (approved):
Good catch

PR #1286, opened 2022-05-10
Keep triangles in mesh communication/filtering in 2D

boris-martin:
## Main changes of this PR

Split of https://github.com/precice/precice/pull/1258. Remove restrictions where triangles woudn't be sent during mesh comunication/filtering in 2D. This is now allowed, and in cases where it shouldn't happen, triangle creation is already prevented elsewhere anyway.

Volumetric coupling: https://github.com/precice/precice/issues/468

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/com/CommunicateMesh.cpp
+++ b/src/com/CommunicateMesh.cpp
@@ -67,26 +67,24 @@ void CommunicateMesh::sendMesh(
     _communication->send(edgeIDs, rankReceiver);
   }
 
-  if (dim == 3) {
-    int numberOfTriangles = mesh.triangles().size();
-    _communication->send(numberOfTriangles, rankReceiver);
-    if (not mesh.triangles().empty()) {
-      //we need to send the edgeIDs first such that the right edges can be created later
-      //contrary to the normal sendMesh, this variant must also work for adding delta meshes
-      std::vector<int> edgeIDs(numberOfEdges);
-      for (int i = 0; i < numberOfEdges; i++) {
-        edgeIDs[i] = mesh.edges()[i].getID();
-      }
-      _communication->send(edgeIDs, rankReceiver);
-
-      std::vector<int> triangleIDs(numberOfTriangles * 3);
-      for (int i = 0; i < numberOfTriangles; i++) {
-        triangleIDs[i * 3]     = mesh.triangles()[i].edge(0).getID();
-        triangleIDs[i * 3 + 1] = mesh.triangles()[i].edge(1).getID();
-        triangleIDs[i * 3 + 2] = mesh.triangles()[i].edge(2).getID();
-      }
-      _communication->send(triangleIDs, rankReceiver);
+  int numberOfTriangles = mesh.triangles().size();
+  _communication->send(numberOfTriangles, rankReceiver);
+  if (not mesh.triangles().empty()) {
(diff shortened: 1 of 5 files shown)


fsimonis:
This changes the API and will need a changelog entry.

> Added the option to set triangles on 2D meshes.

fsimonis:
This changes the API and will need a changelog entry.

> Added the option to set triangles on 2D meshes.

src/com/CommunicateMesh.cpp
```diff
@@ -67,27 +67,27 @@ void CommunicateMesh::sendMesh(
     _communication->send(edgeIDs, rankReceiver);
   }
 
-  if (dim == 3) {
```
fsimonis, line 70: This should be removed instead of being commented out.

src/com/CommunicateMesh.cpp
```diff
@@ -67,27 +67,27 @@ void CommunicateMesh::sendMesh(
...
-      }
-      _communication->send(triangleIDs, rankReceiver);
+    std::vector<int> triangleIDs(numberOfTriangles * 3);
+    for (int i = 0; i < numberOfTriangles; i++) {
+      triangleIDs[i * 3]     = mesh.triangles()[i].edge(0).getID();
+      triangleIDs[i * 3 + 1] = mesh.triangles()[i].edge(1).getID();
+      triangleIDs[i * 3 + 2] = mesh.triangles()[i].edge(2).getID();
     }
+    _communication->send(triangleIDs, rankReceiver);
   }
```
fsimonis, line 90: Same as above

fsimonis (commented):
Can you check again for comments near your changes?

src/com/CommunicateMesh.cpp
```diff
@@ -67,27 +67,27 @@ void CommunicateMesh::sendMesh(
...
+  //}
```
fsimonis, line 90: Same as above

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -828,8 +828,7 @@ void SolverInterfaceImpl::setMeshTriangle(
 {
   PRECICE_TRACE(meshID, firstEdgeID,
                 secondEdgeID, thirdEdgeID);
-  PRECICE_CHECK(_dimensions == 3, "setMeshTriangle is only possible for 3D cases."
-                                  " Please set the dimension to 3 in the preCICE configuration file.");
+
```
DavidSCN, line 831: @boris-martin allow me a question from my naive point of view: how does setting triangles make sense in case a nearest-projection mapping was configured? Shouldn't this check be conditionally disabled?

boris-martin: Right now it doesn't make sense to set triangles in NP mapping indeed. But "3D" is no longer a sufficient condition to disable triangles.
In the long run I think they'll be more common anyways, for instance if we want to allow watch integrals over volume instead of surfaces.

In the case of setMeshTriangle, the connectivity requirement will be reworked to be more specific, so the `context.meshRequirement == mapping::Mapping::MeshRequirement::FULL` will be replaced with a more fine-grained thing. See https://github.com/precice/precice/pull/1203 (which is a draft I'll have to rework)

PR #1290, opened 2022-05-13
Improve communication info publisher

fsimonis:
## Main changes of this PR

This PR extends read and write of connection files to be more strict and robust.

The most common issues are now actively checked and result in error messages.
Messages guide the user to either fix or report problems.

Problems with connection files are tricky to debug as the writers and readers don't actively check conditions.

## Author's checklist

## Reviewers' checklist

* [x] Do you agree with the verbosity and the wording?
* [ ] @KyleDavisSA Can you try this on SuperMUC and see if these errors and warnings get triggered by the synchronization delay of filesystem system?

### DIFF
--- a/src/com/ConnectionInfoPublisher.cpp
+++ b/src/com/ConnectionInfoPublisher.cpp
@@ -61,19 +61,24 @@ std::string ConnectionInfoReader::read() const
 {
   auto path = getFilename();
 
-  PRECICE_DEBUG("Waiting for connection file {}", path);
+  PRECICE_DEBUG("Waiting for connection file \"{}\"", path);
   const auto waitdelay = std::chrono::milliseconds(1);
   while (!bfs::exists(path)) {
     std::this_thread::sleep_for(waitdelay);
   }
   PRECICE_ASSERT(bfs::exists(path));
-  PRECICE_DEBUG("Found connection file {}", path);
+  PRECICE_DEBUG("Found connection file \"{}\"", path);
 
   std::ifstream ifs(path);
-  PRECICE_CHECK(ifs, "Unable to open connection file {}", path);
+  PRECICE_CHECK(ifs,
+                "Unable to establish connection as the connection file \"{}\" couldn't be opened.",
+                path);
   std::string addressData;
   std::getline(ifs, addressData);
-  PRECICE_CHECK(!addressData.empty(), "The connection file {} is empty. Please report this bug to the preCICE developers.", path);
+  PRECICE_CHECK(!addressData.empty(),
+                "Unable to establish connection as the connection file \"{}\" is empty. "
+                "Please report this bug to the preCICE developers.",
+                path);
   boost::algorithm::trim_right(addressData);
   return addressData;
 }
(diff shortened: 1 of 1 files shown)


uekerman (commented):
Error messages look good. Only doubt I had was whether users know what you mean with "connection file". Maybe a intro sentence like "There was a problem establishing the m2n communication between participants X and Y." could help?

I guess you have tested to trigger some of them locally? (e.g. with leftover files from a previous run)

KyleDavisSA:
I managed to test this with a few ranks over 2 nodes on superMUC-NG. No problems found. Next I will try something that uses many ranks over more nodes to really test the issue.

fsimonis:
> Only doubt I had was whether users know what you mean with "connection file". Maybe a intro sentence like "There was a problem establishing the m2n communication between participants X and Y." could help?

This is tricky, as this contains information across multiple layers. The connection info publisher isn't aware of the communication, which isn't aware of the participant names.

uekerman:
> This is tricky, as this contains information across multiple layers. The connection info publisher isn't aware of the communication, which isn't aware of the participant names.

OK, but we could still do a generic: "There was a problem establishing a communication channel." or similar.

src/com/ConnectionInfoPublisher.cpp
```diff
@@ -49,56 +50,103 @@ std::string ConnectionInfoPublisher::getLocalDirectory() const
...
-  fs::create_directories(tmp.parent_path());
+  auto path = getFilename();
+  auto tmp  = bfs::path(path + "~");
+
   {
-    std::ofstream ofs(tmp.string(), std::ofstream::out);
-    ofs << info << "\n";
-    ofs << "Acceptor: " << acceptorName << ", Requester: " << requesterName << ", Tag: " << tag << ", Rank: " << rank << "\n";
+    auto message = "Unable to establish connection as a {}connection file already exists at \"{}\". "
+                   "This is likely a leftover of a previous crash during communication build-up. "
```
uekerman, line 118:
```suggestion
                   "This is likely a leftover of a previous crash or stop during communication build-up. "
```

uekerman (approved):
Did some tests on my laptop. Error messages look good.

src/com/ConnectionInfoPublisher.cpp
```diff
@@ -49,56 +50,103 @@ std::string ConnectionInfoPublisher::getLocalDirectory() const
...
+  }
+  PRECICE_DEBUG("Deleting connection file \"{}\"", path);
   try {
-    fs::path p(getFilename());
-    PRECICE_DEBUG("Deleting connection file {}", p.string());
-    fs::remove(p);
-  } catch (const fs::filesystem_error &e) {
-    PRECICE_WARN("Unable to delete connection file due to error: {}", e.what());
+    bfs::remove(path);
+    /// @TODO Check if we need to sleep to let the underlying filesystem synchronize
```
DavidSCN, line 98: Do you want to check this before merging? How do you want to check this?

KyleDavisSA:
Still struggling to run ASTE on SNG. However the other test cases all ran without issues.

KyleDavisSA (approved):
Earlier SNG tests ran without issue.

src/com/ConnectionInfoPublisher.cpp
fsimonis: We need to run a test case with a distributed filesystem running on separated nodes preferably with a slow inter-node connection.

fsimonis:
@DavidSCN Do you have a large parallel ASTE setup that you can quickly run on a distributed system?

I really want to avoid this breaking for HPC users :wink:

DavidSCN:
> @DavidSCN Do you have a large parallel ASTE setup that you can quickly run on a distributed system?

Lacking a large cluster. I could run it on our institute machine on 16 x 6 nodes?!

fsimonis:
@DavidSCN Sounds like a perfect fit!

DavidSCN (approved):
I ran now the test cases using ASTE on our cluster with 42  cores  (6 x 7 nodes) per participant. One testcase was the our most refined reference paper setup (turbine) with a NN mapping, which ran three times in a loop and one was the most coarse reference paper setup with a NN mapping, which ran five times in a loop. Everything worked fine in `Release` and `Debug` mode using the changes proposed here.

DavidSCN:
Could you rebase before merging? Merging will fail due to the fmt upgrade (boost::filesystem::path).

src/com/ConnectionInfoPublisher.cpp
fsimonis: For the record https://github.com/precice/precice/pull/1290#pullrequestreview-1060192506

DavidSCN:
Only a changelog missing, then we are ready to go.

PR #1291, opened 2022-05-16
adds a pre-commit config for clang-format

renefritze:
## Main changes of this PR

Adds a config for [pre-commit](https://pre-commit.com/)

I saw the release announcement mention a target to run clang-format
and thought preCICE could use some pre-commit hooks to automate that.
Basically `pre-commit` is a framework to configure, install and run
git hooks. For clang-format there's a [hook implementation](https://github.com/pre-commit/mirrors-clang-format) available
that installs clang-format from [binary wheels](https://pypi.org/project/clang-format/#history). That might be valuable
even if you do not want to use pre-commit itself.
I usually run a [CI step](https://github.com/pymor/pymor-deal.II/blob/main/.github/workflows/checks.yml#L24) that executes all hooks and fails if not all hooks pass in my projects.
I can add that to this PR if there's interest for it.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,0 +1,8 @@
+# See https://pre-commit.com for more information
+# See https://pre-commit.com/hooks.html for more hooks
+repos:
+-   repo: https://github.com/pre-commit/mirrors-clang-format
+    rev: 'v8.0.1'
+    hooks:
+    -   id: clang-format
+        exclude: "^thirdparty"


fsimonis:
Thank you very much for this very interesting suggestion!

Especially this pip clang-format package looks super useful!

MakisH (commented):
Thanks a lot for the contribution! I tried this out. Trying to commit a wrongly-formatted file is not allowed: when committing, the hook reports:
```
$ git commit -m "testing hook"
clang-format.............................................................Failed
- hook id: clang-format
- files were modified by this hook
```

The hook updates the local file, which I can stage again and commit. :heavy_check_mark: 

I understand that this project provides a very easy way to add some common pre-commit hooks to Git, without writing scripts yourself (so... something like "Git Actions").

The config looks straight-forward and reasonable, to my understanding.

Some discussion points:
- It is the first time I hear about this project. Any clue how established it is at the moment?
- How can we make sure that we will not forget about this in the future and that this and our `check-format` script don't deviate? Can we integrate the one into the other (in either way)? This does not seem trivial to me, especially since we probably want to keep this as an option and not as a dependency for developers. I guess that we could just have an additional step in the CI (duplicate checks, but we do want to maintain duplicate systems).
- I understand that we should not enforce that hooks are used on the contributor's side: we should just catch any issues on our CI. This should make contribution easier.
- We should also document it here: https://precice.org/dev-docs-dev-tooling.html#formatting-the-code

renefritze:
> * It is the first time I hear about this project. Any clue how established it is at the moment?
I've been using it in projects for over a year now. Judging from their PyPi history it's ~8 years old.

> * How can we make sure that we will not forget about this in the future and that this and our `check-format` script don't deviate? Can we integrate the one into the other (in either way)? This does not seem trivial to me, especially since we probably want to keep this as an option and not as a dependency for developers. I guess that we could just have an additional step in the CI (duplicate checks, but we do want to maintain duplicate systems).

I had a quick look at your `check-format` script and I'd say if you can move the XML formatting to a hook script too, than `check-format` could basically just call `pre-commit` and be done with it. Unless using `parallel` is a must-have option.

Since you already have python as a dependency and using (the right version of) clang-format is much simpler via the python wheel, I'd say making pre-commit a dependency instead would overall be less complex to manage.

> * I understand that we should not enforce that hooks are used on the contributor's side: we should just catch any issues on our CI. This should make contribution easier.

Right. I should mention that there's also a [dedicated service for pre-commit](https://pre-commit.ci), however I haven't been convinced that it brings enough value over just running a github action.

Issue #1206, opened 2022-03-17
Naming of Nearest Projection in a Volume coupling context


boris-martin:
**Please describe the problem you are trying to solve.**
Currently, NP mapping only supports surface (i.e. boundary) coupling and is not suited to volumetric coupling. Extending it raises the question of naming: volumetric interpolation is not actually a *projection*.

-  How do we name the new method? Do we rename the surface version?
-  Should it be one mapping with a surface/volume choice as option, or two completely different mappings? In terms of code, both are doable in a +/- clean way (have a base abstract class and 2 subclasses, or one class with an option)

**Describe the solution you propose.**
On my local prototype, I named "Barycentric mapping" the volumetric version and left the current NP untouched. In terms of config, they are completely different mappings: one use either `mapping:nearest-projection` or `mapping:barycentric` in the preCICE config.
However, the name is misleading, as the surface NP is also barycentric.

**Describe alternatives you've considered**

- Same as above, but name it "volume mapping" instead. Removes the barycentric dilemma, but is weird since NN and RBF can be used in a volumetric context, so defining it as *the* volumetric mapping seems inappropriate.
- Have one unique mapping called "barycentric" (or something like "linear interpolation"?) and a volume/surface flag. Annoying since it "hides" the projection part in the surface version, and it breaks continuity ("nearest projection" has been written many times in a lot of places)
- Keep calling it NP and have a unique class, and keeping the misleading "projection" name even in the volume case.

**Additional context**

As of now, all interpolation is done / planned using edges, triangles and tetrahedras, hence the barycentric name. If we extend to quads, hexaedrons, higher order triangles, etc. the name will become inappropriate. Similarly with "linear interpolation".
=> Maybe think about a name like "local interpolation" to make it more generic?

boris-martin:
Poke @fsimonis @uekerman

uekerman:
I would go for two separate mappings and leave the nearest-projection as it is (backwards compatibility is indeed a strong argument here). The name for the new mapping could `mapping:cell-interpolation` or `mapping:volume-cell-interpolation`? Not sure yet here.
All this is for the user perspective, how to implement and name this internally, is a different question in my opinion.

So far my quick response, I did not yet think about this carefully.

PR #1297, opened 2022-05-18
[V2] Added "Volume Cell Interpolation" mapping support in 2D

boris-martin:
## Main changes of this PR

- Added a new data mapping, "Volume Cell Interpolation", similar to NP but designed for volumetric coupling. Only supported in 2D for now. Usage: `volume-cell-interpolation` in the XML config and set up connectivitiy.
- Added unit tests to check the correct interpolation, and integrations tests for conservative & consistent maps, in serial and parallel.

Milestone of https://github.com/precice/precice/issues/468.
Followup of https://github.com/precice/precice/pull/1258

## Author's checklist

## Reviewers' checklist

@fsimonis : there is still the fall-back question to answer.
For now, we fall-back in NN when the vertex isn't inside a good triangle. However fall-back on edge could be useful in some edge cases (no pun intended).
If we consider the triangle formed by (0,0), (1,0) (0,1) and a point on (-eps, 0.5) with eps a small number, we clearly want to interpolate linearly in the edge. However it is not *in the triangle* thus falls back to NN


boris-martin:
Relevant discussions from the previous PR:

>> We should also cover scaled-consistent mappings. Have a look at NP scaled-consistent tests.
>
> If I understand it right, scaled consistent computes surface integrals, no ? Shouldn't we wait for the 3D version of this to add scaled-consistent ?

And 

> You don't need to gather distance statistics as this is an interpolation-only mapping.
> It would make more sense to count vertices which need to fall-back to NN as they are not inside a cell and display a warning if this is larger than 0.
> 
> Maybe additionally the maximum distance to the nearest vertex could be useful.

src/query/Index.cpp
```diff
@@ -211,6 +211,25 @@ ProjectionMatch Index::findNearestProjection(const Eigen::VectorXd &location, in
...
 }
 
+ProjectionMatch Index::findCellInterpolation(const Eigen::VectorXd &location, int n)
+{
+  if (_mesh->getDimensions() == 2) {
+    auto matchedTriangles = getClosestTriangles(location, n);
+    for (const auto &match : matchedTriangles) {
+      auto polation = mapping::Polation(location, _mesh->triangles()[match.index]);
+      if (polation.isInterpolation()) {
+        return {polation, 0.0};
```
boris-martin, line 221: @fsimonis I made this change (return 0 instead of a distance) such that non-zero distances only occur in case of fall-back.

Maybe I should also do the `distanceStatistics(match.distance);` only when this distance is non-zero ? This way the distance statistics would count the number of fall-backs.

fsimonis: Well, the distance is 0, so this makes sense.

Good idea, but then you need to rename the `distanceStatistics` to `fallbackStatistics` or similar.

src/mapping/VolumeCellInterpolation.hpp
```diff
@@ -0,0 +1,33 @@
+#pragma once
+
+#include <list>
+#include <string>
+#include <vector>
```
fsimonis, line 5: We separate these by a blank line.

```suggestion
#include <vector>

```

fsimonis (commented):
You are right regarding the edge edge-case. What we really want is a nearest-projection which starts one dimension higher.
So, you can basically call the nearest projection logic if we don't find a cell.

If we find a cell, then the distance will be 0. Everything else is a fallback.

This makes me wonder about the name once again. So let's continue in #1206.

_edit: Such edge-cases would mean that the meshes are not really suitable for cell-interpolation. Not sure how valid the use-case of outlying vertices is._

src/query/Index.cpp
```diff
@@ -211,6 +211,25 @@ ProjectionMatch Index::findNearestProjection(const Eigen::VectorXd &location, in
...
+      }
+    }
+
+    // If no triangle is found, fall-back on NN
+    return findVertexProjection(location);
```
fsimonis, line 226: Use `findNearestProjection` here.

src/query/Index.cpp
```diff
@@ -211,6 +211,25 @@ ProjectionMatch Index::findNearestProjection(const Eigen::VectorXd &location, in
   }
 }
 
+ProjectionMatch Index::findCellInterpolation(const Eigen::VectorXd &location, int n)
```
fsimonis, line 214: This will turn in something like `findNearestCell`.

src/mapping/VolumeCellInterpolation.cpp
```diff
@@ -0,0 +1,114 @@
...
+  if (constraint == CONSISTENT) {
+    setInputRequirement(Mapping::MeshRequirement::FULL);
+    setOutputRequirement(Mapping::MeshRequirement::VERTEX);
+  } else if (constraint == CONSERVATIVE) {
+    setInputRequirement(Mapping::MeshRequirement::VERTEX);
+    setOutputRequirement(Mapping::MeshRequirement::FULL);
+  } else {
+    PRECICE_ASSERT(constraint == SCALEDCONSISTENT, constraint);
+    setInputRequirement(Mapping::MeshRequirement::FULL);
+    setOutputRequirement(Mapping::MeshRequirement::FULL);
```
DavidSCN, line 43: I guess in the end it would be nice to move this whole specification into the `Mapping` class, i.e., the base class as I read something like this all over the place in our preCICE mappings. Maybe an override for a specific `VolumeCellInterpolation` would be possible then...

src/mapping/VolumeCellInterpolation.cpp
```diff
@@ -0,0 +1,114 @@
...
+  }
+
+  PRECICE_ASSERT(constraint != SCALEDCONSISTENT, "Volume mapping doesn't support scaled-consistent mappings.");
```
DavidSCN, line 46: Do you catch this case before already? If this is something a user could face, please use `PRECICE_CHECK` instead of the assert.

src/mapping/VolumeCellInterpolation.cpp
```diff
@@ -0,0 +1,114 @@
...
+    }
+  }
+
+  // Amount of nearest elements to fetch for detailed comparison.
+  // This safety margin results in a candidate set which forms the base for the
+  // local nearest projection and counters the loss of detail due to bounding box generation.
+  // @TODO Add a configuration option for this factor
+  constexpr int nnearest = 4;
+
+  auto                                  &index = searchSpace->index();
```
DavidSCN, line 90:
```suggestion
 const auto                                  &index = searchSpace->index();
```
?

src/mapping/VolumeCellInterpolation.hpp
```diff
@@ -0,0 +1,33 @@
...
+ * @brief Mapping using orthogonal projection to nearest triangle/edge/vertex and
+ *        linear interpolation from projected point.
+ */
+class VolumeCellInterpolation : public BarycentricBaseMapping {
+public:
+  /// Constructor, taking mapping constraint.
+  VolumeCellInterpolation(Constraint constraint, int dimensions);
+
+  /// Destructor, empty.
+  virtual ~VolumeCellInterpolation() {}
```
DavidSCN, line 23:
```suggestion
```
In case you don't want to further derive classes from this class, you can omit the `virtual`. Considering that the destructor is empty, you can just omit it as well.

fsimonis:
@boris-martin and I discussed the fallback and there is actually a very valid use-case for this.

One geometry and two meshes of varying discretization. A sphere would is a good example here and the same as we use for the nearest projection mapping.

Decision: We collect fallbacks and print an info statement similar to the one in the nearest projection.
Difference here is that there are 2 cases which are interesting to the user and hence need log statements:
1. there are fallbacks and these are the distance stats
2. there are no fallbacks

@uekerman and I discussed the naming again.
* Adding "nearest" to the name would be confusing.
* We need to inform that we use linear interpolation.
* It's not really volume interpolation, as it also works on the area of cells.

Our best name we could come up with is: `linear-cell-interpolation`.

src/mapping/tests/VolumeCellInterpolationTest.cpp
```diff
@@ -0,0 +1,92 @@
...
+  */
+
+  BOOST_CHECK(!inMesh->triangles().empty());
+  PtrMesh outMesh(new Mesh("OutMesh", dimensions, testing::nextMeshID()));
+  PtrData outDataScalar   = outMesh->createData("OutDataScalar", 1, 2_dataID);
+  int     outDataScalarID = outDataScalar->getID();
+
+  // All vertices to test
+  // Center of triangle = average
+  Vertex &outMidPoint = outMesh->createVertex(Eigen::Vector2d::Constant(1.0 / 3.0));
```
DavidSCN, line 59: I did not look into the tests in detail, but there are quite some unused variables.

DavidSCN (commented):
Minor comments

src/mapping/VolumeCellInterpolation.cpp
```diff
@@ -0,0 +1,114 @@
...
+  const auto &fVertices = origins->vertices();
+
+  if (getDimensions() == 2) {
+    if (!fVertices.empty() && searchSpace->triangles().empty()) {
+      PRECICE_WARN("2D Mesh \"{}\" does not contain triangles. "
+                   "Volume cell interpolation falls back to nearest neighbor mapping.",
+                   searchSpace->getName());
+    }
+  } else {
+    if (!fVertices.empty() && searchSpace->triangles().empty()) {
```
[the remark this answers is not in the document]
boris-martin: There's already a storage for triangles from before. For 3d it'll be replaced by `searchSpace->tetrahedra()`... Once they're implemented

src/query/Index.cpp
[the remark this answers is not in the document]
boris-martin: `findNearestProjection` isn't and is similar so I assumed no. Let's add it in another PR if it works on both, as it's a refactoring of NP then.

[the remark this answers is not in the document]
boris-martin: Actually If we make them const, we'd need to make the functions called inside const too, and it's a too significant change for this PR

src/mapping/VolumeCellInterpolation.cpp
boris-martin: Change to do if we ever refactor to make all the sub-calls const too

fsimonis:
Yes, this is a limitation of the current implementation.

To handle this edge-case correctly, we would have to do this lookup for every vertex and thus require all primitives to be always indexed. This would essentially disable the on-demand evaluation of index trees.
So, the trade-off is a performance penalty for every user to solve a problem that rarely ever pops up in practise. 

I suggest ignoring this edge-case in this PR and open an issue to document it.

src/mapping/LinearCellInterpolation.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+  PRECICE_CHECK(constraint != SCALEDCONSISTENT, "Volume mapping doesn't support scaled-consistent mappings.");
+}
+
+void VolumeCellInterpolation::computeMapping()
+{
+  PRECICE_TRACE(input()->vertices().size(), output()->vertices().size());
+  const std::string     baseEvent = "map.vci.computeMapping.From" + input()->getName() + "To" + output()->getName();
+  precice::utils::Event e(baseEvent, precice::syncMode);
+  PRECICE_CHECK(getDimensions() == 2, "Volume mapping not available in 3D.");
```
fsimonis, line 54: This check should be in the constructor. The earlier this triggers, the better.

Let's use a `PRECICE_ASSERT` here instead.

src/mapping/LinearCellInterpolation.cpp
```diff
@@ -0,0 +1,118 @@
...
+    origins     = output();
+    searchSpace = input();
+  }
+
+  const auto &fVertices = origins->vertices();
+
+  if (getDimensions() == 2) {
+    if (!fVertices.empty() && searchSpace->triangles().empty()) {
+      PRECICE_WARN("2D Mesh \"{}\" does not contain triangles. "
+                   "Volume cell interpolation falls back to nearest neighbor mapping.",
```
fsimonis, line 73:
```suggestion
                   "Volume cell interpolation falls back to nearest projection mapping.",
```

src/mapping/LinearCellInterpolation.cpp
```diff
@@ -0,0 +1,118 @@
...
+                   searchSpace->getName());
+    }
+  } else {
+    if (!fVertices.empty() && searchSpace->triangles().empty()) {
+      PRECICE_WARN("3D Mesh \"{}\" does not contain tetrahedra. "
+                   "Volume cell interpolation will map to primitives of lower dimension.",
```
fsimonis, line 79: We should use the same wording as above.

src/mapping/LinearCellInterpolation.cpp
```diff
@@ -0,0 +1,118 @@
...
+  const auto &fVertices = origins->vertices();
+
+  if (getDimensions() == 2) {
+    if (!fVertices.empty() && searchSpace->triangles().empty()) {
+      PRECICE_WARN("2D Mesh \"{}\" does not contain triangles. "
+                   "Volume cell interpolation falls back to nearest neighbor mapping.",
+                   searchSpace->getName());
+    }
+  } else {
+    if (!fVertices.empty() && searchSpace->triangles().empty()) {
```
fsimonis, line 77: This check currently doesn't make sense.

In fact, this entire else block is dead code as this will be checked before.

src/mapping/LinearCellInterpolation.cpp
```diff
@@ -0,0 +1,118 @@
...
+      // Only push when fall-back occurs, so the number of entries is the number of vertices outside the domain
+      fallbackStatistics(match.distance);
+    }
+  }
+
+  if (!fallbackStatistics.empty()) {
+    PRECICE_WARN("Some points are outisde of the domain defined by connectivity. Fall-back on Nearest-Projection occured."
+                " Statistics: {}", fallbackStatistics);
+    
+  }
```
fsimonis, line 111: As mentioned before, a warning only makes sense if we know the element size.

Thus, this should be a `PRECICE_INFO` statement.
`Statistics` is a bit too vague. This should be something like `Projection distances:`.

Also, the `else` path should result in an info statement along the lines of `All vertices are inside cells, no fallback to Nearest-Projection required.`

src/mapping/LinearCellInterpolation.hpp
```diff
@@ -0,0 +1,31 @@
...
+ *        linear interpolation from projected point.
+ */
+class VolumeCellInterpolation : public BarycentricBaseMapping {
+public:
+  /// Constructor, taking mapping constraint.
+  VolumeCellInterpolation(Constraint constraint, int dimensions);
+
+
+  /// Computes the projections and interpolation relations.
+  virtual void computeMapping() override;
```
fsimonis, line 24: `override` implies `virtual`.

```suggestion
  void computeMapping() override;
```

fsimonis (commented):
The class and the file should be renamed `LinearCellInterpolationMapping` to stay consistent with the other mappings.

Also, let's remove unused variables in the tests.

fsimonis:
Last comment from my side is that the name should be `LinearCellInterpolationMapping` to stay consistent with the other mappings.

src/mapping/LinearCellInterpolationMapping.cpp
```diff
@@ -0,0 +1,116 @@
...
+  utils::statistics::DistanceAccumulator fallbackStatistics;
+
+  _interpolations.clear();
+  _interpolations.reserve(fVertices.size());
+
+  for (const auto &fVertex : fVertices) {
+    // Find tetrahedra (3D) or triangle (2D) or fall-back on NP
+    auto match = index.findCellInterpolation(fVertex.getCoords(), nnearest);
+    _interpolations.push_back(std::move(match.polation));
+    if (match.distance != 0.0) {
```
fsimonis, line 98: We need to use a tolerance-based check here: `math::equals(match.distance, 0.0)`

src/mapping/LinearCellInterpolationMapping.hpp
```diff
@@ -0,0 +1,30 @@
+#pragma once
+
+#include <list>
+#include <string>
+#include <vector>
+#include "logging/Logger.hpp"
+#include "mapping/BarycentricBaseMapping.hpp"
+#include "mapping/Polation.hpp"
```
fsimonis, line 8: Most of these can be removed.

src/mapping/LinearCellInterpolationMapping.cpp
```diff
@@ -0,0 +1,116 @@
+#include "LinearCellInterpolationMapping.hpp"
+
+#include <Eigen/Core>
+#include <algorithm>
+#include <deque>
+#include <memory>
+#include <ostream>
+#include <unordered_set>
+#include <utility>
```
fsimonis, line 9: Most of these can be removed.

src/query/Index.cpp
fsimonis: We are now falling back to nearest-projection by design. So this name is very misleading.

It could also be `findCellOrProjection`, but the name has to change.

src/query/Index.cpp
```diff
@@ -211,6 +211,25 @@ ProjectionMatch Index::findNearestProjection(const Eigen::VectorXd &location, in
...
+      auto polation = mapping::Polation(location, _mesh->triangles()[match.index]);
+      if (polation.isInterpolation()) {
+        return {polation, 0.0};
+      }
+    }
+
+    // If no triangle is found, fall-back on NP
+    return findNearestProjection(location, n);
+  } else {
+    PRECICE_ASSERT(_mesh->getDimensions() == 3, "Volume coupling 3D not  implemented");
```
fsimonis, line 228: This should really be `PRECICE_UNREACHABLE("Volume coupling 3D not implemented")`

tests/parallel/mapping-volume/testParallelSquare1To2.cpp
```diff
@@ -0,0 +1,109 @@
...
+BOOST_AUTO_TEST_SUITE(MappingVolume)
+BOOST_AUTO_TEST_CASE(testParallelSquare1To2)
+{
+  using precice::VertexID;
+  using precice::testing::equals;
+
+  PRECICE_TEST("SolverOne"_on(1_rank), "SolverTwo"_on(2_ranks));
+
+  // Implement your test here.
+  BOOST_TEST(true);
```
fsimonis, line 19: That's what you did :wink:

tests/parallel/mapping-volume/testParallelSquare2To1.cpp
```diff
@@ -0,0 +1,110 @@
...
+
+#include "testing/Testing.hpp"
+
+#include <precice/SolverInterface.hpp>
+#include <vector>
+
+BOOST_AUTO_TEST_SUITE(Integration)
+BOOST_AUTO_TEST_SUITE(Parallel)
+BOOST_AUTO_TEST_SUITE(MappingVolume)
+BOOST_AUTO_TEST_CASE(testParallelSquare2To1)
```
fsimonis, line 11: Same, no test-prefix please.

tests/parallel/mapping-volume/testParallelSquareConservative1To2.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+#include "testing/Testing.hpp"
+
+#include <precice/SolverInterface.hpp>
+#include <vector>
+
+BOOST_AUTO_TEST_SUITE(Integration)
+BOOST_AUTO_TEST_SUITE(Parallel)
+BOOST_AUTO_TEST_SUITE(MappingVolume)
+BOOST_AUTO_TEST_CASE(testParallelSquareConservative1To2)
```
fsimonis, line 11: And one more time

tests/parallel/mapping-volume/testParallelTriangleConservative2To1.cpp
```diff
@@ -0,0 +1,85 @@
...
+
+#include "testing/Testing.hpp"
+
+#include <precice/SolverInterface.hpp>
+#include <vector>
+
+BOOST_AUTO_TEST_SUITE(Integration)
+BOOST_AUTO_TEST_SUITE(Parallel)
+BOOST_AUTO_TEST_SUITE(MappingVolume)
+BOOST_AUTO_TEST_CASE(testParallelTriangleConservative2To1)
```
fsimonis, line 11: You get the idea

src/mapping/LinearCellInterpolationMapping.cpp
boris-martin: Are you sure ? An "almost zero" means there has been some projection, as findCelInterpolation returns an exact 0 if a triangle was used.

fsimonis: Not necessarily. The tolerance check covers rounding errors etc.

tests/parallel/mapping-volume/ParallelSquare2To1.cpp
```diff
@@ -0,0 +1,110 @@
...
+BOOST_AUTO_TEST_SUITE(MappingVolume)
+BOOST_AUTO_TEST_CASE(ParallelSquare2To1)
+{
+  PRECICE_TEST("SolverOne"_on(2_ranks), "SolverTwo"_on(1_rank));
+
+  using precice::VertexID;
+  using precice::testing::equals;
+
+  // Implement your test here.
+  BOOST_TEST(true);
```
uekerman, line 19:
```suggestion
```

src/mapping/LinearCellInterpolationMapping.cpp
```diff
@@ -0,0 +1,100 @@
...
+    }
+  }
+
+  if (!fallbackStatistics.empty()) {
+    PRECICE_INFO("Some points are outisde of the domain defined by connectivity. Fall-back on Nearest-Projection occured."
+                 " Projection distances: {}",
+                 fallbackStatistics);
+  } else {
+    PRECICE_INFO("All vertices are inside cells, no fallback to Nearest-Projection required");
+  }
```
uekerman, line 94: We need to improve the wording here still a bit. Not sure if a user would understand what is happening already.

uekerman (changes requested):
Looks very clean already :+1:
We still need more unit tests, also for the conservative variant. Probably sth you have on the radar anyway; I just mention it for completeness.

boris-martin:
@uekerman all done except the unresolved conversation above.

src/mapping/LinearCellInterpolationMapping.cpp
boris-martin: # Comment PR

Actually, an interesting thing occurs when we have no triangles but enough edges to always find a projection:

```
---[precice]  Compute "read" mapping from mesh "Flow-Mesh" to mesh "Chemistry-Mesh".
---[precice] WARNING:  2D Mesh "Flow-Mesh" does not contain triangles. Linear cell interpolation falls back to nearest projection mapping.
---[precice]  All vertices are inside cells, no fallback to Nearest-Projection required
```

Wording should be more consistent as having just edges trigger contradictory messages ("No triangles ! => then right afterwards "All projections were fine").

That's probably an edge case (no pun intended), as I got this with matching meshes and only edges configured, so projection was trivial. Maybe we should simply don't log anything if no fallback occurs? (But keep the `triangles.empty()` warning)

But back to a more standard case, with actual triangles but some projection, typical log is 

```
---[precice]  Compute "read" mapping from mesh "Mesh1" to mesh "Mesh2".
---[precice]  Some points are outside of the domain defined by connectivity. Fall-back on Nearest-Projection occured. Projection distances: min:0.1 max:0.1 avg: 0.1 var: 0 cnt: 2
```

So, in context, it is clear that it is part of the mapping computation, but not that this is Linear Cell Interpolation specific behavior.
Alternative text suggested could be something like

```
---[precice]  Compute "read" mapping from mesh "Mesh1" to mesh "Mesh2".
---[precice]  Linear Cell Interpolation is used, but some points from "Mesh2" don't lie in the domain defined by the triangles/tetrahedra. These points have been projected on the domain boundary. This could come from rounding errors or wrong connectivity configuration. If distances seem too large, please check your mesh. The projection statistics are: min:0.1 max:0.1 avg: 0.1 var: 0 cnt: 2
```

(where triangles/tetrahedra could be dynamically picked at runtime to show the corresponding on, and "Mesh2" replaced by target mesh name)

Also, should we keep it as PRECICE_INFO or make it a warning ?

Issue #1288, opened 2022-05-12
precice 2.4.0 mingw


3rav:
I have an error when trying to compile version 2.4.0 (while for version 2.3.0 it is OK):

**Info 2.4.0:**
```
-- Found Boost: C:/msys64/mingw64/include (found suitable version "1.79.0", minimum required is "1.71.0") found components: filesystem log log_setup program_options system thread unit_test_framework regex chrono atomic
-- Found Eigen3: C:/msys64/mingw64/include/eigen3 (Required is at least version "3.2")
-- Validating Eigen
-- Validating Eigen - success
-- Found LibXml2: C:/msys64/mingw64/lib/libxml2.dll.a (found version "2.9.14")
-- Validating LibXml2
-- Validating LibXml2 - success
-- Validating JSON
-- Validating JSON - success
-- Validating fmtlib
-- Validating fmtlib - success
-- Found MPI_CXX: C:/msys64/mingw64/lib/libmsmpi.dll.a (found version "2.0")
-- Found MPI: TRUE (found version "2.0")
-- MPI Version: Microsoft MPI 10.1.12498.18
-- Found PkgConfig: C:/msys64/mingw64/bin/pkg-config.exe (found version "1.8.0")
-- Detecting additional PETSc prefixes
-- Detected C:/msys64/mingw64/src/petsc-3.17.0/dmo
-- Skipped the following invalid prefixes: C:/msys64/mingw64/src/petsc-3.17.0
-- pkg-config will search the following paths:
PKG_CONFIG_PATH
   C
   \msys64\mingw64\lib\pkgconfig
   C
   \msys64\mingw64\share\pkgconfig
CMAKE_PREFIX_PATH
   C:/msys64/mingw64/src/petsc-3.17.0/dmo
-- Checking for module 'PETSc>=3.12'
--   Found PETSc, version 3.17.0
-- Found PETSc: TRUE (found suitable version "3.17.0", minimum required is "3.12")
CMake Error at C:/msys64/mingw64/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:230 (message):
  Could NOT find Python3 (missing: Python3_NumPy_INCLUDE_DIRS NumPy) (found
  version "3.8.8")
Call Stack (most recent call first):
  C:/msys64/mingw64/share/cmake/Modules/FindPackageHandleStandardArgs.cmake:594 (_FPHSA_FAILURE_MESSAGE)
[...]

PR #1298, opened 2022-05-18
Rename `uint` to `unsigned int` in RBF test

DavidSCN:
## Main changes of this PR

The data type `uint` is not universal and we need to substitute it with `unsigned int` 

Related to #1288 

## Author's checklist

## Reviewers' checklist


uekerman:
How could we catch this in our CI?

PR #1301, opened 2022-05-23
Refactor mapping into consistent and conservative mapping functions

DavidSCN:
## Main changes of this PR

Splits the `map` function into `mapConsistent` and `mapConservative` and implements the `hasComputedMapping` in the base class (instead of each derived class).

The mapping functions follow the same structure namely

```cpp
PRECICE_ASSERT(...)
if (conservative)
{...}
else
{...}
if(scaledConsistent)
 applyScaledConsistent
```
The map function itself becomes messy quite fast. The new interface defines the structure in the base class and requires the individual branches to be implemented in the derived classes. We do this already in the RBF mapping classes. Open for discussions.

## Author's checklist

## Reviewers' checklist


fsimonis (approved):
Thanks for tidying this up!
Looks all sane.

src/mapping/Mapping.cpp
```diff
@@ -87,6 +87,31 @@ bool Mapping::requireGradient() const
...
+    mapConservative(inputDataID, outputDataID);
+  } else {
+    mapConsistent(inputDataID, outputDataID);
+  }
+  if (hasConstraint(Mapping::SCALEDCONSISTENT)) {
+    scaleConsistentMapping(inputDataID, outputDataID);
+  }
+}
+
 void Mapping::scaleConsistentMapping(int inputDataID, int outputDataID) const
```
boris-martin, line 115: Slightly tangential to this PR, but couldn't this function be refactored to split the validity check and the actual scaling ?

src/mapping/Mapping.cpp
```diff
@@ -87,6 +87,31 @@ bool Mapping::requireGradient() const
...
+  PRECICE_ASSERT(output()->data(outputDataID)->values().size() / output()->data(outputDataID)->getDimensions() == static_cast<int>(output()->vertices().size()),
+                 output()->data(outputDataID)->values().size(), output()->data(outputDataID)->getDimensions(), output()->vertices().size());
+
+  if (hasConstraint(CONSERVATIVE)) {
+    mapConservative(inputDataID, outputDataID);
+  } else {
+    mapConsistent(inputDataID, outputDataID);
+  }
+  if (hasConstraint(Mapping::SCALEDCONSISTENT)) {
+    scaleConsistentMapping(inputDataID, outputDataID);
```
boris-martin, line 111: I'm not a big fan of this structure, with an if coming after an else. Wouldn't it be cleaner to have something like 

```cpp
if (hasConstraint(CONSERVATIVE)) {
mapConservative(...)
} else if (hasConstraint(CONSISTENT) {
mapConsisten(...);
} else if {hasConstraint(SCALEDCONSISTENT) {
mapConsistent(...);
scaleConsistentMapping(...);
}
```

Also, we have `CONSERVATIVE` but `Mapping::SCALEDCONSISTENT`. Prefix isn't consistent.

boris-martin (approved):
Great refactoring. I added some comments on what could be future improvements, but the current changes seem very good.

src/mapping/Mapping.cpp
DavidSCN: You mean the `scaledConsistentMapping` function? Yes, sure. I won't touch this part in this PR, though.

src/mapping/Mapping.cpp
DavidSCN: Yes, I adapted it to the structure you proposed here.

DavidSCN: ... and fixed the inconsistency.

Issue #1196, opened 2022-03-04
Make calling initializeData mandatory?


BenjaminRodenberg:
**Please describe the problem you are trying to solve.**

Simplifying the implementation of coupling schemes and have a stricter control flow with less optional branches.

**Describe the solution you propose.**

I suggest to make the call to `initializeData()` mandatory for every preCICE program. This is clearly breaking and will affect the users. I see the following advantages:

* Data exchange of initial data becomes more explicit for the user. If no data is written before calling `initializeData()`, initial data will be zero. There is an exception for the second participant of serial coupling schemes, because they receive data from the first iteration of the first iteration.
* We know that we don't have to exchange data in `initialize()`. Currently, we don't know whether `initializeData()` will be called after `initialize()`. Therefore the following statements necessary: 
    * receive data in `initialize` of serial coupling scheme, if `initializeData` is not called: [see here](https://github.com/precice/precice/blob/703fbe46165676bc16a4e7cfbad077b1ff8609b4/src/cplscheme/SerialCouplingScheme.cpp#L66-L73)
    * receive data in `initializeData` of serial coupling scheme, if `initializeData` is called: [see here](https://github.com/precice/precice/blob/703fbe46165676bc16a4e7cfbad077b1ff8609b4/src/cplscheme/SerialCouplingScheme.cpp#L90-L93)

**Describe alternatives you've considered**

Leave it as it is. This would be:

* non-breaking
* simpler for users that don't want to initialize data

**Additional context**

* I stumbled across this point when working on https://github.com/precice/precice/projects/7. Here, everything becomes even more complicated when you don't know at the point in time when `initialize()` is called what will happen next.
* If we conclude on this issue, we could require calling `initializeData`, if `experimental=true`. This would simplify the implementation of waveform relaxation and make code of our users future-proof.

BenjaminRodenberg:
This would also simplify the following lines of code:

https://github.com/precice/precice/blob/703fbe46165676bc16a4e7cfbad077b1ff8609b4/src/precice/impl/SolverInterfaceImpl.hpp#L532-L540

uekerman:
Sth to think about: when we make it mandatory, couldn't we fully integrate the functionality into `initialize`?

fsimonis:
If we merge `initialize` and `initializeData`, then we have two options:
1. figure out the need to write initial data in the SolverInterface Ctor and set the action as required. This should be doable.
2. always assume that the user sets the mesh and initializes the data in one go, which is very natural.  
[...]

PR #1307, opened 2022-05-27
Fix first participant method in compositional cplscheme

uekerman:
## Main changes of this PR

I tried to reproduce the problem a user ran into when 

* coupling three solvers A, B, C 
* with two serial-explicit coupling schemes A<-->B and B<-->C
* and prescribing the time window size for B and C by A.

Reported here: https://precice.discourse.group/t/variable-time-steps-without-sub-cycling-in-serial-explicit-coupling/1049/5

I indeed ran into a problem when `CompositionalCouplingScheme` tried to merge different time stepping methods and fixed it.

Afterwards, I ran into another problem, however. "first-participant" has not great integration test coverage yet :/
I reproduced the problem with the integration test `ReadWriteScalarDataFirstParticipant`.

The following is problematic for the first participant of a serial-explicit coupling scheme with dt by "first-participant". Then, the coupling scheme does not define a time window size.

https://github.com/precice/precice/blob/780d4f12aa33aa2e6952debd9031b3d88c59a519/src/precice/impl/SolverInterfaceImpl.cpp#L1699

We need a similar treatment as here:

https://github.com/precice/precice/blob/780d4f12aa33aa2e6952debd9031b3d88c59a519/src/precice/impl/SolverInterfaceImpl.cpp#L431-L435

and then probably store the time window site at `SolverInterfaceImpl`.

@BenjaminRodenberg Could I hand over to you here?

## Author's checklist

* [ ] I checked that this actually solves the user's problem.

## Reviewers' checklist


src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -280,9 +280,15 @@ bool BaseCouplingScheme::hasTimeWindowSize() const
   return not math::equals(_timeWindowSize, UNDEFINED_TIME_WINDOW_SIZE);
 }
 
+bool BaseCouplingScheme::solverSetsTimeWindowSize() const
+{
+  PRECICE_ASSERT(hasTimeWindowSize());
+  return false;
+}
```
uekerman, line 287: Isn't this function a copy of `hastTimeWindowSize`?
I agree that the name `solverSetsTimeWindowSize()` is much clearer.

src/cplscheme/CompositionalCouplingScheme.cpp
```diff
@@ -201,12 +201,24 @@ bool CompositionalCouplingScheme::hasTimeWindowSize() const
   return hasIt;
 }
 
+bool CompositionalCouplingScheme::solverSetsTimeWindowSize() const
+{
+  PRECICE_TRACE();
+  bool setsIt = false;
+  for (const Scheme &scheme : _couplingSchemes) {
+    setsIt &= scheme.scheme->solverSetsTimeWindowSize();
```
uekerman, line 209:
```suggestion
    setsIt |= scheme.scheme->solverSetsTimeWindowSize();
```
?

src/cplscheme/SerialCouplingScheme.cpp
```diff
@@ -122,10 +134,8 @@ bool SerialCouplingScheme::exchangeDataAndAccelerate()
     sendData(getM2N(), getSendData());
     // the second participant does not want new data in the last iteration of the last time window
     if (isCouplingOngoing() || (isImplicitCouplingScheme() && not convergence)) {
-      if (_participantReceivesTimeWindowSize) {
-        receiveAndSetTimeWindowSize();
-      }
       PRECICE_DEBUG("Receiving data...");
+      receiveAndSetTimeWindowSize();
```
uekerman, line 138: What happened to `_participantReceivesTimeWindowSize`?
Also, the location of the debug statement seems wrong now.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: > Isn't this function a copy of `hastTimeWindowSize`?

From the perspective of functionality: Yes. But I think it clearly improves readability to have two functions here. It's from the outside perspective not obvious that `hasTimeWindowSize() == !solverSetsTimeWindowSize()`.

If we remove `hasTimeWindowSize`, we would end up in some situations with something like this:

```
double BaseCouplingScheme::getThisTimeWindowRemainder() const
{
  PRECICE_TRACE();
  double remainder = 0.0;
  if (!solverSetsTimeWindowSize()) {
    remainder = getNextTimestepMaxLength();
  }
  PRECICE_DEBUG("return {}", remainder);
  return remainder;
}
```

instead of 

```
double BaseCouplingScheme::getThisTimeWindowRemainder() const
{
  PRECICE_TRACE();
  double remainder = 0.0;
  if (hasTimeWindowSize()) {
    remainder = getNextTimestepMaxLength();
  }
  PRECICE_DEBUG("return {}", remainder);
  return remainder;
}
```

So I would like to keep both functions for the sake of readability.

src/cplscheme/SerialCouplingScheme.cpp
BenjaminRodenberg: I moved it inside of `receiveAndSetTimeWindowSize()` to make it consistent with `sendTimeWindowSize()`. Moving the statement inside of the function saves some checks in the functions using `receiveAndSetTimeWindowSize`.

I'll move the debug statement.

src/cplscheme/BaseCouplingScheme.cpp
uekerman: Mmh, I actually find the first version easier to read. It gives additional information on when what happens.
In then end, `hasTimeWindowSize` is hard to understand, what does "has a time window size" mean?

src/cplscheme/CompositionalCouplingScheme.cpp
uekerman: Yes, makes sense now as it's just the negative of the old `hasTimeWindowSize`.

BenjaminRodenberg: Actually... then it should be `&=` for `solverSetsTimeWindowSize`. That's simply [De Morgan's law](https://en.wikipedia.org/wiki/De_Morgan%27s_laws). I just reverted the latest commit via [e0f9306](https://github.com/precice/precice/pull/1307/commits/e0f9306823a94be955e922e42b0849d058c0d25a).

tests/serial/three-solvers/ThreeSolversFirstParticipant.xml
```diff
@@ -0,0 +1,69 @@
...
+      <time-window-size method="first-participant" />
+      <exchange data="DataA" mesh="MeshOne" from="SolverOne" to="SolverTwo" />
+    </coupling-scheme:serial-explicit>
+
+    <coupling-scheme:serial-explicit>
+      <participants first="SolverTwo" second="SolverThree" />
+      <max-time value="6" />
+      <time-window-size method="first-participant" />
+      <exchange data="DataB" mesh="MeshTwoB" from="SolverTwo" to="SolverThree" />
+    </coupling-scheme:serial-explicit>
```
BenjaminRodenberg, line 67: I just stumbled here: I think if this case is allowed for a compositional coupling scheme, then the assumption that `hasTimeWindowSize()` and `solverSetsTimeWindowSize()` are inverse is wrong. Both cases can coexist, because SolverTwo has a time window size (received from SolverOne) and at the same time it needs to set the time window size (for SolverThree).

This also has implications for the compositional coupling scheme: I'm not sure that we can simply iterate over all coupling schemes and apply a simple operation like `any_of` or `all_of`. Providing a public method (i.e. reachable from outside the coupling scheme) that is named `solverSetsTimeWindowSize()´ does not make sense for the compositional coupling scheme, because only individual `BaseCouplingSchemes` can set the time window size.

src/cplscheme/CompositionalCouplingScheme.cpp
BenjaminRodenberg: To conclude this discussion: The both functions are not inverse of each other (otherwise solver A sending timestep size to B and B forwarding to C would not work at all). This counters the De Morgan argument. `|=` is the correct implementation, because a compositional coupling scheme might have only a single scheme that sends the timestep size. However: I decided to entirely remove this function (see below). So this explanation is mainly for historic reasons.

BenjaminRodenberg: To conclude this discussion: The both functions are not inverse of each other (otherwise solver A sending timestep size to B and B forwarding to C would not work at all). This counters the De Morgan argument. `|=` is the correct implementation, because a compositional coupling scheme might have only a single scheme that sends the timestep size. However: I decided to entirely remove this function (see below). So this explanation is mainly for historic reasons.

uekerman (commented):
Looks good to me

tests/serial/time/explicit/serial-coupling/ReadWriteScalarDataFirstParticipantInitData.cpp
```diff
@@ -0,0 +1,80 @@
...
+#include "testing/Testing.hpp"
+
+#include <precice/SolverInterface.hpp>
+#include <vector>
+
+using namespace precice;
+
+BOOST_AUTO_TEST_SUITE(Integration)
+BOOST_AUTO_TEST_SUITE(Serial)
+BOOST_AUTO_TEST_SUITE(Time)
```
BenjaminRodenberg, line 12: I don't think that this test (maybe also the other new test) fits very well into `Integration/Serial/Time`. Maybe `Integration/Serial/InitializeData` ? Or should we create a dedicated test suite for the `participant-first` configuration?

BenjaminRodenberg:
> We have another bug in preCICE, I just added an integration test to reproduce.
> 
> https://github.com/precice/precice/blob/cd1b3a4921a557f7a3ab655f8cb472b9f26b5148/src/cplscheme/SerialCouplingScheme.cpp#L60-L74
> 
> If `dt-method = "first participant"`: The second participant always needs to receive the time window size in `initialize` already. This cannot wait till `initializeData` as `initialize` needs to return this value to the user already.
> 
> @BenjaminRodenberg handing back to you grin

I tried to fix this, but we might have a deadlock here:

The first participant sends the timestep size at the end of advance (it does not know about the timestep size provided by the user before the user defines it via advance). But to reach advance, the first participant must call initializeData and receive initial data from the second participant here. Here comes the deadlock: The second participant does not reach initializeData to send initial data, because it is waiting for the timestep size in initialize.

#1196 might help us to resolve this deadlock: 

1) if we merge `initialize` and `initializeData`, we have more possibilities to move around communication.
2) if `initializeData` is mandatory `initializeData` could be used to return `dt`, not `initialize`.

But with our current API design and order of API calls I do not really see a solution. To me it looks like `initializeData` + `first participant` has to be forbidden.

uekerman:
You're right :see_no_evil:

BenjaminRodenberg:
I just pushed [62e3417](https://github.com/precice/precice/pull/1307/commits/62e34179ec0d852f5eba8249724dd6bb4d95b5db). This is from my current point of view as close as we can get to a solution. The deadlock-problem still exists, therefore I had to modify the test.

Depending on how we resolve the deadlock-situation, can can then keep [4431577](https://github.com/precice/precice/pull/1307/commits/4431577e558dae6fe7174ff89652f7043741dbad) and [62e3417](https://github.com/precice/precice/pull/1307/commits/62e34179ec0d852f5eba8249724dd6bb4d95b5db) here or `cherry-pick` it into a feature branch.

tests/serial/time/explicit/serial-coupling/ReadWriteScalarDataFirstParticipantInitData.cpp
uekerman: In a way, that's an inherent problem / feature of integration tests; they always test multiple components simultaneously. Here it is "first-participant" + "data initialization". For me, "first-participant" is the dominating feature here. That's why I added it to `time`. Is somehow fits nicely as it only applies to serial coupling.

PR #1309, opened 2022-06-01
added barycentric coordinates for tetrahedron

boris-martin:
## Main changes of this PR

Added a function for computing the barycentric coordinates of a point with respect to a tetrahedron (with 4 vertices as input), as well as unit tests to check its correctness. Split of #1240 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/math/barycenter.cpp
+++ b/src/math/barycenter.cpp
@@ -100,6 +100,52 @@ Eigen::Vector3d calcBarycentricCoordsForTriangle(
   return barycentricCoords;
 }
 
+Eigen::Vector4d calcBarycentricCoordsForTetrahedron(
+    const Eigen::VectorXd &a,
+    const Eigen::VectorXd &b,
+    const Eigen::VectorXd &c,
+    const Eigen::VectorXd &d,
+    const Eigen::VectorXd &u)
+{
+  using Eigen::Vector3d;
+  using Eigen::Vector4d;
+
+  const int dimensions = a.size();
+
+  PRECICE_ASSERT(dimensions == 3, dimensions);
+  PRECICE_ASSERT(dimensions == b.size(), "A and B need to have the same dimensions.", dimensions, b.size());
+  PRECICE_ASSERT(dimensions == c.size(), "A and C need to have the same dimensions.", dimensions, c.size());
+  PRECICE_ASSERT(dimensions == d.size(), "A and D need to have the same dimensions.", dimensions, d.size());
+  PRECICE_ASSERT(dimensions == u.size(), "A and the point need to have the same dimensions.", dimensions, u.size());
+
+  Vector4d barycentricCoords;
+
+  // Varying per poit
+  Vector3d au = u - a;
+  Vector3d du = u - d;
+  Vector3d cu = u - c;
+
+  // Necessary to compute triangles
+  Vector3d ab = b - a;
+  Vector3d ac = c - a;
+  Vector3d ad = d - a;
+  Vector3d bc = c - b;
+
+  // Triangles
+  Vector3d abc = ab.cross(bc);
+  Vector3d abd = ab.cross(-ad);
+  Vector3d acd = ac.cross(ad);
+
+  auto volume = abc.dot(ad);
+
+  barycentricCoords(3) = abc.dot(au) / volume;
+  barycentricCoords(2) = abd.dot(du) / volume;
+  barycentricCoords(1) = acd.dot(cu) / volume;
+  barycentricCoords(0) = 1 - barycentricCoords(3) - barycentricCoords(2) - barycentricCoords(1);
+
+  return barycentricCoords;
+}
+
 } // namespace barycenter
 } // namespace math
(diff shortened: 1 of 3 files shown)


src/math/barycenter.cpp
```diff
@@ -100,6 +100,52 @@ Eigen::Vector3d calcBarycentricCoordsForTriangle(
...
+
+  PRECICE_ASSERT(dimensions == 3, dimensions);
+  PRECICE_ASSERT(dimensions == b.size(), "A and B need to have the same dimensions.", dimensions, b.size());
+  PRECICE_ASSERT(dimensions == c.size(), "A and C need to have the same dimensions.", dimensions, c.size());
+  PRECICE_ASSERT(dimensions == d.size(), "A and D need to have the same dimensions.", dimensions, d.size());
+  PRECICE_ASSERT(dimensions == u.size(), "A and the point need to have the same dimensions.", dimensions, u.size());
+
+  Vector4d barycentricCoords;
+
+  // Varying per poit
```
fsimonis, line 123: Typo?

src/math/tests/BarycenterTest.cpp
```diff
@@ -237,6 +237,74 @@ BOOST_AUTO_TEST_CASE(BarycenterTriangle2D)
   }
 }
 
+BOOST_AUTO_TEST_CASE(BarycenterTetrahedron)
```
fsimonis, line 240: Please break this big test into smaller individual tests.

Create a test suite, called `BarycenterTetrahedron` and individual tests for these cases.

You can define common coordinates in [a fixture](https://www.boost.org/doc/libs/1_71_0/libs/test/doc/html/boost_test/tests_organization/fixtures/case.html#boost_test.tests_organization.fixtures.case.example_descr)  and reuse it in the tests.

fsimonis (changes requested):
We need some more tests for funny tetrahedra such as

* abc with d projecting outside the triangle abc
* abc with d above the plane abc
* abc with d below the plane abc
* Very slender tet, which barely passes the "is not degenerate" test.

We also need to test extrapolations, as they are used to differentiate inter- and extrapolation.

src/math/tests/BarycenterTest.cpp
boris-martin: By new test suite you mean a new file? Because the pattern here seems to be "one test file per source file" and that would break it.

fsimonis: You can group your tests in a test suite.

```cpp
BOOST_AUTO_TEST_SUITE(Tetrahedron)

// put your tests here

BOOST_AUTO_TEST_SUITE_END()
```

src/math/tests/BarycenterTest.cpp
```diff
@@ -237,6 +238,183 @@ BOOST_AUTO_TEST_CASE(BarycenterTriangle2D)
...
+  Eigen::Vector3d b{0.0, 1.0, 0.0};
+  Eigen::Vector3d c{1.0, 0.0, 0.0};
+  Eigen::Vector3d d{11.0, 11.0, 1.0};
+
+  Eigen::Vector3d center{3.0, 3.0, 0.25};
+};
+
+typedef boost::mpl::vector<TetrahedronFixture, FlippedTetrahedronFixture, AlmostDegenerateTetrahedronFixture, FunnyTetrahedronFixture> TetrahedraFixtures;
+
+BOOST_FIXTURE_TEST_CASE_TEMPLATE(BarycenterTetrahedronVertices, T, TetrahedraFixtures, T)
```
fsimonis, line 282: Why didn't you just use a `BOOST_FIXTURE_TEST_CASE` with your fixture?

src/math/tests/BarycenterTest.cpp
```diff
@@ -237,6 +238,183 @@ BOOST_AUTO_TEST_CASE(BarycenterTriangle2D)
...
+  using Eigen::Vector4d;
+  using precice::testing::equals;
+
+  // Alias for fixture data
+  const auto &a = T::a;
+  const auto &b = T::b;
+  const auto &c = T::c;
+  const auto &d = T::d;
+
+  // Is A?
```
fsimonis, line 295: These should all be separate test cases. They can be as short as 5 lines.

src/math/tests/BarycenterTest.cpp
```diff
@@ -237,6 +238,183 @@ BOOST_AUTO_TEST_CASE(BarycenterTriangle2D)
...
+  PRECICE_TEST(1_rank);
+  using Eigen::Vector3d;
+  using Eigen::Vector4d;
+  using precice::testing::equals;
+
+  // Alias for fixture data
+  const auto &a = T::a;
+  const auto &b = T::b;
+  const auto &c = T::c;
+  const auto &d = T::d;
```
fsimonis, line 293: The point of the test fixture is that you can directly use members of the fixture inside the test. So this is obsolete.

fsimonis (approved):

PR #1310, opened 2022-06-01
Basic Tetrahedron class

boris-martin:
## Main changes of this PR

Added basic components for tetrahedra in meshes: Tetrahedron class which holds vertices, can compute enclosing radius, center and volume. Split of #1240 
Added unit tests for these

## Author's checklist

## Reviewers' checklist


src/mesh/Tetrahedron.cpp
```diff
@@ -0,0 +1,96 @@
...
+
+bool Tetrahedron::operator!=(const Tetrahedron &other) const
+{
+  return !(*this == other);
+}
+
+std::ostream &operator<<(std::ostream &os, const Tetrahedron &t)
+{
+  using utils::eigenio::wkt;
+  return os << "POLYGON (("
```
fsimonis, line 87: There is no way to represent volumes in WKT. I think our best option is to draw the outline using a `MultiLineString` of the 6 edges.

src/mesh/Tetrahedron.hpp
```diff
@@ -0,0 +1,169 @@
...
+  /// Type of the read-only const random-access iterator over Vertex coords
+  /**
+   * This index-based iterator iterates over the vertices of this Tetrahedron.
+   * The returned value is the forwarded result of Vertex::getCoords.
+   * It is thus a read-only random-access iterator.
+   */
+  using const_iterator = IndexRangeIterator<const Tetrahedron, const Vertex::RawCoords>;
+
+  /// Type of the read-only random access vertex iterator
+  using iterator = const_iterator;
```
fsimonis, line 38: We won't need this for Boost.geometry if we use AABBs.

src/mesh/Tetrahedron.hpp
```diff
@@ -0,0 +1,169 @@
...
+  /// Returns a read-only random access iterator to the end (3) of the vertex range [0,1,2]
+  const_iterator end() const;
+
+  /// Returns a read-only random-access iterator to the begin (0) of the vertex range [0,1,2]
+  const_iterator cbegin() const;
+
+  /// Returns a read-only random access iterator to the end (3) of the vertex range [0,1,2]
+  const_iterator cend() const;
+
+  ///@}
```
fsimonis, line 85: Not needed

src/mesh/Tetrahedron.hpp
```diff
@@ -0,0 +1,169 @@
...
+
+inline Tetrahedron::const_iterator Tetrahedron::cbegin() const
+{
+  return begin();
+}
+
+inline Tetrahedron::const_iterator Tetrahedron::cend() const
+{
+  return end();
+}
```
fsimonis, line 159: Not needed

fsimonis (changes requested):
We won't need any iterator access if we use AABBs for the indexing.

src/mesh/Tetrahedron.hpp
```diff
@@ -0,0 +1,105 @@
...
+// ----------------------------------------------------------- CLASS DEFINITION
+
+namespace precice {
+namespace mesh {
+
+/// Tetrahedron of a mesh, defined by 4 vertices
+class Tetrahedron {
+public:
+  /// Fix for the Boost.Test versions 1.65.1 - 1.67
+  using value_type = Vertex::RawCoords;
```
fsimonis, line 30: Is this still required after we removed the iterator interface?

fsimonis (approved):
Last comment, then please squash and merge if all tests pass.

Issue #1153, opened 2021-12-16
Define primitives in terms of vertices


fsimonis:
**Please describe the problem you are trying to solve.**
Currently, triangles are defined in terms of edges an edges are defined in terms of vertices.
This makes defining triangles in adapter codes and integration tests tedious:
```
meshID = getMeshID();
setMeshVertices(...);
for e in edges:
  setMeshEdge(...);
for t in trianges
  setMeshTriangle(...);
```
This will become extremely tedious once we support tetrahedral elements.
For consistency, they would need to be defined using triangles.

As the `mesh::Triangle` is implemented this way as well, we need to still create `mesh::Edge`s for each triangle.
Meaning `setMeshTrianglesWithEdges`, needs to check existing edges in order to avoid duplications, which is the reason why the API call is so slow.
```
meshID = getMeshID();
setMeshVertices(...);
for t in trianges
  setMeshTriangleWithEdges(...);
```

**Describe the solution you propose.**

Implement the following approach for connectivity:
* Elements are defined by vertices only.
* Sub-elements such as the edges of a triangle are implicitly defined via the triangle.
* Update the API implementation to still allow constructing triangles from edgeIDs.

Example Triangle Strip of 6 vertices and 4 triangles:

Scenario | #Triangles | #Edges
--- | --- | ---
All edges |  4 | 12
Unique edges | 4 | 9
Implicit edges | 4 | 0

## Pros:
* Easier to use API with faster calls for setting connectivity information.
* No handling of edge IDs in adapters.
* Faster internals as we don't need to handle edges
* Lower memory consumption as meshes won't contain implicit primitives.
  This will be very noticeable once we implement voluminous primitives such as tetrahedra.
* Lower memory consumption as we don't need special maps to access unique vertices of triangles via the edges any more.
* Lower initialization times as the meshes will be smaller, so there will be less to communicate.
* This would make all tests that require meshes with connectivity as well as integration tests way cleaner.

## Cons:
* The repartitioning will filter out complete primitives including all implicit ones. We could implement an algorithm that generates leftover edges, but this will need some thought.
* This will make primitive ownership #963 more difficult to implement as primitives may not exist in every participant (1Triangle vs 1 Edge due to filtered vertex). This will impact scaled-consistent mappings as well as the watch-integral for parallel participants.
[...]

PR #1311, opened 2022-06-02
Vertex-only primitives

fsimonis:
## Main changes of this PR

This PR simplifies triangles, by defining them in terms of vertices instead of edges.
This reduces their size in memory and avoids having to remap edge IDs in filter and communication operations.
It also simplifies handling triangles in tests.
The change is internal only and has no observable side effects in the API.

Major change is that triangles can now exist independent of edges.

In short:
* No breaking change
* lower memory footprint
* less communication while sending meshes
* faster filtering and sending of meshes
* A triangle ABC **doesn't** imply the existence of edges AB, BC, and AC any longer.

See #1153

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/com/CommunicateMesh.cpp
+++ b/src/com/CommunicateMesh.cpp
@@ -37,7 +37,12 @@ void CommunicateMesh::sendMesh(
   const auto &meshVertices     = mesh.vertices();
   const int   numberOfVertices = meshVertices.size();
   _communication->send(numberOfVertices, rankReceiver);
-  if (not mesh.vertices().empty()) {
+
+  if (mesh.vertices().empty()) {
+    return;
+  }
+
+  {
     std::vector<double> coords(static_cast<size_t>(numberOfVertices) * dim);
     std::vector<int>    globalIDs(numberOfVertices);
     for (size_t i = 0; i < static_cast<size_t>(numberOfVertices); ++i) {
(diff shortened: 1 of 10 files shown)


src/mesh/Triangle.cpp
```diff
@@ -24,54 +26,50 @@ Triangle::Triangle(
...
-    if (&edge(2).vertex(0) == &edge(1).vertex(0)) {
-      _vertexMap[2] = false;
-    } else {
-      PRECICE_ASSERT(&edge(2).vertex(1) == &edge(1).vertex(0));
-      _vertexMap[2] = true;
-    }
-  }
+  _vertices = {&v0, &v1, v2};
+  std::sort(_vertices.begin(), _vertices.end(),
+            [](const Vertex *lhs, const Vertex *rhs) { return *lhs < *rhs; });
```
uekerman, line 56: Why do we need this sorting?

src/com/CommunicateMesh.cpp
```diff
@@ -48,17 +53,25 @@ void CommunicateMesh::sendMesh(
...
   }
 
+  _communication->send(mesh.hasConnectivity(), rankReceiver);
+  if (!mesh.hasConnectivity()) {
+    PRECICE_DEBUG("No connectivity to send");
+    return;
+  }
+
+  // we need to send the vertexIDs first such that the right connectivity can be created later
+  // contrary to the normal sendMesh, this variant must also work for adding delta meshes
```
uekerman, line 63: Do you know what is meant by "normal sendMesh" here? I have the feeling this is some historic left-over comment.

src/com/CommunicateMesh.cpp
```diff
@@ -116,52 +127,50 @@ void CommunicateMesh::receiveMesh(
...
-  PRECICE_DEBUG("Number of edges to receive: {}", numberOfEdges);
+  bool hasConnectivity{false};
+  _communication->receive(hasConnectivity, rankSender);
+  if (!hasConnectivity) {
+    PRECICE_DEBUG("No connectivity to receive");
+    return;
+  }
 
+  // we need to receive the vertexIDs first such that the right connectivity can be created later
+  // contrary to the normal receiveMesh, this variant must also work for adding delta meshes
```
uekerman, line 138: same as above

uekerman (approved):
I did not look at every detail, but I think I understand the changes and the reasoning :+1:

> It also simplifies handling triangles in tests.

which is not yet part of this PR, right?

src/mesh/Triangle.cpp
fsimonis: By storing the vertices in id-ascending order, we can use lexicographical sort to establish an ordering.
The latter is a requirement for sorting.

src/com/CommunicateMesh.cpp
fsimonis: I don't know and cannot quite figure out what this means.
Maybe this used to be part of the broadcast code?

src/mesh/Triangle.cpp
uekerman: What, we need ordering for sorting? :laughing: Aren't these synonyms?
Why do we need to sort vertices in the first place? Not saying that we don't need to do it, but I did not understand why.

src/com/CommunicateMesh.cpp
uekerman: Yes, could very well be.

src/mesh/Triangle.cpp
fsimonis: An order establishes how things are ordered and allows comparing them. Sorting is an operation.

Establishing a weak partial order for vertices allows to sort vertices.
Sorting the vertices of connectivity primitives has two effects:
1. It maps all permutations to a single well defined case ( edges AB and BA both map to AB. triangles BCA, CAB etc map to ABC)
2. It allows to establish a weak partial order for connectivity primitives.
3. The combination of 1 and 2 allows to sort and remove redundant connectivity elements

fsimonis:
> which is not yet part of this PR, right?

@uekerman no, this will require changing many tests.

PR #1314, opened 2022-06-03
API for setting tetrahedra + VTK exporter support for tetra

boris-martin:
## Main changes of this PR

Added a tetrahedra container to Mesh, a `createTetrahedron` function to add them, and a `setMeshTetrahedron` function in the interface to add them. Also updated the VTK exporter to export tetra. (Other exporters will be updated later)

## Motivation and additional information

Superset of #1310, once this gets merged the diff will be much smaller. (If rebasing is done automatically? Not 100% familiar with how Github works here)
Added a VTK export test and it works.

## Author's checklist

* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [x] I ran `make format` to ensure everything is formatted correctly.
* [x] I sticked to C++14 features.
* [ ] I sticked to CMake version 3.16.3.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

PR #1316, opened 2022-06-07
Added Tetrahedron RTree indexing

boris-martin:
## Main changes of this PR

Added R-Tree indexing of Tetrahedra using axis aligned bounding boxes. Main new feature is the `Index::getEnclosingTetrahedra(const Vector3d &location)` function  which returns the list of `TetrahedronID` of tetra whose AABB contains the point "location". This will be used in the linear-cell-interpolation-mapping to find in which tetrahedron lies a vertex.
Other changes are a function to compute the AABB of a tetra and the R-Tree construction.
More tests are planned but the rest of the code is ready for review (ideally once  #1314 is merged).

Superset of #1314, diff should get smaller once this one is merged.
Part of volumetric coupling project https://github.com/precice/precice/issues/468

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mesh/Tetrahedron.hpp
+++ b/src/mesh/Tetrahedron.hpp
@@ -53,9 +53,6 @@ public:
   /// Returns the radius of the sphere enclosing the tetrahedron.
   double getEnclosingRadius() const;
 
-  // Returns the (axis-aligned) Bounding Box enclosing the tetrahedron
-  BoundingBox getBoundingBox() const;
-
   /**
    * @brief Compares two Tetrahedrons for equality
    *
--- a/src/mesh/Tetrahedron.hpp
+++ b/src/mesh/Tetrahedron.hpp
@@ -2,7 +2,6 @@
 
 #include <array>
 #include <iostream>
-#include "mesh/BoundingBox.hpp"
 #include "precice/types.hpp"
 #include "utils/assertion.hpp"
 
--- a/src/mesh/Tetrahedron.hpp
+++ b/src/mesh/Tetrahedron.hpp
@@ -2,6 +2,7 @@
 
 #include <array>
 #include <iostream>
+#include "mesh/BoundingBox.hpp"
 #include "precice/types.hpp"
 #include "utils/assertion.hpp"
 
@@ -51,6 +52,9 @@ public:
 
   /// Returns the radius of the sphere enclosing the tetrahedron.
   double getEnclosingRadius() const;
+
+  // Returns the (axis-aligned) Bounding Box enclosing the tetrahedron
+  BoundingBox getBoundingBox() const;
 
   /**
    * @brief Compares two Tetrahedrons for equality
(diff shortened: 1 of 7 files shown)


fsimonis (commented):
Could you add a test with a location and 3 tetrahedra:
1. the tet that contains the location.
2. a tet that doesn't contain the location, but whose AABB does.
3. a tet that won't be found at all.

We can use this test later with the mapping too.

Apart from that, all looks fine.

src/mesh/Tetrahedron.hpp
```diff
@@ -2,6 +2,7 @@
 
 #include <array>
 #include <iostream>
+#include "mesh/BoundingBox.hpp"
```
fsimonis, line 5: This is obsolete

src/query/Index.cpp
```diff
@@ -110,6 +113,36 @@ TriangleTraits::Ptr Index::IndexImpl::getTriangleRTree(const mesh::Mesh &mesh)
   return indices.triangleRTree;
 }
 
+TetrahedronTraits::Ptr Index::IndexImpl::getTetraRTree(const mesh::Mesh &mesh)
+{
+  if (indices.tetraRTree) {
+    return indices.tetraRTree;
+  }
+
+  precice::utils::Event e("query.index.getTriangleIndexTree." + mesh.getName());
```
fsimonis, line 122: We need to rename the event

src/query/Index.cpp
```diff
@@ -110,6 +113,36 @@ TriangleTraits::Ptr Index::IndexImpl::getTriangleRTree(const mesh::Mesh &mesh)
...
+
+  // We first generate the values for the tetra rtree.
+  // The resulting vector is a random access range, which can be passed to the
+  // constructor of the rtree for more efficient indexing.
+  std::vector<TetrahedronTraits::IndexType> elements;
+  elements.reserve(mesh.tetrahedra().size());
+  for (size_t i = 0; i < mesh.tetrahedra().size(); ++i) {
+    // Instead of using bg::return_envelope, we compute the BB of the tetra ourself
```
fsimonis, line 130: The comment misses the reason why

src/query/Index.hpp
```diff
@@ -74,6 +76,9 @@ class Index {
   /// Return all the vertices inside a bounding box
   std::vector<VertexID> getVerticesInsideBox(const mesh::BoundingBox &bb);
 
+  /// Return all the tetrahedra whose bounding box contains a vertex
```
fsimonis, line 79: We need to specify that the bounding boxes are axis aligned.

fsimonis (changes requested):

src/query/Index.cpp
boris-martin: The thing is that I didn't fully get why :sweat_smile: 
I couldn't find any reference as to why boost's get_envelope wouldn't work.

fsimonis: There is a guarantee that it works and the fact that one version works doesn't mean that all versions do.

boris-martin: So I just write that we don't trust Boost?

fsimonis: No, we don't want to rely on undocumented behaviour.

boris-martin:
Added a safety factor on the AABB and a test that failed without it and passes with it.

PR #1317, opened 2022-06-07
Utility function for volume integrals

boris-martin:
## Main changes of this PR

Added a function (`integrateVolume`) to compute volume integrals of data on a mesh with connectivity. It is similar to the current `integrate` function that does surface integrals, but it uses triangles/tetrahedra instead of edges/tetra (depending on the dimension).

Superset of #1314, the diff will be smaller once it its merged. Only `Utils.hpp`, `Utils.cpp`, `MeshTests.cpp` are actually changed here.
Future usage for scaled-consistent mapping in the context of volumetric coupling, or for watch-point integrals on volumes.

## Author's checklist

## Reviewers' checklist

- Should we rename `integrate` to `integrateSurface` for consistency ?
- New tests are on `MeshTests.cpp` as they were for surface integrals. This file is getting large, shouldn't we split it with a test file like `IntegralTests.cpp`?

### DIFF
--- a/src/mesh/Utils.cpp
+++ b/src/mesh/Utils.cpp
@@ -36,5 +36,35 @@ Eigen::VectorXd integrate(const PtrMesh &mesh, const PtrData &data)
   return integral;
 }
 
+Eigen::VectorXd integrateVolume(const PtrMesh &mesh, const PtrData &data)
+{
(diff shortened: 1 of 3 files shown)


fsimonis:
Looks all fine. The main challenge here will be #1318

Issue #1273, opened 2022-04-28
Partition of unity based data mappings


davidscn:
**Please describe the problem you are trying to solve.**
Our currently available mapping methods are either limited in terms of the convergence order (NN), limited in terms of computational cost (RBFs) or require additional information (NP and NNG). We would like to extend the available mapping concepts by a partition of unity concept, which is supposed to provide a scalable method for large systems with a higher convergence order.

**Describe the solution you propose.**
The partition of unity method splits the domain up in several mildly overlapping regions (usually circles) and associates to each subregion a compactly supported weighting function. A mapping problem is solved in each subregion and the global solution is recovered by a weighted sum of all interpolated values in affected subregions. From the user perspective, the mapping should be easily configurable just as the currently available mapping methods. There might be new new configuration parameters for the size of the subregions and the overlap granularity (tbd). 

The issue here is meant as a project description and tries to track all relevant PRs and potentially further issues.

PR #1319, opened 2022-06-08
Extract RBF solver into a separate class

DavidSCN:
## Main changes of this PR
Shifts the assembly and solving of the Eigen RBF mapping to a separate class.

Related to #1273: we want to reuse the same RBF kernel for both mapping variants. This reduces and avoids duplication in the long run, since #1166 will add additional complexity in the individual steps. 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mapping/RadialBasisFctMapping.hpp
+++ b/src/mapping/RadialBasisFctMapping.hpp
@@ -156,6 +156,7 @@ void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::clear()
 template <typename RADIAL_BASIS_FUNCTION_T>
 void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::mapConservative(DataID inputDataID, DataID outputDataID)
 {
+  precice::utils::Event e("map.rbf.mapData.From" + this->input()->getName() + "To" + this->output()->getName(), precice::syncMode);
   PRECICE_TRACE(inputDataID, outputDataID);
   using precice::com::AsVectorTag;
 
@@ -273,6 +274,7 @@ void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::mapConservative(DataID inpu
 template <typename RADIAL_BASIS_FUNCTION_T>
 void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::mapConsistent(DataID inputDataID, DataID outputDataID)
 {
+  precice::utils::Event e("map.rbf.mapData.From" + this->input()->getName() + "To" + this->output()->getName(), precice::syncMode);
   PRECICE_TRACE(inputDataID, outputDataID);
   using precice::com::AsVectorTag;
 
--- a/src/mapping/RadialBasisFctMapping.hpp
+++ b/src/mapping/RadialBasisFctMapping.hpp
@@ -3,11 +3,11 @@
 #include <Eigen/Core>
 #include <Eigen/QR>
 
-#include "RadialBasisFctSolver.h"
 #include "com/CommunicateMesh.hpp"
 #include "com/Communication.hpp"
 #include "impl/BasisFunctions.hpp"
 #include "mapping/RadialBasisFctBaseMapping.hpp"
+#include "mapping/RadialBasisFctSolver.hpp"
 #include "mesh/Filter.hpp"
 #include "precice/types.hpp"
 #include "utils/EigenHelperFunctions.hpp"
--- a/src/mapping/RadialBasisFctMapping.hpp
+++ b/src/mapping/RadialBasisFctMapping.hpp
@@ -3,6 +3,7 @@
 #include <Eigen/Core>
 #include <Eigen/QR>
 
+#include "RadialBasisFctSolver.h"
 #include "com/CommunicateMesh.hpp"
 #include "com/Communication.hpp"
 #include "impl/BasisFunctions.hpp"
@@ -55,10 +56,7 @@ public:
 private:
   precice::logging::Logger _log{"mapping::RadialBasisFctMapping"};
 
-  Eigen::MatrixXd _matrixA;
-
-  Eigen::ColPivHouseholderQR<Eigen::MatrixXd> _qr;
-
+  RadialBasisFctSolver _rbfSolver;
   /// @copydoc RadialBasisFctBaseMapping::mapConservative
   virtual void mapConservative(DataID inputDataID, DataID outputDataID) override;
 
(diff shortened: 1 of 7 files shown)


src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -3,6 +3,7 @@
 #include <Eigen/Core>
 #include <Eigen/QR>
 
+#include "RadialBasisFctSolver.h"
```
fsimonis, line 6:
```suggestion
#include "mapping/RadialBasisFctSolver.hpp"
```

This will simplify the migration to full prefixes.

src/mapping/RadialBasisFctSolver.h
```diff
@@ -0,0 +1,181 @@
+#include "impl/BasisFunctions.hpp"
```
fsimonis, line 1:
```suggestion
#include "mapping/impl/BasisFunctions.hpp"
```

src/mapping/RadialBasisFctSolver.h
```diff
@@ -0,0 +1,181 @@
...
+namespace precice {
+extern bool syncMode;
+
+struct RadialBasisFctSolver {
+public:
+  RadialBasisFctSolver() = default;
+
+  /// Assembles the system matrices and computes the decomposition of the interpolation matrix
+  template <typename RADIAL_BASIS_FUNCTION_T>
+  void computeDecomposition(RADIAL_BASIS_FUNCTION_T basisFunction, const mesh::Mesh &inputMesh, const mesh::Mesh &outputMesh, std::vector<bool> deadAxis);
```
fsimonis, line 15: `_matrixA` and `_qr` are created here and without them, this class cannot do much.

So, isn't this function really the constructor of this class?

If this is the case, doesn't this type really represent the `Decomposition` itself?

src/mapping/RadialBasisFctSolver.h
```diff
@@ -0,0 +1,181 @@
+#include "impl/BasisFunctions.hpp"
+#include "precice/types.hpp"
+#include "utils/EigenHelperFunctions.hpp"
+#include "utils/Event.hpp"
+
+namespace precice {
+extern bool syncMode;
+
+struct RadialBasisFctSolver {
```
fsimonis, line 9: Is this a class (with members having to preserve some invariance) or just a struct holding some information in a structured way?

A class would require accessors for `_matrixA` and `_qr`.
A struct wouldn't need `public:` and `private:`. 

```suggestion
class RadialBasisFctSolver {
```

src/mapping/RadialBasisFctSolver.h
```diff
@@ -0,0 +1,181 @@
...
+void RadialBasisFctSolver::clear()
+{
+  _matrixA = Eigen::MatrixXd();
+  _qr      = Eigen::ColPivHouseholderQR<Eigen::MatrixXd>();
+}
+
+// ------- Non-Member Functions ---------
+
+/// Deletes all dead directions from fullVector and returns a vector of reduced dimensionality.
+static inline Eigen::VectorXd reduceVector(
```
fsimonis, line 83: Is there a reason why you moved it from EigenHelperFunctions?
I could imagine that this will be used in the RBF partition of unity code as well.

src/mapping/RadialBasisFctSolver.h
```diff
@@ -0,0 +1,181 @@
...
+  precice::logging::Logger _log{"mapping::RadialBasisFctSolver"};
+
+  Eigen::ColPivHouseholderQR<Eigen::MatrixXd> _qr;
+};
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+static Eigen::MatrixXd buildMatrixCLU(RADIAL_BASIS_FUNCTION_T basisFunction, const mesh::Mesh &inputMesh, std::vector<bool> deadAxis);
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+static Eigen::MatrixXd buildMatrixA(RADIAL_BASIS_FUNCTION_T basisFunction, const mesh::Mesh &inputMesh, const mesh::Mesh &outputMesh, std::vector<bool> deadAxis);
```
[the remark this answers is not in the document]
DavidSCN: Due to some template reasons, I need forward declarations here (or reorder the function definitions).

src/mapping/RadialBasisFctSolver.h
DavidSCN: That's a preparation of #1320. I will need it for PU as well, but hopefully only in this file.

src/mapping/RadialBasisFctSolver.h
fsimonis: > Yes, you can see it as a constructor.

If the type provdes default construction and copy assignment, then you could write

```cpp
_rbfSolver = RadialBasicFctSolver{_basisFunction, _inputMesh, _outputMesh, _deadAxis};
```

( Yes, template deduction in constructors is another good reason to move to C++17 )

> Which type are you referring to?

The `RadialBasisFctSolver` really is a `RBFDecomposition`, right?

src/mapping/RadialBasisFctSolver.h
```diff
@@ -0,0 +1,181 @@
...
+  int n = inputSize + polyparams; // Add linear polynom degrees
+
+  Eigen::MatrixXd matrixCLU(n, n);
+  matrixCLU.setZero();
+
+  for (int i = 0; i < inputSize; ++i) {
+    for (int j = i; j < inputSize; ++j) {
+      const auto &u   = inputMesh.vertices()[i].getCoords();
+      const auto &v   = inputMesh.vertices()[j].getCoords();
+      matrixCLU(i, j) = basisFunction.evaluate(reduceVector((u - v), deadAxis).norm());
```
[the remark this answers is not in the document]
fsimonis: Ah, got it. Thanks for the clarification!

src/mapping/RadialBasisFctSolver.h
fsimonis: You could also create an instance of `RadialBasicFctSolver` in `computeDecomposition`, essentially turning it into a factory method.

```cpp
_rbfSolver = computeDecomposition(_basisFunction, _inputMesh, _outputMesh, _deadAxis);
```

The useless default state is kind of covered by the assertions of Eigen, right?

DavidSCN: > The useless default state is kind of covered by the assertions of Eigen, right?

A bit cryptic to track down, but yes. I will add another assertion in the solve subroutine.

DavidSCN: I went now for the two constructor solution you proposed (although we still end up having a useless default state, which is asserted, however).

src/mapping/RadialBasisFctSolver.h
```diff
@@ -0,0 +1,181 @@
...
+};
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+static Eigen::MatrixXd buildMatrixCLU(RADIAL_BASIS_FUNCTION_T basisFunction, const mesh::Mesh &inputMesh, std::vector<bool> deadAxis);
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+static Eigen::MatrixXd buildMatrixA(RADIAL_BASIS_FUNCTION_T basisFunction, const mesh::Mesh &inputMesh, const mesh::Mesh &outputMesh, std::vector<bool> deadAxis);
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+void RadialBasisFctSolver::computeDecomposition(RADIAL_BASIS_FUNCTION_T basisFunction, const mesh::Mesh &inputMesh, const mesh::Mesh &outputMesh, std::vector<bool> deadAxis)
```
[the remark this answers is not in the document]
DavidSCN: See https://github.com/precice/precice/pull/1319#discussion_r901387352

fsimonis (approved):
Definitely improves the readability of the RBF mapping!

Issue #1321, opened 2022-06-13
Clarify mesh API


fsimonis:
**Please describe the problem you are trying to solve.**

The API for setting mesh primitives is confusing and tedious.

* Triangles have to be set using edges
* `setMeshTriangleWithEdges` sounds like it would take edges, but it actually takes vertices. #1057
* There are no bulk functions for setting edges and triangles. #465 
* Adding `setMeshTetrahedron` requiring triangles would be a huge pain for users. #1314 
* Exposing handles to connectivity (EdgeID) prevents us from optimizing meshes #1313 

**Describe the solution you propose.**

1. Change the API to a vertex-only style. **:warning: breaking**
2. Remove `XWithEdges`
3. Add bulk functions

Function | Inputs + MeshID | Outputs | Comment
--- | --- | --- | ---
setMeshVertex | Coords | VertexID | _unchanged_
setMeshVertices | Count, Coords | VertexIDs | _unchanged_
setMeshEdge | VertexIDs | | _no return_
setMeshEdges | Count, VertexIDs | | _no return_
setMeshTriangle | VertexIDs | | _changed input_
setMeshTriangles | Count, VertexIDs | | _new_
setMeshQuad | VertexIDs | | _changed input_
setMeshQuads | Count, VertexIDs | | _new_
setMeshTetrahedron | VertexIDs | | _unchanged_
setMeshTetrahedra | Count, VertexIDs | | _new_

**Describe alternatives you've considered**

Leave it as it and end up with an increasingly confusing API.

PR #1313, opened 2022-06-02
Add mess compression

fsimonis:
## Main changes of this PR

This is an example PR of how mesh compression could look like.
It is implemented by removing duplicate connectivity entries in `intialize()`.

## Motivation and additional information

Smaller mesher means less work

## Author's checklist

* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [ ] I ran `make format` to ensure everything is formatted correctly.
* [ ] I sticked to C++14 features.
* [ ] I sticked to CMake version 3.16.3.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

PR #1322, opened 2022-06-14
Vertex-only API

fsimonis:
## Main changes of this PR

This PR removes all edge ids from the API and converts `setMeshXWithEdges` functions to `setMeshX`. 
The API now guarantees the creation of implicit connectivity information. 

This PR is based on #1312.

See #1321

This only ports the integration tests to the edge-less API.
A further step would be to remove `Mesh::createTriangle(Edge&, Edge&, Edge&)`, which impacts many tests.

## Author's checklist

## Reviewers' checklist


src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -804,7 +804,7 @@ void SolverInterfaceImpl::getMeshVertexIDsFromPositions(
   }
 }
 
-int SolverInterfaceImpl::setMeshEdge(
+void SolverInterfaceImpl::setMeshEdge(
     MeshID meshID,
     int    firstVertexID,
     int    secondVertexID)
```
DavidSCN, line 810: Now that you are touching these API functions, I was wondering whether it is a good time to switch from an `int forstVertexID` to ` unsigned int` or maybe even `VertexID` (of type size_type) or something. I guess it will be  inevitable in the long run when going towards larger problems.

uekerman: Upvote, but I think it could be easier in a separate PR. Is there already a v3.0 issue for this?

fsimonis: Very good point indeed

fsimonis: > Is there already a v3.0 issue for this?

Regarding 64-bit ids: #887

We can change the parameter types (and names) now. This doesn't change the API as is.
Then we change the types in 3.0.0.

fsimonis: To get this PR done soonish, I'll touch the interface types later.

fsimonis:
@uekerman I added bulk functions #469 as well, bindings and tests.

src/precice/SolverInterface.hpp
```diff
@@ -498,87 +498,66 @@ class SolverInterface {
...
    */
-  void setMeshTriangle(
-      int meshID,
-      int firstEdgeID,
-      int secondEdgeID,
-      int thirdEdgeID);
+  void setMeshEdges(
+      int        meshID,
+      int        size,
+      const int *vertices);
```
uekerman, line 526: I think it would be good to give an example here. The necessary ordering of vertices is not yet clearly described. "l1, r1, l2, r2, l3, r3" vs "l1, l2, l3, r1, r2, r3" 
Same concern for all other bulk variants below.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -966,7 +895,7 @@ void SolverInterfaceImpl::setMeshQuadWithEdges(
 {
   PRECICE_TRACE(meshID, firstVertexID,
                 secondVertexID, thirdVertexID, fourthVertexID);
-  PRECICE_CHECK(_dimensions == 3, "setMeshQuadWithEdges is only possible for 3D cases."
+  PRECICE_CHECK(_dimensions == 3, "setMeshQuadis only possible for 3D cases."
```
uekerman, line 898:
```suggestion
  PRECICE_CHECK(_dimensions == 3, "setMeshQuad is only possible for 3D cases."
```

src/mesh/Tetrahedron.cpp
```diff
@@ -34,6 +32,9 @@ Tetrahedron::Tetrahedron(
           (&vertexTwo != &vertexFour) &&
           (&vertexThree != &vertexFour),
       "Tetrahedron vertices are not unique!");
+
+  std::sort(_vertices.begin(), _vertices.end(),
+            [](const Vertex *lhs, const Vertex *rhs) { return *lhs < *rhs; });
```
uekerman, line 37: Needs documentation: why do we need sorting?
Do we have a unit test that tests the sorting?

extras/bindings/fortran/include/precice/SolverInterfaceFortran.hpp
```diff
@@ -546,10 +502,10 @@ void precicef_set_quad_(
  * IN:  meshID, firstVertexID, secondVertexID, thirdVertexID, fourthVertexID
  * OUT: -
  *
- * @copydoc precice::SolverInterface::setMeshQuadWithEdges()
+ * @copydoc precice::SolverInterface::setMeshQuad()
  *
  */
-void precicef_set_quad_we_(
+void precicef_set_quad_(
```
uekerman, line 508: Seems like the bulk methods are still missing in the header of the Fortran bindings?

extras/bindings/fortran/src/SolverInterfaceFortran.cpp
```diff
@@ -334,53 +334,58 @@ void precicef_get_vertex_ids_from_positions_(
...
     const int *secondVertexID,
     const int *thirdVertexID)
 {
   PRECICE_CHECK(impl != nullptr, errormsg);
-  impl->setMeshTriangleWithEdges(*meshID, *firstVertexID, *secondVertexID, *thirdVertexID);
+  impl->setMeshTriangle(*meshID, *firstVertexID, *secondVertexID, *thirdVertexID);
 }
 
-void precicef_set_quad_(
+void precicef_set_mesh_triangles(
```
uekerman, line 362:
```suggestion
void precicef_set_mesh_triangles_(
```
?

extras/bindings/fortran/src/SolverInterfaceFortran.cpp
```diff
@@ -334,53 +334,58 @@ void precicef_get_vertex_ids_from_positions_(
...
     const int *secondVertexID,
     const int *thirdVertexID,
     const int *fourthVertexID)
 {
   PRECICE_CHECK(impl != nullptr, errormsg);
-  impl->setMeshQuadWithEdges(*meshID, *firstVertexID, *secondVertexID, *thirdVertexID, *fourthVertexID);
+  impl->setMeshQuad(*meshID, *firstVertexID, *secondVertexID, *thirdVertexID, *fourthVertexID);
+}
+
+void precicef_set_mesh_quads(
```
uekerman, line 382:
```suggestion
void precicef_set_mesh_quads_(
```
?

extras/bindings/fortran/src/SolverInterfaceFortran.cpp
```diff
@@ -394,6 +399,15 @@ void precicef_set_tetrahedron(
   impl->setMeshTetrahedron(*meshID, *firstVertexID, *secondVertexID, *thirdVertexID, *fourthVertexID);
 }
 
+void precicef_set_mesh_tetrahedra(
```
uekerman, line 402:
```suggestion
void precicef_set_mesh_tetrahedra_(
```
?

tests/serial/mapping-nearest-projection/MappingNearestProjectionEdges.cpp
```diff
@@ -0,0 +1,28 @@
...
+#include <boost/test/data/test_case.hpp>
+#include <precice/SolverInterface.hpp>
+#include "helpers.hpp"
+
+BOOST_AUTO_TEST_SUITE(Integration)
+BOOST_AUTO_TEST_SUITE(Serial)
+BOOST_AUTO_TEST_SUITE(MappingNearestProjection)
+BOOST_DATA_TEST_CASE(MappingNearestProjectionEdges,
+                     boost::unit_test::data::make({true, false}) * boost::unit_test::data::make({true, false}),
+                     defineEdgesExplicitly, useBulkFunctions)
```
uekerman, line 14: I assume this tests the complete 2x2 matrix? Nice!

tests/serial/mapping-nearest-projection/helpers.hpp
```diff
@@ -7,12 +7,32 @@
...
+
+inline void testQuadMappingNearestProjectionTallKite(bool defineEdgesExplicitly, const std::string configFile, const TestContext &context)
+{
+  testQuadMappingNearestProjectionTallKite(defineEdgesExplicitly, false, configFile, context);
+}
+
+inline void testQuadMappingNearestProjectionWideKite(bool defineEdgesExplicitly, const std::string configFile, const TestContext &context)
+{
+  testQuadMappingNearestProjectionWideKite(defineEdgesExplicitly, false, configFile, context);
+}
```
uekerman, line 36: The alternative would be to port all tests do the 4-argument functions, right?
If yes, why did you choose this solution? (not saying it is bad, just to understand ... maybe we need some "why" documentation here)

tests/serial/mapping-nearest-projection/helpers.cpp
```diff
@@ -288,21 +313,31 @@ void testQuadMappingNearestProjectionWideKite(bool defineEdgesExplicitly, const
...
     }
 
     auto &mesh = testing::WhiteboxAccessor::impl(interface).mesh("MeshOne");
     BOOST_REQUIRE(mesh.vertices().size() == 4);
-    BOOST_REQUIRE(mesh.edges().size() == 5);
+    if (defineEdgesExplicitly) {
+      BOOST_REQUIRE(mesh.edges().size() == 9);
+    } else {
+      BOOST_REQUIRE(mesh.edges().size() == 5);
+    }
```
uekerman, line 340: Did we document this behavior somewhere?
This will also have a performance influence for users, no?

uekerman (changes requested):
Looks pretty clean already. I tried to have a thorough look, but I did not try it out.
A bit concerning is still the limited test coverage for the bindings -> We might find issues while porting adapters.

src/mesh/Tetrahedron.cpp
fsimonis: We can test this more efficiently in #1313

tests/serial/mapping-nearest-projection/helpers.hpp
fsimonis: The inline functions were helpers during the migration. I removed them.

This is what you meant, right?

src/precice/SolverInterface.hpp
fsimonis: Good point!

tests/serial/mapping-nearest-projection/helpers.cpp
fsimonis: I extended the documentation of the API functions.
Is it sufficient to mention that the bulk versions are more efficient without stating why?

tests/serial/mapping-nearest-projection/helpers.hpp
uekerman: yes :+1:

tests/serial/mapping-nearest-projection/helpers.cpp
uekerman: > Is it sufficient to mention that the bulk versions are more efficient without stating why?

Are they actually as you now just call the single versions in a loop? 

But this here is sth different, no?
It's more efficient if you define edges yourself (explicitly) and don't let preCICE do it?

fsimonis: Correct, they aren't yet. But they will be.

Upcoming tasks are to prevent duplicates in plural calls and to implement the mesh compression (open PR)

I just wanted to get the API done in this PR.

uekerman: Ah, ok. Then this test will be updated in the mesh compression PR?

uekerman (approved):
Good to go from my side, but I did not try it out.

PR #1330, opened 2022-06-20
Handle tetrahedra in partitioning

boris-martin:
## Main changes of this PR

Followup of https://github.com/precice/precice/pull/1325. Tetrahedra weren't added when gathering meshes.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mesh/Mesh.cpp
+++ b/src/mesh/Mesh.cpp
@@ -11,6 +11,7 @@
 
 #include "Edge.hpp"
 #include "Mesh.hpp"
+#include "Tetrahedron.hpp"
 #include "Triangle.hpp"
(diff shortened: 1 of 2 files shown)


fsimonis:
Didn't you add some tests for the mesh communication?

boris-martin:
For communications yes, but not for this part. So the unit tests pass but the integration test (of the LCI mapping) I'm writing doesn't.

boris-martin:
The addMesh function has 0 unit tests as of now, actually.

fsimonis:
This isn't really a fix as you haven't touched repartitioning of parallel cases yet. So its more of an addition.

PR #1333, opened 2022-06-21
Added tolerance to isInterpolation check

boris-martin:
## Main changes of this PR
In Barycentric mappings (NP and Linear Cell Interpolation) we check that the coefficients of an interpolation are positive (because if they're not, it's act `x >= 0` statement. This is replaced by a check with tolerance to avoid bad behavior when a coefficient is close to zero.

## Author's checklist

## Reviewers' checklist


fsimonis:
There really isn't a need to pass the tolerance to isInterpolation, lets simplify this.

Also, this needs a test.

fsimonis (approved):

PR #1335, opened 2022-06-22
Add all polynomial options to RBF Eigen mappings

DavidSCN:
## Main changes of this PR
Replaces #1166, which was too stale in order to recycle it. Related to #1162.

Depends on #1320. This gives more flexibility for our RBF mappings and how to treat the resulting systems.

## Author's checklist

* [x] Add proper testing for the polynomial options

## Reviewers' checklist


src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -212,7 +219,7 @@ void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::mapConservative(DataID inpu
 
     // Construct Eigen vectors
     Eigen::Map<Eigen::VectorXd> inputValues(globalInValues.data(), globalInValues.size());
-    Eigen::VectorXd             outputValues((_rbfSolver.getEvaluationMatrix().cols() - this->getPolynomialParameters()) * valueDim);
+    Eigen::VectorXd             outputValues((this->output()->getGlobalNumberOfVertices()) * valueDim);
```
DavidSCN, line 222: I was not sure whether it is a good idea to rely on `getGlobalNumberOfVertices`. I needed to add the information in all places in the RBF unit tests. Motivation was here to get rid of the dependency of the polynomial.

uekerman: It should give the correct value AFAIK. Alternative could be to store this in `_rbfSolver`.

uekerman (approved):
Looks good :+1: Did not check every detail, but I understand the code overall.

What do we do with "polynomial = off"? Is it allowed? It seems we have no unit tests for this option.

The class `RadialBasisFctSolver` needs a better name or at least some minimal class documentation. At first sight, it is not clear that this class only does direct solving and has nothing to do with `PetRadialBasisFctMapping`.

DavidSCN:
> What do we do with "polynomial = off"? Is it allowed? It seems we have no unit tests for this option.

I need to add more tests here anyway. `polynomial=off` works, but it doesn't result in the same accuracy for our simple test cases, so I still need to set up a few new test cases to cover this option.

PR #1337, opened 2022-06-23
Extend LCI mapping to 3D

boris-martin:
## Main changes of this PR

- Linear Cell Interpolation Mapping can now be used in 3D (through tetrahedra)
- Relevant changes to index & queries
- Bunch of tests

## Author's checklist

## Reviewers' checklist


docs/changelog/1337.md
```diff
@@ -0,0 +1 @@
+- Linear Cell Interpolation mapping (`linear-cell-interpolation`) can now be used in 3D. Set tetrahedra on the relevant mesh to use it.
```
fsimonis, line 1:
```suggestion
- Added 3D support to Linear Cell Interpolation mapping (`<mapping:linear-cell-interpolation >/`) using tetrahedra.
```

src/query/Index.cpp
```diff
@@ -271,8 +271,17 @@ ProjectionMatch Index::findCellOrProjection(const Eigen::VectorXd &location, int
...
-    PRECICE_UNREACHABLE("Volume coupling 3D not  implemented");
-    return findTriangleProjection(location, n);
+
+    // Find correct tetra, or fall back to NP
+    auto matchedTetra = getEnclosingTetrahedra(location);
+    for (const auto &match : matchedTetra) {
+      // Matches are raw indices, not (indices, distance) pairs
+      auto polation = mapping::Polation(location, _mesh->tetrahedra()[match]);
+      if (polation.isInterpolation()) {
+        return {polation, 0.0};
```
DavidSCN, line 281: Could you comment why you are returning zero here instead of the distance? You wrote it's only the raw index without the distance. Is the zero supposed to be a dummy value? Could we get rid of it?

tests/serial/mapping-volume/helpers.hpp
```diff
@@ -9,5 +9,7 @@ using precice::testing::TestContext;
 
 void testMappingVolumeOneTriangle(const std::string configFile, const TestContext &context);
 void testMappingVolumeOneTriangleConservative(const std::string configFile, const TestContext &context);
+void testMappingVolumeOneTetra(const std::string configFile, const TestContext &context);
+void testMappingVolumeOneTetraConservative(const std::string configFile, const TestContext &context);
 
 #endif
```
DavidSCN, line 15: Could you add a newline at the end of all relevant files?

DavidSCN (commented):
Looks good, only minor comments. Is it possible to try your mapping using aste? Would be interested in trying it out.

tests/parallel/mapping-volume/ParallelCube1To3.xml
```diff
@@ -0,0 +1,42 @@
...
+
+    <mesh name="MeshOne">
+      <use-data name="DataOne" />
+    </mesh>
+
+    <mesh name="MeshTwo">
+      <use-data name="DataOne" />
+    </mesh>
+
+    <participant name="SolverOne1_3">
```
uekerman, line 14:
```suggestion
    <participant name="SolverOne">
```
In my opinion a bit much to encode the test name into the names here.

uekerman (approved):
I did not check all details, but looks very clean to me. Nice job :ok_hand:

tests/parallel/mapping-volume/ParallelCube1To3.xml
boris-martin: It made debugging easier when exporting data (otherwise I ended up with different tests overwriting each other's output) and helped me find relevant logs from irrelevant ones, but I could replace them if you think that's better.

src/query/Index.cpp
fsimonis: The comment needs improvement. The location is within the cell, hence a distance of 0.

boris-martin: The comment above just explains why we use mesh->tetrahedra()[match] instead of the match.first like with triangles. (This comes from the nature of the query)
I'll add another comment
Zero is indeed because we are inside, and non-zero values occur only with fallback to NP

tests/parallel/mapping-volume/ParallelCube1To3.xml
uekerman: My normal workflow is to only run one case when debugging. Then, you don't run into this problem. Are you aware of this trick?

```bash
mpirun -np 4 ./testprecice -x  -t "*/*/*/*/*/ReadWriteScalarDataFirstParticipantInitData"
```

boris-martin: Ah, good to know ! I didn't master this so I stickted to `ctest -R Parallel`which was a bit too general

fsimonis:
Merging this before it gets too much out of date.

PR #1338, opened 2022-06-23
Improve basis function implementation

DavidSCN:
## Main changes of this PR
Switches the basis function implementation from using `std::pow` to a pow variant relying on recursion.

The main problem is that `std::pow` operates on a `double` exponent (there is no `int` variant), but we use it in the basis function implementation only for integer exponents. The basis function evaluation routine is heavily performance critical. With the changes here, we get a considerable performance gain in both RBF implementations (Eigen and PETSc). Here some timings for the (expensive) computation of matrix A as well as the overall mapping computation:

[result-asmA.pdf](https://github.com/precice/precice/files/8966071/result-asmA.pdf)

[result-computet.pdf](https://github.com/precice/precice/files/8966064/result-computet.pdf)

Example: For the c6 Wendland RBF, we see a speed-up of ~4.

Also related to #1320 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mapping/impl/BasisFunctions.hpp
+++ b/src/mapping/impl/BasisFunctions.hpp
@@ -38,7 +38,7 @@ class ThinPlateSplines : public NoCompactSupportBase {
 public:
   inline double evaluate(double radius) const
   {
-    return std::log(std::max(radius, 1e-14)) * math::pow_int<2>(radius);
+    return std::log(std::max(radius, math::NUMERICAL_ZERO_DIFFERENCE)) * math::pow_int<2>(radius);
   }
 };
 
@@ -189,7 +189,7 @@ public:
     double const p = radius * _r_inv;
     if (p >= 1)
       return 0.0;
-    return 1.0 - 30.0 * math::pow_int<2>(p) - 10.0 * math::pow_int<3>(p) + 45.0 * math::pow_int<4>(p) - 6.0 * math::pow_int<5>(p) - math::pow_int<3>(p) * 60.0 * std::log(std::max(p, 1e-15));
+    return 1.0 - 30.0 * math::pow_int<2>(p) - 10.0 * math::pow_int<3>(p) + 45.0 * math::pow_int<4>(p) - 6.0 * math::pow_int<5>(p) - math::pow_int<3>(p) * 60.0 * std::log(std::max(p, math::NUMERICAL_ZERO_DIFFERENCE));
   }
 
 private:
--- a/src/mapping/impl/BasisFunctions.hpp
+++ b/src/mapping/impl/BasisFunctions.hpp
@@ -36,13 +36,9 @@ struct NoCompactSupportBase {
  */
 class ThinPlateSplines : public NoCompactSupportBase {
 public:
-  double evaluate(double radius) const
+  inline double evaluate(double radius) const
   {
-    double result = 0.0;
-    if (math::greater(radius, 0.0)) {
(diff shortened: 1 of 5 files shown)


src/mapping/impl/BasisFunctions.hpp
```diff
@@ -177,31 +173,29 @@ class Gaussian : public CompactSupportBase {
...
   {
-    if (radius >= _r)
+    double const p = radius * _r_inv;
+    if (p >= 1)
       return 0.0;
-    double const p = radius / _r;
-    using std::log;
-    using std::pow;
-    return 1.0 - 30.0 * pow(p, 2.0) - 10.0 * pow(p, 3.0) + 45.0 * pow(p, 4.0) - 6.0 * pow(p, 5.0) - 60.0 * log(pow(p, pow(p, 3.0)));
+    return 1.0 - 30.0 * math::pow_int<2>(p) - 10.0 * math::pow_int<3>(p) + 45.0 * math::pow_int<4>(p) - 6.0 * math::pow_int<5>(p) - math::pow_int<3>(p) * 60.0 * std::log(std::max(p, 1e-15));
```
DavidSCN, line 192: Instead of going for this overly complicated formula in the end log(p^(p^3)) I selected now a fixed cutoff for the log (1e-14). The overall product will be zero anyway, as well multiply it by (1e-14)^3.

DavidSCN:
> It is obscure, but there actually is an overload that matches a double base and a integral exponent. Its overload 7 [of the docs](https://en.cppreference.com/w/cpp/numeric/math/pow).

I was indeed a bit puzzled about this version. The docs say "If any argument has [integral type](https://en.cppreference.com/w/cpp/types/is_integral), it is cast to double."

So, I guess it is still double?

Anyway, the recursion version is still faster than the `std::pow(arg, int(N)` method:

[result-asmA.pdf](https://github.com/precice/precice/files/8969618/result-asmA.pdf)

fsimonis:
Ah I see. You lifted the exponent into the type system. Makes sense that it is faster.

DavidSCN:
> Ah I see. You lifted the exponent into the type system. Makes sense that it is faster.

Yes (although I would have expected a somewhat similar result for small exponents from the STL).

> You did test using a Release build, right?

Yes.

src/math/math.hpp
```diff
@@ -18,5 +18,18 @@ inline int sign(double number)
   return 0;
 }
 
+/// Computes the power of an integer (x^iexp) using recursion, which is much faster than std::pow(x, iexp)
+template <int iexp, typename T>
+inline constexpr T pow_int(const T x)
+{
+  static_assert(iexp >= 0, "Exponent must be an integer greater zero.");
```
fsimonis, line 25: Condition allows 0, while the text doesn't.

src/math/math.hpp
```diff
@@ -18,5 +18,18 @@ inline int sign(double number)
...
+
+  if (iexp == 0)
+    return static_cast<T>(1.);
+  else
```
fsimonis, line 29: It would be interesting to see how this performs when treating iexp=0 separately.

Could be faster as it saves 1 level of recursion, but could also hinder the tail recursion optimisation.

DavidSCN: I just tried and both are quite close to each other. I ran it for several times and sometimes the one or the other is (slightly) faster. Considering that the additional treatment of `iexp=0` leads to more code, I would propose to go for the more compact version here?!

fsimonis (approved):

PR #1339, opened 2022-06-24
Make areas of triangles positive in 2D + Removed dead code

boris-martin:
## Main changes of this PR

Previously, `Triangle::getArea()` had a different behavior depending on the dimension: in 2D it returns a signed area (depending on the ordering of vertices) but a positive area in 3D. This breaks computations of volumetric integrals (https://github.com/precice/precice/pull/1317), and seems to be legacy code, as this signed are was never used (since there were no 2D triangles before!)
Now, this area is positive. It breaks unit tests of unused functions, so these tests & relevant functions have been removed too.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/math/geometry.cpp
+++ b/src/math/geometry.cpp
@@ -15,71 +15,6 @@ namespace geometry {
 
 logging::Logger _log("math::geometry");
 
-bool segmentsIntersect(
-    const Eigen::Ref<const Eigen::Vector2d> &a,
-    const Eigen::Ref<const Eigen::Vector2d> &b,
-    const Eigen::Ref<const Eigen::Vector2d> &c,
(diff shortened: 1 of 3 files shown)


boris-martin:
@uekerman Any idea of the context in which this code was written ?

uekerman (approved):
This should not be problematic.
Back in the days, preCICE had a "geometry mode", participants could query geometry information provided by preCICE. Example: preCICE imports an STL file and then a participant could ask whether a vertex is inside or outside the geometry (needed e.g. for immersed boundary methods). For more information, have a look the respective sections of Bernhard's thesis.
We eventually removed the geometry mode to focus more on the core functionalities of preCICE. This is very likely some leftover dead code.

PR #1341, opened 2022-06-26
Extend warning about MPI ports with OpenMPI

MakisH:

It looks like the warning about MPI ports sounds too intimidating, and users end up using Sockets even if everything is alright with MPI ports.
Main reason seems to be that "if you encounter problems" is too vague.
This adds a clarification on when to ignore this warning.

## Author's checklist

* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release. -> N/A
* [ ] I sticked to C++14 features. -> N/A
* [ ] I sticked to CMake version 3.16.3. -> N/A

## Reviewers' checklist

@fsimonis: Do I understand correctly that if the simulation starts, no other related issues are expected?

### DIFF
--- a/src/m2n/config/M2NConfiguration.cpp
+++ b/src/m2n/config/M2NConfiguration.cpp
@@ -181,7 +181,7 @@ void M2NConfiguration::xmlTagCallback(const xml::ConfigurationContext &context, 
                     "Either switch to a \"sockets\" communication or recompile preCICE with \"PRECICE_MPICommunication=ON\".");
 #else
 #ifdef OMPI_MAJOR_VERSION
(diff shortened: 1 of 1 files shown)


fsimonis:
It occasionally fails during the communication build-up, being part of `initialize()`.
I am not aware of issues later on.

MakisH:
> It occasionally fails during the communication build-up, being part of `initialize()`. I am not aware of issues later on.

Could we maybe provide more concrete details in our documentation? https://github.com/precice/precice.github.io/pull/176

fsimonis (approved):
Then addition looks reasonable. :smiley:

Issue #1318, opened 2022-06-07
API for (future) usage of volume integrals


boris-martin:
**Please describe the problem you are trying to solve.**
With https://github.com/precice/precice/pull/1317 and the use of volumetric coupling, at some point it will be necessary to implement a scaled-consistent mapping that scales according to volume integrals instead of surface integrals. Similarly, watch integrals should be usable with volume integrals (instead of surface integrals or raw values).
The API (the XML config actually) should change to allow this choice, with defaults unchanged for backward compatibility.

**Describe the solution you propose.**
For scaled-consistent mapping: an additional attribute on the mapping that could be a string (surface/volume) or a boolean (useVolume with default to false ?)
For watch-integrals: either replace the bool scaleWithArea by a string (vertex/surface/volume) but that would be breaking, or an additional attribute. Keeping "scaleWithArea" with an additional "useVolume" would be weird linguisitcally speaking, but would work well otherwise.

**Additional context**
Part of #468

PR #1342, opened 2022-06-27
Internal support for volumetric scale-consistent mapping

boris-martin:
## Main changes of this PR

Scale-consistent mapping can now be configured to use volume integrals instead of surface integrals. API support will come in a later PR; this one focuses on internal support.
Integral type is given to the mapping constructor and stored.

See https://github.com/precice/precice/issues/1318 for discussion about future API.
NP mapping always use surface mode, and LCI mapping the volume mode. All the other mappings can use either, with default to surface for backward compatibility.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mapping/BarycentricBaseMapping.hpp
+++ b/src/mapping/BarycentricBaseMapping.hpp
@@ -14,7 +14,7 @@ namespace mapping {
  */
 class BarycentricBaseMapping : public Mapping {
 public:
-  BarycentricBaseMapping(Constraint constraint, int dimensions);
+  BarycentricBaseMapping(Constraint constraint, int dimensions, Mapping::CouplingKind kind);
 
   /// Destructor, empty.
   virtual ~BarycentricBaseMapping() = default;
--- a/src/mapping/BarycentricBaseMapping.hpp
+++ b/src/mapping/BarycentricBaseMapping.hpp
@@ -20,7 +20,7 @@ public:
   virtual ~BarycentricBaseMapping() = default;
 
   /// Computes the projections and interpolation relations. Must be done by inherited subclass
-  virtual void computeMapping() = 0;
+  virtual void computeMapping() override = 0;
 
   /// Removes a computed mapping.
   virtual void clear() final;
--- a/src/mapping/Mapping.cpp
+++ b/src/mapping/Mapping.cpp
@@ -129,20 +129,14 @@ void Mapping::scaleConsistentMapping(int inputDataID, int outputDataID) const
   bool requiresTriangles = (spaceDimension == 2 and _couplingKind == CouplingKind::VOLUME) or (spaceDimension == 3 and _couplingKind == CouplingKind::SURFACE);
   bool requiresTetra     = (spaceDimension == 3 and _couplingKind == CouplingKind::VOLUME);
 
-  if (not input()->vertices().empty()) {
-    if ((requiresEdges and input()->edges().empty()) or
-        (requiresTriangles and input()->triangles().empty()) or (requiresTetra and input()->tetrahedra().empty())) {
-      PRECICE_ERROR("Connectivity information is missing for the mesh {}. "
-                    "Scaled consistent mapping requires connectivity information.",
-                    input()->getName());
-    }
-  }
-  if (not output()->vertices().empty()) {
-    if ((requiresEdges and output()->edges().empty()) or
-        (requiresTriangles and output()->triangles().empty()) or (requiresTetra and output()->tetrahedra().empty())) {
-      PRECICE_ERROR("Connectivity information is missing for the mesh {}. "
-                    "Scaled consistent mapping requires connectivity information.",
-                    output()->getName());
+  for (mesh::PtrMesh mesh : {input(), output()}) {
+    if (not mesh->vertices().empty()) {
+      if ((requiresEdges and mesh->edges().empty()) or
+          (requiresTriangles and mesh->triangles().empty()) or (requiresTetra and mesh->tetrahedra().empty())) {
+        PRECICE_ERROR("Connectivity information is missing for the mesh {}. "
+                      "Scaled consistent mapping requires connectivity information.",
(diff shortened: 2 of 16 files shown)


src/mapping/Mapping.cpp
```diff
@@ -9,16 +9,18 @@ namespace precice {
...
+    bool                  requireGradient,
+    Mapping::CouplingKind couplingKind)
     : _requireGradient(requireGradient),
       _constraint(constraint),
       _inputRequirement(MeshRequirement::UNDEFINED),
       _outputRequirement(MeshRequirement::UNDEFINED),
       _input(),
       _output(),
-      _dimensions(dimensions)
+      _dimensions(dimensions),
```
DavidSCN, line 22: Yes, seems like we have to re-order these things in order to get rid of the warnings here

src/mapping/Mapping.hpp
```diff
@@ -46,8 +46,21 @@ class Mapping {
     FULL = 2
   };
 
+  /**
+   * @brief Specifices what kind of integrals to compute for scaled-consistent mapping.
+   * 
+   * Scaled-Consistent mappings conserve integrals, but they must be computed differently 
+   * depending on the nature of the coupling. It can be either SURFACE coupling (e.g. pressure to conserve forces)
+   * or VOLUMETRIC coupling (e.g. density to preserve mass).
```
DavidSCN, line 54: Do you mean by "nature of the coupling" just volume coupling vs surface coupling? In principle, I would expect that the user provides the desired degree of connectivity he wants to apply (and is also able to fetch the information from preCICE).

Internally, one could then distinguish using the space dimension and the size of the connectivity container. If we still want to keep the `enum`, I would hope for a different name, as 'kind' means kind of anything (could also mean explicit, implicit, consistent etc). What about `CouplingGeometry`, `CouplingDimension`, `CouplingEntity`?!

DavidSCN (commented):
Looks good, thanks! Only some comments regarding the naming.

src/mapping/Mapping.hpp
boris-martin: The way I see it, the user would provide it in the XML config. It feels safer to me than guessing from the connectivity. Would you want something like `if (mesh->tetrahedra().empty) integrateVolume(...) else integrateSurface()` ? (Assuming 3D here)

I'm not a big fan of your suggestions but I agree that "kind" is not ideal. What about `IntegralType`? Or a boolean `integrateOnBoundary` (defaulting to true) ? That's probably a lot of booleans though since there is already the gradient one, so we'll end up with unreadable calls like Mapping(consistent, 3, false, false)

src/mapping/BarycentricBaseMapping.hpp
```diff
@@ -14,7 +14,7 @@ namespace mapping {
  */
 class BarycentricBaseMapping : public Mapping {
 public:
-  BarycentricBaseMapping(Constraint constraint, int dimensions);
+  BarycentricBaseMapping(Constraint constraint, int dimensions, Mapping::CouplingKind kind);
 
   /// Destructor, empty.
   virtual ~BarycentricBaseMapping() = default;
```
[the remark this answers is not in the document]
boris-martin: What additional flag did you use to get this warning ?

src/mapping/Mapping.hpp
DavidSCN: > It feels safer to me than guessing from the connectivity. 

Sounds reasonable

>  What about IntegralType

Maybe `IntegralManifold`? Anyway, feel free to decide on something. It should be as self explanatory as possible.

src/mapping/Mapping.cpp
```diff
@@ -9,16 +9,19 @@ namespace precice {
 namespace mapping {
 
 Mapping::Mapping(
-    Constraint constraint,
-    int        dimensions,
-    bool       requireGradient)
```
DavidSCN, line 14: The initialize order still seems messy, or is this a false alert from the CI? The argument order  of the constructor needs to match the order of the class members, correct?

boris-martin: I kept the order of the arguments instead of the declaration order, then moved dimensions earlier, I'll clean this up

src/mapping/Mapping.hpp
uekerman: IMO, `IntegralType` should do be understandable here.
In the end "volume vs. surface" only makes a difference for mapping methods that make use of integrals in some sense.
But why is this variable only used for "scaled-consistent mapping", shouldn't it also have an influence on the nearest-projection mapping. Do we at least have safeguard assertions there?

boris-martin: NP mapping should be Surface-only, LCI mapping Volume-only, and all the others (NN(G) and RBFs) could be both.
My idea was to have no safeguard internally but safeguards on the API:
- If In precice-config I put volume or surface with NN, _alles gut_
- If I configure NP with volume or LCI with surface, throw an error
- If I configure NP with surface or LCI with volume, either do nothing or warn that configuration wasn't needed.

But in the internals, the scaling is done in the mapping base class, so putting safeguards there would be ugly.

boris-martin: But I don't get what you meant by  "shouldn't it also have an influence on the nearest-projection mapping". The point is that it's one of the few mappings where it shouldn't.

uekerman: Mmh, I thought you would put `LCI` in the config and deduce from this that it is a volume coupling.
In the LCI mapping class I would then put an "assertion(type==surface)" to safeguard against future changes / wrong unit tests etc.

uekerman:
@boris-martin and me briefly discussed this PR: It should be easier to simply add an additional constraint `scaled-consistent-volume` and eventually rename `scaled-consistent` to `scaled-consistent-surface` (breaking).
Then, we save all the logic tests like "If *NP* configured, but *volume* specified ..."
Will come in a separate PR.

PR #1343, opened 2022-06-28
Make calling SolverInterface::initializeData() mandatory

### LINKED ISSUE #1196: Make calling initializeData mandatory?

BenjaminRodenberg:
## Main changes of this PR

Closes #1196.

See #1196. 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -336,13 +336,11 @@ double SolverInterfaceImpl::initialize()
   }
 
   // Needs to be removed for test Integration/Serial/Whitebox/TestExplicitWithDataScaling
-  /**
   if (_couplingScheme->hasDataBeenReceived()) {
     performDataActions({action::Action::READ_MAPPING_PRIOR}, 0.0, 0.0, 0.0, dt);
     mapReadData();
     performDataActions({action::Action::READ_MAPPING_POST}, 0.0, 0.0, 0.0, dt);
   }
-  */
 
   PRECICE_INFO(_couplingScheme->printCouplingState());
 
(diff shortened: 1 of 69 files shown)


fsimonis:
At the end of the test `Integration/Parallel/CouplingOnLine`, the SolverInterface `finalize()`es.
While cleaning the m2ns, the IntraComm tries to broadcast something, but its communicator doesn't exist anymore.
Hence it runs into an assertion.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -494,7 +498,7 @@ void SolverInterfaceImpl::finalize()
   Event                    e("finalize"); // no precice::syncMode here as MPI is already finalized at destruction of this event
   utils::ScopedEventPrefix sep("finalize/");
 
-  if (_state == State::Initialized) {
```
fsimonis, line 497: This has to be both `State::Initialized` and `State::InitializedData`.
We establish communication channels in `initialize()`

BenjaminRodenberg: Locally I added some more checks at the beginning of `finalize()`:
```
void SolverInterfaceImpl::finalize()
{
  PRECICE_TRACE();
  PRECICE_CHECK(_state != State::Finalized, "finalize() may only be called once.")
  PRECICE_CHECK(_state != State::Initialized, "initializeData() has to be called before finalize().");
  PRECICE_CHECK(_state == State::InitializedData, "initializeData() has to be called before advance().")
  ...
```
This asserts that `initialize()` and `initializeData()` have been called before `finalize()`. This is an additional restriction that I would like to introduce here, because I don't really see a use-case for calling `finalize()` without `initialize()`. Does anybody know why we currently support this?

BenjaminRodenberg: Fixed it in [4ff2cbe](https://github.com/precice/precice/pull/1343/commits/4ff2cbe7d7e0ba3b89009ae32274a66295b58bcb).

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -334,11 +335,14 @@ double SolverInterfaceImpl::initialize()
...
   }
 
+  // Needs to be removed for test Integration/Serial/Whitebox/TestExplicitWithDataScaling
+  /**
   if (_couplingScheme->hasDataBeenReceived()) {
     performDataActions({action::Action::READ_MAPPING_PRIOR}, 0.0, 0.0, 0.0, dt);
     mapReadData();
     performDataActions({action::Action::READ_MAPPING_POST}, 0.0, 0.0, 0.0, dt);
   }
+  */
```
BenjaminRodenberg, line 345: Commenting out this part of the code actually makes this PR breaking.

BenjaminRodenberg (commented):
I think we could also implement this in a non-breaking way by replacing `PRECICE_CHECK` with `PRECICE_WARN`. I highlighted the place where a more elaborate treatment would be necessary (I have to check this tomorrow).

We can easily check whether it is really non-breaking by just reverting all the tests to the state from `develop`. It should still work.

Open question:
* If we make `initializeData` mandatory, a warning (if `initializeData` was not called) is the right way to go.
* If we merge `initialize` and `initializeData`, we should instead raise a warning, if `initializeData` is called and just perform everything that is required directly in `initialize`.

Did we already conclude on one of the two options in #1196?

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRodenberg: I played around with this part a bit yesterday: If I comment it out, calling `initializeData()` in every test is unproblematic, but if I revert some tests to the state from `develop`, then they break. There might be a way to make this PR non-breaking, but I think this is complicated and not really worth it. Therefore, I will directly build up on this PR and merge `initialize()` and `initializeData()` now. This is also related to #1349 and https://github.com/precice/precice/pull/1029 (waveform and serial coupling).

BenjaminRodenberg:
Closing in favor of #1350

PR #1344, opened 2022-06-28
Add option to disable precice-tools building

DavidSCN:
## Main changes of this PR
Introduces a cmake option to disable the building of `precice-tools`

Building them might be undesired. The default flag (`ON`) delivers fully backward compatibility.

## Author's checklist

* [x] Check handling of install targets

## Reviewers' checklist


fsimonis:
Is there any reason not to build it? Is has no additional dependencies and takes about a second to compile and link.

If a user can always selectively compile libprecice using `make precice`.

fsimonis (approved):
Still, it could come in handy in case the sym linking doesn't work or similar. So let's keep it.

fsimonis:
I don't think this will need a changelog

PR #1345, opened 2022-06-29
Add dependabot to update package versions in GitHub actions

DavidSCN:
## Main changes of this PR
Adds dependabot to update package versions in our GitHub actions.

Have al look at [Keeping your actions up to date with dependabot](https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot). This way, dependabot will create a PR in case packages get outdated.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,11 @@
+# Set update schedule for GitHub Actions
+
+version: 2
+updates:
+
+  - package-ecosystem: "github-actions"
+    directory: "/"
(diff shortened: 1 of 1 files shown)


fsimonis:
@MakisH This is actually quite useful. We already used some docker-related actions that were discontinued in the past and silently stopped working without realizing it.

fsimonis (approved):

PR #1348, opened 2022-06-30
Improve compile times

fsimonis:
## Main changes of this PR

Slightly improves compile times per TU especially in tests.

## Author's checklist

## Reviewers' checklist


MakisH:
Interesting! For approximately how much difference are we talking here?

fsimonis:
It's about 2 minutes in the CI.  So 10% overall.

There are plenty more const saving opportunities in the test. The downside is that these come at the cost of having to include additional headers whenever they are needed.

Issue #1347, opened 2022-06-30
Fix and enable first-participant timestepping combined with communication of initial data


uekerman:
Currently, we get a problem if we use "first-participant" as timestepping method and, at the same time, initialize any data.
Sadly, we cannot fix this in a non-breaking way. Most promising solution IMO is to merge `initialize` and `initializeData`, see also #1196 .  

An integration test that reproduces the problem is available in a draft PR: #1349 

For a detailed description of the problem and the potential fix, see the discussion in #1307, as a partial fix especially https://github.com/precice/precice/pull/1307/commits/62e34179ec0d852f5eba8249724dd6bb4d95b5db.

davidscn:
@uekerman was this fixed in the course of the current development?

uekerman:
yes

PR #1349, opened 2022-07-01
Add failing test to reproduce first-participant + initData problem

uekerman:
## Main changes of this PR

Adds the remaining failing test from #1307 
Needs breaking changes to be fixed, see also #1347

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/tests/serial/time/explicit/serial-coupling/ReadWriteScalarDataFirstParticipantInitData.cpp
+++ b/tests/serial/time/explicit/serial-coupling/ReadWriteScalarDataFirstParticipantInitData.cpp
@@ -1,0 +1,80 @@
+#ifndef PRECICE_NO_MPI
+
+#include "testing/Testing.hpp"
+
+#include <precice/SolverInterface.hpp>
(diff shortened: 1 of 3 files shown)


BenjaminRodenberg:
I just tried out merging #1350 @ [e8f738d](https://github.com/precice/precice/pull/1350/commits/e8f738d57e09d17dc9540ae77d2ad190f2df00f8) into this PR and the issue is resolved. I will not push the merge, because the changes are trivial and it just makes the history more complicated.

BenjaminRodenberg:
@uekerman what should we do with this PR? I think we can merge it into #1366. We can still keep the issue open, but it will be resolved in preCICE v3.0.0.

uekerman:
> What should we do with this PR? I think we can merge it into https://github.com/precice/precice/pull/1366. We can still keep the issue open, but it will be resolved in preCICE v3.0.0.

Yes, good point. I rebased on `develop-v3.0.0` and changed the target branch.

PR #1222, opened 2022-03-26
Remove mapWriteDataFrom and mapReadDataTo

BenjaminRodenberg:
## Main changes of this PR

Removes API functions `mapWriteDataFrom` and `mapReadDataTo` and all corresponding tests. 

Closes #859.

## Motivation and additional information

These functions do not have a use-case and make implementation more complicated. See #859.

## Author's checklist

* [x] Merge #1205 (changes already included here).
* [x] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [x] I ran `make format` to ensure everything is formatted correctly.
* [x] I sticked to C++14 features.
* [x] I sticked to CMake version 3.10.
* [x] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

PR #1350, opened 2022-07-01
Merge initializeData() into initialize()

### LINKED ISSUE #1196: Make calling initializeData mandatory?

BenjaminRodenberg:
## Main changes of this PR

Closes #1196 by performing data initialization directly in `initialize`. Removes `initializeData()`.

See #1196.
Also helps to close related issue #1349.

## Todo

- [x] merge #1222 
- [x] merge #1362 

## Author's checklist

## Reviewers' checklist


src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -387,7 +361,9 @@ void SolverInterfaceImpl::initializeData()
   _accessor->exportFinal();
   solverInitEvent.start(precice::syncMode);
 
-  _hasInitializedData = true;
+  _state = State::Initialized;
+
+  return _couplingScheme->getNextTimestepMaxLength();
```
fsimonis, line 366: This should be the same as `dt`, correct?
Maybe you could make the `dt` `const` to make this safer.

```suggestion
  return dt;
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -405,12 +381,14 @@ double SolverInterfaceImpl::advance(
   Event                    e("advance", precice::syncMode);
   utils::ScopedEventPrefix sep("advance/");
 
-  PRECICE_CHECK(_state != State::Constructed, "initialize() has to be called before advance().");
+  PRECICE_CHECK(_state != State::Constructed, "initialize() and initializeData() have to be called before advance().");
+  if (_state == State::Initialized) {
+    PRECICE_WARN("initializeData() should be called before advance(). This will become mandatory in preCICE 3.0.0");
+  }
```
fsimonis, line 387: This is obsolete due to the merge, right?

```suggestion
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -548,7 +526,11 @@ int SolverInterfaceImpl::getDimensions() const
 bool SolverInterfaceImpl::isCouplingOngoing() const
 {
   PRECICE_TRACE();
-  PRECICE_CHECK(_state != State::Constructed, "initialize() has to be called before isCouplingOngoing() can be evaluated.");
+  PRECICE_CHECK(_state != State::Constructed, "initialize() and initializeData() have to be called before isCouplingOngoing() can be evaluated.");
+  if (_state == State::Initialized) {
+    PRECICE_WARN("initializeData() should be called before isCouplingOngoing(). This will become mandatory in preCICE 3.0.0");
+  }
+  //PRECICE_CHECK(_state != State::Initialized, "initializeData() has to be called before isCouplingOngoing() can be evaluated.");
```
fsimonis, line 533:
```suggestion
  PRECICE_CHECK(_state == State::Initialized, "initialize() has to be called before isCouplingOngoing() can be evaluated.");
```

fsimonis:
Doesn't run/compile due to:
* initialize data calls in solverdummies
* isActionRequired requiring the state to be initialized

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRodenberg: Yes. Cleaned it up :+1:

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRodenberg: Done.

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRodenberg: This is a small problem (and I'm not sure whether there is an easy solution): If the participant-first method is used for setting the `dt`, we have to call this function twice in initialize: First before all data exchange (because we might need `dt` already for the actions) and then after the initial data exchange (because we might have received a `dt` from first). I updated the source code a bit, but there are still two calls of `_couplingScheme->getNextTimestepMaxLength()`.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -324,70 +327,41 @@ double SolverInterfaceImpl::initialize()
...
   double dt = _couplingScheme->getNextTimestepMaxLength();
 
   performDataActions({action::Action::WRITE_MAPPING_PRIOR}, 0.0, 0.0, 0.0, dt);
   mapWrittenData();
   performDataActions({action::Action::WRITE_MAPPING_POST}, 0.0, 0.0, 0.0, dt);
 
-  _couplingScheme->initializeData();
+  PRECICE_DEBUG("Initialize coupling schemes");
+  _couplingScheme->initialize(time, timeWindow);
+  dt = _couplingScheme->getNextTimestepMaxLength(); // dt can change at this point again, if second participant and participant-first method for setting time window size is used.
```
fsimonis, line 344: Does `dt` for the write mapping of initialize data even make sense at all?

At this point, there hasn't been any time-step and the time-window size isn't known either. 

Why don't we simply pass `0.0` or an invalid value like `-1` instead?
This would allow the user to distinguish between normal and init data actions too.
This would simplify some things.

BenjaminRodenberg: I'm not very deep in the "actions-business". Currently there is clearly a design issue, if write actions are used with the participant first method, because the second participant just cannot know the `dt` at this point in time. If `dt` is given in the config - and therefore fixed - there is no problem. I'll change it to `0.0` for now, feel free to change or comment, if something else looks more reasonable.

BenjaminRodenberg: I changed the `dt` to `0.0`. This is dangerous, because of the `ScaleByDt` action, which divides by `dt`...
I also opened an issue #1358.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -324,70 +327,39 @@ double SolverInterfaceImpl::initialize()
...
   PRECICE_DEBUG("Plot output");
   _accessor->exportFinal();
   solverInitEvent.start(precice::syncMode);
 
-  _hasInitializedData = true;
+  _state = State::Initialized;
+  PRECICE_INFO(_couplingScheme->printCouplingState());
+
+  return _couplingScheme->getNextTimestepMaxLength();
+  ;
```
uekerman, line 362:
```suggestion
```

src/precice/impl/SolverInterfaceImpl.cpp
uekerman: Could you please add this information as comments?

src/precice/impl/SolverInterfaceImpl.cpp
uekerman: I think we can simply remove the argument, see my comment in #1358.
So "0.0" is OK here. Both "breaks" should go into v3.0 anyway.

BenjaminRodenberg:
> My gut feeling tells me that we could get rid of `hasInitialDataBeenReceived`, but cover the need of this flag with `hasDataBeenReceived`?

We need to differentiate between initial data and "normal" data when using waveform interpolation in #1352. I think it also helps with readability to have a dedicated function here. My opinion: Let's keep it.

src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -248,6 +222,8 @@ class BaseCouplingScheme : public CouplingScheme {
    */
   virtual bool hasAnySendData() = 0;
 
+  virtual void determineInitialDataExchange() = 0;
```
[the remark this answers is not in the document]
BenjaminRodenberg: Done: 6865868d0061d632e3cc1846bbf1a46e60fa814f

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRodenberg: Added some comments here: 6865868d0061d632e3cc1846bbf1a46e60fa814f.

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRodenberg: Done. 6865868d0061d632e3cc1846bbf1a46e60fa814f

BenjaminRodenberg:
#1362 has simplified this PR a lot. @uekerman do you want to take another look? From my perspective this PR is ready to merge.

BenjaminRodenberg:
@uekerman: Should we generally remove the restriction that first cannot initializeData already here? See https://github.com/precice/precice.github.io/pull/174#issuecomment-1179517031. I think this might simplify code and the API furtherly.

src/cplscheme/SerialCouplingScheme.cpp
```diff
@@ -106,11 +90,11 @@ void SerialCouplingScheme::exchangeInitialData()
       // The second participant sends the initialized data to the first participant
       // here, which receives the data on call of initialize().
       sendData(getM2N(), getSendData());
-      receiveAndSetTimeWindowSize();
-      // This receive replaces the receive in initialize().
-      receiveData(getM2N(), getReceiveData());
-      checkDataHasBeenReceived();
     }
+    receiveAndSetTimeWindowSize();
+    // This receive replaces the receive in initialize().
```
uekerman, line 95: We can probably remove this comment.

uekerman (approved):
Good to go

uekerman:
> @uekerman: Should we generally remove the restriction that first cannot initializeData already here? See https://github.com/precice/precice.github.io/pull/174#issuecomment-1179517031. I think this might simplify code and the API furtherly.

No, let's do this separately. Can't overview it yet.

src/cplscheme/SerialCouplingScheme.cpp
BenjaminRodenberg: Done 236449928392a750510803a148b1202ec8b6520a.

Issue #1304, opened 2022-05-25
Incorrect projections in the fallback of cell interpolation


fsimonis:
**Describe your setup**

preCICE Version: 2.4.0 + #1297 

**Describe the problem**

When using linear cell interpolation, there may be edge cases where a point is numerically outside a cell. In that case the mapping find the nearest projection to determine the best projection possible.

Currently, this leads to unexpected results.
In 2D, all edges are checked for a match before the vertices.
This leads to the logic picking edges that are further away instead of picking vertices that are closer.

The following image displays the behaviour of the using cell-interpolation in 2D with a defined triangle ABC.
The coloured areas define which primitive a location is mapped to.

![WrongProjection](https://user-images.githubusercontent.com/13552216/170236216-0274cad8-9968-4217-aedf-48607aee2931.svg)

A concrete example is the right red AB section, where the correct projection is to map to the closest point C. Instead, the projection picks the edge AB, as it results in an interpolation.

**Expected behaviour**
The correct behaviour would be to pick the closest primitive as such:

![CorrectProjection](https://user-images.githubusercontent.com/13552216/170236215-38b2b90d-4509-4a2e-9e90-4e7673150c91.svg)

**Possible solution**

Currently, the projection implementation uses a lazy evaluation approach. The closes edge resulting in a valid projection (interpolation) will be picked completely skipping the evaluation of the nearest vertex.

The correct implementation would be to fetch the closest interpolations of all primitive classes, then pick the most suitable one.
The downside is that this completely disables the lazy evaluation of index trees as all primitives of the entire Mesh always need to be indexed.

**Additional context**

@boris-martin stumbled across this in #1297

boris-martin:
Same problem occurs in NP mapping in 3D by the way. Same configuration with everything in the same plane would yield similar results.

boris-martin:
> The downside is that this completely disables the lazy evaluation of index trees as all primitives of the entire Mesh always need to be indexed.

I don't really see how big is the problem. Overall, wouldn't it roughly multiply the complexity by only 2 ? (Index all edges and vertices instead of just all edges)

fsimonis:
> I don't really see how big is the problem. Overall, wouldn't it roughly multiply the complexity by only 2 ? (Index all edges and vertices instead of just all edges)

This changes the best-case to the worst-case of the projection logic, which also impacts the nearest-projection mapping.

[...]

PR #1351, opened 2022-07-01
Changed NP behavior for some edge cases

boris-martin:
## Main changes of this PR

When doing nearest-projection (including the fall-back of linear cell interpolation), we check that the projection we use is closer than the nearest neighbor. If not, we use the NN instead. See #1304 for context.

#### PreferTriangleOverEdge

Consider an edge starting at the origin and moving in the y direction (from (0,0,0) to (0,2,0)) and a triangle living in the plane x = 3. We want to project the point (1,1,1). Before this PR, preCICE tries to map to the triangle, succeeds and thus doesn't check the edge.
Now, we project into the triangle, see that the projection is further than the nearest neighbor vertex (one of the edge), so we decide to reject it. Then we fall back on edges and accept the projection on that edge. **An edge was preferred to a triangle.**

#### AvoidClosestTriangle

Consider two disconnected triangles and a vertex such that only one triangle contains its projection, but it's the furthest of the triangles. Before this PR, we favored interpolation on triangle. Now, we reject this triangle because the projection is further than the nearest neighbor (a vertex from the other triangle). Since the other triangle yields an extrapolation, we reject it, fall-back on edges (there are none) then on vertices. **The far triangle was replaced by the nearest vertex of the other triange**

#### LinearCellInterpolation unit test
See https://github.com/precice/precice/issues/1304: a point fromthe lower-left "AC" region now maps to B, the nearest neighbor.

### 

- Current implementation is a bit suboptimal in 3D: when finding triangle projection, we compare it to NN, and if we fall-back on edges, the NN is recomputed. We could refactor this to gain efficiency at the cost of readability. Let's first discuss the concept.
- Most edge cases happen in volume coupling. If we don't want to change NP for some reason (backward compatibility etc) we could design an "alternative NP" that is only used for the fall-back of LCI. 

Typical example in volume coupling: round domain with different discretizations.
image
The vertex can't be mapped to the nearest edges (it'd be an extrapolation) so it gets mapped to the *opposite* edge.
(In practice, the pentagon would be triangulated and an edge inside it would probably match too)

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/query/Index.cpp
+++ b/src/query/Index.cpp
@@ -217,7 +217,7 @@ std::vector<VertexID> Index::getVerticesInsideBox(const mesh::Vertex &centerVert
   auto coords    = centerVertex.getCoords();
   auto searchBox = query::makeBox(coords.array() - radius, coords.array() + radius);
 
-  const auto &          rtree = _pimpl->getVertexRTree(*_mesh);
+  const auto           &rtree = _pimpl->getVertexRTree(*_mesh);
   std::vector<VertexID> matches;
   rtree->query(bgi::intersects(searchBox) and bg::index::satisfies([&](size_t const i) { return bg::distance(centerVertex, _mesh->vertices()[i]) <= radius; }),
                std::back_inserter(matches));
@@ -228,7 +228,7 @@ std::vector<VertexID> Index::getVerticesInsideBox(const mesh::BoundingBox &bb)
 {
   PRECICE_TRACE();
   // Add tree to the local cache
-  const auto &          rtree = _pimpl->getVertexRTree(*_mesh);
+  const auto           &rtree = _pimpl->getVertexRTree(*_mesh);
   std::vector<VertexID> matches;
   rtree->query(bgi::intersects(query::makeBox(bb.minCorner(), bb.maxCorner())), std::back_inserter(matches));
   return matches;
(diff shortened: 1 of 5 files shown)


fsimonis:
The problem is not as dramatic as it seems to be. As both meshes discretize the same geometry, the vast majority of points will end up finding the correct nearest primitive.
The primary issue are nearly matching meshes which are shifted outside the domain.

The following image features a geometry (gray) and a discretization (blue). Vertices located inside the red cones are the ones that currently do not provide correct results. As they will not map on the closest edges.


image

fsimonis:
@boris-martin Are you still planning to work on this PR? If not, please unassign and we will take over. :smiley:

boris-martin:
> @boris-martin Are you still planning to work on this PR? If not, please unassign and we will take over. smiley

Not really. Feel free to ask for reviews afterwards though!

src/query/Index.cpp
```diff
@@ -299,16 +299,21 @@ ProjectionMatch Index::findEdgeProjection(const Eigen::VectorXd &location, int n
...
     }
   }
-  // Pick the smallest candidate by distance
+
+  // Could not find edge projection element, fall back to vertex projection
+  if (candidates.empty()) {
+    return closestVertex;
+  }
+
+  // Prefer the closet vertex if it closer than the closest edge
```
davidscn, line 308:
```suggestion
  // Prefer the closest vertex if it closer than the closest edge
```

davidscn (approved):
Looks sane.

fsimonis:
Changelog still missing

davidscn:
Ready to merge, guessing that @boris-martin has no final concerns.

boris-martin:
Alles gut!

PR #1352, opened 2022-07-03
Waveform interpolation, read direction for serial implicit coupling

### LINKED ISSUE #1174: Allow to read data from interpolant in time

In #133 the idea of waveform relaxation is drafted. By providing an interpolant instead of a single sample per window higher-order time stepping and more accurate subcycling becomes possible. This issue describes a substep of this major feature: the implementation of the read direction.

preCICE allows to store one sample of data per time window. This sample is currently overridden in every (Quasi-Newton) iteration by the updated sample. If a window is finished an one moves on to the next window, the last sample is used as an initial guess (if no extrapolation is used). 

To support waveform relaxation, preCICE has to create an interpolant and update it based on the provided data in each iteration. As a first step, we want to create a simple linear interpolant from the data at the beginning and end of the time window. Higher order interpolation is generally possible, but out of scope for this issue.

In order to design this feature in a non-breaking way, the following steps have to be performed:

* Add an additional (optional!) argument for the API functions `readBlockScalarData(...)` etc. This argument allows the user to specify where the interpolant has to be sampled.
* Extend the configuration file with options to specify interpolation order.
* Find a consistent treatment for `initializeData`. Also for serial coupling both participants should be able to use `initialize=on`.
* Find a consistent treatment for guards using `precice.isReadDataAvailable()`, if this feature is enabled.
* Use `experimental` tag.

Some important restrictions:

* Only consider implicit coupling, because explicit coupling would require extrapolation.
* Only allow first and zeroth order interpolation, because higher order interpolation requires additional samples (either from past windows or more data in the current window)

**Additional context**

A large part of the work required for this issue is already implemented in #1029. 
#1171 is to some degree a "sibling" of this issue. #133 is the parent of this issue.
Please refer to https://github.com/precice/precice/projects/7 for more related issues.
https://github.com/precice/precice.github.io/pull/109 already provides preliminary documentation for this feature.

BenjaminRodenberg:
## Main changes of this PR

Implements waveform interpolation for serial coupling. Related to #1187 

This PR in combination with #1187 closes #1174.

See #1029. Implements waveform interpolation for serial coupling. Main obstacle here is to allow the second participant to receive two pieces of data in `initialize()`: 1) the initial data from first and 2) the result from the first time window of first send by calling `advance`.

## Todos

* [x] Merge #1350 
* [x] Merge #1362 
* [x] Merge https://github.com/precice/precice/pull/1222

## Author's checklist

## Reviewers' checklist


src/cplscheme/BiCouplingScheme.cpp
```diff
@@ -121,5 +127,51 @@ CouplingData *BiCouplingScheme::getReceiveData(
...
+    checkDataHasBeenReceived();
+  } else { // second participant
+    if (isImplicitCouplingScheme()) {
+      PRECICE_DEBUG("Perform acceleration (only second participant)...");
+      convergence = doImplicitStep();
+      sendConvergence(getM2N(), convergence);
+    }
+    PRECICE_DEBUG("Sending data...");
+    sendData(getM2N(), getSendData());
+  }
```
uekerman, line 172: Where did the `receiveData` go here? Let's not hide this in `doImplicitStep()` (if that's the case)
How can this function be the same for both serial and parallel coupling?

src/cplscheme/CouplingScheme.hpp
```diff
@@ -68,35 +68,29 @@ class CouplingScheme {
...
-   * initialization is necessary. Then it is simply skipped.
-   * It has to be called after initialize() and before
-   * advance(). The second participant has to write the initial data values
-   * to preCICE after initialize() and before initializeData().
+   * Inside SolverInterface::initialize() one has to differentiate between received initial data and received data from
+   * advance. This difference is unimportant as soon as SolverInterface::initialize() and, therefore, cleaned up
+   * by this function.
    */
-  virtual void initializeData() = 0;
+  virtual void completeInitialization() = 0;
```
uekerman, line 90: ... as soon as what? Sth is missing here.

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -984,23 +980,32 @@ void CouplingSchemeConfiguration::addDataToBeExchanged(
     if (from == accessor) {
       scheme.addDataToSend(exchange.data, exchange.mesh, requiresInitialization);
       if (requiresInitialization && (_config.type == VALUE_SERIAL_EXPLICIT || _config.type == VALUE_SERIAL_IMPLICIT)) {
-        PRECICE_CHECK(not scheme.doesFirstStep(),
-                      "In serial coupling only second participant can initialize data and send it. "
-                      "Please check the <exchange data=\"{}\" mesh=\"{}\" from=\"{}\" to=\"{}\" initialize=\"{}\" /> tag in the <coupling-scheme:... /> of your precice-config.xml.",
-                      dataName, meshName, from, to, requiresInitialization);
+        if (_experimental && scheme.doesFirstStep()) {
+          PRECICE_WARN("In serial coupling first participant can only initialize data, if the experimental interface is active. This is only useful for time interpolation of order 1 or higher. Please consider checking the documentation for more details.");
```
uekerman, line 984: I don't understand the first sentence. Could you rewrite?

src/cplscheme/CouplingScheme.hpp
```diff
@@ -68,35 +68,29 @@ class CouplingScheme {
...
+   *
+   * This is only relevant for the second participant of the SerialCouplingScheme, because other coupling schemes only
+   * receive initial data in initialize. This part is implemented as a public function and has to be called explicitly
+   * from the outer scope (usually SolverInterfaceImpl), because the outer scope has to additional actions between
+   * CouplingScheme::initialize and CouplingScheme::receiveResultOfFirstAdvance
    */
-  virtual bool receivesInitializedData() const = 0;
+  virtual void receiveResultOfFirstAdvance(){
+      //@todo: Add security checks?
+  };
```
uekerman, line 81: No big fan of this function. We should carefully validate whether we need these actions between both calls.

src/cplscheme/BiCouplingScheme.cpp
BenjaminRodenberg: I reverted this refactoring. Should be clearer now.

src/cplscheme/CouplingScheme.hpp
BenjaminRodenberg: I now entirely removed this function by merging #1362.

src/cplscheme/CouplingScheme.hpp
BenjaminRodenberg: Done 52ae6c9e6ad74e84f7d47faef1289e094e4d1319

src/cplscheme/config/CouplingSchemeConfiguration.cpp
BenjaminRodenberg: I think we can just remove this warning. If the experimental interface is not activated the user gets an error anyway (we might also soon move the waveform features from the experimental interface to the real interface anyway). Removed it in c985cb315162b305a73efbe7575ba334d7dde281

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -984,18 +980,22 @@ void CouplingSchemeConfiguration::addDataToBeExchanged(
...
-        PRECICE_CHECK(scheme.doesFirstStep(),
-                      "In serial coupling only first participant can receive initial data. "
-                      "Please check the <exchange data=\"{}\" mesh=\"{}\" from=\"{}\" to=\"{}\" initialize=\"{}\" /> tag in the <coupling-scheme:... /> of your precice-config.xml.",
-                      dataName, meshName, from, to, requiresInitialization);
+        if (not _experimental) {
+          PRECICE_CHECK(scheme.doesFirstStep(),
+                        "In serial coupling only first participant can receive initial data. "
+                        "Please check the <exchange data=\"{}\" mesh=\"{}\" from=\"{}\" to=\"{}\" initialize=\"{}\" /> tag in the <coupling-scheme:... /> of your precice-config.xml. You can bypass this restriction by activating the experimental interface for time interpolation. Please refer to the documentation for more details.",
+                        dataName, meshName, from, to, requiresInitialization);
+        }
```
BenjaminRodenberg, line 998: This part is related to #1367

BenjaminRodenberg (commented):
From my perspective ready to review. The only part that might be something we want to deal with before merging this PR is the one related to #1367, where I put a comment.

docs/changelog/1352.md
```diff
@@ -0,0 +1 @@
+- Allow waveform interpolation also for serial coupling. A serial coupling scheme now allows interpolation for both participants.
```
uekerman, line 1:
```suggestion
- Add waveform interpolation also for serial coupling. A serial coupling scheme now allows interpolation for both participants.
```

src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -504,6 +506,11 @@ class BaseCouplingScheme : public CouplingScheme {
    */
   virtual void exchangeInitialData() = 0;
 
+  virtual void performReceiveOfFirstAdvance()
+  {
+    return;
+  }
```
uekerman, line 512: Please add a comment here.

src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -186,6 +186,8 @@ class BaseCouplingScheme : public CouplingScheme {
    */
   void initialize(double startTime, int startTimeWindow) override final;
 
+  void receiveResultOfFirstAdvance() override final;
```
uekerman, line 189:
```suggestion
  /// Receives result of first advance, if this has to happen inside SolverInterface::initialize(), see CouplingScheme.hpp
  void receiveResultOfFirstAdvance() override final;
```

src/cplscheme/tests/DummyCouplingScheme.hpp
```diff
@@ -37,6 +37,14 @@ class DummyCouplingScheme : public CouplingScheme {
       double startTime,
       int    startTimesteps) override final;
 
+  /**
+   * @brief Not implemented.
+   */
+  void receiveResultOfFirstAdvance()
```
uekerman, line 43:
```suggestion
  void receiveResultOfFirstAdvance() override final
```
Please double-check this suggestion, don't know the C++ rules here by heart.

uekerman (approved):
Very clean and readable implementation, nice job :+1:
(I did not check all details of the integration tests)

src/cplscheme/BaseCouplingScheme.hpp
BenjaminRodenberg: Done. e0384bdb4db7e66d5bb3cb6252e31ab37f83821b.

PR #1359, opened 2022-07-06
Add pre-commit hooks

fsimonis:
## Main changes of this PR

This PR adds pre-commit hooks to preCICE including:

* formatting of code using the correct version of clang-format
* formatting of [preCICE config files](https://github.com/precice/precice-pre-commit-hooks)
* checks for merge conflict markers
* checks for mixed line-endings in files
* checks for incorrect XML files (ignores namespaces :wink:)

This PR replaces the code check workflow and doesn't rely on https://pre-commit.ci/

Massive shoutout to @renefritze for showing us this tool!

Closes #1291

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,15 +1,18 @@
 repos:
+# Official repo for the clang-format hook
 - repo: https://github.com/pre-commit/mirrors-clang-format
   rev: 'v8.0.1'
   hooks:
   - id: clang-format
     exclude: "^thirdparty"
+# Official repo for default hooks
 - repo: https://github.com/pre-commit/pre-commit-hooks
   rev: 'v4.3.0'
   hooks:
   - id: check-xml
   - id: check-merge-conflict
   - id: mixed-line-ending
+# Custom repo for the preCICE configuration formatter
 - repo: https://github.com/precice/precice-pre-commit-hooks
   rev: 'v2.1'
   hooks:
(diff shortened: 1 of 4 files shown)


.pre-commit-config.yaml
```diff
@@ -0,0 +1,16 @@
+repos:
```
MakisH, line 1: Where's the docs? :grin: 

To document/steps:
- This is a convenience tool, not a safety measure: you need to explicitly install pre-commit if you want Git to warn you before you commit.
- Install the tool: `pip3 install pre-commit`
- Install the hook: `pre-commit install`

When committing something good, it shows the following output:
```
~/github/precice [add-pre-commit]$ git commit
clang-format.............................................................Passed
check xml............................................(no files to check)Skipped
check for merge conflicts................................................Passed
mixed line ending........................................................Passed
format preCICE configuration files...................(no files to check)Skipped
[add-pre-commit f886de24] test
 1 file changed, 1 insertion(+), 1 deletion(-)
```

MakisH (approved):
I tried it locally and it works. I also understand the code changes. I don't think we need to add any other checks at the moment.

.pre-commit-config.yaml
fsimonis: It's not only a convenience tool, you can use it to format your files using the correct version of clang-format, which is super useful if your distro doesn't provide it.

renefritze: Just FYI, this is thanks to a binary wheel distribution of clang-format via https://github.com/ssciwr/clang-format-wheel

fsimonis:
I extended the formatting readme. Problem is that there is no place for this doc in the source-itself.
I'll add it to the developer documentation of the website.

PR #1363, opened 2022-07-07
Wrong RTree distance computation

boris-martin:
## Main changes of this PR

Added a 3D test with a triangle in the plane z = 0 and a point above it. `getClosestTriangles` should return a match with a distance that equals the z coordinate of the point, but it returns 0.
PR is a draft until a fix is found ?

## Author's checklist

## Reviewers' checklist


DavidSCN:
@boris-martin this was fixed by #1395, right? Do you still want to add the test or is it now already included?

fsimonis:
This test is now obsolete as the distances are not calculated using boost anymore.

PR #1366, opened 2022-07-09
Break develop for v3.0.0

### LINKED ISSUE #834: Check unconditional read mapping in initializeData
`initializeData()` contains an unconditional write mapping. Why?

https://github.com/precice/precice/blob/98b09c8fd184726530c33b23d6400d4ff1cf6d02/src/precice/impl/SolverInterfaceImpl.cpp#L336-L338

In `advance` we have such a condition

https://github.com/precice/precice/blob/a53645cc7bc2ad0a9abba51a1aecbca64aa4022a/src/precice/impl/SolverInterfaceImpl.cpp#L411-L415

Connected to this is the question whether `initializeData` should be called even if no data needs to be initialized (we currently explain it that way in the adapter example)? If yes, how do we treat the mappings there then?

Related to #824

### LINKED ISSUE #859: Remove deprecated API functions mapWriteDataFrom and mapReadDataTo

### LINKED ISSUE #934: ScaleByDtAction with SCALING_BY_DT is misleading

### LINKED ISSUE #1174: Allow to read data from interpolant in time

### LINKED ISSUE #1196: Make calling initializeData mandatory?

### LINKED ISSUE #1223: Use-case for isReadDataAvailable and isWriteDataRequired?
I am not sure that the functions `isReadDataAvailable` and `isWriteDataRequired` are actually needed. Removing them is clearly breaking, but would also help us to simplify the codebase and the API. This issue is meant for discussing the use-case of these function and to find places where these functions are actually used and needed. I would like to find out whether there are actually any adapters or other user codes that really depend on these functions?

[In our documentation](https://precice.org/couple-your-code-timestep-sizes.html) we say the following about the use-case of these two functions:

> If a participant subcycles it is actually not necessary to write data to or read data from preCICE. To avoid unnecessary calls, preCICE offers two optional helper functions:
>
> bool isReadDataAvailable () const;
> bool isWriteDataRequired (double computedTimestepLength) const;

With the current developments around waveform relaxation and subcycling, I don't think that this use-case is relevant anymore:

* If an interpolant is created by preCICE to allow for subcycling, reading data is allowed at any point in time. preCICE cannot know whether the data the user is requesting is new data or the user performs unnecessary read data calls. Therefore, I would suggest to remove `isReadDataAvailable` and let the user decide whether calling a `readData` function makes sense. (these are the current plans for the "read direction" of waveform relaxation, see https://github.com/precice/precice/issues/1174)
* For creating an interpolant on a window, any kind of data provided by the user might be useful. Currently this "write direction" is still in a very early stage, but according to my current plans `isWriteDataRequired` would always return `true`.

Another important point: The two functions are barely used in tests (outside of the tests that I have implemented for waveform relaxation). This shows from my perspective again that these two functions are not really needed.

### LINKED ISSUE #1358: Simplify signature of actions
While working on #1350 I realized that using `dt` for actions during initialization is risky, because it is not really clear what `dt` means at thsi point in time. Under certain circumstances `dt` is not even known (if second participant performs a write actions during initialize and it has not yet received the `dt` from the first participant).

I'm labelling this as a bug, because I think this is currently a generally design issue.

I see two options:

* don't perform any actions during `initialize`
* remove actions that rely on `dt`

Sightly related: https://github.com/precice/precice/pull/1352. This might fix the issue that for the second participant in serial coupling initialize is actually a merge of initialization and receiving results from the first window of the first participant. I've split these two steps up here. So we can also perform no action in the first half and perform all the usual actions in the second half.

### LINKED ISSUE #1367: Generally allow initialize=true for all coupling schemes and all participants

Currently `initialize=true` is forbidden for data that is sent from first to second in a serial coupling scheme (because this data would be ignored). From [our documentation](https://precice.org/couple-your-code-initializing-coupling-data.html):

> In a serial coupling, only the second participant can send data inside initializeData(). In parallel coupling, both participants can initialize data.

I think that this restriction is basically motivated from the technical point of view, but recent changes that I implemented to develop waveform iteration remove the technical reasons (waveform iteration needs to allow initialization in all directions for serial coupling schemes. See #1352).

From the usability perspective I think this restriction makes things more complicated. It makes it more explicit that initial data sent from the first participant to the second is ignored, but this will become less relevant with waveform iteration, anyway.

Remove the error that is raised during configuration time:

https://github.com/precice/precice/blob/67b0e6d9c9629b69d923cc540056bfd63a258a56/src/cplscheme/config/CouplingSchemeConfiguration.cpp#L987-L990

The remaining necessary technical changes should already be included in #1352.

Keep it as it is, but I think this makes the implementation and documentation unnecessarily complicated.

### LINKED ISSUE #1421: Simplification of Waveform class
To implement #1171 the following changes to the Waveform class are useful:

* Remove feature to store and use data from past windows. There is currently no clear use-case and it makes the management of data more complicated.
* Allow to store several data values inside of a window. Always keep the value at the beginning of the window, when a new iteration begins and automatically delete all others.

These changes should not break any existing code, but will help us to implement higher-order interpolation when subcycling is used and multiple data values are available in a window.

BenjaminRodenberg:
## Main changes of this PR

Breaks develop. Only merge this PR as soon as the last release before the breaking release v3.0.0 is published.

Closes #834
Closes #859
Closes #934
Closes #1174
Closes #1196 
Closes #1223 
Closes #1358
Closes #1367
Closes #1421

At some point in time we need to break develop. Because some breaking developments already exist now, but cannot be merged now I created this breaking branch.

Note: We might need to rebase this branch depending on the number of changes on `develop` that will be introduced before being able to merge it.

## Todo

Update lists of Issues being closed.

## Author's checklist

* [x] ~~I added a changelog file with `make changelog` if there are user-observable changes since the last release.~~ not necessary, because the changelog files already come from the individual feature PRs.

## Reviewers' checklist


uekerman:
Please don't forget to add all changes to the [porting guide](https://precice.org/couple-your-code-porting-v2-3.html).

BenjaminRodenberg:
> Please don't forget to add all changes to the [porting guide](https://precice.org/couple-your-code-porting-v2-3.html).

I added a PR https://github.com/precice/precice.github.io/pull/223 and some porting instructions for `initializeData`. We can still merge this PR and break develop. When something explodes somewhere when using the broken develop, this is a good motivation for everybody to look into the porting guide (and if necessary add a section).

PR #1368, opened 2022-07-09
Allow initialize=true all data in serial coupling scheme

### LINKED ISSUE #1367: Generally allow initialize=true for all coupling schemes and all participants

BenjaminRodenberg:
## Main changes of this PR

Closes #1367 

See #1367 

## Todo

- [x] merge https://github.com/precice/precice/pull/1352

## Author's checklist

## Reviewers' checklist


src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -1044,6 +1054,21 @@ void CouplingSchemeConfiguration::checkIfDataIsExchanged(
...
+    for (auto &dataContext : participant->readDataContexts()) {
+      if (dataContext.getDataName() == dataName) {
+        int usedOrder = dataContext.getInterpolationOrder();
+        PRECICE_ASSERT(usedOrder >= 0); // ensure that usedOrder was set
+        return usedOrder;
+      }
+    }
+  }
+  return maxUsedOrder;
+}
```
uekerman, line 1070: I am a bit confused by the variable `maxUsedOrder` here.

Wouldn't it be "more correct" if we we pass the participant name as an argument here and return the order of that participant? In the end, we are only interested in the receiving participant, no?

BenjaminRodenberg: That's a leftover. You are 100% right. I will quickly fix it.

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -979,24 +979,14 @@ void CouplingSchemeConfiguration::addDataToBeExchanged(
...
     if (from == accessor) {
       scheme.addDataToSend(exchange.data, exchange.mesh, requiresInitialization);
-      if (requiresInitialization && (_config.type == VALUE_SERIAL_EXPLICIT || _config.type == VALUE_SERIAL_IMPLICIT)) {
-        if (not _experimental) {
-          PRECICE_CHECK(not scheme.doesFirstStep(),
-                        "In serial coupling only second participant can initialize data and send it. "
-                        "Please check the <exchange data=\"{}\" mesh=\"{}\" from=\"{}\" to=\"{}\" initialize=\"{}\" /> tag in the <coupling-scheme:... /> of your precice-config.xml. You can bypass this restriction by activating the experimental interface for time interpolation. Please refer to the documentation for more details.",
-                        dataName, meshName, from, to, requiresInitialization);
-        }
-      }
```
BenjaminRodenberg, line 989: We only define the waveform-order on the read-data. Therefore, we cannot check it here and raise a warning. Meaning: The first participant will not raise a warning.

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -979,24 +979,14 @@ void CouplingSchemeConfiguration::addDataToBeExchanged(
...
-        if (not _experimental) {
-          PRECICE_CHECK(scheme.doesFirstStep(),
-                        "In serial coupling only first participant can receive initial data. "
-                        "Please check the <exchange data=\"{}\" mesh=\"{}\" from=\"{}\" to=\"{}\" initialize=\"{}\" /> tag in the <coupling-scheme:... /> of your precice-config.xml. You can bypass this restriction by activating the experimental interface for time interpolation. Please refer to the documentation for more details.",
-                        dataName, meshName, from, to, requiresInitialization);
-        }
+      // Checks for a serial coupling scheme, where initial data is received by first participant.
+      if (!scheme.doesFirstStep() && requiresInitialization && (_config.type == VALUE_SERIAL_EXPLICIT || _config.type == VALUE_SERIAL_IMPLICIT) && getWaveformUsedOrder(dataName) == 0) // the order of the statements is crucial! getWaveformUsedOrder should be called last, because it may run into an error, if the data is initialized, but not defined as read data (for example Integration/Serial/AitkenAcceleration)
```
BenjaminRodenberg, line 984: This part is a bit tricky: Usually the second participant to which we are writing the initial data defines this as `read-data`. This is strictly not required (@uekerman, we briefly discussed this in the chat), but I think in most of the cases everything should work.

There is the highly theoretical case that a user does not define data that is initialized as read data. If this every happens, the error in `getWaveformUsedOrder` will be raised (and this might also be something worth knowing...).

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -1060,6 +1050,20 @@ void CouplingSchemeConfiguration::checkIfDataIsExchanged(
...
+  for (const precice::impl::PtrParticipant &participant : _participantConfig->getParticipants()) {
+    for (auto &dataContext : participant->readDataContexts()) {
+      if (dataContext.getDataName() == dataName) {
+        int usedOrder = dataContext.getInterpolationOrder();
+        PRECICE_ASSERT(usedOrder >= 0); // ensure that usedOrder was set
+        return usedOrder;
+      }
+    }
+  }
+  PRECICE_ERROR("Name \"{}\" not found. This is probably a bug. Please report it under https://github.com/precice/precice/issues/new/choose.", dataName);
```
BenjaminRodenberg, line 1064: I'm raising an error here, to make sure that the order that we receive from this function is actually valid.

BenjaminRodenberg (commented):
I added a few comments.

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -979,24 +979,14 @@ void CouplingSchemeConfiguration::addDataToBeExchanged(
...
-                        dataName, meshName, from, to, requiresInitialization);
-        }
+      // Checks for a serial coupling scheme, where initial data is received by first participant.
+      if (!scheme.doesFirstStep() && requiresInitialization && (_config.type == VALUE_SERIAL_EXPLICIT || _config.type == VALUE_SERIAL_IMPLICIT) && getWaveformUsedOrder(dataName) == 0) // the order of the statements is crucial! getWaveformUsedOrder should be called last, because it may run into an error, if the data is initialized, but not defined as read data (for example Integration/Serial/AitkenAcceleration)
+      {
+        PRECICE_WARN("In serial coupling initialized data received by the second participant will only be used, if waveform order > 0 is set. You can avoid unnecessary data initialization by setting initialize=\"false\" in the the <exchange data=\"{}\" mesh=\"{}\" from=\"{}\" to=\"{}\" initialize=\"{}\" /> tag in the <coupling-scheme:... /> of your precice-config.xml. If you want to use the initial data, please consider increasing the waveform-order in the <read-data name=\"{}\" mesh=\"{}\" /> tag in the <participant ... /> named \"{}\". ",
```
uekerman, line 986:
```suggestion
        PRECICE_WARN("In serial coupling, initialized data received by the second participant will only be used if a waveform order > 0 is set. You can avoid unnecessary data initialization by setting initialize=\"false\" in the <exchange data=\"{}\" mesh=\"{}\" from=\"{}\" to=\"{}\" initialize=\"{}\" /> tag in the <coupling-scheme:... /> of your precice-config.xml. If you want to use the initial data, please consider increasing the waveform-order in the <read-data name=\"{}\" mesh=\"{}\" /> tag in the <participant ... /> named \"{}\". ",
```

I assume you tried the warning such that all the formatting here works correctly.

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -1060,6 +1050,20 @@ void CouplingSchemeConfiguration::checkIfDataIsExchanged(
...
+{
+  for (const precice::impl::PtrParticipant &participant : _participantConfig->getParticipants()) {
+    for (auto &dataContext : participant->readDataContexts()) {
+      if (dataContext.getDataName() == readDataName) {
+        int usedOrder = dataContext.getInterpolationOrder();
+        PRECICE_ASSERT(usedOrder >= 0); // ensure that usedOrder was set
+        return usedOrder;
+      }
+    }
+  }
```
uekerman, line 1063: I still would find it easier to pass `participantName` as an argument here and not loop over all participants. Or does this not work?
Could be more robust and would also be easier to read.

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -1060,6 +1050,20 @@ void CouplingSchemeConfiguration::checkIfDataIsExchanged(
...
+  PRECICE_ERROR("Name \"{}\" not as read-data for this participant found. This is probably a bug in preCICE. Please report it under https://github.com/precice/precice/issues/new/choose.", readDataName);
```
uekerman, line 1064: Why not an assertion here?

src/cplscheme/config/CouplingSchemeConfiguration.hpp
```diff
@@ -270,6 +270,9 @@ class CouplingSchemeConfiguration : public xml::XMLTag::Listener {
   void checkIfDataIsExchanged(
       DataID dataID) const;
 
+  // get waveform order for read-data of given name. Will raise an error, if data name is not found in the read-data of this participant.
```
uekerman, line 273:
```suggestion
  /// Get waveform order for read data of given name. Raises an error if data name is not found in the read data of this participant.
```

uekerman (approved):
Minor tweaks.
Feel free to directly merge when done.

src/cplscheme/config/CouplingSchemeConfiguration.hpp
BenjaminRodenberg: Done in b47a2b1ab5f7eb3a6761c23b3efb78893c16b706.

src/cplscheme/config/CouplingSchemeConfiguration.cpp
BenjaminRodenberg: I think we actually need both, because the same read data can be used by multiple participants with different waveform order. I modified the function in 193a431a0db4f651fc087e6068440b96a3269e39.

Issue #1364, opened 2022-07-08
Remove deprecated mapping timings


uekerman:
There is no use case anymore, especially after #1222 

https://github.com/precice/precice/blob/67b0e6d9c9629b69d923cc540056bfd63a258a56/src/mapping/config/MappingConfiguration.hpp#L47-L51

See also #859

PR #1375, opened 2022-07-16
Remove timing "ondemand"

BenjaminRodenberg:
## Main changes of this PR

Remove `ondemand` timing. This is a follow-up for #1222 

## Author's checklist

## Reviewers' checklist


uekerman:
Ideally, we should remove all options at the same time, see #1364

BenjaminRodenberg:
I generally removed the timings via [7bd701b](https://github.com/precice/precice/pull/1375/commits/7bd701b51f7b807bddc3011a8860025a832c45c4) and I'm surprised, but it actually works. I think it would be good to have a critical look from somebody who is into mapping onto this PR (and if something needs to be done, I would also hand this PR over to somebody else).

docs/changelog/1375.md
```diff
@@ -0,0 +1 @@
+- Removes configuration option timingi from mapping, because API function mapWriteDataFrom and mapReadDataTo were removed and there is no use-case anymore. See #1364.
```
uekerman, line 1:
```suggestion
- Removed configuration option `timing` from mapping, because after removing the API functions `mapWriteDataFrom` and `mapReadDataTo`, there is no use case anymore.
```
PRs are added automatically.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1881,10 +1870,7 @@ void SolverInterfaceImpl::clearMappings(utils::ptr_vector<MappingContext> contex
   // Clear non-stationary, non-incremental mappings
```
uekerman, line 1870:
```suggestion
```

uekerman (changes requested):
I have the feeling that `SolverInterfaceImpl::clearMappings()` was used for two different purposes before expecting different things. It looks cleaner now, but we still have to delete the call from `mapWrittenData` and `mapReadData`.

docs/changelog/1375.md
BenjaminRodenberg: Nice. Did not know this :)

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRodenberg: I don't really understand how `SolverInterfaceImpl::clearMappings()` is related to this PR. Why should we remove it? If we remove the calls of `SolverInterfaceImpl::clearMappings()` from `mapWrittenData` and `mapReadData`, we can remove the entire function as far as I understand.

BenjaminRodenberg: Btw: This is something @fsimonis might also veto, because I remember him having some good experiences with resetting data of mappings at the right point in time.

uekerman: The `onadvance` option basically meant that the mapping got cleared after every call. As we no longer have this option, we neither need the call. We basically make the previous `initial` option our default.
`clearMappings`, we only need after the repartitioning.

BenjaminRodenberg:
@uekerman and @davidscn: This PR is very old and outdated. Can we close it? It looks like https://github.com/precice/precice/pull/1536 already took care of everything contained here.

PR #1373, opened 2022-07-14
Accelerate gradient data for constant and Aitken underrelaxation

davidscn:
## Main changes of this PR
Adds a support for gradient data for constant underrelaxation and Aitken underrelaxation.

## Motivation and additional information
In case our data has gradient data, we need to accelerate the corresponding data additionally. This PR implements the acceleration for our  underrelaxation variants, adds unit tests and extends the coupling data class to incorporate iteration dependent gradient data. Part of #1252.

Depends on #1371.

## Author's checklist

* [x] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [x] I ran `make format` to ensure everything is formatted correctly.
* [x] I sticked to C++14 features.
* [x] I sticked to CMake version 3.16.3.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [x] Does the changelog entry make sense? Is it formatted correctly?
* [x] Do you understand the code changes?

PR #1377, opened 2022-07-21
Add a warning when gradient data is used in QN acceleration

DavidSCN:
## Main changes of this PR
...which is currently incompatible.

Extending the QN scheme to handle the additional gradient data is non trivial, as the implementation relies on registered `dataID`s, which we don't have in case of gradient data (gradients are part of the data object itself).

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/acceleration/BaseQNAcceleration.cpp
+++ b/src/acceleration/BaseQNAcceleration.cpp
@@ -83,6 +83,11 @@ void BaseQNAcceleration::initialize(
                    pair.second->values().size(), pair.second->previousIteration().size());
   }
 
+  if (std::any_of(cplData.cbegin(), cplData.cend(), [](const auto &p) { return p.second->hasGradient(); })) {
+    PRECICE_WARN("Gradient data required by the configured mapping is not yet accelerated with quasi-Newton acceleration schemes, which might lead to numerical issues. "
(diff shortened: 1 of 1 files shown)


src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -83,6 +83,11 @@ void BaseQNAcceleration::initialize(
                    pair.second->values().size(), pair.second->previousIteration().size());
   }
 
+  if (std::any_of(cplData.cbegin(), cplData.cend(), [](const auto &p) { return p.second->hasGradient(); })) {
+    PRECICE_WARN("Gradient data required by the configured mapping is not yet accelerated with quasi-Newton acceleration schemes, which might lead to numerical issues. "
```
uekerman, line 87:
```suggestion
    PRECICE_WARN("Gradient data, which is required by at least one of the configured data mappings, is not yet compatible with quasi-Newton acceleration. This combination might lead to numerical issues. "
```

uekerman (approved):
Looks good.
For Aitken, you already added compatibility, right?

src/acceleration/BaseQNAcceleration.cpp
DavidSCN: To be reviewed https://github.com/precice/precice/pull/1373

PR #1383, opened 2022-07-25
Add hint for send/receive errors (sockets)

MakisH:
## Main changes of this PR

Extends error messages:

```diff
- PRECICE_ERROR("Send using sockets failed with system error: {}", e.what());
+ PRECICE_ERROR("Send using sockets failed with system error: {}. This often means that another participant exited with an error (look there).", e.what());
```
and
```diff
- PRECICE_ERROR("Receive using sockets failed with system error: {}", e.what());
+ PRECICE_ERROR("Receive using sockets failed with system error: {}. This often means that another participant exited with an error (look there).", e.what());
```

This error message sounds very cryptic to a new user, and we often receive related support requests, as also pointed out by @DavidSCN. Previously, @fsimonis extended the error message to its current form in https://github.com/precice/precice/pull/740/files#diff-07dcd4c79df0d50ed3443c1b0a273fd00dc61acf24dee7638ae10a6a5f9a57f8, without adding a hint, I assume because the exception caught can be anything. However, this is the most commonly encountered scenario, and such a hint could already help a few users.

Note: Something I don't like is that there is quite some duplication in the actual messages. Not sure if it is worth it trying to reduce duplication here, however.

## Author's checklist

* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release. -> Not sure if I should add one for such a trivial change.

## Reviewers' checklist

@fsimonis Does this make sense? Am I overlooking something? Should we also apply the same elsewhere?

### DIFF
--- a/src/com/SocketCommunication.cpp
+++ b/src/com/SocketCommunication.cpp
@@ -354,7 +354,7 @@ void SocketCommunication::send(std::string const &itemToSend, Rank rankReceiver)
     asio::write(*_sockets[rankReceiver], asio::buffer(&size, sizeof(size_t)));
     asio::write(*_sockets[rankReceiver], asio::buffer(itemToSend.c_str(), size));
   } catch (std::exception &e) {
-    PRECICE_ERROR("Send using sockets failed with system error: {}", e.what());
+    PRECICE_ERROR("Send using sockets failed with system error: {}. This often means that another participant exited with an error (look there).", e.what());
   }
 }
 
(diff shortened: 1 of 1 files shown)


src/com/SocketCommunication.cpp
```diff
@@ -354,7 +354,7 @@ void SocketCommunication::send(std::string const &itemToSend, Rank rankReceiver)
     asio::write(*_sockets[rankReceiver], asio::buffer(&size, sizeof(size_t)));
     asio::write(*_sockets[rankReceiver], asio::buffer(itemToSend.c_str(), size));
   } catch (std::exception &e) {
-    PRECICE_ERROR("Send using sockets failed with system error: {}", e.what());
+    PRECICE_ERROR("Send using sockets failed with system error: {}. This often means that another participant exited with an error (look there).", e.what());
```
DavidSCN, line 357: Actually, I also wanted to change the first sentence, as it confused me for quite a while. I would
1. try to clarify that 'send' refers to an event. Maybe the "send data event" or "while trying to send data"
2. point out that this "send event" refers to inter-participant communication, so that the following hint makes more sense
3. try to clarify that sockets refer to the m2n tag in our config. Many people just use it as a default and never touch it at all so that it's not clear where the sockets belong to. I'm not even sure that the information of communication configuration matters here, as we usually don't want to encourage people to change it when hitting the error message.

MakisH: Very good points, I updated the error messages to reflect all these.

uekerman (approved):
I like :+1:

PR #1387, opened 2022-07-28
Scaled consistent volume mode

boris-martin:
## Main changes of this PR

Followup of https://github.com/precice/precice/pull/1342.
Added a new constraint, `scaled-consistent-volume`that is similar to regular scaled-consistent but computes a different kind of integrals.

(More tests incoming)

## Author's checklist

## Reviewers' checklist


src/mapping/Mapping.hpp
```diff
@@ -21,12 +21,14 @@ class Mapping {
...
-   * should be mapped consistent.
+   * should be mapped consistent. There is also a volume version 
+   * where volume integrals are preserved instead of area integrals.
    */
   enum Constraint {
     CONSISTENT,
     CONSERVATIVE,
-    SCALEDCONSISTENT
+    SCALEDCONSISTENT,
+    SCALEDCONSISTENT_VOLUME
```
uekerman, line 31:
```suggestion
    SCALED_CONSISTENT,
    SCALED_CONSISTENT_VOLUME
```
not directly related to this PR, but looks better to me

src/mapping/Mapping.cpp
```diff
@@ -106,47 +106,63 @@ void Mapping::map(int inputDataID,
...
+  int  spaceDimension    = input()->getDimensions() == 3;
+  bool requiresEdges     = (spaceDimension == 2 and !volumeMode);
+  bool requiresTriangles = (spaceDimension == 2 and volumeMode) or (spaceDimension == 3 and !volumeMode);
+  bool requiresTetra     = (spaceDimension == 3 and volumeMode);
+
+  for (mesh::PtrMesh mesh : {input(), output()}) {
+    if (not mesh->vertices().empty()) {
+      if ((requiresEdges and mesh->edges().empty()) or
+          (requiresTriangles and mesh->triangles().empty()) or (requiresTetra and mesh->tetrahedra().empty())) {
+        PRECICE_ERROR("Connectivity information is missing for the mesh {}. "
```
DavidSCN, line 138:
```suggestion
        PRECICE_ERROR("Connectivity information is missing for the mesh \"{}\". "
```
Not completely sure if it is needed.

src/mapping/Mapping.cpp
```diff
@@ -106,47 +106,63 @@ void Mapping::map(int inputDataID,
...
+                      "Scaled consistent mapping requires connectivity information.",
```
DavidSCN, line 139: It would be nice to see from the error message, which connectivity information is missing/required.

src/mapping/Mapping.cpp
```diff
@@ -106,47 +106,63 @@ void Mapping::map(int inputDataID,
...
+    integralInput  = mesh::integrateVolume(input(), input()->data(inputDataID));
+    integralOutput = mesh::integrateVolume(output(), output()->data(outputDataID));
+  }
 
   // Create reshape the output values vector to matrix
   Eigen::Map<Eigen::MatrixXd> outputValuesMatrix(outputValues.data(), valueDimensions, outputValues.size() / valueDimensions);
 
   // Scale in each direction
   Eigen::VectorXd scalingFactor = integralInput.array() / integralOutput.array();
+  PRECICE_DEBUG("Scale factor in scale-consistent mapping: {}", scalingFactor);
```
DavidSCN, line 165:
```suggestion
  PRECICE_DEBUG("Scaling factor in scaled-consistent mapping: {}", scalingFactor);
```

src/mapping/Mapping.hpp
```diff
@@ -21,12 +21,14 @@ class Mapping {
    * then scales the mapped such that the integrals on both sides of the interface
    * are equal. Values integrated over some area should be mapped conservative or
    * scaled-consistent, while area independent values such as pressure or stresses
-   * should be mapped consistent.
+   * should be mapped consistent. There is also a volume version 
+   * where volume integrals are preserved instead of area integrals.
```
DavidSCN, line 25:
```suggestion
   * where volume integrals are preserved instead of surface integrals.
```

src/mapping/NearestProjectionMapping.cpp
```diff
@@ -36,10 +36,12 @@ NearestProjectionMapping::NearestProjectionMapping(
     setInputRequirement(Mapping::MeshRequirement::VERTEX);
     setOutputRequirement(Mapping::MeshRequirement::FULL);
   } else {
-    PRECICE_ASSERT(constraint == SCALEDCONSISTENT, constraint);
+    PRECICE_ASSERT(constraint == SCALED_CONSISTENT || constraint == SCALED_CONSISTENT_VOLUME, constraint);
     setInputRequirement(Mapping::MeshRequirement::FULL);
     setOutputRequirement(Mapping::MeshRequirement::FULL);
   }
+
+  PRECICE_CHECK(constraint != SCALED_CONSISTENT_VOLUME, "Nearest-projection can't be used with volume version of the scaled-consistent mapping. Use scaled-consistent instead.");
```
DavidSCN, line 44: Why?

DavidSCN (commented):
Hm, I would actually prefer the previous version of this implementation (#1342). From what I understand, you re-implemented this PR after discussing with @uekerman.
What I don't really like: 

1. we have now a somewhat inconsistent xml config of `scaled-consistent` vs `scaled-consistent-volume`. IMO, it should be `scaled-consistent-surface`
2. in terms of the implementation, adding another type to the `constraint` enum treats this new variant as a completely new mapping constraint in terms of configuration and 'what-to-do-when', which is in practice not the case. The actual implementation only branches of in one particular place in the implementation
 ```cpp 
 if (!volumeMode) {
    integralInput  = mesh::integrate(input(), input()->data(inputDataID));
    integralOutput = mesh::integrate(output(), output()->data(outputDataID));
  } else {
    integralInput  = mesh::integrateVolume(input(), input()->data(inputDataID));
    integralOutput = mesh::integrateVolume(output(), output()->data(outputDataID));
  }
```
the multitude of changes of type `(hasConstraint(SCALED_CONSISTENT)` -> `(hasConstraint(SCALED_CONSISTENT) || hasConstraint(SCALED_CONSISTENT_VOLUME))` also suggests that adding an enum type here is not correct. The scaling (scaled-consistent mapping) is anyway rather a post-processing of data than an actual data mapping constraint. At least for the implementation, it might be easier to handle this in a different way.

**Comment by boris-martin** [comment]
>     1. we have now a somewhat inconsistent xml config of `scaled-consistent` vs `scaled-consistent-volume`. IMO, it should be `scaled-consistent-surface`

It's mostly a backwards compatibility issue. @uekerman  and I agree to keep `scaled-consistent` in preCICE 2.5 then rename it `scaled-consistent-surface` in the v3 branch.

>     2. in terms of the implementation, adding another type to the `constraint` enum treats this new variant as a completely new mapping constraint in terms of configuration and 'what-to-do-when', which is in practice not the case. The actual implementation only branches of in one particular place in the implementation

What would you go ? Keep the same XML API but keep one SCALEDCONSISTENT enum, with an additional enumeration VOLUME/SURFACE ? If I remember right, the point of this new PR was to avoid one more attribute than was annoying to bookkeep.

To reduce the number of "SCALED_CONSISTENT_SURFACE || SCALED_CONSISTENT_VOUME" I guess we could refactor by using a helper function called `isScaledMapping` or something like that.

File: src/mapping/NearestProjectionMapping.cpp (hunk 6 above)
Comment (boris-martin) [reply src/mapping/NearestProjectionMapping.cpp:44]: It's not technically impossible, but it's semantically wrong. If you use NP, you should only configure surface primitives. Then, scaling over a volume would be weird. If you want a volume-scaling and have connectivity, linear cell interpolation should be used, not NP.

Similarly, surface scaling is not allowed with linear cell interpolation.

**Comment by uekerman** [comment]
> Hm, I would actually prefer the previous version of this implementation (https://github.com/precice/precice/pull/1342)

@DavidSCN That one would have let to really complex decision trees, only due to a very rare use case (scaled-consistent).
One could have define a `nearest-projection` mapping with `volume`; or an RBF mapping `conservative` and `surface`. We would need checks for all of these cases.

With the solution here, we keep the regular cases simple.

**Comment by DavidSCN** [comment]
>    It's mostly a backwards compatibility issue. @uekerman and I agree to keep scaled-consistent in preCICE 2.5 then rename it scaled-consistent-surface in the v3 branch.

Now that this PR is scheduled for v3, do you want to add this breaking change in this PR?

> To reduce the number of "SCALED_CONSISTENT_SURFACE || SCALED_CONSISTENT_VOUME" I guess we could refactor by using a helper function called isScaledMapping or something like that.

Sounds reasonable, I don't have a better idea right now.

File: src/mapping/Mapping.hpp [hunk 7]
```diff
@@ -21,12 +21,14 @@ class Mapping {
    * then scales the mapped such that the integrals on both sides of the interface
    * are equal. Values integrated over some area should be mapped conservative or
    * scaled-consistent, while area independent values such as pressure or stresses
-   * should be mapped consistent.
+   * should be mapped consistent. There is also a volume version 
+   * where volume integrals are preserved instead of surface integrals.
```
davidscn, line 25: The docs seem a bit outdated. They are not incorrect, but I guess you could immediately merge the volume stuff into the scaled consistent description?!

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -462,7 +462,7 @@ void ParticipantConfiguration::finishParticipantConfiguration(
           (confMapping.direction == mapping::MappingConfiguration::READ &&
            confMapping.mapping->getConstraint() == mapping::Mapping::CONSERVATIVE)) {
         PRECICE_ERROR("For a parallel participant, only the mapping combinations read-consistent and write-conservative are allowed");
-      } else if (confMapping.mapping->getConstraint() == mapping::Mapping::SCALEDCONSISTENT) {
+      } else if (confMapping.mapping->getConstraint() == mapping::Mapping::SCALED_CONSISTENT_SURFACE || confMapping.mapping->getConstraint() == mapping::Mapping::SCALED_CONSISTENT_VOLUME) {
```
davidscn, line 465: Why not using the helper function here?

tests/serial/mapping-scaled-consistent/helpers.cpp
```diff
@@ -106,4 +106,203 @@ void testQuadMappingScaledConsistent(const std::string configFile, const TestCon
...
+    // namespace is required because we are outside the fixture
+    int meshTwoID = interface.getMeshID("MeshTwo");
+
+    // Setup receiving mesh.
+    int idA = interface.setMeshVertex(meshTwoID, coordTwoA.data());
+    int idB = interface.setMeshVertex(meshTwoID, coordTwoB.data());
+    int idC = interface.setMeshVertex(meshTwoID, coordTwoC.data());
+    int idD = interface.setMeshVertex(meshTwoID, coordTwoD.data());
+
+    int idAB = interface.setMeshEdge(meshTwoID, idA, idB);
```
davidscn, line 181: Do we really need to store the return value here? CI claims that this value (and the following) is unused?

davidscn (commented):
We definitely need a changelog entry, as this is now breaking.

src/mapping/Mapping.cpp
```diff
@@ -106,47 +106,60 @@ void Mapping::map(int inputDataID,
...
-                  "Scaled consistent mapping requires connectivity information.",
-                  input()->getName());
-  }
-  if ((output()->edges().empty() and (not output()->vertices().empty())) or
-      (((output()->getDimensions() == 3) and output()->triangles().empty()) and (not output()->vertices().empty()))) {
-    logging::Logger _log{"mapping::Mapping"};
-    PRECICE_ERROR("Connectivity information is missing for the mesh {}. "
-                  "Scaled consistent mapping requires connectivity information.",
-                  output()->getName());
+  int  spaceDimension    = input()->getDimensions() == 3;
```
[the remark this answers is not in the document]
boris-martin: Yep, deleted the == 3

src/mapping/Mapping.cpp
```diff
@@ -106,47 +106,60 @@ void Mapping::map(int inputDataID,
...
+                      mesh->getName());
+      }
+    }
   }
 
   auto &outputValues    = output()->data(outputDataID)->values();
   int   valueDimensions = input()->data(inputDataID)->getDimensions();
 
+  Eigen::VectorXd integralInput;
+  Eigen::VectorXd integralOutput;
```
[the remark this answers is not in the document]
boris-martin: Something like `integralInput = volumeMode ? integrateVolume(...) : integrateSurface(...)` ? Not a big fan.

src/mapping/Mapping.hpp
boris-martin: Updated this. Can you review the new version ?

tests/serial/mapping-scaled-consistent/helpers.cpp
boris-martin: How is that? The edge IDs are used to define the triangles. (Until we update to the Vertex-only API)

boris-martin: Actually we can remove the edges completely since we use `SetMeshTriangleWithEdges`. Or should we rather convert to the vertex-only API?

src/mapping/Mapping.cpp
boris-martin: Updated this.

src/mapping/Mapping.cpp
```diff
@@ -106,49 +106,71 @@ void Mapping::map(int inputDataID,
...
-                  "Scaled consistent mapping requires connectivity information.",
-                  output()->getName());
+  int  spaceDimension    = input()->getDimensions();
+  bool requiresEdges     = (spaceDimension == 2 and !volumeMode);
+  bool requiresTriangles = (spaceDimension == 2 and volumeMode) or (spaceDimension == 3 and !volumeMode);
+  bool requiresTetra     = (spaceDimension == 3 and volumeMode);
+
+  for (mesh::PtrMesh mesh : {input(), output()}) {
+    if (not mesh->vertices().empty()) {
+      if (requiresEdges and mesh->edges().empty()) {
```
davidscn, line 133: Could be a `PRECICE_CHECK` and merged with the error message below, right?

davidscn (approved):
Looks good, however, a changelog is still missing. We need to wait here until breaking changes are allowed to be merged.

davidscn:
> We need to wait here until breaking changes are allowed to be merged.

Waiting time is over. @boris-martin will you add a changelog or do you want me to do that?

PR #1395, opened 2022-08-01
Fix the triangle distance calculation

### LINKED ISSUE #1388: R-Tree weird behavior with respect to the Z-axis (Boost bug ?)
**Describe your setup**

preCICE Version: 2.4.0 develop branch a bit before 2.5.0 release. Followup of https://github.com/precice/precice/pull/1363, found in https://github.com/precice/precice/pull/1361.

**Describe the problem**

This unit test  (insert in RTreeTests.cpp) fails. A point with z=0.3 has a zero distance to a triangle in the plane z=0 (or any other constant), according to Boost. It works when switching axes (e.g. x=0 plane):
```cpp

BOOST_AUTO_TEST_CASE(CorrectTriangleDistance)
{

  // Document the issue
  PRECICE_TEST(1_rank);
  PtrMesh mesh(new Mesh("MyMesh", 3, testing::nextMeshID()));
  Index   indexTree(mesh);

  mesh::Vertex &v1 = mesh->createVertex(Eigen::Vector3d(0.0, 0.0, 0.0));
  mesh::Vertex &v2 = mesh->createVertex(Eigen::Vector3d(1.0, 0.0, 0.0));
  mesh::Vertex &v3 = mesh->createVertex(Eigen::Vector3d(0.0, 1.0, 0.0));
  mesh->createTriangle(v1, v2, v3);

  auto matches = indexTree.getClosestTriangles(Eigen::Vector3d{0.1, 0.2, 0.3}, 4);
  BOOST_REQUIRE(matches.size() == 1);
  BOOST_TEST(matches[0].distance > 0.25);
}

```

**Expected behaviour**
Non-zero distance

fsimonis:
## Main changes of this PR

This PR fixes issues with the triangle to point distance calculation of the mesh index.
Index methods `getClosestX` now don't returned matches sorted by distance anymore, but simply the IDs of the matches.

Closes #1388

Boost geometry was used to calculate distances of a location to n-closest primitives.
This distance calculation returned 0 for the triangle case where the location projection was inside the triangle.
So, the reported projection metrics for triangles in NP were reporting the wrong distance.
Also, this may have lead to further away triangles with correct projection getting picked over closer ones.

Most of these issues do not show up in surface coupling with nearest-projection mapping.
This would require very coarse and thin geometries, where opposing triangles get considered. Then the opposing may be picked first.
Of course, this case is common once a fallback is required in volume coupling.

The root course of this issue is that implementations of boost geometry don't always work with polygons and rings in 3D.
This can be reproduced with a `boost::model::ring` and a `boost::model::point`.

## Author's checklist

## Reviewers' checklist

* [ ] Do you understand the error?
* [ ] Do you understand the fix?

### DIFF
--- a/src/mapping/NearestNeighborBaseMapping.cpp
+++ b/src/mapping/NearestNeighborBaseMapping.cpp
@@ -61,9 +61,11 @@ void NearestNeighborBaseMapping::computeMapping()
 
   auto &index = searchSpace->index();
   for (size_t i = 0; i < verticesSize; ++i) {
-    const auto &matchedVertex = index.getClosestVertex(sourceVertices[i].getCoords());
+    const auto &sourceCoords  = sourceVertices[i].getCoords();
+    const auto &matchedVertex = index.getClosestVertex(sourceCoords);
     _vertexIndices[i]         = matchedVertex.index;
-    distanceStatistics(matchedVertex.distance);
+    auto distance             = (sourceCoords - searchSpace->vertices()[matchedVertex.index].getCoords()).norm();
+    distanceStatistics(distance);
   }
 
   // For gradient mapping, the calculation of offsets between source and matched vertex necessary
--- a/src/mapping/NearestNeighborBaseMapping.cpp
+++ b/src/mapping/NearestNeighborBaseMapping.cpp
@@ -64,7 +64,10 @@ void NearestNeighborBaseMapping::computeMapping()
     const auto &sourceCoords  = sourceVertices[i].getCoords();
     const auto &matchedVertex = index.getClosestVertex(sourceCoords);
     _vertexIndices[i]         = matchedVertex.index;
-    auto distance             = (sourceCoords - searchSpace->vertices()[matchedVertex.index].getCoords()).norm();
+
+    // Compute distance between input and output vertiex for the stats
+    const auto &matchCoords = searchSpace->vertices()[matchedVertex.index].getCoords();
+    auto        distance    = (sourceCoords - matchCoords).norm();
     distanceStatistics(distance);
   }
 
--- a/src/mapping/tests/PolationTest.cpp
+++ b/src/mapping/tests/PolationTest.cpp
@@ -58,6 +58,30 @@ BOOST_AUTO_TEST_CASE(EdgeInterpolation)
   }
 }
 
+BOOST_AUTO_TEST_CASE(EdgeProjectedInterpolation)
+{
+  PRECICE_TEST(1_rank);
+  mesh::Vertex v1(Eigen::Vector3d(0.0, 0.0, 0.0), 0);
+  mesh::Vertex v2(Eigen::Vector3d(0.0, 2.0, 0.0), 1);
+  mesh::Edge   edge(v1, v2, 0);
+
+  Eigen::Vector3d location(0.0, 0.4, 0.12);
+
+  Polation polation(location, edge);
+
+  std::vector<int>    expectedIndices = {0, 1};
+  std::vector<double> expectedWeights = {0.8, 0.2};
+
+  BOOST_TEST(polation.getWeightedElements().size() == 2);
+  BOOST_TEST(polation.isInterpolation());
+  BOOST_TEST(polation.distance() == 0.12);
+
+  for (size_t i = 0; i < polation.getWeightedElements().size(); ++i) {
+    BOOST_TEST(polation.getWeightedElements().at(i).weight == expectedWeights.at(i));
+    BOOST_TEST(polation.getWeightedElements().at(i).vertexID == expectedIndices.at(i));
+  }
+}
+
 BOOST_AUTO_TEST_CASE(TriangleInterpolation)
 {
   PRECICE_TEST(1_rank);
@@ -88,6 +112,36 @@ BOOST_AUTO_TEST_CASE(TriangleInterpolation)
   }
 }
 
+BOOST_AUTO_TEST_CASE(TriangleProjectedInterpolation)
+{
+  PRECICE_TEST(1_rank);
+  mesh::Vertex   v1(Eigen::Vector3d(0.0, 0.0, 0.0), 0);
+  mesh::Vertex   v2(Eigen::Vector3d(2.0, 0.0, 0.0), 1);
+  mesh::Vertex   v3(Eigen::Vector3d(1.0, 2.0, 0.0), 2);
+  mesh::Edge     e1(v1, v2, 0);
+  mesh::Edge     e2(v2, v3, 1);
+  mesh::Edge     e3(v1, v3, 2);
+  mesh::Triangle triangle(e1, e2, e3, 0);
+  triangle.computeNormal();
+
+  Eigen::Vector3d location(1.0, 0.6, 0.14);
+
+  Polation polation(location, triangle);
+
+  std::vector<int>    expectedIndices = {0, 1, 2};
+  std::vector<double> expectedWeights = {0.35, 0.35, 0.3};
+
+  BOOST_TEST(polation.getWeightedElements().size() == 3);
+  BOOST_TEST(polation.isInterpolation());
+  BOOST_TEST(polation.distance() == 0.14);
+
+  for (size_t i = 0; i < polation.getWeightedElements().size(); ++i) {
+
+    BOOST_TEST(polation.getWeightedElements().at(i).weight == expectedWeights.at(i));
+    BOOST_TEST(polation.getWeightedElements().at(i).vertexID == expectedIndices.at(i));
+  }
+}
+
 BOOST_AUTO_TEST_CASE(EdgeExtrapolation)
 {
   PRECICE_TEST(1_rank);
@@ -104,6 +158,8 @@ BOOST_AUTO_TEST_CASE(EdgeExtrapolation)
 
   BOOST_TEST(polation.getWeightedElements().size() == 2);
   BOOST_TEST(not polation.isInterpolation());
+  // The distance to projection is still 0
+  BOOST_TEST(polation.distance() == 0);
 
   for (size_t i = 0; i < polation.getWeightedElements().size(); ++i) {
     BOOST_TEST(polation.getWeightedElements().at(i).weight == expectedWeights.at(i));
@@ -132,6 +188,8 @@ BOOST_AUTO_TEST_CASE(TriangleExtrapolation)
 
   BOOST_TEST(polation.getWeightedElements().size() == 3);
   BOOST_TEST(not polation.isInterpolation());
+  // The distance to projection is still 0
+  BOOST_TEST(polation.distance() == 0);
 
   for (size_t i = 0; i < polation.getWeightedElements().size(); ++i) {
 
@@ -187,6 +245,8 @@ BOOST_AUTO_TEST_CASE(TetrahedronExtrapolation)
 
   BOOST_TEST(polation.getWeightedElements().size() == 4);
   BOOST_TEST(not polation.isInterpolation());
+  // There is no projection distance as such. This should always be 0
+  BOOST_TEST(polation.distance() == 0);
 
   for (size_t i = 0; i < polation.getWeightedElements().size(); ++i) {
 
--- a/src/mapping/tests/PolationTest.cpp
+++ b/src/mapping/tests/PolationTest.cpp
@@ -16,15 +16,17 @@ BOOST_AUTO_TEST_SUITE(Interpolation)
 BOOST_AUTO_TEST_CASE(VertexInterpolation)
 {
   PRECICE_TEST(1_rank);
-  mesh::Vertex vertex(Eigen::Vector3d(1.0, 2.0, 0.0), 0);
+  Eigen::Vector3d location(0.0, 0.0, 0.0);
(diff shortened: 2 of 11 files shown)


boris-martin:
As I said on the other page:

> In that case, I'd suggest replacing the use of `getClosestTriangles(point, n)` by something similar to the approach used with tetrahedra, where boost distances aren't used at all and we get all triangles whose AABB contains the point. Then we'd have to use the polation distance and sort the matches by distance.
> 
> Side advantage would be that we'd get rid of the arbitrary value of the number `n` of closest elements.

Shouldn't we remove the nearest neighbor request since the computation of distance in Boost is wrong ?

fsimonis:
> In that case, I'd suggest replacing the use of getClosestTriangles(point, n) by something similar to the approach used with tetrahedra, where boost distances aren't used at all and we get all triangles whose AABB contains the point. Then we'd have to use the polation distance and sort the matches by distance.

This doesn't work as we are looking for projections onto primitives. So, using intersections between points and AABBs doesn't work. We use the a priori knowledge that spacially far-away primitives are bad for data-mapping and select only the closest-n.

As we use AABBs of Triangles in the rtree, this query is cheap and the rtree doesn't work directly on the triangles. Hence, no boost issues here.

getClosestTriangles uses boost distances only for sorting the results. This is something I removed for all getClosestX functions. You only get the n-closest triangles now.

The getXProjection functions are now responsible to calculate the projection first, remove extrapolations and then pick the match with the lowest projection distance.

> Side advantage would be that we'd get rid of the arbitrary value of the number n of closest elements.

I don't the arbitrary value either, but it works nicely. This should become user-customizable though.

boris-martin:
> > In that case, I'd suggest replacing the use of getClosestTriangles(point, n) by something similar to the approach used with tetrahedra, where boost distances aren't used at all and we get all triangles whose AABB contains the point. Then we'd have to use the polation distance and sort the matches by distance.
> 
> This doesn't work as we are looking for projections onto primitives. So, using intersections between points and AABBs doesn't work. We use the a priori knowledge that spacially far-away primitives are bad for data-mapping and select only the closest-n.

Right.

> As we use AABBs of Triangles in the rtree, this query is cheap and the rtree doesn't work directly on the triangles. Hence, no boost issues here.

> getClosestTriangles uses boost distances only for sorting the results. This is something I removed for all getClosestX functions. You only get the n-closest triangles now.

But to find the closest triangles, Boost still uses the broken distance, no ?

fsimonis:
> Right.

Actually, it could make sense if you are looking for a point on a triangle in 2D for cell-interpolations. I am still in the surface coupling mindset :grin: 

> But to find the closest triangles, Boost still uses the broken distance, no ?

Nope, the rtree is purely based on AABBs. As I wrote, they used to be sorted by distance to the actual triangle. This is no longer the case.
I use the polation for calculating the distance to triangles that result in an interpolation.

Location -> n-nearest triange AABBs -> n polations -> remove extrapolations -> use the polation to calculate the distance from the projectee to the projection.

boris-martin:
My understanding was that the distance between an AABB and a point was broken, since Boost only knows about them. No ?

fsimonis:
AABB distances look fine:

https://godbolt.org/#g:!((g:!((g:!((h:codeEditor,i:(filename:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,selection:(endColumn:67,endLineNumber:17,positionColumn:67,positionLineNumber:17,selectionStartColumn:67,selectionStartLineNumber:17,startColumn:67,startLineNumber:17),source:'%23include%3Ciostream%3E%0A%23include+%3Cboost/geometry.hpp%3E%0A%0Anamespace+bg+%3D+boost::geometry%3B%0A%0Aint+main()%0A%7B%0A++++using+P+%3D+bg::model::point%3Cdouble,+3,+bg::cs::cartesian%3E%3B%0A++++using+B+%3D+bg::model::box%3CP%3E%3B%0A%0A++++B+b%7B%7B0.,0.,0.%7D,+%7B1.,1.,1.%7D%7D%3B%0A%0A++++P+inside%7B0.3,0.5,0.6%7D%3B%0A++++P+aside%7B0.1,0.2,1.3%7D%3B%0A++++P+away%7B1.5,1.5,1.5%7D%3B%0A%0A++++std::cout+%3C%3C+%22inside+%22+%3C%3C+bg::distance(inside,+b)+%3C%3C+%22%5Cnaside+%22+%3C%3C+bg::distance(aside,+b)+%3C%3C+%22%5Cnaway+%22+%3C%3C+bg::distance(away,b)%3B%0A++++return+0%3B%0A%7D'),l:'5',n:'0',o:'C%2B%2B+source+%231',t:'0')),k:50,l:'4',m:100,n:'0',o:'',s:0,t:'0'),(g:!((g:!((h:compiler,i:(compiler:g121,filters:(b:'0',binary:'0',commentOnly:'0',demangle:'0',directives:'0',execute:'0',intel:'0',libraryCode:'0',trim:'1'),flagsViewOpen:'1',fontScale:14,fontUsePx:'0',j:1,lang:c%2B%2B,libs:!((name:boost,ver:'179')),options:'',selection:(endColumn:1,endLineNumber:1,positionColumn:1,positionLineNumber:1,selectionStartColumn:1,selectionStartLineNumber:1,startColumn:1,startLineNumber:1),source:1,tree:'1'),l:'5',n:'0',o:'x86-64+gcc+12.1+(C%2B%2B,+Editor+%231,+Compiler+%231)',t:'0')),header:(),k:50,l:'4',m:50,n:'0',o:'',s:0,t:'0'),(g:!((h:output,i:(compilerName:'x86-64+gcc+12.1',editorid:1,fontScale:14,fontUsePx:'0',j:1,wrap:'1'),l:'5',n:'0',o:'Output+of+x86-64+gcc+12.1+(Compiler+%231)',t:'0')),header:(),l:'4',m:50,n:'0',o:'',s:0,t:'0')),k:50,l:'3',n:'0',o:'',t:'0')),l:'2',n:'0',o:'',t:'0')),version:4

fsimonis:
Further clarification regarding the linear-cell-interpolation mapping:

The mapping is looking for a cell (Tetrahedron in 3D or Triangle in 2D), which contains the search point. In 2D, it queries for n-closest triangles and attempt to find an interpolation. So the order of the triangles does not matter.

src/mapping/NearestNeighborBaseMapping.cpp
```diff
@@ -61,9 +61,11 @@ void NearestNeighborBaseMapping::computeMapping()
 
   auto &index = searchSpace->index();
   for (size_t i = 0; i < verticesSize; ++i) {
-    const auto &matchedVertex = index.getClosestVertex(sourceVertices[i].getCoords());
+    const auto &sourceCoords  = sourceVertices[i].getCoords();
+    const auto &matchedVertex = index.getClosestVertex(sourceCoords);
     _vertexIndices[i]         = matchedVertex.index;
-    distanceStatistics(matchedVertex.distance);
+    auto distance             = (sourceCoords - searchSpace->vertices()[matchedVertex.index].getCoords()).norm();
```
DavidSCN, line 67: We should add a comment why we compute the distance here manually instead of using the boost functionality. Maybe even link to the issue.

src/mapping/tests/PolationTest.cpp
```diff
@@ -16,15 +16,17 @@ BOOST_AUTO_TEST_SUITE(Interpolation)
 BOOST_AUTO_TEST_CASE(VertexInterpolation)
 {
   PRECICE_TEST(1_rank);
-  mesh::Vertex vertex(Eigen::Vector3d(1.0, 2.0, 0.0), 0);
+  Eigen::Vector3d location(0.0, 0.0, 0.0);
```
DavidSCN, line 19: The unit test here is a bit trivial, maybe apply a different input in order to really cover the distance calculation.

src/query/Index.cpp
```diff
@@ -205,9 +204,8 @@ std::vector<TriangleMatch> Index::getClosestTriangles(const Eigen::VectorXd &sou
   std::vector<TriangleMatch> matches;
   rtree->query(bgi::nearest(sourceCoord, n),
                boost::make_function_output_iterator([&](TriangleTraits::IndexType const &match) {
-                 matches.emplace_back(bg::distance(sourceCoord, _mesh->triangles()[match.second]), match.second);
+                 matches.emplace_back(match.second);
                }));
-  std::sort(matches.begin(), matches.end());
```
DavidSCN, line 210: What happened with the sort?

src/query/Index.cpp
```diff
@@ -287,32 +285,45 @@ ProjectionMatch Index::findCellOrProjection(const Eigen::VectorXd &location, int
...
-  for (const auto &match : matchedEdges) {
+  std::vector<ProjectionMatch> candidates;
+  candidates.reserve(n);
+  for (const auto &match : getClosestEdges(location, n)) {
     auto polation = mapping::Polation(location, _mesh->edges()[match.index]);
     if (polation.isInterpolation()) {
-      return {polation, match.distance};
+      candidates.push_back({polation, polation.distance()});
     }
   }
```
DavidSCN, line 302: Looks a bit like `copy_if` could be used, not completely sure.

src/query/Index.hpp
```diff
@@ -25,19 +25,13 @@ constexpr MatchID NO_MATCH{-1};
...
+  MatchID index{NO_MATCH};
   MatchType() = default;
-  MatchType(Distance d, MatchID i)
-      : distance(d), index(i){};
-
-  constexpr bool operator<(MatchType const &other) const
-  {
-    return distance < other.distance;
-  };
+  MatchType(MatchID i)
```
DavidSCN, line 33: Should probably be `explicit`, right?

src/query/tests/RTreeTests.cpp
```diff
@@ -101,7 +101,6 @@ BOOST_AUTO_TEST_CASE(Query2DVertex)
 
   auto result = indexTree.getClosestVertex(location);
   BOOST_TEST(mesh->vertices().at(result.index).getCoords() == Eigen::Vector2d(0, 1));
-  BOOST_TEST(result.distance == 0.28284271247461906);
```
DavidSCN, line 104: In principle, these tests should now move to the polation tests, right?

DavidSCN (commented):
Looks good overall.

src/mapping/NearestNeighborBaseMapping.cpp
fsimonis: Actually, we could still use the distance of boost geometry here.
My long term goal is to limit boost geometry to the query package. These headers are expensive.

DavidSCN: I thought the bug is about the distance calculation? If that's not the case here, we could still rely on boost. I guess with 'expensive' you refer to compile time? How does it look like in terms of runtime performance (if you really want to open these boxes now).

src/query/Index.cpp
fsimonis: The XMatches are now only IDs, so sorting them doesn't make sense any longer.
Conceptually, the `getClosestX` now return sets.

src/query/Index.cpp
fsimonis: It is a mix between a transform `Match -> (Polation, distance)` and a filter operation `(Polation, distance) -> bool`.
We could implement this with boost ranges, but as so often in C++, it's more readable to use loops like this.

src/query/Index.hpp
fsimonis: Well spotted!

src/query/tests/RTreeTests.cpp
fsimonis: The interface is still returning a `ProjectionMatch` which is Polation + distance.

We really need to refactor the interface to return the Polation only, as it contains the distance. Then we should redistribute the test.

I didn't do this to get a solution quickly. I'll open an issue to do this after the release.

DavidSCN: I will leave it up to you. I could imagine that it would be easier and faster to do it now, where you are already into the topic, but opening an issue is also fine.

fsimonis: The mentioned test doesn't check the distance, as getClosestVertex doesn't return the distance any longer. Got muddled up with the projection tests.

I am still going to for the refactoring.

src/mapping/tests/PolationTest.cpp
fsimonis: I added cases with a z difference. This will be the projection distance.

src/mapping/NearestNeighborBaseMapping.cpp
fsimonis: > I thought the bug is about the distance calculation?

It is. I removed the distances from the `getClosestX` results. So, one still needs to compute the distance here.

> If that's not the case here, we could still rely on boost.

I disagree. We use Eigen for coordinates and data, so we should use it here as well.
Boost.geometry and all the boilerplate that comes with it is purely useful for the spacial index and geometric queries (like distances to AABBs etc). 

> I guess with 'expensive' you refer to compile time?

Exactly. Eigen is expensive enough on its own. We should pull in another dependency for something that Eigen was build for.

fsimonis: I clarified the code a bit.

DavidSCN (approved):

PR #1396, opened 2022-08-02
Update third party libs

DavidSCN:
## Main changes of this 
- Upgrade fmt to 9.0.0
- Upgrade json to 3.11.1
- Update tcbrindle span (unversioned)

## Author's checklist

## Reviewers' checklist


DavidSCN:
fmt 9 requires now to provide template specialization for user-defined types. I re-used now our insert operator in order to provide the corresponding specializations.

src/math/geometry.cpp
```diff
@@ -161,7 +161,7 @@ ConvexityResult isConvexQuad(std::array<Eigen::VectorXd, 4> coords)
   Eigen::Vector3d normalVector = e_1.cross(e_2);
 
   PRECICE_CHECK(math::equals(normalVector.dot(coords[3] - coordOrigin), 0.0),
-                "Non-planar quads are not supported. The vertex coordinates are: {}.", coords);
+                "Non-planar quads are not supported. The vertex coordinates are: {}; {}; {}; {}.", coords[0], coords[1], coords[2], coords[3]);
```
fsimonis, line 164: This should still work due to:

https://github.com/precice/precice/blob/0590b349df3e28e0a9a99e4e5e1a05256565d494/src/utils/fmtSTL.hpp#L28-L36

src/mesh/Mesh.cpp
```diff
@@ -481,3 +482,7 @@ std::ostream &operator<<(std::ostream &os, const Mesh &m)
 
 } // namespace mesh
 } // namespace precice
+
+template <>
+struct fmt::formatter<precice::mesh::Mesh> : ostream_formatter {
+};
```
fsimonis, line 488: Do we actually ever print the _entire_ Mesh?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -726,7 +726,7 @@ int SolverInterfaceImpl::setMeshVertex(
   int           index   = -1;
   MeshContext & context = _accessor->usedMeshContext(meshID);
   mesh::PtrMesh mesh(context.mesh);
-  PRECICE_DEBUG("MeshRequirement: {}", context.meshRequirement);
+  PRECICE_DEBUG("MeshRequirement: {}", fmt::underlying(context.meshRequirement));
```
fsimonis, line 729: This should use the ostream formatter

```suggestion
  PRECICE_DEBUG("MeshRequirement: {}", fmt::streamed(context.meshRequirement));
```

Or we need to add this new template specialization below:

https://github.com/precice/precice/blob/0590b349df3e28e0a9a99e4e5e1a05256565d494/src/mapping/Mapping.hpp#L191

src/utils/EigenIO.hpp
```diff
@@ -24,3 +24,8 @@ inline Eigen::IOFormat debug()
 } // namespace eigenio
 } // namespace utils
 } // namespace precice
+
+template <>
```
fsimonis, line 28: Shouldn't this be defined in the [`utils/fmtEigen.hpp`](https://github.com/precice/precice/blob/develop/src/utils/fmtEigen.hpp) ?

src/utils/Parallel.cpp
```diff
@@ -568,4 +569,8 @@ std::ostream &operator<<(std::ostream &out, const Parallel::CommState &value)
 } // namespace utils
 } // namespace precice
 
+template <>
+struct fmt::formatter<precice::utils::Parallel::CommState> : ostream_formatter {
+};
+
```
fsimonis, line 575: Shouldn't this be defined in the header?

fsimonis (changes requested):

src/mesh/Mesh.cpp
DavidSCN: No, I removed it.

src/math/geometry.cpp
DavidSCN: There is apparently a `const` issue with the logger (which is defined in the source file?!). I don't get it fully:
```cpp
error: passing ‘const formatter_type’ {aka ‘const fmt::v9::formatter<Eigen::Matrix<double, -1, 1>, char, void>’} as ‘this’ argument discards qualifiers [-fpermissive]
```

fsimonis: Sounds like `format()` needs to be `const` to allow composition?

Could you try?

```diff
-   auto format(const std::array<T, n> &v, FormatContext &ctx) 
+   auto format(const std::array<T, n> &v, FormatContext &ctx) const
```

DavidSCN: Ah now I got it. Yes, works! Do we need to update the other fmtSTL formatter to be `const` as well? Either we are not using them or we don't need to update, I guess it is the first one.

DavidSCN: Doesn't work either.

DavidSCN: I think the real issue here is that we have a nested sequence of array-like structures, namely a Vector of arrays. Using a simple array works with the current implementation.

fsimonis: I found a solution and pushed it.
The entire chain needs to be const.

fsimonis (approved):
I pushed some finishing touches.
This looks ready now.

PR #1398, opened 2022-08-02
Improve revision detection

### LINKED ISSUE #1397: Getting the preCICE revision from Git while in a third-party repository
## Context

When we build preCICE from source, we print the preCICE version information as, e.g.:

```
---[precice]  This is preCICE version 2.4.0
---[precice]  Revision info: v2.4.0-67-gb8beae6e
```
in CMake and in the beginning of the simulation. This essentially means:

> This is preCICE v2.4.0 (hard-coded in a file). More specifically, it is preCICE v2.4.0 (via a tag) + 94 commits, with the latest one being b8beae6e.

## Issue

While looking at some logs (@carme-hp), I got a very confusing output, which looked essentially like this:

```
---[precice]  This is preCICE version 2.4.0
---[precice]  Revision info: v1.3.0-<number>-g<hash>
```

I was very puzzled, thinking that this may be some version mix-up. After some time, I realized that preCICE was included as a dependency in the repository of the respective adapter, which was itself at its own v1.3.0.

This is probably a corner case (tiny benefit), but maybe it is also very easy to fix (tiny effort). Still, I can imagine that there are projects that are organized like this for reproducibility reasons.

### Solution A

One idea would be to extend the message to explain how the revision info is computed. For example:

```
---[precice]  This is preCICE version 2.4.0
---[precice]  Revision info (<latest git tag>-<commits since>-g<commit hash>): v2.4.0-67-gb8beae6e
```
This does not really fix it, but at least gives a hint that could help the user understand such cases.

### Solution B

Another idea would be to extend the check to only run when this is really the preCICE repository. Not sure how exactly to determine that, though.

https://github.com/precice/precice/blob/6d27ad5db42997bbc06ff97b39eb4999196295b7/cmake/cmake_refresh_git_revision.cmake#L15

fsimonis:
## Main changes of this PR

This PR improves the git revision detection by checking if preCICE itself is a git repository.

Unsuccessful git revisions are now also more descriptive:

Old:
```
no-info [Git failed/Not a repository]
```

New:
```
no-info [git failed to run]
no-info [not a repository]
```

Makes revisions easier to understand.
Prevents preCICE picking up revisions of super-projects.

Closes #1397

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/cmake/cmake_refresh_git_revision.cmake
+++ b/cmake/cmake_refresh_git_revision.cmake
@@ -32,7 +32,7 @@ set(preCICE_REVISION "no-info [git failed to run]")
 
 if(("${PRECICE_REVISION_RET}" EQUAL "0") AND ("${PRECICE_REPO_RET}" EQUAL "0"))
   file(TO_CMAKE_PATH "${PRECICE_REPO_OUT}" _detected_path)
-  if("${CMAKE_SOURCE_DIR}" STREQUAL " ${_detected_path}")
+  if("${CMAKE_SOURCE_DIR}" STREQUAL "${_detected_path}")
     set(preCICE_REVISION "${PRECICE_REVISION_OUT}")
     message(STATUS "Revision status: ${preCICE_REVISION}")
   else()
--- a/cmake/cmake_refresh_git_revision.cmake
+++ b/cmake/cmake_refresh_git_revision.cmake
@@ -11,6 +11,15 @@
 #
 
 find_package(Git REQUIRED)
+
+execute_process(
+  COMMAND ${GIT_EXECUTABLE} rev-parse --show-toplevel
+  RESULT_VARIABLE PRECICE_REPO_RET
+  OUTPUT_VARIABLE PRECICE_REPO_OUT
+  OUTPUT_STRIP_TRAILING_WHITESPACE
+  ERROR_QUIET
+  )
+
 execute_process(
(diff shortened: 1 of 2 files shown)


MakisH (commented):
Nice! :smile: 

Running inside a third-party repository gives:
```
---[precice]  This is preCICE version 2.4.0
---[precice]  Revision info: no-info [not a repository]
```

However, running normally also gives the same. I hope I am doing something wrong, but it looks like I am linking to the right version.

fsimonis:
> However, running normally also gives the same. I hope I am doing something wrong, but it looks like I am linking to the right version.

:eyes: 

Can you run `git describe --tags --dirty` and `git rev-parse --show-toplevel` in that repo?

MakisH:
> Can you run `git describe --tags --dirty` and `git rev-parse --show-toplevel` in that repo?

```
~/github/precice/build [improve-revision-detection]$ git describe --tags --dirty
v2.4.0-77-g213049d6
~/github/precice/build [improve-revision-detection]$ git rev-parse --show-toplevel
/home/makish/github/precice
```

fsimonis:
@MakisH Who committed this space that was used to test this locally? :roll_eyes:

MakisH (approved):
Works correctly for both cases now :+1:

DavidSCN (approved):

PR #1402, opened 2022-08-04
Fix various typos

luzpaz:
## Main changes of this PR

Various typo fized

Typolessness 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/extras/bindings/fortran/include/precice/SolverInterfaceFortran.hpp
+++ b/extras/bindings/fortran/include/precice/SolverInterfaceFortran.hpp
@@ -49,7 +49,7 @@ void precicef_initialize_(double *timestepLengthLimit);
 
 /**
  * Fortran syntax:
- * precicef_intialize_data()
+ * precicef_initialize_data()
  *
  * IN: -
  * OUT: -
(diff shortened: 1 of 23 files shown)


MakisH (approved):
Looks good to me, all changes are in messages and comments and the suggestions look correct.

extras/bindings/fortran/include/precice/SolverInterfaceFortran.hpp
```diff
@@ -49,7 +49,7 @@ void precicef_initialize_(double *timestepLengthLimit);
 
 /**
  * Fortran syntax:
- * precicef_intialize_data()
+ * precicef_initialize_data()
```
DavidSCN, line 52: @MakisH not sure if we really want to change this. The function will be removed soon anyway, but it's still a bit odd to change it before.

MakisH: I don't see why not fixing the comment in any case. It does not rename the function anyway. Zero benefit / zero effort, so I see it as undefined behavior.

DavidSCN: Ah sorry, my fault, I thought it was the signature :see_no_evil:

PR #1403, opened 2022-08-05
Simplify signature of actions and remove ScaleByDtAction

### LINKED ISSUE #934: ScaleByDtAction with SCALING_BY_DT is misleading

### LINKED ISSUE #1358: Simplify signature of actions

BenjaminRodenberg:
## Main changes of this PR

Closes #1358 and closes #934.

See #1358.

## Author's checklist

## Reviewers' checklist


docs/changelog/1403.md
```diff
@@ -0,0 +1 @@
+- Removes ScaleByDtAction and (internally) simplifies signature of all actions for easier maintainment and simpler code. See #1358 and #934.
```
uekerman, line 1:
```suggestion
- Removed action `ScaleByDtAction` and (internally) simplified signature of all actions for easier maintaining and simpler code.
```

uekerman (approved):
Looks good to me

PR #1404, opened 2022-08-05
Make read mapping conditional.

### LINKED ISSUE #834: Check unconditional read mapping in initializeData

BenjaminRodenberg:
## Main changes of this PR

Closes #834.

Expose `CouplingScheme::sendsInitializedData` to the public.
Check for `sendsInitializedData` to determine whether mapping is necessary.

See #834.

## Author's checklist

* [x] ~~I added a changelog file with `make changelog` if there are user-observable changes since the last release.~~ Not visible for users.
* [x] ~~I added a test to cover the proposed changes in our test suite.~~ I updated an existing test.

## Reviewers' checklist


tests/serial/action-timings/ActionTimingsExplicit.cpp
```diff
@@ -62,8 +62,8 @@ BOOST_AUTO_TEST_CASE(ActionTimingsExplicit)
 
   if (context.isNamed("SolverOne")) {
     BOOST_TEST(action::RecorderAction::records.size() == 2);
-    BOOST_TEST(action::RecorderAction::records.at(0).timing == action::Action::WRITE_MAPPING_PRIOR);
-    BOOST_TEST(action::RecorderAction::records.at(1).timing == action::Action::WRITE_MAPPING_POST);
+    BOOST_TEST(action::RecorderAction::records.at(0).timing == action::Action::READ_MAPPING_PRIOR);
+    BOOST_TEST(action::RecorderAction::records.at(1).timing == action::Action::READ_MAPPING_POST);
```
BenjaminRodenberg, line 66: I added `initialize="True"` in this test to actually check whether the branch is entered. The first participant then only reads the initial data. It does now write data at this point.

tests/serial/action-timings/ActionTimingsImplicit.cpp
```diff
@@ -63,21 +63,15 @@ BOOST_AUTO_TEST_CASE(ActionTimingsImplicit)
 
   dt = interface.initialize();
   BOOST_TEST(dt == 1.0);
-
   if (context.isNamed("SolverOne")) {
-    BOOST_TEST(action::RecorderAction::records.size() == 2);
-    BOOST_TEST(action::RecorderAction::records.at(0).timing == action::Action::WRITE_MAPPING_PRIOR);
-    BOOST_TEST(action::RecorderAction::records.at(1).timing == action::Action::WRITE_MAPPING_POST);
+    BOOST_TEST(action::RecorderAction::records.empty());
```
BenjaminRodenberg, line 67: Without initialization, Nothing actually happens here for the first participant.

tests/serial/action-timings/ActionTimingsImplicit.cpp
```diff
@@ -63,21 +63,15 @@ BOOST_AUTO_TEST_CASE(ActionTimingsImplicit)
...
   } else {
     BOOST_TEST(context.isNamed("SolverTwo"));
-    BOOST_TEST(action::RecorderAction::records.size() == 4);
-    BOOST_TEST(action::RecorderAction::records.at(0).timing == action::Action::WRITE_MAPPING_PRIOR);
-    BOOST_TEST(action::RecorderAction::records.at(1).timing == action::Action::WRITE_MAPPING_POST);
-    BOOST_TEST(action::RecorderAction::records.at(2).timing == action::Action::READ_MAPPING_PRIOR);
-    BOOST_TEST(action::RecorderAction::records.at(3).timing == action::Action::READ_MAPPING_POST);
+    BOOST_TEST(action::RecorderAction::records.size() == 2);
+    BOOST_TEST(action::RecorderAction::records.at(0).timing == action::Action::READ_MAPPING_PRIOR);
+    BOOST_TEST(action::RecorderAction::records.at(1).timing == action::Action::READ_MAPPING_POST);
```
BenjaminRodenberg, line 72: Second participant only reads (again: no intialization.)

tests/serial/action-timings/ActionTimingsExplicit.cpp
uekerman: > It does now write data at this point

It does no**t** write data at this point ??

uekerman (approved):
Looks good to me.
Probably worth a changelog entry.

tests/serial/action-timings/ActionTimingsExplicit.cpp
BenjaminRodenberg: Yes. Important difference!

Issue #1393, opened 2022-08-01
precice-tools: Segmentation fault and lack of messages in debug builds


ajaust:
Thanks offering the `precice-tools`. I find them very useful. While checking some preCICE configuration files using `precice-tools check` I ran into two issues:

1. Using `precice-tools check` does not provide error messages if a **release** build of preCICE is used. I have compiled preCICE from source. It might be worth checking if this applies to the provided `deb` packages as this renders the `check` feature useless.
2. Running the checks on an invalid preCICE configuration leads a segmentation fault in `precice-tools check`. This happens in **release** and **debug** builds.

**Describe your setup**

Operating system (e.g. Linux distribution and version): _Ubuntu 20.04_
preCICE Version: 2.4.0 (built from source)

**Describe the problem**

If checking an invalid preCICE configuration with a **release** build of preCICE I get the following.

```text
$ precice-tools check ./dumux-precice/examples/dummysolver/precice-dummy-solver-config.xml_segmentation_fault
Segmentation fault (core dumped)
```

The tool does not generate helpful output and quits with a segmentation fault.

If I check the configuration with a **debug** build, I get the following output:

```text
$ precice-tools check ./dumux-precice/examples/dummysolver/precice-dummy-solver-config.xml_segmentation_fault
ASSERTION FAILED
Location:   const PtrData& precice::mesh::Mesh::data(const string&) const
      /home/jaustar/software/compilescripts/precice/precice-2.4.0/src/mesh/Mesh.cpp:177
Expression: iter != _data.end()
Rank:       0
Arguments:  
  0: "Data not found in mesh" == Data not found in mesh
  1: dataName == dataOne
  2: _name == MeshOne

Stacktrace:
 0# getStacktrace[abi:cxx11]() in /home/jaustar/software/precice/2.4.0-debug-mpi-petsc-python/lib/libprecice.so.2
 1# precice::mesh::Mesh::data(std::__cxx11::basic_string<char, std::char_traits<char>, std::allocator<char> > const&) const in /home/jaustar/software/precice/2.4.0-debug-mpi-petsc-python/lib/libprecice.so.2
[...]

PR #1406, opened 2022-08-05
Add a success statement to the config check

### LINKED ISSUE #1405: No output running precice-tools check on clean config

Running `precice-tools check file.xml` only outputs warnings and errors, hence a clean config outputs nothing at all.

Print a success statement if the config is correct.

_none_

**Additional context**

preCICE version: 2.4
Original report: #1393

fsimonis:
## Main changes of this PR

Adds a success statement to `precice-tools check`.

Checking correctly formatted file will now result in
```
$ precice-tools check valid-config.xml
Checking valid-config.xml for syntax and basic setup issues...
No major issues detected
```

I used the new style support from libfmt to format the last line bold and green.

Closes #1405
See #1393 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/Tooling.cpp
+++ b/src/precice/Tooling.cpp
@@ -1,7 +1,9 @@
 #include "precice/Tooling.hpp"
 
+#include "fmt/color.h"
 #include "precice/config/Configuration.hpp"
 #include "precice/impl/versions.hpp"
+#include "utils/fmt.hpp"
 #include "xml/Printer.hpp"
 
 namespace precice {
@@ -34,6 +36,7 @@ void checkConfiguration(const std::string &filename, const std::string &particip
       0,
       size};
   xml::configure(config.getXMLTag(), context, filename);
+  fmt::print("{}: {}\n", filename, fmt::styled("correct", fmt::emphasis::bold | fg(fmt::color::green)));
 }
 
 } // namespace tooling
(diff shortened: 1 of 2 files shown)


src/precice/Tooling.cpp
```diff
@@ -34,6 +36,7 @@ void checkConfiguration(const std::string &filename, const std::string &particip
       0,
       size};
   xml::configure(config.getXMLTag(), context, filename);
+  fmt::print("{}: {}\n", filename, fmt::styled("correct", fmt::emphasis::bold | fg(fmt::color::green)));
```
MakisH, line 39: That's nice! We could, however, give some hint about the scope of the checking. If people get a "correct" here and an error during their actual simulation, this may be confusing.

fsimonis: Could you propose something? I am not sure how to phrase this well.

MakisH: Maybe just add some very short description in the beginning on what this tool does. For example:

```
$ precice-tools check valid-config.xml
Checking valid-config.xml for syntax and basic setup issues... (see https://precice.org/tooling-builtin.html)
valid-config.xml: correct
```

fsimonis: Good idea. I'll leave the link out though.

fsimonis: See OP for new message.

Issue #1171, opened 2022-02-01
Write data, when subcycling is used


BenjaminRodenberg:
**Please describe the problem you are trying to solve.**

Currently only data at the end of the coupling window is exchanged. This means that all calls of `write...data(...)` are ignored except the last one before `advance()` is called and the end of the window is reached. We could improve the accuracy by using more / all data that is created and written by the individual participants.

**Describe the solution you propose.**

Currently, I see the following steps:

1. Do a linear interpolation on the window from the data at the beginning and at the end of the window. Ignore data inside of the window. This does not require any updates w.r.t `write...data(...)` and data communication. See #1029 
2. Do a linear interpolation between all substeps, if subcycling is used. This requires additional communication and will change the behavior of the API, because every `write...data()` before calling `advance()` will be used by preCICE. This is especially important for subcycling. We will still ignore multiple `write...data()` calls within one time **step**. This will not require an update of the API, because we can implicitly assume that calling `write...data(...)` will provide data associated with the end of the current time step. Note: The way how we want to perform interpolation is still an open question. I can also imagine using higher order interpolation with subcycling instead of piecewise linear interpolation.
3. Support higher-order interpolation, even if no subcycling is used. If we change the API to something like `write...data(..., dt)` a user can provide time stamped samples and therefore, we could provide multiple samples per step. This is, however, a far range goal and currently out-of-scope.

**Additional context**

In https://onlinelibrary.wiley.com/doi/10.1002/nme.6443, the approach described in (2) is implemented by using a wrapper around preCICE. Main goal of this issue is to bring this feature natively into preCICE.

uekerman:
This does not necessarily have to be a breaking change, leaving option 3 aside. Only the behavior of `isWriteDataRequired` would change, which we could enable in the configuration. I currently don't see how option 2 changes the behavior of the API. preCICE just does sth different with the same data. Just like using a different data mapping does not change the behavior of the API, no?

BenjaminRodenberg:
One important question that we will have to answer: Is a piecewise interpolation (zeroth order?) always used when subcycling is used? There are two possibilities:

[...]

PR #1414, opened 2022-08-11
**OUTDATED** writeData when subcycling is used

BenjaminRodenberg:
## Main changes of this PR

Allows to write data to preCICE when subcycling is used. This either results in a piecewise constant or linear interpolation. Possibly also in higher-order interpolation.

See #1171 

## Author's checklist

* [x] merge #1422 
* [x] merge #1445 
* [x] merge https://github.com/precice/precice/pull/1455
* [x] merge https://github.com/precice/precice/pull/1456 ? (or directly do this here?)
* [x] what to do with #1432 (Continue working on https://github.com/BenjaminRodenberg/precice/pull/13 until then)
* [ ] merge https://github.com/precice/precice/pull/1507
* [ ] merge #1504 
* [x] merge #1503 
* [ ] Add tests for MultiCoupling (see https://github.com/precice/precice/pull/1414#issuecomment-1260831100)

## Reviewers' checklist


BenjaminRodenberg:
Summary of important changes of today:

* I'm using Eigen's BSpline interpolation for higher order. This needs a sufficient number of samples on the window. If there is an insufficient number of sample, we will fall back to lower order interpolation on the window.
* I removed the feature to use older samples (only properly implemented for quadratic interpolation and not accessible through API), because it made the implementation much more complicated and has no use-case.

BenjaminRodenberg:
Next steps:

1. store write data every time when advance is called in some buffer in `CouplingScheme` or `CouplingData`. This will probably be another dict.
2. communicate this data to the other participant at the end of the window. Store this data as read data in `CouplingScheme` or `CouplingData`.
3. Tricky part: We have to map every sample that the coupling scheme received and store it in the waveform of the read data. Mapping happens outside of `CouplingScheme` and uses `Mesh::Data`. However, I would rather not touch `Mesh::Data`, because of the many dependencies. Current plan: Use `Mesh::Data` as a buffer for each piece of data from the `CouplingSchemes`. This need some iteration and jumping back and forth between mapping and `CouplingScheme`, but sounds to me like the solution with the least effort.

BenjaminRodenberg:
Steps 1 and 2 described above are now basically complete. Still needs some refactoring (probably could move the whole storing of time steps into an independent class and encapsulate components of coupling scheme a bit better).

Step 3 is now the next goal. With the public function `BaseCouplingScheme::retreiveTimeStepData` it should be possible to let the `SolverInterfaceImpl` load the time steps one by one, map them and put them into the read waveform.

BenjaminRodenberg:
All tests are working now and subcycling is implemented :tada: 

Most todos are now on the code side, feature-wise it should be complete:

- [ ] Add some more tests to test also higher order subcycling (esp. third order with BSplines)
- [ ] Try to remove some of the helper functions and variables (e.g. `_atLeastOneWasReceived`)
- [ ] Biggest todo: A write mapping should work in the context of multi coupling, even, if the sent meshes in time do not agree. For making a read mapping in this situation work as well, we need to be able to refer to received times etc. in an individual fashion (currently a coupling scheme only gives access to the received times globally). See [here](https://github.com/precice/precice/blob/12296ee452c6e64ee645252666d91461551b4723/src/cplscheme/BiCouplingScheme.cpp#L183-L193)
- [ ] Improve encapsulation of time window storage and put serialization / deserialization into a dedicated class (see [here](https://github.com/precice/precice/blob/12296ee452c6e64ee645252666d91461551b4723/src/cplscheme/BaseCouplingScheme.cpp#L114))

src/cplscheme/BiCouplingScheme.cpp
```diff
@@ -148,4 +157,48 @@ void BiCouplingScheme::exchangeInitialData()
...
+    aSendData.second->storeDataAtTime(theData, relativeDt);
+  }
+}
+
+void BiCouplingScheme::storeTimeStepReceiveDataEndOfWindow()
+{
+  if (hasDataBeenReceived()) {
+    for (auto &aReceiveData : getReceiveData()) {
+      auto theData = aReceiveData.second->values();
+      aReceiveData.second->overrideDataAtEndWindowTime(theData);
```
uekerman, line 175:
```suggestion
      receiveData.second->overrideDataAtEndOfWindow(values);
```
Some naming suggestions

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -979,12 +979,6 @@ void CouplingSchemeConfiguration::addDataToBeExchanged(
...
-      // Checks for a serial coupling scheme, where initial data is received by first participant.
-      if (!scheme.doesFirstStep() && requiresInitialization && (_config.type == VALUE_SERIAL_EXPLICIT || _config.type == VALUE_SERIAL_IMPLICIT) && getWaveformUsedOrder(accessor, dataName) == 0) // the order of the statements is crucial! getWaveformUsedOrder should be called last, because it may run into an error, if the data is initialized, but not defined as read data (for example Integration/Serial/AitkenAcceleration)
-      {
-        PRECICE_WARN("In serial coupling, initialized data received by the second participant will only be used if a waveform order > 0 is set. You can avoid unnecessary data initialization by setting initialize=\"false\" in the <exchange data=\"{}\" mesh=\"{}\" from=\"{}\" to=\"{}\" initialize=\"{}\" /> tag in the <coupling-scheme:... /> of your precice-config.xml. If you want to use the initial data, please consider increasing the waveform-order in the <read-data name=\"{}\" mesh=\"{}\" /> tag in the <participant ... /> named \"{}\". ",
-                     dataName, meshName, from, to, requiresInitialization, dataName, meshName, to);
-      }
```
uekerman, line 987: Why can you remove this here?

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -190,11 +256,43 @@ void BaseCouplingScheme::advance()
...
+
+  if (isImplicitCouplingScheme()) {
+    // store data from writeDataContext in buffer, when advance is called.
+    PRECICE_ASSERT(_computedTimeWindowPart > 0);
+    if (not usesFirstParticipantMethod) {
+      relativeDt = _computedTimeWindowPart / _timeWindowSize;
+      PRECICE_ASSERT(math::smallerEquals(relativeDt, 1.0, 10e-9), relativeDt, _computedTimeWindowPart, _timeWindowSize);
+      PRECICE_ASSERT(relativeDt > 0, relativeDt, _computedTimeWindowPart, _timeWindowSize);
+      if (relativeDt > 1.0) {
+        PRECICE_ASSERT(math::smallerEquals(relativeDt, 1.0, 10e-9), relativeDt);
```
uekerman, line 270: the 1e-9 seems arbitrary here :grin:
what are you trying to do?

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -190,11 +256,43 @@ void BaseCouplingScheme::advance()
...
+        relativeDt = 1.0;
+      }
+      storeTimeStepSendData(relativeDt);
+    } else {
+      // We don't support subcycling here, because this is complicated. Therefore, use same strategy like for explicit coupling and just use a single value at end of window.
+      // Possible solution: Don't scale times to [0,1], but leave them as they are. Then we would also allow times > 1. We then have two options:
+      // 1) scale the times back later when the time window size is known (to still benefit from the simpler handling, if all times are scaled to [0,1]).
+      // 2) generally use times in the interval [0, timeWindowSize]. This makes the implementation probably a bit more complicated, but also more consistent.
```
uekerman, line 278: I think I understand the problem here. Not sure what the better solution is, but is would be nice to somehow also cover this case.

uekerman (commented):
Rather readable already :+1:
We need to improve names, short and concise please :grin:

src/cplscheme/BiCouplingScheme.cpp
BenjaminRodenberg: Done in ced11922ff5ec01e43af3cec9bb8bf191b85d8d1.

src/cplscheme/config/CouplingSchemeConfiguration.cpp
BenjaminRodenberg: Good question. I think this change should already have been implemented in #1422. I will try to remove it from this PR and put it into an independent PR/commit. I first have to check this branch and do some experiments to make sure. 

The important thing is the following: zeroth-order interpolation has a jump at the border of two windows. In pre-subcycling and pre-waveform times we did not really worry a lot here and just assumed a constant value for the whole window. If we have two windows: window $AB$ spans $[a,b]$ and window $BC$ spans $[b,c]$, the two windows overlap at $b$. It was possible that on $AB$ and $BC$ we get different values $f_{AB}(b) \neq f_{BC}(b)$ at $b$, because there was just a constant value stored on each window.

With #1422 this changed, because the values get a time stamp and therefore we need consistent behavior. Meaning: We have to decide on one value at b that we use on $AB$ and $BC$. Here, I decided that for zeroth order interpolation $t \in (a,b]$ gives the value associated with $b$ and $t\in (b, c]$ gives the values associated with $c$. Important question: What value do we get at $a$? If we don't initialize the data (independent from the coupling scheme) we have a gap here (set $f(x) = 0$). If we want to allow a value at $f(a) \neq 0$, we have to be able to initialize data for all participants and all coupling schemes.

Important remark: All this is only visible for a user that explicitly asks for the value at the beginning of the window by calling `read...Data` with `relativeDt = 0`. Since `relativeDt` defaults to the end of the window a user that does not use the (experimental) waveform feature will not have to worry about any of these technicalities.

src/time/Storage.cpp
```diff
@@ -15,18 +15,42 @@ void Storage::initialize(Eigen::VectorXd values)
...
 }
 
+void Storage::overrideDataAtEndWindowTime(Eigen::VectorXd data)
+{
+  double endWindowTime = 1.0;
+  if (_sampleStorage.size() == 0) {
+    _sampleStorage.emplace_back(std::make_pair(endWindowTime, data));
+  } else {
+    PRECICE_ASSERT(math::equals(_sampleStorage.back().first, endWindowTime), "Unexpected!", _sampleStorage.back().first);
+    _sampleStorage.back().second = data;
```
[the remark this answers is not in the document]
BenjaminRodenberg: I'm not sure what you mean here. This is exactly what's happening in this line:
```
PRECICE_ASSERT(math::equals(_sampleStorage.back().first, endWindowTime), "Unexpected!", _sampleStorage.back().first);
```
If the storage is not empty we are only allowed to call this function when the last time stamp is equal to `endWindowTime`.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -410,16 +410,19 @@ double SolverInterfaceImpl::advance(
...
 
   PRECICE_DEBUG("Advance coupling scheme");
   _couplingScheme->advance();
 
-  if (_couplingScheme->isTimeWindowComplete()) {
+  if (_couplingScheme->isTimeWindowComplete() && _couplingScheme->moveWindowBeforeMapping()) { /*
+     * move to next window before mapping, if second in serial coupling.
+     * Reason:
+     * Window is complete, but first participant has already performed an iteration in the new window and the data just received is already from the new window. So we want to use it, but before doing this we have to use the data from the old window to initialize the waveform at 0.0.
+     */
```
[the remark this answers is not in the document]
BenjaminRodenberg: Moving the `moveToNextWindow()` call inside the coupling scheme is complicated, because we have to be able to access the `ReadDataContext` where the waveform lives in order to tell it to move to the next window. Alternatively, we could give the coupling scheme access to the `ReadDataContext`, but this would lead to more interaction between the different modules than necessary and lead to some degree of spaghettification. I would rather keep this solution, because here the `SolverInterfaceImpl` clearly stays in control and there is no direct interaction of the coupling scheme and the waveform.

src/cplscheme/config/CouplingSchemeConfiguration.cpp
BenjaminRodenberg: I added a PR: https://github.com/precice/precice/pull/1455 and a todo above to first conclude on https://github.com/precice/precice/pull/1455.

@uekerman feel free to resolve this conversation.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: I removed a large part of this code and it still works. I assume this is a left-over from the "use a double as dict key"-times and the code has become unnecessary. See fb4d0f85d5de99f9b844bf68aad84ee7c47c7122.

src/precice/impl/SolverInterfaceImpl.cpp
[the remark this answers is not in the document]
uekerman: > So we want to use it, but before doing this we have to use the data from the old window to initialize the waveform at 0.0.

Isn't there a solution where we do this at the end of the previous time window. So in the call below?
Let's draw a picture together :blush: 

Anyway, this surely will give problems with `CompostionalCouplingScheme`. We need a different solution, sorry :confused:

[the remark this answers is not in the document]
BenjaminRodenberg: Then we should also try to come up with an integration test for the compositional coupling scheme + time interpolation. This makes it easier to agree on what we want and it also makes sure that the feature will not break in the future without noticing it.

Moving this to the previous time window: Not sure. It's definitely complicated and I don't see a solution with digging deep into the problem.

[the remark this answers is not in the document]
BenjaminRodenberg: I just did two brief experiments:

1) Completely removing the block `if (_couplingScheme->isTimeWindowComplete() && _couplingScheme->moveWindowBeforeMapping()) { ... }`
2) Moving the block `if (_couplingScheme->isTimeWindowComplete() && _couplingScheme->moveWindowBeforeMapping()) { ... }` before `_couplingScheme->advance()`

In both cases we get failing tests for `mpirun -np 4 ./testprecice -t Integration/Serial/Time/Implicit/SerialCoupling`. I already expected this and it means:

* The feature is actually needed
* The convergence communication that happens inside advance is needed for the correct evaluation of `_couplingScheme->moveWindowBeforeMapping()`
* It's unfortunately not that simple ;)

@uekerman and @fsimonis you are currently also working on splitting up advance, correct? Might this be helpful in this context? I think generally there should not be a big problem here and all the information that we need about convergence is available when we need it. It's still a bit tricky with all the communication going on.

BenjaminRodenberg:
Current state:

* Most of the Integration tests for time interpolation are working again
* I did not check the unit tests thoroughly, but there are still some issues, for example in `CplSchemeTests`. Probably minor.
* Code structure is in my opinion much more understandable now, because we require less going back and forth between `CouplingScheme` and `SolverInterfaceImpl`. (I managed to remove `CouplingScheme::receiveResultOfFirstAdvance`, everything happens now directly in `CouplingScheme::initialize()`)

Next steps:

* Get the remaining tests working again (some tests in `MultiCoupling` are optional, because this is basically a new feature).
* Try to also remove `CouplingScheme::moveWindowBeforeMapping`
* Implement Extrapolation again, but inside Waveform.
* Try to factor out changes (refactoring `SolverInterfaceImpl`/`CouplingScheme` and `Extrapolation`) into a smaller PRs.

BenjaminRodenberg:
I tried to merge #1504 into this PR in https://github.com/BenjaminRodenberg/precice/commit/3f35b8cccaa3076fae6e9442b030af6312c5a877, but there are a lot of conflicts and I think it's probably easier to reimplement the changes from #1414 on top of #1504 instead of trying to fix the merge.

BenjaminRodenberg:
This PR is outdated and won't be merged. I will port all relevant work from here to https://github.com/precice/precice/pull/1523 and then close this PR.

BenjaminRodenberg:
Closing this PR, development will continue in #1523

PR #1420, opened 2022-08-23
Improve compatibility with applications using Boost.Log

### LINKED ISSUE #1305: Logger conflicts with other users of boost.log
**Describe your setup**

preCICE Version: 2.4.0

**Describe the problem**

preCICE uses the boost.log in a way, which collides with third-party applications and libraries.

During the preCICE Workshop, @mtree22 discovered that some boost calls in preCICE conflict with Fluent in our Fluent adapter. https://github.com/precice/fluent-adapter/pull/24
In ASTE, @kursatyurt discovered that preCICE overrides the formatting set by ASTE.  https://github.com/precice/aste/issues/73

preCICE currently relies a lot on the global state of the core logger instead of specifying attributes and the like in custom loggers.
It also removes all existing formatters, which is likely due to the testing framework.
The latter shouldn't be an issue though, as the testing framework uses a configuration and then locks the preCICE internals to prevent reconfiguration of the logging facilities.

**Expected behaviour**

The configuration of preCICE should not influence, nor conflict the logging mechanisms of other programs and libraries also relying on boost.log.

**Additional context**

The [Boost.log Design overview](https://www.boost.org/doc/libs/1_79_0/libs/log/doc/html/log/design.html)

fsimonis:
## Main changes of this PR

This PR mainly
- Reduces logger dependence on `boost::log::core`. preCICE will not change existing configuration. 
- Adds the log attributes `preCICE` (always true) for tagging preCICE log entries as well as `Runtime` as it may come in handy.
- Updates the logger to use a "feature" for logging `LogLocation` information.
- Extend the documentation of `<log>`
- Fix `<log enable="false">` being ignored.

These changes allow handle cases where the application/solver/adapter uses Boost.Log itself. (Including ASTE https://github.com/precice/aste/issues/73)
Log entries can now be filtered internally (so preCICE sinks only display preCICE messages) and by the application (using ` and not %preCICE%` in the filter).

Closes #1305 
Related to https://github.com/precice/aste/pull/136

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/logging/LogConfiguration.cpp
+++ b/src/logging/LogConfiguration.cpp
@@ -4,6 +4,7 @@
 #include <boost/log/attributes/mutable_constant.hpp>
 #include <boost/log/core.hpp>
 #include <boost/log/expressions.hpp>
+#include <boost/log/sinks/sink.hpp>
 #include <boost/log/support/date_time.hpp>
 #include <boost/log/trivial.hpp>
 #include <boost/log/utility/setup/console.hpp>
@@ -72,13 +73,39 @@ public:
   }
 };
 
+/// A custom sink that does nothing. It is used to disable the default sink of boost.Log
+class NullSink final : public boost::log::sinks::sink {
+public:
+  NullSink()
+      : boost::log::sinks::sink(false) {}
+
+  bool will_consume(boost::log::attribute_value_set const &) override
+  {
+    return false;
+  }
+
+  void consume(boost::log::record_view const &) override {}
+
+  bool try_consume(boost::log::record_view const &) override
+  {
+    return false;
+  }
+
+  void flush() override {}
+
+  bool is_cross_thread() const noexcept
+  {
+    return false;
+  }
+};
+
 /// A simple backends that outputs the message to a stream
 /**
  * Rationale: The original text_ostream_backend from boost suffered from the great amount of code that lies
  * between the printing of the message and the endline. This leads to high probability that a process switch
  * occurs and the message is severed from the endline.
  */
-class StreamBackend : public boost::log::sinks::text_ostream_backend {
+class StreamBackend final : public boost::log::sinks::text_ostream_backend {
 private:
   boost::shared_ptr<std::ostream> _ostream;
 
@@ -175,7 +202,7 @@ void setupLogging(LoggingConfiguration configs, bool enabled)
       << bl::expressions::message;
 
   // Remove active preCICE sinks
-  using sink_t   = typename boost::log::sinks::synchronous_sink<StreamBackend>;
+  using sink_t   = typename boost::log::sinks::sink;
   using sink_ptr = typename boost::shared_ptr<sink_t>;
 
   static std::vector<sink_ptr> activeSinks;
@@ -186,10 +213,14 @@ void setupLogging(LoggingConfiguration configs, bool enabled)
   }
   activeSinks.clear();
 
-  // If logging sinks are disabled, then we are done
+  // If logging sinks are disabled, then we need to disable the default sink.
+  // We do this by adding a NullSink.
   // We need to exit after the sink removal as the default sink exists before
   // the log configuration is parsed.
   if (!enabled) {
+    auto sink = boost::make_shared<NullSink>();
+    boost::log::core::get()->add_sink(sink);
+    activeSinks.emplace_back(std::move(sink));
     return;
   }
 
@@ -218,7 +249,7 @@ void setupLogging(LoggingConfiguration configs, bool enabled)
     backend->auto_flush(true);
 
     // Setup sink
-    auto sink = boost::make_shared<sink_t>(backend);
+    auto sink = boost::make_shared<boost::log::sinks::synchronous_sink<StreamBackend>>(backend);
     sink->set_formatter(boost::log::parse_formatter(config.format));
 
     if (config.filter.empty()) {
--- a/src/logging/LogConfiguration.cpp
+++ b/src/logging/LogConfiguration.cpp
@@ -177,6 +177,7 @@ void setupLogging(LoggingConfiguration configs, bool enabled)
   using sink_ptr = typename boost::shared_ptr<sink_t>;
 
   static std::vector<sink_ptr> activeSinks;
+
   //remove active sinks
   for (auto &sink : activeSinks) {
     boost::log::core::get()->remove_sink(sink);
@@ -185,9 +186,15 @@ void setupLogging(LoggingConfiguration configs, bool enabled)
   }
   activeSinks.clear();
 
+  // If logging sinks are disabled, then we are done
+  if (!enabled) {
+    return;
+  }
+
   // Add the default config
-  if (configs.empty())
+  if (configs.empty()) {
     configs.emplace_back();
+  }
 
   // Create new sinks
   for (const auto &config : configs) {
--- a/src/logging/LogConfiguration.cpp
+++ b/src/logging/LogConfiguration.cpp
@@ -211,7 +211,7 @@ void setupLogging(LoggingConfiguration configs, bool enabled)
     backend->auto_flush(true);
     boost::shared_ptr<sink_t> sink(new sink_t(backend));
     sink->set_formatter(boost::log::parse_formatter(config.format));
-    // We extend the filter here to remove all log entries not originating from preCICE.
+    // We extend the filter here to filter all log entries not originating from preCICE.
     sink->set_filter(boost::log::parse_filter("( " + config.filter + " ) and %preCICE%"));
     boost::log::core::get()->add_sink(sink);
     activeSinks.emplace_back(std::move(sink));
--- a/src/logging/LogConfiguration.cpp
+++ b/src/logging/LogConfiguration.cpp
@@ -145,9 +145,11 @@ void BackendConfiguration::setOption(std::string key, std::string value)
     filter = value;
   if (key == "format")
     format = value;
-  if (key == "enabled") {
-    enabled = utils::convertStringToBool(value);
-  }
+}
+
+void BackendConfiguration::setEnabled(bool enabled)
+{
+  this->enabled = enabled;
 }
 
 void setupLogging(LoggingConfiguration configs, bool enabled)
@@ -200,6 +202,10 @@ void setupLogging(LoggingConfiguration configs, bool enabled)
 
   // Create new sinks
   for (const auto &config : configs) {
+    if (!config.enabled) {
+      continue;
+    }
+
     boost::shared_ptr<StreamBackend> backend;
     if (config.type == "file")
       backend = boost::make_shared<StreamBackend>(boost::shared_ptr<std::ostream>(new std::ofstream(config.output)));
--- a/src/logging/LogConfiguration.cpp
+++ b/src/logging/LogConfiguration.cpp
@@ -202,8 +202,7 @@ void setupLogging(LoggingConfiguration configs, bool enabled)
       << bl::expressions::message;
 
   // Remove active preCICE sinks
-  using sink_t   = typename boost::log::sinks::sink;
-  using sink_ptr = typename boost::shared_ptr<sink_t>;
+  using sink_ptr = typename boost::shared_ptr<boost::log::sinks::sink>;
 
   static std::vector<sink_ptr> activeSinks;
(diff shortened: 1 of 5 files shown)


src/logging/config/LogConfiguration.cpp
```diff
@@ -18,12 +18,13 @@ LogConfiguration::LogConfiguration(
   tagLog.setDocumentation("Configures logging sinks based on Boost log.");
 
   auto attrLogEnabled = makeXMLAttribute("enabled", true)
-                            .setDocumentation("Enables logging");
+                            .setDocumentation("Enables the creation of log sinks. Disable sinks if you plan to setup Boost.log in your application.");
```
DavidSCN, line 21: I'm not sure I get this sentence right: with the proposed solution, you could still use Boost.log in your own application without conflicting sinks. Why should I now disable the sinks now?

src/logging/config/LogConfiguration.cpp
```diff
@@ -18,12 +18,13 @@ LogConfiguration::LogConfiguration(
...
   tagLog.addAttribute(attrLogEnabled);
 
   XMLTag tagSink(*this, "sink", XMLTag::OCCUR_ARBITRARY);
   tagSink.setDocumentation("Contains the configuration of a single log sink, which allows fine grained control of what to log where. "
-                           "Available attributes in filter and format strings are `%Severity%`, `%ColorizedSeverity%`, `%File%`, `%Line%`, `%Function%`, `%Module%`, `%Rank%`, and `%Participant%`");
+                           "Available attributes in filter and format strings are `%TimeStamp%`, `%Runtime%`, `%Severity%`, `%ColorizedSeverity%`, `%File%`, `%Line%`, `%Function%`, `%Module%`, `%Rank%`, and `%Participant%`. "
+                           "The boolean attribute `%preCICE%` is `true` for all log entries originating from preCICE.");
```
DavidSCN, line 27: That's not visible from the user side, is it?

src/logging/Logger.cpp
```diff
@@ -1,53 +1,158 @@
...
+
+#include "logging/LogConfiguration.hpp"
+#include "logging/Logger.hpp"
 
 namespace precice {
 namespace logging {
 
+/// A feature that adds \ref LogLocation info to log attributes
+template <class BaseT>
+class precice_feature : public BaseT {
```
DavidSCN, line 19: Any reason to deviate from our naming convention? `PreciceFeature` ?

src/logging/Logger.cpp
```diff
@@ -1,53 +1,158 @@
...
+public:
+  using char_type       = typename BaseT::char_type;
+  using threading_model = typename BaseT::threading_model;
+
+  precice_feature()                             = default;
+  precice_feature(const precice_feature &other) = default;
```
DavidSCN, line 25: Do you intentionally leave out other SMFs? What about a destructor?

src/logging/Logger.cpp
```diff
@@ -1,53 +1,158 @@
...
+  template <typename ArgsT>
+  precice_feature(ArgsT const &args){};
+
+  using open_record_lock = typename boost::log::strictest_lock<
+      boost::lock_guard<threading_model>,
+      typename BaseT::open_record_lock,
+      typename BaseT::add_attribute_lock,
+      typename BaseT::remove_attribute_lock>::type;
+
+protected:
```
DavidSCN, line 35: Considering that we inherit things, we should add a virtual destructor.

src/logging/Logger.cpp
```diff
@@ -1,53 +1,158 @@
...
+
+  PRECICE_ASSERT(line_value >= 0);
+  PRECICE_ASSERT(!file_value.empty());
+  PRECICE_ASSERT(!func_value.empty());
+
+  // Process extracted tags
+  using boost::log::attributes::constant;
+  boost::log::attribute_set &attrs = BaseT::attributes();
+  {
+    auto res = BaseT::add_attribute_unlocked("Line", constant<int>(line_value));
```
DavidSCN, line 80: We could put the whole line into the ASSERT, in order to remove our CI warnings here (just a suggestion).

src/logging/Logger.cpp
```diff
@@ -1,53 +1,158 @@
...
+    PRECICE_ASSERT(res.second);
+  }
+  {
+    auto res = BaseT::add_attribute_unlocked("File", constant<std::string>(file_value));
```
DavidSCN, line 84: Same as above

src/logging/Logger.cpp
```diff
@@ -1,53 +1,158 @@
...
+    PRECICE_ASSERT(res.second);
+  }
+  {
+    auto res = BaseT::add_attribute_unlocked("Function", constant<std::string>(func_value));
```
DavidSCN, line 88: Same as above

src/logging/Logger.cpp
```diff
@@ -1,53 +1,158 @@
...
+    PRECICE_ASSERT(res.second);
+  }
+
+  ScopeExit guard([&attrs] {
+    attrs.erase("Line");
+    attrs.erase("File");
+    attrs.erase("Function");
+  });
```
DavidSCN, line 96: Instead of the `ScopeExit` solution here, couldn't we just define a lambda a few lines further up? Would be a bit more readable as the function is closer to the actual call.

DavidSCN (commented):
A few comments. I still have to test all the configuration options of the improved logger.

src/logging/LogConfiguration.cpp
```diff
@@ -210,19 +226,23 @@ void setupLogging(std::string const &logConfigFile)
 
 void setMPIRank(int const rank)
 {
-  boost::log::attribute_cast<boost::log::attributes::mutable_constant<int>>(boost::log::core::get()->get_global_attributes()["Rank"]).set(rank);
+  getGlobalLoggingConfig().rank = rank;
```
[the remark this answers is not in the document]
fsimonis: The lock prevents system tests from changing the configuration of the sink.
Changing the rank is not a problem, as the attribute will be generated from `getGlobalLoggingConfig().rank`.

src/logging/config/LogConfiguration.cpp
fsimonis: An application using Boost.log needs to register sinks.
These sinks will also receive preCICE logs, which is by design of Boost.log.
This application could use this boolean as an easy way to filter out everything from preCICE.

It is not relevant for the vast majority of people, but it could be relevant for some.
I also need to add documentation on this to our docs.

src/logging/Logger.cpp
fsimonis: It forces me to butcher our project name, which hurts my heart.
Alternatives are welcome :smiley:

src/logging/Logger.cpp
fsimonis: Good point. I could remove both of them, really.

src/logging/Logger.cpp
fsimonis: This would disable the code along the assert in Release :wink: 
Perfect candidate for `[[maybe_unused]]`! C++17 :tada:

src/logging/Logger.cpp
fsimonis: Not sure what you mean. Could you provide an example?

src/logging/Logger.cpp
fsimonis: This is compile-time inheritance using RCTP, which does not require a virtual dtor. It would also only be necessary in the base class.

src/logging/config/LogConfiguration.cpp
fsimonis: If an application uses boost.log then it has two choices:

* Use the preCICE provided sinks (default) and remove the preCICE log entries from the application-sinks to provide duplication.
* Or it can handle both the application and the preCICE logs, then it would need to disable the sinks of preCICE, which is what this attribute does.

Feel free to help with a better formulation. I find it confusing :sweat_smile:

DavidSCN: Maybe then `Disable sinks if you want to handle the preCICE logging in your application`? I guess a link to a dev doc or to boost log might be helpful.

src/logging/Logger.cpp
DavidSCN: Thinking about it: I mean the scope exit is right after this. Couldn't we just execute the function in place?

fsimonis: Actually, we could save the result, cleanup and then return, making this way easier.

fsimonis: Done. Thanks for complaining :smiley:

src/logging/config/LogConfiguration.cpp
fsimonis: I updated the text. I would really avoid links from the xmlrefence to the precice website. All of this behaviour should be in the real docs.

src/logging/LogConfiguration.cpp
```diff
@@ -173,16 +173,31 @@ void setupLogging(LoggingConfiguration configs, bool enabled)
...
+  for (auto &sink : activeSinks) {
+    boost::log::core::get()->remove_sink(sink);
+    sink->flush();
+    sink.reset();
+  }
+  activeSinks.clear();
+
+  // If logging sinks are disabled, then we are done
+  if (!enabled) {
+    return;
```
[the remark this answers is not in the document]
fsimonis: Is the updated comment clearer?

davidscn:
I just played around with the changes here:
Configuring the following:
```xml
  <log>
    <sink
      filter="%Severity% > debug and %Rank% = 0"
      format="---[precice] %ColorizedSeverity% %Message%"
      enabled="false" />
  </log>
```
leads to the following log messages (e.g.)
```
(0) 09:43:17 [impl::SolverInterfaceImpl]:2108 in closeCommunicationChannels: Synchronize participants and close communication channels
```
although the sink is disabled. What am I doing wrong?

fsimonis:
For the record: an empty filter should break the current implementation too.

davidscn:
To avoid misunderstandings: preCICE doesn't break, I just picked the last line in order to show the logging message. My point was that there still _is_ (default?) logging.

davidscn:
I still get a default sink when disabling all sinks

```bash
[2022-09-07 14:19:31.033693] [0x00007f77d52f1b80] [trace] %message%
```

is it intended? The second argument looks like a memory address, which seems quite useless.

fsimonis:
This looks like MPI output to me. What are you running to run into this message?

src/logging/LogConfiguration.cpp
```diff
@@ -173,17 +202,41 @@ void setupLogging(LoggingConfiguration configs, bool enabled)
       << bl::expressions::attr<std::string>("Function") << ": "
       << bl::expressions::message;
 
-  // Reset
-  bl::core::get()->remove_all_sinks();
-  bl::core::get()->reset_filter();
+  // Remove active preCICE sinks
+  using sink_t   = typename boost::log::sinks::sink;
```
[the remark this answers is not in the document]
fsimonis: Could be this [compiler bug](https://gcc.gnu.org/bugzilla/show_bug.cgi?id=66159).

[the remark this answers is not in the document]
davidscn: Yes, might be. Considering that it still triggers a (somewhat) weird warning, I would propose to merge this line with the line following.

davidscn:
Disabling the sinks (same as above)

```xml
  <log>
    <sink
      filter="%Severity% > debug and %Rank% = 0"
      format="---[precice] %ColorizedSeverity% %Message%"
      enabled="false" />
  </log>
```

still results in default logging by boost.

fsimonis:
@davidscn Sorry, I didn't see you reply until now.
Could you try again?

src/logging/LogConfiguration.cpp
```diff
@@ -173,17 +202,42 @@ void setupLogging(LoggingConfiguration configs, bool enabled)
...
+  using sink_ptr = typename boost::shared_ptr<sink_t>;
```
davidscn, line 207:
```suggestion
  using sink_ptr = typename boost::shared_ptr<boost::log::sinks::sink>;
```

davidscn (approved):
Yes exactly.

PR #1422, opened 2022-08-23
Simplify waveform class: Don't use data from old window, allow higher-order interpolation through BSplines 

### LINKED ISSUE #1421: Simplification of Waveform class

BenjaminRodenberg:
## Main changes of this PR

Closes #1421.

This implements an important feature needed for #1171. Multiple samples can now be stored in the waveform and higher-order interpolation is possible. Note that this feature is not yet available to the user and only implemented for the Waveform class.

## Author's checklist

* [X] I sticked to CMake version 3.16.3.
* [x] Already introduce (some) changes from https://github.com/precice/precice/pull/1414/commits/c79636b7b62334f8fde2af9128f449235f317a7a here?

## Reviewers' checklist


src/time/Waveform.hpp
```diff
@@ -114,6 +103,21 @@ class Waveform {
    * @return Order that may be used.
    */
   int computeUsedOrder(int requestedOrder, int numberOfAvailableSamples);
+
+  /**
+   * @brief Returns point the closest time stored in _timeStepsStorage that is after normalizedDt
+   *
+   * @param normalizedDt point in time
+   * @return double point in time after normalizedDt in _timeStepsStorage
```
uekerman, line 111:
```suggestion
   * @return double closest point in time
```
No need to duplicate information from the docstring above.

src/time/Waveform.hpp
```diff
@@ -114,6 +103,21 @@ class Waveform {
...
+   */
+  double findTimeAfter(double normalizedDt);
+
+  /**
+   * @brief Get keys of _timeStepsStorage in ascending order. Starting from low to high.
```
uekerman, line 116: normalized?

src/time/Waveform.cpp
```diff
@@ -20,110 +22,159 @@ int Waveform::getInterpolationOrder() const
...
-    this->storeAt(values, sampleIndex);
+  _timeStepsStorage[0.0] = Eigen::VectorXd(values);
+  _timeStepsStorage[1.0] = Eigen::VectorXd(values);
+}
+
+void Waveform::store(const Eigen::VectorXd &values, double normalizedDt)
+{
+  PRECICE_ASSERT(_timeStepsStorage.size() > 0);
+  // dt has to be in interval (0.0, 1.0]
+  PRECICE_ASSERT(normalizedDt > 0.0); // cannot override value at beginning of window. It is locked!
```
uekerman, line 38: Be aware of floating point values. Better use functions from `math::differences.hpp`

src/time/Waveform.cpp
```diff
@@ -20,110 +22,159 @@ int Waveform::getInterpolationOrder() const
...
-void Waveform::storeAt(const Eigen::VectorXd values, int sampleIndex)
+// helper function to compute x(t) from given data (x0,t0), (x1,t1), ..., (xn,tn) via B-spline interpolation (implemented using Eigen).
+Eigen::VectorXd bSplineInterpolationAt(double t, Eigen::VectorXd ts, Eigen::MatrixXd xs, int splineDegree)
 {
-  PRECICE_ASSERT(_timeWindowsStorage.cols() > sampleIndex, maxNumberOfStoredSamples(), sampleIndex);
-  PRECICE_ASSERT(values.size() == this->valuesSize(), values.size(), this->valuesSize());
-  this->_timeWindowsStorage.col(sampleIndex) = values;
+  // organize data in columns. Each column represents one sample in time.
+  PRECICE_ASSERT(xs.cols() == ts.size());
+  const int ndofs = xs.rows(); // number of dofs. Each dof needs it's own interpolant.
```
uekerman, line 69:
```suggestion
  const int ndofs = xs.rows(); // number of dofs. Each dof needs its own interpolant.
```

src/time/Waveform.cpp
```diff
@@ -20,110 +22,159 @@ int Waveform::getInterpolationOrder() const
...
+
+  Eigen::VectorXd interpolated(ndofs);
+
+  const int splineDimension = 1;
+
+  for (int i = 0; i < ndofs; i++) {
+    auto spline     = Eigen::SplineFitting<Eigen::Spline<double, splineDimension>>::Interpolate(xs.row(i), splineDegree, ts);
+    interpolated[i] = spline(t)[0]; // get component of spline associated with xs.row(i)
+  }
```
uekerman, line 78: I did not check, but isn't there a vectorized function for this in Eigen? (such that we don't need the loop)

BenjaminRodenberg: Not sure, but I already spent so much time on Eigen here that I don't want to spend more time on runtime optimizations. Error messages are really difficult to understand in Eigen. I would rather put this into an issue and see whether we really need it. (If somebody else wants to dig deeper here, I'm also fine.)

src/time/Waveform.hpp
```diff
@@ -62,46 +63,34 @@ class Waveform {
...
-   */
-  int maxNumberOfStoredSamples();
-
-  /**
-   * @brief Updates entry in _timeWindowsStorage corresponding to a given sampleIndex with given values.
-   * @param values Input sample.
-   * @param sampleIndex Index of sample to be updated to given input sample.
+   * @brief Get maximum dt that is stored in this waveform.
+   *
+   * Used to check whether a user is trying to add a sample associated with a dt that is smaller than the maximum dt. This is forbidden, because the waveform is locked for times that are smaller than the maximum dt.
```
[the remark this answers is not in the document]
BenjaminRodenberg: It's always normalized dt in this class (but yes: we can make this more explicit).

I wouldn't call it minimum. It's the maximum dt that is currently stored in the waveform. So the stored maximum is the minimum for the next step. Note: It's a private method, so we can also assume that people who use the method will understand this logic.

src/time/Waveform.hpp
BenjaminRodenberg: I added a bit more of documentation here: c818df6fbebebc1daab52707e85a89abe74dbda6

src/time/Waveform.cpp
BenjaminRodenberg: I think I remember the actual reason now: `Eigen::SplineFitting<Eigen::Spline<double, splineDimension>>` is a template. This means: `splineDimension` has to be known at compile time.

src/time/Waveform.hpp
```diff
@@ -62,47 +64,22 @@ class Waveform {
...
   Eigen::VectorXd sample(const double normalizedDt);
 
 private:
-  /// Set by initialize. Used for consistency checks.
-  bool _storageIsInitialized = false;
-
-  /// Stores values for several time windows.
-  Eigen::MatrixXd _timeWindowsStorage;
+  /// Stores values on the current window.
+  Storage _timeStepsStorage;
```
uekerman, line 76:
```suggestion
  Storage _storage;
```
?

src/time/tests/PiecewiseWaveformTest.cpp
```diff
@@ -0,0 +1,306 @@
...
+#include "testing/Testing.hpp"
+#include "testing/WaveformFixture.hpp"
+#include "time/Time.hpp"
+#include "time/Waveform.hpp"
+
+using namespace precice;
+using namespace precice::time;
+
+BOOST_AUTO_TEST_SUITE(TimeTests)
+// @TODO rename to SubcyclingWaveformTests?
```
uekerman, line 12:
```suggestion
```

uekerman (approved):
Looks good to me. I did not read through all tests.

Minor:

* It's a bit odd that the tests of `PiecewiseWaveformTest` are not part of `WaveformTest`. Usually, there should be a one-to-one match between class and test class.
* We need unit tests for the `Storage` class.

src/time/Storage.hpp
```diff
@@ -0,0 +1,98 @@
...
+
+  /**
+   * @brief Returns the value at time following "before" contained in this Storage.
+   *
+   * The stored normalized dt is larger or equal than "before". If "before" is a normalized dt stored in this Storage, this function returns the value at "before"
+   *
+   * @param before a double, where we want to find a normalized dt that comes directly after this one
+   * @return Eigen::VectorXd a value in this Storage at or directly after "before"
+   */
+  Eigen::VectorXd getValueAtTimeAfter(double before);
```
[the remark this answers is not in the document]
BenjaminRodenberg: Sounds good :+1: Renamed in a2c857781ad6df98a1cd3f716f8727352b1d6a81.

src/time/Storage.cpp
```diff
@@ -0,0 +1,101 @@
...
+{
+}
+
+void Storage::initialize(Eigen::VectorXd values)
+{
+  _sampleStorage.emplace_back(std::make_pair(0.0, values));
+  _sampleStorage.emplace_back(std::make_pair(1.0, values));
+}
+
+int findTime(std::vector<std::pair<double, Eigen::VectorXd>> storage, double time)
```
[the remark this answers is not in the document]
BenjaminRodenberg: This function is actually not needed. Maybe we will need it later, but currently its only unnecessary bloat. Removed it in c96a7002c7ed4e5c817cc08b0c2548cc2c4e9ab3

src/time/Waveform.hpp
BenjaminRodenberg: Sounds good :+1: 59dbd2628ee1dbd9b639303f46b84c2748611695

src/time/Waveform.cpp
```diff
@@ -20,108 +21,84 @@ int Waveform::getInterpolationOrder() const
...
+  const int usedOrder = computeUsedOrder(_interpolationOrder, _storage.nTimes());
 
-  const int usedOrder = computeUsedOrder(_interpolationOrder, _numberOfStoredSamples);
+  PRECICE_ASSERT(math::equals(this->_storage.maxStoredNormalizedDt(), 1.0), this->_storage.maxStoredNormalizedDt()); // sampling is only allowed, if a window is complete.
 
-  if (usedOrder == 0) {
+  // @TODO: Improve efficiency: Check whether key = normalizedDt is in _storage. If yes, just get value and return. No need for interpolation.
+
+  if (_interpolationOrder == 0) {
+    // @TODO: Remove constant interpolation in preCICE v3.0? Usecase is unclear and does not generalize well with BSpline interpolation. It's also not 100% clear what to do at the jump.
```
BenjaminRodenberg, line 70: @uekerman what do you think about this point? Should we just remove constant interpolation in 3.0.0 and make linear interpolation the new default? It should still behave the same way as before and I don't see a real benefit providing constant interpolation (maybe for academic reasons but that's it).

src/time/tests/PiecewiseWaveformTest.cpp
BenjaminRodenberg: I merged this file with `WaveformTest.cpp`

src/time/Waveform.cpp
uekerman: It's also a bit undocumented right now which "constant interpolation" we mean, constant over the complete time window or constant per timestep. I guess the later would be more consistent with what we do for order 1 and higher.
I think that the academic use case could be interesting in the long run. To quickly test things (whether the overall order drops or not).
So if it is not overly complicated to keep that case, I would vote for the "constant per timestep".

BenjaminRodenberg: It's no big problem to keep the constant interpolation. So let's keep it. I think the academic argument makes sense. We can still remove it at some point in time when the academic case is not relevant anymore - currently it definitely is for my thesis ;)

On the question interpolation on time step vs. interpolation on window: I would not solve this problem here. The waveform class currently just uses all the data it has. It will use the "per-window-interpolation", if there is only data at the beginning and at the end of the window. If there is more data: It will use all the data that is available and do piecewise constant, linear ... interpolation ("per-time-step-interpolation"). I will add a few more lines of documentation here.

Note: In the end we can still use the "per-window-interpolation", if we just provide data at the beginning and at the end of the window and ignore all the data from subcycling (basically what's happening in preCICE v2). But this is mainly a matter of configuration. Question here: Do we even want to keep this option? If I want to ignore data, why do I even write it in the first place? Again: The only reason I see for still offering the v2 subcycling is academic. But let's discuss this in the "big" PR, where we will also deal with configuration.

BenjaminRodenberg: I added some more documentation in [53fda27](https://github.com/precice/precice/pull/1422/commits/53fda271c4f6a78808dad6874098a446bba28038). I hope it is clear now.

PR #1430, opened 2022-09-08
Add version info of dependencies

fsimonis:
## Main changes of this PR

This PR adds the versions of dependencies to the CMake output.

This should make it easier for preCICE users and developers to detect conflicting versions.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -166,6 +166,7 @@ endif()
 find_package(Boost 1.71.0 REQUIRED
   COMPONENTS filesystem log log_setup program_options system thread unit_test_framework
   )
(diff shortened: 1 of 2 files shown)


davidscn (commented):
Nice, thanks. Only flaw is that the version string is empty when using the explicit eigen dir:
```cmake
cmake -D EIGEN3_INCLUDE_DIR="/home/~/eigen/" ..
```
```
...
-- Found Eigen 
-- Validating Eigen
```

fsimonis:
@davidscn Thanks! Fixed

davidscn (approved):

PR #1435, opened 2022-09-17
Simplify implementation of DataContext.

BenjaminRodenberg:
## Main changes of this PR

* Move DataPtr inside MappingContext.
* Use range-based for loops in DataContext.
* Simplify access of data properties vie context (does not require detour over std::vector in DataContext anymore)

Helpful in scope of #1414 

## Author's checklist

* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release. (not needed?)
* [ ] I added a test to cover the proposed changes in our test suite. (not needed?)

## Reviewers' checklist

### DIFF
--- a/src/precice/impl/DataContext.cpp
+++ b/src/precice/impl/DataContext.cpp
@@ -20,30 +20,15 @@ std::string DataContext::getDataName() const
   return _providedData->getName();
 }
 
-DataID DataContext::getFromDataID(size_t dataVectorIndex) const
-{
-  PRECICE_ASSERT(hasMapping());
-  PRECICE_ASSERT(dataVectorIndex < _fromData.size())
-  PRECICE_ASSERT(_fromData[dataVectorIndex]);
-  return _fromData[dataVectorIndex]->getID();
-}
-
 void DataContext::resetData()
 {
   // See also https://github.com/precice/precice/issues/1156.
   _providedData->toZero();
   if (hasMapping()) {
     PRECICE_ASSERT(hasWriteMapping());
-    std::for_each(_toData.begin(), _toData.end(), [](auto &data) { data->toZero(); });
+    // PRECICE_ASSERT(!hasReadMapping());  // we also need this assertion, because currently we also reset toData from read mappings, if a data context has a read and write mapping! Is this what we want?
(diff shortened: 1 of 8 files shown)


src/precice/impl/DataContext.cpp
```diff
@@ -20,32 +20,17 @@ std::string DataContext::getDataName() const
...
-
 void DataContext::resetData()
 {
   // See also https://github.com/precice/precice/issues/1156.
   _providedData->toZero();
   if (hasMapping()) {
     PRECICE_ASSERT(hasWriteMapping());
-    std::for_each(_toData.begin(), _toData.end(), [](auto &data) { data->toZero(); });
+    // PRECICE_ASSERT(!hasReadMapping());  // we also need this assertion, because currently we also reset toData from read mappings, if a data context has a read and write mapping! Is this what we want?
+    std::for_each(_mappingContexts.begin(), _mappingContexts.end(), [](auto &context) { context.toData->toZero(); });
```
uekerman, line 30: I think I don't understand the comment here. Can a data context have a read and a write mapping? I would guess no.
And where do we reset toData?

Sth different: maybe it's cleaner if we move the resetting to a `MappingContext::resetData`?

uekerman (approved):
Makes things easier, yes.
Only one doubt below, otherwise good to go.

src/precice/impl/DataContext.cpp
BenjaminRodenberg: > I think I don't understand the comment here. Can a data context have a read and a write mapping? I would guess no. And where do we reset toData?

I tried out what happens, if I activate the assertion and it does not get triggered. I thought this was the reason for originally commenting it out, but everything looks fine now: A `DataContext` either has a read or write mapping, not both at the same time (at least, if we reset data).

I added the assertion in [5a0ff28](https://github.com/precice/precice/pull/1435/commits/5a0ff28e61abbfe123a7aa64d9bc7eb225fa0646) and I think we can merge it this way (if the tests run though).

> Sth different: maybe it's cleaner if we move the resetting to a `MappingContext::resetData`?

Yes. But `MappingContext` is a `struct`, so moving something inside `MappingContext` will cause a bigger refactoring and this is from my perspective out of scope for this PR. Opening an isssue would be useful and I guess it would also be a low-hanging fruit, if somebody wants to get his or her hands dirty with C++ and preCICE.

PR #1439, opened 2022-09-20
Axial geometric multiscale mapping

ezonta:
## Main changes of this PR
Addition of Axial Geometric Multiscale Coupling Feature.

Ported previously implemented geometric multiscale mapping to new preCICE version in order to expand functionality of preCICE and enable coupling of dimensionally heterogeneous models.

## Author's checklist

## Reviewers' checklist


ezonta:
@MakisH I will also add you as a reviewer here ASAP.

MakisH:
I converted this to draft, since I was not sure if we would like to merge everything as it comes (option A), or make a feature branch, merge small PRs to that, and in the end merge the whole feature, while regularly rebasing (option B). The only motivation for option B is to keep track of the general progress and have an overview of the required code changes.

@uekerman what do you think? If I remember correctly, you were favoring option A in a recent discussion.

src/mapping/AxialGeoMultiscaleMapping.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+#include "mapping/Mapping.hpp"
+#include "logging/Logger.hpp"
+#include <vector>
+
+namespace precice {
+namespace mapping {
+
+/// Geometric multiscale mapping in axial direction
+class AxialGeoMultiscaleMapping : public Mapping
```
MakisH, line 11: I wonder if eventually we should have a `GeoMultiscaleMapping` parent class. Let's just keep this in mind.

src/mapping/AxialGeoMultiscaleMapping.cpp
```diff
@@ -0,0 +1,143 @@
+#include "AxialGeoMultiscaleMapping.hpp"
+#include "mesh/Mesh.hpp"
+
+namespace precice {
+
+namespace mapping {
+
+AxialGeoMultiscaleMapping:: AxialGeoMultiscaleMapping
```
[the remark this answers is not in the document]
ezonta: Installed and used clang-format-8 and the pre-commit hook now.

src/mapping/AxialGeoMultiscaleMapping.cpp
```diff
@@ -0,0 +1,143 @@
...
+{
+  setInputRequirement(Mapping::MeshRequirement::VERTEX);
+  setOutputRequirement(Mapping::MeshRequirement::VERTEX);
+
+  PRECICE_ASSERT(dimensions == 2 || dimensions == 3);
+  if(dimensions == 2){
+    _scaling = 1.5;
+  } else if(dimensions == 3){
+    _scaling = 2.0;
+  }
```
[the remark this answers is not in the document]
ezonta: The `_scaling` being different for 2D and 3D cases is something that I'm not sure is correct. In my opinion it should be `2.0` no matter the dimensionality. We should keep track of it and investigate it further. However, for now we're only dealing with 3D cases so it can stay.

src/mapping/AxialGeoMultiscaleMapping.cpp
```diff
@@ -0,0 +1,143 @@
...
+{
+  PRECICE_TRACE(inputDataID, outputDataID);
+
+  const Eigen::VectorXd& inputValues = input()->data(inputDataID)->values();
+  int valueDimensions = input()->data(inputDataID)->getDimensions();
+  Eigen::VectorXd& outputValues = output()->data(outputDataID)->values();
+
+  PRECICE_ASSERT ( valueDimensions == output()->data(outputDataID)->getDimensions(),
+              valueDimensions, output()->data(outputDataID)->getDimensions() );
+  PRECICE_ASSERT ( inputValues.size() / valueDimensions == (int)input()->vertices().size(),
```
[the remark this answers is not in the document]
ezonta: Done

src/mapping/AxialGeoMultiscaleMapping.hpp
ezonta: Would be a good idea if we later on also implement `RadialGeoMultiscaleMapping` as a child.

src/mapping/AxialGeoMultiscaleMapping.cpp
```diff
@@ -0,0 +1,143 @@
...
+namespace precice {
+
+namespace mapping {
+
+AxialGeoMultiscaleMapping:: AxialGeoMultiscaleMapping
+(
+  Constraint constraint,
+  int        dimensions,
+  MultiscaleType type,
+  double radius)
```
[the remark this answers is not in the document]
ezonta: I'm currently working on some tests for the 3D-1D and the 1D-3D case.

src/mapping/AxialGeoMultiscaleMapping.cpp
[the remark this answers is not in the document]
MakisH: Let's document the "maybe this is not correct" part. Maybe `_scaling` is just a confusing name in the end.

uekerman:
> @uekerman what do you think? If I remember correctly, you were favoring option A in a recent discussion.

Let's try to do small isolated PRs that we can quickly merge into `develop`. No need to keep this back till we have all possible geometric configurations. If necessary, you can always make the new mappings [experimental](https://precice.org/configuration-xml-reference.html#precice-configuration).

Example:

https://github.com/precice/precice/blob/66c11040713fa186723723feb3d3c81968b4787c/src/cplscheme/config/CouplingSchemeConfiguration.hpp#L128

src/mapping/tests/AxialGeoMultiscaleMappingTest.cpp
```diff
@@ -0,0 +1,130 @@
...
+  // Create mesh to map from
+  PtrMesh inMesh(new Mesh("InMesh", dimensions, testing::nextMeshID()));
+  PtrData inData    = inMesh->createData("InData", 3, 0_dataID);
+  int     inDataID  = inData->getID();
+  Vertex &inVertex0 = inMesh->createVertex(Eigen::Vector3d::Constant(0.0));
+  inMesh->allocateDataValues();
+  Eigen::VectorXd &inValues = inData->values();
+  inValues << 0.0, 0.0, 2.0;
+
+  double radius = 1.0;
```
MakisH, line 37: We probably need a comment for the intention/role of the `radius`. I can imagine that later on we will have more options, or that this test ends up being debugged by someone who does not know much about this mapping method.

src/mapping/tests/AxialGeoMultiscaleMappingTest.cpp
```diff
@@ -0,0 +1,130 @@
...
+  precice::mapping::AxialGeoMultiscaleMapping mapping(mapping::Mapping::CONSISTENT, dimensions, mapping::AxialGeoMultiscaleMapping::SPREAD, radius);
+  mapping.setMeshes(inMesh, outMesh);
+  BOOST_TEST(mapping.hasComputedMapping() == false);
+
+  // Map data
+  mapping.computeMapping();
+  mapping.map(inDataID, outDataID);
+  const Eigen::VectorXd &outValues = outData->values();
+
+  // Check if data is doubled at center node
```
MakisH, line 58: Before the checking, it would be nice to have a comment earlier on regarding the general approach of this test: "Create two meshes so that (blah), then map from `inMesh` to `outMesh` and check if (blah)".

The same for the next test case.

src/mapping/tests/AxialGeoMultiscaleMappingTest.cpp
```diff
@@ -0,0 +1,130 @@
...
+  BOOST_TEST(mapping.hasComputedMapping() == true);
+  BOOST_TEST(outValuesScalar(0) == inValuesScalar(0));
+  BOOST_TEST(outValuesVector(0) == 0.5 * inValuesVector(0));
+  BOOST_TEST(outValuesVector(1) == 0.5 * inValuesVector(1));
+  BOOST_TEST(outValuesVector(2) == 0.5 * inValuesVector(2));
+
+  // Check if invalue vector data at distance = r/2 is 2/3 times outvalue vector data
+  BOOST_TEST(outValuesVector(6) == (2 / 3) * inValuesVector(0));
+  BOOST_TEST(outValuesVector(7) == (2 / 3) * inValuesVector(1));
+  BOOST_TEST(outValuesVector(8) == (2 / 3) * inValuesVector(2));
```
MakisH, line 126: Is this `(2 / 3)` equal to 2/3 or to zero? I would rather write:

```suggestion
  BOOST_TEST(outValuesVector(6) == (2 / 3.0) * inValuesVector(0));
  BOOST_TEST(outValuesVector(7) == (2 / 3.0) * inValuesVector(1));
  BOOST_TEST(outValuesVector(8) == (2 / 3.0) * inValuesVector(2));
```

Also, since we are comparing floating point numbers, shouldn't we compare against some tolerance? Or is that somehow handled by the `BOOST_TEST`? I would not expect that.

_clang-tidy/CMakeCache.txt
```diff
@@ -0,0 +1,863 @@
+# This is the CMakeCache file.
```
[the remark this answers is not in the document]
ezonta: I'm not sure how it ended up here, maybe by calling clang-tidy? I'll just remove it manually

src/mapping/tests/AxialGeoMultiscaleMappingTest.cpp
ezonta: Okay, I will try to document everything more thoroughly!

src/mapping/tests/AxialGeoMultiscaleMappingTest.cpp
ezonta: Will do!

src/mapping/tests/AxialGeoMultiscaleMappingTest.cpp
ezonta: I'll correct that. However, I believe BOOST does indeed compare against a tolerance, because when a test fails it says something like "difference bigger than tolerance". I think the tolerance is aroung 1e-9.

src/mapping/tests/AxialGeoMultiscaleMappingTest.cpp
```diff
@@ -0,0 +1,144 @@
...
+
+  PRECICE_TEST(1_rank);
+  int dimensions = 3;
+  using testing::equals;
+
+  // Create mesh to map from
+  PtrMesh inMesh(new Mesh("InMesh", dimensions, testing::nextMeshID()));
+  PtrData inData    = inMesh->createData("InData", 3, 0_dataID);
+  int     inDataID  = inData->getID();
+  Vertex &inVertex0 = inMesh->createVertex(Eigen::Vector3d::Constant(0.0)); // center
```
MakisH, line 95: Looks like `inVertex0` is unused; the same for more variables (check the compiler warnings on the diff).

MakisH (commented):
Looks good to me already, but please have another look at the automated annotations in the "files changed" view and in the CI:
- formatting
- clang-tidy
- testing coverage

uekerman:
Another potential application of this feature is motivated by @ahmed-h-elsheikh s case.
There we want a parabolic inflow profile on some OpenFOAM geometry by only communicating one value (the max or mean value of the parabola). Alternative to https://github.com/precice/openfoam-adapter/issues/278

ezonta:
> Another potential application of this feature is motivated by @ahmed-h-elsheikh s case.
> There we want a parabolic inflow profile on some OpenFOAM geometry by only communicating one value (the max or mean value of the parabola). Alternative to [precice/openfoam-adapter#278](https://github.com/precice/openfoam-adapter/issues/278)

That sounds like a perfect application of this feature indeed. When providing the mean value, this would already work.

MakisH (approved):
The code looks clean and clear now, all tests pass, and only changes to the axial mapping are implemented here. :rocket:

PR #1446, opened 2022-10-06
Handle interrupts in communication

### LINKED ISSUE #1443: preCICE suspectedly fails to recover from EINTR
**Describe your setup**

Operating system (e.g. Linux distribution and version): _Debian bullseye_
preCICE Version: 2.5.0

**Describe the problem**
When a preCICE application receives a signal while it is waiting on a socket, both sides of the application die with the message `Receiving data from another participant (using sockets) failed with a system error: read: Interrupted system call. This often means that the other participant exited with an error (look there)`. This is presumably caused by preCICE handling EINTR as an unrecoverable error.

**Step To Reproduce**
1. Install the latest development version of Nutils (which uses bottombar 2)
2. Run the partitioned-heat-condition tutorial

**Expected behaviour**
The tutorial code is expected to run without problems.

**Additional context**
The reason that bottombar 2 is causing the code to fail is that it uses setitimer to update itself, which triggers SIGALRM at a set frequency. The previous version used signals only for window resizes via SIGWINCH, which is not triggered as often and therefore caused fewer issues (though we have long received reports of tmux reattach mysteriously killing applications). We are in the process of making point releases for bottombar that set the siginterrupt flag to false, which fixes the issue. Nonetheless, we thought it good to point out that this issue exists and that it could potentially be fixed on your side as well.

fsimonis:
## Main changes of this PR

This PR:
- Extends robustness of the socket communication, correctly handling interrupts
- Extends messages for all communication types, especially handling the remote participant crashing.
- Correctly handles failing async requests
- Adds error messages for async requests

Closes #1443

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/1446.md
+++ b/docs/changelog/1446.md
@@ -1,0 +1,3 @@
+- Fixed interrupts during blocked communication leading to errors instead of restarts.
+- Improved error messages regarding communication issues.
+- Added error checking for asynchronous communication.
--- a/src/acceleration/impl/ParallelMatrixOperations.hpp
+++ b/src/acceleration/impl/ParallelMatrixOperations.hpp
@@ -9,6 +9,7 @@
 #include <vector>
 
 #include "com/Communication.hpp"
+#include "com/ErrorHandling.hpp"
 #include "com/MPIPortsCommunication.hpp"
 #include "com/Request.hpp"
 #include "com/SharedPointer.hpp"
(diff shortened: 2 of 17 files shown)


gertjanvanzwieten:
Today I finally managed to try out the branch and unfortunately the issue does not seem to be fixed, even though the error message changed from `Receiving data from another participant (using sockets) failed with a system error: read: Interrupted system call. This often mean
s that the other participant exited with an error (look there)` (both sides) to `Receiving data from another participant failed with a system error: Interrupted system call.` on one side and `Receiving data from another participant failed with a system error: End of file.` on the other. Tested with the partitioned-heat-conduction tutorial coupling nutils to nutils (development version) with bottombar 2.0.2 (which does not yet contain our siginterrupt fix). I can send you the Dockerfile if it's helpful.

fsimonis:
@gertjanvanzwieten Thank you very much for the report!

In summary:
* There are still places where EINTR isn't handled correctly.
* The error messages are wrong. We need to handle Inter vs Intracomm as well as sending vs receiving.

fsimonis:
Turns out that this is more difficult than I thought as this requires to shift abstraction layers regarding communication.

Unsolvable without major refactoring is a way of restarting asynchronous calls.

fsimonis:
Closing for mentioned reasons

PR #1448, opened 2022-10-17
Handle solverdummy arguments properly

### LINKED ISSUE #1447: Improve solverdummy argument parsing
Our solverdummies are part of the test infrastructure. In case something goes wrong with the argument parsing, users (might) face issues like this one

https://stackoverflow.com/questions/62570882/what-basic-string-m-construct-null-not-valid

which is very cryptic and due to inappropriate handling of the parsed arguments (see the discussion in the link and our current implementation).

https://github.com/precice/precice/blob/e6da8d5d9432f0872ff0e1ecf652477fbbcf6c68/examples/solverdummies/cpp/solverdummy.cpp#L13-L17

davidscn:
## Main changes of this PR

See details in #1447. There was already a check for the cpp solverdummy, but it was in the wrong place. I don't think a changelog entry is necessary.

Closes #1447 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/examples/solverdummies/c/solverdummy.c
+++ b/examples/solverdummies/c/solverdummy.c
@@ -23,16 +23,16 @@ int main(int argc, char **argv)
   const char *writeDataName;
   const char *readDataName;
 
-  const char *configFileName  = argv[1];
-  const char *participantName = argv[2];
-
   if (argc != 3) {
-    printf("Usage: ./solverdummy configFile solverName\n\n");
+    printf("The solverdummy was called without arguments. Usage: ./solverdummy configFile solverName\n\n");
     printf("Parameter description\n");
     printf("  configurationFile: Path and filename of preCICE configuration\n");
     printf("  solverName:        SolverDummy participant name in preCICE configuration\n");
     return 1;
   }
+
+  const char *configFileName  = argv[1];
+  const char *participantName = argv[2];
 
   printf("DUMMY: Running solver dummy with preCICE config file \"%s\" and participant name \"%s\".\n",
          configFileName, participantName);
--- a/examples/solverdummies/c/solverdummy.c
+++ b/examples/solverdummies/c/solverdummy.c
@@ -24,7 +24,7 @@ int main(int argc, char **argv)
   const char *readDataName;
 
   if (argc != 3) {
-    printf("The solverdummy was called without arguments. Usage: ./solverdummy configFile solverName\n\n");
+    printf("The solverdummy was called with an incorrect number of arguments. Usage: ./solverdummy configFile solverName\n\n");
     printf("Parameter description\n");
     printf("  configurationFile: Path and filename of preCICE configuration\n");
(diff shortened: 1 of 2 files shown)


examples/solverdummies/c/solverdummy.c
```diff
@@ -23,17 +23,17 @@ int main(int argc, char **argv)
   const char *writeDataName;
   const char *readDataName;
 
-  const char *configFileName  = argv[1];
-  const char *participantName = argv[2];
-
   if (argc != 3) {
-    printf("Usage: ./solverdummy configFile solverName\n\n");
+    printf("The solverdummy was called without arguments. Usage: ./solverdummy configFile solverName\n\n");
```
IshaanDesai, line 27:
```suggestion
    printf("The solverdummy was called with incorrect arguments. Usage: ./solverdummy configFile solverName\n\n");
```

examples/solverdummies/cpp/solverdummy.cpp
```diff
@@ -10,20 +10,20 @@ int main(int argc, char **argv)
...
 
-  std::string configFileName(argv[1]);
-  std::string solverName(argv[2]);
-  std::string meshName;
-  std::string dataWriteName;
-  std::string dataReadName;
-
   if (argc != 3) {
-    std::cout << "Usage: ./solverdummy configFile solverName\n\n";
+    std::cout << "The solverdummy was called without arguments. Usage: ./solverdummy configFile solverName\n\n";
```
IshaanDesai, line 14:
```suggestion
    std::cout << "The solverdummy was called with incorrect arguments. Usage: ./solverdummy configFile solverName\n\n";
```

IshaanDesai (approved):
Looks good :+1:

examples/solverdummies/c/solverdummy.c
davidscn: It's actually the number of arguments which is checked here, so maybe I can even get more specific 
```suggestion
    printf("The solverdummy was called with an incorrect number of arguments. Usage: ./solverdummy configFile solverName\n\n");
```

davidscn:
Parsing arguments is handled differently for fortran 
https://github.com/precice/precice/blob/3007d5f59af0555c931d2e424a1459f020861b96/examples/solverdummies/fortran/solverdummy.f90#L23-L24

I don't know if we can assert things there at all.

uekerman:
Seems fortran does a decent job there already:

```
$ ./solverdummy
[2022-10-31 14:23:35.501966] [0x00007f26d3e65900] [error]   This participant's name is an empty string. When constructing a preCICE interface you need to pass the name of the participant as first argument to the constructor.
```

meaning `config` becomes an empty string.

Too many arguments does not lead to an error
```
$ ./solverdummy ../precice-config.xml SolverOne bla
preCICE: This is preCICE version 2.5.0
preCICE: Revision info: v2.5.0-22-ge6da8d5d9
preCICE: Build type: Debug
...
```

which I think is tolerable.

PR #1452, opened 2022-10-28
Add CMake presets

fsimonis:
## Main changes of this PR

This PR adds CMake presets, which are available since CMake 3.19.
This is not supported by the Ubuntu 20.04 LTS, so I picked the CMake version 3.22 of Ubuntu 22.04 LTS as a baseline.

This PR adds the presets:
```
cmake --list-presets
Available configure presets:

  "production"   - Default production release
  "develop"      - Default development config
  "debpkg"       - Debian Package
  "bigrelease"   - Big release
  "localinstall" - Release with local installation
  "profiling"    - Profiling
```

One can easily switch to a new preset by running:
```
cmake --preset profiling
```

The default build directory is `./build`, the installation directory for `localinstall` is `./installation`.

This simplifies developers to use the appropriate flags and allows to quickly switch to a common setup without having to manually change variables.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -13,7 +13,7 @@
       "name": "production",
       "inherits": "base",
       "displayName": "Production release",
-      "description": "Default release configuration as preCICE is intended to be used in production.",
+      "description": "Release configuration for preCICE used in production.",
       "cacheVariables": {
         "CMAKE_BUILD_TYPE": "Release",
         "CMAKE_INTERPROCEDURAL_OPTIMIZATION": "ON"
@@ -58,7 +58,7 @@
     {
       "name": "profiling",
       "inherits": "develop",
-      "displayName": "Profiling",
+      "displayName": "Profiling config",
       "description": "Profiling configuration",
       "cacheVariables": {
         "CMAKE_BUILD_TYPE": "RelWithDebInfo",
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -1,0 +1,77 @@
+{
+  "version": 3,
+  "configurePresets": [
+    {
+      "name": "base",
+      "hidden": true,
+      "binaryDir": "${sourceDir}/build",
+      "cacheVariables": {
+        "BUILD_SHARED_LIBS": "ON"
+      }
+    },
+    {
+      "name": "production",
+      "inherits": "base",
+      "displayName": "Production release",
+      "description": "Default release configuration as preCICE is intended to be used in production.",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Release",
+        "CMAKE_INTERPROCEDURAL_OPTIMIZATION": "ON"
+      }
+    },
+    {
+      "name": "develop",
+      "inherits": "base",
+      "displayName": "Development config",
+      "description": "Development config with default settings",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Debug",
+        "CMAKE_CXX_FLAGS": "-Wall -Wextra -Wno-unused-parameter"
+      },
+      "warnings": {
+        "dev": true,
+        "deprecated": true,
+        "uninitialized": true
+      }
+    },
+    {
+      "name": "debpkg",
+      "inherits": "production",
+      "displayName": "Debian package",
+      "description": "Build for the generation of Debian packages",
+      "cacheVariables": {
+        "CMAKE_INSTALL_PREFIX": "/usr",
+        "CPACK_GENERATOR": "DEB",
+        "BUILD_TESTING": "OFF"
+      }
+    },
+    {
+      "name": "bigrelease",
+      "inherits": "production",
+      "displayName": "Big release",
+      "description": "Release build with log and assertions enabled",
+      "cacheVariables": {
+        "PRECICE_RELEASE_WITH_ASSERTIONS": "ON",
+        "PRECICE_RELEASE_WITH_DEBUG_LOG": "ON"
+      }
+    },
+    {
+      "name": "localinstall",
+      "inherits": "production",
+      "displayName": "Locally installed release",
+      "description": "Release build with local installation to ./installation",
+      "installDir": "${sourceDir}/installation"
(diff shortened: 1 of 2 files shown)


uekerman:
> This is not supported by the Ubuntu 20.04 LTS, so I picked the CMake version 3.22 of Ubuntu 22.04 LTS as a baseline.

This means we still have to wait till 24.04 LTS is out, right? Or is this backwards compatible?

fsimonis:
These presets are a newish feature of cmake. It doesn't affect our existing CMake files.

CMakePresets.json
```diff
@@ -0,0 +1,77 @@
...
+        "CMAKE_INSTALL_PREFIX": "/usr",
+        "CPACK_GENERATOR": "DEB",
+        "BUILD_TESTING": "OFF"
+      }
+    },
+    {
+      "name": "bigrelease",
+      "inherits": "production",
+      "displayName": "Big release",
+      "description": "Release build with log and assertions enabled",
```
MakisH, line 52: This term is not self-descriptive to me.
Why not `release-with-assertions`?

Since the presets are only used once in the installation, I think that long, descriptive names are fine.

CMakePresets.json
```diff
@@ -0,0 +1,77 @@
...
+    },
+    {
+      "name": "profiling",
+      "inherits": "develop",
+      "displayName": "Profiling",
+      "description": "Profiling configuration",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "RelWithDebInfo",
+        "CMAKE_INTERPROCEDURAL_OPTIMIZATION": "ON",
+        "CMAKE_CXX_FLAGS": "-Wall -Wextra -Wno-unused-parameter -fno-omit-frame-pointer"
```
MakisH, line 73: Does it make sense to include `-fno-omit-frame-pointer` in the `develop` (debug) preset as well? Is there any overhead? This is also useful for some debugging features, no?

MakisH (commented):
That's a very nice idea! I understand that it does not change the behavior in older CMake versions, so we can just merge it already now.

I tried building with `develop` and `debpkg` on Ubuntu 22.04 (CMake 3.22.1), it worked as expected.

Note that there is currently no changelog or documentation.

CMakePresets.json
```diff
@@ -0,0 +1,77 @@
+{
+  "version": 3,
+  "configurePresets": [
+    {
+      "name": "base",
```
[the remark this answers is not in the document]
fsimonis: CMake requires to explicitly name a preset. So there is no support for a default as such.

The base preset is valid, but incomplete. There are things missing for a usable setup, hence it marked it as hidden.

CMakePresets.json
```diff
@@ -0,0 +1,77 @@
...
+        "CMAKE_CXX_FLAGS": "-Wall -Wextra -Wno-unused-parameter"
+      },
+      "warnings": {
+        "dev": true,
+        "deprecated": true,
+        "uninitialized": true
+      }
+    },
+    {
+      "name": "debpkg",
```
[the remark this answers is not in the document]
fsimonis: Not really. I'll extend it.

CMakePresets.json
```diff
@@ -0,0 +1,77 @@
...
+      }
+    },
+    {
+      "name": "production",
+      "inherits": "base",
+      "displayName": "Production release",
+      "description": "Default release configuration as preCICE is intended to be used in production.",
+      "cacheVariables": {
+        "CMAKE_BUILD_TYPE": "Release",
+        "CMAKE_INTERPROCEDURAL_OPTIMIZATION": "ON"
```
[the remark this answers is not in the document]
fsimonis: The website is probably a better place for this, right?

CMakePresets.json
fsimonis: So, this is a release + debug log + assertions.
Not sure how to call this.

davidscn: For me, this would be equivalent to `CMake`s `RelWithDebInfo`. I would go with something similar, maybe `RelWithDebChecks` or so.

CMakePresets.json
```diff
@@ -0,0 +1,77 @@
...
+        "PRECICE_RELEASE_WITH_ASSERTIONS": "ON",
+        "PRECICE_RELEASE_WITH_DEBUG_LOG": "ON"
+      }
+    },
+    {
+      "name": "localinstall",
+      "inherits": "production",
+      "displayName": "Locally installed release",
+      "description": "Release build with local installation to ./installation",
+      "installDir": "${sourceDir}/installation"
```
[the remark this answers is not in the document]
fsimonis: Production doesn't enforce an installation directory. The user still needs to set it or go with the default.

The idea here was to create an installation in the source directory, which could be useful in scripts.
Thinking about this some more:
For scripting, it is often easier to have stand-alone directories for the source, the build, and the installation. (all 3 in one directory).

There isn't really a benefit to having this one. I'll remove it.

CMakePresets.json
fsimonis: Based on suggestion:
* verbose-release-with-assertions
* checked-release-with-debug-log
* checked-verbose-release
* debug-in-production
* production-debug

I think that we should add a preset for production + assertions (`checked-production`, `production-audit`, `production-with-assertion`)

CMakePresets.json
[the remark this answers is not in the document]
fsimonis: Added it to https://precice.org/installation-source-configuration.html

CMakePresets.json
fsimonis: Or optimized debug

fsimonis:
I'll merge this, so we can try the presets before the release.

uekerman:
Now that we have upgraded the baseline, does it make sense to copy and adjust the PR description to https://precice.org/installation-source-configuration.html ?

fsimonis:
Opened https://github.com/precice/precice.github.io/pull/435

PR #1455, opened 2022-10-31
Remove incorrect warning for waveform order zero and serial coupling schemes

BenjaminRodenberg:
## Main changes of this PR

Removes warning that tells user to not initialize data of second participant in serial coupling, if waveform order is zero. Initialization actually becomes reasonable in this case, if the user wants to sample at the starting time (we cannot know whether the user will sample or not, because the sampling happens over the API).

This PR also removes the warning to avoid the test `Integration/Serial/Time/Implicit/SerialCoupling/ReadWriteScalarDataWithWaveformSamplingZero` to trigger the warning. The test is correct and should not trigger any warning that suggests a change (because this will lead to a failure of the test).

This change is motivated by #1422. See this change in the test: https://github.com/precice/precice/commit/270cc0844bb3cdb1b0bea9997825c421d092bbfc#diff-52e93df8fe0311c9dcda9c30ccdb04ac3fd7d92fc5f7576ae6b261a40e23fb83L81-R81: At the beginning of the window we now get the data from the last window or initial data, even for zeroth order waveforms. This makes data initialization relevant for all coupling schemes and participants, if a user wants to sample the function at the beginning of the simulation time.

See also comments under https://github.com/precice/precice/pull/1414#discussion_r983736569 for further discussion.

## Author's checklist

* [x] I added a test to cover the proposed changes in our test suite. (not relevant)

## Reviewers' checklist

### DIFF
--- a/src/cplscheme/config/CouplingSchemeConfiguration.cpp
+++ b/src/cplscheme/config/CouplingSchemeConfiguration.cpp
@@ -979,12 +979,6 @@ void CouplingSchemeConfiguration::addDataToBeExchanged(
     if (from == accessor) {
       scheme.addDataToSend(exchange.data, exchange.mesh, requiresInitialization);
     } else if (to == accessor) {
-      // Checks for a serial coupling scheme, where initial data is received by first participant.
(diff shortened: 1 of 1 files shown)


uekerman (approved):
Makes sense

PR #1456, opened 2022-11-02
Increase timeouts

BenjaminRodenberg:
## Main changes of this PR

Timouts for tests are increased, because changes in #1414 make this necessary.

precice.cplscheme and precice.integration.Serial take longer with new tests introduced in https://github.com/precice/precice/pull/1414

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/cmake/CTestConfig.cmake
+++ b/cmake/CTestConfig.cmake
@@ -2,8 +2,8 @@
 # CTest
 #
 
-set(PRECICE_TEST_TIMEOUT_LONG 120 CACHE STRING "The timeout in seconds for longer tests.")
-set(PRECICE_TEST_TIMEOUT_SHORT 20 CACHE STRING "The timeout in seconds for shorter tests.")
+set(PRECICE_TEST_TIMEOUT_LONG 300 CACHE STRING "The timeout in seconds for longer tests.")
+set(PRECICE_TEST_TIMEOUT_SHORT 60 CACHE STRING "The timeout in seconds for shorter tests.")
 
 set(PRECICE_TEST_DIR "${preCICE_BINARY_DIR}/TestOutput")
 mark_as_advanced(PRECICE_TEST_DIR)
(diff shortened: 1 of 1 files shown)


BenjaminRodenberg:
@fsimonis would you mind taking a look at the CI pipeline? It looks to me like increasing the timeouts cleared some caching somewhere and broke the macOS pipeline. Not sure what happened, but I created this PR to isolate the breaking pipeline from the big PR #1414, where I actually need this change.

fsimonis:
The macos pipeline occasionally breaks and just needs a rerun.

Instead of increasing timeouts in general, I would prefer to increase timeouts as necessary.

Could you try switching the cplscheme tests from short to long timeouts in the cmake/Ctest file?

cmake/CTestConfig.cmake
```diff
@@ -2,7 +2,7 @@
 # CTest
 #
 
-set(PRECICE_TEST_TIMEOUT_LONG 120 CACHE STRING "The timeout in seconds for longer tests.")
+set(PRECICE_TEST_TIMEOUT_LONG 300 CACHE STRING "The timeout in seconds for longer tests.")
```
BenjaminRodenberg, line 5: @fsimonis I quickly checked the time required by the tests in #1414 and the maximum seems to be somewhere around 130s. Are the 300s too generous? I would like to avoid having to update this again when I add more tests.

fsimonis: Try to keep the timeout as tight as practical.
I would jump to 120 to 180 seconds as a timeout.

Tight timeouts prevent hanging tests from using too much CPU time. We are using the official runners, so we are limited by how many can run in parallel.

fsimonis, line 5:
```suggestion
set(PRECICE_TEST_TIMEOUT_LONG 180 CACHE STRING "The timeout in seconds for longer tests.")
```

fsimonis (approved):

PR #1462, opened 2022-11-04
Simplify compositional couplingscheme

### LINKED ISSUE #1453: Rework CompositionalCouplingScheme
## Problems of current implementation

- The current implementation is nicely explained in [Bernhard's thesis](https://www5.in.tum.de/pub/Gatzhammer2014_preCICE.pdf), section 4.1.5 (be careful, Bernhard calls these scheme "multi coupling", while "multi coupling" is now sth different in preCICE; let's stick with the term "compositional coupling").
- The current implementation is heavily based on the defined order of the coupling schemes in the xml configuration. This is not good as it is hard to understand and easy to mess up, see also #982 .
- There are meaningful, yet simple case that the current implementation cannot handle, #593  .
- On the other hand, the implementation supports cases that are numerically unstable: multiple implicit schemes for one participant, see also https://doi.org/10.1007/s00466-014-1113-2 .
- Parallel coupling schemes were not yet carefully considered for the current design.

## Suggestions

- The split of `CouplingScheme::advance` into `firstExchange` and `secondExchange` (suggested in #1451) could already solve #593 if `CompositionalCouplingScheme` does separate loops over the participants for both functions.
- Let's allow only one implicit scheme **per participant** and refer the user to "multi coupling" otherwise.
- Let's execute parallel-explicit schemes before serial-explicit schemes to allow setups where explicit schemes should be computed right away in every timestep (parallel-explicit) and cases where explicit schemes should be computed after the implicit scheme converged (serial-explicit).
- Let's make the implementation independent of the order in the configuration by defining a unique overall sorting. This is also important to avoid deadlocks in case sending becomes blocking (needs testing).

Suggested sorting:
- Parallel-explicit schemes come first
- Then, the one implicit scheme
- Finally, serial-explicit schemes

```cpp
_firstTime = true;

CompositionalCouplingScheme::firstExchange(){
if (_firstTime){
  for (scheme : parallelExplicitSchemes())
    scheme.firstExchange();
  _firstTime = false;
  }

theOneImplicitScheme.firstExchange();

if (theOneImplicitScheme.isConverged())
  {
  for (scheme : serialExplicitSchemes())
    scheme.firstExchange();
  }
```

and exactly the same for `secondExchange`.

## Problems with these suggestions

After `theOneImplicitScheme.firstExchange()` we cannot yet get information on `theOneImplicitScheme.isConverged()` :confused: 

## More

- Let's create a `coupling-scheme:multi-explicit` scheme and rename the current one to `coupling-scheme:multi-implicit`. 
- Unclear yet whether we should allow using a multi-coupling scheme withing a compositional coupling scheme. If yes, we have to treat it as the one implicit scheme, but could be simpler to forbid this.

fsimonis:
## Main changes of this PR

~~This PR is based on #1451, which is based on the develop-v3.0.0 branch, so **do not merge!**~~
_edit: these changes are superseed #1451_

The PR implements the suggestion from #1453 regarding the `CompositionalCouplingScheme`:

* allow at most one implicit scheme
* handle all explicit schemes in order of definition
* handle the optional implicit scheme last

This simplifies many aspects of the scheme.
Most importantly is that this can be implemented without iterating over steps.

Note that currently, multiple implicit schemes lead to an error message.
The ~~suggested workaround~~ solution is to use a multi-scheme instead.
This conceptually still allows to use many explicit schemes with many implicit schemes.

Closes #1453

I had to move `hasConverged` and `isImplicitScheme` from the `BaseCouplingScheme` into the `CouplingScheme` as these are required to distinguish implicit from explicit DummyCouplingSchemes.

## TODO

* [ ] implement an integration test using a multi coupling scheme

## Author's checklist

## Reviewers' checklist


src/cplscheme/CompositionalCouplingScheme.cpp
```diff
@@ -16,161 +18,116 @@ void CompositionalCouplingScheme::addCouplingScheme(
...
-  _couplingSchemes.emplace_back(couplingScheme);
+
+  if (!couplingScheme->isImplicitCouplingScheme()) {
+    _explicitSchemes.emplace_back(couplingScheme);
+    return;
+  }
+
+  PRECICE_CHECK(_implicitScheme == nullptr,
+                "You attempted to define a second implicit coupling-scheme for a participant, which is not allowed. "
+                "Please use a multi coupling-scheme for true implicit coupling of multiple participants.");
```
uekerman, line 29: Can we work with xml tags here and print the name of the participant? And could be nicer to move this to the configuration class.

src/cplscheme/CompositionalCouplingScheme.cpp
```diff
@@ -321,131 +257,68 @@ void CompositionalCouplingScheme::markActionFulfilled(
...
-    }
-    if (converged) {
-      PRECICE_DEBUG("Active implicit schemes converged");
-      for (SchemesIt it = _activeSchemesBegin; it != _activeSchemesEnd; it++) {
-        it->onHold = false;
-      }
-      _activeSchemesBegin = _activeSchemesEnd;
-    }
+  if (_iterating) {
+    PRECICE_DEBUG("Rerunning implcit scheme");
```
uekerman, line 293:
```suggestion
    PRECICE_DEBUG("Rerunning implicit scheme");
```

tests/serial/circular/Explicit.cpp
```diff
@@ -0,0 +1,21 @@
+#ifndef PRECICE_NO_MPI
+
+#include "helper.hpp"
+#include "testing/Testing.hpp"
+
+#include <precice/SolverInterface.hpp>
+
+BOOST_AUTO_TEST_SUITE(Integration)
+BOOST_AUTO_TEST_SUITE(Serial)
+BOOST_AUTO_TEST_SUITE(Circular)
```
uekerman, line 10: Why not adding this to the `three-solvers` namespace?

uekerman (commented):
Goes in the right direction. We need more testcases to understand whether we really cover all cases. We could sketch some together if you want.

> The suggested workaround is to use a multi-scheme instead

I would not call it "workaround", it's the actual solution.

src/cplscheme/CompositionalCouplingScheme.cpp
fsimonis: Moving this to the configuration class gives us access to the Participant's name, so that would help.

tests/serial/circular/Explicit.cpp
fsimonis: I would like to add more circular tests with more participants.

fsimonis:
Rebased and ready for review.

@uekerman and I agreed that the hanging tests in #1478 are caused by the order the partitions are handled and are not caused by these changes.

src/cplscheme/CompositionalCouplingScheme.cpp
```diff
@@ -15,96 +18,114 @@ void CompositionalCouplingScheme::addCouplingScheme(
...
+    totalChanges.insert(totalChanges.end(), remoteChanges.begin(), remoteChanges.end());
+  }
+  return totalChanges;
+}
+
+void CompositionalCouplingScheme::firstExchange()
+{
+  PRECICE_TRACE();
+  for (const auto scheme : schemesToRun()) {
+    scheme->firstExchange();
```
davidscn, line 89: Maybe a matter of taste but we could use `for_each` here and in several other places.

src/cplscheme/CompositionalCouplingScheme.cpp
```diff
@@ -113,38 +134,30 @@ std::vector<std::string> CompositionalCouplingScheme::getCouplingPartners() cons
...
   PRECICE_TRACE();
-  double time = std::numeric_limits<double>::max();
-  for (const Scheme &scheme : _couplingSchemes) {
-    if (not scheme.onHold) {
-      time = std::min(time, scheme.scheme->getTime());
-    }
+  auto schemes = allSchemes();
+  auto time    = std::numeric_limits<double>::max();
+  for (auto scheme : allSchemes()) {
+    time = std::min(time, scheme->getTime());
```
davidscn, line 160: `std::min_element`?

src/cplscheme/CompositionalCouplingScheme.cpp
```diff
@@ -154,10 +167,8 @@ int CompositionalCouplingScheme::getTimeWindows() const
 {
   PRECICE_TRACE();
   int timeWindows = std::numeric_limits<int>::max();
-  for (const Scheme &scheme : _couplingSchemes) {
-    if (not scheme.onHold) {
-      timeWindows = std::min(timeWindows, scheme.scheme->getTimeWindows());
-    }
+  for (auto scheme : allSchemes()) {
+    timeWindows = std::min(timeWindows, scheme->getTimeWindows());
```
davidscn, line 171: `std::min_element`?

src/cplscheme/CompositionalCouplingScheme.cpp
```diff
@@ -191,12 +200,8 @@ double CompositionalCouplingScheme::getThisTimeWindowRemainder() const
...
   PRECICE_TRACE();
   double maxRemainder = 0.0;
-  for (const Scheme &scheme : _couplingSchemes) {
-    if (not scheme.onHold) {
-      if (scheme.scheme->getThisTimeWindowRemainder() > maxRemainder) {
-        maxRemainder = scheme.scheme->getThisTimeWindowRemainder();
-      }
-    }
+  for (auto scheme : allSchemes()) {
+    maxRemainder = std::max(maxRemainder, scheme->getThisTimeWindowRemainder());
```
davidscn, line 204: `max_element`?

src/cplscheme/CompositionalCouplingScheme.cpp
```diff
@@ -242,9 +241,10 @@ bool CompositionalCouplingScheme::isActionRequired(
 {
   PRECICE_TRACE(actionName);
   bool isRequired = false;
-  for (const Scheme &scheme : _couplingSchemes) {
-    if (not scheme.onHold) {
-      isRequired |= scheme.scheme->isActionRequired(actionName);
+  for (auto scheme : allSchemes()) {
+    if (scheme->isActionRequired(actionName)) {
+      isRequired = true;
+      break;
```
davidscn, line 247: `std::any_of`?

src/cplscheme/CompositionalCouplingScheme.cpp
```diff
@@ -255,131 +255,68 @@ void CompositionalCouplingScheme::markActionFulfilled(
     const std::string &actionName)
 {
   PRECICE_TRACE(actionName);
-  for (const Scheme &scheme : _couplingSchemes) {
-    if (not scheme.onHold) {
-      scheme.scheme->markActionFulfilled(actionName);
-    }
+  for (auto scheme : allSchemes()) {
+    scheme->markActionFulfilled(actionName);
```
davidscn, line 259: for_each?

davidscn (commented):
I didn't check the coupling logic, only the code itself, as my coupling scheme knowledge is not exhaustive enough. I didn't mark all of them, but I guess there are some places where we could use STL algortihms.

src/cplscheme/CompositionalCouplingScheme.hpp
```diff
@@ -204,63 +212,31 @@ class CompositionalCouplingScheme final : public CouplingScheme {
...
-   * @brief Determines the current set of active coupling schemes.
-   *
-   * Is called in advance, after all active sub-schemes are advanced.
-   *
-   * @return True, if active schemes have changed and should be treated within
-   *         the same call of advance().
-   */
-  bool determineActiveCouplingSchemes();
+  /// Returns all schemes at the beginning of a timestep or the implicit one until convergence
+  std::vector<CouplingScheme *> schemesToRun() const;
```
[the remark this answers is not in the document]
fsimonis: It's a collection of non-owning handles to objects.
References don't work as they are not assignable.

This is _still_ a perfectly valid use for raw pointers.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -194,10 +200,31 @@ void BaseCouplingScheme::advance()
...
+
+CouplingScheme::ChangedMeshes BaseCouplingScheme::secondSynchronization()
+{
+  return {};
+}
+
+void BaseCouplingScheme::secondExchange()
+{
+  PRECICE_TRACE(_timeWindows, _time);
+  checkCompletenessRequiredActions();
```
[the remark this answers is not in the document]
fsimonis: I didn't modify the method in this PR. TBH both names sound weird.

fsimonis:
The STL algorithms are a bit of a nuisance, as I need to iterate over a temporary collection.

This gets super verbose without ranges, as I have to first assign to a temporary, pack everything into a lambda, call the algo and often even postprocess the result (minelement returns an iterator).
Ranged for loops are just so much more readable in these situations.

fsimonis:
Concrete example of an easy case.
Are the alternatives really more readable?

## Using ranged for
```cpp
void CompositionalCouplingScheme::addComputedTime(double timeToAdd)
{
  PRECICE_TRACE(timeToAdd);

  for (const auto scheme : schemesToRun()) {
    scheme->addComputedTime(timeToAdd);
  }
}
```

## Using std::for_each
```cpp
void CompositionalCouplingScheme::addComputedTime(double timeToAdd)
{
  PRECICE_TRACE(timeToAdd);
  auto schemes = schemesToRun();
  std::for_each(schemes.begin(), schemes.end(), [timeToAdd](const auto scheme){
    scheme->addComputedTime(timeToAdd);
  });
}
```

## Using boost::range::for_each
```cpp
void CompositionalCouplingScheme::addComputedTime(double timeToAdd)
{
  PRECICE_TRACE(timeToAdd);
  boost::range::for_each(schemesToRun(), [timeToAdd](const auto scheme){
    scheme->addComputedTime(timeToAdd);
  }
}
```

src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -195,10 +195,13 @@ class BaseCouplingScheme : public CouplingScheme {
   /// Receives result of first advance, if this has to happen inside SolverInterface::initialize(), see CouplingScheme.hpp
   void receiveResultOfFirstAdvance() override final;
 
-  /**
-   * @brief Advances the coupling scheme.
-   */
-  void advance() override final;
+  ChangedMeshes firstSynchronization(const ChangedMeshes &changes) override final;
```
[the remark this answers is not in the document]
fsimonis: I extended the docs in CouplingScheme of all steps.

src/cplscheme/CompositionalCouplingScheme.cpp
fsimonis: I used `transform_reduce` instead.

src/cplscheme/CompositionalCouplingScheme.cpp
fsimonis: I used `transform_reduce` instead.

src/cplscheme/CompositionalCouplingScheme.cpp
fsimonis: I used `transform_reduce` instead.

src/cplscheme/CompositionalCouplingScheme.cpp
fsimonis: This will be completely changed in #1487, so I rather leave it for now.

src/cplscheme/CompositionalCouplingScheme.cpp
fsimonis: Again. Just messy until C++20

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1964,6 +1963,19 @@ void SolverInterfaceImpl::syncTimestep(double computedTimestepLength)
   }
 }
 
+void SolverInterfaceImpl::advanceCouplingSchemes()
+{
+  PRECICE_DEBUG("Advance coupling scheme");
```
uekerman, line 1968: That's what TRACE would be for, no?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -415,8 +415,7 @@ double SolverInterfaceImpl::advance(
     performDataActions({action::Action::WRITE_MAPPING_POST}, time);
   }
 
-  PRECICE_DEBUG("Advance coupling scheme");
-  _couplingScheme->advance();
+  advanceCouplingSchemes();
```
uekerman, line 418:
```suggestion
  advanceCouplingScheme();
```
? There is only one.
I am also not sure whether we need this function.
Very likely, we might want to do more things in between both exchanges.

src/cplscheme/CouplingScheme.hpp
```diff
@@ -98,7 +100,54 @@ class CouplingScheme {
...
    * Does not necessarily advance in time.
    */
-  virtual void advance() = 0;
+  void advance()
+  {
+    firstSynchronization({});
+    firstExchange();
+    secondSynchronization();
+    secondExchange();
+  }
```
uekerman, line 109: This is a placeholder, we could now remove, no?

src/cplscheme/CouplingScheme.hpp
```diff
@@ -98,7 +100,54 @@ class CouplingScheme {
...
+   *
+   * @pre \ref firstSynchronization() was called
+   */
+  virtual void firstExchange() = 0;
+
+  /** Receive mesh changes from remote participants in the second step.
+   *
+   * At this point, the remote participant may have changed the meshes if
+   * is coupled using a serial coupling scheme.
+   * So we only need to receive the remote changes here.
```
uekerman, line 134:
```suggestion
   * At this point, the remote participant may have changed the meshes if
   * it is coupled using a serial coupling scheme.
   * So we only need to receive the remote changes here.
```
The logical connection between both sentences is not clear.

src/cplscheme/CouplingScheme.hpp
```diff
@@ -173,6 +222,12 @@ class CouplingScheme {
 
   /// Returns a string representation of the current coupling state.
   virtual std::string printCouplingState() const = 0;
+
+  /// Returns true if the scheme or one subscheme is implicit
+  virtual bool isImplicitCouplingScheme() const = 0;
+
+  /// Returns false if the scheme is implicit and hasnt converged
```
uekerman, line 229:
```suggestion
  /// Returns false if the scheme is implicit and hasn't converged
```

src/cplscheme/BaseCouplingScheme.hpp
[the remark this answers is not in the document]
uekerman: Could we do a "see" or a "copydoc" here?

src/cplscheme/SerialCouplingScheme.cpp
```diff
@@ -87,38 +87,42 @@ void SerialCouplingScheme::performReceiveOfFirstAdvance()
...
-
   if (doesFirstStep()) { // first participant
     PRECICE_DEBUG("Sending data...");
     sendTimeWindowSize();
     sendData(getM2N(), getSendData());
-    PRECICE_DEBUG("Receiving data...");
+    PRECICE_DEBUG("Receiving convergence data...");
     if (isImplicitCouplingScheme()) {
-      convergence = receiveConvergence(getM2N());
+      receiveConvergence(getM2N());
```
uekerman, line 98: Seems wrong that we `receiveConvergence` here and not in `exchangeSecondData`.

uekerman (commented):
It probably makes a lot of sense if @BenjaminRodenberg could have a closer look here as he introduced many of the functions that are now refactored.

With this refactoring, we should also try to simplify (remove) the helper functions that @BenjaminRodenberg introduce for the initialization (such as `receiveResultOfFirstAdvance`).

BenjaminRodenberg:
`receiveResultOfFirstAdvance` and `moveWindowBeforeMapping` are already candidates for removal. See https://github.com/precice/precice/pull/1414#issuecomment-1335029160. I'm planning to put these changes into an independent PR quite soon.

src/precice/impl/SolverInterfaceImpl.cpp
fsimonis: This preserves the previous Debug message. I don't think that trace would be helpful here.

src/cplscheme/SerialCouplingScheme.cpp
fsimonis: Good catch!

src/cplscheme/BaseCouplingScheme.hpp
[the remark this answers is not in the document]
fsimonis: No need. Docs of overriding virtual functions are copied automatically.

src/cplscheme/CouplingScheme.hpp
fsimonis: Removed

PR #1464, opened 2022-11-10
Extend docs of isTimeWindowComplete

uekerman:
## Main changes of this PR

Extend docs of isTimeWindowComplete

This API function is very often leading to questions:
- At the end of every iteration or only at the end of the last iteration?
- Why do I need this function if it is equivalent with "write its check"?

I hope this additional doc clarifies these questions.

## Author's checklist

## Reviewers' checklist

* [ ] Do you understand the docs changes?

### DIFF
--- a/src/precice/SolverInterface.hpp
+++ b/src/precice/SolverInterface.hpp
@@ -275,8 +275,8 @@ public:
    *   than the time window.
    * - An implicit coupling iteration is not yet converged.
    *
-   * A time window is complete if neither reason holds: If we reach the
-   * end of the time window in the last implicit coupling iteration.
+   * Hence, a time window is complete if we reach the end of the time window
+   * and the implicit coupling has converged.
    *
    * For implicit coupling this condition is equivalent with the requirement to
    * write an iteration checkpoint. This is, however, not the case for explicit
--- a/src/precice/SolverInterface.hpp
+++ b/src/precice/SolverInterface.hpp
@@ -266,15 +266,21 @@ public:
   [[deprecated("Will be removed in 3.0.0. See https://github.com/precice/precice/issues/1223 and comment, if you need this function.")]] bool isWriteDataRequired(double computedTimestepLength) const;
 
   /**
-   * @brief Checks if the current coupling window is completed.
+   * @brief Checks if the current time window is completed.
    *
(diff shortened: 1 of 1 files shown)


fsimonis:
Potential question: What is the meaning of calling this before `advance()`?

src/precice/SolverInterface.hpp
```diff
@@ -266,16 +266,22 @@ class SolverInterface {
...
    *
-   * The following reasons require several solver time steps per time window
-   * step:
+   * The following reasons require several solver time steps per time window:
    * - A solver chooses to perform subcycling, i.e. using a smaller timestep
-   *   than the time window..
+   *   than the time window.
    * - An implicit coupling iteration is not yet converged.
    *
+   * A coupling window is complete if neither reason holds: If we reach the
```
BenjaminRodenberg, line 278:
```suggestion
   * A time window is complete if neither reason holds: If we reach the
```

src/precice/SolverInterface.hpp
```diff
@@ -266,16 +266,22 @@ class SolverInterface {
...
+   * end of the time window in the last implicit coupling iteration.
```
BenjaminRodenberg, line 279: * I think the part "neither reason holds" confuses more than it helps.
* What is the "last implicit coupling iteration"? Why not talking about "converged" like above?

Alternative:
> A coupling window is complete, if we reach the end of the time window and the implicit coupling has converged.

BenjaminRodenberg (changes requested):

BenjaminRodenberg:
> Potential question: What is the meaning of calling this before `advance()`?

Before one `advance()` is after another `advance()`. Basically this method asks `didPreviousAdvanceCallCompleteTheTimeWindow()`. Does this answer the question?

I think this method has no interaction with any other methods than `advance()`.

fsimonis:
> Before one advance() is after another advance().

Not if it is before the first advance. The precondition states that initialize must have been called.

So, this is valid:
```py
initialize()
while isCouplingOngoing():
   isTimeWindowComplete() # ????
   advance()
```

uekerman:
> So, this is valid

Should be valid IMO. The method should return `false` as the time window is not yet complete.

PR #1469, opened 2022-11-15
Fix for generating Ninja files

### LINKED ISSUE #1468: Cannot generate Ninja files with CMake
**Describe your setup**

Operating system (e.g. Linux distribution and version): _Arch Linux_
preCICE Version: 2.5.0

Last commit:
a7ae545c9b8a8b228180031eb6eb93c24f610c61

**Describe the problem**
Cannot generate Ninja files with CMake. 
Unix Makefiles are working fine.

**Step To Reproduce**
``` sh
mkdir build
cd build
cmake -GNinja  ..
```
Gives this error message:
```
CMake Error:
  Running

   '/usr/bin/ninja' '-C' '/home/tan/master/advanced_programming/project/precice/build' '-t' 'recompact'

  failed with:

   ninja: warning: phony target 'binprecice' names itself as an input; ignoring [-w phonycycle=warn]

  ninja: error: build.ninja:4102: multiple rules generate binprecice
```

tanaytekin:
## Main changes of this PR

With removing `BYPRODUCTS binprecice`, I was able to generate ninja files.

>
`Ninja requires custom command byproducts to be explicit.`
(https://cmake.org/cmake/help/latest/policy/CMP0058.html)

It seems to be CMake/Ninja doesn't like using same names for `BYPRODUCTS` and the target name. That's why CMake/Ninja is complaining since `binprecice` is  defined as both `BYPRODUCT` and custom target. 

closes #1468 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -369,7 +369,6 @@ if (PRECICE_BUILD_TOOLS)
     COMMAND ${CMAKE_COMMAND} -E create_symlink precice-tools binprecice
     DEPENDS precice-tools
     WORKING_DIRECTORY ${CMAKE_CURRENT_BINARY_DIR}
-    BYPRODUCTS binprecice
     )
 endif()
 


fsimonis:
Thanks for the fix!

We will remove this link in the coming days due to the release preparations of preCICE v 3.0. While short lived, this will prevent issues until then.

PR #1476, opened 2022-11-16
Remove utils::ptr_vector<T>

### LINKED ISSUE #1164: Replace utils::ptr_vector<T> with smart pointers
`utils::ptr_vector<T>` serves a similar purpose like smart pointers from `std`. We should remove this implementation in favor of smart pointers from `std`.

I already took care of replacing `utils::ptr_vector<T>` with `std::unique_ptr` in https://github.com/precice/precice/pull/1159, but there are many more places, where `utils::ptr_vector<T>` is used.

Timo-Schrader:
Removed usage of `utils::ptr_vector` in class Participant and switched to `std::vector<MappingContext>`. This includes changing function headers where necessary.

One important part is to remove `const` qualifiers. In the original implementation, `const` class methods where bypassed by modifying/calling non-const methods from the underlying pointers.

Now, the only occurrence of `utils::ptr_vector` in the repo is in PointerVectorTest.cpp.

## Main changes of this PR

- Removed usage of `utils::ptr_vector` and C-style pointers of class `MappingContext`
- Use `std::vector<MappingContext>` instead
- Change function signatures and qualifiers where necessary

Closes #1164.

## Author's checklist

## Reviewers' checklist


src/precice/impl/Participant.hpp
```diff
@@ -83,10 +83,10 @@ class Participant {
       int                  interpolationOrder);
 
   /// Adds a configured read \ref Mapping to the Participant
-  void addReadMappingContext(MappingContext *mappingContext);
+  void addReadMappingContext(MappingContext mappingContext);
 
   /// Adds a configured write \ref Mapping to the Participant
-  void addWriteMappingContext(MappingContext *mappingContext);
+  void addWriteMappingContext(MappingContext mappingContext);
```
davidscn, line 89: Wouldn't it make sense to use references here as well?

Timo-Schrader: Yes indeed, will change it to const reference.

davidscn:
Changes look good, thanks! @BenjaminRodenberg @fsimonis do we also need to remove the corresponding [PointerVector](https://github.com/precice/precice/blob/develop/src/utils/PointerVector.hpp) class and the [weird test](https://github.com/precice/precice/blob/develop/src/utils/tests/PointerVectorTest.cpp)?

fsimonis:
Yes, let's please get rid of the ptr_vector :+1:

PR #1477, opened 2022-11-17
Fix missing check output

fsimonis:
## Main changes of this PR

`precice-tools check config.xml` does not output anything if the logger is disabled.

This PR enforces a fixed logging sink.

Undisplayed errors are not helpful

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/1477.md
+++ b/docs/changelog/1477.md
@@ -1,0 +1,1 @@
+- Fixed `precice-tools check` not displaying errors if the logger is disabled in the checked config.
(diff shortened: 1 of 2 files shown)


fsimonis:
@davidscn You ran into this issue before. Can you check if this solves it?

davidscn:
Would it make sense to merge #1420 first before testing the proposed changes here?

fsimonis:
@davidscn #1420 merged, rebased and pushed.

davidscn (approved):
Works as expected, thanks! From what I understand you just prescribe an 'appropriate' logger, right? Maybe consider prefixing the logger with `precice-tools` or something similar, simply because it is easy to do

PR #1479, opened 2022-11-17
Add second muscle test (failing)

carme-hp:
## Main changes of this PR

Add a  second muscle test.

The OneActivatedMuscle test consists of 4 participants: M1, M1SM, Tendon and M2SM

@uekerman  test does not pass!

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/tests/serial/compositional/OneActivatedMuscle.cpp
+++ b/tests/serial/compositional/OneActivatedMuscle.cpp
@@ -40,8 +40,8 @@ BOOST_AUTO_TEST_CASE(OneActivatedMuscle)
     auto stretchMeshID = interface.getMeshID("Stretch_M1SM_Mesh");
     interface.setMeshVertices(stretchMeshID, 1, signalCoords.data(), stretchVertexIDs.data());
 
-    activationDataID   = interface.getDataID("Activation1", activationMeshID);
-    stretchDataID      = interface.getDataID("Stretch1", stretchMeshID);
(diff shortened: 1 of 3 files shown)


fsimonis:
The tests currently fail, as the implicitly coupled solvers don't write their checkpoints.
So you need something like this:
```cpp
if (interface.isActionRequired(precice::constants::actionWriteIterationCheckpoint())) {
  interface.markActionFulfilled(precice::constants::actionWriteIterationCheckpoint());
}
```
Same for the reading the checkpoint after calling advance.

After some copy-pasta, I get the following error:
```
error: in "Integration/Serial/Compositional/OneActivatedMuscle": check receivedDisplacements == displacements1 has failed
  - mismatch at position 0: [0 == 4.2000000000000002] is false: absolute value exceeds tolerance [|4.2| > 1e-09]
  - mismatch at position 1: [0 == 1.3999999999999999] is false: absolute value exceeds tolerance [|1.4| > 1e-09]
Failure occurred in a following context:
    Test context represents "Tendon" and runs on rank 0 out of 1.
```

Just realized that I didn't remove the check that prevents composition of the multi scheme.

You may have to remove these lines:
https://github.com/carme-hp/precice/blob/16efa42a64736071635744c8d732a288febfdf00/src/cplscheme/config/CouplingSchemeConfiguration.cpp#L387-L389

PR #1480, opened 2022-11-17
Fix no-return type warnings in release builds

davidscn:
## Main changes of this PR

Fixes `-Wreturn-type` compiler warnings in release build. Probably introduced in #1366 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/time/Storage.cpp
+++ b/src/time/Storage.cpp
@@ -62,12 +62,10 @@ void Storage::clear(bool keepZero)
 
 Eigen::VectorXd Storage::getValueAtOrAfter(double before)
 {
-  for (auto &sample : _sampleStorage) {
-    if (math::greaterEquals(sample.first, before)) {
-      return sample.second;
-    }
-  }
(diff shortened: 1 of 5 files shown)


src/time/Storage.cpp
```diff
@@ -62,12 +62,10 @@ void Storage::clear(bool keepZero)
...
-  for (auto &sample : _sampleStorage) {
-    if (math::greaterEquals(sample.first, before)) {
-      return sample.second;
-    }
-  }
-  PRECICE_ASSERT(false, "no value found!");
+  auto sample = std::find_if(_sampleStorage.begin(), _sampleStorage.end(), [&before](const auto &s) { return math::greaterEquals(s.first, before); });
+  PRECICE_ASSERT(sample != _sampleStorage.end(), "no value found!");
+
+  return (*sample).second;
```
fsimonis, line 68:
```suggestion
  return sample->second;
```

fsimonis (approved):
Only one comment

Issue #1417, opened 2022-08-15
Rename `use-qr-decomposition` to something more appropriate


davidscn:
**Please describe the problem you are trying to solve.**
In general, preCICE has two flavors of data mappings: one using PETSc, which is the default (assuming it was compiled with PETSc) and one using Eigen, which can be queried enabling the option `use-qr-decopmosition=true`. In #1372 we replaced the QR decomposition partially by a Cholesky decomposition and the name of the option doesn't fit any more. In addition, `use-qr-decomposition` might be a bit confusing for users, because they are probably not familiar with the data mapping and the details. 

**Describe the solution you propose.**
We should rename the option to something more appropriate, such that the option a) reflects what's going on under the hood and b) the user gets an intuitive guess when to enable the option (probably for all non-massively parallel cases). @uekerman proposed `use-direct-solver`, which might be a valid option. However, if we decide to also try a (sparse) direct solver in our PETSc implementation, this name will again lead to conflicts in the decision tree. 
As alternatives, I would consider something like `use-distributed-mapping` or `gather-scatter-mapping`, which would reflect the real difference in the implementation. Disadvantage might be that `distributed-mapping` sounds a bit more sophisticated and users simply might not enable it (which I would always recommend in serial runs).

fsimonis:
What would be the user perspective here?
The user selects an rbf and adjusts some shape parameter.
Additionally, they may set some axis as dead.

The implementation/backend really should only need configuration if there are performance issues.
preCICE should select what's most appropriate by default.

Based on this, couldn't we add a "backend" or "method" or "implementation" attribute? This would allow to select gather-scatter-qr or petsc-gmres.
Default should be inferred from the system, so some sort of "guess" value could make that easier to read.

BenjaminRodenberg:
Another line of thought: What would a user search for when looking for this option? How can a user find this option in our documentation? Is there maybe already a fitting name out there?

In our documentation we write

> For small/medium size problems, the QR decomposition is enough and you don’t need to install anything else. However, this follows a gather-scatter approach, which limits the scalability. For large problems, the GMRES solver performs better than the QR decomposition. For this, you need to [build preCICE with PETSc](https://github.com/precice/precice/wiki/Dependencies#petsc-optional). If you built with PETSc, the default is always GMRES. If you still want to use the QR decomposition, you can use the option use-qr-decomposition.

What I understand here is: "I've built preCICE with PETSc, but I want to use the no PETSc option anyway. How can I do this?"

[...]

Issue #1418, opened 2022-08-18
Inconsistent `BoundingBox` implementation


davidscn:
**Describe the problem**
While working on #1273, I stumbled across our `BoundingBox` class and was not completely sure about several design decisions. There are functions which (seemingly arbitrarily) do nothing without telling the user, in case the bounding box is defined 'empty', e.g.,

https://github.com/precice/precice/blob/b51fd77b2bdda66b4942cfbd1c4a6d9d34055c87/src/mesh/BoundingBox.cpp#L133-L141

or 

https://github.com/precice/precice/blob/b51fd77b2bdda66b4942cfbd1c4a6d9d34055c87/src/mesh/BoundingBox.cpp#L143-L157

Interestingly, overloaded versions of these functions don't care about the emptyness of the box

https://github.com/precice/precice/blob/b51fd77b2bdda66b4942cfbd1c4a6d9d34055c87/src/mesh/BoundingBox.cpp#L123-L131

 or 

https://github.com/precice/precice/blob/b51fd77b2bdda66b4942cfbd1c4a6d9d34055c87/src/mesh/BoundingBox.cpp#L115-L121

Overall, the definition of empty is also not completely clear to me, shouldn't this be implemented as `GreaterEqual` and not as `Greater` (and also implemented in a one-liner). The documentation in the code reads that 'empty' is the same as default state, which is strictly speaking not the case (and not what the function checks)

https://github.com/precice/precice/blob/b51fd77b2bdda66b4942cfbd1c4a6d9d34055c87/src/mesh/BoundingBox.cpp#L45-L53

Having said this, the `getArea` function doesn't seem to make sense to me: if we have dead axis, the bounding box is used to be empty in at least one direction, which will fail in debug mode and just return a value in release mode

https://github.com/precice/precice/blob/b51fd77b2bdda66b4942cfbd1c4a6d9d34055c87/src/mesh/BoundingBox.cpp#L95-L103

**Expected behaviour**
More of a place to discuss the points above. In principle: a consistent, well-documented class with tests.

fsimonis:
The BoundingBox class is a weird one.

IIRC, many design decisions are based on the old implementation which was simply a vector of min max pairs for each dimension.

Cleaning this class up would be great.

oguzziya:
[...]

PR #1483, opened 2022-11-18
Add partition of unity RBF data mapping

### LINKED ISSUE #1273: Partition of unity based data mappings

davidscn:
## Main changes of this PR

Adds the partition of unity RBF mapping as a new mapping variant.

There are a few details yet to be clarified and a few TODOs in the code, but I'm sure these changes are ready for a first review. We need to fix a user interface in the configuration file.

Post TODO: Adjust the auto-selection of our mapping-config.

Closes #1273.

## Author's checklist

* [x] Fixing a proper user-interface (related to #1417)

## Reviewers' checklist


src/query/Index.hpp
```diff
@@ -72,6 +72,9 @@ class Index {
   /// Get n number of closest vertices to the given vertex
   VertexMatch getClosestVertex(const Eigen::VectorXd &sourceCoord);
 
+  /// Get n number of closest vertices to the given vertex
+  std::vector<VertexID> getClosestVertices(const mesh::Vertex &centerVertex, unsigned int n);
+
```
uekerman, line 77: Seems like the documentation in line 72 is wrong?
Why `Eigen::VectorXd` vs `mesh::Vertex`?

src/mapping/Partition.hpp
```diff
@@ -0,0 +1,315 @@
...
+#include "precice/types.hpp"
+#include "utils/Event.hpp"
+
+namespace precice {
+
+namespace mapping {
+
+/**
+ * @brief Mapping using partition of unity decomposition strategies
+ */
```
uekerman, line 20: Please extend / modify. "Mapping" might also be misleading here as the class does not inherit from `Mapping`.

src/mapping/Partition.hpp
```diff
@@ -0,0 +1,315 @@
...
+            mesh::Vertex      center,
+            double            radius,
+            double            parameter,
+            std::vector<bool> deadAxis,
+            Polynomial        polynomial,
+            unsigned int      targetSize,
+            mesh::PtrMesh     inputMesh,
+            mesh::PtrMesh     outputMesh);
+
+  /// Removes a computed mapping.
```
uekerman, line 45: Is "remove" the right word here?

src/mapping/Partition.hpp
```diff
@@ -0,0 +1,315 @@
...
+  void mapConservative(mesh::PtrData inputData, mesh::PtrData outputData);
+
+  /// @brief set the normalized weight in the normalizedWeight data structure
+  void setNormalizedWeight(double normalizedWeight, VertexID vertexID);
+
+  /// Compute the weight for a given vertex
+  double computeWeight(const mesh::Vertex &v) const;
+
+  /// Evaluate the interpolant for a given vertex
+  double getInterpolatedValue(const VertexID id, unsigned int direction) const;
```
uekerman, line 61:
```suggestion
  double getInterpolatedValue(const VertexID vertexID, unsigned int direction) const;
```
`dimension` or `axis` instead of `direction`?

src/mapping/Partition.hpp
```diff
@@ -0,0 +1,315 @@
...
+
+  /// Indicate whether a vertex lies within the partition or not
+  bool isVertexInside(const mesh::Vertex &v) const;
+
+  /// Indicate that the mapped data is now considered as outdated by setting _hasMappedData to false
+  void setMappingFinished();
```
uekerman, line 67:
```suggestion
  /// Indicate that the mapped data is now considered as outdated by setting _hasMappedData to false
  void invalidateMapping();
```
or similar?

src/mapping/Partition.hpp
```diff
@@ -0,0 +1,315 @@
...
+  bool _emptyPartition     = false;
+
+  /// Partition center
+  mesh::Vertex _center;
+
+  /// Partition radius
+  const double _radius;
+
+  /// Array with one entry per data component. A plain array would require knowledge about the number of components
+  std::array<Eigen::VectorXd, 3> resultVector;
```
uekerman, line 92: I am not sure I understand what "data component" means? "3" for vector-valued data? Maybe we could also find a better name for "resultVector".

`_resultVector`?

src/mapping/Partition.hpp
```diff
@@ -0,0 +1,315 @@
...
+
+  /// vector containing the normalized weights (only for conservative mappings)
+  Eigen::VectorXd normalizedWeights;
```
uekerman, line 95:
```suggestion
  Eigen::VectorXd _normalizedWeights;
```
more below

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,385 @@
...
+#include "utils/IntraComm.hpp"
+
+namespace precice {
+extern bool syncMode;
+
+namespace mapping {
+
+/**
+ * @brief Mapping using partition of unity decomposition strategies
+ */
```
uekerman, line 24: Please extend. Explain how classes are connected.

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,385 @@
...
+  double averagePartitionRadius = 0;
+
+  /// true if the mapping along some axis should be ignored
+  /// has currently only dim x false entries, as integrated polynomials are irrelevant
+  std::vector<bool> _deadAxis;
+
+  Polynomial _polynomial;
+
+  // Holds the output vertex -> partition association. Outer vector has the size of the output mesh and inner vector size of the associated partitions
+  std::vector<std::vector<VertexID>> _partMap;
```
uekerman, line 82: Would a `map` be easier to understand?
```suggestion
  // Holds the partition -> output vertex association. Outer vector has the size of the output mesh and inner vector size of the associated partitions
  std::vector<std::vector<VertexID>> _partMap;
```

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,385 @@
...
+    Polynomial          polynomial,
+    unsigned int        verticesPerPartition,
+    double              relativeOverlap,
+    bool                projectToInput)
+    : Mapping(constraint, dimension),
+      _parameter(parameter), _verticesPerPartition(verticesPerPartition), _relativeOverlap(relativeOverlap), _projectToInput(projectToInput), _polynomial(polynomial)
+{
+  PRECICE_ASSERT(this->getDimensions() <= 3);
+  PRECICE_CHECK(_relativeOverlap < 1, "The relative overlap has to be smaller than one.");
+  PRECICE_CHECK(_verticesPerPartition > 0, "The number of vertices per partition has to be greater zero.");
```
uekerman, line 109: This should be handled in the configuration. Assertions should be enough here.

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,385 @@
...
+    setOutputRequirement(Mapping::MeshRequirement::FULL);
+  } else {
+    setInputRequirement(Mapping::MeshRequirement::VERTEX);
+    setOutputRequirement(Mapping::MeshRequirement::VERTEX);
+  }
+  // Simply set each axis to be active according to the space dimension.
+  // We could even set each axis to be true, which would just lead to 'zero' columns in the polynomial QR.
+  // Integrated polynomials are anyway not supported with PoU
+  _deadAxis.clear();
+  std::fill_n(std::back_inserter(_deadAxis), getDimensions(), false);
```
uekerman, line 122: I am not sure yet if I agree, but anyway: we should handle this in the configuration.

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,385 @@
...
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+void PartitionOfUnityMapping<RADIAL_BASIS_FUNCTION_T>::computeMapping()
+{
+  PRECICE_TRACE();
+
+  precice::utils::Event e("map.pou.computeMapping.From" + this->input()->getName() + "To" + this->output()->getName(), precice::syncMode);
+
+  // Recompute the whole partitioning
+  this->clear();
```
uekerman, line 177: Feels a bit odd. Normally, I would expect a `assertion(not hasComputedMapping())` or similar here and do the "clearing" somewhere outside.

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,385 @@
...
+    outMesh = this->output();
+  }
+
+  // Step 1: get a tentative partitioning consisting of centers and a radius from one of the available algorithms
+  auto [averagePartitionRadius_, centerCandidates] = partitioner::createUniformBlockPartitioning(inMesh, outMesh, _relativeOverlap, _verticesPerPartition, _projectToInput);
+
+  averagePartitionRadius = averagePartitionRadius_;
+  PRECICE_ASSERT(averagePartitionRadius > 0);
+
+  // Step 2: check, which of the resulting partitions are be non-empty (in term sof the output vertices) and register the partition centers in a mesh
```
uekerman, line 195:
```suggestion
  // Step 2: check, which of the resulting partitions are non-empty (in terms of the output vertices) and register the partition centers in a mesh
```

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,385 @@
...
+  mesh::Mesh centerMesh("partitionCentersMesh", this->getDimensions(), mesh::Mesh::MESH_ID_UNDEFINED);
```
uekerman, line 196: Should we add the name of the input mesh here? What happens if multiple PoU mappings are used?

uekerman (commented):
Looks promising already :+1:
So far, I only had a big picture view and I did not yet try it out.

I struggle a bit with the name of the file `CreatePartitioning` and the namespace `partitioner`. `Partition` vs `Partitioner` vs `Partitioning`, how are they connected?
It gets even more complicated: we already have a package `partition` with a class `Partition`. Maybe we could find a different name here, also to clearly distinguish between parallel partition (mesh on one rank) and partitions meant here.
Maybe the connection between `PoUMapping` and `CreatePartition` could be a "has-a" connection?

`CreatePartitioning` and `Partition` could probably go into `impl`?

Let's add default values to all configuration attributes where we already know that there is a good safe choice.

Should we merge this before or after #1418?

davidscn:
> Should we merge this before or after https://github.com/precice/precice/issues/1418?

Would be better to merge the bounding box refactoring first, the changes here are non-intrusive and can easily be adapted to the new class layout.

davidscn:
While investigating the performance of the OpenFOAM-adapter using the 3D bending-flap tutorial, we observed a rather huge amount of time spent in the mapping part from the more refined solid side to the fluid side. Looking at the total amount of time in the fluid summary log, we had the following timings for the PETSc RBF mappings:

```bash
advance/map.pet.computeMapping.FromSolid-writeToFluid-Mesh-Nodes |          1 |        408 |        408 |        408 |        408 |    0.00163 |
advance/map.pet.mapData.FromSolid-writeToFluid-Mesh-Nodes        |        141 |      43943 |        526 |         17 |        311 |      0.176 |
```
i.e., we have almost 44 seconds spent in the mapData (advance) call of preCICE.

Using the POU method implemented here, we get the following timings:
```bash
 advance/map.pou.computeMapping.FromSolid-writeToFluid-Mesh-Nodes |          1 |         37 |         37 |         37 |         37 |   0.000215 |
        advance/map.pou.mapData.FromSolid-writeToFluid-Mesh-Nodes |        141 |        495 |          3 |          2 |          3 |    0.00287 |
```
i.e., the time spent in `mapData` is around 100 times smaller. The `computeMapping` call is around 10 times faster (note that a rescaling is computed in the PETSc version).

As a disclaimer: I still have to add an early exit from the 'computeMapping' function here in order to bail out at the right time when a rank is empty (has no interface vertices). For the measurement, I added an easy bail-out strategy, but I want think about it again and add a clean (maybe in the end even the same) version .

src/mapping/Partition.hpp
```diff
@@ -0,0 +1,315 @@
...
+  bool isVertexInside(const mesh::Vertex &v) const;
+
+  /// Indicate that the mapped data is now considered as outdated by setting _hasMappedData to false
+  void setMappingFinished();
+
+  /// Number of input vertices this partition operates on
+  unsigned int getNumberOfInputVertices() const;
+
+  /// The center coordinate of this partition
+  std::array<double, 3> getCenterCoords() const;
```
fsimonis, line 73: Is there any reason not to use `Eigen::Vertex3D` or `Eigen::VertexXD` instead?.

I only used this in the `Vertex` to prevent dynamic allocations and simplify the interface to `boost.geometry`.

src/mapping/Partition.hpp
```diff
@@ -0,0 +1,315 @@
...
+
+  /// Indicates whether there are valid output points for this partition
+  bool isEmpty() const;
```
fsimonis, line 76: Maybe sticking to the STL-style `empty()` is easier to understand.

src/mapping/Partition.hpp
fsimonis: `markAsDirty` or `invalidate` could be shorter.

src/mapping/Partition.hpp
```diff
@@ -0,0 +1,315 @@
...
+  double getInterpolatedValue(const VertexID id, unsigned int direction) const;
+
+  /// Indicate whether a vertex lies within the partition or not
+  bool isVertexInside(const mesh::Vertex &v) const;
+
+  /// Indicate that the mapped data is now considered as outdated by setting _hasMappedData to false
+  void setMappingFinished();
+
+  /// Number of input vertices this partition operates on
+  unsigned int getNumberOfInputVertices() const;
```
fsimonis, line 70: What about `size()` or `count()`?

src/mapping/Partition.hpp
```diff
@@ -0,0 +1,315 @@
...
+  PRECICE_DEBUG("Partition radius: {}", _radius);
+
+  // Get vertices to be mapped
+  auto outIDs = outputMesh->index().getVerticesInsideBox(center, radius);
+  // Constructing the partition when we don't have evaluation points is pointless
+  auto inIDs = inputMesh->index().getVerticesInsideBox(center, radius);
+
+  // If we have only very few vertices (10% of the target), we consider this partition as empty
+  // TODO: The second empty criterion is potentially not suitable for conservative mappings
+  _emptyPartition = outIDs.size() == 0 || inIDs.size() < 0.1 * targetSize;
```
fsimonis, line 143: To my understanding, the Partition doesn't is basically non-functional if `_emptyParition == true`.

Suggestion:

```cpp
std::optional<mapping::Partition> makePartition(...) {
  //...
  if (emptyParition) {
    return {};  
  }
  return Partition{...};
};
```

Then you only keep the Partitions that do anything in the PoU mapping.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -243,10 +264,15 @@ void MappingConfiguration::xmlTagCallback(
...
+    if (tag.hasAttribute(ATTR_VERTICES_PER_PARTITION)) {
+      verticesPerPartition = tag.getDoubleAttributeValue(ATTR_VERTICES_PER_PARTITION);
+      relativeOverlap      = tag.getDoubleAttributeValue(ATTR_RELATIVE_OVERLAP);
+      projectToInput       = tag.getBooleanAttributeValue(ATTR_PROJECT_TO_INPUT);
+    }
 
     RBFParameter rbfParameter;
     // Check valid combinations for the Gaussian RBF input
-    if (type == VALUE_RBF_GAUSSIAN || type == VALUE_RBF_INV_MULTIQUADRICS || type == VALUE_RBF_MULTIQUADRICS || type == VALUE_RBF_CTPS_C2 || type == VALUE_RBF_CPOLYNOMIAL_C0 || type == VALUE_RBF_CPOLYNOMIAL_C6 || type == VALUE_RBF_CPOLYNOMIAL_C2 || type == VALUE_RBF_CPOLYNOMIAL_C4) {
+    if (type == VALUE_RBF_GAUSSIAN || type == VALUE_RBF_INV_MULTIQUADRICS || type == VALUE_RBF_MULTIQUADRICS || type == VALUE_RBF_CTPS_C2 || type == VALUE_RBF_CPOLYNOMIAL_C0 || type == VALUE_RBF_CPOLYNOMIAL_C6 || type == VALUE_RBF_CPOLYNOMIAL_C2 || type == VALUE_RBF_CPOLYNOMIAL_C4 || type == VALUE_RBF_POU) {
```
fsimonis, line 275: The partition of unity is a method, not a kind of RBF.

I understand that this is a quick-n-drity way of getting the feature into preCICE, but isn't this a bit too cumbersome?
Given that you hard-coded the actual RBF type, I would move the partition of unity into a completely separate tag. This will be a good starting point for the future refactoring #1417.

If you do, please consider moving the pou to a separate `createMapping` too.

src/mesh/BoundingBox.cpp
```diff
@@ -91,6 +91,29 @@ Eigen::VectorXd BoundingBox::maxCorner() const
   return max;
 }
 
+double BoundingBox::getEdgeLength(int direction) const
```
fsimonis, line 94: Same as https://github.com/precice/precice/pull/1483/files#r1032278299

src/mesh/BoundingBox.cpp
```diff
@@ -91,6 +91,29 @@ Eigen::VectorXd BoundingBox::maxCorner() const
...
+{
+  PRECICE_ASSERT(direction < _dimensions);
+  PRECICE_ASSERT(_bounds.size() > (2 * direction + 1));
+  return std::abs(_bounds[2 * direction + 1] - _bounds[2 * direction]);
+}
+
+double BoundingBox::getMaximumEdgeLength() const
```
fsimonis, line 101: Suggestion

```suggestion
double BoundingBox::longestEdgeLength() const
```

src/mesh/BoundingBox.hpp
```diff
@@ -80,6 +80,15 @@ class BoundingBox {
...
   Eigen::VectorXd maxCorner() const;
 
+  /// returns the edge length in a specific direction
+  double getEdgeLength(int direction) const;
+
+  /// returns the maximum length of the bounding box in a particular direction
+  double getMaximumEdgeLength() const;
+
+  /// returns the coordinates of the bounding box in a specific direction
+  std::pair<double, double> getDirectionsCoordinates(int direction) const;
```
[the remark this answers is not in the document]
fsimonis: Sure, given that it is an axis-aligned bounding-box.

I suggest making this easier to use though.

```cpp
struct MinMax { double min, double max; };

MinMax operator[](int axis) const;

// allows to write bb[1].min bb[1].max instead of the terrible first and second.

// Alternative

MinMax axis(int axis) const;

// allows to write bb.axis(1).min bb(1).max
```

src/query/Index.hpp
fsimonis: > Seems like the documentation in line 72 is wrong?

Yep

> Why Eigen::VectorXd vs mesh::Vertex?

@davidscn True. We should keep the interface consistent. It's also way easier to test.

src/mapping/Partition.hpp
```diff
@@ -0,0 +1,315 @@
...
+  void mapConsistent(mesh::PtrData inputData, mesh::PtrData outputData);
+
+  /// Execute a consistent mapping
+  void mapConservative(mesh::PtrData inputData, mesh::PtrData outputData);
+
+  /// @brief set the normalized weight in the normalizedWeight data structure
+  void setNormalizedWeight(double normalizedWeight, VertexID vertexID);
+
+  /// Compute the weight for a given vertex
+  double computeWeight(const mesh::Vertex &v) const;
```
fsimonis, line 58: This only requires coordinates

```suggestion
  double computeWeight(const Eigen::VertorXD &v) const;
```

src/mapping/Partition.hpp
```diff
@@ -0,0 +1,315 @@
...
+
+  /// Evaluate the interpolant for a given vertex
+  double getInterpolatedValue(const VertexID id, unsigned int direction) const;
+
+  /// Indicate whether a vertex lies within the partition or not
+  bool isVertexInside(const mesh::Vertex &v) const;
```
fsimonis, line 64: Also only requires coordinates.

src/mapping/Partition.hpp
```diff
@@ -0,0 +1,315 @@
...
+  /// The weighting function
+  CompactPolynomialC2 _weightingFunction;
+};
+
+// --------------------------------------------------- HEADER IMPLEMENTATIONS
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+Partition<RADIAL_BASIS_FUNCTION_T>::Partition(
+    int               dimension,
+    mesh::Vertex      center,
```
fsimonis, line 120: What if the center is tagged/untagged before the construction?
Aren't there alternatives to using copies of `mesh::Vertex` inside the Partition?

src/mapping/Partition.hpp
davidscn: ..was a an obsolete leftover..

src/mapping/Partition.hpp
davidscn: I completely removed the function as it was mostly for debugging.

src/mapping/Partition.hpp
davidscn: In principle we do this already right now.

```cpp
Partition p(...)
    if (!p.empty()) {
      _clusters.emplace_back(std::move(p));
    }
```
With the `Partition` constructor bailing out in case the partition is empty. Your suggestion would require a free function, right? The problem is, the `empty()` information is generated in the `Partition` constructor and using `std::optional` would still require a similar logic as above, where we first check if the object was constructed and afterwards move the valid object or skip it, correct?

src/mapping/Partition.hpp
davidscn: Function was obsolete.

src/mapping/Partition.hpp
davidscn: Tried to improve

src/mapping/PartitionOfUnityMapping.hpp
davidscn: I was able to get rid of this data structure

src/mapping/PartitionOfUnityMapping.hpp
davidscn: It's only a local variable used for the mapping construction. I don't see any conflicts with multiple instantiations. Am I missing something?

src/mapping/PartitionOfUnityMapping.hpp
davidscn: Outdated

src/mapping/CreatePartitioning.hpp
```diff
@@ -0,0 +1,464 @@
+#pragma once
+
+#include <Eigen/Core>
+
+#include <algorithm>
+#include <boost/math/constants/constants.hpp>
```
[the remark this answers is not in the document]
davidscn: I removed it and traded the (unmeasured compile time) for run time. `sqrt(2)` is required only once or twice per pou mapping.

src/mapping/config/MappingConfiguration.cpp
davidscn: This is only an implementation to test the mapping itself. The MappingConfiguration will change once we converged in #1417 and the configuration implementation should be trivial then.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -362,6 +392,12 @@ MappingConfiguration::ConfiguredMapping MappingConfiguration::createMapping(
         new NearestNeighborGradientMapping(constraintValue, dimensions));
     configuredMapping.isRBF = false;
     return configuredMapping;
+
+  } else if (type == VALUE_RBF_POU) {
+    configuredMapping.mapping = PtrMapping(
+        new PartitionOfUnityMapping<CompactPolynomialC2>(constraintValue, dimensions, rbfParameter.value, polynomial, verticesPerPartition, relativeOverlap, projectToInput));
+    configuredMapping.isRBF = false;
```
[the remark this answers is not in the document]
davidscn: No, this was just the least intrusive way of getting an additional option. My plan is to first converge and implement #1417 and then add the option here.

src/mesh/BoundingBox.hpp
```diff
@@ -80,6 +80,15 @@ class BoundingBox {
   /// the max corner of the bounding box
   Eigen::VectorXd maxCorner() const;
 
+  /// returns the edge length in a specific direction
+  double getEdgeLength(int direction) const;
+
+  /// returns the maximum length of the bounding box in a particular direction
+  double getMaximumEdgeLength() const;
```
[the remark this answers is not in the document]
davidscn: I will change it, but won't fix too much things here before #1495 is ready

src/mesh/BoundingBox.hpp
```diff
@@ -80,6 +80,15 @@ class BoundingBox {
   /// the max corner of the bounding box
   Eigen::VectorXd maxCorner() const;
 
+  /// returns the edge length in a specific direction
+  double getEdgeLength(int direction) const;
```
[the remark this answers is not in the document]
davidscn: I will change it, but won't fix too much things here before https://github.com/precice/precice/pull/1495 is ready

src/mapping/PartitionOfUnityMapping.hpp
uekerman: We could export the mesh for example (don't you have an option for this already?)
Could also help for debugging in general

src/mapping/Partition.hpp
fsimonis: > Your suggestion would require a free function, right?

Correct.

> The problem is, the empty() information is generated in the Partition constructor and using std::optional would still require a similar logic as above, where we first check if the object was constructed and afterwards move the valid object or skip it, correct?

I see, but then, this information generation step could be moved to `makePartition` (or even a completely separate function) and the result passed to the constructor of the Partition.
Splitting this will make things easier to test too.

If this is not easily doable, then let's leave it as is.

src/mapping/PartitionOfUnityMapping.hpp
davidscn: I have such an option here, yes. Adding the input mesh might be insufficient, though, as the input mesh is 'changes' depending on the constraint.
Exporting the mesh is IMO mostly a developer debugging feature and adding the name is quite easy, so I just added it.

davidscn: Wait: the name for the export mesh is determined in a different place. Right now it is always `pouCenters.vtk`.

src/mapping/PartitionOfUnityMapping.hpp
davidscn: I don't really understand this comment. Why is it odd? We first have to clear the vector containing the clusters, because `compute_mapping` uses `emplace_back` in order to store the clusters in a vector. Calling `compute_mapping` here multiple times would result in erroneous behavior. In usual use-cases, this should never happen, but we use it a lot for the unit testing.

src/mapping/PartitionOfUnityMapping.hpp
davidscn: Changed it now to consider the input mesh name, might conflict for the classical `consistent`/`conservative` case though.

src/mapping/Partition.hpp
davidscn: I was able to get rid of the complete function.

src/mapping/Partition.hpp
davidscn: Tagging has no influence on the class construction.
Similar to other comments here: I mainly used a `Vertex`, because that's what the set of coordinates represent. I can use raw coordinates, i.e., `Eigen::Vector3d` or something similar, but it's less expressive in my opinion. What are your concerns regarding the vertex class here?

src/mapping/PartitionOfUnityMapping.hpp
uekerman: > In usual use-cases, this should never happen, but we use it a lot for the unit testing.

But we want to run into such a problem (if we ever have it), right? So `assertion(not hasComputedMapping())` would be helpful, no?

davidscn: Added the `clear` to the test

uekerman (changes requested):
I had a first more detailed look. So far, I focused on `PartitionOfUnityMapping`.
Looks much better already. It is still hard to follow, but it is also a complicated method.

Is there much code duplication in `PartitionOfUnityMappingTest.cpp` from `RadialBasisFctMappingTest.cpp` or `PetRadialBasisFctMappingTest.cpp`? If yes, maybe we could outsource some things into a common helper file.

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,390 @@
...
+class PartitionOfUnityMapping : public Mapping {
+public:
+  /**
+   * Constructor, which mostly sets the mesh connectivity requirements and initializes member variables.
+   *
+   * @param[in] constraint Specifies mapping to be consistent or conservative.
+   * @param[in] dimension Dimensionality of the meshes
+   * @param[in] parameter shape parameter or support radius of the RB function
+   * @param[in] verticesPerCluster Target number of vertices to be clustered together
+   * @param[in] relativeOverlap Overlap between clusters, where 1 would correspond to a complete overlap, 0 to distance of 2 x radius between clusters
```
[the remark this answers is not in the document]
davidscn: Added the missing ones and tried to clarify.

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,390 @@
...
+
+  /// toggles whether we project the cluster centers to the input mesh
+  const bool _projectToInput;
+
+  /// derived parameter based on the input above: the radius of each cluster
+  double averageClusterRadius = 0;
+
+  /// true if the mapping along some axis should be ignored
+  /// has currently only dim x false entries, as integrated polynomials are irrelevant
+  std::vector<bool> _deadAxis;
```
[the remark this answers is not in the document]
davidscn: I added them now as user input

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,390 @@
...
+
+    // Consider only non-empty clusters
+    if (!p.empty()) {
+      PRECICE_ASSERT(center.getID() == _clusters.size(), center.getID(), _clusters.size());
+      meshVertices.emplace_back(std::move(center));
+      _clusters.emplace_back(std::move(p));
+    }
+  }
+  // Log the average number of resulting clusters
+  PRECICE_INFO("Number of total clusters (final): {}", _clusters.size());
```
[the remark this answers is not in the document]
davidscn: > We are only talking about rank-local information here, right?

Yes. Mostly interesting for debugging, I guess. Should I convert it into a DEBUG statement?

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,390 @@
...
+    unsigned int maxVertices = std::max_element(_clusters.begin(), _clusters.end(), [](auto &v1, auto &v2) { return v1.getNumberOfInputVertices() < v2.getNumberOfInputVertices(); })->getNumberOfInputVertices();
+    unsigned int minVertices = std::min_element(_clusters.begin(), _clusters.end(), [](auto &v1, auto &v2) { return v1.getNumberOfInputVertices() < v2.getNumberOfInputVertices(); })->getNumberOfInputVertices();
+    PRECICE_INFO("Average number of vertices per cluster {}", nVertices / _clusters.size());
+    PRECICE_INFO("Maximum number of vertices per cluster {}", maxVertices);
+    PRECICE_INFO("Minimum number of vertices per cluster {}", minVertices);
+  }
+
+  // Log a bounding box of the center mesh
+  centerMesh.computeBoundingBox();
+  PRECICE_INFO("Bounding Box of the cluster centers {}", centerMesh.getBoundingBox());
```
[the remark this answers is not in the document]
davidscn: Debug is fine.

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,390 @@
...
+
+  // Step 3: index the clusters / the center mesh in order to define the output vertex -> cluster ownership
+  // the ownership is required to compute the normalized partition of unity weight (Step 4)
+  query::Index clusterIndex(centerMesh);
+  // Find all clusters the output vertex lies in, i.e., find all cluster centers which have the distance of a cluster radius from the given output vertex
+  // Here, we do this using the RTree on the clusterMesh: VertexID (queried from the centersMesh) == clusterID, by construction above.
+  // Note: this could also be done on-the-fly in the map data phase for dynamic queries, which would require to make the mesh as well as the indexTree member variables.
+  PRECICE_DEBUG("Computing cluster-vertex association");
+  for (const auto &vertex : outMesh->vertices()) {
+    auto clusterIDs = clusterIndex.getVerticesInsideBox(vertex, averageClusterRadius);
```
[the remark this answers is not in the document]
davidscn: No, it's not an average, I changed it.

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,390 @@
...
+template <typename RADIAL_BASIS_FUNCTION_T>
+void PartitionOfUnityMapping<RADIAL_BASIS_FUNCTION_T>::mapConservative(DataID inputDataID, DataID outputDataID)
+{
+  PRECICE_TRACE(inputDataID, outputDataID);
+
+  precice::utils::Event e("map.pou.mapData.From" + input()->getName() + "To" + output()->getName(), precice::syncMode);
+
+  // Execute the actual mapping evaluation in all clusters
+  // 1. Reset all output data values as we accumulate data in all clusters independent
+  output()->data(outputDataID)->values().setZero();
```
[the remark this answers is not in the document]
davidscn: Could be. For the function here, the setZero is absolutely necessary in order to ensure that it works properly. Do you want to rely on the 'somewhere' reset and transform the above into an assert?
Would require a corresponding reset in the tests, though.

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,390 @@
...
+    return; // Ranks not at the interface should never hold interface vertices
+
+  // TODO: Check again the tagging in combination with the partition construction (which mesh to use)
+  // In order to construct the local partitions, we need all vertices with a distance of 2 x radius,
+  // as the relevant partitions centers have a maximum distance of radius, and the proper construction of the
+  // interpolant requires all vertices with a distance of radius from the center.
+  auto bb = outMesh->getBoundingBox();
+
+  if (averageClusterRadius == 0)
+    averageClusterRadius = impl::estimateClusterRadius(_verticesPerCluster, filterMesh, bb);
```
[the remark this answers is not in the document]
davidscn: Zero is the default state and the state after calling reset. No floating point check required IMO.

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,390 @@
...
+  for (const auto &vertex : outMesh->vertices()) {
+    auto clusterIDs = clusterIndex.getVerticesInsideBox(vertex, averageClusterRadius);
+    // Consider the case where we didn't find any partition (meshes don't match very well)
+    if (clusterIDs.size() == 0) {
+      PRECICE_WARN("Output vertex {} could not be assigned to a cluster. This means that the meshes probably do not match well geometry-wise.", vertex.getCoords());
+      // TODO: Think about a proper way to handle this case, maybe set all radii to distance(v, closestvertex)?
+      clusterIDs.emplace_back(clusterIndex.getClosestVertex(vertex.getCoords()).index);
+    }
+
+    // Step 4: compute the normalized weights of each output vertex for each partition
```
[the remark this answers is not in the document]
davidscn: I used now substeps in the loop and put the step 4 in front of the loop.

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,390 @@
...
+  query::Index clusterIndex(centerMesh);
+  // Find all clusters the output vertex lies in, i.e., find all cluster centers which have the distance of a cluster radius from the given output vertex
+  // Here, we do this using the RTree on the clusterMesh: VertexID (queried from the centersMesh) == clusterID, by construction above.
+  // Note: this could also be done on-the-fly in the map data phase for dynamic queries, which would require to make the mesh as well as the indexTree member variables.
+  PRECICE_DEBUG("Computing cluster-vertex association");
+  for (const auto &vertex : outMesh->vertices()) {
+    auto clusterIDs = clusterIndex.getVerticesInsideBox(vertex, averageClusterRadius);
+    // Consider the case where we didn't find any partition (meshes don't match very well)
+    if (clusterIDs.size() == 0) {
+      PRECICE_WARN("Output vertex {} could not be assigned to a cluster. This means that the meshes probably do not match well geometry-wise.", vertex.getCoords());
```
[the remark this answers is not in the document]
davidscn: Right now, the vertex is then assigned to the 'closest' vertex cluster, i.e., the vertex will get data, but the data might be of a low accuracy (if any).

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,390 @@
...
+      // TODO: Think about a proper way to handle this case, maybe set all radii to distance(v, closestvertex)?
+      clusterIDs.emplace_back(clusterIndex.getClosestVertex(vertex.getCoords()).index);
```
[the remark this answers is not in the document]
davidscn: In my testcases, I usually don't run into this case, no.

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,390 @@
...
+  }
+
+  // Step 1: get a tentative clustering consisting of centers and a radius from one of the available algorithms
+  auto [averageClusterRadius_, centerCandidates] = impl::createClustering(inMesh, outMesh, _relativeOverlap, _verticesPerCluster, _projectToInput);
+
+  averageClusterRadius = averageClusterRadius_;
+  PRECICE_ASSERT(averageClusterRadius > 0 || inMesh->vertices().size() == 0 || outMesh->vertices().size() == 0);
+
+  // Step 2: check, which of the resulting clusters are non-empty and register the cluster centers in a mesh
+  // Here, the VertexCluster computes the matrix decompositions directly in case the cluster is non-empty
```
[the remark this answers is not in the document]
davidscn: We already check for empty-ness in the `createClustering`. We might, however, 'project' the centers afterwards and remove duplicates from the center vertices. In case we don't do the projection, we could in principle merge things from the cluster creation and the index queries here, but it makes the code more complicated. I would keep it this way.

davidscn:
> Looks much better already. It is still hard to follow, but it is also a complicated method.

I don't know how to further improve on this point.

> Is there much code duplication in PartitionOfUnityMappingTest.cpp from `RadialBasisFctMappingTest.cpp` or `PetRadialBasisFctMappingTest.cpp`? If yes, maybe we could outsource some things into a common helper file

I guess there is some overlap between RadialBasisFctMappingTest and the PETSc version, but I would not want to mix a test refactoring with the proposed changes here.

src/mesh/BoundingBox.hpp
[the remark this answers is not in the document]
davidscn: Removed the function

src/mapping/Partition.hpp
fsimonis: The point I am trying to make is that there is some state attached to the Vertex.
If this doesn't affect the construction, then it's fine.

src/mesh/BoundingBox.cpp
```diff
@@ -97,6 +97,18 @@ Eigen::VectorXd BoundingBox::maxCorner() const
...
+  PRECICE_ASSERT(direction < _dimensions);
+  PRECICE_ASSERT(_boundMax.size() > direction && _boundMin.size() > direction);
+  return std::abs(_boundMax(direction) - _boundMin(direction));
+}
+
+double BoundingBox::longestEdgeLength() const
+{
+  return (_boundMax - _boundMin).maxCoeff();
+}
+
```
uekerman, line 111: unit tests for these methods?

src/query/Index.cpp
```diff
@@ -183,6 +183,19 @@ VertexMatch Index::getClosestVertex(const Eigen::VectorXd &sourceCoord)
...
+  PRECICE_TRACE();
+  PRECICE_ASSERT(!(_mesh->vertices().empty()), _mesh->getName());
+  std::vector<VertexID> matches;
+  const auto &          rtree = _pimpl->getVertexRTree(*_mesh);
+
+  rtree->query(bgi::nearest(sourceCoord, n), boost::make_function_output_iterator([&](size_t matchID) {
+                 matches.emplace_back(matchID);
+               }));
+  return matches;
+}
```
uekerman, line 197: unit tests for this method?

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -0,0 +1,401 @@
...
+    for (unsigned int i = 0; i < localNumberOfClusters; ++i) {
+      PRECICE_ASSERT(clusterIDs[i] < static_cast<int>(_clusters.size()));
+      _clusters[clusterIDs[i]].setNormalizedWeight(weights[i] / weightSum, vertex.getID());
+    }
+  }
+
+// Add a VTK export for visualization purposes
+#ifndef NDEBUG
+  exportClusterCentersAsVTU(centerMesh);
+#endif
```
uekerman, line 256: Something to mention in the user docs

src/mapping/PartitionOfUnityMapping.hpp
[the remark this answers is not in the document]
uekerman: Yes, I think it would be better to assert here and check whether things are actually reset. This resetting should be independent of the actual mapping used.

uekerman (approved):
I had a look over everything again with a focus on `SphericalVertexCluster.hpp`. I did not step through all details of `CreateClustering.hpp`.

I think we are in a good state now. Still a few things to tweak, but all minor.

Nice documentation in `SphericalVertexCluster` :+1:

Does it make sense to also implement unit tests for `estimateClusterRadius` (and the other helper functions) ? 

After merging, please add a new section in the [mapping docs](https://precice.org/configuration-mapping.html) explaining the attributes with a picture.

src/mapping/impl/SphericalVertexCluster.hpp
```diff
@@ -0,0 +1,302 @@
...
+ * Hence, the PartitionOfUnity mapping class owns a vector of SphericalVertexClusters in order
+ * to compute the mapping.
+ * In its core, the class consists of a geometric center vertex and a radius representing the
+ * spherical shape of the partition. In order to compute an RBF interpolant, the class stores
+ * VertexIDs of the input mesh and the output mesh lying within the sphere and a
+ * RadialBasisFctSolver to assemble and solve mapping matrices. The solver class here is exactly
+ * the same class used in the plain RBF mapping.
+ */
+template <typename RADIAL_BASIS_FUNCTION_T>
+class SphericalVertexCluster {
```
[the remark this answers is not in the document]
davidscn: The `Mapping' class is an interface for our data mapping methods. Reading the very first sentence should make clear that a `Cluster` itself is a subpart of the `PartitionOfUnityMapping`, no?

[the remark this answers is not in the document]
davidscn: Ah ok, then I will comment on that..

src/mapping/impl/SphericalVertexCluster.hpp
```diff
@@ -0,0 +1,302 @@
...
+  auto inIDs = inputMesh->index().getVerticesInsideBox(center, radius);
+
+  // Transform the vector to the appropriate boost data structure
+  // The IDs are sorted in the boost flat_set, hence, the function here has N log(N) complexity
+  _inputIDs.insert(inIDs.begin(), inIDs.end());
+  _outputIDs.insert(outIDs.begin(), outIDs.end());
+
+  // If the cluster is empty, we return immediately
+  if (empty()) {
+    return;
```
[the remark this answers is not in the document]
davidscn: > that input is non-empty and output is empty

This case is legal and the partition won't be required, i.e., we throw it away at this point.

> and the other way around

this case is illegal and will trigger an error when trying to find a cluster for the output vertex (unless another/neighbor partition was found).

src/mapping/impl/SphericalVertexCluster.hpp
```diff
@@ -0,0 +1,302 @@
...
+  PRECICE_DEBUG("SphericalVertexCluster output size: {}", outIDs.size());
+
+  // TODO
+  // Otherwise the system is underdetermined
+  // if (inIDs.size() < dimension + 1)
+  //   _polynomial = Polynomial::OFF;
+
+  // Construct the solver
+  _rbfSolver          = RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>{function, *inputMesh.get(), _inputIDs, *outputMesh.get(), _outputIDs, deadAxis, _polynomial};
+  _hasComputedMapping = true;
```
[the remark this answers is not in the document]
davidscn: The constructor of the RBFSolver already does all the work-heavy part, e.g., LLT decomposition etc.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -195,10 +205,18 @@ MappingConfiguration::MappingConfiguration(
                                .setDocumentation("Sets kind of preallocation for PETSc RBF implementation")
                                .setOptions({PREALLOCATION_ESTIMATE, PREALLOCATION_COMPUTE, PREALLOCATION_OFF, PREALLOCATION_SAVE, PREALLOCATION_TREE});
 
+  auto verticesPerCluster = XMLAttribute<int>(ATTR_VERTICES_PER_CLUSTER)
+                                .setDocumentation("Vertices per cluster (partition) applied in the rbf partition of unity method.");
+  auto relativeOverlap = makeXMLAttribute(ATTR_RELATIVE_OVERLAP, 0.3)
+                             .setDocumentation("Value between 0 and 1 indicating the relative overlap between clusters. A value of 0.3 is usually a good trade-off between accuracy and efficiency.");
+  auto projectToInput = XMLAttribute<bool>(ATTR_PROJECT_TO_INPUT)
+                            .setDocumentation("If enabled, places the cluster centers at the closest vertex of the input mesh. Should be enabled in case of non-uniform point distributions such as for shell structures.");
```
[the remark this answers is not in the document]
davidscn: Hm, I guess in most of the cases it won't hurt. For the Halton points, e.g. it might hurt in combination with a low number of vertices per cluster. In many cases it is surprisingly beneficial..

davidscn:
@uekerman regarding all the defaults: I added now defaults for the pum specific parameters. I'm just afraid that these settings will be considered as a silver bullet and people don't think and or modify these parameters: Example: going higher with the number of vertices-per-cluster might give more accurate results, but it's also more expensive. The values set here are just reasonable default values.

src/mapping/config/MappingConfiguration.cpp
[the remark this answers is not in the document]
davidscn: See comment below

src/mapping/impl/SphericalVertexCluster.hpp
```diff
@@ -0,0 +1,302 @@
...
+    RADIAL_BASIS_FUNCTION_T function,
+    std::vector<bool>       deadAxis,
+    Polynomial              polynomial,
+    mesh::PtrMesh           inputMesh,
+    mesh::PtrMesh           outputMesh)
+    : _center(center), _radius(radius), _polynomial(polynomial), _weightingFunction(radius)
+{
+  PRECICE_TRACE();
+  // Disable integrated polynomial, as it might cause locally singular matrices
+  PRECICE_ASSERT(_polynomial != Polynomial::ON, "Integrated polynomial is not supported for partition of unity data mappings.")
```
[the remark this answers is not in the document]
davidscn: I removed it. Seems a bit inconvenient to me to offer different options for different methods, but ok.

src/mapping/Partition.hpp
davidscn: I see your point, but would rather delay adjustments in this direction. Here, I need to preserve compatibility towards `vertex.rawCoords`. Maybe something to adjust later on.

src/mapping/Partition.hpp
davidscn: I still need to distinguish input and output and cannot really conform with the stdlib.

src/mapping/Partition.hpp
davidscn: I see your point, but would rather delay adjustments in this direction. Here, I need to preserve compatibility towards `vertex.rawCoords`. Maybe something to adjust later on.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -279,6 +302,11 @@ void MappingConfiguration::xmlTagCallback(
     std::string strPolynomial = tag.getStringAttributeValue(ATTR_POLYNOMIAL, POLYNOMIAL_SEPARATE);
     std::string strPrealloc   = tag.getStringAttributeValue(ATTR_PREALLOCATION, PREALLOCATION_TREE);
 
+    // pum related tags
+    int    verticesPerCluster = tag.getIntAttributeValue(ATTR_VERTICES_PER_CLUSTER, 0);
+    double relativeOverlap    = tag.getDoubleAttributeValue(ATTR_RELATIVE_OVERLAP, 0.3);
+    bool   projectToInput     = tag.getBooleanAttributeValue(ATTR_PROJECT_TO_INPUT, false);
```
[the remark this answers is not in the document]
uekerman: :+1: having this information in a comment could be good

PR #1485, opened 2022-11-23
Selective synchonization

### LINKED ISSUE #1293: Introduce tag for use of adaptive meshes
Introducing re-meshing leads to additional synchronisation points and re-paritioning work.

To reduce this cost as much as possible, I suggest requiring adaptive meshes be tagged as such.

```xml
<mesh name="Mesh">
  ...
</mesh>

<participant name="Sender">
  <use-mesh name="Mesh" provide="true" adaptive="true" />
</participant>

<participant name="Receiver">
  <use-mesh name="Mesh" from="Sender" />
</participant>
```

This allows us to:
* disable all re-meshing features by default.
  Existing configurations will work out-of-the-box without any performance impact.
* enforce that only Sender may use `resetMesh(meshID)` to reset and redefine the mesh.
  This restricts which meshes need repartitioning on which participants.
* enable the intra-participant handshake only for Sender, as it provides an adaptive mesh.
  The Receiver does not need this handshake and can skip it as it will never be able to call `resetMesh` itself.
* enable re-initialization only between Sender and Receiver (everyone using Mesh from Sender)
  Other participants will never re-initialize as they don't come in contact with participants providing adaptive meshes.

fsimonis:
## Main changes of this PR

This PR aims to add selective synchronization of participants.

With the synchronization in-place, we need to reduce the performance impact by selectively enabling synchronization for participants providing adaptive meshes as well as participants receiving dynamic meshes.

Based on #1462 and #1450
Supersedes #1451
Closes #1293 

## Author's checklist

## Reviewers' checklist


davidscn:
Closed in favor of #1589

PR #1486, opened 2022-11-24
Fix include path for eigen unsupported

davidscn:
## Main changes of this PR
If Eigen was installed via spack and not systemwide, the `eigen3` directory is not in the search path.

## Author's checklist

## Reviewers' checklist


BenjaminRodenberg:
Nice find :+1: This should fix a problem that @IshaanDesai discovered in the spack build for the python bindings a few days ago.

fsimonis (approved):

Issue #1365, opened 2022-07-08
precicef_mark_action_fulfilled(writeInitialData) is missing in examples/solverdummies/fortran/solverdummy.f90?


gdenayer:
**Describe your setup**

Operating system (e.g. Linux distribution and version): CentOS 7
preCICE Version: 2.4.0

**Describe the problem**
In order to couple my fluid solver with preCICE, I'm using examples/solverdummies/fortran/solverdummy.f90 as second solver.
I add initialize="yes" to my exchange line in my xml file and I'm getting:

 DUMMY: Writing initial data
---[precice] ERROR:  Initial data has to be written to preCICE by calling an appropriate write...Data() function before calling initializeData(). Did you forget to call markActionFulfilled(precice::constants::actionWriteInitialData()) after writing initial data?

It seems that  precicef_mark_action_fulfilled(writeInitialData) is missing.

**Step To Reproduce**
1.  add initialize="yes" to the xml file
2. then compile and start solverdummy.f90

**Expected behaviour**
No Error :)

**Additional context**
Modifying examples/solverdummies/fortran/solverdummy.f90 as follows solves this issue:
  CALL precicef_is_action_required(writeInitialData, bool)
  IF (bool.EQ.1) THEN
    WRITE (*,*) 'DUMMY: Writing initial data'
    CALL precicef_mark_action_fulfilled(writeInitialData)
  ENDIF
  CALL precicef_initialize_data()

Please note that the different solverdummies seem not to be equivalent.  The section "writeInitialData" is present in solverdummy.f90 but not in solverdummy.c and in  solverdummy.cpp

Regards
Guillaume

BenjaminRodenberg:
Hi @gdenayer,

thanks for pointing us to this issue! The fix that you prepared looks correct. I think that originally data initialization was not part of the solverdummies, but generally it does not harm to add this. I also realized that the implementation of the fortran solverdummy definitely has a problem: https://github.com/precice/precice/blob/c9095ee9f8feceb938e3f9dd8ad6fd55e56a6609/examples/solverdummies/fortran/solverdummy.f90#L68-L71

Actually no data is written.

[...]

PR #1487, opened 2022-11-24
Replace actions with explicit calls

### LINKED ISSUE #719: replace isActionRequired(action) with explicit calls
The preCICE Interface offers a set of functions and actions to allow the user to query whether a checkpoint or initial data has to be written/read and report, if the actions have been completed. I would like to propose an alternative interface, which is in my opinion easier to use.

## Current state

The following functions belong to the current interface:

https://github.com/precice/precice/blob/8bd11debde201d3a0ce78327808570ddb24127ce/src/precice/SolverInterface.hpp#L302-L317

https://github.com/precice/precice/blob/8bd11debde201d3a0ce78327808570ddb24127ce/src/precice/SolverInterface.hpp#L319-L329

https://github.com/precice/precice/blob/8bd11debde201d3a0ce78327808570ddb24127ce/src/precice/SolverInterface.hpp#L834-L841

## Proposed interface

I would prefer a more explicit interface. For example:

* `bool writingInitialDataIsRequired()`
* `void initialDataHasBeenWritten()`
* `bool readingCheckpointIsRequired()`
* `void checkpointHasBeenRead()`
* `bool writingCheckpointIsRequired()`
* `void checkpointHasBeenWritten()`

### Testing

The main reason, why I would prefer this implementation is that it makes testing much more easier. Currently I always have to define a return value, if isActionRequired receives a certain action:

def is_action_required_behavior(py_action):
    if py_action == action_read_iteration_checkpoint():
        return False
    elif py_action == action_write_iteration_checkpoint():
        return True

Interface.is_action_required = MagicMock(side_effect=is_action_required_behavior)

Providing a different Interface allows me to write this differently:

Interface.reading_checkpoint_is_required = MagicMock(return_value=False)
Interface.writing_checkpoint_is_required = MagicMock(return_value=True)

### More reasons?

Do you have more ideas where the proposed interface might be helpful?

## Are there also drawbacks?

First of all, this is an API breaking change. But are there also additional (functional) drawbacks?

Looking forward to your replies and ideas!

## Additional context

(conversation with @uekerman) 

Bernhard's thesis, page 115:

*Finally, we consider some auxiliary API methods shown in Algorithm 3.8. The methods in lines 1–4 can be used by a solver to get additional information about the couplingstate, e.g., to find out whether coupling data should be written or read. The method `isActionRequired` is used as an extensible means to convey required actions to a solver. E.g., coupling iterations for implicit equation coupling require a solver to store and load iteration checkpoints, i.e., old timestep states, while iterating. A solver has to query for the actions write-iteration-checkpoint and read-iteration-checkpoint, and confirm their execution by using the method fulfilledAction. Other coupling tools often have a dedicated method ”iterate” to signal a solver to perform a coupling iteration. However, when subcycling is included and implemented in the coupling library, this is not sufficient. Further actions are provided for synchronized checkpoint writing and reading, as well as regular data output writing. The mechanism is easy to extend by further required solver actions and a solver needs to implement only those actions which correspond to used functionality. In order to avoid errors from not properly adapted solvers, fulfilledAction acts as a checking mechanism. In case a solver does not fulfill a required action, preCICE can detect the missing call of fulfilledAction and report an error. All required action names are gathered in the namespace precice::constants.*

Historic note: In the past, there were more solver actions that we got rid of (e.g. read/write-simulation-checkpoint, writing output). Additionally, the hope *The mechanism is easy to extend by further required solver actions* did not pay off as so far we never did this. For example, we are currently also not considering this interface for the mesh requirement #467. 

The reason for the unpopularity of the solver action interface could be that it feels too heavy. I have seen multiple times that our users do not understand it properly and hack around it. Maybe also sometimes, we don't understand it properly. In particular, the `fulfilledAction` did not pay off so far, in my opinion.

Thus, we could think about how to improve the situation. In my opinion, the reason, however, should not (only) be testing, but usability. Just transforming the interface to explicit calls does not solve the bigger problem. If we touch this, we should think about a proper solution (incorporating what we have learned the last 5 years by watching users doing it wrong). To be honest, I don't yet see a good solution, other then just dropping `fulfilledAction` and finding different names.

fsimonis:
## Main changes of this PR

This PR
* replaces `isActionRequired` and `markActionFulfilled` with explicit calls `requiresInitialData()`, `requiresReadingCheckpoint()`, `requiresWritingCheckpoint()`
* removes `precice::constants`
* extends CouplingScheme to keep track of required as well as fulfilled actions. This allows to check if something was required after it was fulfilled.

Closes #719

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/1487.md
+++ b/docs/changelog/1487.md
@@ -1,3 +1,4 @@
 - Replaced `isActionRequired()` and `markActionFulfilled()` with explicit calls `requiresInitialData()`, `requiresReadingCheckpoint()`, and `requiresWritingCheckpoint()`. Invoking these functions is a promise to preCICE that the caller fulfills the requirement.
 - Removed `precice::constants` from the C++ API and the C and Fortran bindings.
 - Added `precicef_get_version_information_` to Fortran bindings.
+- Renamed `isMeshConnectivityRequired()` to `requiresMeshConnectivityFor()`
--- a/docs/changelog/1487.md
+++ b/docs/changelog/1487.md
@@ -2,3 +2,4 @@
 - Removed `precice::constants` from the C++ API and the C and Fortran bindings.
 - Added `precicef_get_version_information_` to Fortran bindings.
 - Renamed `isMeshConnectivityRequired()` to `requiresMeshConnectivityFor()`
+- Renamed `isGradientDataRequired()` to `requiresGradientDataFor()`
--- a/docs/changelog/1487.md
+++ b/docs/changelog/1487.md
@@ -1,0 +1,3 @@
+- Replaced `isActionRequired()` and `markActionFulfilled()` with explicit calls `requiresInitialData()`, `requiresReadingCheckpoint()`, and `requiresWritingCheckpoint()`. Invoking these functions is a promise to preCICE that the caller fulfills the requirement.
+- Removed `precice::constants` from the C++ API and the C and Fortran bindings.
+- Added `precicef_get_version_information_` to Fortran bindings.
--- a/extras/bindings/c/include/precice/SolverInterfaceC.h
+++ b/extras/bindings/c/include/precice/SolverInterfaceC.h
@@ -449,8 +449,8 @@ PRECICE_API const char *precicec_getVersionInformation();
  */
 ///@{
 
-/// @copydoc precice::SolverInterface::isGradientDataRequired
-PRECICE_API int precicec_isGradientDataRequired(int dataID);
+/// @copydoc precice::SolverInterface::requiresGradientDataFor
+PRECICE_API int precicec_requiresGradientDataFor(int dataID);
 
 /// @copydoc precice::SolverInterface::writeScalarGradientData
 PRECICE_API void precicec_writeScalarGradientData(
--- a/extras/bindings/c/include/precice/SolverInterfaceC.h
+++ b/extras/bindings/c/include/precice/SolverInterfaceC.h
@@ -129,8 +129,8 @@ PRECICE_API int precicec_hasMesh(const char *meshName);
  */
 PRECICE_API int precicec_getMeshID(const char *meshName);
 
-/// @copydoc precice::SolverInterface::isMeshConnectivityRequired()
-PRECICE_API int precicec_isMeshConnectivityRequired(int meshID);
+/// @copydoc precice::SolverInterface::requiresMeshConnectivityFor()
+PRECICE_API int precicec_requiresMeshConnectivityFor(int meshID);
 
 /**
  * @brief Creates a mesh vertex
--- a/extras/bindings/c/include/precice/SolverInterfaceC.h
+++ b/extras/bindings/c/include/precice/SolverInterfaceC.h
@@ -103,22 +103,14 @@ PRECICE_API int precicec_isTimeWindowComplete();
 ///@name Action Methods
 ///@{
 
-/**
- * @brief Checks if the provided action is required.
- * @param[in] action the name of the action
- * @returns whether the action is required
- */
-PRECICE_API int precicec_isActionRequired(const char *action);
+/// @copydoc precice::SolverInterface::requiresInitialData()
+PRECICE_API int precicec_requiresInitialData();
 
-/**
- * @brief Indicates preCICE that a required action has been fulfilled by a solver.
- * @pre The solver fulfilled the specified action.
- *
- * @param[in] action the name of the action
- */
-PRECICE_API void precicec_markActionFulfilled(const char *action);
+/// @copydoc precice::SolverInterface::requiresWritingCheckpoint()
+PRECICE_API int precicec_requiresWritingCheckpoint();
 
-///@}
+/// @copydoc precice::SolverInterface::requiresReadingCheckpoint()
+PRECICE_API int precicec_requiresReadingCheckpoint();
 
 ///@name Mesh Access
 ///@anchor precice-mesh-access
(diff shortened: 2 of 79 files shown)


fsimonis:
Solves #1365

src/cplscheme/CompositionalCouplingScheme.cpp
```diff
@@ -251,13 +251,29 @@ bool CompositionalCouplingScheme::isActionRequired(
...
+{
+  PRECICE_TRACE(actionName);
+  bool isFulfilled = false;
+  for (const Scheme &scheme : _couplingSchemes) {
+    if (not scheme.onHold) {
+      isFulfilled |= scheme.scheme->isActionRequired(actionName);
+    }
+  }
+  PRECICE_DEBUG("return {}", isFulfilled);
+  return isFulfilled;
```
uekerman, line 265: This looks like there is a "not" missing.
Or did you want to call `isActionFulfilled`?

src/cplscheme/CompositionalCouplingScheme.hpp
```diff
@@ -195,6 +195,13 @@ class CompositionalCouplingScheme final : public CouplingScheme {
    */
   bool isActionRequired(const std::string &actionName) const final override;
 
+  /**
+   * @brief Returns true, if the given action has been performed by the accessor.
+   *
+   * True, if any of the composed coupling schemes requires the action.
```
uekerman, line 201:
```suggestion
   * False, if any of the composed coupling schemes requires the action.
```
?

extras/bindings/c/include/precice/SolverInterfaceC.h
```diff
@@ -102,22 +102,38 @@ int precicec_isTimeWindowComplete();
...
-///@}
+/** Checks if the participant is required to read an iteration checkpoint.
+ *
+ * If true, the participant is required to read an iteration checkpoint before
+ * calling advance().
+ *
+ * preCICE refuses to proceed if reading a checkpoint is required,
+ * but this method isn't called prior to advance().
+ *
+ * @pre advance() has been called
```
uekerman, line 134: Same confusion problem as in `SolverInterface`

extras/bindings/fortran/include/precice/SolverInterfaceFortran.hpp
```diff
@@ -634,36 +623,15 @@ void precicef_read_sdata_(
...
-
 /**
  * @brief Name of action for reading iteration checkpoint.
  *
  * Fortran syntax:
- * precicef_action_read_iter_checkpoint( CHARACTER nameAction(*) )
+ * precicef_get_version_information_(
+ *   CHARACTER versionInfo(*) )
+ *
+ * @copydoc precice::SolverInterface::getVersionInformation
```
uekerman, line 633: Sth for the changelog?

uekerman (changes requested):
Looks good. I checked the implementation, but did not go through all integration tests (changes there should be trivial).

In #719, we still have the debate whether we should also update in favor of `requiresMeshConnetivityFor(Mesh)` and `requiresGradientDataFor(Data)`? I would say yes. If the PR now closes the issue, this discussion would be lost. Let's add here or open another issue.

What are pro and con arguments for keeping the "isRequired..." and "markFullfilled..." on the `CouplingScheme` level?

docs/changelog/1487.md
```diff
@@ -0,0 +1,2 @@
+- Replaced `isActionRequired()` and `markActionFulfilled()` with explicit calls `requiresInitialData()`, `requiresReadingCheckpoint()`, and `requiresWritingCheckpoint()`. Invoking these functions is a promise to preCICE that the caller fulfills the requirement.
+- Removed `precice::constants` in all bindings
```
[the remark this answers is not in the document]
fsimonis:
```suggestion
- Removed `precice::constants` from the C++ API, C and Fortran bindings.
```

src/cplscheme/CompositionalCouplingScheme.hpp
fsimonis: Need a rewrite

fsimonis:
> In https://github.com/precice/precice/issues/719, we still have the debate whether we should also update in favor of requiresMeshConnetivityFor(Mesh) and requiresGradientDataFor(Data)? I would say yes. If the PR now closes the issue, this discussion would be lost. Let's add here or open another issue.

:+1: On it.

> What are pro and con arguments for keeping the "isRequired..." and "markFullfilled..." on the CouplingScheme level?

It keeps the scope of the PR small. We should definitely simplify this though as we don't need string representations any more. The isRequired, markFullfilled, isFulfilled interface allows requesting this on a more fine level. Maybe we can reduce this in a future PR.

fsimonis:
@uekerman Rebased on develop and reworded quite a bit.

I also renamed isMeshConnectivtyRequired and isGradientDataRequired.

docs/changelog/1487.md
```diff
@@ -0,0 +1,5 @@
+- Replaced `isActionRequired()` and `markActionFulfilled()` with explicit calls `requiresInitialData()`, `requiresReadingCheckpoint()`, and `requiresWritingCheckpoint()`. Invoking these functions is a promise to preCICE that the caller fulfills the requirement.
+- Removed `precice::constants` from the C++ API, C and Fortran bindings.
```
uekerman, line 2:
```suggestion
- Removed `precice::constants` from the C++ API and the C and Fortran bindings.
```

src/cplscheme/CompositionalCouplingScheme.cpp
```diff
@@ -251,13 +251,29 @@ bool CompositionalCouplingScheme::isActionRequired(
...
 }
 
+bool CompositionalCouplingScheme::isActionFulfilled(
+    const std::string &actionName) const
+{
+  PRECICE_TRACE(actionName);
+  bool isFulfilled = false;
+  for (const Scheme &scheme : _couplingSchemes) {
+    if (not scheme.onHold) {
+      isFulfilled |= scheme.scheme->isActionFulfilled(actionName);
```
uekerman, line 261:
```suggestion
  bool isFulfilled = true;
  for (const Scheme &scheme : _couplingSchemes) {
    if (not scheme.onHold) {
      isFulfilled &= scheme.scheme->isActionFulfilled(actionName);
```
?
It's a bit worrisome that we don't catch this with a test

uekerman (approved):

src/cplscheme/CompositionalCouplingScheme.cpp
fsimonis: This is actually correct.

Not all schemes require the action, so not all will have the action fulfilled.

davidscn:
This is still missing in the porting guide

fsimonis:
Added

PR #1152, opened 2021-12-13
Adds initial QN updates for improvements

KyleDavisSA:
## Main changes of this PR

1. Automatically deletes the first column in time step 4 if an initial sub-vector from at least 1 solver is zero, measured by the L2 norm.
2. Monitors the residual sum preconditioner weights and only updates them if they change by a specified factor. This is required for faster QR filtering.

## Motivation and additional information
Firstly, deleting the first column if there is a zero sub-vector has been shown to be beneficial. 

Secondly, currently the preconditioner weights are updated in every iteration. This is unnecessary if the weights change by a small value, but the entire QR decomposition is performed again. Vast improvements in runtime can be found by not always updating the preconditioner weights. 

## Author's checklist

* [ ] I added a changelog file with this PR number in `docs/changelog/` if there are noteworthy changes.
* [x] I ran `tools/formatting/format-all` to ensure everything is formatted correctly.
* [ ] I sticked to C++14 features.
* [ ] I sticked to CMake version 3.10.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

A few more tasks include 
* [ ] Allow preconditioner weight monitoring to be turned on/off via the config file.
* [ ] Allow column deletion to be turned on/off via the config file.
* [x] Include the fast QR filter when preconditioner weight monitoring is applied.

PR #1493, opened 2022-11-29
Pre-scaling weight monitoring with new QR filter

KyleDavisSA:
## Main changes of this PR
This PR closes #1152 due to being too old. The reason for this PR are copied from #1152.

1. Automatically deletes the first column in time step 4 if an initial sub-vector from at least 1 solver is zero, measured by the L2 norm.
2. Monitors the residual sum preconditioner weights and only updates them if they change by a specified factor. This is required for faster QR filtering.

Currently the preconditioner weights are updated in every iteration. This is unnecessary if the weights change by a small value, but the entire QR decomposition is performed again. Vast improvements in runtime can be found by not always updating the preconditioner weights.

## Author's checklist

## Reviewers' checklist

* [x] Allow preconditioner weight monitoring to be turned on/off via the config file.
* [ ] Allow column deletion to be turned on/off via the config file.
* [ ] Include the fast QR filter when preconditioner weight monitoring is applied.
* [x] Include tests for QR filters
* [ ] Update user documentation


src/acceleration/impl/Preconditioner.hpp
```diff
@@ -193,6 +193,12 @@ class Preconditioner {
     return _requireNewQR;
   }
 
+  /// returns true if pre-scaling weights were updated in the current iteration
+  bool areWeightsUpdated()
+  {
+    return _areWeightsUpdated;
+  }
```
uekerman, line 200: areUpdated or wereUpdated?

src/acceleration/impl/QRFactorization.cpp
```diff
@@ -140,24 +140,54 @@ void QRFactorization::applyFilter(double singularityLimit, std::vector<int> &del
...
-      if (!inserted) {
-        delIndices.push_back(k);
+
+    resetFilter(singularityLimit, delIndices, V);
+
+  } else if (_filter == Acceleration::QR3FILTER) {
+    int index = cols() - 1;
+    // Iterate from the last column to the 2nd column from the left
+    if (computeQR2Filter == true) {
+      PRECICE_DEBUG("  Pre-scaling weights were reset. Reverting to QR2 and rebuilding QR.");
```
uekerman, line 150:
```suggestion
      PRECICE_DEBUG("Pre-scaling weights were reset. Reverting to QR2 and rebuilding QR.");
```

src/acceleration/impl/ResidualSumPreconditioner.cpp
```diff
@@ -67,19 +73,39 @@ void ResidualSumPreconditioner::_update_(bool                   timeWindowComple
     }
 
     offset = 0;
-    for (size_t k = 0; k < _subVectorSizes.size(); k++) {
-      if (not math::equals(_residualSum[k], 0.0)) {
-        for (size_t i = 0; i < _subVectorSizes[k]; i++) {
-          _weights[i + offset]    = 1 / _residualSum[k];
-          _invWeights[i + offset] = _residualSum[k];
+    // Check if the new scaling weights are more or less than 1 order of magnitude from the previous weights
```
uekerman, line 76: The "1" here is now configurable (`updatePreconLimit`), right? Also in the Debug statement below.

src/acceleration/impl/ResidualSumPreconditioner.cpp
```diff
@@ -67,19 +73,39 @@ void ResidualSumPreconditioner::_update_(bool                   timeWindowComple
...
+    if (!_preconForceUpdate) {
+      for (size_t k = 0; k < _subVectorSizes.size(); k++) {
+        double newScalingWeight = (1 / _residualSum[k]);
+        std::cout << "newScalingWeight: " << newScalingWeight << std::endl;
+        std::cout << "_previousResidualSum[k]: " << _previousResidualSum[k] << std::endl;
```
uekerman, line 81: No `std::cout` please.

src/acceleration/impl/ResidualSumPreconditioner.hpp
```diff
@@ -17,7 +17,7 @@ namespace impl {
  */
 class ResidualSumPreconditioner : public Preconditioner {
 public:
-  ResidualSumPreconditioner(int maxNonConstTimeWindows);
+  ResidualSumPreconditioner(int maxNonConstTimeWindows, bool = true, double updatePreconLimit = 10.0);
```
uekerman, line 20:
```suggestion
  ResidualSumPreconditioner(int maxNonConstTimeWindows, bool preconForceUpdate = true, double updatePreconLimit = 10.0);
```
??
Why do you need the defaults here? Aren't these covered in the configuration already?

tests/quasi-newton/serial/TestQN15.cpp
```diff
@@ -0,0 +1,24 @@
...
+#include "../helpers.hpp"
+
+BOOST_AUTO_TEST_SUITE(Integration)
+BOOST_AUTO_TEST_SUITE(QuasiNewton)
+BOOST_AUTO_TEST_SUITE(Serial)
+PRECICE_TEST_SETUP("SolverOne"_on(1_rank), "SolverTwo"_on(1_rank))
+BOOST_DATA_TEST_CASE(TestQN15, boost::unit_test::data::make({true, false}), includeSecondaryData)
+{
+  PRECICE_TEST();
+  // serial coupling, IQN-IMVJ acceleration, to test `RS-SVD` method for restart;
```
uekerman, line 16: This comment probably needs an update?

src/acceleration/config/AccelerationConfiguration.cpp
```diff
@@ -53,17 +53,22 @@ AccelerationConfiguration::AccelerationConfiguration(
...
       VALUE_QR1_ABSFILTER("QR1-absolute"),
       VALUE_QR2FILTER("QR2"),
+      VALUE_QR3FILTER("QR3"),
       VALUE_CONSTANT_PRECONDITIONER("constant"),
       VALUE_VALUE_PRECONDITIONER("value"),
       VALUE_RESIDUAL_PRECONDITIONER("residual"),
       VALUE_RESIDUAL_SUM_PRECONDITIONER("residual-sum"),
+      VALUE_RESIDUAL_SUM_ALWAYS("always"),
+      VALUE_RESIDUAL_SUM_THRESHOLD("on-threshold"),
+      VALUE_RESIDUAL_SUM_LIMIT("limit"),
```
uekerman, line 71: If we keep this limit configurable, we need to make both names consistent.
Either "on-limit" and "limit". Or "on-threshold" and "threshold".

src/acceleration/config/AccelerationConfiguration.cpp
```diff
@@ -492,11 +512,23 @@ void AccelerationConfiguration::addTypeSpecificSubtags(
                                                    VALUE_RESIDUAL_SUM_PRECONDITIONER})
                                       .setDocumentation("The type of the preconditioner.");
     tagPreconditioner.addAttribute(attrPreconditionerType);
+    auto attrPreconditionerUpdate = XMLAttribute<std::string>(ATTR_PRECOND_UPDATE, VALUE_RESIDUAL_SUM_ALWAYS)
+                                        .setOptions({VALUE_RESIDUAL_SUM_ALWAYS,
+                                                     VALUE_RESIDUAL_SUM_THRESHOLD})
+                                        .setDocumentation("To update the preconditioner weights after the first time window:"
+                                                          "- `always`: The preconditioner weights are updated after every iteration.\n"
+                                                          "- `on-threshold`: The preconditioner weights are only updated if the relative difference of the weights is larger than the limit."
+                                                          "If this tag is not provided, the preconditioner weights are updated after every iteration when it's not frozen.");
```
uekerman, line 521: How does "always" and "frozen" interplay?
Shouldn't we make "frozen" an option here as well?
(similarly for IMVJ below)

uekerman (changes requested):
Thanks for taking care of this one, highly appreciated :pray: 

- Do we actually need the "updatePreconLimit" as a configuration option? Would we ever want to use a different value than 10?
- Should we update the default ILS and IMVJ configurations? (If yes, please also update the changelog) I would not consider this breaking.
- Could we unit-test the QR3 filter somehow? Currently, we don't test anywhere that it actually does the right thing, correct?

src/acceleration/config/AccelerationConfiguration.cpp
Fujikawas: As I understand from the paper: 
There is an option to define the time window number to start freezing the scaling weights. 
If this number is larger than 1,  we check the option `always` or `on-threshold`. If it's `always`, the weights will be recomputed after every iteration, otherwise from the 2nd time window, the weights will be updated when it's dramatically changing.
If the number is one, then `always` and  `on-threshold` are the same.

I think to change it into a boolean named "update_on_threshold"  may make this option easier to understand.

uekerman:
> After setting the default filter to QR3, now it's used in several integration tests.

True, but those integration tests typically only test that it doesn't break, not that the filter does the correct thing, no?

docs/changelog/1493.md
```diff
@@ -0,0 +1 @@
+- Added QR3 filter and weight monitoring for residual-sum preconditioner. Set QR3 as default filter.
```
uekerman, line 1:
```suggestion
- Added QR3 filter and use it in default IQN-ILS and IQN-IMVJ configurations (see https://doi.org/10.3390/mca27030040).
- Added weight monitoring for residual-sum preconditioner and use it in default IQN-ILS and IQN-IMVJ configurations (see https://doi.org/10.3390/mca27030040).
```

src/acceleration/config/AccelerationConfiguration.cpp
```diff
@@ -492,6 +498,12 @@ void AccelerationConfiguration::addTypeSpecificSubtags(
                                                    VALUE_RESIDUAL_SUM_PRECONDITIONER})
                                       .setDocumentation("The type of the preconditioner.");
     tagPreconditioner.addAttribute(attrPreconditionerType);
+    auto attrpreconditionerUpdateOnThreshold = XMLAttribute<bool>(ATTR_PRECOND_UPDATE_ON_THRESHOLD, true)
+                                                   .setDocumentation("To update the preconditioner weights after the first time window:"
+                                                                     "- `true`: The preconditioner weights are only updated if the relative difference of the weights is larger than the limit.\n"
+                                                                     "- `false`: The preconditioner weights are updated after every iteration.\n"
+                                                                     "When this tag is not provided, the value is set to `true` by default.");
```
uekerman, line 505: No need to document the default. This is automatically rendered, see https://precice.org/configuration-xml-reference.html

src/acceleration/config/AccelerationConfiguration.cpp
```diff
@@ -565,6 +577,12 @@ void AccelerationConfiguration::addTypeSpecificSubtags(
                                                    VALUE_RESIDUAL_SUM_PRECONDITIONER})
                                       .setDocumentation("Type of the preconditioner.");
     tagPreconditioner.addAttribute(attrPreconditionerType);
+    auto attrpreconditionerUpdateOnThreshold = XMLAttribute<bool>(ATTR_PRECOND_UPDATE_ON_THRESHOLD, true)
+                                                   .setDocumentation("To update the preconditioner weights after the first time window:"
+                                                                     "- `true`: The preconditioner weights are only updated if the relative difference of the weights is larger than the limit.\n"
+                                                                     "- `false`: The preconditioner weights are updated after every iteration.\n"
+                                                                     "When this tag is not provided, the value is set to `true` by default.");
```
uekerman, line 584: see above

src/acceleration/config/AccelerationConfiguration.cpp
```diff
@@ -492,6 +498,12 @@ void AccelerationConfiguration::addTypeSpecificSubtags(
                                                    VALUE_RESIDUAL_SUM_PRECONDITIONER})
                                       .setDocumentation("The type of the preconditioner.");
     tagPreconditioner.addAttribute(attrPreconditionerType);
+    auto attrpreconditionerUpdateOnThreshold = XMLAttribute<bool>(ATTR_PRECOND_UPDATE_ON_THRESHOLD, true)
+                                                   .setDocumentation("To update the preconditioner weights after the first time window:"
+                                                                     "- `true`: The preconditioner weights are only updated if the relative difference of the weights is larger than the limit.\n"
```
uekerman, line 503: now you removed the "limit", right? Then, we should also update the docs here

src/acceleration/config/AccelerationConfiguration.cpp
```diff
@@ -565,6 +577,12 @@ void AccelerationConfiguration::addTypeSpecificSubtags(
                                                    VALUE_RESIDUAL_SUM_PRECONDITIONER})
                                       .setDocumentation("Type of the preconditioner.");
     tagPreconditioner.addAttribute(attrPreconditionerType);
+    auto attrpreconditionerUpdateOnThreshold = XMLAttribute<bool>(ATTR_PRECOND_UPDATE_ON_THRESHOLD, true)
+                                                   .setDocumentation("To update the preconditioner weights after the first time window:"
+                                                                     "- `true`: The preconditioner weights are only updated if the relative difference of the weights is larger than the limit.\n"
```
uekerman, line 582: see above

uekerman (approved):

MakisH:
@uekerman the system tests run you triggered failed immediately, as branches from forks are not yet supported. I triggered a new one on the merged commit: https://github.com/precice/tutorials/actions/runs/14036757914/job/39296515606

This is now failing with results regressions, so the regression was introduced in with this PR, and not in the substeps one. (cc @BenjaminRodenberg)

PR #1494, opened 2022-11-29
Mesh connectivity preprocessing

### LINKED ISSUE #1153: Define primitives in terms of vertices

### LINKED ISSUE #1321: Clarify mesh API

fsimonis:
## Main changes of this PR

This PR adds:
- Weak ordering to primitives, making them sortable
- Add pre-process to mesh which
  - firstly removes duplicates of explicitly specified primitives (user), and
  - secondly adds missing implicit primitives. (edges of triangles etc)
- Fix integration tests
- Add pre-processing tests

Closes #1321 and closes #1153

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mesh/Mesh.cpp
+++ b/src/mesh/Mesh.cpp
@@ -1,7 +1,6 @@
 #include <Eigen/Core>
 #include <algorithm>
 #include <array>
-#include <bits/utility.h>
 #include <boost/container/flat_map.hpp>
 #include <functional>
 #include <memory>
--- a/src/mesh/Mesh.cpp
+++ b/src/mesh/Mesh.cpp
@@ -401,7 +401,7 @@ void Mesh::removeDuplicates()
   auto lastEdge = std::unique(_edges.begin(), _edges.end());
   _edges        = EdgeContainer{_edges.begin(), lastEdge};
 
-  PRECICE_DEBUG("Deduplication removed {} tetrahedra ({} to {}), {} triangles ({} to {}), and {} edges ({} to {})",
+  PRECICE_DEBUG("Compression removed {} tetrahedra ({} to {}), {} triangles ({} to {}), and {} edges ({} to {})",
                 tetrahedraCnt - _tetrahedra.size(), tetrahedraCnt, _tetrahedra.size(),
                 triangleCnt - _triangles.size(), triangleCnt, _triangles.size(),
                 edgeCnt - _edges.size(), edgeCnt, _edges.size());
--- a/src/mesh/Mesh.cpp
+++ b/src/mesh/Mesh.cpp
@@ -1,6 +1,7 @@
 #include <Eigen/Core>
 #include <algorithm>
 #include <array>
+#include <bits/utility.h>
 #include <boost/container/flat_map.hpp>
 #include <functional>
 #include <memory>
@@ -16,6 +17,7 @@
 #include "logging/LogMacros.hpp"
 #include "math/geometry.hpp"
 #include "mesh/Data.hpp"
+#include "mesh/Vertex.hpp"
 #include "precice/types.hpp"
 #include "query/Index.hpp"
 #include "utils/EigenHelperFunctions.hpp"
@@ -372,6 +374,124 @@ const BoundingBox &Mesh::getBoundingBox() const
 void Mesh::expandBoundingBox(const BoundingBox &boundingBox)
 {
   _boundingBox.expandBy(boundingBox);
+}
+
+void Mesh::preprocess()
+{
+  removeDuplicates();
+  generateImplictPrimitives();
+}
+
+void Mesh::removeDuplicates()
+{
+  // Remove duplicate tetrahedra
+  auto tetrahedraCnt = _tetrahedra.size();
+  std::sort(_tetrahedra.begin(), _tetrahedra.end());
+  auto lastTetrahedron = std::unique(_tetrahedra.begin(), _tetrahedra.end());
+  _tetrahedra          = TetraContainer{_tetrahedra.begin(), lastTetrahedron};
+
+  // Remove duplicate triangles
+  auto triangleCnt = _triangles.size();
+  std::sort(_triangles.begin(), _triangles.end());
+  auto lastTriangle = std::unique(_triangles.begin(), _triangles.end());
+  _triangles        = TriangleContainer{_triangles.begin(), lastTriangle};
+
+  // Remove duplicate edges
+  auto edgeCnt = _edges.size();
+  std::sort(_edges.begin(), _edges.end());
+  auto lastEdge = std::unique(_edges.begin(), _edges.end());
+  _edges        = EdgeContainer{_edges.begin(), lastEdge};
+
+  PRECICE_DEBUG("Deduplication removed {} tetrahedra ({} to {}), {} triangles ({} to {}), and {} edges ({} to {})",
+                tetrahedraCnt - _tetrahedra.size(), tetrahedraCnt, _tetrahedra.size(),
+                triangleCnt - _triangles.size(), triangleCnt, _triangles.size(),
+                edgeCnt - _edges.size(), edgeCnt, _edges.size());
+}
+
+namespace {
+
+template <class Primitive, int... Indices>
+auto sortedVertexPtrsForImpl(Primitive &p, std::integer_sequence<int, Indices...>)
+{
+  std::array<Vertex *, Primitive::vertexCount> vs{&p.vertex(Indices)...};
+  std::sort(vs.begin(), vs.end());
+  return std::tuple_cat(vs);
+}
+
+/** returns a tuple of Vertex* sorted by ptr of the Primitive
+ * Requires Primitive to provide static constexpr vertexCount.
+ * Generates an integer sequence based on the vertexCount, which is then expanded to fill the array.
+ */
+template <class Primitive>
+auto sortedVertexPtrsFor(Primitive &p)
+{
+  return sortedVertexPtrsForImpl(p, std::make_integer_sequence<int, Primitive::vertexCount>{});
+}
+} // namespace
+
+void Mesh::generateImplictPrimitives()
+{
+  if (_triangles.empty() && _tetrahedra.empty()) {
+    PRECICE_DEBUG("No implicit primitives required");
+    return; // no implicit primitives
+  }
+
+  // count explicit primitives for debug
+  const auto explTriangles = _triangles.size();
+  const auto explEdges     = _triangles.size();
+
+  // First handle all explicit tetrahedra
+
+  // Build a set of all explicit triangles
+  using ExisitingTriangle = std::tuple<Vertex *, Vertex *, Vertex *>;
+  std::set<ExisitingTriangle> triangles;
+  for (auto &t : _triangles) {
+    triangles.insert(sortedVertexPtrsFor(t));
+  }
+
+  // Generate all missing implicit triangles of explicit tetrahedra
+  // Update the triangles set used by the implicit edge generation
+  auto createTriangleIfMissing = [&](Vertex *a, Vertex *b, Vertex *c) {
+    if (triangles.count({a, b, c}) == 0) {
+      triangles.emplace(a, b, c);
+      createTriangle(*a, *b, *c);
+    };
+  };
+  for (auto &t : _tetrahedra) {
+    auto [a, b, c, d] = sortedVertexPtrsFor(t);
+    // Make sure these are in the same order as above
+    createTriangleIfMissing(a, b, c);
+    createTriangleIfMissing(a, b, d);
+    createTriangleIfMissing(a, c, d);
+    createTriangleIfMissing(b, c, d);
+  }
+
+  // Second handle all triangles, both explicit and implicit from the tetrahedron phase
+  // Build an set of all explicit triangles
+  using ExisitingEdge = std::tuple<Vertex *, Vertex *>;
+  std::set<ExisitingEdge> edges;
+  for (auto &e : _edges) {
+    edges.emplace(sortedVertexPtrsFor(e));
+  }
+
+  // generate all missing implicit edges of implicit and explicit triangles
+  auto createEdgeIfMissing = [&](Vertex *a, Vertex *b) {
+    if (edges.count({a, b}) == 0) {
+      edges.emplace(a, b);
+      createEdge(*a, *b);
+    };
+  };
+  for (auto &t : _triangles) {
+    auto [a, b, c] = sortedVertexPtrsFor(t);
+    // Make sure these are in the same order as above
+    createEdgeIfMissing(a, b);
+    createEdgeIfMissing(a, c);
+    createEdgeIfMissing(b, c);
+  }
+
+  PRECICE_DEBUG("Generated {} implicit triangles and {} implicit edges",
+                _triangles.size() - explTriangles,
+                _edges.size() - explEdges);
 }
 
 bool Mesh::operator==(const Mesh &other) const
--- a/src/mesh/Mesh.cpp
+++ b/src/mesh/Mesh.cpp
@@ -418,6 +418,10 @@ auto sortedVertexPtrsForImpl(Primitive &p, std::integer_sequence<int, Indices...
 }
 
 /** returns a tuple of Vertex* sorted by ptr of the Primitive
+ *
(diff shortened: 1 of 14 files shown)


src/mesh/Mesh.cpp
```diff
@@ -374,6 +375,124 @@ void Mesh::expandBoundingBox(const BoundingBox &boundingBox)
...
+
+  // count explicit primitives for debug
+  const auto explTriangles = _triangles.size();
+  const auto explEdges     = _triangles.size();
+
+  // First handle all explicit tetrahedra
+
+  // Build a set of all explicit triangles
+  using ExisitingTriangle = std::tuple<Vertex *, Vertex *, Vertex *>;
+  std::set<ExisitingTriangle> triangles;
```
davidscn, line 446: Cool algortihm(s). What is the motivation for using a set here? I think I currently don't get where we store the primitives generated here.

src/mesh/tests/MeshTest.cpp
```diff
@@ -821,5 +821,288 @@ BOOST_AUTO_TEST_CASE(AddMesh)
   BOOST_TEST(globalMesh->tetrahedra().size() == 3);
 }
 
+BOOST_AUTO_TEST_CASE(RemoveDuplicates)
```
davidscn, line 824: I guess we test here the implicit primitive generation as well, right?

src/precice/SolverInterface.hpp
```diff
@@ -254,10 +254,9 @@ class SolverInterface {
    * Use isMeshConnectivityRequired() to check if the current participant can make use of the connectivity.
    *
    *
-   * Always set the mesh connectivity of the highest dimensionality available.
-   * preCICE ensures the existence of hierarchical entries for the projection fallback.
-   * Prefer to use bulk versions, as they allows preCICE to efficiently avoid duplicates.
-   * preCICE removes all connectivity duplicates in initialize().
+   * Only set the mesh connectivity that you require.
+   * preCICE removes all connectivity duplicates in initialize() and ensures
+   * the existence of hierarchical entries for the projection fallback.
```
davidscn, line 259: I get what you are saying here, but I'm not sure if users will understand what you mean here. The connectivity is anyway somewhat complicated. I would recommend to make this description a bit more verbose, or even remove the last part (does the user need to know about fallbacks etc.?).

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -763,8 +771,22 @@ void SolverInterfaceImpl::setMeshEdges(
     int        size,
     const int *vertices)
 {
+  PRECICE_TRACE(meshID, size);
+  PRECICE_REQUIRE_MESH_MODIFY(meshID);
+  MeshContext &context = _accessor->usedMeshContext(meshID);
+  if (context.meshRequirement != mapping::Mapping::MeshRequirement::FULL) {
+    return;
```
davidscn, line 778: Do we want to document this behavior for the user and or developers?

davidscn (commented):
Some really nice code included here, only a few comments below.

src/mesh/Mesh.cpp
```diff
@@ -374,6 +375,124 @@ void Mesh::expandBoundingBox(const BoundingBox &boundingBox)
...
+
+template <class Primitive, int... Indices>
+auto sortedVertexPtrsForImpl(Primitive &p, std::integer_sequence<int, Indices...>)
+{
+  std::array<Vertex *, Primitive::vertexCount> vs{&p.vertex(Indices)...};
+  std::sort(vs.begin(), vs.end());
+  return std::tuple_cat(vs);
+}
+
+/** returns a tuple of Vertex* sorted by ptr of the Primitive
```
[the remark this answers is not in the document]
fsimonis: I need to create missing primitives, which requires `Vertex&`.
I also need to keep track of what already exists, which requires something sortable.

Using `VertexID`s turned out to be very cumbersome, as I need to look up the vertices both when creating the existing set of primitives, and when adding a missing primitive.

Using pointers simplifies all of this as they are sortable and reference each Vertex. I now fetch the `Vertex*` once per primitive, then I can query and create implicit primitives without furter lookups.

src/mesh/Mesh.cpp
fsimonis: This is actually very math-like.

A triangle is a tuple of 3 vertices, these are sorted to eliminate permutations.
We keep track of a set of triangles.
We fill this set with existing triangles.
Then we check if implicitly defined triangles are missing, if so, we add them to both the mesh and the set.

src/mesh/tests/MeshTest.cpp
fsimonis: True. I copy the edge deduplication into a separate test.

src/precice/impl/SolverInterfaceImpl.cpp
fsimonis: This is part of the docs on the website. You mean to document this in doxygen, right?

Instead of silently doing nothing, we could also warn that the adapter dev should use `requiresMeshConnectivityFor()` to prevent this.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -763,8 +771,22 @@ void SolverInterfaceImpl::setMeshEdges(
...
+  }
+
+  mesh::PtrMesh &mesh = context.mesh;
+  utils::for_each_unique(vertices, vertices + size, [&_log = _log, &mesh](VertexID vid) {
+    PRECICE_CHECK(mesh->isValidVertexID(vid), impl::errorInvalidVertexID(vid));
```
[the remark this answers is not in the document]
fsimonis: This error message is actually useless, as unactionable.

I have an idea that should make way more sense than this.
The performance fear is a bit silly too.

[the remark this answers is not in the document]
fsimonis: Example of an actionable error:
```
setMeshTriangles|ERROR: The given range of VertexIDs contain invalid IDs at offsets 10-16. Check that they originated from calls to setMeshVertex() or setMeshVertices().
```

src/mesh/Mesh.cpp
```diff
@@ -374,6 +375,124 @@ void Mesh::expandBoundingBox(const BoundingBox &boundingBox)
...
+void Mesh::preprocess()
+{
+  removeDuplicates();
+  generateImplictPrimitives();
+}
+
+void Mesh::removeDuplicates()
+{
+  // Remove duplicate tetrahedra
+  auto tetrahedraCnt = _tetrahedra.size();
```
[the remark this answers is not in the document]
fsimonis: This results in 2 messages per rank and solver. I started with 2 separate before and after messages. This requires searching and matching these messages to find out what the diff is.
One message is less tedious.

src/precice/impl/SolverInterfaceImpl.cpp
fsimonis: Added this to the docs

src/precice/SolverInterface.hpp
uekerman: I also don't understand exactly what you mean by "hierarchical entries".

> Only set the mesh connectivity that you require

Will users understand what we mean with this?

Would it make sense to extend the examples below by what preCICE will remove?

src/precice/impl/CommonErrorMessages.hpp
```diff
@@ -16,6 +16,8 @@ inline std::string errorInvalidEdgeID(int eid)
   return "The given EdgeID \"" + std::to_string(eid) + "\" is invalid. Check that it originated from a call to setMeshEdge().";
 }
 
+static constexpr auto errorInvalidVertexIDRange = "The given range of VertexIDs contains invalid IDs at offsets [{},{}). Check that they originated from calls to setMeshVertex() or setMeshVertices().";
```
uekerman, line 19:
```suggestion
static constexpr auto errorInvalidVertexIDRange = "The given range of VertexIDs contains invalid IDs at offsets [{},{}]. Check that they originated from calls to setMeshVertex() or setMeshVertices().";
```
... could be easier to understand

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -274,6 +274,15 @@ double SolverInterfaceImpl::initialize()
   Event                    e("initialize", precice::syncMode);
   utils::ScopedEventPrefix sep("initialize/");
 
+  PRECICE_DEBUG("Compressing provided meshes");
+  Event e1("mesh compression", precice::syncMode);
```
uekerman, line 278: I think we have not documented this properly, but there is some kind of consistency in event naming.
"mesh.compression" + name of the mesh etc.

src/precice/impl/SolverInterfaceImpl.cpp
uekerman: > Instead of silently doing nothing, we could also warn that the adapter dev should use requiresMeshConnectivityFor() to prevent this.

"should" is a bit too strict here in my opinion. `requiresMeshConnectivityFor()` is mainly meant to save compute effort in the adapter, not in preCICE.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -882,8 +925,59 @@ void SolverInterfaceImpl::setMeshQuads(
...
+    auto did = vertices[4 * i + 3];
+
+    auto vertexIDs = utils::make_array(aid, bid, cid, did);
+    PRECICE_CHECK(utils::unique_elements(vertexIDs), "The four vertex ID's of the quad nr {} are not unique. Please check that the vertices that form the quad are correct.", i);
+
+    auto coords = mesh::coordsFor(mesh, vertexIDs);
+    PRECICE_CHECK(utils::unique_elements(coords),
+                  "The four vertices that form the quad nr {} are not unique. The resulting shape may be a point, line or triangle."
+                  "Please check that the adapter sends the four unique vertices that form the quad, or that the mesh on the interface "
+                  "is composed of quads. A mix of triangles and quads are not supported.",
```
uekerman, line 959: What do you mean by "A mix of triangles and quads are not supported"?

src/mesh/Mesh.cpp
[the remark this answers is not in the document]
uekerman: Good information to add directly as comment in the code

uekerman (approved):
Looks good to me. I mainly focused on the user perspective, not the details of the algorithm.

src/mesh/Mesh.cpp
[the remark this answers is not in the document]
fsimonis: Done

src/precice/SolverInterface.hpp
fsimonis: Extended the docs and listed `{2D, 3D}x{Surface, Volume}`.

src/precice/impl/SolverInterfaceImpl.cpp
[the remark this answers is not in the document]
fsimonis: Changed this to the above error

src/precice/impl/SolverInterfaceImpl.cpp
fsimonis: Let's leave it at the doc then.

src/precice/impl/CommonErrorMessages.hpp
fsimonis: It's an open interval though :wink: so either `[A,B[` or `[A,B)`. The cppreference even uses the `)` version, so I consider it fairly wide-spread.

src/precice/impl/SolverInterfaceImpl.cpp
fsimonis: This was a placeholder, good point.

src/precice/impl/SolverInterfaceImpl.cpp
fsimonis: Problem with warning is that if the user calls `setMeshEdge` 1e6 times, then the logs will be very _very_ verbose.
To we would need to implement some kind of bookkeeping, which then needs to be maintained etc.

src/precice/impl/CommonErrorMessages.hpp
uekerman: > We could of course use a closed interval instead.

That's what I had in mind. As user I would find a closed interval easier to understand.

src/precice/impl/SolverInterfaceImpl.cpp
fsimonis: I explained the behaviour in the API documentation.

PR #1495, opened 2022-11-30
Consistently implement `BoundingBox`

playcatchup:
## Main changes of this PR
1. Add a BoundingBox constructor to accept two VectorXd arguments.
2. Replace _bounds (`vector`) with _boundMin and _boundMax (`VectorXd`).
3. Refactor `empty()` and create `isDefault()`.
4. Refactor functions, excluding those having Non-countable loops, using `Eigen`'s functions.

Co-authored-by: Jeremiah Bintang Santoso @jeremiahbintang

Closes https://github.com/precice/precice/issues/1418

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mesh/BoundingBox.cpp
+++ b/src/mesh/BoundingBox.cpp
@@ -27,9 +27,14 @@ BoundingBox::BoundingBox(Eigen::VectorXd boundMin, Eigen::VectorXd boundMax)
 BoundingBox::BoundingBox(std::vector<double> bounds)
 {
   PRECICE_ASSERT((int) bounds.size() == 4 || (int) bounds.size() == 6, "Dimension of a bounding box can only be 2 or 3. The dimension is {}.", bounds.size() / 2);
+
   _dimensions = bounds.size() / 2;
-  _boundMin   = Eigen::Map<Eigen::VectorXd, 0, Eigen::InnerStride<2>>(bounds.data(), _dimensions);
-  _boundMax   = Eigen::Map<Eigen::VectorXd, 0, Eigen::InnerStride<2>>(bounds.data() + 1, _dimensions);
+
+  // Eigen::Map maps an existing array to a Eigen's matrix or vector
+  // Here, its 1st input is the existing array's pointer (bounds.data()) and 2nd input is output VectorXd's size
+  // "bounds.data() + 1" means that output elements starts from the 2nd element of "bounds"
+  _boundMin = Eigen::Map<Eigen::VectorXd, 0, Eigen::InnerStride<2>>(bounds.data(), _dimensions);
+  _boundMax = Eigen::Map<Eigen::VectorXd, 0, Eigen::InnerStride<2>>(bounds.data() + 1, _dimensions);
 }
 
 BoundingBox::BoundingBox(int dimension)
@@ -65,6 +70,7 @@ bool BoundingBox::isDefault() const
 bool BoundingBox::contains(const mesh::Vertex &vertex) const
 {
   PRECICE_ASSERT(_dimensions == vertex.getDimensions(), "Vertex with different dimensions than this bounding box cannot be checked.");
+
   const auto &coords = vertex.rawCoords();
   for (int d = 0; d < _dimensions; d++) {
     if (coords[d] < _boundMin[d] || coords[d] > _boundMax[d]) {
@@ -77,6 +83,7 @@ bool BoundingBox::contains(const mesh::Vertex &vertex) const
 Eigen::VectorXd BoundingBox::center() const
 {
   PRECICE_ASSERT(!isDefault(), "Data of the bounding box is at default state.");
+
   return (_boundMax + _boundMin) * 0.5;
 }
 
@@ -93,6 +100,7 @@ Eigen::VectorXd BoundingBox::maxCorner() const
 double BoundingBox::getArea(std::vector<bool> deadAxis)
 {
   PRECICE_ASSERT(!isDefault(), "Data of the bounding box is at default state.");
+
   double meshArea = 1.0;
   for (int d = 0; d < _dimensions; d++)
     if (not deadAxis[d])
@@ -117,6 +125,7 @@ const std::vector<double> BoundingBox::dataVector() const
 void BoundingBox::expandBy(const BoundingBox &otherBB)
 {
   PRECICE_ASSERT(_dimensions == otherBB.getDimension(), "Other BoundingBox with different dimensions than this bounding box cannot be used to expand it.");
+
   _boundMin = _boundMin.cwiseMin(otherBB._boundMin);
   _boundMax = _boundMax.cwiseMax(otherBB._boundMax);
 }
@@ -124,6 +133,7 @@ void BoundingBox::expandBy(const BoundingBox &otherBB)
 void BoundingBox::expandBy(const Vertex &vertices)
 {
   PRECICE_ASSERT(_dimensions == vertices.getDimensions(), "Vertex with different dimensions than this bounding box cannot be used to expand it.");
+
   const auto coords = vertices.getCoords();
   _boundMin         = _boundMin.cwiseMin(coords);
   _boundMax         = _boundMax.cwiseMax(coords);
--- a/src/mesh/BoundingBox.cpp
+++ b/src/mesh/BoundingBox.cpp
@@ -46,8 +46,8 @@ BoundingBox::BoundingBox(int dimension)
   _boundMax = Eigen::VectorXd(dimension);
 
   // Define 'illegal' BoundingBox: _boundMin > _boundMax
-  std::fill(_boundMin.begin(), _boundMin.end(), std::numeric_limits<double>::max());
-  std::fill(_boundMax.begin(), _boundMax.end(), std::numeric_limits<double>::lowest());
+  _boundMin.fill(std::numeric_limits<double>::max());
+  _boundMax.fill(std::numeric_limits<double>::lowest());
 }
 
 bool BoundingBox::operator==(const BoundingBox &otherBB) const
--- a/src/mesh/BoundingBox.cpp
+++ b/src/mesh/BoundingBox.cpp
@@ -15,7 +15,8 @@ logging::Logger BoundingBox::_log{"mesh::BoundingBox"};
 
 BoundingBox::BoundingBox(Eigen::VectorXd boundMin, Eigen::VectorXd boundMax)
 {
-  PRECICE_ASSERT(boundMin.rows() == boundMax.rows(), "Dimension of min {} and max {} vertices should be the same.", boundMin.rows(), boundMax.rows());
+  PRECICE_ASSERT((boundMin.rows() == 2 && boundMax.rows() == 2) || (boundMin.rows() == 3 && boundMax.rows() == 3),
+                 "Dimension of min {} and max {} vertices should be the same and both 2 or 3.", boundMin.rows(), boundMax.rows());
   PRECICE_ASSERT((boundMin - boundMax).maxCoeff() < 0, "Each component of min vertex {} must be <= max vertex {} in the same axis direction.", boundMin, boundMax);
 
   _boundMin   = std::move(boundMin);
--- a/src/mesh/BoundingBox.cpp
+++ b/src/mesh/BoundingBox.cpp
@@ -51,6 +51,11 @@ bool BoundingBox::operator==(const BoundingBox &otherBB) const
   return _boundMin.isApprox(otherBB._boundMin) && _boundMax.isApprox(otherBB._boundMax);
 }
 
+bool BoundingBox::empty() const
+{
+  return (_boundMax - _boundMin).isZero();
+}
+
 bool BoundingBox::isDefault() const
 {
   return _boundMin.isApproxToConstant(std::numeric_limits<double>::max()) && _boundMax.isApproxToConstant(std::numeric_limits<double>::lowest());
--- a/src/mesh/BoundingBox.cpp
+++ b/src/mesh/BoundingBox.cpp
@@ -40,8 +40,8 @@ BoundingBox::BoundingBox(int dimension)
   _boundMax = Eigen::VectorXd(dimension);
 
   // Define 'illegal' BoundingBox: _boundMin > _boundMax
-  std::fill(_boundMin.data(), _boundMin.data() + _boundMin.rows(), std::numeric_limits<double>::max());
-  std::fill(_boundMax.data(), _boundMax.data() + _boundMax.rows(), std::numeric_limits<double>::lowest());
+  std::fill(_boundMin.begin(), _boundMin.end(), std::numeric_limits<double>::max());
+  std::fill(_boundMax.begin(), _boundMax.end(), std::numeric_limits<double>::lowest());
 }
 
 bool BoundingBox::operator==(const BoundingBox &otherBB) const
@@ -53,7 +53,7 @@ bool BoundingBox::operator==(const BoundingBox &otherBB) const
 
 bool BoundingBox::empty() const
 {
-  return (_boundMax - _boundMin).isZero();
+  return (_boundMax - _boundMin).isZero() || isDefault();
 }
 
 bool BoundingBox::isDefault() const
@@ -76,11 +76,7 @@ bool BoundingBox::contains(const mesh::Vertex &vertex) const
 Eigen::VectorXd BoundingBox::center() const
 {
   PRECICE_ASSERT(!isDefault(), "Data of the bounding box is at default state.");
-  Eigen::VectorXd cog(_dimensions);
-  for (int d = 0; d < _dimensions; d++) {
-    cog[d] = (_boundMax[d] + _boundMin[d]) / 2.0;
-  }
-  return cog;
+  return (_boundMax + _boundMin) * 0.5;
 }
 
 Eigen::VectorXd BoundingBox::minCorner() const
@@ -120,29 +116,23 @@ const std::vector<double> BoundingBox::dataVector() const
 void BoundingBox::expandBy(const BoundingBox &otherBB)
 {
   PRECICE_ASSERT(_dimensions == otherBB.getDimension(), "Other BoundingBox with different dimensions than this bounding box cannot be used to expand it.");
-  for (int d = 0; d < _dimensions; d++) {
-    _boundMin[d] = std::min(_boundMin[d], otherBB._boundMin[d]);
-    _boundMax[d] = std::max(_boundMax[d], otherBB._boundMax[d]);
-  }
+  _boundMin = _boundMin.cwiseMin(otherBB._boundMin);
+  _boundMax = _boundMax.cwiseMax(otherBB._boundMax);
 }
 
 void BoundingBox::expandBy(const Vertex &vertices)
 {
   PRECICE_ASSERT(_dimensions == vertices.getDimensions(), "Vertex with different dimensions than this bounding box cannot be used to expand it.");
-  const auto coords = vertices.rawCoords();
-  for (int d = 0; d < _dimensions; ++d) {
-    _boundMin[d] = std::min(coords[d], _boundMin[d]);
-    _boundMax[d] = std::max(coords[d], _boundMax[d]);
-  }
+  const auto coords = vertices.getCoords();
+  _boundMin         = _boundMin.cwiseMin(coords);
+  _boundMax         = _boundMax.cwiseMax(coords);
 }
 
 void BoundingBox::expandBy(double value)
 {
   if (!isDefault()) {
-    for (int d = 0; d < _dimensions; d++) {
-      _boundMin[d] -= value;
-      _boundMax[d] += value;
-    }
+    _boundMin.array() -= value;
+    _boundMax.array() += value;
   }
 }
 
--- a/src/mesh/BoundingBox.cpp
+++ b/src/mesh/BoundingBox.cpp
@@ -148,15 +148,12 @@ void BoundingBox::expandBy(double value)
 
 void BoundingBox::scaleBy(double safetyFactor)
 {
-  // TODO: the reason why BoundingBox should not be default state here is still unknown
   if (!isDefault()) {
     double maxSideLength = 1e-6; // we need some minimum > 0 here
     maxSideLength        = (_boundMax - _boundMin).maxCoeff();
-    for (int d = 0; d < _dimensions; d++) {
-      _boundMax[d] += safetyFactor * maxSideLength;
-      _boundMin[d] -= safetyFactor * maxSideLength;
-      PRECICE_DEBUG("Merged BoundingBox {}", *this);
-    }
+    _boundMax.array() += safetyFactor * maxSideLength;
+    _boundMin.array() -= safetyFactor * maxSideLength;
+    PRECICE_DEBUG("Merged BoundingBox {}", *this);
   }
 }
 
--- a/src/mesh/BoundingBox.cpp
+++ b/src/mesh/BoundingBox.cpp
@@ -133,7 +133,6 @@ void BoundingBox::expandBy(const Vertex &vertices)
 
 void BoundingBox::expandBy(double value)
 {
-  // BoundingBox should not be default state; otherwise, it will make ReceivedPartitionTest fail because of incorrect connectionMapSize
   if (!isDefault()) {
     for (int d = 0; d < _dimensions; d++) {
       _boundMin[d] -= value;
--- a/src/mesh/BoundingBox.cpp
+++ b/src/mesh/BoundingBox.cpp
@@ -64,12 +64,12 @@ bool BoundingBox::operator==(const BoundingBox &otherBB) const
 
 bool BoundingBox::empty() const
 {
-  for (int i = 0; i < _dimensions; ++i) {
-    if (_boundMin[i] > _boundMax[i]) {
-      return true;
-    }
-  }
-  return false;
+  return (_boundMax - _boundMin).isZero();
+}
+
+bool BoundingBox::isDefault() const
+{
+  return _boundMin.isApproxToConstant(std::numeric_limits<double>::lowest()) && _boundMax.isApproxToConstant(std::numeric_limits<double>::max());
 }
 
 bool BoundingBox::contains(const mesh::Vertex &vertex) const
--- a/src/mesh/BoundingBox.cpp
+++ b/src/mesh/BoundingBox.cpp
@@ -111,7 +111,7 @@ int BoundingBox::getDimension() const
 
 const std::vector<double> BoundingBox::dataVector() const
 {
-  if (_boundMin.rows() == 3) {
+  if (_dimensions == 3) {
     std::vector<double> _bounds{_boundMin[0], _boundMax[0], _boundMin[1], _boundMax[1], _boundMin[2], _boundMax[2]};
   } else {
     std::vector<double> _bounds{_boundMin[0], _boundMax[0], _boundMin[1], _boundMax[1]};
--- a/src/mesh/BoundingBox.cpp
+++ b/src/mesh/BoundingBox.cpp
@@ -16,19 +16,19 @@ logging::Logger BoundingBox::_log{"mesh::BoundingBox"};
 BoundingBox::BoundingBox(Eigen::VectorXd boundMin, Eigen::VectorXd boundMax)
 {
   // PRECICE_ASSERT 1
-  PRECICE_ASSERT(boundMin.rows() == boundMax.rows(), "Dimension of min and max vertices should be the same.", boundMin.rows(), boundMax.rows());
+  PRECICE_ASSERT(boundMin.rows() == boundMax.rows(), "Dimension of min ({}) and max ({}) vertices should be the same.", boundMin.rows(), boundMax.rows());
 
   // PRECICE_ASSERT 2
-  PRECICE_ASSERT((boundMin - boundMax).maxCoeff() > 0, "Each component of min vertex must be less than or equal to that of max vertex in the same axis direction.", boundMin, boundMax);
+  PRECICE_ASSERT((boundMin - boundMax).maxCoeff() < 0, "Each component of min vertex ({}) must be <= max vertex ({}) in the same axis direction.", boundMin, boundMax);
 
   _boundMin   = std::move(boundMin);
   _boundMax   = std::move(boundMax);
-  _dimensions = boundMin.rows();
+  _dimensions = _boundMin.rows();
 }
 
 BoundingBox::BoundingBox(std::vector<double> bounds)
 {
-  PRECICE_ASSERT((int) bounds.size() == 4 || (int) bounds.size() == 6, "Dimension of a bounding box can only be 2 or 3.", bounds.size() / 2);
+  PRECICE_ASSERT((int) bounds.size() == 4 || (int) bounds.size() == 6, "Dimension of a bounding box can only be 2 or 3. The dimension is ({}).", bounds.size() / 2);
   _dimensions = bounds.size() / 2;
   _boundMin   = Eigen::Map<Eigen::VectorXd, 0, Eigen::InnerStride<2>>(bounds.data(), _dimensions);
   _boundMax   = Eigen::Map<Eigen::VectorXd, 0, Eigen::InnerStride<2>>(bounds.data() + 1, _dimensions);
@@ -37,17 +37,19 @@ BoundingBox::BoundingBox(std::vector<double> bounds)
 BoundingBox::BoundingBox(int dimension)
     : _dimensions(dimension)
 {
-  PRECICE_ASSERT(dimension == 2 || dimension == 3, "Dimension of a bounding box can only be 2 or 3.", dimension);
+  PRECICE_ASSERT(dimension == 2 || dimension == 3, "Dimension of a bounding box ({}) can only be 2 or 3.", dimension);
 
   _boundMin = Eigen::VectorXd(dimension);
   _boundMax = Eigen::VectorXd(dimension);
-  std::fill(_boundMin.data(), _boundMin.data() + _boundMin.rows(), std::numeric_limits<double>::lowest());
-  std::fill(_boundMax.data(), _boundMax.data() + _boundMax.rows(), std::numeric_limits<double>::max());
+
+  // Define 'illegal' BoundingBox: _boundMin > _boundMax
+  std::fill(_boundMin.data(), _boundMin.data() + _boundMin.rows(), std::numeric_limits<double>::max());
+  std::fill(_boundMax.data(), _boundMax.data() + _boundMax.rows(), std::numeric_limits<double>::lowest());
 }
 
 bool BoundingBox::operator==(const BoundingBox &otherBB) const
 {
-  PRECICE_ASSERT(_dimensions == otherBB._dimensions, "Bounding boxes with different dimensions cannot be compared.", _dimensions, otherBB._dimensions);
+  PRECICE_ASSERT(_dimensions == otherBB._dimensions, "Bounding boxes ({} and {}) with different dimensions cannot be compared.", _dimensions, otherBB._dimensions);
 
   return _boundMin.isApprox(otherBB._boundMin) && _boundMax.isApprox(otherBB._boundMax);
 }
@@ -59,12 +61,12 @@ bool BoundingBox::empty() const
 
 bool BoundingBox::isDefault() const
 {
-  return _boundMin.isApproxToConstant(std::numeric_limits<double>::lowest()) && _boundMax.isApproxToConstant(std::numeric_limits<double>::max());
+  return _boundMin.isApproxToConstant(std::numeric_limits<double>::max()) && _boundMax.isApproxToConstant(std::numeric_limits<double>::lowest());
 }
 
 bool BoundingBox::contains(const mesh::Vertex &vertex) const
 {
-  PRECICE_ASSERT(_dimensions == vertex.getDimensions(), "Vertex with different dimensions than bounding box cannot be checked.");
+  PRECICE_ASSERT(_dimensions == vertex.getDimensions(), "Vertex with different dimensions than this bounding box cannot be checked.");
   const auto &coords = vertex.rawCoords();
   for (int d = 0; d < _dimensions; d++) {
     if (coords[d] < _boundMin[d] || coords[d] > _boundMax[d]) {
@@ -76,7 +78,7 @@ bool BoundingBox::contains(const mesh::Vertex &vertex) const
 
 Eigen::VectorXd BoundingBox::center() const
 {
-  PRECICE_ASSERT(!empty(), "The BoundingBox is empty, i.e. it has zero area or zero volume.");
+  PRECICE_ASSERT(!isDefault(), "Data of the bounding box is at default state.");
   Eigen::VectorXd cog(_dimensions);
   for (int d = 0; d < _dimensions; d++) {
     cog[d] = (_boundMax[d] + _boundMin[d]) / 2.0;
@@ -96,7 +98,7 @@ Eigen::VectorXd BoundingBox::maxCorner() const
 
 double BoundingBox::getArea(std::vector<bool> deadAxis)
 {
-  PRECICE_ASSERT(!empty(), "The BoundingBox is empty, i.e. it has zero area or zero volume.");
+  PRECICE_ASSERT(!isDefault(), "Data of the bounding box is at default state.");
   double meshArea = 1.0;
   for (int d = 0; d < _dimensions; d++)
     if (not deadAxis[d])
@@ -122,7 +124,7 @@ const std::vector<double> BoundingBox::dataVector() const
 
 void BoundingBox::expandBy(const BoundingBox &otherBB)
 {
-  PRECICE_ASSERT(_dimensions == otherBB.getDimension(), "Other BoundingBox with different dimensions than bounding box cannot be used to expand bounding box");
+  PRECICE_ASSERT(_dimensions == otherBB.getDimension(), "Other BoundingBox with different dimensions than this bounding box cannot be used to expand it.");
   for (int d = 0; d < _dimensions; d++) {
     _boundMin[d] = std::min(_boundMin[d], otherBB._boundMin[d]);
     _boundMax[d] = std::max(_boundMax[d], otherBB._boundMax[d]);
@@ -131,26 +133,31 @@ void BoundingBox::expandBy(const BoundingBox &otherBB)
 
 void BoundingBox::expandBy(const Vertex &vertices)
 {
-  PRECICE_ASSERT(_dimensions == vertices.getDimensions(), "Vertex with different dimensions than bounding box cannot be used to expand bounding box");
+  PRECICE_ASSERT(_dimensions == vertices.getDimensions(), "Vertex with different dimensions than this bounding box cannot be used to expand it.");
   const auto coords = vertices.rawCoords();
   for (int d = 0; d < _dimensions; ++d) {
     _boundMin[d] = std::min(coords[d], _boundMin[d]);
     _boundMax[d] = std::max(coords[d], _boundMax[d]);
   }
 }
-// TODO: empty() definition
+
 void BoundingBox::expandBy(double value)
 {
-  PRECICE_ASSERT(!empty(), "The BoundingBox is empty, i.e. it has zero area or zero volume.");
+  // Adding or substracting a constant to the default state will still make an 'illegal' box
+  if (isDefault()) {
+    return;
+  }
   for (int d = 0; d < _dimensions; d++) {
     _boundMin[d] -= value;
     _boundMax[d] += value;
   }
 }
-// TODO: empty() definition
+
 void BoundingBox::scaleBy(double safetyFactor)
 {
-  PRECICE_ASSERT(!empty(), "The BoundingBox is empty, i.e. it has zero area or zero volume.");
+  if (isDefault()) {
+    return;
+  }
   double maxSideLength = 1e-6; // we need some minimum > 0 here
   for (int d = 0; d < _dimensions; d++) {
     if (_boundMax[d] > _boundMin[d])
--- a/src/mesh/BoundingBox.cpp
+++ b/src/mesh/BoundingBox.cpp
@@ -122,7 +122,7 @@ const std::vector<double> BoundingBox::dataVector() const
 
 void BoundingBox::expandBy(const BoundingBox &otherBB)
 {
-  PRECICE_ASSERT(_dimensions == otherBB.getDimensions(), "Other BoundingBox with different dimensions than bounding box cannot be used to expand bounding box");
+  PRECICE_ASSERT(_dimensions == otherBB.getDimension(), "Other BoundingBox with different dimensions than bounding box cannot be used to expand bounding box");
   for (int d = 0; d < _dimensions; d++) {
     _boundMin[d] = std::min(_boundMin[d], otherBB._boundMin[d]);
     _boundMax[d] = std::max(_boundMax[d], otherBB._boundMax[d]);
--- a/src/mesh/BoundingBox.cpp
+++ b/src/mesh/BoundingBox.cpp
@@ -53,10 +53,9 @@ BoundingBox::BoundingBox(int dimension)
   double max = std::numeric_limits<double>::max();
 
   if (dimension == 2) {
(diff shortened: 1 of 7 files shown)


BenjaminRodenberg:
Can you please use a more descriptive title for the PR? If it solves #1418 ("Inconsistent `BoundingBox` implementation") something like "Consistently implement BoundingBox" might work.

playcatchup:
Thank you @BenjaminRodenberg for correcting the unsuitable title! The title name is changed according to your advice.

src/mesh/BoundingBox.cpp
```diff
@@ -44,7 +76,7 @@ bool BoundingBox::operator==(const BoundingBox &otherBB) const
 bool BoundingBox::empty() const
 {
   for (int i = 0; i < _dimensions; ++i) {
-    if (_bounds[2 * i] > _bounds[2 * i + 1]) {
+    if (_boundMin[i] > _boundMax[i]) {
```
davidscn, line 79: To refer to your question: 

> but here comes a question: can 3D BoundingBox has zero length along one of the axes, i.e. has zero volume

I guess having 'dead' directions should still be a valid bounding box. We have such cases (also 2D dimensions with only a 1D interface, i.e., the bounding box would be 1D). In my opinion we have two options here: either we clearly return `false` in case we have at least one vertex inside the box (min and max bounds are the same) or we introduce a slightly different version of this function like `getActiveDimensions()`, which returns an `int` indicating the non-zero dimensions. The `empty` function is anyway mainly used for asserts, so that we can then ask in the code `PRECICE_ASSERT(getActiveDimensions >0)` or similar.

src/mesh/BoundingBox.cpp
```diff
@@ -108,14 +132,15 @@ int BoundingBox::getDimension() const
 
 const std::vector<double> BoundingBox::dataVector() const
 {
+  std::vector<double> _bounds{_boundMin[0], _boundMax[0], _boundMin[1], _boundMax[1], _boundMin[2], _boundMax[2]};
```
davidscn, line 135: This will lead to a segfault, when running in 2D, as `boundMin` is either 2D or 3D, right?

davidscn (commented):
> Could we define the isDefault() as "a BoundingBox have one vertex whose entries having the same value equal to std::numeric_limits<double>::lowest() and another one whose entries having the same value equal to std::numeric_limits<double>::max()"

Yes, sounds reasonable. Just make sure to document this properly.

src/mesh/BoundingBox.cpp
playcatchup: Thank @davidscn for your response : )

Quick questions please: what are 'dead' directions? What are them for? Does they relate to `deadAxis` as an input arguments in the function `getArea` of BoundingBox?
```cpp
double BoundingBox::getArea(std::vector<bool> deadAxis);
```

davidscn: No, I was just referring to directions which have zero expansion, i.e., min == max

src/mesh/BoundingBox.cpp
```diff
@@ -13,28 +13,60 @@ namespace precice::mesh {
...
+
+  // PRECICE_ASSERT 2
+  bool isMinLessThanMax = true;
+  for (int i = 0; i < dimMin; i++) {
+    if (boundMin[i] >= boundMax[i]) {
+      isMinLessThanMax = false;
+      break;
+    }
+  }
+  PRECICE_ASSERT(isMinLessThanMax, "Each component of min vertex must be less than that of max vertex in the same axis direction.", boundMin, boundMax);
```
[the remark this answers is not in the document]
jeremiahbintang: We changed it to check using maxCoeff from substracted bounds as we think it is more intuitive

src/mesh/BoundingBox.cpp
jeremiahbintang: Fixed @davidscn

src/mesh/BoundingBox.cpp
```diff
@@ -13,28 +13,60 @@ namespace precice::mesh {
...
+    _boundMax << max, max, max;
   }
 }
 
 bool BoundingBox::operator==(const BoundingBox &otherBB) const
 {
   PRECICE_ASSERT(_dimensions == otherBB._dimensions, "Bounding boxes with different dimensions cannot be compared.", _dimensions, otherBB._dimensions);
   for (int i = 0; i < _dimensions; ++i) {
-    if (_bounds.at(i) != otherBB._bounds.at(i)) {
+    if (_boundMin[i] != otherBB._boundMin[i] && _boundMax[i] != otherBB._boundMax[i]) {
```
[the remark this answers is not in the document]
jeremiahbintang: We decided to use the built-in isApprox method from Eigen, because we think it is more intuitive @davidscn

src/mesh/BoundingBox.cpp
jeremiahbintang: We tried to implement the getActiveDimensions, but we think the function would be the same as the getDimension method, because, in the new BB, the non-zero dimensions are the number of dimensions instantiated in the Eigen::VectorXd. @davidscn

playcatchup: We choose to implement empty() directly via Eigen's ```isZero()``` on difference of two vertices (shown below) instead of implementing getActiveDimensions() since it's quite simple and short.
```cpp
/// empty() returning true means that the length in every BoundingBox's dimension is equal to zero
bool BoundingBox::empty() const
{
  return (_boundMax - _boundMin).isZero();
}
```

@davidscn An additional question as regards empty() please. Which of member functions should be added the empty() check, i.e. ``` PRECICE_ASSERT(!empty(), "The BoundingBox is empty, i.e. it has zero area or zero volume.");```?

playcatchup:
@davidscn One question as regards ```BoundingBox::getArea``` please. What are the reasons that ```BoundingBox::getArea``` is not a ```const``` funciton and its argument is not a ```const``` reference? ```BoundingBox::getArea``` doesn't change anything of the BoundingBox object, and its vector<bool> argument won't be modified in it.

```
double BoundingBox::getArea(const std::vector<bool>& deadAxis) const
{
  PRECICE_ASSERT(!empty(), "The BoundingBox is empty, i.e. it has zero area or zero volume.");
  double meshArea = 1.0;
  for (int d = 0; d < _dimensions; d++)
    if (not deadAxis[d])
      meshArea *= _boundMax[d] - _boundMin[d];
  return meshArea;
}
```

playcatchup:
@davidscn @BenjaminRodenberg @MakisH   We (@jeremiahbintang ) wish you good health and happiness in 2023🥳 We're so thankful to have you to guide us in our AdvProg bonus project😊

playcatchup:
@davidscn Could you please again approve running actions for this pull request? We (@jeremiahbintang ) finally correct the errors, reported by actions, via:
1. use pre-commit to deal with the format of codes
2. solve the extra qualification errors
Thanks~

playcatchup:
@davidscn For this check, we've (@jeremiahbintang) found the reason why some checks were not successful: the initializer syntax of ```Eigen::VectorXd``` was wrong😟 They are now fixed! Would you be able to again approve running actions for this pull request?
Thanks for your time :)

davidscn:
> 

@playcatchup just to be clear: You can also run the tests locally on your machine using (after compiling) `ctest`.

playcatchup:
@davidscn Thanks for your information! We've not yet built preCICE in our local computer. We would follow the instructions here (https://precice.org/installation-source-preparation.html) to build one.

davidscn:
@playcatchup What's the status of this PR? Are you awaiting a review, do you need help somewhere  or are you still compiling on your local machine?

playcatchup:
@davidscn We (@jeremiahbintang ) are still trying to compile on our machine, and we will raise questions if we can't solve the errors from ctest as soon as possible.

playcatchup:
@davidscn By the way, our (@jeremiahbintang ) Advanced Programming bonus project requires us to make improvement of the performance on our project. We tried to do some refactoring on the BoundingBox class. We removed some loops and changed them to use the built-in functions of Eigen.

We did it in a separate branch which you can see [here](https://github.com/playcatchup/precice/tree/RefactorByEigen), especially on the third and second last commits: 
1. https://github.com/precice/precice/commit/93dd9a56774414dd178d07c693c955f09fe3faeb
2. https://github.com/precice/precice/commit/2ad139f0a2767103ae9a4f2e721da18227acf6f3

What do you think? Do you have any suggestions? Thank you very much😃

src/mesh/BoundingBox.cpp
```diff
@@ -65,39 +76,31 @@ bool BoundingBox::contains(const mesh::Vertex &vertex) const
 
 Eigen::VectorXd BoundingBox::center() const
 {
-  PRECICE_ASSERT(!empty(), "Data of the bounding box is at default state.");
+  PRECICE_ASSERT(!empty(), "The BoundingBox is empty, i.e. it has zero area or zero volume.");
   Eigen::VectorXd cog(_dimensions);
   for (int d = 0; d < _dimensions; d++) {
-    cog[d] = (_bounds[2 * d + 1] - _bounds[2 * d]) / 2.0 + _bounds[2 * d];
+    cog[d] = (_boundMax[d] + _boundMin[d]) / 2.0;
```
playcatchup, line 82: we could remove the loop since we have Eigen::VectorXd built-in function

src/mesh/BoundingBox.cpp
```diff
@@ -108,14 +111,21 @@ int BoundingBox::getDimension() const
...
+    bounds.insert(bounds.end(), {_boundMin[0], _boundMax[0], _boundMin[1], _boundMax[1], _boundMin[2], _boundMax[2]});
+  } else {
+    bounds.insert(bounds.end(), {_boundMin[0], _boundMax[0], _boundMin[1], _boundMax[1]});
+  }
+  return bounds;
 }
 
 void BoundingBox::expandBy(const BoundingBox &otherBB)
 {
+  PRECICE_ASSERT(_dimensions == otherBB.getDimension(), "Other BoundingBox with different dimensions than bounding box cannot be used to expand bounding box");
```
playcatchup, line 125:
```suggestion
  PRECICE_ASSERT(_dimensions == otherBB.getDimension(), "Other BoundingBox with different dimensions than this bounding box cannot be used to expand it");
```

src/mesh/BoundingBox.cpp
```diff
@@ -124,42 +134,40 @@ void BoundingBox::expandBy(const Vertex &vertices)
...
+    _boundMin[d] = std::min(coords[d], _boundMin[d]);
+    _boundMax[d] = std::max(coords[d], _boundMax[d]);
   }
 }
-
+// TODO: empty() definition
 void BoundingBox::expandBy(double value)
 {
-  if (empty())
-    return;
```
playcatchup, line 135: why return when it's empty?

src/mesh/tests/BoundingBoxTest.cpp
```diff
@@ -13,6 +13,29 @@ using namespace precice::mesh;
 BOOST_AUTO_TEST_SUITE(MeshTests)
 BOOST_AUTO_TEST_SUITE(BoundingBoxTests)
 
+BOOST_AUTO_TEST_CASE(Constructor)
+{
+  PRECICE_TEST(1_rank);
```
playcatchup, line 18: is it fine to executing this in 1_rank?

playcatchup (commented):
Review with @MakisH

MakisH:
We found the issues the tests were failing:
1. The definition of a default box was assumed as the opposite of what it was before: instead of an "illegal box", this PR was defining the biggest possible box.
2. In two places, there was an undocumented assumption that the method should directly return if a box is empty, instead of expanding. This PR was introducing assertions for this, which in principle makes sense. I suggest that we document that assumption in a comment.

@playcatchup @jeremiahbintang please document in more detail, resolve any conversation that is not relevant anymore, cleanup, and request again a review by me.

src/mesh/BoundingBox.cpp
playcatchup: Will be within the scope of sprint 3

playcatchup:
@MakisH We (@jeremiahbintang) have done all the requests you suggested. Would you be able to review again? We are now moving to the performance analysis. Thank you in advance😃

src/mesh/BoundingBox.cpp
```diff
@@ -13,50 +13,63 @@ namespace precice::mesh {
...
-  PRECICE_ASSERT(_dimensions == otherBB._dimensions, "Bounding boxes with different dimensions cannot be compared.", _dimensions, otherBB._dimensions);
-  for (int i = 0; i < _dimensions; ++i) {
-    if (_bounds.at(i) != otherBB._bounds.at(i)) {
-      return false;
-    }
-  }
-  return true;
+  PRECICE_ASSERT(_dimensions == otherBB._dimensions, "Bounding boxes ({} and {}) with different dimensions cannot be compared.", _dimensions, otherBB._dimensions);
+
+  return _boundMin.isApprox(otherBB._boundMin) && _boundMax.isApprox(otherBB._boundMax);
```
MakisH, line 54: In case anyone is interested in checking: https://eigen.tuxfamily.org/dox/classEigen_1_1DenseBase.html#ae8443357b808cd393be1b51974213f9c

src/mesh/BoundingBox.cpp
```diff
@@ -108,58 +113,68 @@ int BoundingBox::getDimension() const
 
 const std::vector<double> BoundingBox::dataVector() const
 {
-  return _bounds;
+  std::vector<double> bounds;
+  if (_dimensions == 3) {
+    bounds.insert(bounds.end(), {_boundMin[0], _boundMax[0], _boundMin[1], _boundMax[1], _boundMin[2], _boundMax[2]});
+  } else {
+    bounds.insert(bounds.end(), {_boundMin[0], _boundMax[0], _boundMin[1], _boundMax[1]});
+  }
```
MakisH, line 121: Would this work as well, to reduce duplication?

```suggestion
  bounds.insert(bounds.end(), {_boundMin[0], _boundMax[0], _boundMin[1], _boundMax[1]});
  if (_dimensions == 3) {
    bounds.insert(bounds.end(), {_boundMin[2], _boundMax[2]});
  }
```

(why not `.push_back`?)

src/mesh/BoundingBox.cpp
```diff
@@ -108,58 +113,68 @@ int BoundingBox::getDimension() const
...
-    _bounds.at(2 * d + 1) = std::max(coords[d], _bounds.at(2 * d + 1));
+    _boundMin[d] = std::min(coords[d], _boundMin[d]);
+    _boundMax[d] = std::max(coords[d], _boundMax[d]);
   }
 }
 
 void BoundingBox::expandBy(double value)
 {
-  if (empty())
+  // Adding or substracting a constant to the default state will still make an 'illegal' box
```
MakisH, line 146: But what is the reason this method should just return if the box is at its default state? I think what you did before with the assertion was better.

In any case, a typo:

```suggestion
  // Adding or subtracting a constant to the default state will still make an 'illegal' box
```

src/mesh/BoundingBox.hpp
```diff
@@ -45,9 +53,14 @@ class BoundingBox {
...
   bool operator==(const BoundingBox &otherBB) const;
 
-  /// Check whether the bounding box is at default state or not
+  /// empty() returning true means that the length in every BoundingBox's dimension is equal to zero
   bool empty() const;
 
+  /// Check whether the bounding box is at default state or not
+  /// all the values of _boundMin = std::numeric_limits<double>::max()
+  /// all the values of _boundMax = std::numeric_limits<double>::lowest()
+  bool isDefault() const;
```
MakisH, line 62: Maybe it would be useful to document (here or at least in the pull request conversation) what the purpose of the `empty` and of the `isDefault` is. At which situation is a box expected to be empty and in which default?

src/mesh/BoundingBox.hpp
```diff
@@ -95,11 +108,18 @@ class BoundingBox {
...
-  std::vector<double> _bounds;
+  /// Container of a Eigen::VectorXd of a bounding box with min components in each dimension
+  /// (x_min, y_min) when _dimensions=2;
+  /// (x_min, y_min, z_min) when _dimensions=3;
+  Eigen::VectorXd _boundMin;
+
+  /// Container of a Eigen::VectorXd of a bounding box with max components in each dimension
+  /// (x_max, y_max) when _dimensions=2;
+  /// (x_max, y_max, z_max) when _dimensions=3;
+  Eigen::VectorXd _boundMax;
```
MakisH, line 122: I guess here the important part is not that this is a container, but that it is a point/vertex. And we now have two points, instead of a strange `std::vector`.

Maybe (not sure about `Point A` and `Point B`):

```suggestion
  /// Point A defining the bounding box, with minimum coordinates in each direction
  /// (x_min, y_min) when _dimensions=2;
  /// (x_min, y_min, z_min) when _dimensions=3;
  Eigen::VectorXd _boundMin;

  /// Point B defining the bounding box, with maximum coordinates in each direction
  /// (x_max, y_max) when _dimensions=2;
  /// (x_max, y_max, z_max) when _dimensions=3;
  Eigen::VectorXd _boundMax;
```

src/mesh/Mesh.cpp
```diff
@@ -277,7 +277,7 @@ void Mesh::computeBoundingBox()
   PRECICE_TRACE(_name);
 
   // Keep the bounding box if set via the API function.
-  BoundingBox bb = _boundingBox.empty() ? BoundingBox(_dimensions) : BoundingBox(_boundingBox);
+  BoundingBox bb = _boundingBox.isDefault() ? BoundingBox(_dimensions) : BoundingBox(_boundingBox);
```
MakisH, line 280: I do not understand this change (it may be correct, it may be wrong). Could you please explain?

src/mesh/tests/BoundingBoxTest.cpp
```diff
@@ -13,6 +13,29 @@ using namespace precice::mesh;
...
+  { // 3D
+    Eigen::VectorXd boundMin(3);
+    boundMin << 1.0, 2.0, 3.0;
+    Eigen::VectorXd boundMax(3);
+    boundMax << 4.0, 5.0, 6.0;
+    BoundingBox bb(boundMin, boundMax);
+    BoundingBox compareData({1.0, 4.0, 2.0, 5.0, 3.0, 6.0});
```
MakisH, line 25: So, here you are testing the new constructor compared to the old constructor, right?
If we remove the old constructor, what happens here?

src/partition/ReceivedPartition.cpp
```diff
@@ -104,7 +104,7 @@ void ReceivedPartition::compute()
       // Filter out vertices not laying in the bounding box
       mesh::Mesh filteredMesh("FilteredMesh", _dimensions, mesh::Mesh::MESH_ID_UNDEFINED);
       // To discuss: maybe check this somewhere in the SolverInterfaceImpl, as we have now a similar check for the parallel case
-      PRECICE_CHECK(!_bb.empty(), "You are running this participant in serial mode and the bounding box on mesh \"{}\", is empty. Did you call setMeshAccessRegion with valid data?", _mesh->getName());
+      PRECICE_CHECK(!_bb.isDefault(), "You are running this participant in serial mode and the bounding box on mesh \"{}\", is empty. Did you call setMeshAccessRegion with valid data?", _mesh->getName());
```
MakisH, line 107: I do not understand this change, and it looks like the message is not consistent anymore. Could you please explain your motivation?

MakisH (commented):
I had another detailed look, and left some more comments. Please check, implement those you understand, and keep open the ones you don't.

Then, ask for a review by @davidscn, probably.

src/mesh/BoundingBox.cpp
```diff
@@ -13,50 +13,63 @@ namespace precice::mesh {
...
+
+  // PRECICE_ASSERT 2
+  PRECICE_ASSERT((boundMin - boundMax).maxCoeff() < 0, "Each component of min vertex ({}) must be <= max vertex ({}) in the same axis direction.", boundMin, boundMax);
+
+  _boundMin   = std::move(boundMin);
+  _boundMax   = std::move(boundMax);
+  _dimensions = _boundMin.rows();
+}
+
 BoundingBox::BoundingBox(std::vector<double> bounds)
```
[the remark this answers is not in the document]
playcatchup: It is still used in the existing codes. Do you think we should change all implementations now?

src/mesh/BoundingBox.cpp
```diff
@@ -13,50 +13,63 @@ namespace precice::mesh {
...
 {
-  PRECICE_ASSERT((int) bounds.size() == 4 || (int) bounds.size() == 6, "Dimension of a bounding box can only be 2 or 3.", bounds.size() / 2);
-  _bounds     = std::move(bounds);
-  _dimensions = _bounds.size() / 2;
+  PRECICE_ASSERT((int) bounds.size() == 4 || (int) bounds.size() == 6, "Dimension of a bounding box can only be 2 or 3. The dimension is ({}).", bounds.size() / 2);
```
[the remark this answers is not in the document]
playcatchup: `({})` are corrected now.

src/mesh/BoundingBox.cpp
```diff
@@ -13,50 +13,63 @@ namespace precice::mesh {
...
 
 BoundingBox::BoundingBox(int dimension)
     : _dimensions(dimension)
 {
-  PRECICE_ASSERT(dimension == 2 || dimension == 3, "Dimension of a bounding box can only be 2 or 3.", dimension);
-  for (int i = 0; i < _dimensions; ++i) {
-    _bounds.push_back(std::numeric_limits<double>::max());
-    _bounds.push_back(std::numeric_limits<double>::lowest());
-  }
+  PRECICE_ASSERT(dimension == 2 || dimension == 3, "Dimension of a bounding box ({}) can only be 2 or 3.", dimension);
```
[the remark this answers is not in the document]
playcatchup: Corrected.

src/mesh/BoundingBox.hpp
playcatchup: We change the comment to the following:
``` 
/// _boundMin defining the bounding box, with minimum coordinates in each direction
...
/// _boundMax defining the bounding box, with minimum coordinates in each direction
```

We use _boundMin and _boundMax directly describe the properties of the variables.

src/mesh/tests/BoundingBoxTest.cpp
playcatchup: The purpose of the test is to make sure that the new BB constructor returns the correct values. But we get what you mean and we think we should remove the old constructor and just check if the new constructor works as expected.

```cpp
BOOST_AUTO_TEST_CASE(Constructor)
{
  PRECICE_TEST(1_rank);
  { // 3D
    Eigen::VectorXd boundMin(3);
    boundMin << 1.0, 2.0, 3.0;
    Eigen::VectorXd boundMax(3);
    boundMax << 4.0, 5.0, 6.0;
    BoundingBox bb(boundMin, boundMax);

    BOOST_TEST(bb.getDimension() == 3);
    BOOST_TEST(bb.minCorner() == boundMin);
    BOOST_TEST(bb.maxCorner() == boundMax);
  }
  { // 2D
    Eigen::VectorXd boundMin(2);
    boundMin << 1.0, 2.0;
    Eigen::VectorXd boundMax(2);
    boundMax << 4.0, 5.0;
    BoundingBox bb(boundMin, boundMax);
    
    BOOST_TEST(bb.getDimension() == 2);
    BOOST_TEST(bb.minCorner() == boundMin);
    BOOST_TEST(bb.maxCorner() == boundMax);
  }
} // Constructor
```

src/mesh/BoundingBox.cpp
playcatchup: Agree

src/mesh/BoundingBox.cpp
playcatchup: Correctec

src/mesh/BoundingBox.cpp
```diff
@@ -13,50 +13,63 @@ namespace precice::mesh {
...
+BoundingBox::BoundingBox(Eigen::VectorXd boundMin, Eigen::VectorXd boundMax)
+{
+  // PRECICE_ASSERT 1
+  PRECICE_ASSERT(boundMin.rows() == boundMax.rows(), "Dimension of min ({}) and max ({}) vertices should be the same.", boundMin.rows(), boundMax.rows());
+
+  // PRECICE_ASSERT 2
+  PRECICE_ASSERT((boundMin - boundMax).maxCoeff() < 0, "Each component of min vertex ({}) must be <= max vertex ({}) in the same axis direction.", boundMin, boundMax);
+
+  _boundMin   = std::move(boundMin);
+  _boundMax   = std::move(boundMax);
```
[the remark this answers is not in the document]
davidscn: Is this a performance comment? Pass-by.reference doesn't make sense here, as we anyway want to store the data in the class. If we pass an rvalue into the constructor, we should be able to directly move it into the class. If we pass an lvalue, we make a copy. If we pass-by-reference, we would make a copy in any case.

src/mesh/BoundingBox.cpp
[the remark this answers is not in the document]
davidscn: The problem is a bit that we use this constructor in the API `setMeshAccessRegion`. If we remove it here, we probably have to copy the code over to other places. I would try to use the new constructor wherever it simplifies the code, but keep the constructor here.

src/mesh/BoundingBox.cpp
```diff
@@ -108,58 +105,67 @@ int BoundingBox::getDimension() const
 
 const std::vector<double> BoundingBox::dataVector() const
 {
-  return _bounds;
+  std::vector<double> bounds;
+  bounds.insert(bounds.end(), {_boundMin[0], _boundMax[0], _boundMin[1], _boundMax[1]});
```
davidscn, line 109: I guess you can merge these two lines as the vector takes an initializer list anyway.

src/mesh/BoundingBox.cpp
```diff
@@ -108,58 +105,67 @@ int BoundingBox::getDimension() const
...
   }
 }
 
 void BoundingBox::expandBy(double value)
 {
-  if (empty())
+  // Adding or subtracting a constant to the default state will still make an 'illegal' box
+  if (isDefault()) {
     return;
+  }
```
davidscn, line 140: Could you add this behavior to the function description? I'm still a bit unhappy that this function overloads with the previous two, but still behaves differently (in this way).

Also, you can just wrap the code into the if block and remove the return, as it is a void function.

src/mesh/BoundingBox.cpp
```diff
@@ -108,58 +105,67 @@ int BoundingBox::getDimension() const
...
-  if (empty())
+  if (isDefault()) {
     return;
+  }
   double maxSideLength = 1e-6; // we need some minimum > 0 here
   for (int d = 0; d < _dimensions; d++) {
-    if (_bounds.at(2 * d + 1) > _bounds.at(2 * d))
-      maxSideLength = std::max(maxSideLength, _bounds[2 * d + 1] - _bounds[2 * d]);
+    if (_boundMax[d] > _boundMin[d])
+      maxSideLength = std::max(maxSideLength, _boundMax[d] - _boundMin[d]);
```
davidscn, line 155: We could use Eigen functions here, right?

src/mesh/Mesh.cpp
davidscn: The code change itself is only a renaming of the function. We have to preserve the bounding box in case there is reasonable data in there.

src/mesh/tests/BoundingBoxTest.cpp
davidscn: Yes

src/partition/ReceivedPartition.cpp
davidscn: Maybe it would be helpful to still keep the `empty` function in order to check whether the BB is actually empty or not (which is still different from default).

src/mesh/BoundingBox.cpp
davidscn: What happens here if we have single axis being in the default state? Then we would have negative entries in the result and `isZero` would return `false`, which is not what we are looking for, right?

>  PRECICE_ASSERT(!empty(), "The BoundingBox is empty, i.e. it has zero area or zero volume.")

You can add it whenever you think it makes sense, will have a look in the final review.

src/mesh/BoundingBox.cpp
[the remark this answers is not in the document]
MakisH: > Pass-by.reference doesn't make sense here, as we anyway want to store the data in the class.

Can't we do something like this, to avoid copying? I am not sure if it would even compile.

```c++
BoundingBox::BoundingBox(Eigen::VectorXd& boundMin, Eigen::VectorXd& boundMax)
:
_boundMin(std::move(boundMin)), _boundMax(std::move(boundMax))
{
// ...
}
```

But anyway, performance-wise it does not make any sense (this is happening only once and the vectors are trivially small). My main point was that I did not really like stealing the resource in the middle of the body. I would prefer if such an operation was in the initializer list (and generally trying to stick to prefer initializers, as e.g., https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rc-initialize suggests)

You can safely ignore my comment.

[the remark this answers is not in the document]
davidscn: You could, but the user (we) would be very surprised if the vectors are all of a sudden empty when you construct a bounding box out of them.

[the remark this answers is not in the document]
davidscn: But of course we can first put the arguments in the initializer list and still keep the PRECICE_ASSERTs afterwards.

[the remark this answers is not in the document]
MakisH: Isn't the same happening already? That is also bad, but I assumed it was ok so far.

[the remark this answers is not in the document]
davidscn: > Isn't the same happening already? That is also bad, but I assumed it was ok so far.

No, there are no references involved in the constructor right now, i.e., we first copy the data into the local scope and afterwards move it to the class members. If you pass arguments as an rvalue, the compiler is (usually) able to get rid of the unnecessary copy (but only when it is an rvalue).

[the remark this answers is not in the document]
MakisH: Ah, right. Just disregard my comment, the current approach is better.

src/mesh/BoundingBox.cpp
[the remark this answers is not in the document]
playcatchup: We'll (@jeremiahbintang ) keep the old constructor.

src/mesh/BoundingBox.cpp
playcatchup: We (@jeremiahbintang ) merge these two lines.
```
std::vector<double> bounds({_boundMin[0], _boundMax[0], _boundMin[1], _boundMax[1]});
```

src/mesh/BoundingBox.cpp
playcatchup: We (@jeremiahbintang ) decide to use `insert` instead of `push_back` since it's faster to add elements for `vector` (https://quick-bench.com/q/vqA7scnhikRpd8aHdzy_NtqJ3ig).

src/mesh/BoundingBox.cpp
playcatchup: We add this behavior inside this function as a comment.
```cpp
void BoundingBox::expandBy(double value)
{
  // BoundingBox should not be default state; otherwise, it will make ReceivedPartitionTest fail because of incorrect connectionMapSize
  if (!isDefault()) {
    for (int d = 0; d < _dimensions; d++) {
      _boundMin[d] -= value;
      _boundMax[d] += value;
    }
  }
}
```
Error message of `ReceivedPartitionTest` fail because of incorrect `connectionMapSize`:
```
gney@Hsiao:~/projects/advprog_bp/precice/build$ ctest --rerun-failed --output-on-failure
Test project /home/gney/projects/advprog_bp/precice/build
    Start 10: precice.partition
1/1 Test #10: precice.partition ................***Timeout  20.02 sec
[2023-01-29 18:42:33.002000] [0x00007f04d711b780] [debug]   Initialize MPI
[2023-01-29 18:42:33.001894] [0x00007f98e492a780] [debug]   Initialize MPI
[2023-01-29 18:42:33.002640] [0x00007f0e91707780] [debug]   Initialize MPI
[2023-01-29 18:42:33.005487] [0x00007f9353450780] [debug]   Initialize MPI
This test suite runs on rank 1 of 4
This test suite runs on rank 2 of 4
This test suite runs on rank 0 of 4
This test suite runs on rank 3 of 4
Running 20 test cases...
Running 20 test cases...
Running 20 test cases...
Setup up logging
Setup up logging
Setup up logging
Running 20 test cases...
Setup up logging
18:42:33.586897|Fluid|0|partition::ReceivedPartition|l891|createOwnerInformation|WARNING: 2 of 6 vertices of mesh SolidzMesh have been filtered out since they have no influence on the mapping.
18:42:33.638039|Fluid|0|partition::ReceivedPartition|l891|createOwnerInformation|WARNING: 1 of 3 vertices of mesh NastinMesh have been filtered out since they have no influence on the mapping.
/home/gney/projects/advprog_bp/precice/src/partition/tests/ReceivedPartitionTest.cpp(949): fatal error: in "PartitionTests/ReceivedPartitionTests/TestCompareBoundingBoxes2D": critical check connectionMapSize == 2 has failed [3 != 2]
Failure occurred in a following context:
    Test context represents "SOLIDZ" and runs on rank 0 out of 1. Initialized: { Events}
/home/gney/projects/advprog_bp/precice/src/partition/tests/ReceivedPartitionTest.cpp(1017): error: in "PartitionTests/ReceivedPartitionTests/TestCompareBoundingBoxes3D": check connectionMapSize == 2 has failed [3 != 2]
Failure occurred in a following context:
    Test context represents "SOLIDZ" and runs on rank 0 out of 1. Initialized: { Events}
18:42:33.764675||3|partition::ReceivedPartition|l745|createOwnerInformation|WARNING: 1 of 6 vertices of mesh mesh have been filtered out since they have no influence on the mapping.
18:42:33.764673||0|partition::ReceivedPartition|l745|createOwnerInformation|WARNING: 1 of 6 vertices of mesh mesh have been filtered out since they have no influence on the mapping.
18:42:33.764959||1|partition::ReceivedPartition|l745|createOwnerInformation|WARNING: 1 of 6 vertices of mesh mesh have been filtered out since they have no influence on the mapping.
18:42:33.771773||1|partition::ReceivedPartition|l745|createOwnerInformation|WARNING: 2 of 12 vertices of mesh mesh have been filtered out since they have no influence on the mapping.
18:42:33.771742||3|partition::ReceivedPartition|l745|createOwnerInformation|WARNING: 2 of 12 vertices of mesh mesh have been filtered out since they have no influence on the mapping.
18:42:33.771858||2|partition::ReceivedPartition|l745|createOwnerInformation|WARNING: 2 of 12 vertices of mesh mesh have been filtered out since they have no influence on the mapping.


0% tests passed, 1 tests failed out of 1

Total Test time (real) =  20.02 sec

The following tests FAILED:
         10 - precice.partition (Timeout)
Errors while running CTest
```

src/mesh/BoundingBox.cpp
playcatchup: Yes.
``` cpp
maxSideLength        = (_boundMax - _boundMin).maxCoeff();
```
On the other hand, what's `PRECICE_DEBUG("Merged BoundingBox {}", *this);` for in this function? We would like to also use Eigen functions to eliminate loop in the following code, but it looks like the above debug command is used as long as _boundMax and _boundMin are changed.
```cpp
  for (int d = 0; d < _dimensions; d++) {
      _boundMax[d] += safetyFactor * maxSideLength;
      _boundMin[d] -= safetyFactor * maxSideLength;
      PRECICE_DEBUG("Merged BoundingBox {}", *this);
    }
```

src/partition/ReceivedPartition.cpp
playcatchup: @davidscn Would you be able answer that
1. does the `empty` function here mean "every dimension of the `BoundingBox` is zero"?
2. why should it be changed? The original `empty` function is defined like `isDefault` function.

src/mesh/BoundingBox.cpp
davidscn: > We (@jeremiahbintang) add this behavior inside this function as a comment.

There is no real value in adding the docs to the function implementation. We generate code documentation based on doxygen, which usually looks at the function declaration (in the header file). Please at the documentation there such that we don't have to look for the implementation each time.

src/mesh/BoundingBox.cpp
davidscn: `PRECICE_DEBUG` is a logging macro for additional information when debugging code. If you can get rid of the loop (i.e. move the loop into eigen), there is no need to have this information dimension-wise (it's anyway a bit duplicated).
You can safely move it behind the loop.

src/partition/ReceivedPartition.cpp
davidscn: Problem is that the code following this check only works if the bounding box is really non-empty. Example: you call `setMeshAccessRegion`, but only define a single vertex. This will make the box non-default, but the check/code should should stop at this point, because the whole mesh will be filtered afterwards.
How to define empty is a good question. In principle, I guess looking for all dimensions of the BB are zero would be appropriate. I'm not sure if this could lead to issues when running a multi-dimensional coupling and the interface might indeed be 0D. @MakisH any opinions?

MakisH: >  In principle, I guess looking for all dimensions of the BB are zero would be appropriate. I'm not sure if this could lead to issues when running a multi-dimensional coupling and the interface might indeed be 0D. @MakisH any opinions?

I would not worry about issues with mixed dimensions in this PR, and definitely not about 0D. We will revisit the bounding boxes later, if needed.

Keeping both `empty()` and `isDefault()` makes sense to me, I understand that they are different concepts. The `isDefault()` essentially means "I am not initialized - don't use me".

In a 0D simulation, we would probably still have one mesh vertex. The area would be zero, but the box would still have one point.

src/mesh/BoundingBox.cpp
playcatchup: Thank you. We've done it.
```cpp
void BoundingBox::scaleBy(double safetyFactor)
{
  if (!isDefault()) {
    double maxSideLength = 1e-6; // we need some minimum > 0 here
    maxSideLength        = (_boundMax - _boundMin).maxCoeff();
    _boundMax.array() += safetyFactor * maxSideLength;
    _boundMin.array() -= safetyFactor * maxSideLength;
    PRECICE_DEBUG("Merged BoundingBox {}", *this);
  }
}
```

src/mesh/BoundingBox.cpp
playcatchup: We (@jeremiahbintang ) have added the description in the hpp file.
```cpp
/// Expand bounding box using double value
/// Using this method, BoundingBox should not be in the default state; otherwise, it will make ReceivedPartitionTest fail because of incorrect connectionMapSize
  void expandBy(double value);
```

src/partition/ReceivedPartition.cpp
playcatchup: We've (@jeremiahbintang ) added the `empty` method back.
```cpp
bool BoundingBox::empty() const
{
  return (_boundMax - _boundMin).isZero();
}
```
src/partition/ReceivedPartition.cpp
```cpp
PRECICE_CHECK(!_bb.empty(), "You are running this participant in serial mode and the bounding box on mesh \"{}\", is empty. Did you call setMeshAccessRegion with valid data?", _mesh->getName());
```

src/mesh/BoundingBox.cpp
MakisH: > Expand bounding box using double value

To which direction / how?

> otherwise, it will make ReceivedPartitionTest fail because of incorrect connectionMapSize

I am not sure if referring to the test is a good idea here. It is not that the function is made for the test, but the test for the function.

But I think you probably just don't have enough context to write a more detailed comment about the motivation (neither do I :grimacing: ).

src/mesh/BoundingBox.cpp
```diff
@@ -13,50 +13,60 @@ namespace precice::mesh {
...
+
 BoundingBox::BoundingBox(std::vector<double> bounds)
 {
-  PRECICE_ASSERT((int) bounds.size() == 4 || (int) bounds.size() == 6, "Dimension of a bounding box can only be 2 or 3.", bounds.size() / 2);
-  _bounds     = std::move(bounds);
-  _dimensions = _bounds.size() / 2;
+  PRECICE_ASSERT((int) bounds.size() == 4 || (int) bounds.size() == 6, "Dimension of a bounding box can only be 2 or 3. The dimension is {}.", bounds.size() / 2);
+  _dimensions = bounds.size() / 2;
+  _boundMin   = Eigen::Map<Eigen::VectorXd, 0, Eigen::InnerStride<2>>(bounds.data(), _dimensions);
+  _boundMax   = Eigen::Map<Eigen::VectorXd, 0, Eigen::InnerStride<2>>(bounds.data() + 1, _dimensions);
```
davidscn, line 31: Can you code comment what these lines do? Where is the `1` coming from? Is it really valid for all dimensions?

src/mesh/BoundingBox.cpp
```diff
@@ -13,50 +13,60 @@ namespace precice::mesh {
...
-    _bounds.push_back(std::numeric_limits<double>::lowest());
-  }
+  PRECICE_ASSERT(dimension == 2 || dimension == 3, "Dimension of a bounding box can only be 2 or 3 (in this case, it is {}).", dimension);
+
+  _boundMin = Eigen::VectorXd(dimension);
+  _boundMax = Eigen::VectorXd(dimension);
+
+  // Define 'illegal' BoundingBox: _boundMin > _boundMax
+  std::fill(_boundMin.data(), _boundMin.data() + _boundMin.rows(), std::numeric_limits<double>::max());
+  std::fill(_boundMax.data(), _boundMax.data() + _boundMax.rows(), std::numeric_limits<double>::lowest());
```
davidscn, line 44: Wouldn't it be more natural to use `begin()` and `end()` instead of `data()` + ...

src/mesh/BoundingBox.cpp
```diff
@@ -13,50 +13,60 @@ namespace precice::mesh {
...
 
 bool BoundingBox::empty() const
 {
-  for (int i = 0; i < _dimensions; ++i) {
-    if (_bounds[2 * i] > _bounds[2 * i + 1]) {
-      return true;
-    }
-  }
-  return false;
+  return (_boundMax - _boundMin).isZero();
```
davidscn, line 56: What would happen here if the BB is default, i.e., `min>max`?

src/mesh/BoundingBox.cpp
```diff
@@ -108,58 +110,58 @@ int BoundingBox::getDimension() const
...
 }
 
 void BoundingBox::expandBy(double value)
 {
-  if (empty())
-    return;
-  for (int d = 0; d < _dimensions; d++) {
-    _bounds[2 * d] -= value;
-    _bounds[2 * d + 1] += value;
+  if (!isDefault()) {
```
davidscn, line 141: What happens if we use here an `PRECICE_ASSERT` instead of a conditional action?

src/mesh/BoundingBox.cpp
jeremiahbintang: Hi David, thanks for the catch. We decided to make the method to return true if it is default

src/mesh/BoundingBox.cpp
playcatchup: 1. Yes. One question please: should I code the comment also in hpp file or in the cpp file?

2. `1` is to move the pointer `bounds.array()` to point at its second element.

3. Yes, because the dimensions are only 2 and 3. This reminds us to add an assert in the constructor `BoundingBox(Eigen::VectorXd boundMin, Eigen::VectorXd boundMax)` to ensure `boundMin` and `boundMax` to have the same dimensions of 2 or 3.
```cpp
PRECICE_ASSERT((boundMin.rows() == 2 && boundMax.rows() == 2) || (boundMin.rows() == 3 && boundMax.rows() == 3),
                 "Dimension of min {} and max {} vertices should be the same and both 2 or 3.", boundMin.rows(), boundMax.rows());
```


- [Note for 2.](https://eigen.tuxfamily.org/dox/classEigen_1_1Map.html)
`Eigen::Map` maps an existing array to a Eigen's matrix or vector. In this case, its first input is the existing array and its second input is the size of output `Eigen::VectorXd`. The existing array is from the vector `bounds` by using `bounds.array()` to return its array pointer. The `bounds.array()+1` means that the elements of output `Eigen::VectorXd` starts from the second element of the existing array `bounds.array()`.

davidscn: Here, directly in the cpp file, thanks!

src/mesh/BoundingBox.cpp
jeremiahbintang: Thank you for the catch. We have changed it

src/mesh/BoundingBox.cpp
jeremiahbintang: There is some logic in the testing scheme that we haven't got the time to look into it deeply, sorry for that. But the testing fails when we remove the conditional with `PRECICE_ASSERT`.

src/mesh/BoundingBox.cpp
playcatchup: Yes. Thanks for your catch! (This was due to the VSCode include path setting makes `begin()` and `end()` undetectable😅 )
```cpp
 std::fill(_boundMin.begin(), _boundMin.end(), std::numeric_limits<double>::max());
 std::fill(_boundMax.begin(), _boundMax.end(), std::numeric_limits<double>::lowest());
```

src/mesh/BoundingBox.cpp
playcatchup: We (@jeremiahbintang ) modify it as follows:
```cpp
return (_boundMax - _boundMin).isZero() || isDefault();
```

src/mesh/BoundingBox.cpp
fsimonis: Eigen also provides a [`fill()`](https://eigen.tuxfamily.org/dox-3.3/classEigen_1_1DenseBase.html#a9be169c308801411aa24be93d30930bf) :wink: : `_boundMin.fill(std::numeric_limits<double>::max());`

playcatchup: @fsimonis Thanks you!

src/mesh/BoundingBox.cpp
playcatchup: 1. 'expandBy(double value)' Expand bounding box using a double value in all dimensions.
2. Yes, we still don't know why it fails the test😅 The reason why we wrote it down is to record the behavior clearly for other developers' reference.

src/mesh/BoundingBox.cpp
playcatchup: Done.

src/mesh/BoundingBox.cpp
playcatchup: Done.

src/mesh/BoundingBox.hpp
```diff
@@ -36,6 +36,14 @@ class BoundingBox {
   */
   explicit BoundingBox(std::vector<double> bounds);
 
+  /**
+  * @brief Construct a new Bounding Box:: Bounding Box object
+  *
+  * @param boundMin
+  * @param boundMax
```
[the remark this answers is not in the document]
playcatchup: Done.

src/mesh/BoundingBox.cpp
playcatchup: Done.

playcatchup:
Dear @davidscn, we (@jeremiahbintang ) have updated the code as per all the above comments.  Would you be able to do a code review and then consider accepting our pull request? Thank you in advance :)

davidscn:
Could you add a changelog?

playcatchup:
@davidscn Sure. Could you please tell us how to do that?

davidscn (approved):
Thanks a lot for your efforts!

playcatchup:
@davidscn @fsimonis @MakisH Thank you for guiding us (@jeremiahbintang ) generously! I'm really lucky to have you in my first FOSS contribution :)

MakisH:
:tada: Thank you both very much for contributing!

PR #1499, opened 2022-12-01
Remove the ability to install tests

fsimonis:
## Main changes of this PR

This PR removes the ability to install tests.

The feature bloats the CMakeLists.txt of preCICE without adding value.

I will leave this PR open until we agree on this.

## Author's checklist

## Reviewers' checklist


uekerman:
What was the motivation for this in the first place? Can indeed go in my opinion.

fsimonis:
I can vaguely remember that we intended  these tests to be used by the user as smoke tests. But it turned out to be a pain to use, so we decided to make it optional

PR #1502, opened 2022-12-02
Add framework for benchmarks

fsimonis:
## Main changes of this PR

This PR adds a framework for implementing benchmarks using the popular [benchmark](https://github.com/google/benchmark).
As this introduces another unnecessary dependency, the generation of the benchmark suite is disabled by default.

`cmake -DPRECICE_ENABLE_BENCHMARKS=ON ..` enables the generation of the benchmark executable `benchprecice`.

Add your benchmark as a `.cpp` file in the `benchmarks/` directory and run `tools/building/updateSourceFiles.py`.

This is a first step towards performance regression testing.

There are currently no benchmarks.

## Author's checklist

## Reviewers' checklist


fsimonis:
Open question: should we build these in the CI pipeline as well?

CMakeLists.txt
```diff
@@ -439,6 +439,33 @@ if (PRECICE_ENABLE_FORTRAN)
   # include((${CMAKE_CURRENT_LIST_DIR}/extras/bindings/fortran/CMakeLists.txt)
   add_subdirectory(${CMAKE_CURRENT_LIST_DIR}/extras/bindings/fortran)
 endif()
+
+# Build benchmarks if enabled
+option(PRECICE_ENABLE_BENCHMARKS "Build the preCICE benchmarks" OFF)
+if(PRECICE_ENABLE_BENCHMARKS)
+  find_package(benchmark REQUIRED)
+
+  add_executable(benchprecice)
```
davidscn, line 448: In accordance to other executables around preCICE, I would call this `precice-bench`.  I would even consider merging it optionally with the `precice-tools`

davidscn (commented):
For now, I only benchmarked 'non-coupled' scenarios. I'm not sure if we can benchmark coupled API functions at all.

CMakeLists.txt
fsimonis: Good point. Renaming `testprecice` would also make sense.

davidscn:
> Open question: should we build these in the CI pipeline as well?

Experience states that it will become outdated otherwise.

fsimonis:
> For now, I only benchmarked 'non-coupled' scenarios. I'm not sure if we can benchmark coupled API functions at all.

Right. We will be able to benchmark internals, such as Mappings, Meshes, Exporters as well as everything SolverInterface-related until the call to initialize.

fsimonis:
I'll unassign myself as I don't have plans to push this further. I keep it open as 
@davidscn may still need it.

fsimonis:
I rebased this PR and extended it to

- work with MPI (1 rank only) so we can benchmark everything in Participant before calling initialize.
- ported test to v3
- extended write mapping tests to run for ranges of inputs with complexity estimation

I investigated how to vendor benchmark as a third party library, but this gets ugly pretty quickly. It's easier to rely on system packages.

CMakeLists.txt
fsimonis: Done. Renamed it to `precice-bench`

CMakeLists.txt
```diff
@@ -620,6 +620,33 @@ if (PRECICE_BINDINGS_FORTRAN)
   # include((${CMAKE_CURRENT_LIST_DIR}/extras/bindings/fortran/CMakeLists.txt)
   add_subdirectory(${CMAKE_CURRENT_LIST_DIR}/extras/bindings/fortran)
 endif()
+
+# Build benchmarks if enabled
+option(PRECICE_ENABLE_BENCHMARKS "Build the preCICE benchmarks" OFF)
```
davidscn, line 625:
```suggestion
option(PRECICE_ENABLE_BENCHMARKS "Build preCICE-internal benchmarks" OFF)
```

davidscn (approved):
Looks good. I would suggest to add at least documentation to all the benchmarks we add to the repository in terms of 

- what do they benchmark
- how does it benchmark

Otherwise, it seems unnecessary complicated to figure this out. Furthermore, at least a note about the existence of this feature in the dev docs seems necessary.

Regarding the CI: I would enable the benchmarks at least for one of our configurations, such that changes are correspondingly updated.

CMakeLists.txt
fsimonis: Done

fsimonis:
* Gave the tests better names and documented them
* Renamed the CMake option to PRECICE_BUILD_BENCHMARKS
* The benchmarks run in archlinux + release (fastest full release builds)

I'll add developer docs after merging

PR #1504, opened 2022-12-04
WIP: Remove receiveResultOfFirstAdvance

BenjaminRodenberg:
## Main changes of this PR

Removes `CouplingScheme::receiveResultOfFirstAdvance` and simplifies usage of `CouplingScheme`

simplifies implementation in #1414 and #1462 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -253,35 +253,34 @@ add_compile_definitions("$<$<CONFIG:DEBUG>:_GLIBCXX_ASSERTIONS>")
 print_empty()
 print_section("TARGETS & PACKAGES")
 
-# Core of preCICE bundling the sources and dependencies.
-# The main library and tests link against this.
-add_library(preciceCore OBJECT)
-set_target_properties(preciceCore PROPERTIES
+# Add precice as an empty target
+add_library(precice)
+set_target_properties(precice PROPERTIES
+  # precice is a C++17 project
   CXX_STANDARD 17
   CXX_STANDARD_REQUIRED Yes
   CXX_EXTENSIONS No
-  POSITION_INDEPENDENT_CODE YES
-  CXX_VISIBILITY_PRESET hidden
-  VISIBILITY_INLINES_HIDDEN YES
+  VERSION ${preCICE_VERSION}
+  SOVERSION ${preCICE_SOVERSION}
   )
 
 # Setup release override options
 if(PRECICE_RELEASE_WITH_DEBUG_LOG)
-  target_compile_definitions(preciceCore PUBLIC PRECICE_RELEASE_WITH_DEBUG_LOG)
+  target_compile_definitions(precice PRIVATE PRECICE_RELEASE_WITH_DEBUG_LOG)
 endif()
 
 if(PRECICE_RELEASE_WITH_TRACE_LOG)
-  target_compile_definitions(preciceCore PUBLIC PRECICE_RELEASE_WITH_TRACE_LOG)
+  target_compile_definitions(precice PRIVATE PRECICE_RELEASE_WITH_TRACE_LOG)
 endif()
 
 if(PRECICE_RELEASE_WITH_ASSERTIONS)
-  target_compile_definitions(preciceCore PUBLIC PRECICE_RELEASE_WITH_ASSERTIONS)
+  target_compile_definitions(precice PRIVATE PRECICE_RELEASE_WITH_ASSERTIONS)
 endif()
 
 
 # Setup Boost
-target_compile_definitions(preciceCore PUBLIC BOOST_ALL_DYN_LINK BOOST_ASIO_ENABLE_OLD_SERVICES BOOST_GEOMETRY_DISABLE_DEPRECATED_03_WARNING)
-target_link_libraries(preciceCore PUBLIC
+target_compile_definitions(precice PRIVATE BOOST_ALL_DYN_LINK BOOST_ASIO_ENABLE_OLD_SERVICES BOOST_GEOMETRY_DISABLE_DEPRECATED_03_WARNING)
+target_link_libraries(precice PRIVATE
   Boost::boost
   Boost::filesystem
   Boost::log
@@ -292,68 +291,42 @@ target_link_libraries(preciceCore PUBLIC
   Boost::unit_test_framework
   )
 if(UNIX OR APPLE OR MINGW)
-  target_compile_definitions(preciceCore PUBLIC _GNU_SOURCE)
-  target_link_libraries(preciceCore PUBLIC ${CMAKE_DL_LIBS})
+  target_compile_definitions(precice PRIVATE _GNU_SOURCE)
+  target_link_libraries(precice PRIVATE ${CMAKE_DL_LIBS})
 endif()
 
 # Setup Eigen3
-target_link_libraries(preciceCore PUBLIC Eigen3::Eigen)
-target_compile_definitions(preciceCore PUBLIC "$<$<CONFIG:DEBUG>:EIGEN_INITIALIZE_MATRICES_BY_NAN>")
+target_link_libraries(precice PRIVATE Eigen3::Eigen)
+target_compile_definitions(precice PRIVATE "$<$<CONFIG:DEBUG>:EIGEN_INITIALIZE_MATRICES_BY_NAN>")
 
 # Setup LIBXML2
-target_link_libraries(preciceCore PUBLIC LibXml2::LibXml2)
+target_link_libraries(precice PRIVATE LibXml2::LibXml2)
 
 # Setup JSON
-target_link_libraries(preciceCore PUBLIC JSON fmt-header-only)
+target_link_libraries(precice PRIVATE JSON fmt-header-only)
 
 # Setup MPI
 if (PRECICE_MPICommunication)
-  target_link_libraries(preciceCore PUBLIC MPI::MPI_CXX)
+  target_link_libraries(precice PRIVATE MPI::MPI_CXX)
 else()
-  target_compile_definitions(preciceCore PUBLIC PRECICE_NO_MPI)
+  target_compile_definitions(precice PRIVATE PRECICE_NO_MPI)
 endif()
 
 # Setup PETSC
 if (PRECICE_PETScMapping AND PRECICE_MPICommunication)
-  target_link_libraries(preciceCore PUBLIC PETSc::PETSc)
+  target_link_libraries(precice PRIVATE PETSc::PETSc)
 else()
-  target_compile_definitions(preciceCore PUBLIC PRECICE_NO_PETSC)
+  target_compile_definitions(precice PRIVATE PRECICE_NO_PETSC)
 endif()
 
 # Option Python
 if (PRECICE_PythonActions)
-  target_link_libraries(preciceCore PUBLIC Python3::NumPy Python3::Python)
-  target_compile_definitions(preciceCore PUBLIC NPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION)
+  target_link_libraries(precice PRIVATE Python3::NumPy Python3::Python)
+  target_compile_definitions(precice PRIVATE NPY_NO_DEPRECATED_API=NPY_1_7_API_VERSION)
 else()
-  target_compile_definitions(preciceCore PUBLIC PRECICE_NO_PYTHON)
+  target_compile_definitions(precice PRIVATE PRECICE_NO_PYTHON)
 endif()
 
-# Includes configuration for the core
-target_include_directories(preciceCore PUBLIC
-  $<BUILD_INTERFACE:${preCICE_SOURCE_DIR}/src>
-  $<BUILD_INTERFACE:${preCICE_BINARY_DIR}/src>
-  $<INSTALL_INTERFACE:include>
-  )
-
-# Add precice as an empty target
-add_library(precice)
-set_target_properties(precice PROPERTIES
-  VERSION ${preCICE_VERSION}
-  SOVERSION ${preCICE_SOVERSION}
-  CXX_VISIBILITY_PRESET hidden
-  VISIBILITY_INLINES_HIDDEN YES
-  )
-
-target_include_directories(precice PUBLIC
-  $<BUILD_INTERFACE:${preCICE_SOURCE_DIR}/src>
-  $<BUILD_INTERFACE:${preCICE_BINARY_DIR}/src>
-  $<INSTALL_INTERFACE:include>
-  )
-
-target_link_libraries(precice PRIVATE preciceCore)
-
-# Sources Configuration
-include(${CMAKE_CURRENT_LIST_DIR}/src/sources.cmake)
 
 # File Configuration
 include(GenerateVersionInformation)
@@ -361,18 +334,16 @@ include(${CMAKE_CURRENT_LIST_DIR}/cmake/DetectGitRevision.cmake)
 configure_file("${PROJECT_SOURCE_DIR}/src/precice/impl/versions.hpp.in" "${PROJECT_BINARY_DIR}/src/precice/impl/versions.hpp" @ONLY)
 configure_file("${PROJECT_SOURCE_DIR}/src/precice/Version.h.in" "${PROJECT_BINARY_DIR}/src/precice/Version.h" @ONLY)
 
-# Setup export header
-include(GenerateExportHeader)
-GENERATE_EXPORT_HEADER(preciceCore
-  EXPORT_FILE_NAME "src/precice/export.h"
-  BASE_NAME PRECICE
-  EXPORT_MACRO_NAME PRECICE_API
-  INCLUDE_GUARD_NAME PRECICE_EXPORT
+# Includes Configuration
+target_include_directories(precice PUBLIC
+  $<BUILD_INTERFACE:${preCICE_SOURCE_DIR}/src>
+  $<BUILD_INTERFACE:${preCICE_BINARY_DIR}/src>
+  $<INSTALL_INTERFACE:include>
   )
 
-# Add the export header to the public headers of the preCICE lib
-set_property(TARGET precice PROPERTY PUBLIC_HEADER
-    ${CMAKE_BINARY_DIR}/src/precice/export.h)
+# Sources Configuration
+include(${CMAKE_CURRENT_LIST_DIR}/src/sources.cmake)
+
 
 #
 # Configuration of Target precice-tools
@@ -408,9 +379,21 @@ IF (BUILD_TESTING)
   add_executable(testprecice "src/testing/main.cpp")
   target_link_libraries(testprecice
     PRIVATE
-    preciceCore
+    Threads::Threads
+    precice
+    Eigen3::Eigen
+    fmt-header-only
+    Boost::boost
+    Boost::filesystem
+    Boost::log
+    Boost::log_setup
+    Boost::program_options
+    Boost::system
+    Boost::thread
+    Boost::unit_test_framework
     )
   set_target_properties(testprecice PROPERTIES
+    # precice is a C++17 project
     CXX_STANDARD 17
     CXX_STANDARD_REQUIRED Yes
     CXX_EXTENSIONS No
@@ -419,6 +402,18 @@ IF (BUILD_TESTING)
     ${preCICE_SOURCE_DIR}/src
     ${preCICE_SOURCE_DIR}/tests
     )
+  # Copy needed properties from the lib to the executatble. This is necessary as
+  # this executable uses the library source, not only the interface.
+  copy_target_property(precice testprecice COMPILE_DEFINITIONS)
+  copy_target_property(precice testprecice COMPILE_OPTIONS)
+
+  # Testprecice fully depends on MPI and PETSc.
+  if(PRECICE_MPICommunication)
+    target_link_libraries(testprecice PRIVATE MPI::MPI_CXX)
+  endif()
+  if(PRECICE_MPICommunication AND PRECICE_PETScMapping)
+    target_link_libraries(testprecice PRIVATE PETSc::PETSc)
+  endif()
 
   message(STATUS "Including test sources")
   # Test Sources Configuration
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -107,6 +107,7 @@ void BaseCouplingScheme::receiveData(const m2n::PtrM2N &m2n, const DataMap &rece
   for (const DataMap::value_type &pair : receiveData) {
     // Data is only received on ranks with size>0, which is checked in the derived class implementation
     m2n->receive(pair.second->values(), pair.second->getMeshID(), pair.second->getDimensions());
+    pair.second->storeDataAtTime(pair.second->values(), time::Storage::WINDOW_END);
 
     if (pair.second->hasGradient()) {
       m2n->receive(pair.second->gradientValues(), pair.second->getMeshID(), pair.second->getDimensions() * pair.second->meshDimensions());
@@ -115,6 +116,14 @@ void BaseCouplingScheme::receiveData(const m2n::PtrM2N &m2n, const DataMap &rece
     receivedDataIDs.push_back(pair.first);
   }
   PRECICE_DEBUG("Number of received data sets = {}", receivedDataIDs.size());
+}
+
+void BaseCouplingScheme::initializeZeroReceiveData(const DataMap &receiveData)
+{
+  for (const DataMap::value_type &pair : receiveData) {
+    auto values = pair.second->values();
+    pair.second->storeDataAtTime(values, time::Storage::WINDOW_START);
+  }
 }
 
 void BaseCouplingScheme::setTimeWindowSize(double timeWindowSize)
@@ -157,22 +166,9 @@ void BaseCouplingScheme::initialize(double startTime, int startTimeWindow)
   }
 
   exchangeInitialData();
-
-  if (isImplicitCouplingScheme()) {
-    if (not doesFirstStep()) {
-      storeExtrapolationData();
-      moveToNextWindow();
-    }
-  }
+  performReceiveOfFirstAdvance();
 
   _isInitialized = true;
-}
-
-void BaseCouplingScheme::receiveResultOfFirstAdvance()
-{
-  PRECICE_ASSERT(_isInitialized, "Before calling receiveResultOfFirstAdvance() one has to call initialize().");
-  _hasDataBeenReceived = false;
-  performReceiveOfFirstAdvance();
 }
 
 bool BaseCouplingScheme::sendsInitializedData() const
--- a/src/cplscheme/BiCouplingScheme.cpp
+++ b/src/cplscheme/BiCouplingScheme.cpp
@@ -134,17 +134,45 @@ void BiCouplingScheme::exchangeInitialData()
       sendData(getM2N(), getSendData());
     }
     if (receivesInitializedData()) {
+      for (const DataMap::value_type &pair : getReceiveData()) {
+        pair.second->clearTimeStepsStorage(false);
+      }
       receiveData(getM2N(), getReceiveData());
+      for (const DataMap::value_type &pair : getReceiveData()) {
+        pair.second->moveTimeStepsStorage();
+      }
       checkDataHasBeenReceived();
+    } else {
+      initializeZeroReceiveData(getReceiveData());
     }
   } else { // second participant
     if (receivesInitializedData()) {
+      for (const DataMap::value_type &pair : getReceiveData()) {
+        pair.second->clearTimeStepsStorage(false);
+      }
       receiveData(getM2N(), getReceiveData());
+      for (const DataMap::value_type &pair : getReceiveData()) {
+        pair.second->moveTimeStepsStorage();
+      }
       checkDataHasBeenReceived();
+    } else {
+      initializeZeroReceiveData(getReceiveData());
     }
     if (sendsInitializedData()) {
       sendData(getM2N(), getSendData());
     }
+  }
+}
+
+void BiCouplingScheme::retreiveTimeStepReceiveData(double relativeDt)
+{
+  PRECICE_ASSERT(math::greaterEquals(relativeDt, time::Storage::WINDOW_START), relativeDt);
+  PRECICE_ASSERT(math::greaterEquals(time::Storage::WINDOW_END, relativeDt), relativeDt);
+  for (auto &receiveData : getReceiveData()) {
+    auto dataId    = receiveData.second->getDataID();
+    auto allData   = getAllData();
+    auto data      = allData[dataId];
+    data->values() = data->getDataAtTime(relativeDt);
   }
 }
 
--- a/src/cplscheme/CompositionalCouplingScheme.hpp
+++ b/src/cplscheme/CompositionalCouplingScheme.hpp
@@ -81,8 +81,6 @@ public:
       double startTime,
       int    startTimeWindow) final override;
 
-  void receiveResultOfFirstAdvance() override final;
-
   /// Returns true, if any of the composed coupling schemes sendsInitializedData for this participant
   bool sendsInitializedData() const override final;
 
@@ -117,6 +115,13 @@ public:
    * @returns true, if data has been exchanged in last call of advance().
    */
   bool hasDataBeenReceived() const final override;
+
+  /**
+   * @brief retreives time step data from CouplingData into mesh values
+   *
+   * @param relativeDt relative dt associated with the data.
+   */
+  void retreiveTimeStepReceiveData(double relativeDt) final override;
 
   /**
    * @brief Returns the currently computed time of the coupling scheme.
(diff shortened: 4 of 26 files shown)


BenjaminRodenberg:
Currently, some tests are not working. But the basic functionality is fine.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -342,23 +338,10 @@ double SolverInterfaceImpl::initialize()
...
-
-  if (_couplingScheme->hasDataBeenReceived()) {
-    performDataActions({action::Action::READ_MAPPING_PRIOR}, 0.0);
-    mapReadData();
-    performDataActions({action::Action::READ_MAPPING_POST}, 0.0);
-  }
+  performDataActions({action::Action::READ_MAPPING_PRIOR}, 0.0);
+  std::vector<double> receiveTimes{time::Storage::WINDOW_START, time::Storage::WINDOW_END};
+  mapReadData(receiveTimes);
+  performDataActions({action::Action::READ_MAPPING_POST}, 0.0);
```
BenjaminRodenberg, line 344: Possible reason for failing tests: We now always perform the mapping to be able to also store zero initial data in the waveform.

src/precice/impl/SolverInterfaceImpl.hpp
```diff
@@ -525,7 +525,7 @@ class SolverInterfaceImpl {
   void mapWrittenData();
 
   /// Computes, performs, and resets all suitable read mappings.
-  void mapReadData();
+  void mapReadData(std::vector<double> receiveTimes);
```
BenjaminRodenberg, line 528: This part needs a better solution. Compare to #1414

src/time/Storage.cpp
```diff
@@ -4,29 +4,46 @@
...
     : _sampleStorage{}
 {
 }
 
 void Storage::initialize(Eigen::VectorXd values)
 {
-  _sampleStorage.emplace_back(std::make_pair(0.0, values));
-  _sampleStorage.emplace_back(std::make_pair(1.0, values));
+  _sampleStorage.emplace_back(std::make_pair(WINDOW_START, values));
+  _sampleStorage.emplace_back(std::make_pair(WINDOW_END, values));
```
BenjaminRodenberg, line 19: These kind of changes we could already directly merge into develop to reduce the diff. Ok @uekerman?

fsimonis:
Could you try to rebase this PR on #1462 I would prefer to merge it swiftly.

BenjaminRodenberg:
> Could you try to rebase this PR on #1462 I would prefer to merge it swiftly.

I will try to synchronize this PR with develop as soon as I figured out what's going wrong with the Quasi-Newton Test. I guess I will need some more unit tests here first.

BenjaminRodenberg:
I merged #1507 into this PR, but the new unit tests did not give me any new insights regarding the failure of parallel coupling + QN

src/precice/impl/SolverInterfaceImpl.cpp
uekerman: Mmh, not sure why this would lead to failing tests.
But why do we need to map zero initial data in the first place?
And in any case, this sounds like sth that should be documented directly in the code.

src/time/Storage.cpp
uekerman: sure

src/precice/impl/SolverInterfaceImpl.cpp
BenjaminRodenberg: I found the problem with the last failing test. Will write a comment at the right place. The problem was not here.

src/time/Storage.cpp
BenjaminRodenberg: Done. 1e6a41ec6d7178b84f8f4ac6e726af26507ee2ae

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -675,6 +679,23 @@ void BaseCouplingScheme::doImplicitStep()
...
+       * We are using strategy 2), because it's easier to access getAccelerationData() from here. However, this also means that
+       * we have to make sure that the send data is stored in the CouplingData::_timeStepsStorage - even though this is not
+       * needed at the moment. Important note: In https://github.com/precice/precice/pull/1414 also send data requires to keep
+       * track of _timeStepsStorage for subcycling. So it will become simpler as soon as subcycling is fully implemented.
+       */
+      // @todo For other Acceleration schemes as described in "Rüth, B, Uekermann, B, Mehl, M, Birken, P, Monge, A, Bungartz, H-J. Quasi-Newton waveform iteration for partitioned surface-coupled multiphysics applications. Int J Numer Methods Eng. 2021; 122: 5236– 5257. https://doi.org/10.1002/nme.6443" we need a more elaborate implementation.
+      for (auto &pair : getAccelerationData()) {
+        bool override = true;
+        pair.second->storeDataAtTime(pair.second->values(), time::Storage::WINDOW_END, override);
+      }
```
BenjaminRodenberg, line 698: The reason for the last failing test was here: After acceleration we again have to store the accelerated data in the storage, because the acceleration only updates `Mesh::Data`, not `CouplingData::_timeStepsStorage`. This also required some updates of how the storage of send data is managed. See 5f789ac952af9f2aa39ec3da7cfd24a71b2325e4 for the complete picture. I think it should be possible to clean this up a bit more, but most important for now: It's working as expected.

This issue was only a problem for parallel coupling schemes, because only in parallel coupling schemes it is allowed to accelerate send data. For receive data this problem was not visible, because the data in `Mesh::Data` was updated correctly and this is also the data that finally arrives at the user.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -109,7 +109,8 @@ void BaseCouplingScheme::receiveData(const m2n::PtrM2N &m2n, const DataMap &rece
   PRECICE_ASSERT(m2n->isConnected());
   for (const DataMap::value_type &pair : receiveData) {
     // Data is only received on ranks with size>0, which is checked in the derived class implementation
-    m2n->receive(pair.second->values(), pair.second->getMeshID(), pair.second->getDimensions());
+    m2n->receive(pair.second->values(), pair.second->getMeshID(), pair.second->getDimensions()); // @todo don't receive into pair.second->values(), but receive data into Storage.
+    pair.second->storeDataAtTime(time::Storage::WINDOW_END);
```
BenjaminRodenberg, line 113:
```suggestion
void BaseCouplingScheme::receiveData(const m2n::PtrM2N &m2n, const DataMap &receiveData, bool recvInit = false)
{
    ...
    Eigen::VectorXd recvBuffer;
    m2n->receive(buffer, pair.second->getMeshID(), pair.second->getDimensions());
    // will be changed via https://github.com/precice/precice/pull/1414
    if (recvInit) {
        pair.second->storeDataAtTime(recvBuffer, time::Storage::WINDOW_START);
    } else {
        pair.second->storeDataAtTime(recvBuffer, time::Storage::WINDOW_END);    
    }
```

src/cplscheme/SerialCouplingScheme.cpp
```diff
@@ -111,14 +119,34 @@ void SerialCouplingScheme::exchangeSecondData()
...
+    for (const DataMap::value_type &pair : getReceiveData()) {
+      pair.second->clearTimeStepsStorage(true);
+    }
     receiveData(getM2N(), getReceiveData());
+    if (hasConverged()) {
+      // received converged result of this window, trigger move
+      for (const DataMap::value_type &pair : getReceiveData()) {
+        pair.second->moveTimeStepsStorage();
+      }
+    }
```
BenjaminRodenberg, line 133: Try to move this part inside receive. Possible problem: In serial coupling scheme this has to be triggered before receive for the first and after receive for the second participant.

src/cplscheme/SerialCouplingScheme.cpp
```diff
@@ -111,14 +119,34 @@ void SerialCouplingScheme::exchangeSecondData()
     if (isImplicitCouplingScheme()) {
       receiveConvergence(getM2N());
     }
+
     PRECICE_DEBUG("Receiving data...");
+    for (const DataMap::value_type &pair : getReceiveData()) {
+      pair.second->clearTimeStepsStorage(true);
+    }
```
BenjaminRodenberg, line 126: Should not be needed with proper implementation in `receiveData`

BenjaminRodenberg (commented):
Next steps below identified during discussion with @uekerman. Goal: Make this PR a reasonable intermediate step towards https://github.com/precice/precice/pull/1414.

src/cplscheme/CouplingScheme.hpp
```diff
@@ -173,6 +163,17 @@ class CouplingScheme {
...
+  // @todo find a better name. This is too low level.
+  virtual void storeTimeStepReceiveData(double relativeDt) = 0;
+
+  // @todo find a better name. This is too low level.
+  /**
+   * @brief retreives time step data from CouplingData into mesh values
+   *
+   * @param relativeDt relative dt associated with the data.
+   */
+  virtual void retreiveTimeStepReceiveData(double relativeDt) = 0;
```
BenjaminRodenberg, line 175: Can we also use `exchangeSecondData()` here?

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: Done: 0aa0a6899fc0610496c5e720388118cdf92eea50

src/cplscheme/SerialCouplingScheme.cpp
BenjaminRodenberg: Done. 0aa0a6899fc0610496c5e720388118cdf92eea50

src/cplscheme/SerialCouplingScheme.cpp
BenjaminRodenberg: Still looks difficult. 0aa0a6899fc0610496c5e720388118cdf92eea50

src/cplscheme/CouplingScheme.hpp
BenjaminRodenberg: Situation improved a lot, but still looks like we need a special treatment for `SerialCouplingScheme`. 0aa0a6899fc0610496c5e720388118cdf92eea50

BenjaminRodenberg:
There is currently a problem with the interaction of extrapolation and the time window storage. Since one of the next goals is to merge extrapolation into the waveform, it might be reasonable to do this directly in this PR.

BenjaminRodenberg:
I think this PR should be fine. Extrapolation is still not ideal, but I will entirely move extrapolation inside of the storage in a follow-up pre. Let's try to merge this one soon. @fsimonis and @uekerman any more comments here?

src/cplscheme/BiCouplingScheme.cpp
```diff
@@ -138,23 +138,67 @@ CouplingData *BiCouplingScheme::getReceiveData(
 void BiCouplingScheme::exchangeInitialData()
 {
   // F: send, receive, S: receive, send
+  bool recvInitialData = true;
+  bool sendInitialData = true;
```
uekerman, line 142: Naming here is really unfortunate as these variable do not store whether initial data is send or received.
Wouldn't it be easier if the variable of `sendData` and `receiveData` had the same name, e.g. `initialCommunication` or `initialize`?

src/cplscheme/BiCouplingScheme.cpp
```diff
@@ -138,23 +138,67 @@ CouplingData *BiCouplingScheme::getReceiveData(
...
+      sendData(getM2N(), getSendData(), sendInitialData);
+    }
+  }
+}
+
+void BiCouplingScheme::storeTimeStepReceiveData(double relativeDt)
+{
+  PRECICE_ASSERT(math::greaterEquals(relativeDt, time::Storage::WINDOW_START), relativeDt);
+  PRECICE_ASSERT(math::greaterEquals(time::Storage::WINDOW_END, relativeDt), relativeDt);
+  if (hasDataBeenReceived()) {
```
uekerman, line 171: Wouldn't it be easier to understand if this `if` was outside of this function?

src/cplscheme/CouplingData.cpp
```diff
@@ -132,4 +132,27 @@ void CouplingData::storeExtrapolationData()
...
+  PRECICE_ASSERT(math::greaterEquals(relativeDt, time::Storage::WINDOW_START), relativeDt);
+  PRECICE_ASSERT(math::greaterEquals(relativeDt, _timeStepsStorage.maxStoredNormalizedDt()), relativeDt, _timeStepsStorage.maxStoredNormalizedDt());
+  PRECICE_ASSERT(math::greaterEquals(time::Storage::WINDOW_END, relativeDt), relativeDt);
+  _timeStepsStorage.setValueAtTime(relativeDt, data, mustOverrideExisting);
+}
+
+Eigen::VectorXd CouplingData::getValuesAtTime(double relativeDt)
+{
+  return _timeStepsStorage.getValueAtTime(relativeDt);
+}
```
uekerman, line 156: `value` vs `values`?

src/cplscheme/SerialCouplingScheme.cpp
```diff
@@ -77,13 +77,14 @@ void SerialCouplingScheme::receiveAndSetTimeWindowSize()
...
-  } else { // second participant
+  if (!doesFirstStep()) { // second participant
+    // similar to SerialCouplingScheme::exchangeSecondData()
     receiveAndSetTimeWindowSize();
     PRECICE_DEBUG("Receiving data...");
     receiveData(getM2N(), getReceiveData());
-    checkDataHasBeenReceived();
+    if (not hasDataBeenReceived()) {
+      checkDataHasBeenReceived();
+    }
```
uekerman, line 87: If we already know that data has been received, why do we have to check it?

uekerman (changes requested):
I am still struggling quite a bit to understand what happens just by reading the names.

src/cplscheme/SerialCouplingScheme.cpp
BenjaminRodenberg: Because for the second participant in a `SerialCouplingScheme` there are two places where data can be received: When we initialize or when we receive it from the other participant's advance. We have to check whether `hasDataBeenReceived()` is true before calling `checkDataHasBeenReceived()`, because an assertion is triggered, if `checkDataHasBeenReceived` is called twice (safeguard).

src/cplscheme/BiCouplingScheme.cpp
BenjaminRodenberg: I think this `if` is not even needed. Will try to remove it.

src/cplscheme/BiCouplingScheme.cpp
BenjaminRodenberg: Sure. The name is only temporary and will vanish again in #1414, but I don't really care.

src/cplscheme/CouplingData.cpp
BenjaminRodenberg: Will use `values`, because it is consistent with `mesh::Data::values()`.

src/cplscheme/BiCouplingScheme.cpp
BenjaminRodenberg: Done. c58b6128a522af256756516fbfc81b8613a52fe4

src/cplscheme/CouplingData.cpp
BenjaminRodenberg: Done. 1acc0af70eea926c8691bd9a203e8bdb1ebc2ede

src/cplscheme/SerialCouplingScheme.cpp
uekerman: ok, thanks, makes more sense now. please add this information as a comment here

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -676,6 +704,23 @@ void BaseCouplingScheme::doImplicitStep()
...
+       * We are using strategy 2), because it's easier to access getAccelerationData() from here. However, this also means that
+       * we have to make sure that the send data is stored in the CouplingData::_timeStepsStorage - even though this is not
+       * needed at the moment. Important note: In https://github.com/precice/precice/pull/1414 also send data requires to keep
+       * track of _timeStepsStorage for subcycling. So it will become simpler as soon as subcycling is fully implemented.
+       */
+      // @todo For other Acceleration schemes as described in "Rüth, B, Uekermann, B, Mehl, M, Birken, P, Monge, A, Bungartz, H-J. Quasi-Newton waveform iteration for partitioned surface-coupled multiphysics applications. Int J Numer Methods Eng. 2021; 122: 5236– 5257. https://doi.org/10.1002/nme.6443" we need a more elaborate implementation.
+      for (auto &pair : getAccelerationData()) {
+        bool mustOverride = true;
+        pair.second->storeValuesAtTime(time::Storage::WINDOW_END, pair.second->values(), mustOverride);
+      }
```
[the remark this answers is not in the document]
BenjaminRodenberg: Fixed override vs. overwrite in 43c54159f6d0757c2aaefdf2c33058a0138eb102.

In `receiveData` we clear the storage first and then store the values. There is no need for overwriting and all data (except acceleration data) is final. Only for the acceleration data we have to override all data after acceleration, because the acceleration puts the accelerated data into `pair.second->values()`, not into `time::Storage`.

We could also let the acceleration directly access the storage, but this would lead to many changes (again) and probably a lot of spaghetti. (Same argumentation like for the mapping)

src/cplscheme/CouplingData.cpp
```diff
@@ -142,12 +142,12 @@ void CouplingData::moveTimeStepsStorage()
   _timeStepsStorage.move();
 }
 
-void CouplingData::storeValuesAtTime(double relativeDt, Eigen::VectorXd data, bool mustOverrideExisting)
+void CouplingData::storeValuesAtTime(double relativeDt, Eigen::VectorXd data, bool mustOverwiteExisting)
```
uekerman, line 145:
```suggestion
void CouplingData::storeValuesAtTime(double relativeDt, Eigen::VectorXd data, bool mustOverwriteExisting)
```

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -101,15 +109,27 @@ void BaseCouplingScheme::sendData(const m2n::PtrM2N &m2n, const DataMap &sendDat
...
+    // will be changed via https://github.com/precice/precice/pull/1414
+    if (recvInitialData) {
+      pair.second->storeValuesAtTime(time::Storage::WINDOW_START, recvBuffer);
+      pair.second->storeValuesAtTime(time::Storage::WINDOW_END, recvBuffer);
+    } else {
+      pair.second->clearTimeStepsStorage();
+      pair.second->storeValuesAtTime(time::Storage::WINDOW_END, recvBuffer);
+    }
+
+    pair.second->values() = recvBuffer; // @todo should happen somewhere else!
```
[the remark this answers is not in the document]
BenjaminRodenberg: Not 100% sure, but generally it would be nice to take care of putting the "right" data into `pair.second->values()` just before we move from the `cplscheme` package to another package (acceleration, mapping ...). That's the reason I marked it here. But it is also fine the way it is right now. It might become easier to put it somewhere else when we continue working on #1414 

See also aa2765b47c6ef81039ca41647f30814884e5ee34

src/cplscheme/BaseCouplingScheme.cpp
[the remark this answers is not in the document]
BenjaminRodenberg: I changed this part in aa2765b47c6ef81039ca41647f30814884e5ee34. This is strictly not needed, but I think it's easier to read, if we first put data from the storage into `pair.second->values()` before acceleration, then acceleration processed the data and after acceleration the coupling scheme takes care of putting the data at the right place into `pair.second->values()`.

src/cplscheme/SerialCouplingScheme.cpp
BenjaminRodenberg: Done. 9a518593e72e6e600c2cf5397ece974ecd3b0aaa

src/cplscheme/BaseCouplingScheme.cpp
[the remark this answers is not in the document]
uekerman: Did you take into consideration that acceleration data is used to compute the acceleration (primary data), but then **all** data is accelerated (secondary data)? See also: https://precice.org/configuration-acceleration.html#coupling-data-and-primary-data

src/cplscheme/CompositionalCouplingScheme.hpp
```diff
@@ -126,6 +124,12 @@ class CompositionalCouplingScheme final : public CouplingScheme {
    */
   bool hasDataBeenReceived() const final override;
 
+  /// @copydoc CouplingScheme::storeReceiveData
```
fsimonis, line 127: This is done automatically for virtual functions.

src/cplscheme/CouplingData.hpp
```diff
@@ -71,11 +72,23 @@ class CouplingData {
...
   void storeExtrapolationData();
 
+  /// clears _timeStepsStorage. Called after data was written or before data is received.
+  void clearTimeStepsStorage();
+
+  /// moves _timeStepsStorage. Called after converged data was received.
+  void moveTimeStepsStorage(); // @todo very easy to mix up with moveToNextWindow. Try to rename or merge!
+
+  /// stores data at key relativeDt in _timeStepsStorage for later use.
+  void storeValuesAtTime(double relativeDt, Eigen::VectorXd data, bool mustOverwriteExisting = false);
```
fsimonis, line 87: Prefer an `enum struct X : bool {...}` over a blank bool.

fsimonis (commented):
Some general comments to improve readability.

src/cplscheme/BaseCouplingScheme.cpp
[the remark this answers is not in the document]
BenjaminRodenberg: I guess this might cause trouble as soon as we apply acceleration to a scenario where subcycling is used. Here, I think no special treatment is needed: The acceleration only "knows" about the data returned from in `getAccelerationData()` and we also take care of all this data. If there is a possible problem, we should first come up with an integration test to reproduce this problem. Otherwise we will break it again. I see two possibilities:

* We are missing a test. Then we should add a test on develop, not here.
* We are not missing a test. Then everything should work.

For subcycling i.e. #1414 I'm quite sure we are missing a test here. But let's worry about this as soon as #1504 is merged (already complicated enough).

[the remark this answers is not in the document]
uekerman: My bad: acceleration data already contains primary and secondary data. So "taking care of acceleration data" should be enough.

I could not find a suitable existing integration or unit test right away. Could be very well that we are lacking tests here.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -120,6 +140,15 @@ void BaseCouplingScheme::receiveData(const m2n::PtrM2N &m2n, const DataMap &rece
   PRECICE_DEBUG("Number of received data sets = {}", receivedDataIDs.size());
 }
 
+void BaseCouplingScheme::initializeZeroReceiveData(const DataMap &receiveData)
+{
+  for (const DataMap::value_type &pair : receiveData) {
```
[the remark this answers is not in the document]
BenjaminRodenberg: Done 715d89ed2a8031164e73b57297fb08e5cc343027.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -676,6 +704,23 @@ void BaseCouplingScheme::doImplicitStep()
...
+       * 1) Only override the data in CouplingData::_timeStepsStorage that is part of the receive data
+       * 2) Override the data in CouplingData::_timeStepsStorage for all data in getAccelerationData()
+       *
+       * We are using strategy 2), because it's easier to access getAccelerationData() from here. However, this also means that
+       * we have to make sure that the send data is stored in the CouplingData::_timeStepsStorage - even though this is not
+       * needed at the moment. Important note: In https://github.com/precice/precice/pull/1414 also send data requires to keep
+       * track of _timeStepsStorage for subcycling. So it will become simpler as soon as subcycling is fully implemented.
+       */
+      // @todo For other Acceleration schemes as described in "Rüth, B, Uekermann, B, Mehl, M, Birken, P, Monge, A, Bungartz, H-J. Quasi-Newton waveform iteration for partitioned surface-coupled multiphysics applications. Int J Numer Methods Eng. 2021; 122: 5236– 5257. https://doi.org/10.1002/nme.6443" we need a more elaborate implementation.
+      for (auto &pair : getAccelerationData()) {
```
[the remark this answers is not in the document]
BenjaminRodenberg: 715d89ed2a8031164e73b57297fb08e5cc343027

src/cplscheme/BiCouplingScheme.cpp
```diff
@@ -138,23 +138,67 @@ CouplingData *BiCouplingScheme::getReceiveData(
...
+  for (auto &receiveData : getReceiveData()) {
+    auto dataId               = receiveData.second->getDataID();
+    auto allData              = getAllData();
+    allData[dataId]->values() = allData[dataId]->getValuesAtTime(relativeDt);
+  }
+}
+
+void BiCouplingScheme::clearTimeStepSendStorage()
+{
+  for (const DataMap::value_type &pair : getSendData()) {
```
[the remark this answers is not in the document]
BenjaminRodenberg: 715d89ed2a8031164e73b57297fb08e5cc343027

src/cplscheme/BiCouplingScheme.cpp
```diff
@@ -138,23 +138,67 @@ CouplingData *BiCouplingScheme::getReceiveData(
...
+    pair.second->clearTimeStepsStorage();
+  }
+}
+
+void BiCouplingScheme::storeTimeStepSendData(double relativeDt)
+{
+  for (const DataMap::value_type &pair : getSendData()) {
```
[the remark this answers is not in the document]
BenjaminRodenberg: 715d89ed2a8031164e73b57297fb08e5cc343027

src/cplscheme/BaseCouplingScheme.cpp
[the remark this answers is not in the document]
BenjaminRodenberg: > Could be very well that we are lacking tests here.

Wouldn't be too surprised. I will create make this a todo in #1414, because we definitely want to check it there. Everything will become more complicated with subcycling and without proper testing at least I will get 100% lost.

src/cplscheme/CompositionalCouplingScheme.hpp
BenjaminRodenberg: Done. abdd415a50a8edb822189ae63462849e73dea404

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -676,6 +704,23 @@ void BaseCouplingScheme::doImplicitStep()
...
+        bool mustOverride = true;
+        pair.second->storeValuesAtTime(time::Storage::WINDOW_END, pair.second->values(), mustOverride);
```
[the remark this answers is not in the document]
BenjaminRodenberg: To me this looks quite complicated and this is also a candidate for some changes in #1414. I will keep the bools as they are for now and we can come back to this when I'm taking care of #1414.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -120,6 +140,15 @@ void BaseCouplingScheme::receiveData(const m2n::PtrM2N &m2n, const DataMap &rece
   PRECICE_DEBUG("Number of received data sets = {}", receivedDataIDs.size());
 }
 
+void BaseCouplingScheme::initializeZeroReceiveData(const DataMap &receiveData)
+{
+  for (const DataMap::value_type &pair : receiveData) {
+    auto zeroData = Eigen::VectorXd::Zero(pair.second->getSize());
+    pair.second->storeValuesAtTime(time::Storage::WINDOW_START, zeroData);
+    pair.second->storeValuesAtTime(time::Storage::WINDOW_END, zeroData);
```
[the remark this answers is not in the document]
BenjaminRodenberg: Done. bb839cf5cd2f52ff65491811b6363c935221752b

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -267,21 +310,43 @@ void BaseCouplingScheme::secondExchange()
...
 {
   PRECICE_TRACE(_timeWindows);
+  // @todo breaks for CplSchemeTests/SerialImplicitCouplingSchemeTests/testConfiguredAbsConvergenceMeasureSynchronized. Why? @fsimonis
+  // for (auto &data : getAllData() | boost::adaptors::map_values) {
+  //   data->storeExtrapolationData();
+  // }
   for (auto &pair : getAllData()) {
-    PRECICE_DEBUG("Store data: {}", pair.first);
     pair.second->storeExtrapolationData();
   }
```
BenjaminRodenberg, line 319: I was only partially successful when trying to use `map_values`. I marked the corresponding sections. Do you see what I'm doing wrong here?

CMakeLists.txt
```diff
@@ -164,8 +164,6 @@ endif()
 find_package(Boost 1.71.0 REQUIRED
   COMPONENTS filesystem log log_setup program_options system thread unit_test_framework
   )
-mark_as_advanced(Boost_FILESYSTEM_LIBRARY_RELEASE Boost_INCLUDE_DIR Boost_LOG_LIBRARY_RELEASE Boost_LOG_SETUP_LIBRARY_RELEASE Boost_PROGRAM_OPTIONS_LIBRARY_RELEASE Boost_SYSTEM_LIBRARY_RELEASE Boost_THREAD_LIBRARY_RELEASE Boost_UNIT_TEST_FRAMEWORK_LIBRARY_RELEASE)
-mark_as_advanced(Boost_FILESYSTEM_LIBRARY_DEBUG Boost_INCLUDE_DIR Boost_LOG_LIBRARY_DEBUG Boost_LOG_SETUP_LIBRARY_DEBUG Boost_PROGRAM_OPTIONS_LIBRARY_DEBUG Boost_SYSTEM_LIBRARY_DEBUG Boost_THREAD_LIBRARY_DEBUG Boost_UNIT_TEST_FRAMEWORK_LIBRARY_DEBUG)
```
uekerman, line 168: sth went wrong while rebasing

BenjaminRodenberg: Thanks for the hint. Should work now.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: I think I found a hint now: removing the function `getAllData()` and instead creating and maintaining another map `_cplData` allows us to use `boost::adaptors::map_values`. For `getAccelerationData()` the situation is similar, but I did not want to create another map for acceleration data. See [708617b](https://github.com/precice/precice/pull/1523/commits/708617b14a741ac1f3fb7cb74e28192311ece277).

I will also open an issue, because `getAllData()` has some more problems in connection with multi coupling.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1856,11 +1844,19 @@ void SolverInterfaceImpl::performDataActions(
...
+    _couplingScheme->loadReceiveDataFromStorage(time::Storage::WINDOW_END);
+  }
   for (action::PtrAction &action : _accessor->actions()) {
     if (timings.find(action->getTiming()) != timings.end()) {
       action->performAction(time);
     }
   }
+  if (_couplingScheme->hasDataBeenReceived()) {
+    _couplingScheme->storeReceiveData(time::Storage::WINDOW_END);
+  }
```
BenjaminRodenberg, line 1859: In #1523 the current implementation of actions turned out to become more and more problematic. So I removed them via [91d76c6](https://github.com/precice/precice/pull/1523/commits/91d76c678f246670d3fedf0c1034bfb46d84ae62). This here is basically already the first step into the problematic direction. I guess we have to try to find a better solution here.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: #1535 fixes the problems here. Like mentioned above: `getAccelerationData()` is still a problem.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1841,16 +1816,25 @@ void SolverInterfaceImpl::mapWrittenData()
   }
 }
 
-void SolverInterfaceImpl::mapReadData()
+void SolverInterfaceImpl::mapReadDataAndPerformReadAction(std::vector<double> receiveTimes)
```
uekerman, line 1819: IMO, this function (name) is overkill. But let's re-evaluate once we have a symmetric treatment of read and write.

BenjaminRodenberg:
To give a short summary on the status and next steps of this PR: 

Events have been simplified in #1614. Our next goal is to reach a state where read and write data is more symmetric. For this, we need to find a conclusion in https://github.com/precice/precice/pull/1612.

BenjaminRodenberg:
Replaced by #1639

PR #1510, opened 2022-12-09
Skip QN update with zero difference residual

### LINKED ISSUE #1496: IQN implementation does not handle starting from zero-state
**Describe your setup**

preCICE Version: v2.5.0

**Describe the problem**

Users starting simulations from a zero/steady state, often face the following error when they apply an IQN acceleration:

https://github.com/precice/precice/blob/e920870a8848e1f4fbf194c6f545c52b88083d2b/src/acceleration/BaseQNAcceleration.cpp#L219-L222

This often leads them to fall back to much slower methods, such as Aitken or even constant under-relaxation. Typically, IQN works just fine in these cases if we don't start from a zero initial state.

This has already affected at least:
- [This user simulating a Fluid-Fluid coupling scenario](https://precice.discourse.group/t/fluid-fluid-coupling/1185/2)
- [This user simulating an FSI scenario](https://precice.discourse.group/t/mapping-issue-in-fsi-simulation/647/6)
- Our ATHLET-preCICE project (starting from zero flow and slowly opening a valve)

**Step To Reproduce**

I have not managed to reproduce this situation with any of our tutorials. I already tried the following:
- Modifying the `flow-over-heated-plate` to use IQN instead of Aitken (see https://github.com/precice/tutorials/issues/313)
- Modifying the `partitioned-pipe` tutorial to start with zero flow and to set stricter convergence measures
- Modifying the `partitioned-heat-conduction` case to use stricter convergence measures. In a currently not publicly available variant of this case where we set the right-hand-side to zero, we can reproduce the issue with relative convergence measures of 1e-9.

@uekerman any further ideas on what could trigger this error?

**Expected behaviour**

In case this situation is encountered, IQN could just ignore the column and keep running, only showing a warning.

### LINKED ISSUE #1558: Sending zero displacements in a serial-implicit coupling crashes precice (perpendicular-flap tutorial)
**Describe your setup**

Operating system (e.g. Linux distribution and version): _Linux Rocky 8.0_
preCICE Version: 2.5.0

**Describe the problem**
I am developing a new fsi adapter and due to programming error on my side, the coupling data was lagging behind 1 coupling step. This resulted in sending zero displacement to the fluid-solver once too often. When using a parallel-implicit coupling, there was no problem, but when using a _serial-implicit_ coupling, precice crashes with a vague warning message and core-dump due to std::dequeue failure (see below).

NOTES: 
- the problem was solved by sending the correct displacements in the new fsi adapter.
- the problem can be reproduced with the attached solid participant that always sends zero displacement.
- the problem has nothing to do with my programming error, it is just how I found the problem.

**Error message**
```
---[precice]  relative convergence measure: relative two-norm diff of data "Displacement" = inf, limit = 1.00e-03, normalization = 0.00e+00, conv = true
---[precice]  All converged
---[precice] WARNING:  The coupling residual equals almost zero. There is maybe something wrong in your adapter. Maybe you always write the same data or you call advance without providing new data first or you do not use available read data. Or you just converge much further than actually necessary.
/cm/local/apps/gcc/9.2.0/include/c++/9.2.0/bits/stl_deque.h:1477: std::_Deque_base<_Tp, _Alloc>::_Alloc_traits::reference std::deque<_Tp, _Alloc>::front() [with _Tp = int; _Alloc = std::allocator<int>]: Assertion '__builtin_expect(!this->empty(), true)' failed.
Aborted (core dumped)
[precice-crash.zip](https://github.com/precice/precice/files/10660838/precice-crash.zip)

```

**Step To Reproduce**
1. Run the attached adapter (solid-python.py) with the attached config (precice-config.serial-implicit.xml) with the tutorial perpendicular-flap example and the fluid-openfoam solver.

**Expected behaviour**
If sending zero displacement should be possible, that needs to be fixed. 
Otherwise, I would have liked to get more explanation on "The coupling residual equals almost zero", especially while the line above says that the residual is infinity. Furthermore, a better explanation on why sending zero displacement is not suitable for the serial implicit coupling. 

**Additional context**
None.

KyleDavisSA:
## Main changes of this PR
This PR does not add a new column to V and W if the residual difference is zero. If this is the case, a simple under-relaxation step is performed instead. 

Closes #1496
Closes #1558

## Author's checklist

## Reviewers' checklist


src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -216,42 +216,43 @@ void BaseQNAcceleration::updateDifferenceMatrices(
         residualMagnitude /= utils::IntraComm::l2norm(_values);
       }
 
-      PRECICE_CHECK(not math::equals(residualMagnitude, 0.0),
-                    "Attempting to add a zero vector to the quasi-Newton V matrix. This means that the residuals "
-                    "in two consecutive iterations are identical. If a relative convergence limit was selected, "
-                    "consider increasing the convergence threshold.");
+      if (math::equals(residualMagnitude, 0.0)) {
+        _forceUnderRelaxationStep = true;
+      }
```
uekerman, line 221: Let's inform the user about this. Could be an INFO or a WARNING, not sure yet.

uekerman (commented):
Isn't it a bit random that we do an underrelaxation with the initial underrelaxation factor again? Couldn't we not also do a QN iteration but just not add the current iteration?

MakisH:
The [corresponding branch based on v2.5](https://github.com/KyleDavisSA/precice/commits/zero_residual_main) seems to fix the problem described in https://github.com/precice/precice/issues/1496#issuecomment-1347389600

Instead of failing in the first iteration, the simulation completes and the results look smooth.

src/acceleration/BaseQNAcceleration.cpp
MakisH: I guess the question is: "Is this a normal step of the procedure, or a special case we are trying to recover from and which may blow up later?".

If normal, then `INFO`. If "may blow up", then `WARNING`.

For my use case, an `INFO` would be more expected.

KyleDavisSA: With the issue Makis described, there was not even one column added to V and W, therefore a QN step can't be performed. We could add some more conditions, such as, perform a QN step if V and W already have 5 columns, otherwise do a under-relaxation step. But this becomes a weird condition imposed by the developers. I think that having very few columns might not converge quickly, but then again, would simple under-relaxation?

MakisH: Don't we anyway always do an under-relaxation for the first few iterations?

KyleDavisSA: No, only the very first iteration. But it cannot be guaranteed that if a column is not added after the first iteration, that it will be added after the second. We need to safeguard against the case where the solvers are essentially static in the beginning and V and W are not being appended with columns.

BenjaminRodenberg (changes requested):
@uekerman and @MakisH is there a good reason for not merging this PR (as it is)? With https://github.com/precice/precice/pull/1863 we have just started reinventing the wheel.

From my perspective this PR still lacks a test.

MakisH:
> @uekerman and @MakisH is there a good reason for not merging this PR (as it is)? With #1863 we have just started reinventing the wheel.
> 
> From my perspective this PR still lacks a test.

I have no clue why this is not merged yet. Feel free to take it over.

BenjaminRodenberg:
@Fujikawas this PR and https://github.com/precice/precice/pull/1863 cover more or less the same thing, correct? Can we merge one and close the other?

@MakisH I don't have time to take over responsibility for this PR. Just trying to push things a bit ;)

Fujikawas:
> @Fujikawas this PR and https://github.com/precice/precice/pull/1863 cover more or less the same thing, correct? Can we merge one and close the other?
> 


Yes, they are motivated by the same thing. I would support to merge https://github.com/precice/precice/pull/1863 and close this one :)

uekerman:
This was not merged as there are still critical open discussions.

Closing in favor of #1863

PR #1513, opened 2022-12-13
Overhaul Serialization

fsimonis:
## Main changes of this PR

This PR overhauls the serialization logic regarding the Mesh and Partition-related data structures.
This includes sending/receiving/broadcasting of `Mesh`, `mesh::BoundingBoxMap`, `mesh::BoundingBox` and `mesh::CommunicationMap`/`mesh::ConnectionMap`.

These functions used to interleave serialization and traversal.
This is error-prone and slow, as the data structures are send bit by bit.

This PR refactors this into clear sub-steps:

```mermaid
flowchart LR;
    
    T([Type]) -->|serialize| S([Serialized Type]);
    S -->|toType| T;
    S -->|broadcastSend| C;
    S -->|send| C([Communication]);
    C -->|receive| S;
    C -->|broadcastReceive| S;
    S -->|assertValid| S;
```

Given an instance of type `T`, one can `serialize()` it.
This serialized representation of `T` can now be sent or broadcast.
On the receiving side, one can receive a serialized representation of a type `T` and then call `to<Type>()` to build the desired type.

Upsides are:
* The serialized types require a fixed amount of communication calls. Most types are sent as two ranges. Prior to this, each key-value pair in maps was sent individually. (linear changed to constant complexity, while size of sends got bigger)
* The communication functions are super simple, hence easier to debug.
* The serialized types are validated after serialization and after reception, thus safer to handle.

This PR was created with `std::size_t VertexID`s in mind.

## Author's checklist

## Reviewers' checklist


src/com/SerializeMesh.hpp
```diff
@@ -0,0 +1,47 @@
...
+
+class SerializedMesh {
+public:
+  static SerializedMesh serialize(const mesh::Mesh &mesh);
+  void                  addToMesh(mesh::Mesh &mesh) const;
+
+  void assertValid() const;
+
+  void                  sendTo(Communication &communication, int rankReceiver);
+  static SerializedMesh receiveFrom(Communication &communication, int rankSender);
```
uekerman, line 22:
```suggestion
  void                  send(Communication &communication, int rankReceiver);
  static SerializedMesh receive(Communication &communication, int rankSender);
```
To be consistent with names we use in `com`?

src/mesh/Mesh.hpp
```diff
@@ -49,6 +49,7 @@ class Mesh {
 
   /// A mapping from remote local ranks to the IDs that must be communicated
   using CommunicationMap = std::map<Rank, std::vector<VertexID>>;
+  using ConnectionMap    = CommunicationMap; // until we decide on a name
```
uekerman, line 52: because we currently use both? Should we try to resolve this here?

uekerman (approved):
Looks very clean. I did not go through all the logic, but only had a look at the interfaces.

* A further upside: Things are easier to unit-test. You still hide this fact a bit because you didn't refactor `CommunicateBoundingBoxTest` and `CommunicateMeshTest.cpp`. Could be easier to understand if we had `Serialize(d)MeshTest` and `Serialize(d)PartitionTest` instead.
* You named the files `SerializeX` while the classes are called `SerializedX`. Did you do this on purpose? Might confuse more than it helps.
* What's the rationale to not remove `CommunicateBoundingBox` completely?
* What you wrote in the PR description: Could we add this somewhere in the code?

src/mesh/Mesh.hpp
fsimonis: I could, but this mostly unrelated to these changes.

src/com/SerializeMesh.hpp
uekerman: IMHO, there is no additional information in the "from/to". From MPI, we are used to plain send / receive.

src/mesh/Mesh.hpp
uekerman: Then let's open an issue?

fsimonis: https://github.com/precice/precice/issues/1539

fsimonis:
> A further upside: Things are easier to unit-test. You still hide this fact a bit because you didn't refactor CommunicateBoundingBoxTest and CommunicateMeshTest.cpp. Could be easier to understand if we had Serialize(d)MeshTest and Serialize(d)PartitionTest instead.

True. I would prefer to handle all of this after #712. 

> You named the files SerializeX while the classes are called SerializedX. Did you do this on purpose? Might confuse more than it helps.

True

> What's the rationale to not remove CommunicateBoundingBox completely?

There are still instances where individual boundingboxes need to be transferred.

> What you wrote in the PR description: Could we add this somewhere in the code?

Which part? The motivation?

uekerman:
> True. I would prefer to handle all of this after https://github.com/precice/precice/pull/712.

You probably mean #1418 

Then, please open an issue before merging here.

> There are still instances where individual boundingboxes need to be transferred.

But why shouldn't we serialize these the same way?

> Which part? The motivation?

Yes, mainly the figure + explanation.

fsimonis:
> Then, please open an issue before merging here.

Done #1542 

> But why shouldn't we serialize these the same way?

Just checked again, and I actually could already implement this here. Will have a look.

> Yes, mainly the figure + explanation.

Will add this to the `precice::com::serialize` namespace.

fsimonis:
Added docs and replaced the CommunicateBoundingBox by SerializedBoundingBox.

uekerman:
Nice job :+1: 
One last thing: Could you please mention in the class documentation of `SerializedMesh` and `SerializedPartition` where to find the documentation.

Otherwise, please merge when ready from your side

PR #1520, opened 2022-12-18
Perform extrapolation inside Storage

BenjaminRodenberg:
## Main changes of this PR

Removes extrapolation from cplscheme package. Storage now directly takes care of extrapolation.

Removes a lot of bookkeeping logic, because both extrapolation and storage need to keep track of data at the beginning and at the end of a window.

## Author's checklist

* [x] ~~I added a test to cover the proposed changes in our test suite.~~ Not needed. Refactoring.

## Reviewers' checklist


src/cplscheme/tests/SerialImplicitCouplingSchemeTest.cpp
```diff
@@ -736,8 +736,8 @@ BOOST_AUTO_TEST_CASE(FirstOrderWithAcceleration)
     // first order extrapolation
     BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 4); // = 2*3 - 2
   } else if (context.isNamed(second)) {
-    // extrapolation only applied to accelerated data. So data written by first participant.
-    BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 3);
+    // first order extrapolation
+    BOOST_TEST(mesh->data(receiveDataIndex)->values()(0) == 5); // = 2*3 - 1
```
BenjaminRodenberg, line 740: This is an important change in the test which might make this PR breaking: Extrapolation is now applied to all data, not only to the accelerated data. This simplifies the implementation and this change is also not observable with our integration tests. Performing the extrapolation is also not expensive.

src/cplscheme/MultiCouplingScheme.cpp
```diff
@@ -142,12 +198,19 @@ void MultiCouplingScheme::exchangeSecondData()
...
-
     for (auto &receiveExchange : _receiveDataVector) {
       receiveData(_m2ns[receiveExchange.first], receiveExchange.second);
     }
     checkDataHasBeenReceived();
   }
+  if (hasConverged()) {
+    for (const DataMap::value_type &data : getAllData()) {
+      data.second->moveTimeStepsStorage();
+    }
```
BenjaminRodenberg, line 209: Here, we call for `getAllData()` the function `moveTimeStepsStorage()`. This performs extrapolation to compute an initial guess for all data (not only acceleration data). We have to call `moveTimeStepsStorage()` for all data anyway. Differentiating between accelerated data (use extrapolation when moving) and data that is not accelerated (don't use extrapolation when moving) would be overly complicated.

BenjaminRodenberg:
Replaced by #1639

PR #1524, opened 2022-12-30
Correct a typo in error message: "before finalize" to "after finalize"

kanishkbh:
## Main changes of this PR

Data writing functions in the API must be called before `finalize()` and there is a check for this.
The error message for this check in `writeVectorData(...)` has a typo: it reports the error as "cannot be called **before** finalize".
This PR corrects this to "cannot be called **after** finalize".

## Author's checklist

* [NA] I added a test to cover the proposed changes in our test suite.
* [NA] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).
* [ NA] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist


uekerman (approved):
Makes sense. Thanks!

PR #1523, opened 2022-12-22
WIP: write data when subcycling is used (outdated)

BenjaminRodenberg:
## Main changes of this PR

Allows to write data to preCICE when subcycling is used. This either results in a piecewise constant or linear interpolation. Possibly also in higher-order interpolation.

## Motivation and additional information

See #1171 

This is a rebased version of https://github.com/precice/precice/pull/1414, based on https://github.com/precice/precice/pull/1520.

## Author's checklist

* [x] merge #1422 
* [x] merge #1445 
* [x] merge https://github.com/precice/precice/pull/1455
* [x] merge https://github.com/precice/precice/pull/1456 ? (or directly do this here?)
* [x] what to do with #1432 (Continue working on https://github.com/BenjaminRodenberg/precice/pull/13 until then)
* [x] merge https://github.com/precice/precice/pull/1507
* [x] merge #1504 
* [x] merge #1520 
* [ ] https://github.com/precice/precice/pull/1639
* [x] merge #1503 
* [x] merge #1555 
* [x] merge https://github.com/precice/precice/pull/1545
* [x] merge https://github.com/precice/precice/pull/1538
* [x] merge https://github.com/precice/precice/pull/1526 or delete test + feature
* [ ] Add tests for MultiCoupling: A write mapping should work in the context of multi coupling, even, if the sent meshes in time do not agree. For making a read mapping in this situation work as well, we need to be able to refer to received times etc. in an individual fashion (currently a coupling scheme only gives access to the received times globally).
* [x] Add test for Direct Mesh Access + Waveform relaxation (done: https://github.com/precice/precice/pull/1584)
* [ ] Add test for Participant First method + Waveform relaxation (partially done: 3e82443, rest https://github.com/precice/precice/issues/1570)
* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [ ] I added a test to cover the proposed changes in our test suite.
* [ ] I ran `make format` to ensure everything is formatted correctly.
* [ ] I sticked to C++14 features.
* [ ] I sticked to CMake version 3.16.3.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

PR #1526, opened 2023-01-02
Fix incorrect read data in watch integral test

BenjaminRodenberg:
## Main changes of this PR

Removes unnecessary read data `"DataTwo"`.

`DataTwo` looks unnecessary to me and it's hard to understand what's the purpose from looking at the configuration. Is it necessary for the watch integral to work?

Additional issue: It looks like preCICE is not checking for read data that that is actually never receiving any data from any participant. Is there a use case for this setup? If not, I think we should have a check for to avoid this faulty config.

## Author's checklist

## Reviewers' checklist


BenjaminRodenberg:
From the results of the CI it looks like the `"DataTwo"` needs to be defined and used for the test to work properly. I still think the user interface here is not very clear. @davidscn Any opinion? Should we open an issue here?

tests/serial/watch-integral/WatchIntegralScaleAndNoScale.xml
```diff
@@ -26,7 +24,6 @@
 
     <participant name="SolverTwo">
       <provide-mesh name="MeshTwo" />
-      <read-data name="DataTwo" mesh="MeshTwo" />
```
davidscn, line 29: I guess this needs to be `DataOne`, right? Otherwise, which data are you actually 'watching' here (as there is also no write data tag.

davidscn (commented):
> Additional issue: It looks like preCICE is not checking for read data that that is actually never receiving any data from any participant. Is there a use case for this setup? If not, I think we should have a check for to avoid this faulty config.

This should go into an issue, I guess. Not sure how easily we can check this.

tests/serial/watch-integral/WatchIntegralScaleAndNoScale.xml
BenjaminRodenberg: Not sure. I don't know what this test is supposed to be doing. I just stumbled over the config and found it odd. I currently don't have time to play around with this test and try to thoroughly understand it. Feel free to play around with it, if you want. Maybe some input from @fsimonis?

BenjaminRodenberg:
> > Additional issue: It looks like preCICE is not checking for read data that that is actually never receiving any data from any participant. Is there a use case for this setup? If not, I think we should have a check for to avoid this faulty config.
> 
> This should go into an issue, I guess. Not sure how easily we can check this.

Done. https://github.com/precice/precice/issues/1527

uekerman:
> Is there a use case for this setup? If not, I think we should have a check for to avoid this faulty config.

There is, I answered in #1527

Still, the test looks wrong.

tests/serial/watch-integral/WatchIntegralScaleAndNoScale.xml
uekerman: Should probably be `DataOne`, yes.

BenjaminRodenberg: Fixed in [f172994](https://github.com/precice/precice/pull/1526/commits/f17299490b3a2979595f9459c3cf870082bb1a35)

BenjaminRodenberg:
I tried to fix the test as far as I can. Does anybody know what this test is supposed to be doing? I don't have the time to go deep into watch integrals and currently it is failing (probably because the test was always wrong). If nobody wants to take over responsibility here, I would suggest to delete the entire test ( + potentially the feature as well, because this makes the feature untested), because it is currently causing me headaches in #1523.

tests/serial/watch-integral/WatchIntegralScaleAndNoScale.cpp
```diff
@@ -76,26 +76,26 @@ BOOST_AUTO_TEST_CASE(WatchIntegralScaleAndNoScale)
...
+      interface.readScalarData(dataOneID, idC, valueC);
 
       dt = interface.advance(dt);
     }
     interface.finalize();
 
     {
       std::string fileName = "precice-SolverTwo-watchintegral-WatchIntegral.log";
-      auto        result   = readDoublesFromTXTFile(fileName, 4);
+      auto        result   = readDoublesFromTXTFile(fileName, 3);
```
BenjaminRodenberg, line 94: Magic number? I don't understand what this is doing.

tests/serial/watch-integral/WatchIntegralScaleAndNoScale.cpp
```diff
@@ -76,26 +76,26 @@ BOOST_AUTO_TEST_CASE(WatchIntegralScaleAndNoScale)
...
       auto        expected = std::vector<double>{
-          1.0, 9.5, 0.0, 3.0,
-          2.0, 12.5, 0.0, 3.0,
-          3.0, 12.5, 0.0, 3.0};
+          1.0, 9.5, 3.0,
+          2.0, 12.5, 3.0,
+          3.0, 0.0, 3.0};
```
BenjaminRodenberg, line 98: Took these numbers from `precice-SolverTwo-watchintegral-WatchIntegral.log`. Cannot judge whether this is correct or not.

BenjaminRodenberg (commented):
I decided to fix this test for my PR. If somebody can review and judge whether the output makes sense, we can probably merge it.

tests/serial/watch-integral/WatchIntegralScaleAndNoScale.cpp
uekerman: It is skipping the first line of the log file. You could maybe call the variable `numberOfEntriesToSkip`.

tests/serial/watch-integral/WatchIntegralScaleAndNoScale.cpp
```diff
@@ -108,11 +108,11 @@ BOOST_AUTO_TEST_CASE(WatchIntegralScaleAndNoScale)
...
       std::string fileName = "precice-SolverTwo-watchintegral-WatchIntegralNoScale.log";
-      auto        result   = readDoublesFromTXTFile(fileName, 4);
+      auto        result   = readDoublesFromTXTFile(fileName, 3);
       auto        expected = std::vector<double>{
-          1.0, 9.0, 0.0, 3.0,
-          2.0, 12.0, 0.0, 3.0,
-          3.0, 12.0, 0.0, 3.0};
+          1.0, 9.0, 3.0,
+          2.0, 12.0, 3.0,
+          3.0, 12.0, 3.0};
```
uekerman, line 115:
```suggestion
          1.0, 6.0, 3.0,
          2.0, 9.0, 3.0,
          3.0, 12.0, 3.0};
```
... is what I would actually expect.

uekerman (changes requested):
The feature seems to give wrong results over time.

> And: potentially delete the feature as well, because this makes the feature untested.

There are many unit tests.

tests/serial/watch-integral/WatchIntegralScaleAndNoScale.cpp
BenjaminRodenberg: Ok. Makes sense. 3 columns * 1 line :+1:

tests/serial/watch-integral/WatchIntegralScaleAndNoScale.cpp
BenjaminRodenberg: I opened an issue: https://github.com/precice/precice/issues/1579

BenjaminRodenberg:
@uekerman We can merge this PR, right? I opened an issue #1579 that we don't forget that the feature is actually buggy.

uekerman (approved):

PR #1535, opened 2023-01-12
Introduce _allData in BaseCouplingScheme to avoid duplicates

BenjaminRodenberg:
* `_allData` is used to collect all `CouplingData` of a `CouplingScheme`
* Avoids duplicates, especially for `MultiCouplingSchemes`
* In cases where a single `CouplingData` is used at multiple places (e.g. in a `MultiCouplingScheme` where the same data is sent and/or received to different participants), multiple pointers will exist.

## Main changes of this PR

* add `_allData` to collect all `CouplingData`
* use `PtrCouplingData` to allow to point from different places to the same data

In #1523 maintaining `CouplingData` becomes more involved. I observed that the old approach for storing `CouplingData` in a `CouplingScheme` created duplicates under certain conditions.

## Author's checklist

* [x] I added a changelog file with `make changelog` if there are user-observable changes since the last release. (no user obervable changes)
* [x] I added a test to cover the proposed changes in our test suite. (no new feature, not visible before #1523)
* [x] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html). (not breaking)

## Reviewers' checklist

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -22,6 +22,7 @@
 #include "mesh/Mesh.hpp"
 #include "precice/types.hpp"
 #include "utils/EigenHelperFunctions.hpp"
+#include "utils/Helpers.hpp"
 #include "utils/IntraComm.hpp"
 #include "utils/assertion.hpp"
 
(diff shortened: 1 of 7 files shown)


uekerman (approved):
Nice refactoring, I don't see any potential problems.

PR #1536, opened 2023-01-13
Remove mapping timings

### LINKED ISSUE #1364: Remove deprecated mapping timings

davidscn:
## Main changes of this PR
Closes #1364 and remotely related to the clean-up in #1417.

## Author's checklist

* [x] <del> I added a test to cover the proposed changes in our test suite. </del>

## Reviewers' checklist

### DIFF
--- a/src/precice/impl/SolverInterfaceImpl.cpp
+++ b/src/precice/impl/SolverInterfaceImpl.cpp
@@ -651,12 +651,6 @@ void SolverInterfaceImpl::resetMesh(
   PRECICE_TRACE(meshID);
   PRECICE_VALIDATE_MESH_ID(meshID);
   impl::MeshContext &context = _accessor->usedMeshContext(meshID);
-  /*
-  bool               hasMapping = context.fromMappingContext.mapping || context.toMappingContext.mapping;
-  bool               isStationary =
-      context.fromMappingContext.timing == mapping::MappingConfiguration::INITIAL &&
-      context.toMappingContext.timing == mapping::MappingConfiguration::INITIAL;
-  */
 
   PRECICE_DEBUG("Clear mesh positions for mesh \"{}\"", context.mesh->getName());
   _meshLock.unlock(meshID);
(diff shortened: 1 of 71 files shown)


src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1816,10 +1805,7 @@ void SolverInterfaceImpl::clearMappings(std::vector<MappingContext> &contexts)
   // Clear non-stationary, non-incremental mappings
   using namespace mapping;
   for (impl::MappingContext &context : contexts) {
-    bool isStationary = context.timing == MappingConfiguration::INITIAL;
-    if (not isStationary) {
-      context.mapping->clear();
-    }
+    context.mapping->clear();
```
uekerman, line 1808: This seems wrong. Doesn't this lead to a re-computation of the mapping in every iteration?

uekerman (changes requested):
Thanks for pushing this!
I have the feeling we could remove more. Is `hasMappedData` still required? For what reason?
If no, `isMappingRequired` still required?

src/precice/impl/SolverInterfaceImpl.cpp
davidscn: This would clear mapping if we call `SolverInterfaceImpl::clearMappings`, which seems to be the way to go here.

davidscn:
> Is hasMappedData still required?

I'm not sure. I can mostly answer 'how to', not when to. In sourcetrail, the boolean looks rather stray, yes.

src/precice/impl/SolverInterfaceImpl.cpp
fsimonis: By dropping the timings, preCICE can internally handle the mappings whenever it needs to (Partitioning), meaning there is also no need to collectively clear mappings. Hence, the complete `clearMappings` function can go.

davidscn: Done

uekerman (approved):
Looks good :+1:

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1796,59 +1790,38 @@ void SolverInterfaceImpl::computeMappings(std::vector<MappingContext> &contexts,
...
   PRECICE_TRACE();
   computeMappings(_accessor->writeMappingContexts(), "write");
   for (auto &context : _accessor->writeDataContexts()) {
-    if (context.isMappingRequired()) {
+    if (context.hasMapping()) {
       PRECICE_DEBUG("Map write data \"{}\" from mesh \"{}\"", context.getDataName(), context.getMeshName());
       context.mapData();
     }
   }
-  clearMappings(_accessor->writeMappingContexts());
```
BenjaminRodenberg, line 1837: @fsimonis Wasn't there a reason why we cleared the mapping here? I remember a discussion we once had here when I wanted to remove this line at some point in time...

PR #1514, opened 2022-12-13
Add muscle tests

fsimonis:
## Main changes of this PR

This PR adds muscle tests from @carme-hp.

## Motivation and additional information

These tests add:
* a complex compositional scheme
* a compositional test with MultiCouplingScheme

Currently, the two muscle test hangs in initialize. ~~@uekerman is looking into this.~~

The one muscle test still needs to handle requirements correctly. @carme-hp have you made any progress regarding this test?

Related to #1478 #1479 

## Author's checklist

* [ ] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [ ] I added a test to cover the proposed changes in our test suite.
* [ ] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).
* [ ] I sticked to C++17 features.
* [ ] I sticked to CMake version 3.16.3.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

PR #1541, opened 2023-01-16
Fix composition of compositional coupling

### LINKED ISSUE #1534: Raw assertion for illegal configuration of compositional coupling scheme
**Describe your setup**

latest state of develop: 8f09ec2 (+ some unrelated changes on my own fork)

**Describe the problem**

I just tried to run `Integration/Serial/threeSolvers/ThreeSolversImplicitExplicit` with a modified configuration using two implicit coupling schemes:

```
<precice-configuration>
  <solver-interface dimensions="2">
    ...
      <max-time-windows value="10" />
      <max-iterations value="10" />
      <min-iteration-convergence-measure min-iterations="3" data="Data1" mesh="MeshA" />
      <exchange data="Data0" mesh="MeshA" from="SolverOne" to="SolverTwo" />
      <exchange data="Data1" mesh="MeshA" from="SolverTwo" to="SolverOne" />

      <max-time-windows value="10" />
      <max-iterations value="10" />
      <min-iteration-convergence-measure min-iterations="3" data="Data0" mesh="MeshB" />
      <exchange data="Data0" mesh="MeshB" from="SolverOne" to="SolverThree" />
      <exchange data="Data2" mesh="MeshB" from="SolverThree" to="SolverOne" />
  </solver-interface>
</precice-configuration>
```

Since #1462 this is illegal:

> allow at most one implicit scheme

As expected this configuration does not work, but I run into a raw assertion that is hard to understand for users. A proper preCICE check would be better here.

**Step To Reproduce**
1. Modify `tests/serial/three-solvers/ThreeSolversImplicitExplicit.xml` as suggested above
2. Run `precice/build$ mpirun -np 4 ./testprecice -t Integration/Serial/ThreeSolvers/ThreeSolversImplicitExplicit --log_level=all`

We observe that the following assertion is triggered:

```
ASSERTION FAILED
Location:   void precice::cplscheme::CompositionalCouplingScheme::addCouplingScheme(const PtrCouplingScheme&)
File:      precice/src/cplscheme/CompositionalCouplingScheme.cpp:38
Expression: _implicitScheme == nullptr
Rank:       0
Arguments:  none
```

**Expected behaviour**

A error message that is more understandable for a user.

fsimonis:
## Main changes of this PR

Fixes the composition of multiple coupling schemes.
Also simplifies and hardens the code.

Closes #1534

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/cplscheme/config/CouplingSchemeConfiguration.cpp
+++ b/src/cplscheme/config/CouplingSchemeConfiguration.cpp
@@ -372,33 +372,32 @@ void CouplingSchemeConfiguration::addCouplingScheme(
     const std::string &      participantName)
 {
   PRECICE_TRACE(participantName);
-  if (utils::contained(participantName, _couplingSchemes)) {
-    PRECICE_DEBUG("Coupling scheme exists already for participant");
-    if (utils::contained(participantName, _couplingSchemeCompositions)) {
-      PRECICE_DEBUG("Coupling scheme composition exists already for participant");
-      // Fetch the composition and add the new scheme.
-      PRECICE_ASSERT(_couplingSchemeCompositions.count(participantName) == 1);
-      auto composition = _couplingSchemeCompositions.at(participantName);
-      PRECICE_CHECK(!cplScheme->isImplicitCouplingScheme() || !composition->isImplicitCouplingScheme(),
-                    "You attempted to define a second implicit coupling-scheme for the participant \"{}\", which is not allowed. "
-                    "Please use a multi coupling-scheme for true implicit coupling of multiple participants.",
-                    participantName);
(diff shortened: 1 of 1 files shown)


fsimonis:
This could also be part of the issue why #1514 fails

uekerman:
I don't understand the fix yet. Don't both versions do the same things?

fsimonis:
When composing a couplingscheme [here](https://github.com/precice/precice/pull/1541/files#diff-cb7b50384ae536aad8d8e561e25e510a6f84bd3b95c357c4532a7f27be8e34c4L388-L395), it isn't registered in `_couplingSchemeCompositions`.
So we always compose.

uekerman (approved):

PR #1545, opened 2023-01-17
Fix multi coupling configs that use non-central controller

uekerman:
## Main changes of this PR

While discussing waveform stuff with @BenjaminRodenberg, we saw that some of our multi-coupling configuration files do not make sense completely. This PR fixes these configs.

## Author's checklist

## Reviewers' checklist


BenjaminRodenberg (approved):
Looks ok. Also no alert signs when using these modified configs with #1523 :+1:

PR #1547, opened 2023-01-19
Simplify FindPETSc to directly use pkgconfig target

fsimonis:
## Main changes of this PR

This PR simplifies the FindPETSc.cmake module to directly use pkg-config to generate a target.
It removes all kinds of guesswork based on `PETSC_DIR` and `PETSC_ARCH`.
The user now has to ensure that `pkg-config --libs --cflags PETSc` returns meaningful results. This also works for in-source builds without installing PETSc.

Simplify the code and reduce overhead on our sides.
Directly using the target simplifies the library handling by **a lot**.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/cmake/modules/FindPETSc.cmake
+++ b/cmake/modules/FindPETSc.cmake
@@ -1,7 +1,7 @@
 # FindPETSc
 # ---------
 #
-# Locates the PETSc library using pkg-config
+# Locates the PETSc library using pkg-config module PETSc
 #
 # Imported Targets
 # ^^^^^^^^^^^^^^^^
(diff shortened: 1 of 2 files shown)


fsimonis:
Tested it with spack@53d5011192 and it works.

fsimonis:
@davidscn Given that you are regularly benchmarking preCICE on clusters, could you try this PR to see if you run into any issues?

davidscn (approved):
Tested it on SuperMUC and our IPVS machines and it works. Supermuc explicitly exports the pkg-config in the module; on our machine, I had a local installation and the pkg-config path was already exported due to preCICE. Just two comments:

- PETSc exports two versions of the file: `PETSc.pc` and `petsc.pc`. Do we support both or only one? I'm not sure about best practices here.
- I guess what works would be either 1) handing over a prefix path to the PETSc build and then installing PETSc (and exporting PKG_CONFIG) or 2) not handing over a prefix path and then exporting the pc file in the build tree. However, I assume that handing over a prefix path when compiling and exporting the pc file from the build tree would not work as the pc file would refer to the prefix path locations (that's also what I see in my local build tree). Maybe this would be worth documenting somewhere.

Also, this change is breaking, as people who solely relied on any env export mechanisms so far would run into potential issues. I would expect an entry in the docs regarding the change.

fsimonis:
PETSc creates and installs both pc files. There doesn't seem to be a reason to prefer one over the other. 
The Fedora packages used to have a bug resulting in different content of both files, but this was fixed after I reported it.
I didn't run into other problems so far.

PETSc generates the pc files based on the prefix given during configuration. This could lead to problems and this is definitely something we should document.

Good point regarding the breaking nature of this.

PR #1574, opened 2023-02-20
Add missing xml error handlers

### LINKED ISSUE #1564: Silent crash for mismatching opening and endig tags
preCICE Version: 1965ceb

**Describe the problem**
In our RBF mappings, we have now closing and ending tags due to the basis-function being a subtag (related to #1417). However, this leads to the situation that  we have multiple 'legal' opening tags (e.g. `rbf-global-iterative` ...) and multiple 'legal' closing tags. 
Mixing the opening and the closing tag leads to a silent crash of preCICE without any useful information (even with trace log enabled).

**Step To Reproduce**
Use a config with mismatching ending and closing tags for the mapping. Example:
```xml
      <mapping:rbf
        direction="write"
        from="Fluid-Mesh"
        to="Solid-Mesh"
        constraint="conservative" >
         <basis-function:compact-polynomial-c0 support-radius="1" />
      </mapping:rbf-global-direct>
```

**Expected behaviour**
A clear error message.

fsimonis:
## Main changes of this PR

This PR adds missing error handlers for libxml2.

Tag mismatches are correctly handled now.

```
ERROR: Opening and ending tag mismatch: configuration line 2 and figuration
```

Closes #1564

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/1574.md
+++ b/docs/changelog/1574.md
@@ -1,0 +1,1 @@
+- Fixed missing error on mismatched closing xml tags
--- a/src/logging/LogConfiguration.cpp
+++ b/src/logging/LogConfiguration.cpp
@@ -217,7 +217,7 @@ void setupLogging(LoggingConfiguration configs, bool enabled)
   // We need to exit after the sink removal as the default sink exists before
   // the log configuration is parsed.
   if (auto noconfigs = std::none_of(configs.begin(), configs.end(), [](const auto &config) { return config.enabled; });
-      !enabled || noconfigs) {
+      !enabled && noconfigs) {
     auto sink = boost::make_shared<NullSink>();
     boost::log::core::get()->add_sink(sink);
     activeSinks.emplace_back(std::move(sink));
--- a/src/xml/ConfigParser.cpp
+++ b/src/xml/ConfigParser.cpp
@@ -93,6 +93,16 @@ void OnStructuredErrorFunc(void *userData, xmlError *error)
   ConfigParser::MessageProxy(error->level, message);
 }
 
+void OnErrorFunc(void *userData, const char *error, ...)
+{
+  ConfigParser::MessageProxy(XML_ERR_ERROR, error);
+}
+
+void OnFatalErrorFunc(void *userData, const char *error, ...)
+{
+  ConfigParser::MessageProxy(XML_ERR_FATAL, error);
+}
+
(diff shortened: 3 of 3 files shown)


davidscn (changes requested):
I built preCICE in release and added the following lines to my config

```xml
      <mapping:rbf
        direction="read"
        from="Solid-Mesh"
        to="Fluid-Mesh"
        constraint="consistent" >
         <basis-function:compact-polynomial-c0 support-radius="1" />
      </mapping:rbf-global-iterative>
```

which still triggers the silent crash, i.e., the mismatch is not detected properly.

This might be related to the mapping, though, es we have a `nullptr` in the mapping object for the RBF mapping. The mapping is only instantiated in the subtag, see
https://github.com/precice/precice/blob/fcefb9d6620f8d4f8dfefa574bbcd44002ce7adf/src/mapping/config/MappingConfiguration.cpp#L453-L455

fsimonis:
@davidscn I get the following error for your config in both Debug and Release builds:

```
ERROR: Opening and ending tag mismatch: rbf line 28 and rbf-global-iterative
```

davidscn:
I only get the error message you mention when using `precice-tools`.
Running _the same_ config with a real solver still leads to the silent crash (tested with OpenFOAM and deal.II). Even the debug build fails immediately, it seems that preCICE dies on the spot. Have you tried with a real solver?

fsimonis:
This took me some while to reproduce, but I am finally managed it.


perpendicular-flap with {fluid-nutils, solid-dealii} doesn't output any error in case of mismatching tags, but it silently crashes.

Using the `log.conf` to configure an output stream works and the error is nicely displayed.
```
[HelpMe]
Filter = 
Type = stream
output = stdout
```

Hence, we are running into an old problem: 
During the parsing stage of the configuration, preCICE doesn't have a logger setup.
Hence, error messages are not displayed.

Tests and `precice-tools check` create a logger beforehand, hence it doesn't affect them.
Hence, this is only visible the using solvers that link to the library.
So, this can easily be reproduced using the solverdummies.

The behaviour was most likely introduced with #1305

fsimonis:
Fixed the bug here. I am not adding a changelog entry ass this was introduced after the last release.

davidscn (approved):
Works now, nice!

PR #1578, opened 2023-02-22
Fix PETSc mappings crashing when checking the config

### LINKED ISSUE #1562: config-checker crashes when checking PETSc RBF mappings
**Describe your setup**

Operating system (e.g. Linux distribution and version): Ubuntu 20.04 (VM) & Ubuntu 22.04
preCICE Version: 2.5.0 + develop

**Describe the problem**
`precice-tools check` crashes if the config includes an RBF mapping and preCICE is build with PETSc.

```
Checking precice-config.xml for syntax and basic setup issues...
*** The MPI_Comm_rank() function was called before MPI_INIT was invoked.
*** This is disallowed by the MPI standard.
*** Your MPI job will now abort.
[tickly:111257] Local abort before MPI_INIT completed completed successfully, but am not able to aggregate error messages, and not able to guarantee that all other processes were killed!
```

**Step To Reproduce**
1. Pick a configuration that uses `<mapping:rbf-X />` 
3. Run `precice-tools check precice-config.xml`
4. Notice the crash

**Expected behaviour**
No crash

**Additional context**

The PETRBFMapping accesses the `rank()` in the constructor.
MPI is not initialized when calling `tooling::checkConfiguration`.

**Likely fix**
In `src/precice/Tooling.cpp` add `include<utils/Parallel.hpp` and in `checkConfiguration`:

```diff
+  utils::Parallel::initializeMPI(nullptr, nullptr);
  xml::configure(config.getXMLTag(), context, filename);
  fmt::print(fmt::emphasis::bold | fg(fmt::color::green), "No major issues detected\n", filename);
+  utils::Parallel::finalizeMPI();
```

fsimonis:
## Main changes of this PR

This PR fixes a crash when checking configurations that require PETSc mappings.

The constructor request the MPI ranks, which leads to a crash.

Closes #1562

## Author's checklist

### DIFF
--- a/src/utils/Parallel.cpp
+++ b/src/utils/Parallel.cpp
@@ -2,6 +2,7 @@
 #include <algorithm>
 #include <map>
 #include <memory>
+#include <mpi.h>
 #include <numeric>
 #include <ostream>
 #include <utility>
@@ -185,6 +186,17 @@ void Parallel::pushState(CommStatePtr newState)
 //   return nullptr;
 // #endif
 // }
+
+bool Parallel::isMPIInitialized()
+{
+#ifndef PRECICE_NO_MPI
+  int isMPIInitialized{-1};
+  MPI_Initialized(&isMPIInitialized);
+  return isMPIInitialized != 0;
+#else
+  return false;
+#endif // not PRECICE_NO_MPI
+}
 
 void Parallel::initializeManagedMPI(
     int *   argc,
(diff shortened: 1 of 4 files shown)


src/utils/Parallel.cpp
```diff
@@ -2,6 +2,7 @@
 #include <algorithm>
 #include <map>
 #include <memory>
+#include <mpi.h>
```
fsimonis, line 5:
```suggestion
```

Issue #1484, opened 2022-11-22
Implementing GPU support for data mapping methods


Timo-Schrader:
## Motivation
With increasing number of available GPUs in HPC clusters, it becomes clear that utilizing their immense data-parallel calculation power can significantly increase specific calculations in preCICE. 
For example, RBF data mapping consists of building the system matrix and solving the resulting system of linear equations. The first step is purely data-parallel, whereas the second one can also potentially accelerated using General-Purpose GPU programming.

Making use of GPUs in data-mapping methods such as RBF and barycentric interpolation (and further calculations) is to be investigated (in my thesis).

**Describe the solution you propose.**
We want to make use of the numerical algebra library Ginkgo (https://github.com/ginkgo-project/ginkgo). It offers a lot of numerical features out-of-the-box, e.g., pre-conditioning, matrix data structures and system solvers. Furthermore, it supports the usage of different HPC backends (OpenMP, CUDA, AMD HIP, ...) without having to change the self-written code. 

The goal is to use Ginkgo for all parallelization efforts. Currently, we are in discussion with the Ginkgo authors about how to bring their nice feature of kernel dispatches into the main branch (for more information cf. https://github.com/ginkgo-project/ginkgo/discussions/1190 ). We believe that Ginkgo offers a great support and a wide variety of HPC backends and maintenance. 

**Describe alternatives you've considered**
We investigated whether using Kokkos might be an alternative to Ginkgo. However, Kokkos seems to represent good HPC, but not great numeric library, whereas Ginkgo seems to be much more self-contained when it comes to numerical applications.

**Additional context**
This project is done in the context of my Master's thesis at the University of Stuttgart (supervised by @davidscn ).

fsimonis:
> Making use of GPUs in data-mapping methods such as RBF and barycentric interpolation (and further calculations) is to be investigated (in my thesis).

Just a hint that barycentric interpolations are quite inexpensive and most likely not worth porting to GPUs. Most of them are fundamentally a combination of random access and weighted sums.

You could absolutely go all-in in terms of RBFs, as they are occasionally prohibitively expensive. Local RBFs should also benefit from sparse matrices. So, there is a lot of information already in preCICE that could be used for potential improvements.

Timo-Schrader:
Thank you for your input! 

[...]

PR #1581, opened 2023-02-23
Added Ginkgo + GPU support for RBF data mapping

Timo-Schrader:
## Main changes of this PR
**PR for reviewing and discussion purposes**
Addresses #1484 

This PR proposes GPU support via the linear algebra library [Ginkgo](https://github.com/ginkgo-project/ginkgo). More precisely, it enables RBF interpolation data mapping to be solved with the help of Nvidia and AMD GPUs. Our Ginkgo solver uses the pre- and postprocessing code from the Eigen solver, but uses all values, puts it into Ginkgo data structures and does everything on the GPU (or CPU using OpenMP). Therefore, it introduces almost no changes to already existing code.

The following features are implemented:

* Support for Nvidia and AMD GPUs and OpenMP (data-parallel paradigm for CPUs)
* Highly optimized matrix assembly kernels for GPUs that are much faster in constructing these matrices
* CG, GMRES and Multigrid Solver (however, the last one is very likely to be removed as stand-alone solver)
* (Block-)Jacobi, Ic and and Ilu preconditioner (best configuration not clear yet)
* XML configuration tags enabling selection of executor (sequential, OpenMP, Nvidia CUDA (and AMD HIP)), solver, preconditioner, stopping criteria

GPUs offer a high amount of data-parallel calculation units. Therefore, we want to make use of them when it comes to setting up very large interpolation matrices, solving linear systems of equations and executing [BLAS](https://de.wikipedia.org/wiki/Basic_Linear_Algebra_Subprograms) routines. Ginkgo provides execution backends for both CPU (sequential and parallel) and GPU, which makes it very easy to run cross-platform. However, we currently make use of experimental Ginkgo features, so in order to test it, you need branch [a195f856ee](https://github.com/ginkgo-project/ginkgo/tree/a195f856ee946b1b9387d30513dd992df4501df3). You can enable it in `CMakeLists.txt` using the `PRECICE_GinkgoParallelization` flag.

## Future Work and Todos

There are still some open aspects that need to be addressed before this feature can be merged into preCICE:

* Test cases covering all Ginkgo configurations
* Multigrid preconditioning is currently very slow
* Conservative mapping is not implemented yet
* Ginkgo does not have the `run_kernel` public interface in the main branch yet.
* Renaming some newly introduced flags and variables to make it more consistent
* Cover build process in Github Actions pipleline
* Probably a few more

## Author's checklist

## Reviewers' checklist


Timo-Schrader:
Please feel free to discuss! I'd be glad about any feedback!

src/mapping/ginkgo_kernels/GinkgoRBFKernels.cpp
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        dist = sqrt(dist);
+
+#else
+        const unsigned int supportPointOffset = dataDimensionality * j; // Point of current column
+        const unsigned int targetPointOffset  = dataDimensionality * i; // Point of current row
+        // Loop over each dimension and calculate euclidian distance
+        for (size_t k = 0; k < dataDimensionality; ++k) {
+          dist += std::pow(supportPoints[supportPointOffset + k] - targetPoints[targetPointOffset + k], 2) * static_cast<int>(activeAxis.at(k));
```
davidscn, line 110: No `pow_int` here?

src/mapping/ginkgo_kernels/GinkgoRBFKernels.cpp
```diff
@@ -0,0 +1,188 @@
...
+            y    = supportPoints[k * inputRowLength + j] - prefetchedEvalPoint[k]; // <- This should not impose a bank conflict since CUDA offers broadcasting if a warp accesses the same shared memory adress
+            dist = fma(y, y, dist);
+          }
+        }
+
+        dist = sqrt(dist);
+
+#else
+        const unsigned int supportPointOffset = dataDimensionality * j; // Point of current column
+        const unsigned int targetPointOffset  = dataDimensionality * i; // Point of current row
```
davidscn, line 107: Above you write 64 bit ints are required? Is `unsigned int` enough here?

src/mapping/ginkgo_kernels/GinkgoRBFKernels.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+        dist                                  = std::sqrt(dist);
+#endif
+
+        mtx[i * rowLength + j] = f(dist, rbf_params);
+      },
+      gko::dim<2>{n1, n2}, n2, dataDimensionality, activeAxis, mtx, supportPoints, targetPoints, f, rbf_params, inputRowLength, outputRowLength, addPolynomial, extraDims);
+}
+
+// Here, we need to instantiate all possible variants for each basis function
```
davidscn, line 121: I understand that this is necessary by the way Ginkgo works. Internally, they often use macro expansions for such instantiations. I would also consider doing this here as editing the kernel function signature will become very tedious. However, feel free to decide this on your own, just a comment.

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -1,5 +1,25 @@
 #pragma once
 
+#ifdef __NVCC__
```
davidscn:
```suggestion
// the basis functions are parsed by the host and the device compiler
#ifdef __NVCC__
```

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -1,5 +1,25 @@
...
+
+#include <cuda_runtime.h>
+#define BOOST_PP_VARIADICS 1
+#define PRECICE_HOST_DEVICE __host__ __device__
+#define PRECICE_MEMORY_SPACE __device__
```
davidscn, line 8:
```suggestion
#define PRECICE_DEVICE_MEMORY __device__
```
the memory space itself could be both, no?

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -52,8 +72,26 @@ class ThinPlateSplines : public NoCompactSupportBase,
...
+
+  PRECICE_HOST_DEVICE inline double operator()(const double radius, const std::array<double, 3> &params) const
+  {
+    // We don't need to read any values from params since there is no need here
+    return std::log(std::max(radius, NUMERICAL_ZERO_DIFFERENCE)) * math::pow_int<2>(radius);
+  }
+
+#ifndef PRECICE_NO_GINKGO
+
+  std::array<double, 3> getFunctionParameters()
```
davidscn: This function can always be `const`, right?

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -52,8 +72,26 @@ class ThinPlateSplines : public NoCompactSupportBase,
...
+    return this->operator()(radius, this->_params);
+  }
+
+  PRECICE_HOST_DEVICE inline double operator()(const double radius, const std::array<double, 3> &params) const
+  {
+    // We don't need to read any values from params since there is no need here
+    return std::log(std::max(radius, NUMERICAL_ZERO_DIFFERENCE)) * math::pow_int<2>(radius);
+  }
+
+#ifndef PRECICE_NO_GINKGO
```
davidscn, line 101: I would add the function unconditionally, shouldn't hurt

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -52,8 +72,26 @@ class ThinPlateSplines : public NoCompactSupportBase,
 public:
   double evaluate(double radius) const
   {
-    return std::log(std::max(radius, math::NUMERICAL_ZERO_DIFFERENCE)) * math::pow_int<2>(radius);
+    return this->operator()(radius, this->_params);
```
davidscn, line 92:
```suggestion
    return operator()(radius, _params);
```
?

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -205,16 +334,36 @@ class CompactThinPlateSplinesC2 : public CompactSupportBase,
...
+
+  PRECICE_HOST_DEVICE inline double operator()(const double radius, const std::array<double, 3> &params) const
+  {
+    double       r_inv = params.at(0);
+    double const p     = radius * r_inv;
+    if (p >= 1) {
       return 0.0;
-    return 1.0 - 30.0 * math::pow_int<2>(p) - 10.0 * math::pow_int<3>(p) + 45.0 * math::pow_int<4>(p) - 6.0 * math::pow_int<5>(p) - math::pow_int<3>(p) * 60.0 * std::log(std::max(p, math::NUMERICAL_ZERO_DIFFERENCE));
+    } else {
+      return 1.0 - 30.0 * std::pow(p, 2) - 10.0 * std::pow(p, 3) + 45.0 * std::pow(p, 4) - 6.0 * std::pow(p, 5) - std::pow(p, 3) * 60.0 * std::log(std::max(p, NUMERICAL_ZERO_DIFFERENCE));
```
davidscn, line 375:
```suggestion
      return 1.0 - 30.0 * math::pow_int(p, 2) - 10.0 * std::pow(p, 3) + 45.0 * std::pow(p, 4) - 6.0 * std::pow(p, 5) - std::pow(p, 3) * 60.0 * std::log(std::max(p, NUMERICAL_ZERO_DIFFERENCE));
```
similar the others..

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -245,14 +397,32 @@ class CompactPolynomialC0 : public CompactSupportBase,
...
+
+  PRECICE_HOST_DEVICE inline double operator()(const double radius, const std::array<double, 3> &params) const
+  {
+    double       r_inv = params.at(0);
+    double const p     = radius * r_inv;
+    if (p >= 1) {
       return 0.0;
-    return math::pow_int<2>(1.0 - p);
+    } else {
+      return std::pow(1.0 - p, 2);
```
davidscn, line 437: Same as above

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -284,14 +457,32 @@ class CompactPolynomialC2 : public CompactSupportBase,
...
+
+  PRECICE_HOST_DEVICE inline double operator()(const double radius, const std::array<double, 3> &params) const
+  {
+    double       r_inv = params.at(0);
+    double const p     = radius * r_inv;
+    if (p >= 1) {
       return 0.0;
-    return math::pow_int<4>(1.0 - p) * (4 * p + 1);
+    } else {
+      return std::pow(1.0 - p, 4) * (4 * p + 1);
```
davidscn, line 500: same as above

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -323,14 +517,32 @@ class CompactPolynomialC4 : public CompactSupportBase,
...
+
+  PRECICE_HOST_DEVICE inline double operator()(const double radius, const std::array<double, 3> &params) const
+  {
+    double       r_inv = params.at(0);
+    double const p     = radius * r_inv;
+    if (p >= 1) {
       return 0.0;
-    return math::pow_int<6>(1.0 - p) * (35 * math::pow_int<2>(p) + 18 * p + 3);
+    } else {
+      return std::pow(1.0 - p, 6) * (35 * std::pow(p, 2) + 18 * p + 3, 2);
```
davidscn, line 563: There is a weird `,2` at the end. `pow_int` is missing as well.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -337,6 +364,10 @@ void MappingConfiguration::xmlTagCallback(
       delete[] arg;
 
       mapping.mapping = getRBFMapping<RBFBackend::PETSc>(basisFunction, constraintValue, mapping.fromMesh->getDimensions(), supportRadius, shapeParameter, _rbfConfig.deadAxis, _rbfConfig.solverRtol, _rbfConfig.polynomial, _rbfConfig.preallocation);
+
+#elif !defined(PRECICE_NO_GINKGO)
+      mapping.mapping = getRBFMapping<RBFBackend::Ginkgo>(basisFunction, constraintValue, mapping.fromMesh->getDimensions(), supportRadius, shapeParameter, _rbfConfig.deadAxis, _rbfConfig.polynomial, false, _ginkgoParameter);
```
davidscn, line 369: This needs to go into a separate if branch. If one builds preCICE with both, PETSc and ginkgo, it will override the mapping and initialize petsc. We could potentially discuss a bit about the config interface at the next coding days.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -195,10 +204,25 @@ MappingConfiguration::MappingConfiguration(
...
 
+  auto attrExecutor = makeXMLAttribute(ATTR_EXECUTOR, "reference-executor")
+                          .setDocumentation("Specifies the execution backend used by Ginkgo.")
+                          .setOptions({"reference-executor", "omp-executor", "cuda-executor", "hip-executor"});
+
+  auto attrSolver = makeXMLAttribute(ATTR_SOLVER, "cg-solver")
+                        .setDocumentation("Specifies the iterative solver used by Ginkgo.")
+                        .setOptions({"cg-solver", "gmres-solver", "mg-solver"});
+
+  auto attrPreconditioner = makeXMLAttribute(ATTR_PRECONDITIONER, "jacobi-preconditioner")
```
davidscn, line 215: I guess we also need a way to handle the GPU index. Let's discuss in person here.

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -65,19 +81,29 @@ class RadialBasisFctMapping : public RadialBasisFctBaseMapping<RADIAL_BASIS_FUNC
 
   /// Treatment of the polynomial
   Polynomial _polynomial;
+
+  /// Ginkgo Configuration
+  MappingConfiguration::GinkgoParameter _ginkgoParameter;
```
davidscn, line 86: Please fix the Wreorder here.

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -56,7 +67,12 @@ class RadialBasisFctMapping : public RadialBasisFctBaseMapping<RADIAL_BASIS_FUNC
 private:
   precice::logging::Logger _log{"mapping::RadialBasisFctMapping"};
 
+#ifndef PRECICE_NO_GINKGO
+  GinkgoRadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T> _ginkgoRbfSolver;
+#endif
```
davidscn, line 72: I would probably prefer a `std::conditional` here and make the backend-type a template argument, such that we have.

```cpp
SolverBackend = std::conditional<GinkgoRadialBasisFctSolver, RadialBasisFctSolver, useEigen>;
```
Should remove all the if's in the code and we avoid the (large) branches for the compiler.

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -6,13 +6,22 @@
...
 #include "mesh/Filter.hpp"
 #include "precice/types.hpp"
 #include "utils/Event.hpp"
 #include "utils/IntraComm.hpp"
 
+#ifndef PRECICE_NO_GINKGO
+
+#include <ginkgo/ginkgo.hpp>
+using GinkgoVector = gko::matrix::Dense<double>;
+using GinkgoMatrix = gko::matrix::Dense<double>;
```
davidscn, line 21: These variables seem unused here.

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -145,8 +171,17 @@ void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::computeMapping()
...
 
-    _rbfSolver = RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>{this->_basisFunction, globalInMesh, boost::irange<Eigen::Index>(0, globalInMesh.vertices().size()),
-                                                               globalOutMesh, boost::irange<Eigen::Index>(0, globalOutMesh.vertices().size()), this->_deadAxis, _polynomial};
+    if (_useEigen) {
+      _rbfSolver = RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>{this->_basisFunction, globalInMesh, boost::irange<Eigen::Index>(0, globalInMesh.vertices().size()),
+                                                                 globalOutMesh, boost::irange<Eigen::Index>(0, globalOutMesh.vertices().size()), this->_deadAxis, _polynomial};
+    }
+#ifndef PRECICE_NO_GINKGO
+    else {
+      _ginkgoRbfSolver.~GinkgoRadialBasisFctSolver();
```
davidscn, line 180: Why do we need to destroy the object here first? In the above version, we override the former object, which implicitly calls the destructor..

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -145,8 +171,17 @@ void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::computeMapping()
...
+      new (&_ginkgoRbfSolver) GinkgoRadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>{this->_basisFunction, globalInMesh, boost::irange<Eigen::Index>(0, globalInMesh.vertices().size()),
```
davidscn, line 181: Interesting syntax. I guess there is a good reason for that? We could make the whole object a `unique_ptr` in order to unify the syntax..

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -220,15 +255,35 @@ void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::mapConservative(DataID inpu
     // Construct Eigen vectors
     Eigen::Map<Eigen::VectorXd> inputValues(globalInValues.data(), globalInValues.size());
     Eigen::VectorXd             outputValues((this->output()->getGlobalNumberOfVertices()) * valueDim);
-    Eigen::VectorXd             in(_rbfSolver.getEvaluationMatrix().rows()); // rows == outputSize
+
+    Eigen::VectorXd in; // rows == outputSize
+
+    if (_useEigen) {
+      in.resize(_rbfSolver.getEvaluationMatrix().rows()); // rows == outputSize
```
davidscn, line 262: That's actually a flaw of the eigen solver. We should rather replace the eval getter by a size getter ( if we want to unify the code here).

davidscn (commented):
Really nice, thanks a lot. I left a few comments.  

We definitely need to integrate this into the RBF test suite and implement the `conservative` case as well (as you mentioned).

Also related to  https://github.com/ginkgo-project/ginkgo/pull/938

CMakeLists.txt
```diff
@@ -227,6 +228,33 @@ else()
...
 
+# Option: Ginkgo
+if (PRECICE_GinkgoParallelization)
+
+  # Check if CUDA is available on the system
+  include(CheckLanguage)
+  check_language(CUDA)
+
+  # Enable CUDA if driver and libraries are available; otherwise it'll be ignored
+  if(CMAKE_CUDA_COMPILER)
```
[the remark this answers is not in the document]
davidscn: I think we can use something linke `GINKGO_BUILD_CUDA` to check for the cuda awareness of ginkgo itself. There are even flags like `GINKGO_CUDA_ARCH_FLAGS` or `GINKGO_HIP_COMPILER_FLAGS` which might be able to replace the manual flags below (and ensure interoperability).

fsimonis:
I am having a look into the CMake side of things.

There are some issues with `ginkgo_add_unified_kernels`:
* Its cmake file isn't installed, so one needs the sources of ginkgo (tricky with spack and system packages)
* It generates a set of targets for the given target. Calling this multiple times per target will lead to errors. `T -> T_omp T_cuda ...`
* It currently isn't easy to set properties on these targets. Thus, includes need to be set globally, which is ugly. Alternatively, we could guess the targets and add properties ourselves.

It would simplify many things if they provided an output argument used to return the generated targets.

Ideally, this would look like this:
```cmake
find_package(Gingko REQUIRED)
ginkgo_add_unified_kernels(preciceCore _ginkgo_kernels src/mapping/ginkgo_kernels/GinkgoRBFKernels.cpp)
foreach(kernel in LISTS _ginkgo_kernels)
  target_include_directories($kernel PRIVATE src)
endforeach()
```

Looking at https://github.com/ginkgo-project/ginkgo/blob/a195f856ee946b1b9387d30513dd992df4501df3/cmake/unified_kernels.cmake , the implementation itself is rather trivial. It would be way less trouble to do this ourselves, which would simplify many things.

fsimonis:
An alternative to setting a global include for ginkgo kernels to work is to use `#include "@CMAKE_SOUCE_DIR@/src/mapping/..."` and the use [`configure_file()`](https://cmake.org/cmake/help/latest/command/configure_file.html?highlight=configure_file) to create an absolute path. This would go into the binary directory similar to how we deal with the version files.

CMakeLists.txt
```diff
@@ -227,6 +228,33 @@ else()
...
+    set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+    set(CMAKE_CUDA_SEPARABLE_COMPILATION ON)
+    include_directories(${EIGEN3_INCLUDE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/thirdparty/fmt/include)
+
+  endif()
+
+  # Check if OpenMP is available
+  find_package(OpenMP)
+  if(OpenMP_FOUND)
+    set(CMAKE_CXX_FLAGS "-fopenmp")
```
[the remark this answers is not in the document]
fsimonis: Also, this overrides all cxx flags.

Use `target_compile_options` (per target, preferred) or `add_compile_options` (globally) instead.

src/mapping/RadialBasisFctMapping.hpp
fsimonis: I would highly prefer the solver to be a template argument to the RadialBasisFctMapping.

In fact, this means that the RadialBasisFunction isn't even required by the mapping itself any more, as it is purely responsible to shovel data around and call the RBFSolver.

```diff
-template <typename RADIAL_BASIS_FUNCTION_T>
+template <typename RBFSolver>
```

Then you would just create the RBFSolver here

```suggestion
  RBFSolver _solver;
```

src/mapping/RadialBasisFctMapping.hpp
fsimonis: This is very dangerous territory! Placement new and in-place destruction are meant for dealing with uninitialized memory.

We should consider (in this order) reconstruction, `.reset()`/`.clear()`, `std::optional` (if you don't like heap memory), or `std::unique_ptr`.

CMakeLists.txt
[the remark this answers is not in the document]
Timo-Schrader: I just checked it: It builds without any error/warning message. However, during runtime, it'll crash, telling the user that the corresponding module has not been compiled in Ginkgo.

CMakeLists.txt
```diff
@@ -227,6 +228,33 @@ else()
...
+    set(CMAKE_CUDA_STANDARD_REQUIRED ON)
+    set(CMAKE_CUDA_FLAGS "-lineinfo --allow-unsupported-compiler --use_fast_math --std=c++17 --expt-extended-lambda --expt-relaxed-constexpr -Xcompiler -fopenmp -Wall")
+    set(CMAKE_POSITION_INDEPENDENT_CODE ON)
+    set(CMAKE_CUDA_SEPARABLE_COMPILATION ON)
+    include_directories(${EIGEN3_INCLUDE_DIR} ${CMAKE_CURRENT_SOURCE_DIR}/thirdparty/fmt/include)
+
+  endif()
+
+  # Check if OpenMP is available
+  find_package(OpenMP)
```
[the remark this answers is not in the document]
Timo-Schrader: See above.

CMakeLists.txt
```diff
@@ -320,6 +348,27 @@ else()
...
 
+# Setup Ginkgo
+if (PRECICE_GinkgoParallelization)
+  set(CMAKE_CXX_FLAGS "-fPIC")
+  include_directories(${CMAKE_CURRENT_SOURCE_DIR}/src)
+  set(Ginkgo_Path "" CACHE STRING "Path to the experimental Ginkgo Repository of branch a195f856e.") # TODO: Remove as soon as GPU feature is in main branch
+  find_package(Ginkgo REQUIRED HINTS "${Ginkgo_Path}/build" NO_CMAKE_SYSTEM_PATH)
+  target_include_directories(preciceCore PUBLIC "${Ginkgo_Path}/include")
+  include(${Ginkgo_Path}/cmake/unified_kernels.cmake)
+  file(GLOB ginkgo_lib "${Ginkgo_Path}/build/lib/*")
```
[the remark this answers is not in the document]
Timo-Schrader: Build directory. 
But this is a good point: I just checked in a bare VM and one issue might be that this unified_kernel.cmake seems not to be installed. We definitely need to address this when talking to the authors.

But as soon as this feature is in the main branch, we can hopefully just use `find_package()` without this hacky workaround.

(And my local path shouldn't have been committed..)

src/mapping/GinkgoRadialBasisFctSolver.hpp
```diff
@@ -0,0 +1,533 @@
...
+  /// Assembles the system matrices and computes the decomposition of the interpolation matrix
+  template <typename IndexContainer>
+  GinkgoRadialBasisFctSolver(RADIAL_BASIS_FUNCTION_T basisFunction, const mesh::Mesh &inputMesh, const IndexContainer &inputIDs,
+                             const mesh::Mesh &outputMesh, const IndexContainer &outputIDs, std::vector<bool> deadAxis, Polynomial polynomial,
+                             const MappingConfiguration::GinkgoParameter &ginkgoParameter);
+
+  ~GinkgoRadialBasisFctSolver() = default;
+
+  GinkgoRadialBasisFctSolver(const GinkgoRadialBasisFctSolver &solver) = delete;
+  GinkgoRadialBasisFctSolver &operator=(const GinkgoRadialBasisFctSolver &solver) = delete;
```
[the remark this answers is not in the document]
Timo-Schrader: I added them because I also added constructor and deconstructor. In my eyes, copying this class is not necessary in preCICE. Do you think it is? If so, I can remove it of course.

src/mapping/GinkgoRadialBasisFctSolver.hpp
```diff
@@ -0,0 +1,533 @@
...
+                                                                                                  {"omp-executor", [] { return gko::OmpExecutor::create(); }},
+                                                                                                  {"cuda-executor", [] { return gko::CudaExecutor::create(0, gko::OmpExecutor::create(), true, gko::allocation_mode::device); }},
+                                                                                                  {"hip-executor", [] { return gko::HipExecutor::create(0, gko::OmpExecutor::create(), true); }}};
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+class GinkgoRadialBasisFctSolver {
+public:
+  GinkgoRadialBasisFctSolver() = default;
+
+  GinkgoRadialBasisFctSolver(const MappingConfiguration::GinkgoParameter &ginkgoParameter);
```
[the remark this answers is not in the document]
Timo-Schrader: Got an error without the first one, but I'll check again.

src/mapping/GinkgoRadialBasisFctSolver.hpp
```diff
@@ -0,0 +1,533 @@
...
+  // Access to the evaluation matrix (output x input)
+  const std::shared_ptr<GinkgoMatrix> getEvaluationMatrix() const;
+
+  std::shared_ptr<gko::Executor> getReferenceExecutor() const;
+
+private:
+  mutable precice::logging::Logger _log{"mapping::GinkgoRadialBasisFctSolver"};
+
+  std::shared_ptr<gko::Executor>        _deviceExecutor;
+  static std::shared_ptr<gko::Executor> _hostExecutor;
```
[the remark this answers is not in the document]
Timo-Schrader: Yes, it's the same. It is necessary to manage data transfers as it would be by using plain CUDA. At least I never found another way, and looking at the Ginkgo examples, they almost do the same.

src/mapping/GinkgoRadialBasisFctSolver.hpp
```diff
@@ -0,0 +1,533 @@
...
+  static std::shared_ptr<gko::Executor> _ompExecutor;
+
+  // Stores the RBF interpolation matrix
+  std::shared_ptr<GinkgoMatrix> _rbfSystemMatrix;
+
+  /// Decomposition of the interpolation matrix
+  std::shared_ptr<GinkgoMatrix> _decMatrixC;
```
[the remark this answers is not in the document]
Timo-Schrader: Yes, should be removed of course.

src/mapping/GinkgoRadialBasisFctSolver.hpp
```diff
@@ -0,0 +1,533 @@
...
+
+  precice::utils::Event _assemblyEvent{"map.rbf.ginkgo.assembleMatrices", false, false};
+
+  std::shared_ptr<gko::log::Convergence<>> _logger;
+};
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+GinkgoRadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::GinkgoRadialBasisFctSolver(const MappingConfiguration::GinkgoParameter &ginkgoParameter)
+{
+  this->_deviceExecutor = ginkgoExecutorLookup.at(ginkgoParameter.executor)();
```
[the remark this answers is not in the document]
Timo-Schrader: I think `this` will never make any difference. I'd personally prefer to have it, but I can remove it of course.

src/mapping/GinkgoRadialBasisFctSolver.hpp
```diff
@@ -0,0 +1,533 @@
...
+  auto outputVertices = gko::share(GinkgoMatrix::create(this->_hostExecutor, gko::dim<2>{outputVerticesM, outputVerticesN}));
+  for (std::size_t i = 0; i < inputMeshSize; ++i) {
+    for (std::size_t j = 0; j < meshDim; ++j) {
+      if ("cuda-executor" == ginkgoParameter.executor) {
+        inputVertices->at(j, i) = inputMesh.vertices().at(i).rawCoords()[j];
+      } else {
+        inputVertices->at(i, j) = inputMesh.vertices().at(i).rawCoords()[j];
+      }
+    }
+    // Initial guess is required since memory chunk could lead to never converging system
```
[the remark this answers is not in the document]
Timo-Schrader: Yes

src/mapping/GinkgoRadialBasisFctSolver.hpp
```diff
@@ -0,0 +1,533 @@
...
+
+  this->_rbfCoefficients = gko::clone(this->_deviceExecutor, this->_rbfCoefficients);
+  this->_deviceExecutor->synchronize();
+  this->_copyEvent.pause();
+
+  this->_rbfSystemMatrix = gko::share(GinkgoMatrix::create(this->_deviceExecutor, gko::dim<2>{n, n}));
+  this->_matrixA         = gko::share(GinkgoMatrix::create(this->_deviceExecutor, gko::dim<2>{outputSize, n}));
+
+  if (polynomial == Polynomial::SEPARATE) {
+    const unsigned int polyParams = 4 - std::count(activeAxis.begin(), activeAxis.end(), false);
```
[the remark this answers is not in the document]
Timo-Schrader: You're right of course.

src/mapping/GinkgoRadialBasisFctSolver.hpp
```diff
@@ -0,0 +1,533 @@
...
+
+  // Wait for the kernels to finish
+  this->_deviceExecutor->synchronize();
+  this->_assemblyEvent.stop();
+
+  dInputVertices->clear();
+  dOutputVertices->clear();
+
+  auto iterationCriterion = gko::share(gko::stop::Iteration::build()
+                                           .with_max_iters(static_cast<std::size_t>(1e6))
```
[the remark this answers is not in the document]
Timo-Schrader: I also thought about that and we probably should just add another config tag for that. Likewise, I'd probably add a config tag for the GPU device ID s.t. in a multi-GPU setting, the user is able to select one that is available (instead on relying on device 0).

src/mapping/GinkgoRadialBasisFctSolver.hpp
```diff
@@ -0,0 +1,533 @@
...
+                                           .on(this->_deviceExecutor));
+
+  auto residualCriterion = gko::share(gko::stop::ResidualNormReduction<>::build()
+                                          .with_reduction_factor(ginkgoParameter.residualNorm)
+                                          .on(this->_deviceExecutor));
+
+  iterationCriterion->add_logger(this->_logger);
+  residualCriterion->add_logger(this->_logger);
+
+  if (this->_solverType == SolverType::MG) {
```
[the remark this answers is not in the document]
Timo-Schrader: Probably makes sense for now. We could also implement it as possible preconditioner for CG and GMRES and hope for Ginkgo to improve it.

src/mapping/GinkgoRadialBasisFctSolver.hpp
```diff
@@ -0,0 +1,533 @@
...
+
+  if (polynomial == Polynomial::SEPARATE) {
+    auto polynomialSolverFactory = cg::build()
+                                       .with_criteria(gko::stop::Iteration::build()
+                                                          .with_max_iters(static_cast<std::size_t>(1e6))
+                                                          .on(this->_deviceExecutor),
+                                                      gko::stop::ResidualNormReduction<>::build()
+                                                          .with_reduction_factor(1e-4)
+                                                          .on(this->_deviceExecutor))
+                                       .on(this->_deviceExecutor);
```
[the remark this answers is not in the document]
Timo-Schrader: You mean the instantiation of the solver? if so, of course, will move up. Same with matrices Q^T and Q^TQ

src/mapping/GinkgoRadialBasisFctSolver.hpp
```diff
@@ -0,0 +1,533 @@
...
+}
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+Eigen::VectorXd GinkgoRadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::solveConservative(const Eigen::VectorXd &rhsValues, Polynomial polynomial)
+{
+  return Eigen::VectorXd(1, 1);
+}
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+std::shared_ptr<gko::Executor> GinkgoRadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::_hostExecutor = gko::ReferenceExecutor::create();
```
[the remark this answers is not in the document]
Timo-Schrader: Will check.

src/mapping/GinkgoRadialBasisFctSolver.hpp
```diff
@@ -0,0 +1,533 @@
...
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+void GinkgoRadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::clear()
+{
+  if (nullptr == this->_rbfSystemMatrix) {
+    this->_rbfSystemMatrix = gko::share(GinkgoMatrix::create(this->_deviceExecutor, gko::dim<2>{0, 0}));
+    this->_matrixA         = gko::share(GinkgoMatrix::create(this->_deviceExecutor, gko::dim<2>{0, 0}));
+  } else {
+    this->_rbfSystemMatrix = gko::share(GinkgoMatrix::create(this->_deviceExecutor, gko::dim<2>{this->_rbfSystemMatrix->get_size()[0], this->_rbfSystemMatrix->get_size()[1]}));
+    this->_matrixA         = gko::share(GinkgoMatrix::create(this->_deviceExecutor, gko::dim<2>{this->_matrixA->get_size()[0], this->_matrixA->get_size()[1]}));
```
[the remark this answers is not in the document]
Timo-Schrader: Thanks for noticing, this was one of the very first functions available. We should replace it with calls to Ginkgo's `clear()` since it deallocates memory (I suspect memory leaks in the case here).

src/mapping/RadialBasisFctMapping.hpp
Timo-Schrader: Sounds good, but I need to check how to do correctly.

src/mapping/RadialBasisFctMapping.hpp
Timo-Schrader: First time I see these errors, but will do!

src/mapping/RadialBasisFctMapping.hpp
Timo-Schrader: It produced some error. Will later elaborate on that one.

src/mapping/RadialBasisFctMapping.hpp
Timo-Schrader: Sounds good. We could then also think of a common interface class.

src/mapping/config/MappingConfiguration.cpp
Timo-Schrader: Yes (as mentioned above). We should enable the user to provide the GPU device id (which always starts with 0). We can just hand this number over to the executor instantiation then.

src/mapping/config/MappingConfiguration.cpp
Timo-Schrader: Also thought about that. Currently, PETSc is the preferred variant.

src/mapping/ginkgo_kernels/GinkgoRBFKernels.cpp
```diff
@@ -0,0 +1,188 @@
...
+TO, THE IMPLIED WARRANTIES OF MERCHANTABILITY AND FITNESS FOR A
+PARTICULAR PURPOSE ARE DISCLAIMED. IN NO EVENT SHALL THE COPYRIGHT
+HOLDER OR CONTRIBUTORS BE LIABLE FOR ANY DIRECT, INDIRECT, INCIDENTAL,
+SPECIAL, EXEMPLARY, OR CONSEQUENTIAL DAMAGES (INCLUDING, BUT NOT
+LIMITED TO, PROCUREMENT OF SUBSTITUTE GOODS OR SERVICES; LOSS OF USE,
+DATA, OR PROFITS; OR BUSINESS INTERRUPTION) HOWEVER CAUSED AND ON ANY
+THEORY OF LIABILITY, WHETHER IN CONTRACT, STRICT LIABILITY, OR TORT
+(INCLUDING NEGLIGENCE OR OTHERWISE) ARISING IN ANY WAY OUT OF THE USE
+OF THIS SOFTWARE, EVEN IF ADVISED OF THE POSSIBILITY OF SUCH DAMAGE.
+******************************<GINKGO LICENSE>*******************************/
```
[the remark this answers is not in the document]
Timo-Schrader: Yes, initally took that from the Ginkgo repo, but of cours, the code is fully ours.

src/mapping/ginkgo_kernels/GinkgoRBFKernels.cpp
```diff
@@ -0,0 +1,188 @@
...
+
+#include "mapping/impl/BasisFunctions.hpp"
```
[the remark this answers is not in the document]
Timo-Schrader: This file is handed over to the unified kernel CMake macro, so it's not part of the preCICE build process itself.

src/mapping/ginkgo_kernels/GinkgoRBFKernels.cpp
```diff
@@ -0,0 +1,190 @@
...
+
+          // Since this block spans across two lines, we have to use global memory and cannot use prefetched memory
+          for (size_t k = 0; k < dataDimensionality; ++k) {
+            y    = supportPoints[k * inputRowLength + j] - targetPoints[k * outputRowLength + i];
+            dist = fma(y, y, dist);
+          }
+        } else {
+
+          // If this block is indeed only in one row, we can make thread 0 in each block responsible for prefetching values into shared memory
+          if (0 == threadIdx.x) {
```
Timo-Schrader, line 89: Will adjust that for data dimensionality < 3

src/mapping/ginkgo_kernels/GinkgoRBFKernels.cpp
Timo-Schrader: As long as the amount of vertices * dimensionality does not exceed the 32bit, we should be fine. Above, we deal with millions of threads, so this obviously overflows very fast when it comes to O(n^2) matrices.

src/mapping/ginkgo_kernels/GinkgoRBFKernels.cpp
Timo-Schrader: Will replace.

src/mapping/ginkgo_kernels/GinkgoRBFKernels.cpp
Timo-Schrader: What about having this function signature as `#define` macro which takes one input which is the name of the RBF class?

src/mapping/impl/BasisFunctions.hpp
Timo-Schrader: Ah, this is probably what the compiler wants to tell me. But why does this compile anyways?

src/mapping/GinkgoRadialBasisFctSolver.hpp
```diff
@@ -0,0 +1,533 @@
...
+  GMRES,
+  MG
+};
+
+enum PreconditionerType {
+  Jacobi,
+  Cholesky,
+  Ilu,
+  None
+};
```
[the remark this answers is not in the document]
Timo-Schrader: I would prefix them with "Ginkgo". I basically introduced them in order to not always use "magic strings" which could change later, e..g, `cg-solver`.

src/mapping/GinkgoRadialBasisFctSolver.hpp
```diff
@@ -0,0 +1,533 @@
...
+
+  const std::size_t inputMeshSize  = inputMesh.vertices().size();
+  const std::size_t outputMeshSize = outputMesh.vertices().size();
+  const std::size_t meshDim        = inputMesh.vertices().at(0).getDimensions();
+
+  this->_scalarOne         = gko::share(gko::initialize<GinkgoScalar>({1.0}, this->_deviceExecutor));
+  this->_scalarNegativeOne = gko::share(gko::initialize<GinkgoScalar>({1.0}, this->_deviceExecutor));
+
+  // Now we fill the RBF system matrix on the GPU (or any other selected device)
+  this->_rbfCoefficients = gko::share(GinkgoVector::create(this->_hostExecutor, gko::dim<2>{n, 1}));
```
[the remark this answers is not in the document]
Timo-Schrader: I don't think it initializes with a default value, this is why we have to set every initial entry in the solution vector to 0.

src/mapping/RadialBasisFctMapping.hpp
Timo-Schrader: So I finally checked again and my thoughts were that copying all the executor pointers etc. might be a bit unclean. I added a proper destructor deleting all pointers and calling `clear()` which frees the memory of all Ginkgo matrices now. Therefore, this "placement new" shouldn't make any problems anymore, or do you think we should still rather remove it?

src/mapping/config/MappingConfiguration.cpp
Timo-Schrader: I added a XML tag which enables the user to provide a GPU ID.

src/mapping/RadialBasisFctMapping.hpp
davidscn: I would rather formulate the question the other way around: you are trying to use a rather inconvenient language feature: why should we do that? I cannot imagine a whole lot of performance benefits using this style (proof me wrong). We call this loc per simulation exactly once and the large matrices are only allocated once we we hand over the system-size information here.
 
Referring to the  discussion above (https://github.com/precice/precice/pull/1581#discussion_r1117195631), making the solver the actual template parameter will kind of merge the line here with the other `if` branch, so we anyway need to find a solution here working for 'all' solvers.

Timo-Schrader: I also think making it a template will solve this. Let me try to figure out a clean way to do this. We can then discuss this in person.

davidscn:
Will merge this into a preCICE branch for now, such that we can work on a common platform.

PR #1586, opened 2023-02-27
Remove getThisTimeWindowRemainder(), use getNextTimestepMaxLength() instead 

BenjaminRodenberg:
## Main changes of this PR

Removes function `getThisTimeWindowRemainder()` and uses `getNextTimestepMaxLength()` instead, because both functions have the same functionality.

Simplification of code base.

## Author's checklist

* [x] ~~I added a changelog file with `make changelog` if there are user-observable changes since the last release.~~ not necessary
* [x] ~~I added a test to cover the proposed changes in our test suite.~~ pure refactoring, but added a few tests in this direction in [3e82443](https://github.com/precice/precice/commit/3e824439a4e4ac477b5f5aafe1976162f3105143)

## Reviewers' checklist


BenjaminRodenberg:
@uekerman Do you have an idea where we should additionally do testing here?

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -687,7 +676,7 @@ void BaseCouplingScheme::advanceTXTWriters()
 
 bool BaseCouplingScheme::reachedEndOfTimeWindow()
 {
-  return math::equals(getThisTimeWindowRemainder(), 0.0, _eps);
+  return math::equals(getNextTimestepMaxLength(), 0.0, _eps) || not hasTimeWindowSize();
```
BenjaminRodenberg, line 679: I think this is the most critical change with respect to functionality of `getNextTimestepMaxLength` vs. `getThisTimeWindowRemainder`.

src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -140,21 +140,13 @@ class BaseCouplingScheme : public CouplingScheme {
    */
   double getTimeWindowSize() const override final;
 
-  /**
-   * @brief Returns the remaining timestep length within the current time window.
-   *
-   * If no time window size is prescribed by the coupling scheme, always 0.0 is
```
BenjaminRodenberg, line 146: Returning 0.0 here I think is not a good idea (looking at the behavior of `method="participant-first"`

uekerman (approved):
Makes sense :+1:

Issue #1461, opened 2022-11-04
Use names instead of IDs in API


uekerman:
## Motivation

When explaining the API of preCICE, there is always this weird moment arguing what these mesh and data IDs are for.

```python
mesh_id = interface.get_mesh_id("Fluid-Mesh")
vertex_ids = interface.set_mesh_vertices(mesh_id, positions)
```

We could merge these two calls into:

```python
vertex_ids = interface.set_mesh_vertices("Fluid-Mesh", positions)
```

and similarly for all other calls. Tricky is the data access as we would now need both names, data and mesh (as the data ID depends on the mesh ID):

```python
write_block_scalar_data("Force", "Fluid-Mesh", vertex_ids, u)
```

## Technical level

In principle one could implement this, for example, directly in the Python bindings without changing preCICE itself. But we could also directly modify the C++ API. 

Opinions?

fsimonis:
We could encapsulate the internal mesh ids in a proxy object:

```py
vertexids = interface.mesh("Fluid-Mesh").set_vertices(positions)
interface.mesh("Fluid-Mesh").write_scalar_data("Force", vertexids, u)
```

If we pin vertex ids directly to a mesh, then we could even do:

```py
fluid_vertices = interface.mesh("Fluid-Mesh").set_vertices(positions)
fluid_vertices.write_scalar_data("Force", u)
```

It is worth noting, that moving away from the IDs will result in more complex C bindings.
The big question is what is more important to us: Consistency or comfort

Related to #717

BenjaminRodenberg:
> It is worth noting, that moving away from the IDs will result in more complex C bindings. The big question is what is more important to us: Consistency or comfort

I thought the same about the python bindings (and probably it applies generally): If we move away from low level types we probably have to think for every language about an individual solution that fits. This might improve usability, but also makes the bindings less consistent (think about the numpy arrays vs. raw double arrays in the python bindings, required some discussions and thinking work).

fsimonis:
Disucssion result:
* We like this a lot
* Does this impact performance if called in a hot loop (setMeshVertex per vertex of a mesh, or read write per vertex)

davidscn:
I benchmarkd  the write data functions we have in preCICE using Googlebench. In the table below, I tested 1e6 data sites using our bulk version (`*Block*`) and our singular versions within a loop. The `*String` version of each function has additionally the `getMeshID` and `geDataID` call in front of the write data function such that these calls appear in the hot loop for our singular data write versions:  

[...]

PR #1587, opened 2023-02-28
Remove getMeshIDs

fsimonis:
## Main changes of this PR

This PR removes the deprecated API method `getMeshIDs()`.

Unclear use-case and will be obsolete with #1461 

## Author's checklist


uekerman (approved):
Can't think of a use case.
I assume you checked the bindings as well.

PR #1593, opened 2023-03-06
Fix info log for linear-cell interpolation mapping

### LINKED ISSUE #1565: Confusing fall-back warnings in linear-cell interpolation
**Describe your setup**

preCICE Version: 1965ceb

**Describe the problem**
Using the `linear-cell-interpolation` mapping without providing any connectivity leads to the following warnings/logs

```bash
---[precice] WARNING:  2D Mesh "Solid-Mesh" does not contain triangles. Linear cell interpolation falls back to nearest projection mapping.
---[precice]  All vertices are inside cells, no fallback to Nearest-Projection required
```

This is very confusing and some people might think that the LCI mapping just works (without providing connectivity.

**Step To Reproduce**
1. Select a tutorial case without connectivity (I used the perpendicular-flap)
2. exchange one of the mappings to use linear-cell-interpolation

**Expected behaviour**
A clear warning saying what is actually wrong and what should be changed.

FYI: @boris-martin

davidscn:
## Main changes of this PR

Fixes the info log for linear-cell interpolation mapping.

See #1565, closes #1565 

## Author's checklist

<del> [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release </del>
<del> I added a test to cover the proposed changes in our test suite. </del>

## Reviewers' checklist

### DIFF
--- a/src/mapping/LinearCellInterpolationMapping.cpp
+++ b/src/mapping/LinearCellInterpolationMapping.cpp
@@ -48,22 +48,25 @@ void LinearCellInterpolationMapping::computeMapping()
     searchSpace = input();
   }
 
-  const auto &fVertices           = origins->vertices();
-  bool        missingConnectivity = false;
+  const auto &fVertices       = origins->vertices();
+  bool        hasConnectivity = true;
 
   if (getDimensions() == 2) {
     if (!fVertices.empty() && searchSpace->triangles().empty()) {
-      PRECICE_WARN("2D Mesh \"{}\" does not contain triangles. "
-                   "Linear cell interpolation falls back to nearest projection mapping.",
-                   searchSpace->getName());
-      missingConnectivity = true;
+      const bool hasEdges = searchSpace->hasEdges();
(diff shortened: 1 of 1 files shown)


src/mapping/LinearCellInterpolationMapping.cpp
```diff
@@ -90,16 +93,21 @@ void LinearCellInterpolationMapping::computeMapping()
     }
   }
 
-  if (!fallbackStatistics.empty() && !missingConnectivity) {
```
boris-martin, line 93: Would probably be cleaner with two levels of block: `if (!fallbackStatistics.empty())` then check for `hasConnectivity`

boris-martin (approved):

src/mapping/LinearCellInterpolationMapping.cpp
davidscn: I don't like nested of blocks, as they somehow make the control flow more complex. However, I adjusted it

PR #1594, opened 2023-03-06
Add exceptions to precice errors

### LINKED ISSUE #280: Exception Handling of preCICE errors

### LINKED ISSUE #1189: Let preCICE only exit on critical errors

I'm trying to use preCICE in an interactive fashion via the python bindings. Here, even non fatal but erroneous calls to the preCICE API lead to an critical error and exit. This makes working interactively difficult, because every error is fatal.

I would suggest to use `PRECICE_ERROR` for non-fatal errors and introduce `PRECICE_FATAL` for fatal errors. After `PRECICE_ERROR` the user session is still valid and the user can continue execution. `PRECICE_FATAL` should only be raised, if there is no possibility to recover from the erroneous state.

We could also use `PRECICE_WARNING` for the purpose I'm proposing. However, I think raising an error is good, if a user provides input that is clearly wrong and therefore discarded. A warning may be ignored and you can continue, if you have a good reason. Erroneous input should be discarded.

**Additional context**

* This is a breaking change, because it will change how preCICE programs behave.
* #280 might be related, because we could then differentiate better
* Using the python bindings in an interactive fashion brings, of course, another layer of complexity. This issue is still an issue of the main library, because I think the bindings cannot handle the error in a reasonable way.

### Example:

```
$ python3
Python 3.8.10 (default, Nov 26 2021, 20:14:08) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import precice
>>> interface = precice.Interface("precice-config.xml", "SolverOne", 0, 1)
(0) 12:44:57 [xml::XMLParser]:152 in readXmlFile: ERROR: XML parser was unable to open configuration file "SolverOne"
$
```

I think killing the session of the user just because the arguments were mixed up is too harsh. It's clearly an error, but it should not be a fatal error. This also leaves the user with no possibility for error handling.

An example for a nicer solution:
```
$ python3
Python 3.8.10 (default, Nov 26 2021, 20:14:08) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> import os
>>> os.listdir('non-existing')
  File "<stdin>", line 1, in <module>
FileNotFoundError: [Errno 2] No such file or directory: 'non-existing'
>>> 
```

fsimonis:
## Main changes of this PR

This PR 
* adds an exception hierarchy.
* extends `PRECICE_ERROR(precice::APIError, "Oh no!")` to take an exception type first.
* migrates all error calls to exceptions
* catches exceptions and aborts in the C and Fortran bindings

This allows us to test failing cases.

Closes #280
Closes #1189

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/exceptions.hpp
+++ b/src/precice/exceptions.hpp
@@ -10,4 +10,46 @@ public:
       : std::runtime_error(what_arg){};
 };
 
+class ActionError : public precice::Error {
+public:
+  ActionError(const std::string &what_arg)
+      : precice::Error(what_arg){};
+};
+
+class AccelerationError : public precice::Error {
+public:
+  AccelerationError(const std::string &what_arg)
+      : precice::Error(what_arg){};
+};
+
+class CouplingSchemeError : public precice::Error {
+public:
+  CouplingSchemeError(const std::string &what_arg)
+      : precice::Error(what_arg){};
+};
+
+class ExportError : public precice::Error {
+public:
+  ExportError(const std::string &what_arg)
+      : precice::Error(what_arg){};
+};
+
+class MappingError : public precice::Error {
+public:
+  MappingError(const std::string &what_arg)
+      : precice::Error(what_arg){};
+};
+
+class InterfaceError : public precice::Error {
+public:
+  InterfaceError(const std::string &what_arg)
+      : precice::Error(what_arg){};
+};
+
+class ConfigurationError : public precice::Error {
+public:
+  ConfigurationError(const std::string &what_arg)
+      : precice::Error(what_arg){};
+};
+
 } // namespace precice
--- a/src/precice/exceptions.hpp
+++ b/src/precice/exceptions.hpp
@@ -34,27 +34,15 @@ public:
       : precice::Error(what_arg){};
 };
 
-class IOError : public precice::Error {
+class ExportError : public precice::Error {
 public:
-  IOError(const std::string &what_arg)
+  ExportError(const std::string &what_arg)
       : precice::Error(what_arg){};
 };
 
 class MappingError : public precice::Error {
 public:
   MappingError(const std::string &what_arg)
-      : precice::Error(what_arg){};
-};
-
-class MeshError : public precice::Error {
-public:
-  MeshError(const std::string &what_arg)
-      : precice::Error(what_arg){};
-};
-
-class M2NError : public precice::Error {
-public:
-  M2NError(const std::string &what_arg)
       : precice::Error(what_arg){};
 };
 
--- a/src/precice/exceptions.hpp
+++ b/src/precice/exceptions.hpp
@@ -28,9 +28,15 @@ public:
       : precice::Error(what_arg){};
 };
 
-class ExportError : public precice::Error {
+class CommunicationError : public precice::Error {
 public:
-  ExportError(const std::string &what_arg)
+  CommunicationError(const std::string &what_arg)
+      : precice::Error(what_arg){};
+};
+
+class IOError : public precice::Error {
+public:
+  IOError(const std::string &what_arg)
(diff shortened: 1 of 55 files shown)


davidscn (commented):
Looks good and will make out live easier in the future.

Only a few comments:

- although it is somewhat of an open question on how we treat the exceptions in C and Fortran, we duplicate quite a lot of code due to the usage of `try{} catch (...) {}`
- our test coverage will now complain if we don't test the throw's in our CI
- we need to export the `exceptions` header to the (user) installation include target

I would propose to keep the number of exceptions in preCICE rather small (maybe the four most significant) and extend later on demand. We could summarize smaller groups in `precice::InternalError` or something similar. Let's discuss in person.

src/precice/exceptions.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+#include <stdexcept>
+
+namespace precice {
+
+class Error : public std::runtime_error {
+public:
+  Error(const std::string &what_arg)
+      : std::runtime_error(what_arg){};
+};
```
BenjaminRodenberg, line 11: I agree with the suggestion of @davidscn to keep the number of errors small for the beginning and find out what we need later. The main motivation here is to simplify using the correct error in the correct situation. Two concrete ideas:

* Only use a single exception for now: `Error`. We can extend this later? Then there is no risk of wrong or inconsistent usage of error classes, because everything is simple.
* The errors below are basically one error class per package. From my perspective this means that there is implicitly the rule: "Use ActionError, if you want to throw an error from `precice::action`". This rule is simple enough to not mess up things from the beginning, but we should also put this into the documentation of each error class.

src/precice/exceptions.hpp
```diff
@@ -0,0 +1,79 @@
...
+
+class ActionError : public precice::Error {
+public:
+  ActionError(const std::string &what_arg)
+      : precice::Error(what_arg){};
+};
```
BenjaminRodenberg, line 17: Related to my comment above. I think we should put this error into the namespace `precice::action`. This makes it clear that the error should only be thrown from this package.

fsimonis:
We decided to:
* drop MeshError
* drop M2NError (unused)
* rename IOError to ExportError
* Use ExportError for the WatchPoint and WatchIntegral. Move them from `precice/impl` -> `io`
* eventually rename the `io` package to export

fsimonis:
I finally got around to implementing the simpler version proposed by @BenjaminRodenberg in #2065.

Closing this as it is outdated and doesn't add any value.

PR #1606, opened 2023-03-21
Print all reasonable tag hints instead of only the first

davidscn:
## Main changes of this PR

Prints all tag hints instead of only the first.

Which turns something like this:

```bash
ERROR: The configuration contains an unknown tag <basis-function:compact-polynomial-c3>. Did you mean <basis-function:compact-polynomial-c0>?
```
 into 

```bash
ERROR: The configuration contains an unknown tag <basis-function:compact-polynomial-c3>. Did you mean <basis-function:compact-polynomial-c0>,<basis-function:compact-polynomial-c2>,<basis-function:compact-polynomial-c4>,<basis-function:compact-polynomial-c6>?
```

I don't like the first version, because it looks as if we only have the `c0` function, but we have more (better) ones.

## Reviewers' checklist

### DIFF
--- a/src/xml/ConfigParser.cpp
+++ b/src/xml/ConfigParser.cpp
@@ -221,7 +221,10 @@ void ConfigParser::connectTags(const ConfigurationContext &context, std::vector<
 
       auto matches = utils::computeMatches(expectedName, names);
       if (matches.front().distance < 3) {
-        PRECICE_ERROR("The configuration contains an unknown tag <{}>. Did you mean <{}>?", expectedName, matches.front().name);
+        matches.erase(std::remove_if(matches.begin(), matches.end(), [](auto &m) { return m.distance > 2; }), matches.end());
+        std::vector<std::string> stringMatches;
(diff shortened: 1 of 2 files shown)


fsimonis:
The actual bug here is that in case of an obvious typo (edit distance < 3), the current method only suggests a single option, even if there are multiple options under the threshold.

Question: Do we want to suggest all options under the threshold, or only suggest the options sharing the lowest edit distance?

Example let's consider that we also have a `rbf-compact-polynomial-c99`:
* all under threshold: `rbf-compact-polynomial-c0 rbf-compact-polynomial-c2 rbf-compact-polynomial-c4 rbf-compact-polynomial-c6 rbf-compact-polynomial-c99`
* all with lowest threshold (of 1): `rbf-compact-polynomial-c0 rbf-compact-polynomial-c2 rbf-compact-polynomial-c4 rbf-compact-polynomial-c6`

fsimonis:
Thinking about this, it may actually be a better idea to be slightly verbose instead of being too brief.

fsimonis (approved):

Issue #1009, opened 2021-04-28
Cleanup Action Interface


BenjaminRodenberg:
The actions are a rarely used feature, which needs a lot of maintainment. We want to use the 3.0.0 release for removing unneeded actions. Let's use this issue for collecting comments related to the action interface.

BenjaminRodenberg:
[This code](https://github.com/precice/precice/pull/935/files/0ee5d5af9f5a1393c49812dd6869808a45ce6aeb#diff-77d4c4cc6ff34a09a88b532f34bf52b866f14df95ab2d29643121a460877410cR28-R30) needs an update, if #935 gets merged.

fsimonis:
This also includes the deprecated vertex normals #1000

fsimonis:
Related to #1438

uekerman:
We should also remove the deprecated ones:

```
REGULAR_PRIOR
REGULAR_POST
ON_EXCHANGE_PRIOR
ON_EXCHANGE_POST
```

They still exist on config level.

PR #1611, opened 2023-04-02
Remove occurences of deprecated action timings

BenjaminRodenberg:
## Main changes of this PR

* Remove deprecated action timings REGULAR_PRIOR, REGULAR_POST, ON_EXCHANGE_PRIOR, ON_EXCHANGE_POST from tests.
* Remove corresponding tests using these timings.

Comment in #1009 

## Author's checklist

## Reviewers' checklist


BenjaminRodenberg:
As suggested in #1009 I cleaned up the deprecated action timings. I did this in a very quick and rough way and also deleted all the tests that rely on these action timings. If someone could instead port these tests that we could still use them I guess this would be much better, because at the current state not too many tests are left (actually there is only a single unit test left).

uekerman:
I am already working towards this; might redo this here in another PR.

uekerman:
Replaced by #1614

Issue #1438, opened 2022-09-20
Which data actions are really needed?


BenjaminRodenberg:
While working on waveform relaxation I had a lot of trouble with providing the right data at the right time to the data actions. We might be able to simplify this part of preCICE by removing some features here. The following function is called at several places in `SolverInterfaceImpl`:

https://github.com/precice/precice/blob/b81e53f5bd5d63b18b45692756d894211050f381/src/precice/impl/SolverInterfaceImpl.cpp#L2028-L2041

There are the following timings that are used with this function:

* `action::Action::READ_MAPPING_PRIOR`
* `action::Action::READ_MAPPING_POST`
* `action::Action::WRITE_MAPPING_PRIOR`
* `action::Action::WRITE_MAPPING_POST`
* `action::Action::ON_TIME_WINDOW_COMPLETE_POST`

Ideally, I would like to remove all data actions that are used with these timings and only keep `action::Action::ON_TIME_WINDOW_COMPLETE_POST`. We could potentially also introduce a new timing at the end of the window, if the iteration has not converged. The purpose of this is to move all actions after the waveform relaxation and mapping is done to have less interference of actions, mapping and waveform relaxation.

BenjaminRodenberg:
@uekerman can we close this issue? I think #1614 pretty much resolves this.

uekerman:
yes

PR #1614, opened 2023-04-05
Cleanup actions

### LINKED ISSUE #1009: Cleanup Action Interface

uekerman:
## Main changes of this PR

- Removes deprecated timings `REGULAR_PRIOR, REGULAR_POST, ON_EXCHANGE_PRIOR, ON_EXCHANGE_POST`
- Further removes timings `READ_MAPPING_PRIOR, WRITE_MAPPING_PRIOR, ON_TIME_WINDOW_COMPLETE_POST` as I could not identify current use cases for these.
  - This means, only `READ_MAPPING_POST` and `WRITE_MAPPING_POST` remain. 
  - If the mesh coordinates are not involved in an action, there is no difference between "prior" and "post".
  - For the summation action, all data fields need to be on the same mesh. This is typically the case *after* the mapping. Before the mapping, one could also sum in the adapter if data fields are on the same mesh.
  - `ON_TIME_WINDOW_COMPLETE_POST`: I could not think of a realistic action one would only want to perform after convergence. 
- Removes the unused `ComputeCurvatureAction`.
- Removes `vertexCallback` and `postAction` from the callback interface of `PythonAction`.
  - If there is no dependence on the mesh coordinates, there can again be no difference between "prior" and "post".
  - If `vertexCallback` is gone, also `postAction` makes no sense any more. 

Closes #1009, #1438 
Replaces #1611

- Mainly removing unused, but not necessarily unusable features to keep things manageable. If ever needed, some of these features should not be too complicated to add again. I looked at all currently used actions I am aware of: #840, [forum](https://precice.discourse.group/t/multiple-fluid-domains-interacting-with-a-shell-structure-precice-configuration-on-the-openfoam-side/1250), [thesis Gibert](https://theses.hal.science/tel-03859464/document)  
- `ScaleByAreaAction` does only work 100% in serial. In parallel, it suffers from the same "connectivity over ranks" issue as scaled-consistent mappings, see also #986. We should mention this in the user documentation.

## Remarks on waveform handling

Currently only "prior" actions were supported. "post" ones do not work as data is already moved to the waveform within the mapping. Some tests now fail as I adjusted these to work with "post" actions.
In general, I recommend to perform actions and mappings en-block for all timesteps, meaning to treat both "read" and "write" the same way. 

https://github.com/precice/precice/blob/135167ac32a5131bfed62d79cb75eafdf9d98bd4/src/precice/impl/SolverInterfaceImpl.cpp#L439-L443

https://github.com/precice/precice/blob/135167ac32a5131bfed62d79cb75eafdf9d98bd4/src/precice/impl/SolverInterfaceImpl.cpp#L1844-L1855

## Author's checklist

* ~[ ] I added a test to cover the proposed changes in our test suite.~
* [ ] Adjust [user documentation](https://precice.org/configuration-action.html), but only merge once v3 is released.

## Reviewers' checklist

* [x] @BenjaminRodenberg Fix waveform handling (here or somewhere else??) such that tests work


BenjaminRodenberg:
The tests should work now. [5912ed2](https://github.com/precice/precice/pull/1614/commits/5912ed235739090b37c6bd703ec1f19c9a5407bb) is mainly a work-around until we have a symmetric treatment of the read and write data context.

BenjaminRodenberg (approved):
From my perspective this simplifies the actions a lot. Not sure if this is really the last and final step here, but definitely going into the right direction. I also merged this PR into "my" PR https://github.com/precice/precice/pull/1504 and it integrates well. But it also shows the problems with the current non-symmetric implementation of the read/write data context, actions and mapping more clearly (which is good).

PR #1616, opened 2023-04-05
Map rbf alias tag onto PUM mapping

### LINKED ISSUE #1608: Partition of unity follow-up tasks
Now that the basic version of the PUM mapping was added in https://github.com/precice/precice/pull/1483 there are a couple of follow-up tasks to keep track of:

- [ ] Document the feature in the mapping user-docs (including the exports in debug https://github.com/precice/precice/pull/1483#discussion_r1138492960)
- [ ] add the PUM method to the auto-configuration of our mapping configuration. This will probably become a bit more intricate than initially thought, as we don't have access to the global number of unknowns in the configuration (and instantiation) phase.

davidscn:
## Main changes of this PR

Let's the `<mapping:rbf ..>` tag resolve to `<mapping:rbf-pum-direct >`.

See #1608. I ran our perpendicular-flap tutorial and as pointed out in #1608, there are cases, where the global rbf would be faster. However, the overall mapping runtime for these examples was of $\mathcal{O}(1\cdot 10^{-2})$ seconds and negligible compared to the global runtime (<1%).
 
I also cleaned up a few logging information and added the mapping name in the info log, as users often times post the log without the config. Advantage for the users: they can see what happens due to the autoconfig. 

Closes #1608 .

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mapping/PartitionOfUnityMapping.hpp
+++ b/src/mapping/PartitionOfUnityMapping.hpp
@@ -401,5 +401,11 @@ void PartitionOfUnityMapping<RADIAL_BASIS_FUNCTION_T>::clear()
   _clusterRadius            = 0;
   this->_hasComputedMapping = false;
 }
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+std::string PartitionOfUnityMapping<RADIAL_BASIS_FUNCTION_T>::getName() const
+{
+  return "partition of unity rbf";
+}
 } // namespace mapping
 } // namespace precice
--- a/src/mapping/PartitionOfUnityMapping.hpp
+++ b/src/mapping/PartitionOfUnityMapping.hpp
@@ -405,7 +405,7 @@ void PartitionOfUnityMapping<RADIAL_BASIS_FUNCTION_T>::clear()
 template <typename RADIAL_BASIS_FUNCTION_T>
 std::string PartitionOfUnityMapping<RADIAL_BASIS_FUNCTION_T>::getName() const
 {
-  return "partition of unity rbf";
+  return "partition-of-unity RBF";
 }
 } // namespace mapping
 } // namespace precice
--- a/src/mapping/PartitionOfUnityMapping.hpp
+++ b/src/mapping/PartitionOfUnityMapping.hpp
@@ -195,7 +195,7 @@ void PartitionOfUnityMapping<RADIAL_BASIS_FUNCTION_T>::computeMapping()
     }
   }
   // Log the average number of resulting clusters
-  PRECICE_INFO("Partition of unity data mapping between mesh \"{}\" and mesh \"{}\": mesh \"{}\" on rank {} was decomposed into {} clusters.", this->input()->getName(), this->output()->getName(), inMesh->getName(), utils::IntraComm::getRank(), _clusters.size());
+  PRECICE_DEBUG("Partition of unity data mapping between mesh \"{}\" and mesh \"{}\": mesh \"{}\" on rank {} was decomposed into {} clusters.", this->input()->getName(), this->output()->getName(), inMesh->getName(), utils::IntraComm::getRank(), _clusters.size());
 
   if (_clusters.size() > 0) {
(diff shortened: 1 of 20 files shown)


uekerman:
> I ran our perpendicular-flap tutorial and as pointed out in https://github.com/precice/precice/issues/1608, there are cases, where the global rbf would be faster.

I couldn't find such information in #1608. Which cases would run faster?

I think it is OK to always use `mapping:rbf-pum-direct` for the moment. If then the mesh is small, we only have one cluster and basically implicitly the same functionality as global-direct RBF. 

Later, we could still tweak more.

davidscn:
> > I ran our perpendicular-flap tutorial and as pointed out in #1608, there are cases, where the global rbf would be faster.
> 
> I couldn't find such information in #1608. Which cases would run faster?

Cases with (heuristic value) 500 vertices, where PUM would maybe create 7 or 8 clusters (in surface coupling). As noted above, the overall runtime of these cases is, however, very small.

src/mapping/impl/CreateClustering.hpp
```diff
@@ -341,7 +341,7 @@ inline std::tuple<double, Vertices> createClustering(mesh::PtrMesh inMesh, mesh:
   // Step 2: Now we pick random samples from the input mesh and ask the index tree for the k-nearest neighbors
   // in order to estimate the point density and determine a proper cluster radius (see also the function documentation)
   double clusterRadius = estimateClusterRadius(verticesPerCluster, inMesh, localBB);
-  PRECICE_INFO("VertexCluster Radius: {}", clusterRadius);
+  PRECICE_DEBUG("VertexCluster Radius: {}", clusterRadius);
```
uekerman, line 344:
```suggestion
  PRECICE_DEBUG("Vertex cluster radius: {}", clusterRadius);
```

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -398,5 +401,11 @@ void PartitionOfUnityMapping<RADIAL_BASIS_FUNCTION_T>::clear()
   _clusterRadius            = 0;
   this->_hasComputedMapping = false;
 }
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+std::string PartitionOfUnityMapping<RADIAL_BASIS_FUNCTION_T>::getName() const
+{
+  return "partition of unity rbf";
```
uekerman, line 408:
```suggestion
  return "partition-of-unity RBF";
```

src/mapping/PetRadialBasisFctMapping.hpp
```diff
@@ -584,6 +587,12 @@ void PetRadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::clear()
   this->_hasComputedMapping = false;
 }
 
+template <typename RADIAL_BASIS_FUNCTION_T>
+std::string PetRadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::getName() const
+{
+  return "rbf global iterative";
```
uekerman, line 593:
```suggestion
  return "global-iterative RBF";
```

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -160,6 +163,12 @@ void RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::clear()
   this->_hasComputedMapping = false;
 }
 
+template <typename RADIAL_BASIS_FUNCTION_T>
+std::string RadialBasisFctMapping<RADIAL_BASIS_FUNCTION_T>::getName() const
+{
+  return "rbf global direct";
```
uekerman, line 169:
```suggestion
  return "global-direct RBF";
```

uekerman (approved):
Functionality looks good to me, only some concerns about the naming.
I tried to make things a bit more consistent. Before there was kind of a mixture between "mapping name" and "name of tag". Not sure which one is better actually.

PR #1618, opened 2023-04-12
Update eigen version requirement

davidscn:
## Main changes of this PR
Updates the eigen version requirement to match our documentation. I tried to test that it actually fails with older versions. That's, unfortunately, not too easy as older versions do not rely on CMake.


MakisH (approved):
This actually seems to already be our documented requirement, CMake just did not know about it: https://precice.org/installation-source-dependencies.html#eigen

fsimonis (approved):
Thanks!

PR #1620, opened 2023-04-17
Re-arrange delimiters in log files

davidscn:
## Main changes of this PR

Removes the trailing whitespace in the watchpoint, convergence and iterator log files.

## Additional information

working with trailing whitespaces in hit is tedious.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/io/TXTTableWriter.cpp
+++ b/src/io/TXTTableWriter.cpp
@@ -35,8 +35,8 @@ void TXTTableWriter::addData(
   if ((type == INT) || (type == DOUBLE)) {
     _outputStream << delimiter << name;
   } else {
+    _outputStream << delimiter << name << 0;
     if (type == VECTOR2D) {
-      _outputStream << delimiter << name << 0;
       for (int i = 1; i < 2; i++) {
         _outputStream << "  " << name << i;
       }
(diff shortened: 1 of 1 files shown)


src/io/TXTTableWriter.cpp
```diff
@@ -30,16 +30,21 @@ void TXTTableWriter::addData(
   data.name = name;
   data.type = type;
   _data.push_back(data);
+
+  std::string delimiter = _data.empty() ? "" : "  ";
```
fsimonis, line 34: `delimiter` is a tricky choice as the space separating vectorial components is also a delimiter.

I would handle this edge case here and concentrate on just the data section below.
```
  if (!_data.empty()) {
    _outputStream << ' ';
  }
```

Also, this has to move above `_data.push_back()`.

davidscn: > delimiter is a tricky choice as the space separating vectorial components is also a delimiter.

I don't see the trickyness here. What exactly is the problem with that?

> Also, this has to move above _data.push_back().

Yes, that's what happens in last minute changes :see_no_evil:

fsimonis: The point I am trying to make, is that we can simplify and clarify this.
fmtlibs makes a lot of fiddly string handling very clean.

fsimonis:
Merging this now. We can always clean this up further.

PR #1621, opened 2023-04-20
Add check for C++17 std::transform_reduce

### LINKED ISSUE #1617: Insufficient C++17 compiler support check in CMake

Using GCC 9.2.0 and our latest develop, I am able to configure precice using `CMake`. However, at compile time, the compiler generates the following error message:

```bash
~/precice/src/com/SerializedPartitioning.cpp: In static member function ‘static precice::com::serialize::SerializedConnectionMap precice::com::serialize::SerializedConnectionMap::serialize(const ConnectionMap&)’:
~/precice/src/com/SerializedPartitioning.cpp:26:8: error: no matching function for call to ‘transform_reduce(std::map<int, std::vector<int> >::const_iterator, std::map<int, std::vector<int> >::const_iterator, int, std::plus<long unsigned int>, precice::com::serialize::SerializedConnectionMap::serialize(const ConnectionMap&)::<lambda(const auto:1&)>)’
   26 |       });
      |        ^
```

The problem here is that this GCC version does not fully support the required [`std::transform_reduce`](https://en.cppreference.com/w/cpp/algorithm/transform_reduce) C++17 feature.

We should catch this already at compile time and inform the user about this. Looking at our CMake C++ checks, our standard support check [only checks for C++11 compliance](https://github.com/precice/precice/blob/develop/cmake/CheckCXX11Library.cmake). It would be easy to either extend the check to C++17 or add an additional C++17 check.

For me, it was sufficient to modify the C++11 check for 

```cpp
#include <execution>
```

Do nothing and let the user figure out.

fsimonis:
## Main changes of this PR

This PR adds a check for C++17 std::transform_reduce during the configuration of the CMake project.

Closes #1617

## Author's checklist


cmake/CheckSTL.cmake
```diff
@@ -0,0 +1,72 @@
...
+
+  set(CMAKE_CXX_STANDARD 17)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
+  check_cxx_source_compiles("${CheckSourceTransformReduce}" CXX17_TRANSFORM_REDUCE)
+  if(NOT "${CXX17_TRANSFORM_REDUCE}")
+    message(FATAL_ERROR "
+    The standard library used by your compiler ${CMAKE_CXX_COMPILER_ID} ${CMAKE_CXX_COMPILER_VERSION} is not C++17 compliant.
+    It does not support std::transform_reduce (P0024R2)!
+    Please upgrade your compiler at least to:
+    * GCC 9.2
```
davidscn, line 50: I used this GCC version when facing the issue in the first place. I guess this should be 9.3?

cmake/CheckSTL.cmake
```diff
@@ -0,0 +1,72 @@
...
+  int main() { std::time_t t = std::time(nullptr); std::tm tm = *std::localtime(&t); std::cout << std::put_time(&tm, \"%c %Z\"); }"
+  )
+
+  set(CMAKE_CXX_STANDARD 11)
+  set(CMAKE_CXX_STANDARD_REQUIRED ON)
+  check_cxx_source_compiles("${CheckSourceN2071N2072}" CXX11_N2071_N2072)
+  if(NOT "${CXX11_N2071_N2072}")
+    message(FATAL_ERROR "
+    The standard library used by your compiler ${CMAKE_CXX_COMPILER_ID} ${CMAKE_CXX_COMPILER_VERSION} is not C++11 compliant.
+    It does not support time manipulation (N2071 and N2072)!
```
davidscn, line 18:
```suggestion
    It does not support time manipulation (N2071 and N2072).
```

cmake/CheckSTL.cmake
```diff
@@ -0,0 +1,72 @@
...
+    Please upgrade your compiler at least to:
+    * GCC 5
```
davidscn, line 20: I guess this is not really of high practical relevance, as almost all compilers will succeed here, but it is somewhat weird to ask for GCC v5, as it won't be sufficient to compile preCICE.

cmake/CheckSTL.cmake
fsimonis:
```suggestion
    * GCC 9.3
```

cmake/CheckSTL.cmake
fsimonis: True. We should provide a global recommended baseline for compilers.

cmake/CheckSTL.cmake
```diff
@@ -0,0 +1,71 @@
+include(CheckCXXSourceCompiles)
+
+# These are the compiler/STL versions required to compile preCICE
+set(_precice_recommended_baseline "Please upgrade your compiler at least to:
+    * GCC 9.2
+    * LLVM 5.0
+    * MSVC 19.14
+    * Intel 15 in conjunction with GCC 9.2
+    * PGI 2015 in conjunction with GCC 9.2
```
davidscn, line 9:
```suggestion
    * GCC 9.3
    * LLVM 5.0
    * MSVC 19.14
    * Intel 15 in conjunction with GCC 9.3
    * PGI 2015 in conjunction with GCC 9.3
```

Issue #1559, opened 2023-02-06
Simplify the Data API


fsimonis:
**Please describe the problem you are trying to solve.**

The API to access data is cumbersome as it implements all possibilities of the cross-product  
`{read, write, writeGradient} x {Single, Block} x {Scalar, Vector}`

This leads to 12 API calls that are cumbersome to maintain as the input checks need to preserve context while their implementation logic is extremely similar.
It also makes using the API cumbersome as these functions slightly differ in syntax and return types.

**Describe the solution you propose.**

We use the block variant as a replacement for singular versions.
We omit the data type (Vector or Scalar) in the API name and deduce which of the two to use inside the function call. In the code, the scalar version can often be considered a vector of 1 element.
Currently, we check if the user calls the correct variant for the configured datatype. We can reuse this logic to select the correct process internally.

This would replace 12 API functions with:
```cpp
void SolverInterface::readData(         DataID did, int size, const int* indices, double* values);
void SolverInterface::writeData(        DataID did, int size, const int* indices, const double* values);
void SolverInterface::writeGradientData(DataID did, int size, const int* indices, const double* gradientValues);
```

Further upside is that this would simplify the documentation, as the code now doesn't explicitly refer to vectorial data any longer.

**Downside** for read-data is that the user can no-longer directly use/assign single values, but always need to use references using the more verbose variant.
```cpp
double d;
interface.readData(did, 1, &vertexID, &d); 

// previously
double d = interface.readScalarData(did, vertexID); 
```

An alternative could be a singular overload for the readScalarData version, which may result in confusion:
```cpp
void SolverInterface::readData(DataID did, int size, const int* indices, double* values);
double SolverInterface::readData(DataID did, int index);  // assumes scalar
```

**Describe alternatives you've considered**

Keep the Scalar/Vector differentiation, but merge Single/Block variants.
This at least reduced forwarding or other duplications.

IshaanDesai:
As far as the Python bindings go, this simplification should not cause any issues. Actually it is more *pythonic* to deduce the data type (vector or scalar) and to not have the single and block variants but just one general API call. We already do this to some degree [in the fenics-adapter](https://github.com/precice/fenics-adapter/blob/9aa3e22496117725ba993ac677a1239dc6f156f8/fenicsprecice/fenicsprecice.py#L193-L230), wherein we deduce all the necessary information about the data within the `read_data` function.

uekerman:
[...]

PR #1622, opened 2023-04-21
Make `relativeReadTime` mandatory in API

### LINKED ISSUE #1603: Make `dt` argument in read data functions mandatory
The technical discussion in #1559 actually brought us to a bigger usability discussion about making the `dt` argument in all read data functions mandatory. This is a big step for our users, because they will always have to provide a `dt`, but this also means they have to realize that read data is always connected to some kind of timing and that there is a waveform behind the scenes.

Make `dt` mandatory. The default behavior of v2 could still be realized by users easily, if they provide `dt = precice_dt` (where `precice_dt` is obtained from `advance`)

In version 2 we provide a default for `dt`. This makes `dt` optional. However, there is a technical problem with the C/Fortran bindings (see #1559)

BenjaminRodenberg:
## Main changes of this PR

Make `relativeReadTime` mandatory in read API and remove unnecessary functions. Also update tests correspondingly. Closes #1603 

https://github.com/precice/precice/issues/1559 depends on this PR.

Note that this makes the previously experimental waveform API the new default. Using waveform order bigger 0 should, however, still be considered experimental.

## Author's checklist

* [x] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html). See https://github.com/precice/precice.github.io/pull/256

## Reviewers' checklist


src/precice/SolverInterface.hpp
fsimonis, line 1: We should directly refer to `getMaxTimeStepSize` in doxygen:

```
 * @see SolverInterface::getMaxTimeStepSize()
```

fsimonis (approved):
Be aware of/ready for the merge conflicts wrt #1623

PR #1624, opened 2023-04-24
Change `IDs` in Fortran bindings to `ids`

davidscn:

...as that's what we used before (when we had our other ID functions) and that's what our users are used to. See https://precice.discourse.group/t/an-error-occurs-when-i-want-to-access-the-mesh-information-directly/1386/3

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/1624.md
+++ b/docs/changelog/1624.md
@@ -1,0 +1,1 @@
+- Fixed the Fortran API to use lowercase `id` in functions setting mesh vertices.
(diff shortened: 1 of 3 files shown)


MakisH (approved):
Looks clear enough. 

Will need updates in https://github.com/precice/fortran-module/blob/develop/precice.f03 (binded functions).

Fun fact: Fortran is case-insensitive. But even if this works by accident, we should keep it consistent.

fsimonis:
I created a stub issue so that we don't forget this https://github.com/precice/fortran-module/issues/13

Issue #1605, opened 2023-03-20
preCICE C++ API requires C++ 17 in user codes


davidscn:
**Describe the problem**

Due to our changes in #1471, we use now a `std::string_view` in order to pass string arguments to preCICE. However, now that we use `C++17` features explicitly in the API, we essentially force all adapter compiler to use and be compatible with a `-std=c++17`.

**Expected behaviour**
Either document this change or provide a backwards compatible version of `string_view`. Expecting C++17 support from the compiler is probably not overly problematic, but expecting all adapter/user-code to be compiled with C++17 might be too restrictive.

**Additional context**
I discovered this during the OpenFOAM-adapter porting: OpenFOAM (precompiled binary) was built with C++14. Calling the OpenFOAM own `wmake` compiler wrapper will automatically invoke `-std=c++14`, which leads to errors in the `SolverInterface`

```bash
precice/SolverInterface.hpp:55:23: error: expected ‘)’ before ‘participantName’
   55 |       std::string_view participantName,
      |                       ^~~~~~~~~~~~~~~~
      |                       )
```

IshaanDesai:
Not only adapters, but also some of the [bindings](https://github.com/precice/python-bindings/blob/develop/setup.py#L67).

fsimonis:
A danger of `using string_view = precice::span` for compatibility reasons is, that we will never make the shift to STL `std::string_view` and `std::span`.

With the release of Ubuntu 24.04 LTS, our baseline shifts to Ubuntu 22.04 LTS, which already ships with a C++20 `span` implementation.
So, it is realistic that we could require C++20 internally and by user-code (due to `std::span`) in April 2024.

davidscn:
> So, it is realistic that we could require C++20 internally and by user-code (due to std::span) in April 2024.

The main question here remains: Do we want to enforce constraints on the user build-system or not? Although _our_ baseline supports all these STL features, it doesn't imply that our user codes go with the same strategy. Example: OpenFOAM build its packages as of now using c++14, which is -speaking in our dependency structure- deprecated.

MakisH:
>     * In the worst case, we can simply backport the API from `std::string_view` to `const std::string&`, which will come with a performance / convenience impact. An alternative to `std::string_view` is to use the span `using string_view = precice::span<const char>` after [Using spans instead of raw pointers in the API #1577](https://github.com/precice/precice/issues/1577) (in case we conclude on using this).

[...]

PR #1625, opened 2023-04-24
Add cstring support to span

fsimonis:
## Main changes of this PR

This PR adds tests for the span type and a conditional cstring constructor.
This makes `span<const char>` usable as a replacement for `std::string_view`.

Preparation for tackling #1605

## Author's checklist

### DIFF
--- a/src/sources.cmake
+++ b/src/sources.cmake
@@ -4,7 +4,7 @@
 
 target_sources(preciceCore
     PRIVATE
-    ${PROJECT_BINARY_DIR}/src/precice/impl/versions.cpp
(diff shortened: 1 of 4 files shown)


src/sources.cmake
```diff
@@ -4,7 +4,7 @@
 
 target_sources(preciceCore
     PRIVATE
-    ${PROJECT_BINARY_DIR}/src/precice/impl/versions.cpp
+    ${CMAKE_BINARY_DIR}/src/precice/impl/versions.cpp
```
davidscn, line 7: How is this change related? I guess it is incorrect, right?

PR #1623, opened 2023-04-21
1457 add api get max time step size

BenjaminRodenberg:
## Main changes of this PR

Introduces new API method `getMaxTimeStepSize` and updates tests correspondingly. `advance` and `initialize` don't return the maximum time step size anymore.

closes #1457 

## Motivation and additional information

see #1457 

## Author's checklist

* [x] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [x] I added a test to cover the proposed changes in our test suite.
* [ ] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).
* [x] I sticked to C++17 features.
* [x] I sticked to CMake version 3.16.3.
* [x] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

PR #1626, opened 2023-04-24
Replace std::string_view by span in API

### LINKED ISSUE #1605: preCICE C++ API requires C++ 17 in user codes

fsimonis:
## Main changes of this PR

This PR replaces the use of std::string_view in the `SolverInterface` with precice::span.
The latter is now part of the API and installed alongside the other headers.
The `SolverInterfaceImpl` still uses `std::string_view`, hence the `SolverInterface.cpp` converts between types.

This drops the API requirement from C++17 down to C++11.

I also checked for updates of the span.hpp file, our version is the latest.

Closes #1605

## Author's checklist

## Reviewers' checklist


examples/solverdummies/cpp/CMakeLists.txt
```diff
@@ -4,7 +4,7 @@ project(SolverdummyCPP LANGUAGES CXX DESCRIPTION "preCICE CPP solver dummy")
 find_package(precice REQUIRED CONFIG)
 
 add_executable(solverdummy solverdummy.cpp)
-set_target_properties(solverdummy PROPERTIES CXX_STANDARD 17)
+set_target_properties(solverdummy PROPERTIES CXX_STANDARD 11)
```
davidscn, line 7: Our solverdummies could be a good test case for our API compatibility with C++11. Why upgrading here?

src/precice/SolverInterface.cpp
```diff
@@ -1,26 +1,34 @@
...
 #include "precice/impl/versions.hpp"
 
 namespace precice {
 
+namespace {
+std::string_view toSV(precice::string_view sv)
+{
+  return {sv.data(), sv.size()};
+}
+} // namespace
```
davidscn, line 14: Wouldn't it be way easier to define an implicit conversion operator from `precice::span` to `std::span`? Or do you want to make the conversion explicit? Adds a bit of visual clutter IMO.

src/precice/SolverInterface.cpp
```diff
@@ -72,296 +80,296 @@ bool SolverInterface::requiresWritingCheckpoint()
   return _impl->requiresWritingCheckpoint();
 }
 
-bool SolverInterface::hasMesh(std::string_view meshName) const
+bool SolverInterface::hasMesh(::precice::string_view meshName) const
 {
-  return _impl->hasMesh(meshName);
+  return _impl->hasMesh(toSV(toSV(meshName)));
```
davidscn, line 85:
```suggestion
  return _impl->hasMesh(toSV(meshName));
```
Seems like there are a couple of instances where this occurs twice. Please check

src/precice/SolverInterface.hpp
```diff
@@ -1,7 +1,7 @@
 #pragma once
 
 #include <memory>
-#include <string_view>
+#include <precice/span.hpp>
```
davidscn, line 4:
```suggestion
#include "precice/span.hpp"
```

davidscn (commented):
Mostly trivial changes due to the altered type. Main question is probably regarding the conversion handling.

examples/solverdummies/cpp/CMakeLists.txt
fsimonis: Have a look at the colour :stuck_out_tongue_winking_eye:

davidscn: Time to go home

src/precice/SolverInterface.cpp
fsimonis: Agreed.
I would have preferred an implicit conversion, but I couldn't find a way how to add an implicit conversion as a non-member function.

I could add it to the span class, but this would enable it everywhere and make upgrading the span more difficult/error-prone. 
Given that the conversion is only used in this one cpp file, I added a separate function.

src/precice/SolverInterface.hpp
fsimonis: Given that this is part of the API, I would prefer the compiler to search system includes first, hence the `<>`.

davidscn: In principle fine, but why are the API includes (e.g. Version) searched the other way?

fsimonis: That's a very good point. They shouldn't be :see_no_evil:

src/precice/SolverInterface.cpp
fsimonis: Done, thanks for pointing it out.

src/precice/SolverInterface.cpp
fsimonis: I ran into issues where the span allows NULL characters inside the span.
So we also need to trim the view in the conversion. This kind of justifies a separate function.

fsimonis: This reason for this was also new to me. :wink:  

`span<const char>("123")` dispatches to the overload that handles arrays `template <class T, int N> span(T[N])`.
The C string literal is a C array `char const[4]` with value `{ '1', '2', '3', '\0' }`, hence size 4.

Passing the c string literal as a `const char *` however will dispatch to the c string constructor `span(const char *)`, which uses `std::strlen` to determine the size of the string.

So, we cannot dispatch directly to the `std::string_view(char*, size)` constructor, as the span may include a NULL char. Thus, we need to extract everything until the first (possibly absent) `\0` char.

davidscn: Thanks for the example, might be worth a small note in the code, unless you want anyone else in the future to unravel this.

fsimonis:
Rebased on #1623 to prepare for merge.

src/precice/SolverInterface.hpp
fsimonis: Fixed

PR #1630, opened 2023-04-25
Add Ginkgo and hipSOLVER GPU support for RBF mappings [continued]

davidscn:
## Main changes of this PR

Continuation of #1581.

FYI (and credits to) @Timo-Schrader 

## Open TODOs:

* [x] Migrate open discussions from #1581 to this PR (see open discussions in #1581)
* [x] Add unit tests for the GinkgoRBFSolver backend
    * [x] Might need adjustments of the `perform2DTestConservativeMappingVector` for the iterative solvers
    * [x] Add Ginkgo to precice-ci images
    * [x] we can recycle the existing unit tests of the RBF mapping and either use `BOOST_TEST_DATA` ([NP Example](https://github.com/precice/precice/blob/691fccdc406f6c58d4c222da5622bc26144a1b2f/tests/serial/mapping-nearest-projection/MappingNearestProjectionEdges.cpp#L12-L14)) or we create a separate test suite for the solver backend
* [x] Enable multiple mapping class instantiations when using `cuSOLVER`/`hipSOLVER`
* [x] Move QR solver config into `rbf-global-direct` mapping tag
* [x] Decide on user xml config options and the CMake variable name

## Author's checklist

## Reviewers' checklist


src/mapping/cuda_kernels/qr_decomp.cu
uekerman, line 1: should we move these files to `src/mapping/impl/`?

davidscn: We should consider creating a sub directory for the RBF mappings in general, i.e., re-arrange all files here. It becomes a bit messy right now.

CMakeLists.txt
```diff
@@ -108,6 +109,15 @@ add_feature_info(PETScMapping PRECICE_PETScMapping
...
    Requires MPICommunication.
   ")
+  add_feature_info(GinkgoParallelization PRECICE_GinkgoParallelization
+  "Enables the Ginkgo library backend which supports the usage of GPUs and OpenMP in radial basis function mappings.
+
+   Currently, this enables radial basis function mappings to leverage the compute power of Nvidia and AMD GPUs.
+   It uses iterative solvers to compute the solution of interpolation problems.
+
+   This feature can be enabled/disabled by setting the PRECICE_GinkgoParallelization CMake option.
+   Requires the Ginkgo library.
```
fsimonis, line 119:
```suggestion
   Requires the Ginkgo library and Eigen 3.4.
```

CMakeLists.txt
```diff
@@ -182,7 +192,9 @@ if(TPL_ENABLE_EIGEN3)
   # Use EIGEN3_ROOT to set the directory
   xsdk_tpl_require(EIGEN3 EIGEN3_INCLUDE_DIR)
 endif()
-find_package(Eigen3 3.3.7 REQUIRED)
+# We need Eigen 3.4.0 such that Eigen deals properly eith cuda
+# TODO: We should only conditionally increase the requirement
+find_package(Eigen3 3.4.0 REQUIRED)
```
fsimonis, line 197:
```suggestion
if(PRECICE_GinkgoParallelization)
  find_package(Eigen3 3.4.0 REQUIRED)
else()
  find_package(Eigen3 3.3.7 REQUIRED)
endif()
```

CMakeLists.txt
```diff
@@ -236,6 +248,69 @@ else()
...
 
+# Option: Ginkgo
+if (PRECICE_GinkgoParallelization)
+  find_package(Ginkgo 1.5 REQUIRED)
+  message(STATUS "Found Ginkgo ${GINKGO_PROJECT_VERSION} at ${GINKGO_INSTALL_PREFIX}")
+
+  # We aim for the same configuration as Ginkgo itself and check immediately for device compiler
+  # such that preciceCore inherits all relevant flags directly
+  # 1. Check if Ginkgo was built with CUDA and if CUDA is available on the system
+  if(${GINKGO_BUILD_CUDA})
```
fsimonis, line 259: You can directly specify the CMake var in an if

```suggestion
  if(GINKGO_BUILD_CUDA)
```

CMakeLists.txt
```diff
@@ -236,6 +248,69 @@ else()
...
+    # Similar to Ginkgo, we don't let CMake deduce the architecture, but inherit the Ginkgo flags
+    if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.18)
+      set(CMAKE_CUDA_ARCHITECTURES OFF)
+    endif()
+    add_compile_definitions("PRECICE_WITH_CUDA")
+    message(STATUS "Enabled CUDA support for preCICE via the Ginkgo library.")
+  endif()
+
+  # 2. Check if Ginkgo was built with HIP and if HIP is available on the system
+  if(${GINKGO_BUILD_HIP})
```
fsimonis, line 277:
```suggestion
  if(GINKGO_BUILD_HIP)
```

CMakeLists.txt
```diff
@@ -236,6 +248,69 @@ else()
...
+    enable_language(CUDA)
+    set(CMAKE_CUDA_STANDARD 17)
+    set(CMAKE_CUDA_STANDARD_REQUIRED ON)
+    set(CMAKE_CUDA_FLAGS "${GINKGO_CUDA_ARCH_FLAGS} --expt-relaxed-constexpr")
+
+    # Similar to Ginkgo, we don't let CMake deduce the architecture, but inherit the Ginkgo flags
+    if(CMAKE_VERSION VERSION_GREATER_EQUAL 3.18)
+      set(CMAKE_CUDA_ARCHITECTURES OFF)
+    endif()
+    add_compile_definitions("PRECICE_WITH_CUDA")
```
fsimonis, line 272: This should be added to the preciceCore target using `target_compile_definitions`.

CMakeLists.txt
```diff
@@ -236,6 +248,69 @@ else()
...
+    message(STATUS "Enabled CUDA support for preCICE via the Ginkgo library.")
+  endif()
+
+  # 2. Check if Ginkgo was built with HIP and if HIP is available on the system
+  if(${GINKGO_BUILD_HIP})
+    if(NOT DEFINED HIP_PATH)
```
fsimonis, line 278: It may be a good idea to be forwards compatible with CMake 3.21, which adds HIP as a language similar to CUDA.

CMakeLists.txt
```diff
@@ -338,6 +413,51 @@ else()
...
+# Setup Ginkgo
+if (PRECICE_GinkgoParallelization)
+  # Next we configure the common kernels for the matrix assembly
+  # This CMake part is still a shortcoming from the experimental Ginkgo we are using right now
+  # TODO: Remove as soon as GPU feature is in main branch
+  set(Ginkgo_Path "${Ginkgo_DIR}/.." CACHE STRING "Path to the experimental Ginkgo Repository of branch a195f856e.")
+  include(${Ginkgo_Path}/cmake/unified_kernels.cmake)
+
+  # This command registers Ginkgo kernels for different execution backends
+  ginkgo_add_unified_kernels(preciceCore src/mapping/ginkgo_kernels/GinkgoRBFKernels.cpp)
```
fsimonis, line 425: As a remember: This is still a tricky command as it adds various dependencies directly to the preciceCore target and relies on global includes and definitions, which will be a pain to maintain.

If the script ends up in the next release unchanged, then we need to either hardcode the kernel handling or copy it to our code-base and patch it.

fsimonis (commented):
Some CMake comments

davidscn:
One question which is kind of a leftover from #1417 is how this feature looks like from the configuration perspective. Right now we have the following tags:

```xml
<!-- partition of unity -->
<mapping:rbf-pou-direct vertices-per-cluster="..." relative-overlap="..."    />

<!-- the classical global rbf mapping -->
<mapping:rbf-global-direct  /> 
<mapping:rbf-global-iterative tolerance="..." max-iterations="..." preconditioner="..."  />

<!-- alias -->
<mapping:rbf  ... />
```

and we want to offer the user to specify and configure the executor (OpenMP, AMD via HIP, Cuda (via HIP or native), PETsc (which has also some GPU support), Eigen/CPU).

### Solution 1: introduce an additional attribute

Example:

```xml
<mapping:rbf-pou-direct vertices-per-cluster="..." relative-overlap="..." executor="..."   />
```
 where attribute describes the options above. However, then we need another attribute or solution for the configuration of the executor. Example: The GPU requires to specify a device ID, OpenMP requires the number of executor threads. Here, we could either introduce yet another attribute or rely on environment variables.

Advantage:

- selecting an executor is simple and intuitive from the user perspective

Disadvantage:

- configuring the device is rather nasty. Either we use generic attribute names to configure them (`option="..."`, where the option refers to e.g. number of threads or the device ID) or we need individual environment variable solutions
- the device configuration might be over-simplified. Example: Supporting multi-gpu environments at some point or things like cuda unified memory (which we don't need right now, but it was considered in the thesis). We might eventually need additional configuration options rather than just an integer value here. 

### Solution 2: add (yet another) subtag

The basis-function was already wrapped into a subtag and one of the main drawbacks of using subtags is the tedious closing tag of the parent tag (which we anyway have to live with already).

Example:

```xml
<mapping:rbf-pou-direct vertices-per-cluster="..." relative-overlap="..." >
  <basis-function:compact-polynomial-c4 support-radius="..." />
  <executor:open-mp n-threads=".." />
</mapping:rbf-pou-direct>
```

Advantage:

- tailored attributes for each corresponding option
- more freedom to add more or less configuration option (also in the future)

Disadvantage:

- configuration syntax is more intricate and complex
- implementation will become more intricate


An open question in general: how to treat this option with the alias tag.

MakisH:
@davidscn if we make all subtags optional, could we have both a start-close tag and a self-closing tag?

I get the impression that all of

```xml
<mapping:rbf-pou-direct vertices-per-cluster="..." relative-overlap="..." />
```

and

```xml
<mapping:rbf-pou-direct vertices-per-cluster="..." relative-overlap="..." >
</mapping:rbf-pou-direct>
```

and

```xml
<mapping:rbf-pou-direct vertices-per-cluster="..." relative-overlap="..." >
  <basis-function:compact-polynomial-c4 support-radius="..." />
  <executor:open-mp n-threads=".." />
</mapping:rbf-pou-direct>
```

could work at the same time, as an empty start-end tag is equivalent to a self-closing tag. Consider `</img>` and `` in HTML.

Besides the implementation details, I think that such a structured way makes the most sense for the user.

davidscn:
@MakisH thank you for the feedback. In principle, it is true that one could use self-closing tags, if no subtag is required. However, the basis -function is a mandatory tag and I would not want to make it default in any way, because there is no reasonable default value. For the executor subtag, we could/should create a default subtag, of course.

As a consequence, the self-closing tag is does not work (although it looks like it works above). I edited the above examples.

CMakeLists.txt
davidscn: We need the compile definitions in our tests as well. This would duplicate the line then, right?

MakisH:
But then, a subtag looks like the straight-forward best way here. Option 1 has the clear disadvantage that would need extra (seemingly unrelated and different) attributes for each executor. Regarding the disadvantages of option 2:

> configuration syntax is more intricate and complex

I actually see the subtag as more naturally structured, as it groups the attributes specific to each executor. I could imagine having multiple executor subtags and commenting-out/-in a different one each time. Changing the executor in option 1 sounds trickier to get right, as attributes would depend on each other.

> implementation will become more intricate

Only in length of code, or also in terms of main concepts and combinations to check? I assume it is only a bit more of what we aready have, in which case this is a weak disadvantage.

CMakeLists.txt
davidscn: We also need the deduced variables in the test configuration and in the testprecice target (cf [3a28c7c](https://github.com/precice/precice/pull/1630/commits/3a28c7cbe268026609beacb956e70523e7fbe093))

uekerman:
I also like option 2 (dedicated subtag), especially if we provide a default.

Let's assume we go for this option, which executors would we have then exactly (after this PR)?

davidscn:
> Let's assume we go for this option, which executors would we have then exactly (after this PR)?

We would have the following options:

For the direct solver

```xml
<mapping:rbf-global-direct >
  <basis-function:compact-polynomial-c4 support-radius="..." />

  <!-- the default eigen version -->
  <executor:cpu />

  <!-- for nvidia GPUs -->
  <executor:cuda device-id="..." />

  <!-- for amd GPUs, should also work with nvidia gpus (still have to look up if this is supported in ginkgo) -->
  <executor:hip device-id="..." />

</mapping:rbf-global-direct>
```

here the cuda and hip executor applies the QR factorization.

For the iterative solver

```xml
<mapping:rbf-global-iterative >
  <basis-function:compact-polynomial-c4 support-radius="..." />

  <!-- the default PETSc version (no idea about petsc and GPUs) -->
  <executor:cpu />

  <!-- thread parallel -->
  <executor:openmp n-threads="..." />

  <!-- for nvidia GPUs -->
  <executor:cuda device-id="..." />

  <!-- for amd GPUs-->
  <executor:hip device-id="..." />

</mapping:rbf-global-iterative>
```

The solver option would go into the main tag. However, I would probably only offer one solver per executor such that the option is not exposed for the user. In principle, we have for PETSc GMRES, all other executors have CG, GMRES, potentially with a preconditioner.

fsimonis:
The separate subtag looks very clear. We just need to communicate clearly what we use for the sake of reproducibility.

uekerman:
> The solver option would go into the main tag. However, I would probably only offer one solver per executor such that the option is not exposed for the user. In principle, we have for PETSc GMRES, all other executors have CG, GMRES, potentially with a preconditioner.

Yes, let's go with this. In case we need it, we could then still later add a solver attribute in a non-breaking way.

If a user builds without PETSc or Gingko, I guess we would then just throw an error:

> You picked an iterative solver on CPU, this requires PETSc, but you built without PETSc ...

src/mapping/device/CudaQRSolver.cu
```diff
@@ -0,0 +1,94 @@
...
+  cudaErrorCode  = cudaDeviceSynchronize();
+  assert(cusolverStatus == CUSOLVER_STATUS_SUCCESS);
+  assert(cudaSuccess == cudaErrorCode);
+
+  A_T->transpose(gko::lend(A_Q));
+
+  cudaDeviceSynchronize();
+  calculateQRDecompEvent.stop();
+
+  // Free the utilizted memory
```
Timo-Schrader, line 85: typo

src/mapping/device/CudaQRSolver.cuh
```diff
@@ -0,0 +1,11 @@
+#ifdef PRECICE_WITH_CUDA
+#pragma once
+
+#include <ginkgo/ginkgo.hpp>
+
+/**
+ * Computes the QR decomposition using CUDA
+*/
+void computeQRDecompositionCuda(const int deviceId, const std::shared_ptr<gko::Executor> &exec, gko::matrix::Dense<> *A_Q, gko::matrix::Dense<> *R);
```
Timo-Schrader, line 9: Should we rather use `std::shared_ptr<>`  (or a reference to it) instead of a `std::shared_ptr<>*`?

src/mapping/device/HipQRSolver.hip.cpp
```diff
@@ -0,0 +1,81 @@
...
+  hipsolverStatus = hipsolverDnDorgqr(solverHandle, M, N, k, A_T->get_values(), lda, dTau, (double *) dWork, lwork, devInfo);
+  hipErrorCode    = hipDeviceSynchronize();
+  assert(hipsolverStatus == HIPSOLVER_STATUS_SUCCESS);
+  assert(hipSuccess == hipErrorCode);
+
+  A_T->transpose(gko::lend(A_Q));
+
+  hipDeviceSynchronize();
+
+  // Free the utilized memory
```
Timo-Schrader, line 72: typo

davidscn: Not here, right?

davidscn:
Due to #1682 we now have to exclude all files located in `src/mapping/device*` (as they are handled by the device compiler).

I followed more or less the `tools/building/updateSourceFiles.py` hint and added the path to the script as follows

```diff
-IGNORE_PATTERNS = ["drivers"]
+IGNORE_PATTERNS = ["drivers", "${PROJECT_SOURCE_DIR}/src/mapping/device/*"]
```

but this doesn't seem to work. Since @fsimonis wrote the script, I could imagine that there is a very simple way to do what I want.

docs/changelog/1630.md
```diff
@@ -0,0 +1,3 @@
+- Added experimental support for building preCICE with Ginkgo.
+- Added a mapping configuration option for the executor, in order to compute and configure RBF data mappings for computations on GPUs/OpenMP. The executor can now be specified as a subtag.
+- Added support for executing RBF data mappings on different executors. preCICE now supports RBF data mappings executed on CUDA, HIP and OpenMP (as well as the previously supported CPU, which is also used by default).
```
uekerman, line 3:
```suggestion
- Added a mapping configuration subtag `executor` to specify where and how RBF data mappings should be computed: `cpu` (default), `cuda`, `hip`, or `openmp`.
```

src/testing/GlobalFixtures.cpp
```diff
@@ -47,7 +47,7 @@ std::string filterFromLogLevel(boost::unit_test::log_level logLevel)
   namespace bu = boost::unit_test;
 
   if (logLevel == bu::log_successful_tests || logLevel == bu::log_test_units) {
-    return "%Severity% >= debug";
+    return "";
   }
```
uekerman, line 51: Why this change?

src/testing/GlobalFixtures.cpp
```diff
@@ -87,7 +87,7 @@ void setupTestLogging()
 
     // The full debug log
     config.output = "test.debug.log";
-    config.filter = "%Severity% >= debug";
+    config.filter = "";
     logConfigs.push_back(config);
```
uekerman, line 91: Why this change?

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -1,11 +1,43 @@
...
 #include "math/math.hpp"
 
 namespace precice {
 namespace mapping {
 
+struct RadialBasisParameters {
+  double parameter1;
+  double parameter2;
+  double parameter3;
+};
```
uekerman, line 39: We probably need some documentation here. Unclear to me what these parameters mean exactly.

src/mapping/impl/BasisFunctions.hpp
uekerman, line 1: Many changes in this class seem to have no test coverage.

src/mapping/RadialBasisFctMapping.hpp
uekerman, line 1: Apparently, changes in this class have limited test coverage.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -204,6 +214,10 @@ MappingConfiguration::MappingConfiguration(
 
   auto attrSolverRtol = makeXMLAttribute(ATTR_SOLVER_RTOL, 1e-9)
                             .setDocumentation("Solver relative tolerance for convergence");
+  // TODO: Discuss whether we wanto to introduce this attribute
+  // auto attrMaxIterations = makeXMLAttribute(ATTR_MAX_ITERATIONS, 1e6)
+  //                              .setDocumentation("Maximum number of iterations of the solver");
```
uekerman, line 219: TODO
Do we really need this? I guess a good robust value could be enough. What did we use so far for PETSc, 10,000, right?

src/mapping/config/MappingConfiguration.hpp
```diff
@@ -169,6 +185,22 @@ class MappingConfiguration : public xml::XMLTag::Listener {
...
+  const std::string EXECUTOR_OMP    = "openmp";
+
+  const std::string ATTR_DEVICE_ID = "gpu-device-id";
+  const std::string ATTR_N_THREADS = "n-threads";
+  // const std::string ATTR_ENABLE_UNIFIED_MEMORY = "enable-unified-memory";
+  // const std::string ATTR_SOLVER                = "solver";
+  // const std::string ATTR_USE_PRECONDITIONER    = "use-preconditioner";
+  // const std::string ATTR_PRECONDITIONER        = "preconditioner";
+  // const std::string ATTR_JACOBI_BLOCK_SIZE     = "jacobi-block-size";
+  // const std::string ATTR_MAX_ITERATIONS        = "max-iterations";
```
uekerman, line 202: still needed?

src/mapping/tests/PartitionOfUnityMappingTest.cpp
```diff
@@ -366,9 +366,9 @@ void performTestConsistentMapDeadAxis(Mapping &mapping, int dim)
   double value2 = outData->values()(2);
   BOOST_TEST(mapping.hasComputedMapping() == true);
   if (dim == 2) {
-    BOOST_TEST(value == 19935.268150244759);
-    BOOST_TEST(value1 == 19935.266962237896);
-    BOOST_TEST(value2 == 19935.276990769267);
+    BOOST_TEST(value == 19935.374757794027);
+    BOOST_TEST(value1 == 19935.37795225389);
+    BOOST_TEST(value2 == 19935.35164938603);
```
uekerman, line 371: Why was this necessary? Seems that we should use less valid digits here.

uekerman (commented):
I only had a brief look. Would probably be helpful if @fsimonis could have a closer look, also to increase the bus factor here.

Could we also add integration tests? This way we could test the different configurations.

We need some user documentation of these new methods (and the config porting, didn't we do this already??) in the [mapping config docs](https://github.com/precice/precice.github.io/blob/precice-v3/pages/docs/configuration/configuration-mapping.md#radial-basis-function-mapping).

src/mapping/impl/BasisFunctions.hpp
Timo-Schrader: In this form, they don't have an exact meaning apart from being RBF parameters (shape, radius, etc.). Initally, I wanted to move RBF parameters to the GPU using `std::array` that is part of each RBF kernel class. However, AMD does not support this container on its GPUs because of `constexpr` of the `.at()` method. Hence, each RBF kernel fills 0-3 parameters in this struct, which is then given to the RBF kernel lambda that is executed on the GPU. Having it this way, we can have one single evaluation device kernel on both Nvidia and AMD by always passing this parameter struct to the GPU kernel. Does this clarify it?

uekerman: yes, thanks :pray: 
but let's at this documentation also in the source code

Timo-Schrader: Hi, sorry for the delay, but I added a docstring here: https://github.com/Timo-Schrader/precice/tree/patch-1

Can open a PR if you are fine with that version. Let me know if you need anything else!

src/testing/GlobalFixtures.cpp
davidscn: I was facing a (quite intricate) bug and our previous configuration did not provide any option to inspect trace logging from out tests. We should either add an option to get additional trace logging from our test suite or (what this does) add the trace logging by default to the most verbose DEBUG log stream boost test provides.

src/mapping/impl/BasisFunctions.hpp
davidscn: As far as I can say from now, I will wrap up the whole PR next week and then cherry-pick your changes. Thanks!

src/testing/GlobalFixtures.cpp
uekerman: Why couldn't you get trace output from our tests? For me, this always worked in the past via `log.conf`.

src/mapping/tests/PartitionOfUnityMappingTest.cpp
uekerman: but then it would be more sustainable to reduce the test accuracy here, right? otherwise, we eventually run into the same problem again.

CMakeLists.txt
```diff
@@ -436,6 +563,9 @@ IF (BUILD_TESTING)
     ${preCICE_SOURCE_DIR}/src
     ${preCICE_SOURCE_DIR}/tests
     )
+  if(PRECICE_WITH_OMP)
+    target_compile_options(testprecice PRIVATE "-fopenmp")
```
fsimonis, line 567: This should be handled in preciceCore as explained above.

cmake/CTestConfig.cmake
```diff
@@ -55,11 +59,12 @@ function(add_precice_test)
   set(PAT_WDIR "${PRECICE_TEST_DIR}/${PAT_NAME}")
   file(MAKE_DIRECTORY "${PAT_WDIR}")
   # Setting properties
+  # Question: Do we want to set the OMP env only if PAT_GINKGO_OMP?
```
fsimonis, line 62: I would say that it's fine. It impacts only our tests and it is less code.

src/testing/GlobalFixtures.cpp
davidscn: Using a CLI (from `ctest`) still sounds more easy to me.

CMakeLists.txt
```diff
@@ -346,6 +418,61 @@ else()
...
+        NVCC_OPTIONS ${GINKGO_HIP_NVCC_OPTIONS}
+        STATIC)
+        target_link_libraries(preciceCore_qr_hip PRIVATE Ginkgo::ginkgo roc::hipsolver)
+        target_link_libraries(preciceCore PRIVATE preciceCore_qr_hip)
+    message(STATUS "Enabled HIP support for preCICE via the Ginkgo library.")
+  endif()
+
+  if(PRECICE_WITH_OMP)
+    target_compile_definitions(preciceCore PUBLIC PRECICE_WITH_OMP)
+    # There is nothing special required for OpenMP, looking for OpenMP and setting the flag is handled by the unified kernels
```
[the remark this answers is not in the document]
davidscn: In principle, the FindOpenMP module is already used by the Ginkgo cmake unified kernel interface. Then, we call it here twice, in the overall workflow.

src/mapping/config/MappingConfiguration.cpp
davidscn: Yes, in PETSc we use a hard-coded value of 10.000. I'm not sure about a robust default. In principle, the max iteration should depend on the system size . For small systems, reaching this iteration count is a waste of time and for large systems it might be too less.
Do we really need this? I don't have a strong opinion on this. The solver tolerance would be the more natural parameter to modify, I guess.

src/mapping/config/MappingConfiguration.hpp
davidscn: It would be good to keep an overview of available options for the moment, yes.

src/mapping/impl/BasisFunctions.hpp
davidscn: The coverage is now good, this was maybe a problem with missing coverage reports.

davidscn:
There is still some room for improvements, for example in the CMake setup or in some code duplication. However, I think the changes are good to go for now. Given that it comes with rather large changes, I don't want to merge this too short before the upcoming release. The future of this implementation will also heavily depend on the development of Ginkgo.

PR #1633, opened 2023-04-27
Introduce Wrappers for better encapsulation of Read/WriteDataContext

BenjaminRodenberg:
## Main changes of this PR

Introduces wrappers in Read/WriteDataContext for better encapsulation of mesh.

* ~~Reduces amount of code duplication and preparatory refactoring for #1612.~~ (Code duplication was already taken care of in #1636)
* Improves encapsulation, especially allows to remove getter for `_providedData`.

## Author's checklist

* [x] ~~I added a changelog file with `make changelog` if there are user-observable changes since the last release.~~ (not needed, because refactoring)

## Reviewers' checklist

### DIFF
--- a/src/precice/impl/DataContext.hpp
+++ b/src/precice/impl/DataContext.hpp
@@ -48,6 +48,13 @@ public:
   int getDataDimensions() const;
 
   /**
+   * @brief Get the spatial dimensions of _providedData.
+   *
+   * @return int Spatial dimensions of _providedData.
+   */
+  int getSpatialDimensions() const;
+
+  /**
    * @brief Get the name of _mesh.
    *
    * @return std::string Name of _mesh.
@@ -62,9 +69,12 @@ public:
   MeshID getMeshID() const;
 
   /**
-   * @brief Get the Mesh
+   * @brief Returns whether _providedData has gradient
+   *
+   * @return true, if it has gradient
+   * @return false, if it has gradient
    */
-  const mesh::Mesh &getMesh() const;
+  bool hasGradient() const;
 
   /**
    * @brief Perform the mapping for all mapping contexts and the corresponding data context (from and to data)
@@ -131,6 +141,16 @@ protected:
    * @return True, if DataContext has any write mapping.
    */
   bool hasWriteMapping() const;
+
+  /**
+   * @brief Get the number of vertices of mesh
+   *
+   * @return int number of vertices
+   */
+  int getMeshVertexCount() const;
+
+  /// Returns true if the given vertexID is valid
+  bool isValidVertexID(const VertexID id) const;
 
 private:
   /// Unique mesh associated with _providedData.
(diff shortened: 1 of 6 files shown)


BenjaminRodenberg:
@davidscn Can you please especially check the gradient data interface? I think I go the dimensions right, but it probably does not harm. If we can also use the `.block` functionality from Eigen in the other functions, this would be nice, but I don't really know how it works and don't have the time to also change this part of the code.

src/precice/impl/WriteDataContext.hpp
```diff
@@ -32,6 +32,22 @@ class WriteDataContext : public DataContext {
...
+   */
+  void writeValues(std::vector<int> indices, const Eigen::Map<const Eigen::VectorXd> values);
+
+  /**
+   * @brief Store gradients in _providedData.gradientValues()
+   *
+   * @param[in] indices ids of data
+   * @param[in] gradients gradients of data
+   */
+  void writeGradientValues(std::vector<int> indices, const Eigen::Map<const Eigen::MatrixXd> gradients);
```
davidscn, line 49: These are potentially large data sets. I would rather use a reference instead of copies?!

davidscn (approved):
The gradient API functions are very well tested. If the tests pass, I happily approve.

src/precice/impl/WriteDataContext.hpp
BenjaminRodenberg: As far as I understand an `Eigen::Map` works more or less like a reference:

> While one option is to make a copy of the data, most commonly you probably want to re-use this memory as an Eigen type. Fortunately, this is very easy with the [Map](https://eigen.tuxfamily.org/dox/classEigen_1_1Map.html) class.

https://eigen.tuxfamily.org/dox/group__TutorialMapClass.html

BenjaminRodenberg: Done. [5cec86c](https://github.com/precice/precice/pull/1633/commits/5cec86c584d550de4f13ea538308819dd343a3ee)

src/precice/impl/DataContext.hpp
```diff
@@ -44,6 +44,20 @@ class DataContext {
    */
   int getDataDimensions() const;
 
+  /**
+   * @brief Get the spatial dimensions of _providedData.
+   *
+   * @return int Spatial dimensions of _providedData.
+   */
+  int getSpatialDimensions() const;
```
uekerman, line 52: in the API, we now call this `getMeshDimensions`. Does this make sense here as well?

BenjaminRodenberg: Puh... I'm a bit unsure. This has quite some implications that don't really belong here in my opinion. There is, for example, also a `Data::getMeshDimensions()` and `Data::_spatialDimensions`. I would rather rename this is another PR.

fsimonis:
I cleaned the PR up. It's ready to merge once the CI agrees.

fsimonis (approved):

PR #1636, opened 2023-05-03
Simplify Data API

fsimonis:
## Main changes of this PR

This PR simplifies the data API to only take ranges, which need to be consistent.

Closes https://github.com/precice/precice/issues/1559

## Author's checklist

## Reviewers' checklist


BenjaminRodenberg (approved):
Did not check the tests in detail, but nice that you were able to remove so much code and if the tests still work it should be fine.

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -978,364 +978,58 @@ void SolverInterfaceImpl::setMeshTetrahedra(
...
+    ::precice::span<const double>   values)
 {
-  PRECICE_TRACE(meshName, dataName, size);
-  PRECICE_CHECK(_state != State::Finalized, "writeBlockVectorData(...) cannot be called after finalize().");
+  PRECICE_TRACE(meshName, dataName, vertices.size());
+  PRECICE_CHECK(_state != State::Finalized, "writeData(...) cannot be called after finalize().");
   PRECICE_REQUIRE_DATA_WRITE(meshName, dataName);
-  if (size == 0)
+  if (vertices.empty()) {
     return;
```
uekerman, line 991: Couldn't we add a check here that `values` is also empty?

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1345,157 +1039,86 @@ void SolverInterfaceImpl::readBlockVectorData(
     PRECICE_CHECK(relativeReadTime == _couplingScheme->getNextTimeStepMaxSize(), "Waveform relaxation is not allowed for solver that sets the time step size");
     normalizedReadTime = 1; // by default read at end of window.
   }
+
   PRECICE_REQUIRE_DATA_READ(meshName, dataName);
-  if (size == 0)
+
+  if (vertices.empty()) {
     return;
```
uekerman, line 1046: Same as above

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -1345,157 +1039,86 @@ void SolverInterfaceImpl::readBlockVectorData(
...
-  }
-  PRECICE_REQUIRE_DATA_READ(meshName, dataName);
-  if (size == 0)
+  PRECICE_EXPERIMENTAL_API();
+
+  // Asserts and checks
+  PRECICE_TRACE(meshName, dataName, vertices.size());
+  PRECICE_CHECK(_state != State::Finalized, "writeGradientData(...) cannot be called after finalize().");
+  PRECICE_REQUIRE_DATA_WRITE(meshName, dataName);
+  if (vertices.empty() || !requiresGradientDataFor(meshName, dataName)) {
```
uekerman, line 1085: same as above

uekerman (approved):
Nice job. Only a few things to tweak. No further review required from my side.

src/precice/impl/DataContext.hpp
```diff
@@ -58,6 +58,11 @@ class DataContext {
    */
   MeshID getMeshID() const;
 
+  /**
+   * @brief Get the Mesh
+   */
+  mesh::PtrMesh getMesh() const;
```
BenjaminRodenberg, line 64: I'm not a big fan of exposing the mesh via the data context, because one might do bad things with the mesh then. As far as I see, you are just exposing the mesh for 1) getting the dimension (`mesh.getDimensions()`), 2) getting the number of vertices (`mesh.vertices().size()`), and 3) checking the vertex ids (`isValidVertexID(vid)`). Why not provide dedicated functions for this for better encapsulation?

extras/bindings/c/src/SolverInterfaceC.cpp
```diff
@@ -251,72 +251,19 @@ void precicec_setMeshTetrahedra(
...
     const char *  meshName,
     const char *  dataName,
     int           size,
     const int *   valueIndices,
     const double *values)
 {
   PRECICE_CHECK(impl != nullptr, errormsg);
-  impl->writeBlockVectorData(meshName, dataName, size, valueIndices, values);
+  auto dataSize = size * impl->getDataDimensions(meshName, dataName);
+  impl->writeData(meshName, dataName, {valueIndices, static_cast<unsigned long>(size)}, {values, static_cast<unsigned long>(dataSize)});
```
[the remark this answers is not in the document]
fsimonis: Constructions using the braced initializer syntax forbid implicit conversions. The span size is unsigned, so this leads to an error.

src/precice/impl/DataContext.hpp
fsimonis: The solution to this is `const mesh::Mesh& getMesh() const;`.

I am not a big fan of unnecessary wrappers. They make refactoring unnecessary painful.

src/precice/tests/ParallelTests.cpp
```diff
@@ -102,7 +102,7 @@ void multiCouplingThreeSolversParallelControl(const std::string configFile, cons
         }
 
         maxDt = cplInterface.getMaxTimeStepSize();
-        cplInterface.readScalarData(meshName, dataBAID, vertex2, maxDt, valueRead);
+        cplInterface.readData(meshName, dataBAID, {&vertex2, 1}, maxDt, {&valueRead, 1});
```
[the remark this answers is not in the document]
fsimonis: True, this hits our tests hard, but shouldn't be an issue for real use-cases.

src/precice/impl/DataContext.hpp
BenjaminRodenberg: I don't see these wrappers as unnecessary. In the past introducing such wrappers helped me a lot when refactoring the code base. From my perspective one main problem in our design is that often there are long chains of calls that make it really hard to understand in which context, for example, `Mesh::vertices()` is used. `vertices().size()` has around 200 hits in our code base. Introducing a wrapper like `Mesh::verticesSize()` allows me to ignore these 200 hits and concentrate on the rest.

PR #1639, opened 2023-05-04
Directly perform extrapolation in time::Storage, only allow CouplingScheme to moveToNextWindow

BenjaminRodenberg:
## Main changes of this PR

Replaces #1520 and #1504. Extrapolation now happens directly inside Storage and Storage is directly maintained in CouplingData.

* Allows to remove dedicated `Extrapolation` class
* Allows to remove `receiveResultOfFirstAdvance` from `ParticipantImpl`
* Simplifies maintainment of storage from `CouplingScheme` via `CouplingData`
* Only `CouplingScheme` takes care of `moveToNextWindow` (previously also partially done by `ParticipantImpl`)

## Author's checklist

* [x] ~~I added a changelog file with `make changelog` if there are user-observable changes since the last release.~~ (not needed, basically refactoring)
* [x] Merge #1612 

## Reviewers' checklist


src/time/Storage.cpp
```diff
@@ -26,9 +26,9 @@ void Storage::setSampleAtTime(double time, Sample sample)
   PRECICE_ASSERT(math::smallerEquals(WINDOW_START, time), "Setting values outside of valid range!");
   PRECICE_ASSERT(math::smallerEquals(time, WINDOW_END), "Sampling outside of valid range!");
   // check if key "time" exists.
-  auto existingStample = std::find_if(_stampleStorage.begin(), _stampleStorage.end(), [&time](const auto &s) { return math::equals(s.timestamp, time); });
-  if (existingStample == _stampleStorage.end()) { // key does not exist yet
-    PRECICE_ASSERT(math::smaller(maxStoredNormalizedDt(), time), maxStoredNormalizedDt(), time, "Trying to write values with a time that is too small. Please use trim() or clear(), if you want to write new samples to the storage.");
+  auto existingSample = std::find_if(_stampleStorage.begin(), _stampleStorage.end(), [&time](const auto &s) { return math::equals(s.timestamp, time); });
+  if (existingSample == _stampleStorage.end()) { // key does not exist yet
+    PRECICE_ASSERT(math::smaller(maxStoredNormalizedDt(), time), maxStoredNormalizedDt(), time, "Trying to write values with a time that is too small. Please use clear(), if you want to write new samples to the storage.");
```
uekerman, line 31: Is `math::smaller` here the right thing to do? Shouldn't it be `not math::larger`?

src/time/tests/StorageTest.cpp
```diff
@@ -102,5 +102,106 @@ BOOST_AUTO_TEST_CASE(testGetTimesAndValues)
   }
 }
 
+BOOST_AUTO_TEST_SUITE(ExtrapolationTests)
+BOOST_AUTO_TEST_CASE(testExtrapolateDataZerothOrder)
+{
+  PRECICE_TEST(1_rank);
+
+  // Test first order extrapolation
```
uekerman, line 110:
```suggestion
  // Test zeroth order extrapolation
```

src/cplscheme/CouplingScheme.cpp
```diff
@@ -8,7 +8,7 @@ const int CouplingScheme::UNDEFINED_TIME_WINDOWS = -1;
 
 const double CouplingScheme::UNDEFINED_TIME_WINDOW_SIZE = -1.0;
 
-const int CouplingScheme::UNDEFINED_EXTRAPOLATION_ORDER = -1;
+const int CouplingScheme::UNDEFINED_EXTRAPOLATION_ORDER = 0; // by default use zeroth order extrapolation.
```
uekerman, line 11: Why this change? Shouldn't the configuration handle this?
At least, the name of the variable is confusing now.

src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -254,8 +254,9 @@ class BaseCouplingScheme : public CouplingScheme {
    *
    * @param m2n M2N used for communication
    * @param receiveData DataMap associated with received data
+   * @param initialCommunication if true, will store received data for WINDOW_START, else store received data for WINDOW_END
```
uekerman, line 257:
```suggestion
   * @param initialCommunication if true, will store received data for WINDOW_START and WINDOW_END, else store received data only for WINDOW_END
```
Right?

src/cplscheme/SerialCouplingScheme.cpp
```diff
@@ -93,10 +93,13 @@ void SerialCouplingScheme::performReceiveOfFirstAdvance()
   if (doesFirstStep()) {
     // do nothing
   } else { // second participant
+    // similar to SerialCouplingScheme::exchangeSecondData()
     receiveAndSetTimeWindowSize();
     PRECICE_DEBUG("Receiving data...");
     receiveData(getM2N(), getReceiveData());
-    checkDataHasBeenReceived();
+    if (not hasDataBeenReceived()) { // check is required, because calling checkDataHasBeenReceived() if hasDataBeenReceived() == true would trigger an assertion. This situation can only occur during initialization of the second participant of a SerialCouplingScheme, because it may call receive twice: Once for receiving initial data, once for receiving the result of the first advance.
+      checkDataHasBeenReceived();
```
uekerman, line 101: This construction looks odd. Looks a lot like `hasDataBeenReceived` should be set to `false` somewhere.

src/cplscheme/SerialCouplingScheme.cpp
```diff
@@ -108,26 +111,46 @@ void SerialCouplingScheme::exchangeFirstData()
...
 {
+  // @todo bundle receiveConvergence, specific moveToNextWindow and sendData into one function?
   if (doesFirstStep()) { // first participant
-    PRECICE_DEBUG("Receiving convergence data...");
     if (isImplicitCouplingScheme()) {
+      PRECICE_DEBUG("Receiving convergence data...");
       receiveConvergence(getM2N());
     }
+    if (hasConverged() || isExplicitCouplingScheme()) {
+      moveToNextWindow(); // extrapolation result for receive data of first is directly overwritten in next step
```
uekerman, line 142: To what do you refer with "next step" here?

src/cplscheme/ParallelCouplingScheme.cpp
```diff
@@ -37,22 +37,43 @@ void ParallelCouplingScheme::exchangeFirstData()
 
 void ParallelCouplingScheme::exchangeSecondData()
 {
+  // @todo bundle receiveConvergence, specific moveToNextWindow and sendData into one function?
```
uekerman, line 40:
```suggestion
```
IMO no. Would only make things harder to read. More occasions below.

src/cplscheme/ParallelCouplingScheme.cpp
```diff
@@ -37,22 +37,43 @@ void ParallelCouplingScheme::exchangeFirstData()
...
+    if (hasConverged() || isExplicitCouplingScheme()) {
+      moveToNextWindow();
+    }
     PRECICE_DEBUG("Sending data...");
     sendData(getM2N(), getSendData());
   }
+
+  if (isImplicitCouplingScheme()) {
+    storeIteration();
+  }
```
uekerman, line 76: Readability-wise, wouldn't it be simpler to restructure here into on big:
```cpp
if (isExplicitCouplingScheme()) {
...
} else {
  assert(isImplicitCouplingScheme());
 ...
}
```
(same in `SerialCouplingScheme` class)

uekerman (approved):
Looks good, only minor tweaks

src/cplscheme/SerialCouplingScheme.cpp
BenjaminRodenberg: We could do this before `performReceiveOfFirstAdvance()` in `BaseCouplingScheme::initialize`, but then we again need to differentiate between the different coupling schemes, because this only has to be done for the `SerialCouplingScheme`. If we want to set `_hasDataBeenReceived = false` inside of `SerialCouplingScheme::performReceiveOfFirstAdvance()`, we do not really gain a lot and we will have to make `_hasDataBeenReceived` protected (currently it's private).

Be aware that this work-around will also go away as soon as we can remove `performReceiveOfFirstAdvance()` (which will be the case in #1523 when we are able to send and receive the complete storage at once).

src/cplscheme/CouplingScheme.cpp
BenjaminRodenberg: I changed this a bit in cb867ca3b11c66b8ece01def1abbfce11296f387. Does this look better?

Why do we need this? We need to always perform an extrapolation now, because we always extrapolate, if we move (in most cases just constant). Previously we did not always perform the extrapolation. But I agree: It's reasonable to be able to differentiate between no extrapolation and zeroth order extrapolation (explicitly set in the config).

src/precice/impl/WriteDataContext.cpp
```diff
@@ -31,7 +31,7 @@ void WriteDataContext::resetData(bool atEndOfWindow)
 
   if (atEndOfWindow) {
     // need to not only clear _providedData->timeStepsStorage(), but also _toData->timeStepsStorage() as soon as we map from storage to storage.
-    _providedData->timeStepsStorage().clear();
+    _providedData->timeStepsStorage().trim();
```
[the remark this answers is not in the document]
BenjaminRodenberg: Yes. This is outdated. Removed it. b5844bebfaa4edff2f7a3b10ea45a1a69a83e026

src/cplscheme/SerialCouplingScheme.cpp
BenjaminRodenberg: That was unclear. Changed it in 600e855b290b0e95e884812000cdf05a35fe41e4

src/cplscheme/ParallelCouplingScheme.cpp
BenjaminRodenberg: Ok. Removed the comments in 1934853707403ada41aa9c8b838351ecf883fb89

src/time/Storage.cpp
BenjaminRodenberg: No. Technically `not math::greater` and `math::smallerEquals` are the same, but if we have `time == mexStoredNormalizedDt()` we would overwrite and not append. So the other branch of the `if`.

src/cplscheme/ParallelCouplingScheme.cpp
BenjaminRodenberg: Not so sure. Changed it in e3c1f51484c2ca3817b98efc4021712b359d418d, we can still revert. But this way we at least have both variants to compare.

src/cplscheme/SerialCouplingScheme.cpp
BenjaminRodenberg: Actually. Let's just remove `performReceiveOfFirstAdvance` here. It's so much simpler. d73a326c734986c24553ed4fe7731d7da710eacd

src/cplscheme/CouplingScheme.cpp
uekerman: Looks better already. A bit nicer could be if the order is always set in the configuration. If the user doesn't set it, it defaults to zero. I thought this is how we handled it so far as well. Also before waveform things, we always extrapolated at least at order zero.

src/cplscheme/ParallelCouplingScheme.cpp
uekerman: Much easier to read IMO. One starts with the easy thing (explicit) and understands the basics and then moves to the complicated thing (implicit) and sees what else is still important. 
There are even more `isImplicitCouplingScheme()` you could remove.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -310,18 +300,10 @@ void BaseCouplingScheme::secondExchange()
...
-}
-
 void BaseCouplingScheme::moveToNextWindow()
 {
   PRECICE_TRACE(_timeWindows);
-  for (auto &data : getAccelerationData() | boost::adaptors::map_values) {
+  for (auto &data : _allData | boost::adaptors::map_values) {
     data->moveToNextWindow();
   }
 }
```
BenjaminRodenberg, line 309: Regarding extrapolation this is one important bit I changed: Before, we only called `moveToNextWindow` for the accelerated data. This applied the extrapolation. Now, we move `_allData`.

Especially for explicit coupling schemes this part changed a lot: Previously this function was never called for explicit coupling schemes, because extrapolation is not reasonable for explicit coupling schemes:
```
    if (not doesFirstStep()) {
      storeExtrapolationData();
      moveToNextWindow();
    }
  }
```
Now, we always call it for all coupling schemes, because all coupling schemes have to move the storage and extrapolation is just an optional feature of moving.

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -408,12 +398,6 @@ void ParticipantImpl::advance(
 
   advanceCouplingScheme();
 
-  if (_couplingScheme->isTimeWindowComplete()) {
-    for (auto &context : _accessor->readDataContexts()) {
-      context.moveToNextWindow();
-    }
-  }
```
BenjaminRodenberg, line 415: Also important regarding extrapolation: Previously we called `moveToNextWindow` for the storage of the `ReadDataContext` here. I moved this inside the coupling scheme, because then we don't have to worry about convergence etc. in `ParticipantImpl` anymore.

src/cplscheme/CouplingData.cpp
```diff
@@ -129,26 +129,14 @@ std::vector<int> CouplingData::getVertexOffsets()
...
-  PRECICE_ASSERT(stamples.size() > 0);
-  this->values() = stamples.back().sample.values;
-  _extrapolation.store(values());
+  if (this->timeStepsStorage().stamples().size() > 0) {
+    this->timeStepsStorage().move();
+    auto atEnd = this->timeStepsStorage().stamples().back();
+    PRECICE_ASSERT(math::equals(atEnd.timestamp, time::Storage::WINDOW_END));
+    _data->sample() = atEnd.sample;
+  }
 }
```
BenjaminRodenberg, line 140: Previously, this did not move the Storage of the coupling data, but computed the extrapolation (remember: for acceleration data) and stored it at the `WINDOW_END` of the Storage. Technically this is the same like moving, but the implementation is different.

BenjaminRodenberg (commented):
I added a few comments on how we dealt with extrapolation previously and what we do now. Hopefully this clarifies the open point about `UNDEFINED_EXTRAPOLATION_ORDER`.

One important part that I did not touch in this PR, but that is relevant: An explicit coupling scheme must use `UNDEFINED_EXTRAPOLATION_ORDER`, while an implicit coupling scheme must use `(_extrapolationOrder == 0) || (_extrapolationOrder == 1)`. See in `BaseCouplingScheme.cpp`:

```
  if (isExplicitCouplingScheme()) {
    PRECICE_ASSERT(_extrapolationOrder == UNDEFINED_EXTRAPOLATION_ORDER, "Extrapolation is not allowed for explicit coupling");
  } else {
    PRECICE_ASSERT(isImplicitCouplingScheme());
    PRECICE_CHECK((_extrapolationOrder == 0) || (_extrapolationOrder == 1),
                  "Extrapolation order has to be 0 or 1.");
  }
```

BenjaminRodenberg:
[1c676d6](https://github.com/precice/precice/pull/1639/commits/1c676d6ac8d0a18dbc6fb0d51d8f8d26267a8f08) is interesting: I think I found a fix, but at the moment this test also fails on develop. Meaning: There is probably a bug with respect to this combination of features on develop. I will try to understand this better on develop by using this test there tomorrow.

BenjaminRodenberg:
As directly discussed with @uekerman: This PR is ready to go and I will merge as soon as the tests have completed. There are still some details that we did not 100% finalize yet, but this will change and we will reiterate on these details soon again when subcycling + higher interpolation is implemented.

PR #1641, opened 2023-05-05
Port Mesh API to spans

fsimonis:
## Main changes of this PR

This PR ports the mesh API of preCICE to using spans.

C and Fortran assume correctly sized memory sections.

Better error messages based on stricter rules and fewer arguments for C++ users of the interface.

## Author's checklist

* [ ] I updated the doxygen documentation

## Reviewers' checklist

### DIFF
--- a/docs/documents/span.dox
+++ b/docs/documents/span.dox
@@ -1,0 +1,10 @@
+/** @class precice::span
+ *
+ * @copyright Copyright Tristan Brindle 2018.
+ *
+ * @brief A C++ 11 implementation of the non-owning C++20 std::span type.
+ *
+ * A span represents a non-owning view into a contiguous block of memory of a known size.
+ *
+ * This version contains an additional constructor for `const char *`-style C strings, allowing the span to mimic C++17 std::string_view.
+ */
--- a/src/precice/SolverInterface.hpp
+++ b/src/precice/SolverInterface.hpp
@@ -594,7 +594,7 @@ public:
    * @param[in] values the values to write to preCICE.
    *
    * @pre every VertexID in vertices is a return value of setMeshVertex or setMeshVertices
-   * @pre values.size() = getDataDimensions(meshName, dataName) * vertices.size()
+   * @pre values.size() == getDataDimensions(meshName, dataName) * vertices.size()
    *
    * @see SolverInterface::setMeshVertex()
    * @see SolverInterface::setMeshVertices()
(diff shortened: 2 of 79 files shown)


docs/documents/span.dox
```diff
@@ -0,0 +1,10 @@
+/** @class precice::span
+ *
+ * @copyright Copyright Tristan Brindle 2018.
```
uekerman, line 3: Do we need to ship a third-party license with this?

src/precice/SolverInterface.hpp
```diff
@@ -602,7 +594,7 @@ class PRECICE_API SolverInterface {
    * @param[in] values the values to write to preCICE.
    *
    * @pre every VertexID in vertices is a return value of setMeshVertex or setMeshVertices
-   * @pre values has the size = getDataDimensions(meshName, dataName) * vertices.size
+   * @pre values.size() = getDataDimensions(meshName, dataName) * vertices.size()
```
uekerman, line 597:
```suggestion
   * @pre values.size() == getDataDimensions(meshName, dataName) * vertices.size()
```

src/precice/SolverInterface.hpp
```diff
@@ -636,7 +628,7 @@ class PRECICE_API SolverInterface {
    * @param[out] values the destination memory to read the data from.
    *
    * @pre every VertexID in vertices is a return value of setMeshVertex or setMeshVertices
-   * @pre values has the size = getDataDimensions(meshName, dataName) * vertices.size
+   * @pre values.size() = getDataDimensions(meshName, dataName) * vertices.size()
```
uekerman, line 631:
```suggestion
   * @pre values.size() == getDataDimensions(meshName, dataName) * vertices.size()
```

src/precice/impl/SolverInterfaceImpl.cpp
```diff
@@ -701,30 +703,33 @@ void SolverInterfaceImpl::setMeshEdge(
...
   PRECICE_REQUIRE_MESH_MODIFY(meshName);
   MeshContext &context = _accessor->usedMeshContext(meshName);
   if (context.meshRequirement != mapping::Mapping::MeshRequirement::FULL) {
     return;
   }
 
   mesh::PtrMesh &mesh = context.mesh;
+  PRECICE_CHECK(vertices.size() % 2 == 0,
+                "Cannot interpret passed vertex IDs attempting to set edges of mesh \"{}\" . "
+                "You passed {} vertices, but we expected an odd number.",
```
uekerman, line 719:
```suggestion
                "You passed {} vertices, but we expected an even number.",
```

uekerman (approved):
Looks good to me. I did not check the integration tests and had a rather quick look over the rest.

Changelog still missing.

docs/documents/span.dox
fsimonis: It's included in the header file. That should suffice, right?

fsimonis: After a chat with @uekerman regarding this:
* comment was regarding the licence text itself. The header only states which licence it uses
* This should suffice.

PR #1627, opened 2023-04-24
Rename XML tag solver-interface

MakisH:
## Main changes of this PR

Related to #1458. Renames the XML tag `<solver-interface>` to `<simulation-setup>` and adjusts the tests accordingly.

## Motivation and additional information

More related PRs follow, this is a separate one to facilitate reviewing (too many test configs).

## Author's checklist

* [ ] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* I added a test to cover the proposed changes in our test suite. -> N/A
* [ ] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).
* [x] I sticked to C++17 features.
* [x] I sticked to CMake version 3.16.3.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

Issue #1458, opened 2022-11-04
Rename SolverInterface?


uekerman:
As we are breaking, we could open the box and discuss whether it makes sense to rename `SolverInterface`.

I always found the name confusing, what is it an interface for / to? From user perspective it is an interface or a handle to preCICE. Or should it be an interface to the other solver? From dev perspective, it is also does not really represent a solver ... especially as we also have the `Participant` class. An "interface to solver" sounds a lot like a framework approach.

Why not calling it `Precice`? (other than the ugly spelling)
When explaining preCICE and in many integration tests, we call the object `precice` anway

From [step-by-step guide](https://precice.org/couple-your-code-steering-methods.html):

```cpp
precice::SolverInterface precice("FluidSolver","precice-config.xml",rank,size);
```

What about the name of the header? I think back in the day, it made sense to not call it `precice.hpp` as we had multiple objects (`Constants`, `MeshHandle`, and `SolverInterface`), but now we only have one header to include.

Opinions?

It would also mean that we rename the configuration, see also #1416.

davidscn:
Upvote, also the header renaming. Note that this is only somewhat breaking, we can always add a redirect from old names to updated names in order to enforce compatibility.

uekerman:
Thinking of #655 I could make more sense to call the object `Participant`.

BenjaminRodenberg:
We should merge https://github.com/precice/precice/pull/1366 before starting to work on this, because otherwise rebasing #1366 will be a nightmare. Then we can also directly break it. I would not worry about redirections, because this introduces additional code that just exists for maintaining legacy.

On the actual issue: I agree. I don't like `SolverInterface`, as well. Be aware that there is already a `precice::impl::Participant`. In some of our tutorials we simply call it [`interface`](https://github.com/precice/tutorials/blob/eaf55b17ae88f8b37c20ae8791467f16ed1e2f0e/elastic-tube-1d/solid-cpp/src/SolidSolver.cpp#L28) or [`Adapter`](https://github.com/precice/tutorials/blob/eaf55b17ae88f8b37c20ae8791467f16ed1e2f0e/partitioned-heat-conduction/fenics/heat.py#L103).

[...]

PR #1643, opened 2023-05-08
Rename SolverInterface class

MakisH:
## Main changes of this PR

- [x] Renames the class `SolverInterface` to `Participant` in the C++ API
- [x] Renames the files `SolverInterface.xpp` to `Participant.xpp`
- [x] Renames `SolverInterfaceImpl` to `ParticipantImpl`
- [x] Renames the files `SolverInterfaceImpl.xpp` to `ParticipantImpl.xpp`
- [x] Applies the same changes in the C bindings (file: `ParticipantC.h`)
- [x] Applies the same changes in the Fortran bindings (file: `ParticipantFortran.hpp`)
- ~~Renames `SolverInterfaceConfiguration` to ?? + files~~ -> suggestion: skip for now
- ~~Reorder `sources.cmake`~~ -> already not strictly ordered
- Does not yet introduce the central `precice.hpp` file -> All user-facing changes should end up there
- Does not yet modify the XML configuration and keeps the `SolverInterfaceConfiguration` class for now

As discussed in #1458, we want to rename the main API class from `SolverInterface` to `Participant`, as the name is more descriptive.

## Author's checklist

* I added a changelog file with `make changelog` if there are user-observable changes since the last release. -> User-facing changes from this PR will be overwritten when introducing a global `precice.hpp`.
* For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html). -> Should happen when introducing a central `precice.hpp` and renaming the bindings files.

## Reviewers' checklist


MakisH:
What would a good name be for `SolverInterfaceConfiguration`? `ParticipantConfiguration` is already taken, and covers the XML tag `<participant>`.

I would prefer to not change everything in a giant PR, so I would like again (similarly to the closed #1627) to not yet remove the `<solver-interface>` tag and do that separately, to keep things easier to review.

I suggest keeping it as-is in this PR, remove when dropping the `<solver-interface>` tag.

fsimonis (approved):
> I suggest keeping it as-is in this PR, remove when dropping the <solver-interface> tag.

Definitely a good idea. It will disappear anyhow.

Only thing that could also be done in a different PR is to rename the variable used for the Participant in the solverdummies. They currently use `interface`.

I already find the interface way easier to read! :+1:

MakisH:
@fsimonis I merged `develop`. Could you please check for any renaming mentions of solver interface that are not stemming from the XML configuration?

MakisH:
Follow-up regarding the central header file: https://github.com/precice/precice/pull/1643

kanishkbh:
I noticed the copydoc lines still say `SolverInterface::<foo>`, which can now be changed to `Participant::<foo>`

MakisH:
> I noticed the copydoc lines still say `SolverInterface::<foo>`, which can now be changed to `Participant::<foo>`

Thanks! I thought I changed it, but it probably got lost in one of the conflict resolutions. I will update it.

PR #1649, opened 2023-05-10
Replace `BaseCouplingScheme::_time` with `BaseCouplingScheme::_timeWindowStartTime`

BenjaminRodenberg:
## Main changes of this PR

* Replace _time with _timeWindowStartTime for easier tracking of time.

Makes tracking of time a bit simpler, because we do not need to roll back the `_time`, if an iteration did not converge, but instead we increase the `_timeWindowStartTime`, if the iteration converged. At all places, where `_time` is used, we now use `getTime()`.

Important: Similar to https://github.com/precice/precice/pull/1109 and might come with similar risks.

## Author's checklist

## Reviewers' checklist


uekerman (approved):
Looks correct.

PR #1654, opened 2023-05-11
Add a wrapper header file precice.hpp

MakisH:
## Main changes of this PR

- Adds a very minimal `precice.hpp`, which only includes headers (`Participant.hpp`, `Version.h`, `span.hpp`).
- Renames the bindings files to `precicec.h` and `precicefortran.hpp`.

This probably has many side-effects. I have not checked the packaging or any downstream components.

To discuss:

- [x] Do we still want to include the additional headers (version, span) in the `Participant` files?

As discussed in https://github.com/precice/precice/issues/1458, we want to have a central header file `precice.hpp` to include the API and any auxiliary headers. The name is also more intuitive than `SolverInterface.hpp` or `Participant.hpp`.

Depends on https://github.com/precice/precice/pull/1643.

## Author's checklist

## Reviewers' checklist


src/precice/precice.hpp
```diff
@@ -0,0 +1,5 @@
+#pragma once
+
+#include <precice/Participant.hpp>
+#include <precice/Version.h>
+#include <precice/span.hpp>
```
fsimonis, line 5: Isn't this primarily an implementation detail of `Participant.hpp` similar to `export.hpp` and `types.hpp`?
If so, then this doesn't really belong here.

MakisH: But then, would the `precice.hpp` just be a redirection/alias for `Participant.hpp`?

MakisH: I thought we wanted to expose the `Version.h` and `span.hpp` to the user, to be able to use, e.g., the preCICE span directly in their code.

We can achieve this with either location, but I thought that putting them directly into the `precice.hpp` makes the intention clearer.

fsimonis: Key point here is that the user may have to create a `precice::span` in the adapter.
So, let's keep it in the `precice.hpp`.

MakisH: And same for `Version.h`, right? So, everything right, conversation resolved, and PR approved?

fsimonis (approved):

Issue #1466, opened 2022-11-13
Location of dimensions option in the configuration format


MakisH:
Split from https://github.com/precice/precice/issues/1416.

**Please describe the problem you are trying to solve.**

preCICE configuration files currently start with:

```xml
<solver-interface dimensions="..." />
```

In the context of the Geometric Multiscale feature, we want to be able to define different dimensions for each participant. Even further, we may want to define different dimensions for each coupling scheme. Defining a global dimensions size for all participants would lead to some duplication or additional input checks.

**Describe the solution you propose.**

I am not sure about a concrete solution yet.

Specifically about the dimensions, I think we could move it to each participant:

```xml
<participant name="{string}" dimensions={integer}">
```

If needed, we could then derive the largest number of dimensions automatically.

MakisH:
> > Even further, we may want to define different dimensions for each coupling scheme.
> 
> This part, I don't understand yet. Could you elaborate, why coupling schemes may need a dimension?

Imagine that we are coupling 1D flow with a 2D solid, and the 2D solid with a 3D flow. In principle, we could use a 2D interface for the first one and a 3D interface for the latter one.

But the "coupling scheme" term was probably a bit unfortunate here.

carme-hp:
I think would rather have the dimensionality specific to the mesh than to the participant, since for my muscle case, a participant should be able to send meshes of different dimensionality to different participants.

My (simplified) case  is the following:
I have a solid mechanics muscle solver, a solid mechanics tendon solver and a neuronal-feedback muscle solver. 
The solid mechanics solvers exchange 3D meshes between each other. However between the solid mechanics muscle solver and the neuronal-feedback muscle solver a 1D mesh is exchanged.

MakisH:
@carme-hp could you please sketch roughly how you would imagine an ideal configuration for your case? Documenting your setup drawing here would be helpful.

Is the mesh you exchange with the neuronal-feedback participant 1D (one or more scalars) or 0D (one scalar, assuming it is an ODE solver)? Of course, the terminology here is a bit broken.

I understand it is not enough to have it in the participant level, because one of your participants is used in couplings with both a 3D and a 1D other participants. But it still remains a 3D participant that could directly benefit from the geometric multiscale mapping, having preCICE automatically convert from the 3D world to the 1D world, right?

carme-hp:
Hi, sorry for my extremely late reply.

Here is my case setup:

[...]

PR #1662, opened 2023-05-16
Remove XML tag solver-interface

### LINKED ISSUE #1416: Confusing solver-interface XML tag and impact on geometric multiscale

Currently, every preCICE configuration file starts with:

```xml
<solver-interface dimensions="..." />
```

[Everything, apart from the logger, is wrapped by this tag](https://precice.org/configuration-xml-reference.html). The reason seems to be an implementation detail, as this is the configuration of the `SolverInterface` class.

However, I see two main issues with this tag:

- **Issue 1: Confusing name "solver-interface".**
    As a "solver interface", I would imagine the (physical or code) interface between two participants. This is already ambiguous, as a new user could think: "is it the part of the mesh that belongs to the interface? is it the code coupling pair?". Besides, all coupling schemes are defined inside this tag, which could confuse people in terms of singular vs plurar. But overall, this is always awkward to explain to people: "just use this in your config".
- **Issue 2: Global dimensions.**
  Moved to https://github.com/precice/precice/issues/1466
  This issue was originally more about the dimensions from my side, but the discussion focused on the more fundamental issue of the tag itself.

We could consider alternative names for the tag.

Removing the tag altogether and bringing everything to the top level would mean that we need to find a new place not only for the `dimensions` attribute, but also for the `experimental` flag. It would also make an extension to support more `SolverInterface` objects in a participant more complicated (see https://github.com/precice/precice/issues/655).

MakisH:
## Main changes of this PR

Closes #1416.

Our XML configuration files currently look like this:

```xml
<precice-configuration>
  <log>
    <sink
      type="stream"
      output="stdout"
      filter="%Severity% > debug"
      format="preCICE:%ColorizedSeverity% %Message%"
      enabled="true" />
  </log>

  <solver-interface dimensions="3">
    <data:vector name="Data-One" />
    <data:vector name="Data-Two" />
    ...
  </solver-interface>
</precice-configuration>
```

This PR changes the configuration reader, all XML files, and related error messages to this form:

```xml
<precice-configuration dimensions="3">
  <log>
    <sink
      type="stream"
      output="stdout"
      filter="%Severity% > debug"
      format="preCICE:%ColorizedSeverity% %Message%"
      enabled="true" />
  </log>

  <data:vector name="Data-One" />
  <data:vector name="Data-Two" />
  ...
</precice-configuration>
```

Notes:

1. The class `SolverInterfaceConfiguration` is deleted and all functionality is merged into the `Configuration` class.
2. All attributes previously in `solver-interface` are moved to the `precice-configuration`. The `dimensions` will later be moved elsewhere (see #1466).
3. The integration tests do not complain, but I don't know if anything else changed in their behavior/output (@uekerman)

Related to https://github.com/precice/precice/issues/1458, where we decided to drop the `<solver-interface>`.

## Author's checklist

## Reviewers' checklist

* [ ] Have a look over the complete `Configuration` class and the changes in the `ParticipantImpl`. Anything strange?
* [ ] Anything to consider regarding the integration tests? Is it expected that it just works without the split between `Configuration` and `SolverInterfaceConfiguration`?

It probably makes sense to hide whitespace changes when reviewing. And yes, I have used the config formatter via the pre-commit hook, I am not sure why it added so many blank lines. Maybe because of the indentation level change?


fsimonis (approved):
Changes look as expected :+1:

PR #1664, opened 2023-05-18
Always perform send/recv for both `WINDOW_START` and `WINDOW_END`

BenjaminRodenberg:
## Main changes of this PR

Removes special receive for initialization and always performs communication for both samples at `WINDOW_START` and `WINDOW_END`.

This is a preparatory step for time interpolation and subcycling, where the whole storage will be serialized and communicated.

~~**Important**: Runtime increases significantly! See tests.~~ (fixed: see https://github.com/precice/precice/pull/1664#issuecomment-1556007767)

## Author's checklist

* [x] Merge ~~#1520~~ #1639
* [x] Merge https://github.com/precice/precice/pull/1684

## Reviewers' checklist


BenjaminRodenberg:
With data serialization there is no significant performance degradation anymore:

comparing `precice.integration.Serial` for `ubuntu-2004 g++ MPI Debug` on our CI pipeline:

* 66.21s for d3485cd614f29f82e54269028be8a737f9325954 (current version on develop), see https://github.com/precice/precice/actions/runs/5032242045/jobs/9025778054#step:13:46
* 105.94s for ecda717 (no serialization), see https://github.com/precice/precice/actions/runs/5032899888/jobs/9026791679#step:13:46
* 66.78s for 8c069fe (uses serialization), see https://github.com/precice/precice/actions/runs/5034070628/jobs/9028669568?pr=1664#step:13:46

src/cplscheme/CouplingData.hpp
```diff
@@ -91,6 +91,41 @@ class CouplingData {
...
+   * Serialization of the data is performed per mesh node. The dimension of one node is then data dimension (scalar or vector) * number of time steps.
+   *
+   * @return Eigen::VectorXd a vector containing all data for all time steps in serialized fashion.
+   */
+  Eigen::VectorXd getSerializedValues();
+
+  /**
+   * @brief Returns the gradients of all time steps stored in this coupling data in a serialized fashion
+   *
+   * Serialization of the data is performed per mesh node. The dimension of one node is then data dimension (scalar or vector) * number of time steps.
```
uekerman, line 106:
```suggestion
   * Serialization of the data is performed per mesh node. The dimension of one node is then data dimension (scalar or vector) * space dimension * number of time steps.
```
?

src/cplscheme/CouplingData.hpp
```diff
@@ -91,6 +91,41 @@ class CouplingData {
...
+  void storeFromSerialized(Eigen::VectorXd timesAscending, Eigen::VectorXd serializedValues);
+
+  /**
+   * @brief accepts serialized values and gradients and stores them in this coupling data
+   *
+   * @param timesAscending times associated with data
+   * @param serializedValues values in serialized form
+   * @param serializedGradients gradients in serialized form
+   */
+  void storeFromSerialized(Eigen::VectorXd timesAscending, Eigen::VectorXd serializedValues, Eigen::MatrixXd serializedGradients);
```
uekerman, line 127: It's a bit odd that the "store" methods are not symmetric to the "get" methods.

Instead of `getSerializedX()`, there could also be one `serialize()`.

Compare:

https://github.com/precice/precice/blob/d3485cd614f29f82e54269028be8a737f9325954/src/com/SerializedMesh.hpp#L21

src/cplscheme/CouplingData.cpp
```diff
@@ -139,6 +139,98 @@ void CouplingData::moveToNextWindow()
...
+  int  timeId   = 0;
+  PRECICE_ASSERT(atBeginn.timestamp == time::Storage::WINDOW_START, atBeginn.timestamp);
+  auto sliceBeginn = atBeginn.sample.values;
+  for (int valueId = 0; valueId < nValues; valueId++) {
+    serializedData(valueId * nTimeSteps) = sliceBeginn(valueId);
+  }
+
+  auto atEnd = timeStepsStorage().stamples().back();
+  timeId     = 1;
+  // PRECICE_ASSERT(atEnd.timestamp == time::Storage::WINDOW_END, atEnd.timestamp); // triggered during exchangeInitialData
```
uekerman, line 158: What do we do with this one?

uekerman (changes requested):
Very clear design :+1: 

I did not look carefully through the implementation details of `CouplingData`, but I think we need a good unit test coverage here.

Only doubt: If I understood it correctly, it is now no longer possible to couple the "old way", i.e. not communicate `WINDOW_START` if waveform relaxation is switched off (order==0?).
I think this would still be good for debugging (my coupled simulation fails with v3.0, is this due to the waveform relaxation?) and for performance comparison for your thesis. Re-enabling this, could also be part of the next PR, where you anyway have to tinker with the number of stamples to communicate.
Using the integration tests for a first performance indication is a good idea, but they certainly do not show the complete picture. For real cases, communication will now be twice as expensive.

With some PR, we will need a changelog entry. Maybe not yet here.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -104,35 +104,44 @@ void BaseCouplingScheme::sendData(const m2n::PtrM2N &m2n, const DataMap &sendDat
   for (const auto &data : sendData | boost::adaptors::map_values) {
     const auto stamples = data->stamples();
     PRECICE_ASSERT(stamples.size() > 0);
-    data->sample() = stamples.back().sample;
+
+    auto serializedValues = data->getSerializedValues();
```
fsimonis, line 108: This will copy a bunch of data. You could simplify this by creating a `std::vector<double> serializedValues` outside the loop.

Then you pass this as `std::vector<double>&` to:
```suggestion
    data->serializeValuesTo(serializedValues);
```

Inside you would call `.resize()`.
This should save you allocations.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -104,35 +104,44 @@ void BaseCouplingScheme::sendData(const m2n::PtrM2N &m2n, const DataMap &sendDat
...
+    int  nTimeSteps       = 2;
 
     // Data is actually only send if size>0, which is checked in the derived classes implementation
-    m2n->send(data->values(), data->getMeshID(), data->getDimensions());
+    m2n->send(serializedValues, data->getMeshID(), data->getDimensions() * nTimeSteps);
 
     if (data->hasGradient()) {
-      m2n->send(data->gradients(), data->getMeshID(), data->getDimensions() * data->meshDimensions());
+      auto serializedGradients = data->getSerializedGradients();
```
fsimonis, line 115: Same logic as above applies.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -104,35 +104,44 @@ void BaseCouplingScheme::sendData(const m2n::PtrM2N &m2n, const DataMap &sendDat
...
-void BaseCouplingScheme::receiveData(const m2n::PtrM2N &m2n, const DataMap &receiveData, bool initialCommunication)
+void BaseCouplingScheme::receiveData(const m2n::PtrM2N &m2n, const DataMap &receiveData)
 {
   PRECICE_TRACE();
   PRECICE_ASSERT(m2n.get());
   PRECICE_ASSERT(m2n->isConnected());
   for (const auto &data : receiveData | boost::adaptors::map_values) {
+
+    int  nTimeSteps     = 2;
+    auto timesAscending = Eigen::VectorXd(nTimeSteps);
```
fsimonis, line 129: Why not?
```suggestion
    Eigen::VectorXd timesAscending(nTimeSteps);
```

src/cplscheme/CouplingData.cpp
BenjaminRodenberg: Fixed. c185890655af6a6a18e03c527112830d89114b15

BenjaminRodenberg:
> Only doubt: If I understood it correctly, it is now no longer possible to couple the "old way", i.e. not communicate `WINDOW_START` if waveform relaxation is switched off (order==0?).
> I think this would still be good for debugging (my coupled simulation fails with v3.0, is this due to the waveform relaxation?) and for performance comparison for your thesis. Re-enabling this, could also be part of the next PR, where you anyway have to tinker with the number of stamples to communicate.

Took care of this part in [46f76d9](https://github.com/precice/precice/pull/1664/commits/46f76d9c41ae2414ec052f8486a87bbbe1e772b6). The corresponding toggle is the experimental tag in the configuration. We could also pick something else here, but we always have to hand it down from the config all the way to send/recv. I decided against the waveform order, because this is only defined on the read data, but the sender and receiver need to know which communication strategy to pick.

src/cplscheme/CouplingData.hpp
BenjaminRodenberg: I think the best solution is to use the same pattern as `SerializedMesh` for `SerializedStamples` (and add this class in `com`) and then do something like the following inside `BaseCouplingScheme::sendData`

```
const auto serializedStamples = com::SerializedStamples::serialize(data->stamples());

// Data is actually only send if size>0, which is checked in the derived classes implementation
m2n->send(serializedStamples.values(), data->getMeshID(), data->getDimensions() * serializedStamples.nTimeSteps);

if (data->hasGradient()) {
  m2n->send(serializedStamples.gradients(), data->getMeshID(), data->getDimensions() * data->meshDimensions() * serializedStamples.nTimeSteps());
}
```

and `BaseCouplingScheme::receiveData`:
 
```
int             nTimeSteps = 2;
Eigen::VectorXd timesAscending(nTimeSteps);
timesAscending << time::Storage::WINDOW_START, time::Storage::WINDOW_END;
com::SerializedStamples = serializedStamples();

serializedStamples.allocateValues(nTimeSteps * data->getSize());
m2n->receive(serializedStamples.values(), data->getMeshID(), data->getDimensions() * nTimeSteps);

if (data->hasGradient()) {
  serializedStamples.allocateGradients(nTimeSteps * data->getSize() * data->meshDimensions());
  m2n->receive(serializedStamples.gradients(), data->getMeshID(), data->getDimensions() * data->meshDimensions() * nTimeSteps);
}

data->stamples() = serializedStamples.toStamples();
```

@uekerman does this look reasonable?

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: See comment above. I'll follow the pattern provided by `SerializedMesh` etc. This should also have the same effect.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: See above.

src/cplscheme/CouplingData.hpp
BenjaminRodenberg: Implementation of this part is done.

uekerman (approved):
For test coverage, it would probably be good to have an integration test that uses `substeps="false"`.
And we are missing a few usability checks: waveform configuration vs exchange configuration.
Otherwise, good to go.

BenjaminRodenberg:
> For test coverage, it would probably be good to have an integration test that uses `substeps="false"`.

Sure. I think it might be tricky to come up with a case that already shows a difference between `substeps="false"` and `substeps="true"` now that subcycling is not implemented. If this is not possible now, I might move this test to #1671. Here, it should definitely work.

> And we are missing a few usability checks: waveform configuration vs exchange configuration.

Just to make sure. I guess we want to check the following:

1) If waveform order is zero, then substeps must be set to false. [^1]
2) If waveform order is first, then substeps may be set to true (allows piecewise linear interpolation) or false
3) If waveform order is second or greater, substeps must be set to true (higher order is only reachable, if substeps are exchange)

How do we want to deal with a false configuration? Warning or error? I would suggest an error, because I don't see a reason for permitting the false configurations. If we discover a use-case, we can still relax and use a warning instead. The other way around would be breaking.

[^1]: This eliminates subcycling with piecewise constant interpolation, but this is not a useful feature anyways, because piecewise linear interpolation comes at the same cost with better performance.

BenjaminRodenberg:
> * If waveform order is zero, then substeps must be set to false. [1](#user-content-fn-1-6927fd3cf0a25b18e3cf55ed88da1bc7)
> 
> * If waveform order is first, then substeps may be set to true (allows piecewise linear interpolation) or false
> 
> * If waveform order is second or greater, substeps must be set to true (higher order is only reachable, if substeps are exchange)

I realized the following when working on the point mentioned above: Should we already now set `waveform-order = "1"` to the new default and make the whole waveform API non-experimental? Main reason: If we now require "If waveform order is zero, then substeps must be set to false", our current default setting `waveform-order="0"` and `substeps="true"` would lead to a lot of warnings or errors in our tests. We could fix this by updating the tests, but we would later just change them again.

Suggestion: Let's use `waveform-order="1"` as the new default. I can open a corresponding PR.

PR #1671, opened 2023-05-27
WIP: write data when subcycling is used

BenjaminRodenberg:
## Main changes of this PR

Allows to write data to preCICE when subcycling is used. This either results in a piecewise constant or linear interpolation. Possibly also in higher-order interpolation.

see #1171, this is an updated version of #1171 and #1523 implementing the changes introduced (especially) via #1639 and #1612.

## Author's checklist

## Reviewers' checklist


BenjaminRodenberg:
Some interesting findings on this PR: When running the experiments from https://github.com/BenjaminRodenberg/oscillator-example/tree/main/precice-3 with this PR and #1664 there is a clear performance difference. Just some preliminary experiments on my notebook, but for this PR we are in the order of magnitude of three minutes, while for #1664 we only need a few seconds.

BenjaminRodenberg:
I did a performance comparison of #1671 (this PR at c05c3c63206ecc820912d46016c1c4c10fa7da7a), #1664 (at c2b83f7b3252a30400a70ef950868a037032023a), and the current state of develop (at 49ba127e4ada06a1cbb2d81d2ac71aa7845ab665). As a basis, I used https://github.com/BenjaminRodenberg/oscillator-example/commit/05382282b52c7776585ce294fca9dcd605767fad and ran `oscillator-example/precice-3$ python3 mass-spring.py Mass-Right & python3 mass-spring.py Mass-Left`.

For develop and #1671 the experiment runs quickly, while for #1664 there is a clear performance degradation. Numerical results are all identical. There is also consistent behavior for some other tests that I ran, but I think the case above is the simplest one. Next step: Try to find the reason for the performance degradation of #1671.

src/precice/impl/WriteDataContext.cpp
```diff
@@ -31,7 +31,7 @@ void WriteDataContext::resetData(bool atEndOfWindow, bool isTimeWindowComplete)
 
   if (isTimeWindowComplete && hasWriteMapping()) { // manually overwrite value at beginning with value from end. Need this exception for WriteDataContext with write mapping, because CouplingScheme is not able to update _providedData, if write mapping sits between _providedData and _toData. CouplingScheme in this case only has access to _toData.
     PRECICE_ASSERT(atEndOfWindow, "isTimeWindowComplete without atEndOfWindow is forbidden!");
-    auto atEnd = _providedData->timeStepsStorage().stamples().back().sample;
+    const auto &atEnd = _providedData->timeStepsStorage().stamples().back().sample;
```
davidscn, line 34: I did not read through the whole code, but considering that you are investigating performance, [this section about Eigen and `auto`](https://eigen.tuxfamily.org/dox/TopicPitfalls.html#title3) might be of interest.

BenjaminRodenberg: Thank you for the hint! That's very good to know, but looks to me that the problem is somewhere else. I changed the `auto` to `Eigen::VectorXd` at several places without a observable difference in runtime. However, we actually use `auto` at many places where we access `Data::values()` (e.g. `Acceleration::applyRelaxation`).

BenjaminRodenberg:
I found a simple way to reproduce the performance degradation. See [b8b1779](https://github.com/BenjaminRodenberg/precice/commit/b8b17793899b7ea9c525f9e53278127aa393b11a). Here, I branched off from the latest version of #1664 and just added some communication that is similar to the communication I introduced in #1671, but without serialization and only exchanging 2 samples for any window.

BenjaminRodenberg:
> I found a simple way to reproduce the performance degradation. See [b8b1779](https://github.com/BenjaminRodenberg/precice/commit/b8b17793899b7ea9c525f9e53278127aa393b11a). Here, I branched off from the latest version of #1664 and just added some communication that is similar to the communication I introduced in #1671, but without serialization and only exchanging 2 samples for any window.

@uekerman and I decided to make this an issue for the future (https://github.com/precice/precice/issues/1679). Main reason: We expect that the performance degradation is less relevant for realistic cases.

BenjaminRodenberg:
I removed [9cdcbd2](https://github.com/precice/precice/pull/1696/commits/9cdcbd227e5cbf237254995653e112188f81c17d) and [b818e43](https://github.com/precice/precice/pull/1696/commits/b818e43e2e8b27c5edce5a268c712bf173b5b10e) from here. The goal of this PR is to provide a basic implementation for higher-order waveforms + subcycling. For a simple fixed-point iteration this also works nicely. For constant underrelaxation and more advanced acceleration schemes there is still a lot of work to do. This will be done in follow-up work packages (like #1696).

src/com/SerializedStamples.hpp
```diff
@@ -86,13 +86,17 @@ class SerializedStamples {
    */
   void serializeValues(const cplscheme::PtrCouplingData data);
 
+  void serializeValuesInitialization(const cplscheme::PtrCouplingData data);
```
uekerman, line 89: Could you add here and below a quick comment explaining the difference between those methods?

src/m2n/M2N.cpp
```diff
@@ -330,6 +346,28 @@ void M2N::receive(double &itemToReceive)
   PRECICE_DEBUG("receive(double): {}", itemToReceive);
 }
 
+void M2N::receive(precice::span<double> itemsToReceive)
+{
+  PRECICE_TRACE(utils::IntraComm::getRank());
+  if (not utils::IntraComm::isSecondary()) { //coupling mode
```
uekerman, line 352:
```suggestion
  if (not utils::IntraComm::isSecondary()) {
```
"coupling mode" as a term should be rather obsolete, I hope

uekerman (approved):
Looks good. I did not look in detail through the integration tests.

src/m2n/M2N.cpp
BenjaminRodenberg: Yes. Copy-pasted this from somewhere else. I cleaned this up via e4c706d341bff0d5156e7918fba4d53f5afccfbf.

src/com/SerializedStamples.hpp
BenjaminRodenberg: Done. [9fb3bd7](https://github.com/precice/precice/pull/1671/commits/9fb3bd77800d65057d303b788d2100f467b10e19)

Issue #1596, opened 2023-03-07
Extend test coverage of waveform iteration to adaptive timestep sizes


NiklasKotarsky:
WIP

**Please describe the problem you are trying to solve.**
A clear and concise description of what the problem is.

**Describe the solution you propose.**

**Describe alternatives you've considered**

**Additional context**

PR #1628, opened 2023-04-24
Fix timestep vs. time step.

BenjaminRodenberg:
## Main changes of this PR

* Replace "timestep" with "timeStep" in code
* Replace "timestep" with "time step" in documentation
* Replace "timestep length" with "time step size" in documentation
* Replace "timestepLength" with "timeStepSize" in code

## Motivation and additional information

Make code and documentation more consistent. Especially w.r.t #1623 

## Author's checklist

* [x] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [x] I added a changelog file with `make changelog` if there are user-observable changes since the last release. (not necessary)
* [x] I added a test to cover the proposed changes in our test suite.
* [x] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).
* [x] I sticked to C++17 features.
* [x] I sticked to CMake version 3.16.3.
* [x] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

PR #1681, opened 2023-06-19
Add test for time adaptivity and waveform iteration

NiklasKotarsky:
## Main changes of this PR

Add test for waveform iterations and under relaxation.

Mentioned in issue https://github.com/precice/precice/issues/1596. This is in preparation for the full waveform iterations. 

Adds the following tests
Write and read data with non constant timesteps
Write and read data with varying number of timesteps
Check if acceleration is applied to every timestep for relaxation

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/impl/ParticipantState.hpp
+++ b/src/precice/impl/ParticipantState.hpp
@@ -187,17 +187,6 @@ public:
     return _readDataContexts | boost::adaptors::map_values;
   }
 
-  /** @brief Determines and returns the maximum order of all read waveforms of this participant
-   */
-  int maxReadWaveformOrder() const
-  {
-    int maxOrder = -1;
-    for (const auto &context : _readDataContexts | boost::adaptors::map_values) {
-      maxOrder = std::max(maxOrder, context.getInterpolationOrder());
-    }
-    return maxOrder;
-  }
-
   /// Is the dataID know to preCICE?
   bool hasData(std::string_view mesh, std::string_view data) const;
 


BenjaminRodenberg (changes requested):
I added a few comments regarding the updated API calls. I think you test were not actually build. But as soon as the compiler tries to build the tests, it should probably give you useful error messages, if you are using the outdated API. You can also use other tests as a template.

Also useful: You can use the following command to run an individual test locally:

```
precice/build$ mpirun -np 4 ./testprecice -t Integration/Serial/Time/Explicit/ParallelCoupling/ReadWriteScalarDataWithSubcycling
```

The part behind the `-t` defines the test (or you can also run a complete suite).

BenjaminRodenberg (changes requested):
Thanks for the updates!

Another important piece of tooling: Please install and run the pre-commit hooks. See https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit. This helps you to take care about formatting issues. Then the test "Code Style" should also pass.

NiklasKotarsky:
> Thanks for the updates!
> 
> Another important piece of tooling: Please install and run the pre-commit hooks. See https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit. This helps you to take care about formatting issues. Then the test "Code Style" should also pass.

Thanks for pointing it out. It appears that my pre-commit hook was out of date.

tests/serial/time/implicit/serial-coupling/WaveformSubsyclingWithConstantAcceleration.cpp
BenjaminRodenberg, line 1: Please fix the spelling in the name of this file.

tests/serial/time/implicit/serial-coupling/WaveformSubsyclingWithConstantAcceleration.cpp
```diff
@@ -0,0 +1,135 @@
...
+
+  if (precice.requiresInitialData()) {
+    writeData = writeFunction(time);
+    precice.writeData(meshName, writeDataName, {&vertexID, 1}, {&writeData, 1});
+  }
+
+  precice.initialize();
+  double maxDt     = precice.getMaxTimeStepSize();
+  double windowDt  = maxDt;
+  double dt        = windowDt / nSubsteps; // Timestep length desired by solver. E.g. 7 steps  with size 1/7
```
BenjaminRodenberg, line 78:
```suggestion
  double dt        = windowDt / nSubsteps; // Timestep length desired by solver. E.g. 7 steps with size 1/7
```

tests/serial/time/implicit/serial-coupling/WaveformSubsyclingWithConstantAcceleration.cpp
```diff
@@ -0,0 +1,135 @@
...
+    readFunction  = dataOneFunction;
+  }
+
+  double   writeData = 0;
+  double   readData  = 0;
+  double   v0[]      = {0, 0, 0};
+  VertexID vertexID;
+  vertexID = precice.setMeshVertex(meshName, v0);
+
+  int    nSubsteps          = 7; // perform subcycling on solvers. 4 steps happen in each window.
```
BenjaminRodenberg, line 62:
```suggestion
  int    nSubsteps          = 7; // perform subcycling on solvers. 7 steps happen in each window.
```
Is there a specific reason for using 7 steps? Should 2 steps not already suffice?

tests/serial/time/implicit/serial-coupling/WaveformSubsyclingWithConstantAcceleration.cpp
```diff
@@ -0,0 +1,135 @@
...
+  int    nWindows           = 5; // perform 5 windows.
```
BenjaminRodenberg, line 63: Why not just one window? Should be sufficient for reproducing errors in the acceleration scheme.

tests/serial/time/implicit/serial-coupling/WaveformSubsyclingWithConstantAcceleration.cpp
```diff
@@ -0,0 +1,135 @@
...
+    } else { // in the following iterations, use data at the end of window.
+      BOOST_TEST(readData == readFunction(time + maxDt));
+    }
+
+    precice.readData(meshName, readDataName, {&vertexID, 1}, currentDt, {&readData, 1});
+    if (context.isNamed("SolverOne")) { // in the first iteration of each window, use data from previous window.
+      if (iterations == 0) {
+        BOOST_TEST(readData == readFunction(timeCheckpoint));
+      } else {
+        BOOST_TEST(readData == factor * readFunction(time + currentDt));
```
BenjaminRodenberg, line 105: If I merge this PR into my PR #1671 I actually get a failure here. This is good, because it means that your test is sensitive. Of course, it it also to some degree bad, because it means that there is a problem with the implementation of acceleration schemes and waveforms at the moment.

tests/serial/time/implicit/serial-coupling/WaveformSubsyclingWithConstantAcceleration.cpp
```diff
@@ -0,0 +1,135 @@
...
+using namespace precice;
+
+BOOST_AUTO_TEST_SUITE(Integration)
+BOOST_AUTO_TEST_SUITE(Serial)
+BOOST_AUTO_TEST_SUITE(Time)
+BOOST_AUTO_TEST_SUITE(Implicit)
+BOOST_AUTO_TEST_SUITE(SerialCoupling)
+
+/**
+ * @brief Simple test to ensure that underrelaxation is applied to every  timestep.
```
BenjaminRodenberg, line 17:
```suggestion
 * @brief Simple test to ensure that underrelaxation is applied to every time step.
```
Please write `time step` (in text) and `timeStep` (in code). We agreed on this at some point in time to make things consistent. See https://github.com/precice/precice/pull/1628 for some more spelling rules.

Similar for `time window` (in text) and `timeWindow` (in code).

tests/serial/time/implicit/parallel-coupling/ReadWriteScalarDataWithWaveformSubcyclingNonConstantTimeSteps.cpp
```diff
@@ -0,0 +1,123 @@
...
+    precice.writeData(meshName, writeDataName, {&vertexID, 1}, {&writeData, 1});
+  }
+
+  precice.initialize();
+  double              maxDt = precice.getMaxTimeStepSize();
+  std::vector<double> dt_v; // the different dt used
+  if (context.isNamed("SolverOne")) {
+    dt_v = {maxDt / 5, 2 * maxDt / 5, 2 * maxDt / 5, 0}; // initialize the different timesteps and one extra timestep
+  } else {
+    dt_v = {maxDt / 6, 2 * maxDt / 6, maxDt / 6, 2 * maxDt / 6, 0}; // initialize the different timesteps and one extra timestep
```
BenjaminRodenberg, line 72:
```suggestion
    dt_v = {maxDt / 5, 2 * maxDt / 5, 2 * maxDt / 5, 0}; // initialize the different time steps and one extra time step
  } else {
    dt_v = {maxDt / 6, 2 * maxDt / 6, maxDt / 6, 2 * maxDt / 6, 0}; // initialize the different time steps and one extra time step
```
Why do you need this extra time step?

tests/serial/time/implicit/parallel-coupling/ReadWriteScalarDataWithWaveformSubcyclingVaryingTimeGrids.cpp
```diff
@@ -0,0 +1,114 @@
...
+    precice.writeData(meshName, writeDataName, {&vertexID, 1}, {&writeData, 1});
+  }
+
+  precice.initialize();
+  double time_window_size = precice.getMaxTimeStepSize();
+  double maxDt            = time_window_size;
+  double dt               = maxDt;
+  double timeCheckpoint;
+  int    iterations;
+  int    nbr_timestep = 0;
```
BenjaminRodenberg, line 73:
```suggestion
  int    nbrTimeStep = 0;
```
Is `nbr` number? Then why not `n` like `nSubsteps` in other tests.

tests/serial/time/implicit/parallel-coupling/ReadWriteScalarDataWithWaveformSubcyclingVaryingTimeGrids.cpp
```diff
@@ -0,0 +1,114 @@
...
+  double time = 0;
+
+  if (precice.requiresInitialData()) {
+    writeData = writeFunction(time);
+    precice.writeData(meshName, writeDataName, {&vertexID, 1}, {&writeData, 1});
+  }
+
+  precice.initialize();
+  double time_window_size = precice.getMaxTimeStepSize();
+  double maxDt            = time_window_size;
```
BenjaminRodenberg, line 69:
```suggestion
  double timeWindowSize = precice.getMaxTimeStepSize();
  double maxDt            = timeWindowSize;
```
Please use `camelCase`, not `snake_case`. @fsimonis Interesting that our linter does not fix this...

tests/serial/time/implicit/parallel-coupling/ReadWriteScalarDataWithWaveformSubcyclingVaryingTimeGrids.cpp
BenjaminRodenberg: I actually don't really understand why this variable is needed. To me it looks like you could just remove it.

tests/serial/time/implicit/parallel-coupling/ReadWriteScalarDataWithWaveformSubcyclingVaryingTimeGrids.cpp
BenjaminRodenberg, line 1: I'm not really sure what this test is doing. Is something missing here? I don't really see the "varying time grids" part here.

BenjaminRodenberg (changes requested):
I currently see two independent sets of tests:

1) `WaveformSubcyclingWithConstantAcceleration` tests for a bug in the acceleration if waveforms are used.
2) The other two tests are target towards the broader topic of time adaptivity. I think I'm missing a bit context here. Why do we need these tests? Is there a feature that you have in mind?

My suggestion: Let's split the two groups above into independent PRs, because things are a bit mixed up:

1) `WaveformSubcyclingWithConstantAcceleration` we should be able to merge soon, because it helps us to fix a bug in #1671. I'm not sure whether we should also add tests for other acceleration schemes, but this is something @uekerman should know.
2) For the other two tests we should create another PR with a more specific title and description (for me to understand things better and also for documentation reasons).

Btw: Did you already open an issue for the topic of time adaptivity? I think this would be useful to have an overview and provide some general description of where you want to go. I'm currently missing the following points: Are there actually any features missing for time adaptivity? Which important parts that you need are untested? You could also link to this issue from the PR with the two "time adaptivity tests".

uekerman:
> I'm not sure whether we should also add tests for other acceleration schemes, but this is something @uekerman should know.

Eventually, we want tests for every acceleration method, of course. But that's beyond this PR in my opinion.

PR #1684, opened 2023-06-20
Make waveform non-experimental and first order the default

BenjaminRodenberg:
## Main changes of this PR

* Removes requirement for setting `experimental="true"`, if waveform interpolation is used.
* Changes default waveform order to first order.

This is our intended default behavior for version 3.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/1684.md
+++ b/docs/changelog/1684.md
@@ -1,0 +1,1 @@
+* Use `waveform-order="1"` as default. Makes time interpolation non-experimental, therefore no need to use `experimental="true"`. Note on explicit schemes: Depending on the scheme for some participants `waveform-order="1"` technically leads to a zeroth order interpolation.
--- a/src/cplscheme/config/CouplingSchemeConfiguration.cpp
+++ b/src/cplscheme/config/CouplingSchemeConfiguration.cpp
@@ -301,31 +301,8 @@ void CouplingSchemeConfiguration::xmlEndTagCallback(
   PRECICE_TRACE(tag.getFullName());
   if (tag.getNamespace() == TAG) {
     if (_config.type == VALUE_SERIAL_EXPLICIT) {
-
-      //Check the waveform order of both participants in the explicit coupling
-      const auto first  = _config.participants[0];
-      const auto second = _config.participants[1];
-
-      auto first_participant = _participantConfig->getParticipant(first);
-      for (const auto &dataContext : first_participant->readDataContexts()) {
-        const int usedOrder = dataContext.getInterpolationOrder();
-        // The first participants waveform order has to be 1 for serial explicit coupling
-        int allowedOrder = 1;
-        if (usedOrder > allowedOrder) {
-          PRECICE_ERROR(
-              "You configured <read-data name=\"{}\" mesh=\"{}\" waveform-order=\"{}\" />, but for the serial explicit coupling scheme only a waveform-order of \"{}\" is allowed for the first participant.",
-              dataContext.getDataName(), dataContext.getMeshName(), usedOrder, allowedOrder);
-        }
-      }
-      auto second_participant = _participantConfig->getParticipant(second);
-      for (const auto &dataContext : second_participant->readDataContexts()) {
-        const int usedOrder = dataContext.getInterpolationOrder();
-        if (usedOrder < 0) {
-          PRECICE_ERROR(
-              "You configured <read-data name=\"{}\" mesh=\"{}\" waveform-order=\"{}\" />, but for the serial explicit coupling scheme the waveform-order must be non-negative for the second participant.",
-              dataContext.getDataName(), dataContext.getMeshName(), usedOrder);
-        }
-      }
+      int maxAllowedOrder = 1; // explicit coupling schemes do not allow higher-order waveform iteration
+      checkWaveformOrderReadData(maxAllowedOrder);
 
       std::string       accessor(_config.participants[0]);
       PtrCouplingScheme scheme = createSerialExplicitCouplingScheme(accessor);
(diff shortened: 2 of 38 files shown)


src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -307,33 +301,8 @@ void CouplingSchemeConfiguration::xmlEndTagCallback(
...
-          const int usedOrder = dataContext.getInterpolationOrder();
-          if (usedOrder < 0) {
-            PRECICE_ERROR(
-                "You configured <read-data name=\"{}\" mesh=\"{}\" waveform-order=\"{}\" />, but for the serial explicit coupling scheme the waveform-order must be non-negative for the second participant.",
-                dataContext.getDataName(), dataContext.getMeshName(), usedOrder);
-          }
-        }
-      }
+      int maxAllowedOrder = 1; // explicit coupling schemes do not allow higher-order waveform iteration
+      checkWaveformOrderReadData(maxAllowedOrder);
```
BenjaminRodenberg, line 305: We don't really need these fine-grained checks anymore. For most explicit schemes, `waveform-order="1"` will be used automatically, but it will result in a zeroth order interpolation, because there is no data available for `WINDOW_END`, but only for `WINDOW_START` when we enter the window. Exception: The second participant in a serial-explicit scheme. Here, a first order interpolant is created. `tests/serial/time/explicit/serial-coupling/ReadWriteScalarDataWithWaveform.xml` corresponds to this setup.

BenjaminRodenberg (commented):
Ready for review.

docs/changelog/1684.md
```diff
@@ -0,0 +1 @@
+* Use `waveform-order="1"` as default. Makes time interpolation non-experimental, therefore no need to use `experimental="true"`. Note on explicit schemes: Depending on the scheme for some participants `waveform-order="1"` technically leads to a zeroth order interpolation.
```
uekerman, line 1:
```suggestion
* Changed `waveform-order="1"` to be default and time interpolation to non-experimental (no need to use `experimental="true"` anymore). Note on explicit schemes: Depending on the scheme for some participants `waveform-order="1"` technically leads to a zeroth order interpolation.
```
Explicit schemes probably needs some revisiting. Maybe, we need some error messages there.
Not for this PR, but for v3: Why not cubic as default?

src/cplscheme/config/CouplingSchemeConfiguration.cpp
uekerman: We could still print warnings for parallel-explicit and first participant in serial-explicit, right?

uekerman (approved):

docs/changelog/1684.md
BenjaminRodenberg: > Explicit schemes probably needs some revisiting.

The current situation is not ideal, but it is also not easy to make the connection from `read-data` to the scheme being used and then raise the right error. Any idea for a situation where we want to raise an error? Generally, it is not wrong to ask for higher-order when using an explicit scheme. The user might just not get anything with higher order, because there is just no higher-order data available.

>  Not for this PR, but for v3: Why not cubic as default?

For cubic interpolation we need to do at least 3 substeps per window. Is see two competing arguments here:

* From the user perspective it might be surprising, if we say "the default is third order" and then third order is not reached for the setting that most users will probably run (no subcycling).
* Using a higher-order interpolation by default is the "smartest" option, because it would automatically use a higher-order interpolation, if possible. Otherwise, it would use a lower-order interpolation. Using third order is an arbitrary, but common option (see, e.g. [interpolation in scipy](https://docs.scipy.org/doc/scipy/reference/generated/scipy.interpolate.make_interp_spline.html#scipy.interpolate.make_interp_spline))

Main usability question here is: simple and easy to understand behavior of preCICE vs. a smart strategy that might need some more explanation. I tend towards using third order as default, because this is also what I think most libraries often do, if there is different algorithms are available ([scipy.optimize.minimize](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html#scipy.optimize.minimize) applies some "magic" for picking the best `method`, if the user does not explicitly ask for a specific method).

src/cplscheme/config/CouplingSchemeConfiguration.cpp
BenjaminRodenberg: If we use third order as default, we would always warn about the default in this case. This looks odd to me.

docs/changelog/1684.md
uekerman: I see the doubts with cubic as default. Let's stick with "1" for the moment. That's the safe bet.

uekerman: Explicit coupling: don't you have everything needed in the code below?
I would do warnings, not errors. "You picked order 3, but we can only do order 0".

uekerman: > But if our default is order 1, then we would also have to raise a warning for our default value. 

Could still prevent users form wrong choices, no?

BenjaminRodenberg: I created https://github.com/precice/precice/issues/1690 to conclude the discussion about the default order for this PR. Third order is not available yet, so let's use 1st order for now.

W.r.t error messages I will also develop a solution based on 1st order interpolation as default.

BenjaminRodenberg: I decided to remove the usability checks for explicit coupling schemes. Main reason: If we also think about compositional coupling schemes, then we might actually run into contradictions, if we trigger a warning based on the `waveform-order` provided to `read-data`. And: Even for an explicit coupling scheme it is not wrong to ask for a higher waveform order (you might just not get it) @uekerman: We can also continue the discussion in an issue, if you want.

Issue #1645, opened 2023-05-08
Reduce number of references to Data::values()


BenjaminRodenberg:
Currently all packages of preCICE reference `Data::values()` to read or write data. Mapping, acceleration, communication and actions operate on the data in `Data::values()`. For time interpolation we recently introduced a write data buffer and a storage (#1612) in `Data`. In #1612 data has to be moved from the storage to `Data::values()` and vice-versa when mapping, acceleration, communication or actions are executed to not interfere with the currently existing implementation. In the future, we should, however avoid using `Data::values()` and instead use the data from the storage (see example below).

The following packages need a review:

- [x] mapping (see https://github.com/precice/precice/pull/1660, https://github.com/precice/precice/issues/1651)
- [x] communication (see #1639)
- [ ] acceleration
- [ ] actions
- [ ] exporter / io

Note that this issue is mainly meant as an overview.

### Example for mapping

Current state (#1612)

https://github.com/precice/precice/blob/f272d0ad35c69375ac95881e383ba40dac5f7df4/src/precice/impl/DataContext.cpp#L88-L102

`map` accesses `Data::values()` using `fromDataID` and `toDataID`.

Proposed implementation:

```cpp
 for (auto &stample : context.fromData->getStamples()) { 
   const Eigen::VectorXd fromData = stample.sample.values;
   const Eigen::VectorXd toData = context.mapping->map(fromData); 
   // Store data from mapping buffer in storage 
   context.toData->setSampleAtTime(stample.timestamp, toData);
 } 
```

`map` does not need to access `Data::values()`, because it would just operate on `Eigen::VectorXd`.

**Benefits:**

* No need for moving data back and forth
* Better encapsulation of packages

fsimonis:
I had another shot at this as this is making the samples in the remeshing unnecessarily difficult to follow.

Actions are in a Limbo. The actions work, but they will be applied multiple times to some samples. Good example is the initial sample of the time-window, which will be handled every iteration.

Acceleration is easy to port 1-1 by only accelerating the last sample in the time storage. There are a lot of tests that need changing though.

Exporters are relatively trivial to change as they can also only use the last sample.

fsimonis:
The implementation of the tests are a massive burden. After the migration of internals, there are still around 1200 occurrences to deal with.


Uses of "values()" as of now:

| File | Matches |
--- | ---
| src/acceleration/test/AccelerationIntraCommTest.cpp | 222 |
[...]

PR #1696, opened 2023-06-27
Acceleration for Waveforms

BenjaminRodenberg:
## Main changes of this PR

WIP: Currently provides a draft implementation.

Similar to mapping and actions, we need also a proper implementation for acceleration that supports waveforms (see also https://github.com/precice/precice/issues/1645).

Some todos:

- [ ] Plan this for 3.0.0 or 3.x.x ? Technically the changes are not breaking. However, performance of Quasi-Newton acceleration will currently suffer if waveform relaxation is used. We still need to do some testing here.
- [ ] Need to take care of previous iterations in an appropriate way (already deal with time grid changes from one iteration to the next here?)
- [ ] Implement tests for all acceleration schemes
- [ ] Which variants of Quasi-Newton schemes do we want to implement for waveform relaxation?

## Author's checklist

* [ ] Merge #1671 

## Reviewers' checklist


src/time/Storage.cpp
```diff
@@ -124,6 +139,73 @@ time::Sample Storage::computeExtrapolation()
...
+  const int usedOrder = computeUsedOrder(_interpolationOrder, nTimes());
+
+  PRECICE_ASSERT(math::equals(this->maxStoredNormalizedDt(), time::Storage::WINDOW_END), this->maxStoredNormalizedDt()); // sampling is only allowed, if a window is complete.
+
+  if (_interpolationOrder == 0) {
+    return this->getGradientsAtOrAfter(normalizedDt);
+  }
+
+  PRECICE_WARN("You specified interpolation degree of {}, but only degree 0 is supported for gradient interpolation"); // @todo implement this like for sampleAt
+  return this->getGradientsAtOrAfter(normalizedDt);
```
BenjaminRodenberg, line 188: Here, we also should offer higher-order interpolation for gradients. We should also add a corresponding unit test, because this feature is rarely used.

src/time/Storage.cpp
```diff
@@ -92,6 +98,15 @@ Eigen::VectorXd Storage::getValuesAtOrAfter(double before) const
   return stample->sample.values;
 }
 
+//@todo merge getValuesAtOrAfter and getGradientsAtOrAfter into getSampleAtOrAfter, then let user draw data from Sample
```
BenjaminRodenberg, line 101: This requires also some changes in the tests, but makes sense, if we don't want to always have to do some specific implementation just for gradients.

BenjaminRodenberg (commented):
I maked some todos. Generally the feature is working now. I used a large portion of code provided by @NiklasKotarsky via #1698 here.

The todos that are left are mainly refactorings that we could also solve independently from this PR.

Most important point: We will have to move the ownership of the `Waveform` from `ReadDataContext` to another place, because it is not sufficient to only allow `ReadDataContext` to perform interpolations. Currently, there is some code duplication in `CouplingData` to solve this problem. But this is not a good solution, because we might end up with inconsistent configuration of interpolation degree etc. I currently have the feeling that we should move from the following configuration:

```
<mesh name="MeshOne">
    <use-data name="DataOne" />
    <use-data name="DataTwo" />
</mesh>
...
  <provide-mesh name="MeshOne" />
  <write-data name="DataOne" mesh="MeshOne" />
  <read-data name="DataTwo" mesh="MeshOne" waveform-order="3" />
</participant>
```

to

```
<mesh name="MeshOne">
    <use-data name="DataOne" />
    <use-data name="DataTwo" waveform-order="3" />
</mesh>
...
  <provide-mesh name="MeshOne" />
  <write-data name="DataOne" mesh="MeshOne" />
  <read-data name="DataTwo" mesh="MeshOne" />
</participant>
```

this moves the `Waveform` from `ReadDataContext` to `mesh::Data`. Any opinion on this @uekerman?

uekerman:
> this moves the Waveform from ReadDataContext to mesh::Data. Any opinion on this @uekerman?

I agree that

```xml
<read-data name="DataTwo" mesh="MeshOne" waveform-order="3" />
```

is no good solution then anymore.

But what about moving this to

```xml
<exchange data="DataOne" mesh="MeshOne" from="SolverOne" to="SolverTwo" waveform-order="3"/>
```

?

This would correspond to `CouplingData`, right?. And then we might also be able to get rid of the `substeps` option.

Only concern I see right now: What happens if we

```xml
<exchange data="DataOne" mesh="MeshOne" from="SolverOne" to="SolverTwo" waveform-order="3"/>
<exchange data="DataOne" mesh="MeshOne" from="SolverOne" to="SolverThree" waveform-order="1"/>
```

but AFAIS, this should be two different `CouplingData`.

Any such solution would move this PR to v3.0.

BenjaminRodenberg:
> Only concern I see right now: What happens if we
> 
> ```
> <exchange data="DataOne" mesh="MeshOne" from="SolverOne" to="SolverTwo" waveform-order="3"/>
> <exchange data="DataOne" mesh="MeshOne" from="SolverOne" to="SolverThree" waveform-order="1"/>
> ```
> 
> but AFAIS, this should be two different `CouplingData`.

Let's continue discussing in https://github.com/precice/precice/issues/1709

BenjaminRodenberg:
I just merged #1714 into this PR. This should help us to resolve some of the SW engineering issues I mentioned above. We will not, however, need to resolve #1714 first before we are able to merge #1696.

src/time/Storage.hpp
```diff
@@ -109,6 +113,16 @@ class Storage {
...
 
+  /**
+   * @brief Need to use interpolation for the case with changing time grids
+   *
+   * @param normalizedDt a double, where we want to sample the waveform
+   * @return Eigen::VectorXd values in this Storage at or directly after "before"
+  */
+  Eigen::VectorXd sampleAt(double normalizedDt); // @todo try to solve this differently. Currently duplicates a lot of code from Waveform::sample. Maybe even move Waveform inside Storage, if every Storage needs to interpolate anyway?
+
+  Eigen::MatrixXd sampleGradientsAt(double normalizedDt); // @todo try to solve this differently. Currently duplicates a lot of code from Waveform::sample. Maybe even move Waveform inside Storage, if every Storage needs to interpolate anyway?
```
BenjaminRodenberg, line 124: todo

src/cplscheme/CouplingData.cpp
```diff
@@ -17,11 +17,15 @@ CouplingData::CouplingData(
...
-      _exchangeSubsteps(exchangeSubsteps)
+      _exchangeSubsteps(exchangeSubsteps),
+      _timeStepsStoragePrevious()
 {
   PRECICE_ASSERT(_data != nullptr);
   timeStepsStorage().setExtrapolationOrder(extrapolationOrder);
+  _timeStepsStoragePrevious.setExtrapolationOrder(extrapolationOrder);
+  _timeStepsStoragePrevious.setInterpolationOrder(3); // @todo hard-coded for now, but we need to somehow link this to <read-data waveform-order="ORDER" />
+  _timeStepsStoragePrevious.setSampleAtTime(time::Storage::WINDOW_START, time::Sample{getDimensions(), Eigen::VectorXd::Zero(getSize())});
+  _timeStepsStoragePrevious.setSampleAtTime(time::Storage::WINDOW_END, time::Sample{getDimensions(), Eigen::VectorXd::Zero(getSize())});
```
BenjaminRodenberg, line 28: todo

BenjaminRodenberg:
I just merged `develop` into this PR. @NiklasKotarsky you might also want to do this, because there are quite a lot of changes. Relevant is especially #1738.

BenjaminRodenberg:
Closing this PR in favor of https://github.com/precice/precice/pull/1794, because it does not depend on #1721.

PR #1697, opened 2023-06-27
Fix DivByZero for scaled constraints

### LINKED ISSUE #1686: Mapping that may result in `nan` values
The following line might introduce a scaling factor `nan`, if `integralOutput.array()` is equal to zero. 

https://github.com/precice/precice/blob/c1f6cb902452c429e4a84c8bd34a61680cd7db0c/src/mapping/Mapping.cpp#L166

This can happen easily, if data has not been initialized (defaults to zero). I ran into this problem while working on #1684. Here the tests from `Integration/Serial/MappingScaledConsistent` do not initialize data. This results in `nan` values, that are propagated to the user, if `waveform-order = "1"` is used.

fsimonis:
## Main changes of this PR

This PR reproduces a test which fails due to div by zero and then fixes the issue.

Closes #1686

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mapping/Mapping.cpp
+++ b/src/mapping/Mapping.cpp
@@ -225,7 +225,19 @@ void Mapping::scaleConsistentMapping(const Eigen::VectorXd &input, Eigen::Vector
   Eigen::Map<Eigen::MatrixXd> outputValuesMatrix(output.data(), valueDimensions, output.size() / valueDimensions);
 
   // Scale in each direction
-  const Eigen::VectorXd scalingFactor = integralInput.binaryExpr(integralOutput, [](double lhs, double rhs) { return (rhs == 0.0) ? 0.0 : (lhs / rhs); });
+  // We cannot handle the case with zero output data and non-zero input data.
+  // To fulfill the constraint, we would need to scale the output data in such a way, that the integral sum of the input is preserved.
+  // That's not possible using a constant scaling factor as the output sum will always be zero. Here, we return 1 and emit a warning afterwards.
+  const Eigen::VectorXd scalingFactor = integralInput.binaryExpr(integralOutput, [](double lhs, double rhs) { return (rhs == 0.0) ? 1.0 : (lhs / rhs); });
+   PRECICE_DEBUG("Scaling factor in scaled-consistent mapping: {}", scalingFactor);
+
+  // check whether the constraint is fulfilled
+  for (Eigen::Index i = 0; i < scalingFactor.size(); ++i) {
+    double consistency = std::abs(scalingFactor[i] * integralOutput[i] - integralInput[i]);
+    if (consistency > 0)
+      PRECICE_WARN("Failed to fulfill consistency constraint for scaled-consistent mapping from mesh \"{}\" to mesh \"{}\". Consistency difference between input and output is \"{}\".", this->input()->getName(), this->output()->getName(), integralInput[i] - integralOutput[i]);
+  }
+
   PRECICE_DEBUG("Scaling factor in scaled-consistent mapping: {}", scalingFactor);
   outputValuesMatrix.array().colwise() *= scalingFactor.array();
 } // namespace mapping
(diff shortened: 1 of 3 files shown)


src/mapping/Mapping.cpp
```diff
@@ -225,7 +225,7 @@ void Mapping::scaleConsistentMapping(const Eigen::VectorXd &input, Eigen::Vector
   Eigen::Map<Eigen::MatrixXd> outputValuesMatrix(output.data(), valueDimensions, output.size() / valueDimensions);
 
   // Scale in each direction
-  Eigen::VectorXd scalingFactor = integralInput.array() / integralOutput.array();
+  const Eigen::VectorXd scalingFactor = integralInput.binaryExpr(integralOutput, [](double lhs, double rhs) { return (rhs == 0.0) ? 0.0 : (lhs / rhs); });
```
davidscn, line 228: I'm a bit in doubt about this: When having no flux, then the solution of the applied mapping here is undefined. Why is zero the desired solution? Shouldn't we abort at this point or at least give a warning? What about 
```suggestion
  const Eigen::VectorXd scalingFactor = integralInput.binaryExpr(integralOutput, [](double lhs, double rhs) { return (rhs == 0.0) ? lhs : (lhs / rhs); });
```

uekerman: `0.0` seems to be the right value if the input is zero everywhere, which is the bug we ran into.
It could, indeed, however also be that the integral is zero, but the values are not.

```suggestion
  const Eigen::VectorXd scalingFactor = integralInput.binaryExpr(integralOutput, [](double lhs, double rhs) { return (rhs == 0.0) ? 1.0 : (lhs / rhs); });
```

??

I don't understand the `lhs` suggestion. It's "only" the scaling factor, right? The values still get multiplied below.

fsimonis:
Open questions on how the scaled consistent mapping should behave:

1. The data is 0.
2. The data has 0 components. Example, all 3D vector data is (1 1 0) or (1 0 0).
3. The scalar data is non-zero, but the output integral is
4. The vectorial data is non-zero, but the output integral is zero in one or more components

davidscn:
> There are still open questions on how the scaled consistent mapping should behave in the situations where:
> 
>     1. The input data is 0.

The problem is more the output data. If all input data is zero, then all output data is zero as well (independent of the applied mapping method).  In this case, all suggestions (0/1/lhs) lead in theory to the same result.

> 
>     2. The input data has 0 components. Example: all 3D vector data is (1 1 0) or (1 0 0).

Not an isolated issue, as the scaling is handled component-wise (just as all mappings). Hence, the case is the same as above. (A bit odd that the scaling factors are `Eigen::VectorXd`, since this data structure is typically used for global vectors.

>     3. The input scalar data is non-zero, but the output integral is 0.
> 

That's the difficult one and a singularity of the constraint in general. To fulfill the constraint, we would need to scale the output data in such a way, that the integral sum of the input is preserved. That's not possible using a constant scaling factor as the output sum will always be zero.
I'm still not sure what's the correct answer here. We should at least warn the user that things (whatever flux) is not preserved in the domain, as it might affect the correctness of the results. Probably Benjamin's suggestion makes most sense of all this 

```cpp
  const Eigen::VectorXd scalingFactor = integralInput.binaryExpr(integralOutput, [](double lhs, double rhs) { return (rhs == 0.0) ? 1.0 : (lhs / rhs); });
  ```

>     4. The input vectorial data is non-zero, but the output integral contains zero components.
> 

Same as 2.

uekerman:
> I'm still not sure what's the correct answer here. We should at least warn the user that things (whatever flux) is not preserved in the domain, as it might affect the correctness of the results.

warning is a good idea here :+1:

fsimonis:
> The problem is more the output data.

Of course :see_no_evil:

To summarize:

For each component:
1. If input data is zero, do nothing as output should already be zero.
2. If input integral is zero, set output to zero as scaling factor is 0.
3. If output integral is zero, emit a warning and do nothing. (damage reduction mode)

Question: Do we need a warning / info statement for point 2?

uekerman:
> 1. If input data is zero, do nothing as output should already be zero.

i would still set it to zero just to be sure

davidscn:
> To summarize:
> 
> For each component:
> 
>     1. If input data is zero, do nothing as output should already be zero.

If input data is zero, we still run into the original issue -> set output integral sum to 1

> 
>     2. If input integral is zero, set output to zero as scaling factor is 0.

If we set the output to zero, we run into the original issue -> set output sum to 1, the 0 sum from the input will do the rest

> 
>     3. If output integral is zero, emit a warning and do nothing. (damage reduction mode)
> 

if do nothing means "set output integral to 1" -> yes

> Question: Do we need a warning / info statement for point 2?

No, this case is not problematic at all.

davidscn:
@fsimonis could you please review and add a changelog-entry? Functionality is good to go, I just cannot push to your fork.

fsimonis:
Changelog still missing. I'll add one before I merge

fsimonis:
Changelog still missing. I'll add one before I merge

PR #1702, opened 2023-06-29
Try GSL for B-Spline interpolation

fsimonis:
## Main changes of this PR

This is a prototype PR to try out the GNU scientific library for B-spline fit.
The PR is meant as a sandbox for @BenjaminRodenberg to compare results to the Eigen impl

Note that this would introduce GSL as an additional dependency (Available everywhere and there is CMake support)

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -188,6 +188,10 @@ find_package(Eigen3 3.3.7 REQUIRED)
 message(STATUS "Found Eigen ${EIGEN3_VERSION}")
 precice_validate_eigen()
 
+# GNU Scientific Library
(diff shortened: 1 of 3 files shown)


BenjaminRodenberg:
@fsimonis I found the bug leading to https://github.com/precice/precice/pull/1704. I already found a fix and it's definitely not a problem with Eigen or the interpolation routine (see https://github.com/precice/precice/pull/1704). Is there still a motivation for trying out the GSL implementation now that we get good results with the Eigen implementation?

fsimonis:
No, let's leave it at Eigen.

Issue #1680, opened 2023-06-18
Allow convergence measurement w.r.t complete time storage, not only value at window end


BenjaminRodenberg:
**Please describe the problem you are trying to solve.**

 The convergence measurement of preCICE only takes the values at the end of the window into consideration. With time interpoloation and waveforms, we need a new convergence measurement that considers the whole waveform (including sub steps) on the current time window. To support this, I a reference to some results below.

**Describe the solution you propose.**

Our current implementation boils down to the following line:

https://github.com/precice/precice/blob/ddc9aea4c99a5420bd7a94f47f8c0b835e0be173/src/cplscheme/BaseCouplingScheme.cpp#L578

Here, `CouplingData::previousIteration()` and `CouplingData::values()` are `Eigen::VectorXd`, i.e. just a single value at the end of the time window. Instead, we have to compare the current and previous `time::Storage` representing the waveform. Important detail: The current and previous iteration do not necessarily live on the same time grids. Therefore, for a general solution, we would need to compare the `time::Storage`s using an appropriate method, such as, a L2 norm applied to the two waveforms.

**Describe alternatives you've considered**

I did not consider alternatives. It would be useful to verify that the proposed solution has the potential for improving the error by performing additional experiments that use the already existing implementation (with only slight modifications) to measure the convergence of substeps.

**Additional context**

I presented results from a convergence study based on #1671 at Coupled conference 2023. See [slides](https://mediatum.ub.tum.de/node?id=1713068). The gap between the monolithic experiment and the partitioned experiment with $\Delta t = 4 \delta t$ in the plot below is especially interesting:

![Screenshot from 2023-06-18 14-55-20](https://github.com/precice/precice/assets/5740604/3c9b803d-04be-46ff-8624-23844fa67660)

My current interpretation here is the following: For the partitioned experiment we do not measure the convergence with respect to all sub steps inside the window, but only at the end of the window. Therefore, we only see $\mathcal{O}(\Delta t ^4)$, but not $\mathcal{O}(\delta t^4)$. The error is (roughly) shifted by a the difference between the time step size and the time window size.

BenjaminRodenberg:
I experimented with a simple implementation assuming that the time grid does not change in [f0bc0a5](https://github.com/BenjaminRodenberg/precice/commit/f0bc0a5c016876ad76cfbb1f9596513be3325c54). I saw convergence with respect to every sub step after a few iterations, but no improvement of the error.

[...]

PR #1704, opened 2023-06-30
Fix bug that prevents requiring checkpoint in last window

BenjaminRodenberg:
## Main changes of this PR

Make sure that `isCouplingOngoing()` does not accidentally evaluate `false` due to wrong update of internal time.

Fix bug. Found it while debugging case from https://github.com/precice/precice/issues/1680.

Original implementation has a bug with checkpointing: When entering the last window, `isCouplingOngoing()` evaluates `false`, if we increase the `_timeWindowStartTime` before checking `isCouplingOngoing()`. This will avoid writing a checkpoint for the beginning of the last window, meaning we jump back to the second to last window when reading the checkpoint.

## Author's checklist

* [x] ~~I added a changelog file with `make changelog` if there are user-observable changes since the last release.~~ (not needed, bug does not exist in version 2.5)
* [x] I added a test to avoid reintroducing the bug.
* [X] I sticked to CMake version 3.16.3.

## Reviewers' checklist

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -348,16 +348,18 @@ void BaseCouplingScheme::secondExchange()
         // coupling iteration.
         PRECICE_ASSERT(math::greater(_computedTimeWindowPart, 0.0));
         _timeWindows -= 1;
-      } else { // write output, prepare for next window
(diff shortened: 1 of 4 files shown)


uekerman (approved):
Looks good. Needs a changelog entry.

PR #1705, opened 2023-06-30
Fix subcycling and mix time-window-sizes in `CompositionalCouplingScheme`

### LINKED ISSUE #1713: Mixed time window sizes in compositional coupling schemes

Mixing different time window sizes in a compositional coupling schemes is allowed, but not completely clear.

Example:
```xml
<coupling-scheme:serial-explicit>
  <participants first="A" second="B" />
  <time-window-size value="2" />
</coupling-scheme:serial-explicit>

<coupling-scheme:serial-explicit>
  <participants first="B" second="C" />
  <time-window-size value="3"  />
</coupling-scheme:serial-explicit>
```

## Maximum time step size

The max time step size of 
* A is the time-window-size 2
* C is the time-window-size 3
* B aligns with A and B:
  1. Meet A: 2 (t=2)
  2. Meet B: 1 (t=3)
  3. Meet A: 1 (t=4)
  4. Meet A and B: 2 (t=6)

Is it reasonable for B to expect this fluctuation in time step sizes?

## Data Mapping

A participant performs data mapping and communication at the end of a time window.
The communication is handled by the coupling schemes, which thus supports mixed time windows.
The data mapping is handled by the participant, resulting in unnecessary mapping executions.
Only the participants synchronized with provide/require data which may need to be mapped.
This gets magnified by https://github.com/precice/precice/pull/1664 as we now map two samples per data.

If mappings for exchanged data is defined on B, then this would result in unnecessary mappings on B:

Step | Meeting who | DT | t | useless map in advance 
--- | --- | --- | --- | ---
1 | A | 2 | 2 | 2 calls
2 | B | 1 | 3 | 2 calls
3 | A | 1 | 4 | 2 calls
4 | A & B | 2 | 6 | 0 call

## Implicit coupling

Assuming:
```xml
<coupling-scheme:serial-explicit>
  <participants first="A" second="B" />
  <time-window-size value="1" />
</coupling-scheme:serial-explicit>

  <participants first="B" second="C" />
  <time-window-size value="2"  />
</coupling-scheme:serial-explicit>
```

Then we get the following get timestep sizes:
* A is the time-window-size 1
* C is the time-window-size 2
* B aligns with partners
  1. 1 as it meets A (at t=1)
  2. 1 as it meets A and B (at t=2)

If the implicit scheme doesn't converge, then the checkpoints are restored and the schemes revert to t=0.
Now the max step size for B and C are 2 and they both meet at t=2.

What data of A do we expect at this point in time? t=0, t=1, or t=2?
How do we sample from this partial data in time? Do we need to keep all samples?

## Prescribed time window

One or multiple serial coupling schemes may also prescribe the time window. 

As an example A and C could prescribe the time window for B.

We need to clarify what use-cases we support, test the supported ones and prevent the unsupported ones.

**Additional context**

### LINKED ISSUE #1730: Data storage can be moved multiple times per time window to the next time window
**Describe your setup**

preCICE Version: develop (> 2.5.0)

**Describe the problem**

At the end of a time window, the data storage is moved.
The end sample becomes the new start sample and the new end sample is extrapolated. Then old samples are discarded.

This is triggered in **each coupling scheme**. If different schemes read the same data, then this is done repeatedly.

Let f be the first and l be the last sample, then 2 coupling schemes using the same data results in the following samples when moving to a new time window:
`[f, … ,l] -> [l, e(f, … ,l)] -> [e(f, … ,l), e(l, e(f, … ,l)]`

This hasn't been noticed, as the extrapolation defaults to 0th order:
`[f, … ,l] -> [l, l] -> [l, l]`

Using 1st order will show this bug.

**Step To Reproduce**
1. Configure a data to be exchanged by 2 coupling schemes
2. Use an implicit coupling scheme with extrapolation order 1

**Expected behaviour**
Each data storage should be moved exactly once.

**Additional context**
#1723 would allow to handle this centrally inside the participant.

BenjaminRodenberg:
## Main changes of this PR

This PR fixes the compositional coupling scheme in the follow ways:
* Subcycling now works with implicit coupling in a compositional scheme
* Mixing different time-window-sizes now works correctly (even though mappings are still duplicate)

The compositional CouplingScheme flat out doesn't support subcycling.

Supersedes #1822
Closes #1713
Closes #1730 as the time storage is now only "moved" for received Data and receiving the same Data from multiple coupling schemes at the same time is nonsense. 

## Author's checklist

## Reviewers' checklist


BenjaminRodenberg:
@fsimonis Can you take a look here? I tried to debug this, but got a bit lost in the `CompositionalCouplingScheme`. Looks to me like the maximum allowed time step size is not computed correctly, if we have to iterate on the implicit scheme. This leads to `reachedEndOfTimeWindow()` returning `true` before actually reaching the end of the window. The data exchange then runs into an assertion, because the Storage is not complete.

fsimonis:
Is it possible that some coupling schemes already enter the next time window?

So this would need to be "schemes to run"? 

https://github.com/precice/precice/blob/develop/src/cplscheme/CompositionalCouplingScheme.cpp#L245

fsimonis:
After an extensive debugging session we found the bug.

The Compositional coupling scheme uses the `hasConverged()` of the implicit scheme to deduce weather all schemes are supposed to run, or only the iterative scheme.

When combining this with subcycling results in the behaviour, that the implicit scheme returns `hasConverged() == false` after the first time **step**. At this point in time it has never even computed a convergence measure. Hence, it only runs the implicit scheme for subsequent time steps, leading to rubbish results.

We thus need to update `CompositionalCouplingScheme::_iterating` only if the end of the time window has been reached.

We could allow calls to `hasConverged()` only at the end of the time window.

BenjaminRodenberg:
> We could allow calls to `hasConverged()` only at the end of the time window.

I think this makes a lot of sense, because asking the coupling scheme for `hasConverged()` before reaching the end of the time window is not a meaningful question and the answer will not be useful in any case. We could alternatively let `hasConverged()` return something like `undefined`, but then we need to move from a `bool` to something more fancy. And currently I don't see a reason why we should ask for `hasConverged()` from inside the window.

BenjaminRodenberg:
@fsimonis just to make sure: The issue here (subcycling with identical window sizes does not work) is independent from #1713. Right?

fsimonis:
Correct, this is due to the fact that we update the active schemes after each step.

This leads to the implicit scheme reporting that it hasn't converged even though it never even evaluated a convergence measure.

BenjaminRodenberg:
I did a quick merge with `develop` and updated the config files.

BenjaminRodenberg:
Note: As soon as #1732 is merged, we will also need to update the config files here.

BenjaminRodenberg:
Problematic tests are `Integration/Serial/Time/Implicit/Compositional`. Rest works. @fsimonis: Are you taking care about this?

fsimonis:
Current state: time stepping and isTimeWindowComple() seem to work correctly now.

Open TODOs:
* [ ] Potentially move some functions from CouplingScheme to BaseCouplingScheme
* [ ] Handle time storage of "write-data" in the ParticipantImpl, not in the CouplingScheme.

src/acceleration/test/AccelerationSerialTest.cpp
```diff
@@ -72,10 +75,8 @@ BOOST_AUTO_TEST_CASE(testIQNIMVJPP)
   forces->values() << 0.2, 0.2, 0.2, 0.2;
   forces->setSampleAtTime(0, forces->sample());
 
-  bool exchangeSubsteps = true;
-
-  cplscheme::PtrCouplingData dpcd(new cplscheme::CouplingData(displacements, dummyMesh, false, exchangeSubsteps));
-  cplscheme::PtrCouplingData fpcd(new cplscheme::CouplingData(forces, dummyMesh, false, exchangeSubsteps));
+  cplscheme::PtrCouplingData dpcd = makeCouplingData(displacements, dummyMesh, false);
+  cplscheme::PtrCouplingData fpcd = makeCouplingData(forces, dummyMesh, false);
```
BenjaminRodenberg, line 79:
```suggestion
  bool exchangeSubsteps = true;
  
  cplscheme::PtrCouplingData dpcd = makeCouplingData(displacements, dummyMesh, exchangeSubsteps);
  cplscheme::PtrCouplingData fpcd = makeCouplingData(forces, dummyMesh, exchangeSubsteps);
```

BenjaminRodenberg: Otherwise you are changing what the test actually does.

src/acceleration/test/AccelerationSerialTest.cpp
```diff
@@ -154,10 +155,8 @@ BOOST_AUTO_TEST_CASE(testVIQNPP)
...
```
BenjaminRodenberg, line 159:
```suggestion
  bool exchangeSubsteps = true;
  
  cplscheme::PtrCouplingData dpcd = makeCouplingData(displacements, dummyMesh, exchangeSubsteps);
  cplscheme::PtrCouplingData fpcd = makeCouplingData(forces, dummyMesh, exchangeSubsteps);
```

src/acceleration/test/AccelerationSerialTest.cpp
```diff
@@ -225,10 +224,8 @@ BOOST_AUTO_TEST_CASE(testConstantUnderrelaxationWithSubsteps)
   forces->values() << 0.2, 0.2, 0.2, 0.2;
   forces->setSampleAtTime(0, forces->sample());
 
-  bool exchangeSubsteps = true;
-
-  cplscheme::PtrCouplingData dpcd = std::make_shared<cplscheme::CouplingData>(displacements, dummyMesh, false, exchangeSubsteps);
-  cplscheme::PtrCouplingData fpcd = std::make_shared<cplscheme::CouplingData>(forces, dummyMesh, false, exchangeSubsteps);
+  cplscheme::PtrCouplingData dpcd = makeCouplingData(displacements, dummyMesh, false);
+  cplscheme::PtrCouplingData fpcd = makeCouplingData(forces, dummyMesh, false);
```
BenjaminRodenberg, line 228:
```suggestion
  bool exchangeSubsteps = true;
  
  cplscheme::PtrCouplingData dpcd = makeCouplingData(displacements, dummyMesh, exchangeSubsteps);
  cplscheme::PtrCouplingData fpcd = makeCouplingData(forces, dummyMesh, exchangeSubsteps);
```

src/acceleration/test/AccelerationSerialTest.cpp
```diff
@@ -302,10 +299,8 @@ BOOST_AUTO_TEST_CASE(testConstantUnderrelaxationWithGradientWithSubsteps)
   forces->gradients().setConstant(-2);
   forces->setSampleAtTime(0.0, forces->sample());
 
-  bool exchangeSubsteps = true;
-
-  cplscheme::PtrCouplingData dpcd = std::make_shared<cplscheme::CouplingData>(displacements, dummyMesh, false, exchangeSubsteps);
-  cplscheme::PtrCouplingData fpcd = std::make_shared<cplscheme::CouplingData>(forces, dummyMesh, false, exchangeSubsteps);
+  cplscheme::PtrCouplingData dpcd = makeCouplingData(displacements, dummyMesh, false);
+  cplscheme::PtrCouplingData fpcd = makeCouplingData(forces, dummyMesh, false);
```
BenjaminRodenberg, line 303:
```suggestion
  bool exchangeSubsteps = true;
  
  cplscheme::PtrCouplingData dpcd = makeCouplingData(displacements, dummyMesh, exchangeSubsteps);
  cplscheme::PtrCouplingData fpcd = makeCouplingData(forces, dummyMesh, exchangeSubsteps);
```

src/acceleration/test/AccelerationSerialTest.cpp
```diff
@@ -396,10 +391,8 @@ BOOST_AUTO_TEST_CASE(testConstantUnderrelaxationWithoutSubsteps)
   forces->values() << 0.2, 0.2, 0.2, 0.2;
   forces->setSampleAtTime(0.0, forces->sample());
 
-  bool exchangeSubsteps = false;
-
-  cplscheme::PtrCouplingData dpcd = std::make_shared<cplscheme::CouplingData>(displacements, dummyMesh, false, exchangeSubsteps);
-  cplscheme::PtrCouplingData fpcd = std::make_shared<cplscheme::CouplingData>(forces, dummyMesh, false, exchangeSubsteps);
+  cplscheme::PtrCouplingData dpcd = makeCouplingData(displacements, dummyMesh, false);
+  cplscheme::PtrCouplingData fpcd = makeCouplingData(forces, dummyMesh, false);
```
BenjaminRodenberg, line 395:
```suggestion
  bool exchangeSubsteps = false;
  
  cplscheme::PtrCouplingData dpcd = makeCouplingData(displacements, dummyMesh, exchangeSubsteps);
  cplscheme::PtrCouplingData fpcd = makeCouplingData(forces, dummyMesh, exchangeSubsteps);
```

src/acceleration/test/AccelerationSerialTest.cpp
```diff
@@ -473,10 +466,8 @@ BOOST_AUTO_TEST_CASE(testConstantUnderrelaxationWithGradientWithoutSubsteps)
   forces->gradients().setConstant(-2);
   forces->setSampleAtTime(0, forces->sample());
 
-  bool exchangeSubsteps = false;
-
-  cplscheme::PtrCouplingData dpcd = std::make_shared<cplscheme::CouplingData>(displacements, dummyMesh, false, exchangeSubsteps);
-  cplscheme::PtrCouplingData fpcd = std::make_shared<cplscheme::CouplingData>(forces, dummyMesh, false, exchangeSubsteps);
+  cplscheme::PtrCouplingData dpcd = makeCouplingData(displacements, dummyMesh, false);
+  cplscheme::PtrCouplingData fpcd = makeCouplingData(forces, dummyMesh, false);
```
BenjaminRodenberg, line 470:
```suggestion
  bool exchangeSubsteps = false;
  
  cplscheme::PtrCouplingData dpcd = makeCouplingData(displacements, dummyMesh, exchangeSubsteps);
  cplscheme::PtrCouplingData fpcd = makeCouplingData(forces, dummyMesh, exchangeSubsteps);
```

src/com/tests/helper.hpp
```diff
@@ -0,0 +1,15 @@
...
+#include <memory>
+
+#include "cplscheme/CouplingData.hpp"
+#include "cplscheme/SharedPointer.hpp"
+#include "mesh/SharedPointer.hpp"
+
+namespace precice::testing {
+inline cplscheme::PtrCouplingData makeCouplingData(mesh::PtrData data, mesh::PtrMesh mesh, bool exchangeSubsteps)
+{
+  bool requiresInitialization = false;
```
BenjaminRodenberg, line 12:
```suggestion
inline cplscheme::PtrCouplingData makeCouplingData(mesh::PtrData data, mesh::PtrMesh mesh)
{
  bool requiresInitialization = false;
  bool exchangeSubsteps = true;
```

BenjaminRodenberg: Why not? I think for the serialization tests we will not use `exchangeSubsteps = false` anytime.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -213,12 +213,12 @@ void BaseCouplingScheme::initializeWithZeroInitialData(const DataMap &receiveDat
...
 }
 
-PtrCouplingData BaseCouplingScheme::addCouplingData(const mesh::PtrData &data, mesh::PtrMesh mesh, bool requiresInitialization, bool communicateSubsteps)
+PtrCouplingData BaseCouplingScheme::addCouplingData(const mesh::PtrData &data, mesh::PtrMesh mesh, bool requiresInitialization, bool communicateSubsteps, CouplingData::Direction direction)
 {
   int             id = data->getID();
   PtrCouplingData ptrCplData;
   if (!utils::contained(id, _allData)) { // data is not used by this coupling scheme yet, create new CouplingData
-    ptrCplData = std::make_shared<CouplingData>(data, std::move(mesh), requiresInitialization, communicateSubsteps);
+    ptrCplData = std::make_shared<CouplingData>(data, std::move(mesh), requiresInitialization, communicateSubsteps, direction);
```
BenjaminRodenberg, line 221: I think we should here at least have some kind of assertion. Imagine there is already a `CouplingData` with `id = 3` in `_allData` for the `Direction::Send`. Then `if (!utils::contained(id, _allData)) ` would prevent adding a duplicate of this `CouplingData` with `id = 3` (good), even if the data is now in `Direction::Receive` (bad).

BenjaminRodenberg:
> I am in doubt that the current way of cleaning only received coupling data is the way to go.
> 
> All problems are pointing in the direction that the ParticipantImpl should handle the cleaning. Fundamental part of that is #1822.
> 
> Only exception to the above is cleaning the storage when receiving samples.

Meaning we should prioritize #1822 for v3.0.0 ? Currently this issue is scheduled for v3.0.0 and #1822 for v3.x.x.

fsimonis:
I think that we need #1882 as another piece of the cpl puzzle.

fsimonis:
I finally got things working, and I also added a bunch of tests to ensure that mixed time window sizes lead to the correct stepping behaviour.

src/acceleration/test/AccelerationSerialTest.cpp
fsimonis: This one is supposed to stay `false`

fsimonis:
While the direction of coupling data isn't used for its initial purpose here. It is now used to prevent sending and receiving the same data at the same time in the same coupling scheme.

PR #1715, opened 2023-07-06
Add check for NaN values in setSampleAtTime

kanishkbh:
## Main changes of this PR

Add check for NaN values in setSampleAtTime

This helps to find bugs in initialization of buffer (observed during global data buffer creation).
Thanks @BenjaminRodenberg 

## Author's checklist

## Reviewers' checklist


kanishkbh:
> @kanishkbh I slightly changed the assertion. I think it makes more sense to check the data that we actually want to store and not the data that is already there. Can you briefly check whether this assertion also gets triggered in your setup where we originally found the problem? Please briefly comment here and if the assertion still triggers, then we can merge.

Agree.
And yes the assertion does get triggered. We can merge

Issue #1710, opened 2023-07-05
Require `substeps="false"`, if using any acceleration scheme


BenjaminRodenberg:
**Please describe the problem you are trying to solve.**

Integration of subcycling with acceleration schemes is ongoing work. Currently, acceleration is only integrated in a preliminary fashion here:

https://github.com/precice/precice/blob/394d6483effeee43c4d23cff709fafe7106ca566/src/cplscheme/BaseCouplingScheme.cpp#L798-L816

With the current implementation behavior is not clear, if acceleration is used in combination with subcycling and we need further testing. A first approach into this direction is provided in #1696 (see this PR for details).

**Describe the solution you propose.**

It will take some time until all acceleration schemes are integrated with subcycling and properly tested. Until then we could disallow using incompatible acceleration schemes by requiring `substeps="false"`, if an acceleration scheme is used. This would deactivate subcycling and, therefore, lead to the behavior from preCICE v2. We can then also one-by-one remove this restriction when we provide a proper implementation for acceleration schemes.

**Describe alternatives you've considered**

Just use the "dirty" implementation for acceleration that we currently have. I did a few experiments with my mass-spring example case and did not observe any problems, if using subcycling with acceleration. I will soon provide an overview in the comments below. Still: The unit tests from #1696 fail for the current implementation and we should opt for the proper implementation provided in #1696 and follow-up PRs.

**Additional context**

The whole topic of acceleration + subcycling is under development. Related: #1689

BenjaminRodenberg:
A brief update on my experiments here:

* For the mass-spring example the "dirty" solution overall works ok. I assume that the main reason for this is that a plain fixed-point iteration - which is what we actually apply on all substeps except the last one - already converges nicely. So this test case is not really sensitive.
* For the partitioned heat equation I actually see a clear difference between the current version of develop ([f1adf55](https://github.com/precice/precice/commit/f1adf55c6a34bb7b6e87e7691f69785cf0558741), not converging) and #1696 (converging with overrelaxation factor 0.5 after approx. 18/100 iterations).

Conclusion: Let's only support `substeps="true"` for acceleration schemes, where we have a "clean" implementation. I will take care of implementing this safeguard.

Another note: Currently, there seems to be a bug with `substeps="false"` and data initialization. I first need to figure this out and fix it to have a proper baseline.

BenjaminRodenberg:
[...]

PR #1549, opened 2023-01-19
Add global/meshless data

kanishkbh:
Adds functionality to exchange data that is not associated with any mesh. So far the name is "global-data".

## Motivation and additional information

Closes #202  

## Main changes of this PR

- API
  - Provide methods to read/write global data.
  - Initialize and advance global data.

- Configuration
  - Allow reading the xml tag `global-data`
  - Allow participants to configure read/write global data without specifying any mesh name.
  - Allow coupling to configure exchanges for global data without specifying any mesh name.
  - ~~Allow implicit coupling to configure convergence measures for global data without specifying any mesh name.~~ (Open TODO for a future PR)

- Data
  - ~~Add a `GlobalData` class (meshless equivalent of `Data`).~~
  - `DataConfiguration`:
    -  Configure the `global-data` tag.
    - Add a vector `_globalData` in `DataConfiguration` to store global data here as `Data` objects.

- Participants
  - Add a `GlobalRead/WriteDataContexts` class (meshless and mappingless equivalents of `Read/WriteDataContext`) which inherit from `DataContext`.
  - `ParticipantState` and `ParticipantConfiguration`: Configure and store global read/write contexts.

- Coupling
  - ~Add a `GlobalCouplingData` class, analogous to `CouplingData`for mesh data.~ Add an `isGlobal` attribute to `CouplingData` and use it for global coupling as well.
  - `CouplingSchemeConfiguration`: Add a struct to store global exchanges and functions for global data (analogous to mesh-associated data).

- Com
  - `M2N`: always use intracom in case of global data.

## Open Questions and TODOs:
Documented in https://github.com/precice/precice/issues/1716

## Author's checklist

* [x] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [x] I added a test to cover the proposed changes in our test suite.
* [ ] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).
* [x] I sticked to C++17 features.
* [x] I sticked to CMake version 3.16.3.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

PR #1719, opened 2023-07-12
Fix initialization for `substeps="false"`

BenjaminRodenberg:
## Main changes of this PR

Bugfix, if initializing data and `substeps="false"`. Ensures that data at `WINDOW_START` and `WINDOW_END` is set during initialization. Previously, only data for `WINDOW_END` was set.

Bug detected for partitioned heat equation with `waveform-degree="1"` and `substeps="false"`. Mentioned in #1710.

## Author's checklist

* [x] ~~I added a changelog file with `make changelog` if there are user-observable changes since the last release.~~ Not needed. Buggy state was not released.
* [x] ~~For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).~~ Not needed.

## Reviewers' checklist

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -110,7 +110,7 @@ void BaseCouplingScheme::sendTimes(const m2n::PtrM2N &m2n, const Eigen::VectorXd
   m2n->send(times);
 }
 
-void BaseCouplingScheme::sendData(const m2n::PtrM2N &m2n, const DataMap &sendData, bool initialDataExchange)
+void BaseCouplingScheme::sendData(const m2n::PtrM2N &m2n, const DataMap &sendData, bool initialCommunication)
 {
   PRECICE_TRACE();
   PRECICE_ASSERT(m2n.get() != nullptr);
@@ -148,7 +148,7 @@ void BaseCouplingScheme::sendData(const m2n::PtrM2N &m2n, const DataMap &sendDat
         m2n->send(serialized.gradients(), data->getMeshID(), data->getDimensions() * data->meshDimensions() * serialized.nTimeSteps());
       }
     } else {
-      if (initialDataExchange) { // send data for WINDOW_START. Will be used for WINDOW_START and WINDOW_END.
+      if (initialCommunication) { // send data for WINDOW_START. Will be used for WINDOW_START and WINDOW_END.
         PRECICE_ASSERT(math::equals(stamples.front().timestamp, time::Storage::WINDOW_START));
         PRECICE_ASSERT(math::equals(stamples.back().timestamp, time::Storage::WINDOW_START));
       } else { // send data for WINDOW_END.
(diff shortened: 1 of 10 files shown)


BenjaminRodenberg:
@uekerman there are some merge conflicts with https://github.com/precice/precice/pull/1549, but I'm already in the process of merging both branches to test whether it works.

src/cplscheme/ParallelCouplingScheme.cpp
```diff
@@ -25,26 +25,28 @@ ParallelCouplingScheme::ParallelCouplingScheme(
 
 void ParallelCouplingScheme::exchangeInitialData()
 {
+  bool initialDataExchange = true;
```
BenjaminRodenberg, line 28: We had a similar `bool` previously, but I think we need it again, if we want to properly support `substeps="false"`. Upside: We are closer to the v2 behavior and might also be able to remove some unnecessary communication of data at `WINDOW_START` for `substeps="true"`, if we have such a `bool` (Would suggest to take care of this outside of this PR).

src/cplscheme/SerialCouplingScheme.cpp
```diff
@@ -90,21 +90,29 @@ void SerialCouplingScheme::receiveAndSetTimeWindowSize()
...
     if (receivesInitializedData()) {
-      receiveData(getM2N(), getReceiveData());
+      receiveData(getM2N(), getReceiveData(), initialDataExchange);
       checkDataHasBeenReceived();
     } else {
       initializeWithZeroInitialData(getReceiveData());
     }
+    if (sendsInitializedData()) { // this send/recv pair is only needed, if no substeps are exchanged.
+      sendData(getM2N(), getSendData(), initialDataExchange);
+    }
```
BenjaminRodenberg, line 105: I don't like this additional send/recv pair a lot, but I think it is the easiest way to fix the bug for `substeps="false"`. Comes with some overhead for `substeps="true"` during initialization, but as mentioned above: We might also benefit from this, if we generally don't communicate data for `WINDOW_START` if `initialDataExchange = false`.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -148,6 +148,12 @@ void BaseCouplingScheme::sendData(const m2n::PtrM2N &m2n, const DataMap &sendDat
         m2n->send(serialized.gradients(), data->getMeshID(), data->getDimensions() * data->meshDimensions() * serialized.nTimeSteps());
       }
     } else {
+      if (initialCommunication) { // send data for WINDOW_START. Will be used for WINDOW_START and WINDOW_END.
+        PRECICE_ASSERT(math::equals(stamples.front().timestamp, time::Storage::WINDOW_START));
+        PRECICE_ASSERT(math::equals(stamples.back().timestamp, time::Storage::WINDOW_START));
+      } else { // send data for WINDOW_END.
+        PRECICE_ASSERT(math::equals(stamples.back().timestamp, time::Storage::WINDOW_END));
+      }
```
uekerman, line 156: Currently, `initialCommunication` only triggers some assertions in `sendData`. Looks suspicious, no?
Or better remove the argument?

BenjaminRodenberg: I think we can keep it. The assertions are generally a good safeguard in my opinion and we might add more functionality later, if we don't want to send/receive values for `WINDOW_START` to save some computational cost.

Issue #1690, opened 2023-06-23
Use `waveform-degree="1"` or `waveform-degree="3"` as default?


BenjaminRodenberg:
With https://github.com/precice/precice/pull/1684 using waveforms became non-experimental and the default. First degree interpolation is at the moment the default we agreed on. However, there is also the possibility to use a higher degree as default as soon as #1671 is merged. We should conclude on this point until the release of version 3, because changing the default would be breaking.

I will provide some information on the two candidates below:

### Use `waveform-degree="1"` as default

1st degree interpolation is generally always possible, because we only need two samples to construct a 1st degree interpolant. Using data from the beginning and the end of the window is always available. If additional data is available due to subcycling, a piecewise linear interpolation would be used instead.

Note that a piecewise linear interpolation produces the same results like a linear interpolation as long as we work on the same time grid.

From the user perspective this would lead to a behavior of preCICE that is easy to understand: No matter how many samples I provide in window, there is always a linear interpolation between two samples.

### Use `waveform-degree="3"` as default

3rd degree interpolation is only possible, if 4 or more samples are available in a time window. Meaning, we need subcycling to actually perform 3rd degree interpolation. If less than 4 samples exist, our interpolation routine will automatically pick a lower-degree interpolation; if no subcycling is used this leads to 1st degree interpolation.

If we compare results from runs with and without subcycling there might be a difference -- even if time grids agree -- due to the different interpolation routines and smoothness of the interpolant.

From the user perspective the behavior of preCICE becomes a bit harder to understand: The user has to understand that subcycling and higher-order interpolation go hand in hand. However, if a user uses subcycling, this would automatically lead to a higher-order interpolation routine (likely to also lead to better results) without the need for any further user interaction.

Note: Looking at other libraries for scientific computing might also help us here: In scipy the function [scipy.optimize.minimize](https://docs.scipy.org/doc/scipy/reference/generated/scipy.optimize.minimize.html#scipy.optimize.minimize) also applies some "magic" for picking the best `method`, if the user does not explicitly ask for a specific method. This corresponds more or less to using 3rd degree interpolation as default, because this would automatically fallback to 1st or 2nd degree (different `method`) in certain situations.

### Explicit coupling schemes

[...]

PR #1721, opened 2023-07-13
Add error, if `substeps="true"` for unsupported acceleration scheme.

BenjaminRodenberg:
## Main changes of this PR

Raise error, if `substeps="true"` for unsupported acceleration scheme. Closing #1710.

See #1710.

## Author's checklist

* [x] ~~I added a changelog file with `make changelog` if there are user-observable changes since the last release.~~ (Not needed, feature was not released yet)
* [x] ~~For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).~~

## Reviewers' checklist


src/acceleration/AitkenAcceleration.cpp
```diff
@@ -52,6 +53,12 @@ void AitkenAcceleration::performAcceleration(
 {
   PRECICE_TRACE();
 
+  for (const auto &data : cplData | boost::adaptors::map_values) {
+    if (data->exchangeSubsteps()) {
+      PRECICE_ERROR("Acceleration scheme does not support subcycling. Please pick a different acceleration scheme or set substeps=\"false\" in the exchange tag of data \"{}\".", data->getDataName());
```
uekerman, line 58: Isn't subcycling here the wrong term?
How I would use the term, subcycling is supported, but waveform iteration is not.
Maybe we could use sth like

> Aitken acceleration does not yet support using data from all substeps.

"Pick a different acceleration scheme" could also be dangerous. User will start testing all methods. I can already imagine the forum post. I would just drop this sentence. 

Why not triggering the error in intialize already?

src/acceleration/test/AccelerationIntraCommTest.cpp
```diff
@@ -67,6 +67,8 @@ BOOST_AUTO_TEST_CASE(testVIQNILSpp)
   mesh::PtrData displacements(new mesh::Data("dvalues", -1, 1));
   mesh::PtrData forces(new mesh::Data("fvalues", -1, 1));
 
+  bool exchangeSubsteps = false; // @todo set "true" as soon as acceleration scheme supports subcycling
```
uekerman, line 70: Would be good to have test for both then.

uekerman (changes requested):
Do we now plan to make `substeps` default to `true` or `false` with v3.0?

src/acceleration/test/AccelerationIntraCommTest.cpp
BenjaminRodenberg: True. Will take care of this in #1696 and duplicate the test correspondingly.

BenjaminRodenberg:
> Do we now plan to make `substeps` default to `true` or `false` with v3.0?

This is probably similar to what we are currently discussing in https://github.com/precice/precice/issues/1690. IIRC the current default is `true`. But this would - in many cases - directly trigger an error message in the current state.

Going for some kind of `auto` option is probably complicated, because determining whether we should use `substeps="true"` or `substeps="false"` based on the acceleration scheme is tricky on the technical level. Therefore, I would vore for `substeps="true"`, because we want to prioritize waveform iterations with exchange of substeps. This will require some UX compromises and manual labor until quasi Newton is 100% working, but I think this is ok. 

Another idea, not sure if it is a good one: we could also add an additional check and only trigger the error above, if we detect that the user actually performed substeps. I guess this would bypass the error for most users, but it also makes the implementation more complicated, harder to test and the behavior does not solely depend on the configuration, but also on the runtime.

uekerman:
> Another idea, not sure if it is a good one: we could also add an additional check and only trigger the error above, if we detect that the user actually performed substeps. I guess this would bypass the error for most users, but it also makes the implementation more complicated, harder to test and the behavior does not solely depend on the configuration, but also on the runtime.

I agree and share the concerns. In general, sounds too complicated to me.

> Going for some kind of auto option is probably complicated

Agree, let's not do this here.

I see the situation as follows:

- With v3.0, some waveform iteration will be functional and we will have many important building blocks in the backend ready. But some important features will not yet be supported, such that `true` as default would lead to many errors and slowly all users will add `substeps="false"` to their configuration without even thinking about it. So `false` as default might be better.
- Eventually in some v3.x waveform iteration will be functional including all features. Then, defaulting to `true` would be better.
- Can / should we change the default behavior in a minor release? I think we did similar things in the past already. We also improve features in minor releases. I think it is not uncommon to do that.

So, I would vote for `false` now and `true` eventually.

MakisH:
To the extent I understand the situation:

- Currently, typical FSI scenarios use subcycling and IQN. This still remains supported with the current (v2) limitations (effectively no time interpolation applied).
- The term "subcycling" is already used to describe something different, so we should definitely change it in the error messages. I got a mini heart attack from thinking that at least half of our users will be affected by disabling subcycling.
- We are adding time interpolation, but it will not be as good as it could be till later in the v3.x release cycle. But we still need to explicitly disable half-baked issues, which then makes sense to disable by default.

> * Can / should we change the default behavior in a minor release? I think we did similar things in the past already. We also improve features in minor releases. I think it is not uncommon to do that.

As far as I see it, we should document the option already with its limitations and plans, and later advertise the "now time interpolation supports X, set Y to use it" and use it in the tutorials. It will be a new feature.

I would vote to stick to the policy of not changing the default in a minor release. We can guide users to explicitly switch something on with different ways (documentation, tutorials, warnings).

Let's set the option to `false` and keep it like that.

What would be the downside of asking the user to explicitly use the new features when available?

BenjaminRodenberg:
@MakisH mostly agree with the following two comments:

> * Currently, typical FSI scenarios use subcycling and IQN. This still remains supported with the current (v2) limitations (effectively no time interpolation applied).

I would say a typical FSI scenario uses no subcycling (see our tutorials). Generally subcycling may lead to numerical problems, if you don't do it right. I you know about users that (consciously) use subcycling, please let me know, because these might be great application cases for my thesis :smirk:  

> I would vote to stick to the policy of not changing the default in a minor release. We can guide users to explicitly switch something on with different ways (documentation, tutorials, warnings).
> 
> Let's set the option to `false` and keep it like that.
> 
> What would be the downside of asking the user to explicitly use the new features when available?

Ideally we would like to exchange substeps as the default option as soon as everything is working, because it would finally result in the behavior of subcycling that a user would also expect: I perform substeps and the data also arrives on the other side. Until then (half-baked situation) defaulting to `false` makes a lot of sense, but when everything is ready I think defaulting to `true` would cause the least pain for users. This would also help us that the feature is actually regularly used and tested (not like other features that are rarely used and where we lack experience and test coverage).

I'm not sure how far we are stretching the basic principles of semantic versioning, but if possible, I would also like to relax a bit and go change the default from `false` to `true` in some 3.x release as proposed by @uekerman.

MakisH:
> @MakisH mostly agree with the following two comments:
> 
> > * Currently, typical FSI scenarios use subcycling and IQN. This still remains supported with the current (v2) limitations (effectively no time interpolation applied).
> 
> I would say a typical FSI scenario uses no subcycling (see our tutorials).

hmm... true. None of our tutorials (OpenFOAM cases) currently uses subcycling. We do need a tutorial/test case for it!

src/acceleration/AitkenAcceleration.cpp
BenjaminRodenberg: Triggering error in initialize already and updated the error message.

BenjaminRodenberg:
Depending on the order of merging: If we merge #1732 first, I can also reduce the diff of this PR here quite a bit and remove `substep="false"` at several places.

src/acceleration/ConstantRelaxationAcceleration.cpp
```diff
@@ -37,7 +37,7 @@ void ConstantRelaxationAcceleration::performAcceleration(const DataMap &cplData)
 
   for (const auto &data : cplData | boost::adaptors::map_values) {
     if (data->exchangeSubsteps()) {
-      PRECICE_ERROR("Acceleration scheme does not support subcycling. Please pick a different acceleration scheme or set substeps=\"false\" in the exchange tag of data \"{}\".", data->getDataName());
+      PRECICE_ERROR("Constant acceleration does not yet support using data from all substeps. Please set substeps=\"false\" in the exchange tag of data \"{}\".", data->getDataName());
```
uekerman, line 35:
```suggestion
      PRECICE_ERROR("Constant underrelaxation does not yet support using data from all substeps. Please set substeps=\"false\" in the exchange tag of data \"{}\".", data->getDataName());
```

BenjaminRodenberg:
@uekerman can you take a look again? Changes are trivial. I think it generally makes sense introducing this error message. It does not matter what the default value of `substeps` is as both options are generally allowed and only some configurations are valid. We should merge now, because what's merged is merged and every open PR is a possible integration problem in the future ;)

#1732 is related to this one, but still has some minor problems and needs more time.

uekerman (approved):
Let's keep the `substeps="false"` in most tests despite being the default. Doesn't hurt and make things explicit

BenjaminRodenberg:
> Let's keep the `substeps="false"` in most tests despite being the default. Doesn't hurt and make things explicit

Well, the nice thing about using the default in the config is that it will also automatically change, if we change the default. This also makes sure that the tests are aligned what most of our users are likely to do. We can remove the `substeps="false"` or keep it as it is. In the end it's not worth the discussion in my opinion.

PR #1734, opened 2023-07-25
Exclude boundary vertices from index queries

### LINKED ISSUE #1733: Partition of Unity RBF mapping: assertion fails for tutorial
**Describe your setup**

- Operating system (e.g. Linux distribution and version): _Ubuntu 22.04_
- preCICE Version: `develop` (ef76221)
- OpenFOAM adapter: branch `workaround-connectivity-v3`, commit `91f4f0f`
- Tutorials: branch `develop`, commit `73a5edcb`
- OpenFOAM v2306

**Describe the problem**

Running the `perpendicular-flap` tutorial with `fluid-openfoam` and `solid-openfoam`, I stumble upon [this assertion](https://github.com/precice/precice/blob/ef76221a3ea9d5b4ba52e4ca524905ba275cb23c/src/mapping/impl/SphericalVertexCluster.hpp#L177):

```
---[precice]  Automatic RBF mapping alias from mesh "Fluid-Mesh" to mesh "Solid-Mesh" in "write" direction resolves to "partition-of-unity RBF" .
---[precice]  Computing "partition-of-unity RBF" mapping from mesh "Fluid-Mesh" to mesh "Solid-Mesh" in "write" direction.
ASSERTION FAILED
Location:   void precice::mapping::SphericalVertexCluster<RADIAL_BASIS_FUNCTION_T>::setNormalizedWeight(double, precice::VertexID) [with RADIAL_BASIS_FUNCTION_T = precice::mapping::CompactPolynomialC6; precice::VertexID = int]
File:       /home/gc/repos/precice/precice/src/mapping/impl/SphericalVertexCluster.hpp:177
Expression: normalizedWeight > 0
Rank:       0
Arguments:  none
```

[Configuration](https://github.com/precice/tutorials/blob/73a5edcb211ca3d06dffcbad1b6233b9c840d14a/perpendicular-flap/precice-config.xml):

```xml
    <mapping:rbf direction="write" from="Fluid-Mesh" to="Solid-Mesh" constraint="conservative">
      <basis-function:compact-polynomial-c6 support-radius="1." />
    </mapping:rbf>
    <mapping:rbf direction="read" from="Solid-Mesh" to="Fluid-Mesh" constraint="consistent">
      <basis-function:compact-polynomial-c6 support-radius="1." />
    </mapping:rbf>
``` 

**Step To Reproduce**

1. Go to the `tutorials/perpendicular-flap/fluid-openfoam` and run `./run.sh`
2. Go to the `tutorials/perpendicular-flap/solid-openfoam` and run `./run.sh`
3. After initialization, the assertion should be hit.

davidscn:
## Main changes of this PR

Excludes boundary vertices from index query in `getVerticesInsideBox`.

The applied change affects other RBF mappings as well. Since the basis-function implementation returns anyway zero when using the support radius as argument, it makes sense to not collect the vertices with a distance of `radius` in the first place. 

Fixes #1733

### DIFF
--- a/src/query/Index.cpp
+++ b/src/query/Index.cpp
@@ -230,7 +230,7 @@ std::vector<VertexID> Index::getVerticesInsideBox(const mesh::Vertex &centerVert
 
   const auto &          rtree = _pimpl->getVertexRTree(*_mesh);
   std::vector<VertexID> matches;
-  rtree->query(bgi::intersects(searchBox) and bg::index::satisfies([&](size_t const i) { return bg::distance(centerVertex, _mesh->vertices()[i]) <= radius; }),
+  rtree->query(bgi::intersects(searchBox) and bg::index::satisfies([&](size_t const i) { return bg::distance(centerVertex, _mesh->vertices()[i]) < radius; }),
                std::back_inserter(matches));
   return matches;
 }
--- a/src/query/Index.hpp
+++ b/src/query/Index.hpp
@@ -81,7 +81,7 @@ public:
   /// Get n number of closest triangles to the given vertex
   std::vector<TriangleMatch> getClosestTriangles(const Eigen::VectorXd &sourceCoord, int n);
 
-  /// Return all the vertices inside the box formed by vertex and radius
+  /// Return all the vertices inside the box formed by vertex and radius (boundary exclusive)
   std::vector<VertexID> getVerticesInsideBox(const mesh::Vertex &centerVertex, double radius);
 
   /// Return all the vertices inside a bounding box


src/query/Index.cpp
```diff
@@ -230,7 +230,7 @@ std::vector<VertexID> Index::getVerticesInsideBox(const mesh::Vertex &centerVert
 
   const auto &          rtree = _pimpl->getVertexRTree(*_mesh);
   std::vector<VertexID> matches;
-  rtree->query(bgi::intersects(searchBox) and bg::index::satisfies([&](size_t const i) { return bg::distance(centerVertex, _mesh->vertices()[i]) <= radius; }),
+  rtree->query(bgi::intersects(searchBox) and bg::index::satisfies([&](size_t const i) { return bg::distance(centerVertex, _mesh->vertices()[i]) < radius; }),
```
uekerman, line 233: Numerically, there is anyway no real difference between "<=" and "<", no? Then, I don't understand the motivation for this  change.
Shouldn't there be a

https://github.com/precice/precice/blob/fbe44d289a9d257d0e8ab1e226cf75f35eef5e23/src/math/differences.hpp#L89-L93

anyway?

davidscn: Hitting the 'exact' zero was no coincidence in the first place. Hence, there is a difference in both versions. To some extend, the problem was solved already with #1736 . I will subtract the numerical difference from the computed radius as well to make it a bit more robust, but this shouldn't really be necessary IMO.

PR #1735, opened 2023-07-25
Remove Broyden acceleration

### LINKED ISSUE #885: Remove Broyden acceleration
This acceleration type was added for the sole purpose of comparing it to the different quasi-Newton variants. There is no use case where the single-vector Broyden is superior to IQN-ILS or IQN-IMVJ.
On the other hand, the configuration of the method is a bit ugly as all quasi-Newton parameters are required as `BroydenAcceleration` inherits from `BaseQNAcceleration`. Those parameters are not used, however.
Therefore, I would suggest to simply remove the method.

Fujikawas:
## Main changes of this PR

Removes the Broyden acceleration, closes #885  

## Author's checklist

* ~~[ ] I added a test to cover the proposed changes in our test suite.~~

## Reviewers' checklist


uekerman (approved):
First PR merged, congrats :tada:

PR #1736, opened 2023-07-25
PUM mapping: use only a single cluster for small cases

davidscn:
## Main changes of this PR

If the user-specified target number of vertices per cluster/cluster size is higher than the number of vertices in the (local) mesh, we use just a single cluster and bypass the clustering algorithm. 

Open for discussion. In principle, that's not what the PUM was made for and a user could just use the 'available' global RBF mappings to achieve the same effect. Now that we have the alias tag, it would be somewhat tedious to go back and forth between the alias tag and a specific kind of mapping in the precice config (more arguments, different opening/closing tags). Therefore, I would just go with the proposed changes such that the PUM does the same by default -> no config changes needed. We could revert the change at anytime, if needed.

Will add a changelog, if approved.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mapping/impl/CreateClustering.hpp
+++ b/src/mapping/impl/CreateClustering.hpp
@@ -334,6 +334,15 @@ inline std::tuple<double, Vertices> createClustering(mesh::PtrMesh inMesh, mesh:
   inMesh->computeBoundingBox();
   auto localBB = inMesh->getBoundingBox();
 
+  // If we have less vertices in the whole domain than our target cluster size,
+  // we just use a single cluster. The clustering result of the algorithm further
+  // down is in this case not optimal.
+  // The single cluster has in principle a radius of inf. We use here twice the
+  // length of the longest bounding box edge length and the center of the bounding
+  // box for the center point.
+  if (inMesh->vertices().size() < verticesPerCluster)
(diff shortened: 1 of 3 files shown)


src/mapping/impl/CreateClustering.hpp
```diff
@@ -334,6 +334,15 @@ inline std::tuple<double, Vertices> createClustering(mesh::PtrMesh inMesh, mesh:
...
   auto localBB = inMesh->getBoundingBox();
 
+  // If we have less vertices in the whole domain than our target cluster size,
+  // we just use a single cluster. The clustering result of the algorithm further
+  // down is in this case not optimal.
+  // The single cluster has in principle a radius of inf. We use here twice the
+  // length of the longest bounding box edge length and the center of the bounding
+  // box for the center point.
+  if (inMesh->vertices().size() < verticesPerCluster)
+    return {localBB.longestEdgeLength() * 2, Vertices{mesh::Vertex({localBB.center(), 0})}};
```
uekerman, line 344: The radius here has no influence as long as the sphere covers all vertices, right?

uekerman (approved):
Sounds reasonable to me.

PR #1738, opened 2023-07-25
Remove extrapolation

### LINKED ISSUE #1731: The future of extrapolation

With current changes, extrapolation is becoming increasingly problematic to maintain (#1713) at a questionable practical benefit.
The second order extrapolation was very mystical before we dropped it (#1089 #1503), so we currently only support 1st order extrapolation. 

Due to the current refactoring, the configuration location for the extrapolation doesn't make sense any more. It should move to the `<data:...>` tag just like we did with the `waveform-degree` #1709.

A possible extension of the waveform interpolation would be to use extrapolation with explicit coupling. This is uncharted territory though and a possible future research topic for our waveform experts @BenjaminRodenberg and @NiklasKotarsky.

* Remove the support for extrapolation
* Only reintroduce it as `<data:scalar name="d" extrapolation-order="n" />` once we conducted enough research on it.

BenjaminRodenberg:
## Main changes of this PR

Removes extrapolation feature from coupling scheme. Closes #1731.

See #1731.

## Author's checklist

* [x] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html). -> https://github.com/precice/precice.github.io/commit/0b791b6f5477cb6b7b4e342fb5e94e21a0917bff

## Reviewers' checklist


uekerman (approved):
LGTM

PR #1744, opened 2023-07-26
Name CMake variables consistently

### LINKED ISSUE #1516: Naming convention for CMake variables
The current naming convention of our CMake variables has no clear pattern: we have underscore separated all capital variables such as `PRECICE_ENABLE_C` and `PRECICE_RELEASE_WITH_ASSERTION`, we have a mixture of underscore and CamelCase variables such as in `PRECICE_MPICommunication` and `PRECICE_PETScMapping` and no clear guideline what the variable is supposed to express: the dependency and configuration of (external) software or the purpose of the configuration in the library.  

CMake itself uses exclusively all caps and underscore naming conventions (e.g. `CMAKE_PREFIX_PATH`). I would propose to stick to the convention as it comes with certain benefits

1. you don't need to think about capitalization (PETSc and preCICE) and the random mixture we use `PRECICE_PETSc...`
2. underscore separation is better readable
3. we stick to the CMake style and convention. Other projects I looked at (e.g. gismo, deal.II or ginkgo) stick to the same naming convention.

Open question is then still what to encode into the variable name. What I have seen quite often is just to use the external package (without any semantics, e.g. `PRECICE_WITH_MPI` or `PRECICE_WITH_PETSC`), but maybe there are other ideas. See also additional context below. 

Invent a new naming convention people have to lookup every time it is being used and try to convince others that this new naming convention is better.

**Additional context**
In the context of recent GPU data mapping investigations, it becomes somewhat difficult to encode the purpose of the dependency into the CMake variable name. Using a similar style as the PETSc variable name we would maybe use `PRECICE_GinkgoMapping` (assuming we decided for Ginkgo as a GPU support). However, from the user perspective it is not clear that this method comes with a certain type of parallelization concept and hardware requirements (OpenMP or GPU). Looking at the variables it might be a user question, whether preCICE can perform data mapping without these dependencies or whether these dependencies are mandatory. The actual benefit of using Ginkgo would be the (shared memory) parallelization, so wouldn't `PRECICE_GinkgoMappingParallelization` be a better name? Using the dependency only for the CMake variables might be the most straightforward solution.

**Remotely related**
Looking at the CMake-GUI, it seems a bit arbitrary, which variables we expose by default (in the non-detailed view) to the user. Example: The GUI shows all boost component paths as well as a (private?) PETSc variable called `_petsc_lib_petsc_real`, the MPI paths are not shown. In my opinion, only the user-relevant variables should be given in the default gui view.

davidscn:
## Main changes of this PR

Names our CMake variables consistently and cleans-up some cmake variables, which are currently not marked as advanced. Current suggestions/renamings:

- `PRECICE_PETScMapping` -> `PRECICE_FEATURE_PETSC_MAPPING`
- `PRECICE_MPICommunication` -> `PRECICE_FEATURE_MPI_COMMUNICATION`
- `PRECICE_Packages` -> `PRECICE_CONFIGURE_PACKAGE_GENERATION`
- `PRECICE_PythonActions` -> `PRECICE_FEATURE_PYTHON_ACTIONS`
- `PRECICE_ENABLE_C` -> `PRECICE_BINDINGS_C`
- `PRECICE_ENABLE_FORTRAN` ->`PRECICE_BINDINGS_FORTRAN`
- `PRECICE_ENABLE_LIBBACKTRACE`  -> `PRECICE_FEATURE_LIBBACKTRACE_STACKTRACES`

All remaining variables remain unchanged. As discussed in #1516 the naming scheme is supposed to follow the convention `PRECICE_FEATURE`, i.e., `<PREFIX>_<DEPENDENCY>_<FEATURE>`. The only violation I can spot right now is `PRECICE_ENABLE_LIBBACKTRACE`, or should we go for `PRECICE_LIBBACKTRACE_STACKTRACES`? 

See #1516 and closes #1516 

## Author's checklist

* [ ] update spack and more downstream package manager

## Reviewers' checklist


CMakeLists.txt
```diff
@@ -53,15 +53,15 @@ if(NOT CMAKE_BUILD_TYPE)
...
 endif()
 
-option(PRECICE_MPICommunication "Enables MPI-based communication and running coupling tests." ON)
-option(PRECICE_PETScMapping "Enable use of the PETSc linear algebra library." ON)
-option(PRECICE_PythonActions "Python support" ON)
-option(PRECICE_Packages "Configure package generation." ON)
+option(PRECICE_MPI_COMMUNICATION "Enables MPI-based communication and running coupling tests." ON)
+option(PRECICE_PETSC_MAPPING "Enable use of the PETSc linear algebra library for global-iterative RBF data mappings." ON)
+option(PRECICE_PYTHON_ACTIONS "Enable Python support for preCICE actions." ON)
+option(PRECICE_CONFIGURE_PACKAGE_GENERATION "Configure package generation." ON)
```
MakisH, line 59: Why the `CONFIGURE_` prefix here? Sounds inconsistent, as everything else should also be `CONFIGURE_*`. `PRECICE_PACKAGE_GENERATION` sounds enough right to me right now. The `CONFIGURE_` sounds better if we also group things with `FEATURE_`, `BINDINGS_`, etc.

CMakeLists.txt
```diff
@@ -53,15 +53,15 @@ if(NOT CMAKE_BUILD_TYPE)
...
 option(BUILD_SHARED_LIBS "Build shared instead of static libraries" ON)
 option(BUILD_TESTING "Build tests" ON)
 option(PRECICE_ALWAYS_VALIDATE_LIBS "Validate libraries even after the validatation succeeded." OFF)
-option(PRECICE_ENABLE_C "Enable the native C bindings" ON)
-option(PRECICE_ENABLE_FORTRAN "Enable the native Fortran bindings" ON)
+option(PRECICE_C_BINDINGS "Enable the native C bindings" ON)
+option(PRECICE_FORTRAN_BINDINGS "Enable the native Fortran bindings" ON)
```
MakisH, line 64: To more easily find these in ccmake, would it make more sense to name them as:

```
PRECICE_BINDINGS_C
PRECICE_BINDING_FORTRAN
```
?

CMakeLists.txt
```diff
@@ -53,15 +53,15 @@ if(NOT CMAKE_BUILD_TYPE)
   set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS Debug Release RelWithDebInfo MinSizeRel)
 endif()
 
-option(PRECICE_MPICommunication "Enables MPI-based communication and running coupling tests." ON)
-option(PRECICE_PETScMapping "Enable use of the PETSc linear algebra library." ON)
-option(PRECICE_PythonActions "Python support" ON)
-option(PRECICE_Packages "Configure package generation." ON)
+option(PRECICE_MPI_COMMUNICATION "Enables MPI-based communication and running coupling tests." ON)
+option(PRECICE_PETSC_MAPPING "Enable use of the PETSc linear algebra library for global-iterative RBF data mappings." ON)
+option(PRECICE_PYTHON_ACTIONS "Enable Python support for preCICE actions." ON)
```
MakisH, line 58: We could also group these under the prefix `FEATURE_`, to signify to a completely new user (e.g., cluster admin), that toggling this variable toggles a feature:

```
PRECICE_FEATURE_MPI_COMMUNICATION
PRECICE_FEATURE_PETSC_MAPPING
PRECICE_FEATURE_PYTHON_ACTIONS
```

MakisH:
> The only violation I can spot right now is `PRECICE_ENABLE_LIBBACKTRACE`, or should we go for `PRECICE_LIBBACKTRACE_STACKTRACES`?

I also don't have any good idea on how to handle this. Does it toggle a feature, or choose between two different dependencies?

CMakeLists.txt
davidscn: In principle, these features are now already grouped together (for lexicographic reasons). We could go for the `_FEATURE_<feat>` naming (again would be nice for ccmake). Only counter argument is that the variables get rather lengthy.  What do you like more:

image

image

CMakeLists.txt
davidscn: I was a bit afraid that the plain `PRECICE_PACKAGE_GENERATION` would indicate that cmake actually generates the packages for you, but I will change it back once we decided below

CMakeLists.txt
MakisH: I would not rely to the fact that the `MPI_`, `PETSC_`, and `PYTHON_` are currently next to each other, as this can easily change in the future.

I see the issue with the truncation of names in ccmake, but this sounds like a technicality that should not concern us. ccmake could set this width arbitrarily, and more variables are already truncated.

I prefer the `FEATURE_` grouping, unless it becomes unclear to define what should belong into this group.

CMakeLists.txt
davidscn: > The CONFIGURE_ sounds better if we also group things with FEATURE_, BINDINGS_, etc.

Ok so do we keep it then?

MakisH: yes, makes sense.

CMakeLists.txt
davidscn: Yes, checked and it does. Just in the same way as ccmake does, no?

davidscn:
More on a side-note: do we also keep the default for all `FEATURE=ON` ? Some of them are more in use for corner-cases. Some projects use autodetection mechanisms for such purposes. Any opinions on this?

MakisH:
> More on a side-note: do we also keep the default for all `FEATURE=ON` ? Some of them are more in use for corner-cases. Some projects use autodetection mechanisms for such purposes. Any opinions on this?

As long as the user receives a clear message "PETSc not found - Set `PRECICE_FEATURE_PETSC_MAPPING=OFF`", I would be fine with the current behavior. The autodetection could lead to silently disabling some features, which may be making reproducibility more complicated.

CMakeLists.txt
fsimonis: image

Ok. It only groups on the first part.

fsimonis:
I like the new format. The feature prefix really makes them easier to detect in ccmake too. 

> Some projects use autodetection mechanisms for such purposes. Any opinions on this?

Autodetection is a huge pain as your entire system leaks into your builds, making them non-reproducible. I already ran into this with some other projects. Explicit options and hard errors are the way to go. CMake has some support for this, but it is super confusing from the developer and user perspective.

`libbacktrace` is intended to be a developer tool, so I wouldn't make it too prominent. Technically, it toggles more extensive stacktraces, which could be seen as a feature.

Some TODOs:
* Update `cmake/GenerateVersionInformation.cmake` (output of  `precice-tools version`)
* Update 'CMakePresets.txt`

davidscn:
> Update 'CMakePresets.txt`

I don't see any update requirements in the `CMakePresets.json`

fsimonis:
Oops right.

So it's just a question what to do with the libbacktrace now.

davidscn:
> So it's just a question what to do with the libbacktrace now.

Yes, for me, libbacktrace fits more into the category of developer features and not into our `FEATURE` category. Googletest might eventually end-up in the same. I would either

- `PRECICE_ENABLE_LIBBACKTRACE` -> leave it as it is right now
- `PRECICE_DEVEL_LIBBACKTRACE` -> to indicate development features
- `PRECICE_CONFIGURE_LIBBACKTRACE` -> in analogy to our package generation

Once we agreed here, I would merge this PR, as I understand that a corresponding spack patch is ready.

MakisH:
>     * `PRECICE_ENABLE_LIBBACKTRACE` -> leave it as it is right now

The `ENABLE_` prefix sounds a bit arbitrary now. Why should this be prefixed, but not, e.g., `PRECICE_MPI_MAPPING`?

>     * `PRECICE_DEVEL_LIBBACKTRACE` -> to indicate development features

This is the option I like the most, even though I would find it hard to argue that it is only a feature for developers of preCICE. I could imagine that people use stacktraces when debugging their codes, which use preCICE. But no real objection here.

>     * `PRECICE_CONFIGURE_LIBBACKTRACE` -> in analogy to our package generation

I thought that the package generation was really that we prepare the configuration for the package generation. The `CONFIGURE_` here sounds strange to me (again: why for this but not also for everything else).

davidscn:
> This is the option I like the most, even though I would find it hard to argue that it is only a feature for developers of preCICE. I could imagine that people use stacktraces when debugging their codes, which use preCICE. But no real objection here.

I mean we can also go for `PRECICE_FEATURE_LIBBACKTRACE_STACKTRACES` if you argue this way.

MakisH:
> > This is the option I like the most, even though I would find it hard to argue that it is only a feature for developers of preCICE. I could imagine that people use stacktraces when debugging their codes, which use preCICE. But no real objection here.
> 
> I mean we can also go for `PRECICE_FEATURE_LIBBACKTRACE_STACKTRACES` if you argue this way.

Sounds consistent, I guess let's do that.

PR #1746, opened 2023-07-26
Use absolute timestamps

### LINKED ISSUE #1729: Use absolute timestamps for samples

Currently, we save user-provided data as samples in a reference time frame from `[0, 1]`, where 0 is the beginning and 1 is the end of the current time window.

When a participant completes a time window, the sample at the end becomes the sample at the beginning `1 -> 0` and all other samples are discarded.

This only works if all coupling schemes use the same time window size, which isn't guaranteed #1713.
Otherwise, a participant may be at different relative locations in the differently sized time windows.

Example:
**todo**

As samples are stored globally in the `time::storage`, there is no unique relative time frame to store them to.

Furthermore, it is not clear which samples to remove after at least one coupling scheme completes a time window. Some coupling schemes may subcycle.  

Use absolute time stamps for saving samples.

* A participant is always at a known point in time, hence, time-stamping user-provided data becomes clear.
* If a coupling scheme completes a time window, we discard everything before the oldest time window start. For this, we replace `time::storage::move()` and `time::storage::trim()` with `time::storage::discardBefore()`.
* The interpolation functions operate on the global time frame instead of the relative one.

* Restrict samples to the compositional time windows, which is the union of all time window ends in time. For time window sizes 2 (0, 4, 6,…) and 3 (0, 3, 6,…), the union would be (0, 2, 3, 4, 6,…). **Problem** we only keep samples of the final compositional time window, breaking interpolation.
* Save samples in the coupling schemes, where the relative time frame is clear. **Problems:** Data duplication and mappings, actions, and exporters still need access to these samples.
* Save samples globally and save references to them from inside coupling schemes with relative time stamps. **Problems** Access and lifetime of samples will become tricky to get right. 

**Additional context**

A further upside of global time stamps is that this simplifies the implementation of data extrapolation in the future.

#1713

This could impact the work of @BenjaminRodenberg and @NiklasKotarsky

### LINKED ISSUE #1747: Refactor B-Spline interpolation into a separate file

We are currently using B-Spline interpolation in waveforms provided by Eigen.
This is implemented as part of the `time::Storage`.

The interface is clear though: `(timestamps, samples, timeToSample) -> sample`

We should extract this into a separate file, and test it.

**Additional context**

#1746

BenjaminRodenberg:
## Main changes of this PR

Use absolute time in storage for stamples. Remove scaled time window.

Closes #1729 
Closes #1747 

See #1729 

## Author's checklist

* [x] Merge https://github.com/precice/precice/pull/1820

## Reviewers' checklist


fsimonis:
I think a major issue is that the time window size of the second participant when dealing with prescribed time windows is not transferred to the second participant while it is waiting in `initialize()`.

This seems to be independent of these changes though.

fsimonis:
@uekerman @BenjaminRodenberg I have a conceptional question regarding time-window sizes (TWS) in combination with the participant-first method in serial-implicit coupling, aka first prescribing the TWS to second.

If the subsequent iteration prescribes a shorter TWS, then we can interpolate in the read data, but the end time sample needs to be removed at some point as it is outside the time-window.

If the subsequent iteration prescribes a longer TWS, then the end sample of the previous iteration is not at the end of the time-window. This allows the user to read past this time stamp, which essentially leads to an extrapolation.

Concrete test which fails:
https://github.com/BenjaminRodenberg/precice/blob/1729-use-absolute-timestamps/tests/serial/time/implicit/serial-coupling/ReadWriteScalarDataFirstParticipant.cpp

How is this supposed to work? Should we "lock" the TWS after it has been prescribes until once until it converges?

BenjaminRodenberg:
@fsimonis when I wrote the test you are mentioning I was mainly concerned about getting the participant first method working at all. The actual data values are not tested at all. Therefore, I wrote [the following](
https://github.com/BenjaminRodenberg/precice/blob/6e311463a02bf9f5c8cad84795dd7884a71389f9/tests/serial/time/implicit/serial-coupling/ReadWriteScalarDataFirstParticipant.cpp#L33-L34):

```
  // some dummy values, to check the actual values is not the point of this test,
  // but more to test whether reading / writing is possible at all
```

For the case you describe above there was no proper solution in the past (before waveforms were introduced). The first participant just used the data it received from second, no matter what time it was associated with.

Locking the time windows size would change the functionality quite a bit and require additional logic. Let's not do this.

I think we should introduce some kind of simple extrapolation that allows the first solver to get reasonable values, if it reads past the end of the time window of the second solver. Constant extrapolation should be fine (some more advanced schemes are a future work package #1731).

(Because this is a niche feature, I would suggest to deactivate this test and similar ones here, document the bug and fix it in another PR. It did not work properly before, it just somehow worked through the relative timestamps)

src/math/bspline.cpp
```diff
@@ -22,6 +22,8 @@ Eigen::VectorXd interpolateAt(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int
   const int ndofs = xs.rows(); // number of dofs. Each dof needs it's own interpolant.
 
   // transform all times to the relative interval [0; 1]
+  PRECICE_ASSERT(ts(0) != ts(ts.size() - 1));
+  // PRECICE_ASSERT((t >= ts(0)) && (t <= ts(ts.size() - 1)), "t lies outside of ts!", t, ts); // Only allowed to use BSpline for interpolation, not extrapolation.  // @todo fix this! Currently leads to failures in tests.
```
BenjaminRodenberg, line 26: I think we should try to also use this assertion to avoid unintended B-Spline extrapolation. Besides this point we are getting closer to the finish line here.

BenjaminRodenberg: We can also possibly add these assertions later via #1775 in favor of merging this PR earlier. The possible bugs I mention #1775 already exist on develop. Whether we fix them here or in #1775 does not really matter.

fsimonis:
I restricted the failing participant-first test to stick to a time window size in all iterations of a time window.

The MultiCoupling tests fail, as the storage does not get advanced properly. The stample from t=0 remains in the storage after having moved to the next time window.
So the expected constant interpolation which happens when there is only one sample is not used and the existing stamples don't overlap with the requested time leading to the assertion.

fsimonis:
At this point, only the `Integration/Serial/MultiCoupling/MultiCoupling` test is failing.


@uekerman Can you check the MultiCoupling tests changed in https://github.com/precice/precice/pull/1746/commits/d3cfe786b18e934cec486cde3644d185f3fb4983 ? Do these changes make sense?

uekerman:
> @uekerman Can you check the MultiCoupling tests changed in https://github.com/precice/precice/commit/d3cfe786b18e934cec486cde3644d185f3fb4983 ? Do these changes make sense?

You are doing two things:
- [ ] You remove the acceleration from all tests. This still gives valid tests, but, of course, lowers the test coverage. If this is on purpose, I would recommend documenting this in an issue, such that we add the acceleration again eventually.
- You change the three-solvers tests to always read 0 at the **end** of the first time window. This looks wrong or, at least, I don't understand why. Also suspicious that you don't update the four-solvers tests.

BenjaminRodenberg:
I reverted [6de6fcb](https://github.com/precice/precice/pull/1746/commits/6de6fcb0051d9a44cf47436832fc916b6101a7dd) and [8a79f35](https://github.com/precice/precice/pull/1746/commits/8a79f35d939ce6f2715673f46ced59ff248e2b1b), because the changes confused me and led to many failing tests. Will try to get my head around the multi coupling soon and hopefully find the problem. We can, of course, recover the changes from above, if there is a good reason. As a first step I would like to get the tests working.

src/cplscheme/CompositionalCouplingScheme.cpp
```diff
@@ -161,7 +161,7 @@ bool CompositionalCouplingScheme::hasDataBeenReceived() const
 double CompositionalCouplingScheme::getTime() const
 {
   PRECICE_TRACE();
-  auto schemes = allSchemes();
+  auto schemes = schemesToRun(); // analogous change to https://github.com/precice/precice/pull/1792, but if we use allSchemes() tests do not trigger. Why?
```
BenjaminRodenberg, line 164: @fsimonis I think this change would also make sense, but there is not really an observable effect. How to treat time with compositional coupling schemes is generally an important question as soon as we have moved to absolute timestamps.

BenjaminRodenberg:
Current status:

* Failing test (runs into assertion): `mpirun -np 4 ./testprecice -t Integration/Serial/MultiCoupling`
* PR still needs reviews and polishing.

BenjaminRodenberg:
@fsimonis and @uekerman I think this PR is ready for review. Tests are working and I did some cleaning up. I'm offline for the next two weeks. Feel free to merge this PR, if you think it's ready :wave:

uekerman:
@fsimonis OK if I leave the (first) review to you here? Please ping if I should also have a look.

fsimonis:
Going through the commits and notices https://github.com/precice/precice/pull/1746/commits/19ccbae9894307f0ae37ae18debea47aa5d9e779. The point of `ensureDataHasNotYetBeenReceived` is to trigger the assertion before receiving data a second time, essentially preventing a second receive.

src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -262,9 +258,16 @@ class BaseCouplingScheme : public CouplingScheme {
...
-  void receiveData(const m2n::PtrM2N &m2n, const DataMap &receiveData, bool initialCommunication = false);
+  void receiveData(const m2n::PtrM2N &m2n, const DataMap &receiveData);
+
+  /**
+   * @brief Like receiveData, but sets window time to end of window. Needed by SerialCouplingScheme.
+   *
+   * @param m2n M2N used for communication
+   * @param receiveData DataMap associated with received data
+   */
+  void receiveDataForWindowEnd(const m2n::PtrM2N &m2n, const DataMap &receiveData);
```
uekerman, line 270: Couldn't we move this function to `SerialCouplingScheme`?

Could you please write more in the comment on why this function is needed? This one is a bit tricky, we will wonder in the future.

uekerman (approved):
Looks good to me. Seems to simplify things :+1:
Only minor comments.

tests/serial/mapping-volume/helpers.cpp
```diff
@@ -83,7 +83,7 @@ void testMappingVolumeOneTriangle(const std::string configFile, const TestContex
     // Expected value in the middle of the triangle is the average of inputs (1, 10, 100)
     expected << 111.0 / 3;
 
-    participant.readData(meshName, dataName, vertexIDs, dt, readData);
+    participant.readData(meshName, dataName, vertexIDs, 0, readData);
```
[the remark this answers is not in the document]
BenjaminRodenberg: Yes. Let's try to resolve this independently.

src/cplscheme/BaseCouplingScheme.hpp
BenjaminRodenberg: Sure. But then `BaseCouplingScheme::_computedTimeWindowPart` needs a protected setter to allow `SerialCouplingScheme` to access it. Not sure that this really simplifies things. I would rather keep it as it is, because it's less risky to expose `receiveDataForWindowEnd` to `ParallelCouplingScheme` than exposing `_computedTimeWindowPart` to all derived coupling schemes.

BenjaminRodenberg: I added some more information on the function.

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -1348,7 +1339,6 @@ void ParticipantImpl::mapReadData()
   computeMappings(_accessor->readMappingContexts(), "read");
   for (auto &context : _accessor->readDataContexts()) {
     if (context.hasMapping()) {
-      PRECICE_DEBUG("Map read data \"{}\" to mesh \"{}\"", context.getDataName(), context.getMeshName());
```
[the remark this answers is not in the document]
BenjaminRodenberg: Reverted this change.

tests/serial/mapping-volume/helpers.cpp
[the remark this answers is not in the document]
BenjaminRodenberg: With proper initialization this change is not needed anymore. Still: Not sure why this modification is needed. There seems to be something strange, if no initial values are provided in some mapping situations (and this problem only becomes apparent here). I'll try to isolate it and might open an issue, if I know more.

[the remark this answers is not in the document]
BenjaminRodenberg: I created https://github.com/precice/precice/pull/1820 and will also merge this fix into this PR.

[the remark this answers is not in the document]
BenjaminRodenberg: Just reverted the changed in this file to the state from develop. The changes are not needed anymore. I guess this was fixed in a different way. We can still decide to merge or not merge #1820, but this is now independent of this PR.

src/cplscheme/CompositionalCouplingScheme.cpp
BenjaminRodenberg: I just reconsidered this point: for `getTime()` using `allSchemes()` vs. `schemesToRun()` does not make a difference, if we take the minimum, because the schemes that are included in `allSchemes()` but not in `schemesToRun()` (i.e. the `convergedSchemes`) have already reached the end of the time window. Meaning: Their `getTime()` is always greater or equal to the time of the `schemesToRun()` and therefore does not matter for the minimum.

I'll remove this comment and we can go ahead.

Btw: Situation was different in #1792, because there we looked at the relative time, which is set to zero for converged schemes. Therefore, this influenced the minimum.

BenjaminRodenberg:
@fsimonis and @uekerman: I'm done with polishing here. I'll merge tomorrow, if nobody stops me until then ;)

PR #1754, opened 2023-07-31
Add default settings to IQN-ILS and IQN-IMVJ acceleration methods 

### LINKED ISSUE #1505: Defining default values for the IQN configuration
When creating a new coupled simulation, there are many user-defined parameters for the acceleration scheme required in the configuration XML file. This is fine for experienced users, but can be confusing for new users. Especially as the parameters may significantly influence convergence. It would be helpful to allow for default configuration values to be used in the case where they are simply ignored. For example, a typical IQN-ILS acceleration scheme requires:

```xml
      <acceleration:IQN-ILS>
        <data name="Displacement" mesh="Solid-Mesh" />
        <data name="Stress" mesh="Fluid-Mesh-Centers" />
        <preconditioner type="residual-sum" />
        <filter type="QR2" limit="1.2e-3" />
        <initial-relaxation value="0.1" />
        <max-used-iterations value="60" />
        <time-windows-reused value="15" />
      </acceleration:IQN-ILS>
```
Each line needs to be defined by the user. 

Add a set of default configuration parameters for the `implicit` coupling schemes, such that they are selected if the user does not specify a different parameter. The example above can then be reduced to

```xml
      <acceleration:IQN-ILS>
        <data name="Displacement" mesh="Solid-Mesh" />
        <data name="Stress" mesh="Fluid-Mesh-Centers" />
      </acceleration:IQN-ILS>
```
and pre-configured default values would be selected. The variables that need to be exchanged are still required in the configuration as this cannot be assumed. 

The default values will not necessarily be optimal for every test case, but provide a good set of default values. A first attempt to define them would be:
- `<preconditioner type="residual-sum" />`  - Only if parallel-implicit coupling is defined. This ensures good scaling between data values that are potentially orders of magnitude different.
- `<filter type="QR2" limit="1e-2" />` - The QR2 filter uses a relative criterion compared to the QR1 filter, and a value of `1e-2` has been shown to be a good choice. 
- `<initial-relaxation value="0.1" />` - this has always tended to be the default for setting up test cases, and never shown to be problematic.
- ` <max-used-iterations value="100" />` - This defines the maximum number of columns in the QR decomposition. This can be set quite high and the size of QR can be controlled by the number of time windows reused. 
-  `<time-windows-reused value="10" />` - The previous 10 time steps has often been used. However 15 or 20 may also be valid options. In the case of instability in a solver, this information will remain in the IQN method for longer and potentially hurt convergence. 

None

**Additional context**

Fujikawas:
## Main changes of this PR
add default settings to IQN acceleration methods.

closes #1505 

## Author's checklist

* [ ] ~~For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).~~

## Reviewers' checklist


Fujikawas:
The pre-conditioner type is set to "residual-sum" as default for all coupling schemes, not only for parallel-implicit coupling as mentioned in the issue.
I'm not sure about if we need to provide the default settings according to the coupling scheme types and if yes, how could we do it elegantly.

docs/changelog/1754.md
```diff
@@ -0,0 +1 @@
+- Add default settings about initial-relaxation, max-used-iterations, time-windows-reused, filter and preconsitioner to IQN-ILS and IQN-IMVJ acceleration methods
```
uekerman, line 1:
```suggestion
- Added default values for the IQN-ILS and IQN-IMVJ configuration options `initial-relaxation`, `max-used-iterations`, `time-windows-reused`, `filter`, and `preconditioner`.
```

tests/quasi-newton/serial/DefaultConfig.xml
```diff
@@ -0,0 +1,57 @@
...
+    <exchange data="Data1" mesh="MeshOne" from="SolverOne" to="SolverTwo" />
+    <exchange data="Data2" mesh="MeshOne" from="SolverTwo" to="SolverOne" />
+    <max-iterations value="100" />
+    <relative-convergence-measure limit="1e-7" data="Data2" mesh="MeshOne" />
+    <acceleration:IQN-ILS>
+      <data name="Data2" mesh="MeshOne" />
+      <!-- <filter type="QR2" limit="1e-1" /> -->
+      <!-- <initial-relaxation value="1.0" /> -->
+      <!-- <max-used-iterations value="56" />  -->
+      <!-- <time-windows-reused value="0" /> -->
```
uekerman, line 54:
```suggestion
```

src/acceleration/config/AccelerationConfiguration.hpp
```diff
@@ -117,6 +117,34 @@ class AccelerationConfiguration : public xml::XMLTag::Listener {
...
+    int         timeWindowsReused          = 10;
+    int         filter                     = Acceleration::QR2FILTER;
+    double      singularityLimit           = 1e-2;
+    std::string preconditionerType         = "residual-sum";
+    int         precond_nbNonConstTWindows = -1;
+  } _defaultValuesIQNILS;
+
+  const struct DefaultValuesIMVJ {
+    double      relaxationFactor           = 0.1;
+    int         maxIterationsUsed          = 10;
```
uekerman, line 132:
```suggestion
    int         maxIterationsUsed          = 20;
```
Isn't 10 a bit too small?

src/acceleration/config/AccelerationConfiguration.cpp
```diff
@@ -263,11 +266,13 @@ void AccelerationConfiguration::xmlEndTagCallback(
...
       } else {
         // no preconditioner defined
+        /*if ()
         std::vector<double> factors;
         for (int id = 0; id < static_cast<int>(_config.dataIDs.size()); ++id) {
           factors.push_back(1.0);
         }
-        _preconditioner = PtrPreconditioner(new ConstantPreconditioner(factors));
+        _preconditioner = PtrPreconditioner(new ConstantPreconditioner(factors));*/
+        _preconditioner = PtrPreconditioner(new ResidualSumPreconditioner(_defaultValuesIQNILS.precond_nbNonConstTWindows));
```
uekerman, line 275: Please don't comment lines out, but directly delete them. Will only confuse someone in three years. "Why is this commented out, should I comment it in again?"

src/acceleration/config/AccelerationConfiguration.cpp
```diff
@@ -359,7 +374,7 @@ void AccelerationConfiguration::addCommonIQNSubtags(xml::XMLTag &tag)
                              "Please note that a QR1 is based on Given's rotations whereas QR2 uses "
                              "modified Gram-Schmidt. This can give different results even when no columns "
                              "are filtered out.");
-  XMLAttribute<double> attrSingularityLimit(ATTR_SINGULARITYLIMIT, 1e-16);
+  XMLAttribute<double> attrSingularityLimit(ATTR_SINGULARITYLIMIT, 1e-2);
```
uekerman, line 377: I am not sure if we should change it here as well. 1e-2 is not a good default for QR1. 1e-16 is in a way as it means filter is off.

src/acceleration/config/AccelerationConfiguration.cpp
```diff
@@ -399,7 +414,8 @@ void AccelerationConfiguration::addTypeSpecificSubtags(
     tagData.addAttribute(attrMesh);
     tag.addSubtag(tagData);
   } else if (tag.getName() == VALUE_IQNILS) {
-    XMLTag tagInitRelax(*this, TAG_INIT_RELAX, XMLTag::OCCUR_ONCE);
+
+    XMLTag tagInitRelax(*this, TAG_INIT_RELAX, XMLTag::OCCUR_NOT_OR_ONCE);
     tagInitRelax.setDocumentation("Initial relaxation factor.");
```
uekerman, line 419: I think we should add information on the default behavior in all of these documentation strings.

```suggestion
    tagInitRelax.setDocumentation("Initial relaxation factor. If tag is not provided, an initial relaxation of 0.1 is used.");
```

similarly below; also for filter and preconditioner

uekerman (changes requested):
Very clean code, only a few additional tweaks needed.

uekerman:
After merging this, let's consider updating some of our tutorials and also the documentation.

PR #1745, opened 2023-07-26
Test precice-events analyze if pandas is available

davidscn:
## Main changes of this PR

Tests `precice-events analyze` if pandas is available and adds pandas to our CI.

## Motivation and additional information

Closes #1708 

## Author's checklist

* [ ] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [ ] I added a test to cover the proposed changes in our test suite.
* [ ] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).
* [ ] I sticked to C++17 features.
* [ ] I sticked to CMake version 3.16.3.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

Issue #1749, opened 2023-07-27
Simple performance regression tests


MakisH:
We sometimes discover too late that changes in `develop` may be increasing the runtime impact of preCICE.

As discussed in the coding days, let's set up a simple GitHub Actions workflow that:

- Is triggered at some stage (e.g., when a PR is approved)
- Runs the tests (`make test`) for the current and a previous (develop/released) version of preCICE
- Reports the runtime used by each test
- Does not succeed or fail, but only reports results that a human can judge

To ensure that we check this:

- Add a reminder to the PR template checklist
- Have a bot (e.g., @precice-bot) report the results in the PR

uekerman:
Trigger could also work by talking to the bot. Maybe we don't want to do this for every PR.

fsimonis:
We could write the performance regression tests purely as user-facing tests using the API of preCICE and then use the profiling tools to extract the measurements we care about.

Conceptually, they would be something between integration tests and tutorials/ASTE.

These could either live in `precice/precice` in `extra/performance/` or in a separate repository.

BenjaminRodenberg:
> These could either live in `precice/precice` in `extra/performance/` or in a separate repository

I think they should definitely live in this repository to make sure we detect performance regressions early (before merging a PR).

fsimonis:
We have some basic benchmarks #1502

moksha-hub:
@uekerman can i get assigned.

MakisH:
@moksha-hub please [do not ask to be assigned](https://github.com/precice/precice/issues/2457).

Contributions are welcome if they happen and are at a state to review. But keep in mind that this issue might be more complicated than one thinks.

Eshaan-byte:
Hi @MakisH I've been looking into this. A simple approach that builds on what's already in the repo:

-Trigger via a label (trigger-perf-test) rather than on every PR — similar to trigger-system-tests
-Check out base and HEAD in the same job so the comparison runs on the same machine, reducing runner noise ([ref](https://werat.dev/blog/running-benchmarks-for-pull-requests-via-github-actions/))
-Reuse the existing precice-bench (Google Benchmark) with --benchmark_repetitions=3, compare using gbench.report, and post the output as a PR comment
-No pass/fail — just numbers for a human to judge

Some questions that i want to get clarified:
-The existing benchmarks are mostly micro-benchmarks on internals. The issue mentions timing make test too — should the workflow also time the full ctest run, or is the Google Benchmark suite enough for a first version?

[...]

Issue #1708, opened 2023-07-04
`precice-events analyze` crashes


davidscn:
**Describe your setup**

Recent develop snapshot  394d648 and python 3.10

**Describe the problem**
I'm trying to use the profiler. I have a very basic serial run and after running `precice-events merge <dir>` I would like to investigate the result of `precice-events analyze`. However, the latter command fails with the following backtrace:

```bash
Reading events file events.json
Output timing are in us.
Traceback (most recent call last):
  File "~/precice/tools/profiling/precice-events", line 719, in <module>
    sys.exit(dispatcher.get(args.cmd, showHelp)(args))
  File "~/precice/tools/profiling/precice-events", line 710, in <lambda>
    "analyze": lambda ns: analyzeCommand(ns.eventfile, ns.participant, ns.output,  ns.unit),
  File "~/precice/tools/profiling/precice-events", line 472, in analyzeCommand
    joined.reset_index(inplace=True, names="event")
  File "~/.local/lib/python3.10/site-packages/pandas/util/_decorators.py", line 311, in wrapper
    return func(*args, **kwargs)
TypeError: DataFrame.reset_index() got an unexpected keyword argument 'names'

```

**Step To Reproduce**
The behavior can easily be reproduced using our test data (e.g. [`from-common-dir`](https://github.com/precice/precice/tree/develop/tools/profiling/tests/from-common-dir))
1. `precice-events merge`
2. `precice-events analyze B`

**Expected behaviour**
No crash

**Additional context**
This feature seems to be untested, so we cannot expect it to work in the first place-

davidscn:
The trivial change of

https://github.com/precice/precice/blob/394d6483effeee43c4d23cff709fafe7106ca566/tools/profiling/precice-events#L472

```diff
- joined.reset_index(inplace=True, names="event") 
+joined.reset_index(inplace=True) 
```

fixed it for me. Not sure whether this is a proper solution, though.

fsimonis:
Looks like a problem with the pandas version.

We currently don't test this as I didn't want to make pandas a dependency of preCICE for the sole sake of testing.
Of course, we could add an optional switch to run these tests is pandas is found and install pandas in the ci-images.

uekerman:
Having this in the CI would indeed be good moving forward :+1:

fsimonis:
A question here is which baseline we are going to pick for pandas.

The above error is due to a keyword required by pandas 1.5.0.

[...]

PR #1756, opened 2023-07-31
Test precice-events analyse with venv

fsimonis:
Use python venv to install polars and use it in tests

## Main changes of this PR

Adds tests for `precie-events analyze`.
They rely on the python package polars, which is installed in a venv.

I couldn't push to the PR ref, so I squashed and created an new branch

Superseeds #1745

## Author's checklist

## Reviewers' checklist


davidscn:
> I couldn't push to the PR ref, so I squashed and created an new branch

Maybe we should discuss once more or anybody else could explain to me what 'assignment' in issues means. My understanding: clarify responsibility and avoid race conditions in development (#1745 was WIP). One could think #1708 is the only issue left in our repo. I only took responsibility there to distribute work, but will unassign, there is apparently no need for load balancing.

fsimonis:
> Maybe we should discuss once more or anybody else could explain to me what 'assignment' in issues means. My understanding: clarify responsibility and avoid race conditions in development (https://github.com/precice/precice/pull/1745 was WIP).

Correct, assignment means responsibility, which I broke.
When I reviewed the changes, I noticed that pip pandas pulled in a newer numpy version, which lead to conflicting version when importing python files in the python actions, once I figured that out I tried virtual environments, which triggered problems in CMake and the generated Debian packages, so I had a look at using a single virtualenv for only the tests, but the FindPython logic of CMake seems to be buggy and doesn't prefer the virtualenv even though it should.

After, I had the full picture, I just decided to go the extra step and port the tests.

On a more important side node, there really seems to be an issue with pushing to other users forks.

fsimonis:
I think the venv infrastructure could be important in case we decide to do more elaborate testing in the future such as #1749.

PR #1757, opened 2023-07-31
Bump version on develop to 3.0.0

fsimonis:
## Main changes of this PR

This PR bumps the version on develop to 3.0.0.

It also adds a manual conflict with libprecice2 and declares it to provide the `preCICE` virtual package. This should cover future major versions.

This should simplify porting dependants of the preCICE core.

Generated Debian packages and pkg-config should work too.
Note that this will change the major version, which is part of the Debian package name. So this one directly conflicts with the old one.

## Author's checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,5 +1,5 @@
 cmake_minimum_required (VERSION 3.16.3)
-project(preCICE VERSION 2.5.0 LANGUAGES CXX)
+project(preCICE VERSION 3.0.0 LANGUAGES CXX)
 set(preCICE_SOVERSION ${preCICE_VERSION_MAJOR})
 
 #
--- a/cmake/CPackConfig.cmake
+++ b/cmake/CPackConfig.cmake
@@ -87,6 +87,11 @@ set(CPACK_DEBIAN_PACKAGE_CONTROL_EXTRA "${preCICE_SOURCE_DIR}/tools/releasing/pa
 set(CPACK_DEBIAN_PACKAGE_GENERATE_SHLIBS TRUE)
 set(CPACK_DEBIAN_PACKAGE_GENERATE_SHLIBS_POLICY "=")
 
+# Handling version conflicts
+set(CPACK_DEBIAN_PACKAGE_PROVIDES precice)
+set(CPACK_DEBIAN_PACKAGE_CONFLICTS "precice, libprecice2")
+set(CPACK_DEBIAN_PACKAGE_REPLACES precice)
+
 # Install doc files
 install(FILES tools/releasing/packaging/debian/copyright
   DESTINATION ${CMAKE_INSTALL_DATAROOTDIR}/doc/${CPACK_PACKAGE_NAME}


davidscn:
> Note that this will change the major version, which is part of the Debian package name. So this one directly conflicts with the old one.

Yes, but the symlink should point to the latest installation such that linking works as expected. Also, one could (and probably should) always run `make uninstall` in the previous build dir to prevent such conflicts.

I'm in favor of such a change. I'm not sure how correct our version string will be rendered, though `This is preCICE version 3.0.0 - <commits after v2.5.0>. Some pre-release marking could still be useful to not get confused when jumping back and forth between dev and release version in the future.

davidscn:
> I have an idea. Maybe we can declare preCICE as a virtual package and declare a conflict with the virtual preCICE package for upcoming major releases as well as explicitly declaring a conflict for libprecice2.

I think we are not the first package to have a new major release. How do other packages handle this? To some extend this is a problem of the user. I would just go with the default debian way, but if this sounds convincing to you, go ahead.

fsimonis:
Tried this in a docker container and it works!
It even works both ways, so installing the libprecice2 will replace the libprecice3 package. 

```console
# apt install ./libprecice2_2.5.0_focal.deb
Reading package lists... Done
Building dependency tree
Reading state information... Done
Note, selecting 'libprecice2' instead of './libprecice2_2.5.0_focal.deb'
The following NEW packages will be installed:
  libprecice2
0 upgraded, 1 newly installed, 0 to remove and 12 not upgraded.

# precice-tools version
2.5.0;no-info [git failed to run];PRECICE_MPICommunication=Y;PRECICE_PETScMapping=Y;PRECICE_PythonActions=Y;PRECICE_ENABLE_C=Y;PRECICE_ENABLE_FORTRAN=Y;CXX=GNU;CXXFLAGS= -O3 -DNDEBUG;LDFAGS=

# apt install ./libprecice3_3.0.0_focal.deb
Reading package lists... Done
Building dependency tree
Reading state information... Done
Note, selecting 'libprecice3' instead of './libprecice3_3.0.0_focal.deb'
The following packages will be REMOVED:
  libprecice2
The following NEW packages will be installed:
  libprecice3
0 upgraded, 1 newly installed, 1 to remove and 12 not upgraded.
Need to get 0 B/3391 kB of archives.
After this operation, 26.5 MB of additional disk space will be used.
Do you want to continue? [Y/n]

# precice-tools version
3.0.0;no-info [not a repository];PRECICE_MPICommunication=Y;PRECICE_PETScMapping=Y;PRECICE_PythonActions=Y;PRECICE_ENABLE_C=Y;PRECICE_ENABLE_FORTRAN=Y;CXX=GNU;CXXFLAGS= -g;LDFLAGS=

# apt install ./libprecice2_2.5.0_focal.deb
Reading package lists... Done
Building dependency tree
Reading state information... Done
Note, selecting 'libprecice2' instead of './libprecice2_2.5.0_focal.deb'
The following packages will be REMOVED:
  libprecice3
The following NEW packages will be installed:
  libprecice2
0 upgraded, 1 newly installed, 1 to remove and 12 not upgraded.
Need to get 0 B/1662 kB of archives.
After this operation, 26.5 MB disk space will be freed.
Do you want to continue? [Y/n]
```

fsimonis:
Squashed some fixes and rebased. The arch CI still fails until openmpi 4.1.6 is released.

PR #1760, opened 2023-08-01
Link developer tooling from CONTRIBUTING.md

### LINKED ISSUE #1481: Check and update CONTRIBUTING
Two things I stumbled upon:

- no mention of `pre-commit`, we should probably also avoid the duplication of content
- a dead link still pointing to the doxygen dev docs

There might be more things that are outdated.

MakisH:
Closes #1481.

The links are already fine, main remaining issue was linking to the pre-commit hook description: https://precice.org/dev-docs-dev-tooling.html

It also removes the duplication regarding formatting.

### DIFF
--- a/docs/CONTRIBUTING.md
+++ b/docs/CONTRIBUTING.md
@@ -10,18 +10,17 @@ In order to get your contributions into the code base as smoothly as possible, p
   * Make sure you fill in the earliest version that you know has the issue.
 * Fork the repository on GitHub.
 * Create a feature branch, based on `develop`, from where you want to base your work. For simplicity, prefix the branch name either with `add-` or `fix-`.
-* Make commits of logical units.  
+* Make commits of logical units.
   Write [good commit messages][commit].
   Check for unnecessary whitespace with `git diff --check` before committing.
-* Write tests to assure your feature works as expected and prevent it from getting broken in the future.  
+* Write tests to assure your feature works as expected and prevent it from getting broken in the future.
   See our example tests in `src/testing/tests/ExampleTests.cpp` and the [documentation of boost.test][boosttest] for more information.
 * Run _all_ the tests to assure nothing else accidentally broke.
-* Follow our [style guide][style].
-  Install `clang-format-8` and run `tools/formatting/format-all` to format your code.
-* Submit a pull request to the repository in the preCICE organization.  
-  See [Collaboration workflow with pull requests and issues][workflow] and [Creating a pull request][pullrequest]
+* Use the available [tools to format and check your contribution](https://precice.org/dev-docs-dev-tooling.html), in particular the pre-commit hook that will format the code following our [style guide][style].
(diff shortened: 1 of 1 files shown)


davidscn (approved):
Looks good in general. I'm wondering whether the full links make more sense in such documents. It makes obvious where the link targets and one can follow the links from an offline editor as well.

MakisH:
What do you mean by "full links"?

docs/CONTRIBUTING.md
```diff
@@ -10,18 +10,17 @@ In order to get your contributions into the code base as smoothly as possible, p
...
+* Write tests to assure your feature works as expected and prevent it from getting broken in the future.
   See our example tests in `src/testing/tests/ExampleTests.cpp` and the [documentation of boost.test][boosttest] for more information.
 * Run _all_ the tests to assure nothing else accidentally broke.
-* Follow our [style guide][style].
-  Install `clang-format-8` and run `tools/formatting/format-all` to format your code.
-* Submit a pull request to the repository in the preCICE organization.  
-  See [Collaboration workflow with pull requests and issues][workflow] and [Creating a pull request][pullrequest]
+* Use the available [tools to format and check your contribution](https://precice.org/dev-docs-dev-tooling.html), in particular the pre-commit hook that will format the code following our [style guide][style].
+* Submit a pull request to the repository in the preCICE organization and fill the provided template.
+  See [Collaboration workflow with pull requests and issues][workflow] and [Creating a pull request][pullrequest] to get started.
```
davidscn, line 21: @MakisH
```suggestion
  See [Collaboration workflow with pull requests and issues](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests) and [Creating a pull request](https://docs.github.com/en/pull-requests/collaborating-with-pull-requests/proposing-changes-to-your-work-with-pull-requests/creating-a-pull-request) to get started.
```

MakisH: The `[pullrequest]` is actually a reference to the bottom of the document, not a relative link. I was also confused about this before.

In that sense, one could argue that my new link is inconsistent with the rest.

davidscn: Are you sure? Where is it pointing to then. Looking here https://github.com/precice/precice/blob/develop/docs/CONTRIBUTING.md it clearly refers to the links given above, no?

MakisH: I guess we have some misunderstanding here.

When rendering the file on GitHub, I see full, absolute links, which work.
When looking at the source, the `[workflow]` points to `[workflow]: https://docs.github.com/en/pull-requests/collaborating-with-pull-requests`. This seems to be the same link as the one you are adding, just different Markdown syntax.

davidscn: Ah ok, didn't look at the offline version, this is probably GH standard then, thanks for clarification!

Issue #1752, opened 2023-07-28
Do not interpolate data for existing timestamps


fsimonis:
**Please describe the problem you are trying to solve.**

When reading data for a given dt, then `time::storage` creates a b-spline and interpolates based on it.
Only exception is if there is only a single data point.

If a sample already exists at the given dt, then the time storage still interpolates instead of returning the matching sample.

**Describe the solution you propose.**

Directly use a sample if the requested time matches its timestamp.

**Additional context**

https://github.com/precice/precice/pull/1746

BenjaminRodenberg:
In the [documentation](https://www.eigen.tuxfamily.org/dox/unsupported/structEigen_1_1SplineFitting.html#adc80b6f0dd0dbbea28130fb254626874) of Eigen it says

> Fits an interpolating [Spline](https://www.eigen.tuxfamily.org/dox/unsupported/classEigen_1_1Spline.html) to the given data points.

Interpolating spline means it fulfils the interpolation condition, meaning $f(t_i) = x_i$ for the input data points $(t_i,x_i)$. The fitting is here only how this is done algorithmically, because you need to somehow find the control points that fulfill the interpolation condition. I'm not sure about the tolerance, but it should be small enough, if Eigen claims the resulting spline is interpolating.

Under the hood I assume what's happening is similar to scipy, where the documentation is a bit more helpful. See https://docs.scipy.org/doc/scipy/tutorial/interpolate/smoothing_splines.html#tutorial-interpolate-fitpack.

fsimonis:
I ran into cases where for pretty simple tests, I required a higher tolerance than we use by default.

Concrete example of a midpoint in case of linear interpolation:

https://github.com/precice/precice/blob/b4c5df1f2e059906b43c20543f1c96f74d331b13/src/math/tests/BSplineTest.cpp#L66C14

Maybe we should extend the test coverage here to prevent bad surprises.

BenjaminRodenberg:
> Concrete example of a midpoint in case of linear interpolation:

But that's a different issue, because here you are interpolating and you don't use data from an existing timestamp.

The [following](https://github.com/precice/precice/blob/b4c5df1f2e059906b43c20543f1c96f74d331b13/src/math/tests/BSplineTest.cpp#L96) looks more fitting, no? Here it looks like the interpolation condition is only satisfied up to a certain tolerance.

> Maybe we should extend the test coverage here to prevent bad surprises.

But aren't you already testing this here?

[...]

Issue #1699, opened 2023-06-28
Rename `waveform-order` to `waveform-degree`


BenjaminRodenberg:
Because `waveform-order="1"` leads to (piecewise) linear interpolation (i.e. polynomial degree = 1, but the interpolation order of the scheme or convergence order of the time stepping scheme can be higher). So let's talk about the `waveform-degree="1"` instead.

Alternatives:

* `interpolating-polynomial-degree="1"`
* `bspline-degree="1"`

BenjaminRodenberg:
Another thing to keep in mind: Sooner or later we might want to use alternative interpolation schemes to B-splines. Example: Create a polynomial fit of degree $p$ that approximates $n$ given substeps.

uekerman:
`waveform-degree` is good :+1:

BenjaminRodenberg:
Closed via #1714

PR #1765, opened 2023-08-09
Speed up bspline interpolation

NiklasKotarsky:
## Main changes of this PR
This pull request reworks the spline interpolation in preCICE.

~~Closes issue #1752~~ See comment by @fsimonis below.
This branch is built upon #1746 where the bspline interpolation was separated into its own file. 

## Author's checklist

## Reviewers' checklist


src/time/Waveform.hpp
BenjaminRodenberg, line 1: What's the reason again for removing this class? I think we might have discussed this and it might also be a good idea to put all the functionality into `Storage`, but I do not exactly remember. In any case, we should definitely separate the *refactoring* "Remove `Waveform` class" from the *feature* "Speed up bspline interpolation" and create indepdenent PRs. I'm quite sure we can just go for the refactoring first.

src/math/bspline.cpp
```diff
@@ -34,17 +33,55 @@ Eigen::VectorXd interpolateAt(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int
...
+  }
+  // @todo implement cache to avoid unnecessary recomputation. Important! Need to reset cache when entering next window or iteration.
+  interpolated = interpolate(xs, splineDegree, ts, t, ndofs);
 
-  const int splineDimension = 1;
+  return interpolated;
+}
+
+void bspline::ComputeMatrices(const Eigen::MatrixXd pts, int degree, Eigen::VectorXd knot_parameters, int ndofs)
+{
```
NiklasKotarsky, line 47: Add comment that this comes from eigen.

src/time/Storage.hpp
```diff
@@ -23,7 +24,7 @@ class Storage {
    *
    * This Storage is used in the context of Waveform relaxation where samples in time are provided. Starting at the beginning of the window with time 0.0 and reaching the end of the window with time 1.0.
    */
-  Storage();
+  Storage(int _interpolationOrder);
```
BenjaminRodenberg, line 27: Do we really need to change the constructor here to accept `_interpolationOrder`? (btw: Please write `interpolationOrder` for this argument name. The `_variableName` is reserved for class members.

If possible, please use `Storage::setInterpolationOrder`. I remember that I introduced this function, because at the time where we have to construct the `Storage` (somewhere in the configuration), we do not know the interpolation degree yet (somewhere *later* in the configuration).

Also important: Please write `interpolationDegree`, not `interpolationOrder`. We fixed the naming in https://github.com/precice/precice/issues/1699, because we actually talk about the B-Spline degree and not necessarily about the order or interpolation scheme.

src/math/bspline.hpp
```diff
@@ -11,6 +11,16 @@ namespace precice::math::bspline {
  * @pre splineDegree must be larger than 0.
  * @pre t must be within [ts.min; ts.max].
  */
-Eigen::VectorXd interpolateAt(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree, double t);
+class bspline {
```
BenjaminRodenberg, line 14:
```suggestion
class Bspline {
```
Because it's a class.

src/math/bspline.hpp
```diff
@@ -11,6 +11,16 @@ namespace precice::math::bspline {
...
 
+public:
+  Eigen::VectorXd interpolateAt(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree, double t, bool recomputeBuffer);
```
BenjaminRodenberg, line 17:
```suggestion
  Eigen::VectorXd interpolateAt(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree, double t);
```
Not needed, if we make `_recomputeBuffer` a member of `Bspline`.

src/time/Waveform.cpp
```diff
@@ -23,6 +23,7 @@ int Waveform::getDegree() const
 
 time::Storage &Waveform::timeStepsStorage()
 {
+  _recomputeBuffer = true;
```
BenjaminRodenberg, line 26:
```suggestion
  _bspline.requireBufferUpdate();
```
Instead of maintaining `_recomputeBuffer` here, you can just tell `Bspline` that a buffer update will be required. I think this would look the following:
```
void Bspline::requireBufferUpdate() {
    _recomputeBuffer = true;
}
```

src/time/Waveform.cpp
```diff
@@ -43,9 +44,11 @@ Eigen::VectorXd Waveform::sample(double normalizedDt) const
 
   PRECICE_ASSERT(usedDegree >= 1);
 
-  const auto data = _timeStepsStorage.getTimesAndValues();
+  const auto data   = _timeStepsStorage.getTimesAndValues();
+  auto       sample = _bspline.interpolateAt(data.first, data.second, usedDegree, normalizedDt, _recomputeBuffer);
+  _recomputeBuffer  = false;
```
BenjaminRodenberg, line 49: If `_recomputeBuffer` is inside `Bspline`, you can also just take care of setting it `false` inside `Bspline::computeMatrices()`

src/math/bspline.hpp
```diff
@@ -11,6 +11,16 @@ namespace precice::math::bspline {
...
+
+private:
+  //interpolate(xs, splineDegree, ts, tRelative, ndofs);
+  Eigen::VectorXd interpolate(const Eigen::MatrixXd pts, int degree, Eigen::VectorXd knot_parameters, double tRelative, int ndofs);
+  void            ComputeMatrices(const Eigen::MatrixXd pts, int degree, Eigen::VectorXd knot_parameters, int ndofs);
```
BenjaminRodenberg, line 22:
```suggestion
  void            computeMatrices(const Eigen::MatrixXd pts, int degree, Eigen::VectorXd knot_parameters, int ndofs);
```
Because it is a function.

src/math/tests/BSplineTest.cpp
BenjaminRodenberg, line 1: Looks like some tests are currently failing, but this is probably the work-in-progress part, isn't it?

BenjaminRodenberg (changes requested):
Overall very nice implementation :+1: I Especially like the "lazy" way of recomputing the buffer. @NiklasKotarsky can you take care of the remaining todos?

With respect to the refactoring: Let's take care of the refactoring as soon as this PR and #1794 is merged. I think the refactoring is not really critical and we have enough more important things to do at the moment. The code duplication does not really harm a lot.

src/math/bspline.hpp
```diff
@@ -11,6 +11,16 @@ namespace precice::math::bspline {
...
  * @pre t must be within [ts.min; ts.max].
  */
-Eigen::VectorXd interpolateAt(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree, double t);
+class bspline {
 
+public:
+  Eigen::VectorXd interpolateAt(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree, double t, bool recomputeBuffer);
+
+private:
+  //interpolate(xs, splineDegree, ts, tRelative, ndofs);
```
BenjaminRodenberg, line 20:
```suggestion
```

src/math/bspline.hpp
BenjaminRodenberg, line 1: Please also generally add some more documentation for this class.

BenjaminRodenberg (changes requested):

src/time/Waveform.hpp
BenjaminRodenberg: Let's postpone the refactoring, because it is not critical. I took care of splitting the two issues.

NiklasKotarsky:
> Overall very nice implementation 👍 I Especially like the "lazy" way of recomputing the buffer. @NiklasKotarsky can you take care of the remaining todos?
> 
> With respect to the refactoring: Let's take care of the refactoring as soon as this PR and #1794 is merged. I think the refactoring is not really critical and we have enough more important things to do at the moment. The code duplication does not really harm a lot.

Yes, I should be able to fix the remaining things.

src/cplscheme/CouplingData.cpp
```diff
@@ -119,19 +119,16 @@ void CouplingData::storeIteration()
 
 const Eigen::VectorXd CouplingData::previousIteration() const
 {
-  PRECICE_ASSERT(!_previousTimeStepsStorage.stamples().empty());
```
BenjaminRodenberg, line 122: Why did you remove these assertions? Or was this an accident from the force-pushing here?

src/cplscheme/SerialCouplingScheme.cpp
```diff
@@ -84,7 +84,7 @@ void SerialCouplingScheme::receiveAndSetTimeWindowSize()
     PRECICE_ASSERT(not doesFirstStep(), "Only second participant can receive time window size.");
 
     if (hasTimeWindowSize() && isImplicitCouplingScheme() && not hasConverged()) { // Restriction necessary as long as extrapolation is not implemented. See https://github.com/precice/precice/issues/1770 for details.
-      PRECICE_CHECK(dt == getTimeWindowSize(), "May only use a larger time window size in the first iteration of the window. Otherwise old time window size must equal new time window size.");
+      PRECICE_ASSERT(dt == getTimeWindowSize(), "May only use a larger time window size in the first iteration of the window. Otherwise old time window size must equal new time window size.");
```
BenjaminRodenberg, line 87: Please revert this change. The `PRECICE_CHECK` is correct.

src/time/Storage.cpp
```diff
@@ -126,30 +140,30 @@ std::pair<Eigen::VectorXd, Eigen::MatrixXd> Storage::getTimesAndValues() const
...
 
-Eigen::VectorXd Storage::sample(double normalizedDt) const
+Eigen::VectorXd Storage::sample(double normalizedDt)
 {
   const int usedDegree = computeUsedDegree(_degree, nTimes());
 
   PRECICE_ASSERT(math::equals(this->maxStoredNormalizedDt(), time::Storage::WINDOW_END), this->maxStoredNormalizedDt()); // sampling is only allowed, if a window is complete.
 
-  if (usedDegree == 0) {
+  if (_degree == 0) {
```
BenjaminRodenberg, line 149: Please revert this change. `usedDegree` is correct.

src/time/Storage.cpp
```diff
@@ -126,30 +140,30 @@ std::pair<Eigen::VectorXd, Eigen::MatrixXd> Storage::getTimesAndValues() const
...
 
-Eigen::MatrixXd Storage::sampleGradients(double normalizedDt) const
+Eigen::MatrixXd Storage::sampleGradients(double normalizedDt)
 {
   const int usedDegree = computeUsedDegree(_degree, nTimes());
 
   PRECICE_ASSERT(math::equals(this->maxStoredNormalizedDt(), time::Storage::WINDOW_END), this->maxStoredNormalizedDt()); // sampling is only allowed, if a window is complete.
 
-  if (usedDegree == 0) {
+  if (_degree == 0) {
```
BenjaminRodenberg, line 166: See above.

src/time/Waveform.cpp
BenjaminRodenberg, line 1: The code removal here already introduces some changes that thematically belong into https://github.com/precice/precice/pull/1795/files. I would suggest to wrap up https://github.com/precice/precice/pull/1795/files first and then continue here. This makes this PR much more readable and simplifies things a lot.

src/time/Waveform.hpp
BenjaminRodenberg: See comment above: Refactoring and this PR is still a bit mixed. I think we can solve this by wrapping up https://github.com/precice/precice/pull/1795/files first.

src/math/bspline.hpp
```diff
@@ -1,16 +1,44 @@
...
 
-namespace precice::math::bspline {
+namespace precice::math {
+
+class Bspline {
 
-/** Uses B-Spline interpolation to compute compute x(t) from given data (x0,t0), (x1,t1), ..., (xn,tn).
+public:
+  /**
+ * @brief Uses B-Spline interpolation to compute compute x(t) from given data (x0,t0), (x1,t1), ..., (xn,tn).
```
BenjaminRodenberg, line 10:
```suggestion
 * @brief Uses B-Spline interpolation to compute x(t) from given data (x0,t0), (x1,t1), ..., (xn,tn).
```

src/math/bspline.hpp
```diff
@@ -1,16 +1,44 @@
...
+  /**
+   * @brief Calculates the knot vector and control points for the Bspline interpolation
+   *
+   * @param pts the values that are supposed to be interpolated
+   *
+   * @param knot_parameters the timestamps or values along which we want to interpolate
+   */
+  void computeMatrices(const Eigen::MatrixXd pts, int degree, Eigen::VectorXd knot_parameters, int ndofs);
+
+  bool            _recomputeBuffer = true; // Signals that the buffer should be recomputed next time interpolateAt is called
```
BenjaminRodenberg, line 40: Rather than calling this a "buffer" I would call it a "cache", because it is more specific. A buffer just stores some data, a cache stores data to speed up computations, but we might have to recompute, if there is a cache miss.

src/time/Storage.cpp
```diff
@@ -84,6 +92,9 @@ void Storage::move()
   PRECICE_ASSERT(math::equals(_stampleStorage.front().timestamp, time::Storage::WINDOW_START), _stampleStorage.front().timestamp);
   PRECICE_ASSERT(math::equals(_stampleStorage.back().timestamp, time::Storage::WINDOW_END), _stampleStorage.back().timestamp);
 
+  // The buffer in _bsplines has to be recomputed since the data changed
+  _bspline.requireRecomputeBuffer();
```
BenjaminRodenberg, line 96: Interface of the `BSpline` class looks good and simple :+1:

src/time/Storage.cpp
BenjaminRodenberg, line 1: To make it more secure: If we change `_degree`, we also need to call `_bspline.requireRecomputeBuffer()`, correct? This is not a use-case at the moment, but we should still cover this case, if the API allows us to change the degree.

docs/changelog/1765.md
```diff
@@ -0,0 +1 @@
+- Improved the runtime of readData with subcycling.
```
BenjaminRodenberg, line 1:
```suggestion
- Reduced the runtime of readData by optimizing BSpline interpolation.
```

BenjaminRodenberg (changes requested):
Some minor changes are still needed, but overall design looks good. I think we should first go for #1795 to tidy up things. Can you ask @uekerman for a final review, because I'm not available for the next two weeks?

src/time/Waveform.cpp
BenjaminRodenberg: I took care of merging #1795 into `develop` and will merge the updated `develop` into this PR. I think this should now be fine.

src/math/bspline.cpp
BenjaminRodenberg: Still a todo?

src/math/bspline.hpp
BenjaminRodenberg, line 1: One more thing I just realized: Since `Bspline` is now a class, please also rename `bspline.cpp` and `bspline.hpp` to `Bspline.cpp` and `Bspline.hpp`.

NiklasKotarsky:
> Some minor changes are still needed, but overall design looks good. I think we should first go for #1795 to tidy up things. Can you ask @uekerman for a final review, because I'm not available for the next two weeks?

Thanks, will do.

src/time/Storage.cpp
```diff
@@ -16,6 +16,7 @@ const double Storage::WINDOW_END = 1.0;
 Storage::Storage()
     : _stampleStorage{}, _degree(0)
 {
+  math::Bspline _bspline = math::Bspline();
 }
```
uekerman, line 20:
```suggestion
Storage::Storage()
    : _stampleStorage{}, _degree(0), _bspline()
{
}
```

@fsimonis Please correct me if this is nonsense :smile:

src/math/Bspline.hpp
```diff
@@ -0,0 +1,44 @@
...
+namespace precice::math {
+
+class Bspline {
+
+public:
+  /**
+ * @brief Uses B-Spline interpolation to compute x(t) from given data (x0,t0), (x1,t1), ..., (xn,tn).
+ * Also updates the cache if the flag _recomputeCache is true and t does not equal any of the timestamps in ts.
+ *
+ * @param ts the Timestamps which must be sorted from lowest to highest and contain at least 2 samples.
```
uekerman, line 13:
```suggestion
 * @param ts the timestamps which must be sorted from lowest to highest and contain at least 2 samples.
```

src/math/Bspline.cpp
```diff
@@ -0,0 +1,91 @@
...
+  PRECICE_ASSERT(ts.size() >= 2, "Interpolation requires at least 2 samples");
+#if EIGEN_VERSION_AT_LEAST(3, 4, 0)
+  PRECICE_ASSERT(std::is_sorted(ts.begin(), ts.end()), "Timestamps must be sorted");
+#else
+  PRECICE_ASSERT(std::is_sorted(ts.data(), ts.data() + ts.size()), "Timestamps must be sorted");
+#endif
+
+  // organize data in columns. Each column represents one sample in time.
+  PRECICE_ASSERT(xs.cols() == ts.size());
+  const int ndofs = xs.rows(); // number of dofs. Each dof needs it's own interpolant.
```
uekerman, line 25:
```suggestion
  const int ndofs = xs.rows(); // number of dofs. Each dof needs its own interpolant.
```

uekerman (approved):
Looks good to me. I did not check the actual Bspline implementation details, but believe that the test coverage is good enough.

Design-wise, it could be nice if the storage could signal the changes implicitly. Then, we could avoid the explicit calls and are sure that we won't forget them in the future.
@fsimonis Wasn't there some listener functionality in boost we could use? What would you recommend?

uekerman:
@NiklasKotarsky Please resolve any comments above once done such that we can keep a better overview.

fsimonis (changes requested):
Semi-automatic caching is always error-prone. Using listeners as suggested by @uekerman doesn't make this better.
I think this suggestion just needs some design feedback to make it more robust.

The initial idea of moving the interpolation to the function  `bspline::interpolateAt()` was to have one place that takes samples, timestamps and interpolates with a given spline degree.

The current solution converts this to a semi-automatic caching method, which needs to be manually reset. This is very error-prone, and the recomputation prevents the interpolation and thus sampling methods to be `const`, which really doesn't make sense.

I propose the following design:

* BSpline is a class that contains control points, the knot vector, the spline degree, and the min and max timestamp.
* It can be constructed from timestamps, samples, and a spline degree
* It can be sampled at any t within the min and max timestamps.

```cpp
class BSpline {
public:
  BSpline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree);

  Eigen::VectorXd interpolateAt(double t) const;
private:
  Eigen::VectorXd _knots;
  Eigen::MatrixXd _ctrls;
  double minT, maxT;
  int splineDegree;
};
```

This way, both timestamps and data are only required to be passed during construction. The constructed BSpline can be sampled only with a t.

The `time::Storage` then needs to handle the validity of these BSplines. We can model that by using an `std::optional<BSpline>` in the `time::Storage`. We delete it whenever the Storage changes and recreate it only if we attempt to interpolate, and the BSpline doesn't yet exist.

src/math/Bspline.cpp
```diff
@@ -0,0 +1,75 @@
...
+#else
+  PRECICE_ASSERT(std::is_sorted(ts.data(), ts.data() + ts.size()), "Timestamps must be sorted");
+#endif
+
+  // organize data in columns. Each column represents one sample in time.
+  PRECICE_ASSERT(xs.cols() == ts.size());
+  _ndofs            = xs.rows(); // number of dofs. Each dof needs its own interpolant.
+  _tsMin            = ts(0);
+  _tsMax            = ts(ts.size() - 1);
+  auto relativeTime = [this](double t) -> double { return (t - _tsMin) / (_tsMax - _tsMin); };
```
fsimonis, line 29: I honestly don't know if this is well-defined, as you capture `this` by value inside the constructor. I would go safe instead and simplify this by capturing `_tsMin` and `_tsMax` by value instead.

```suggestion
  auto relativeTime = [_tsMin, _tsMax](double t) -> double { return (t - _tsMin) / (_tsMax - _tsMin); };
```

src/math/Bspline.cpp
```diff
@@ -0,0 +1,75 @@
...
+  A(n - 1, n - 1) = 1.0;
+
+  auto qr = A.householderQr();
+
+  Eigen::MatrixXd controls = Eigen::MatrixXd::Zero(n, _ndofs);
+
+  for (int i = 0; i < _ndofs; i++) {
+    controls.col(i) = qr.solve(xs.row(i).transpose());
+  }
+  _ctrls = controls;
```
fsimonis, line 55: This is the last use of `controls`, so we can move it and save a copy of the Matrix.

```suggestion
  _ctrls = std::move(controls);
```

src/math/Bspline.cpp
```diff
@@ -0,0 +1,75 @@
...
+{
+  // transform t to the relative interval [0; 1]
+  const double tRelative = (t - _tsMin) / (_tsMax - _tsMin);
+  PRECICE_ASSERT(math::greaterEquals(tRelative, 0.0), "t is before the first sample!", tRelative, 0.0, _tsMin); // Only allowed to use BSpline for interpolation, not extrapolation.
+  PRECICE_ASSERT(math::smallerEquals(tRelative, 1.0), "t is after the last sample!", tRelative, 1.0, _tsMax);   // Only allowed to use BSpline for interpolation, not extrapolation.
+
+  Eigen::VectorXd interpolated(_ndofs);
+  constexpr int   splineDimension = 1;
+
+  //@todo: solve this differently
```
fsimonis, line 68: Is this still a problem?

src/time/Storage.cpp
```diff
@@ -148,7 +163,19 @@ Eigen::VectorXd Storage::sample(double normalizedDt) const
...
+  //If t equals to one timestep in data.first then we want to return the corresponding sample
+  for (int i = 0; i < data.first.size(); i++) {
+    if (math::equals(data.first[i], normalizedDt)) {
+      return data.second.col(i);
+    }
+  }
+
+  //Create a new bspline if _bspline does not already contain a spline
+  if (!_bspline.has_value()) {
+    _bspline = std::make_optional<math::Bspline>(data.first, data.second, usedDegree);
```
fsimonis, line 175: At this point, we know that the bspline is empty. So we can directly construct the B-Spline inside the optional using [`.emplace()`](https://en.cppreference.com/w/cpp/utility/optional/emplace).

```suggestion
    _bspline.emplace(data.first, data.second, usedDegree);
```

docs/changelog/1765.md
```diff
@@ -0,0 +1 @@
+- Reduced the runtime of readData by optimizing BSpline interpolation.
```
fsimonis, line 1: Optimizations can be many things :wink:
Interesting from a user-perspective is that these changes reduce the cost of subsequent evaluations as they construct the B-Spline only once, which makes higher-order schemes more cost-effective.

fsimonis (commented):
Looks very clean now!
Some last comments from my side then we can merge.

fsimonis:
Sorry about that!

> auto relativeTime = [_tsMin, _tsMax](double t) -> double { return (t - _tsMin) / (_tsMax - _tsMin); };

should have been
```cpp
auto relativeTime = [tMin = _tsMin, tMax = _tsMax](double t) -> double { return (t - tMin) / (tMax - tMin); };
```

fsimonis:
This PR prevents recomputations, but doesn't really close the issue in https://github.com/precice/precice/issues/1752.

BenjaminRodenberg:
> This PR prevents recomputations, but doesn't really close the issue in #1752.
> 
> _Not a problem, just a comment_ 😉

True. I would suggest to keep #1752 open and merge this PR as soon as it is ready.

src/time/Storage.cpp
```diff
@@ -148,7 +163,19 @@ Eigen::VectorXd Storage::sample(double normalizedDt) const
 
   auto data = getTimesAndValues();
 
-  return math::bspline::interpolateAt(data.first, data.second, usedDegree, normalizedDt);
+  //If t equals to one timestep in data.first then we want to return the corresponding sample
+  for (int i = 0; i < data.first.size(); i++) {
+    if (math::equals(data.first[i], normalizedDt)) {
+      return data.second.col(i);
+    }
+  }
```
BenjaminRodenberg, line 171: @NiklasKotarsky you are referring to this part resolving #1752, correct? @fsimonis and I have some concerns w.r.t possible jumps in the interpolant. Is this part here really necessary to solve your performance problem or is it not really relevant? If there is no clear performance gain, I would leave it out in favor of a simpler and more consistent implementation. But you probably know the pros and cons best.

NiklasKotarsky: Yes, after some thought I would also say that this is not going to matter from a performance perspective, since one could always use 0:th order interpolation.

fsimonis:
> I moved it to the storage class, since the Bspline class does not store the time steps.

@NiklasKotarsky I totally missed this. :sweat_smile: 
IMHO this is the right location.
There may be more performance improvements possible here, but I would rather get this PR merged and incrementally improve performance.

Also, the conflicts possibly originate from me squashing #1746. They should be minor, no worries.

fsimonis (approved):
Thanks for the substantial contribution!

Time to merge

PR #1769, opened 2023-08-11
Correct definition of convergence measure

### LINKED ISSUE #1768: Docs/XML reference for residual-relative-convergence-measure incorrect
The [XML reference for `residual-relative-convergence-measure`](https://precice.org/configuration-xml-reference.html#residual-relative-convergence-measure-1) states that the residual of the current iteration is compared to the residual of the previous iteration. However, checking [the implementation of this convergence measure](https://github.com/precice/precice/blob/develop/src/cplscheme/impl/ResidualRelativeConvergenceMeasure.hpp#L58) reveals that the residual of the current iteration is compared to `_normFirstDiff`, which is only computed in the first iteration of a time window[^1].

I have not managed to find out where the documentation for this XML tag is located, but the formula should be updated to reflect what is actually implemented: i.e., we divide by `\| H(x^0) - x^0 \|`, not `\| H(x^{k-1}) - x^{k-1} \|`.

[^1]: a new measurement series is created every time window, I checked this in a different part of the code

valentinaschueller:
## Main changes of this PR

I fixed the formula given in the description of the XML tag `residual-relative-convergence-measure`, this PR thus closes #1768 

This is a small update to in-code documentation, no changes w.r.t. code behavior etc. Please let me know if you would like me to add anything to the PR, I don't see any reason to add tests or a changelog file.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/cplscheme/config/CouplingSchemeConfiguration.cpp
+++ b/src/cplscheme/config/CouplingSchemeConfiguration.cpp
@@ -543,7 +543,7 @@ void CouplingSchemeConfiguration::addTagResidualRelativeConvergenceMeasure(
                                XMLTag::OCCUR_ARBITRARY);
   tagConvergenceMeasure.setDocumentation(
       "Residual relative convergence criterion based on the relative two-norm differences of data values between iterations.\n"
-      "\\$$\\frac{\\left\\lVert H(x^k) - x^k \\right\\rVert_2}{\\left\\lVert H(x^{k-1}) - x^{k-1} \\right\\rVert_2} < \\text{limit}\\$$");
+      "\\$$\\frac{\\left\\lVert H(x^k) - x^k \\right\\rVert_2}{\\left\\lVert H(x^0) - x^0 \\right\\rVert_2} < \\text{limit}\\$$");
   addBaseAttributesTagConvergenceMeasure(tagConvergenceMeasure);
   XMLAttribute<double> attrLimit(ATTR_LIMIT);
   attrLimit.setDocumentation("Limit under which the measure is considered to have converged. Must be in \\((0, 1]\\).");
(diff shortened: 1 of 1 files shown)


valentinaschueller:
@davidscn would you like to take care of this PR, since you also looked at the issue? Is there anything I should add or is it fine as is?

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -543,7 +543,7 @@ void CouplingSchemeConfiguration::addTagResidualRelativeConvergenceMeasure(
                                XMLTag::OCCUR_ARBITRARY);
   tagConvergenceMeasure.setDocumentation(
       "Residual relative convergence criterion based on the relative two-norm differences of data values between iterations.\n"
-      "\\$$\\frac{\\left\\lVert H(x^k) - x^k \\right\\rVert_2}{\\left\\lVert H(x^{k-1}) - x^{k-1} \\right\\rVert_2} < \\text{limit}\\$$");
+      "\\$$\\frac{\\left\\lVert H(x^k) - x^k \\right\\rVert_2}{\\left\\lVert H(x^0) - x^0 \\right\\rVert_2} < \\text{limit}\\$$");
```
davidscn, line 546: May I ask you to also update the documentation sentence: "values between iterations" doesn't fit to the formula IMO

davidscn (commented):
Only one comment (see below). I'm not overly familiar with this class or the corresponding notation. So, I was a bit hesitant to take over, but can do.

src/cplscheme/config/CouplingSchemeConfiguration.cpp
valentinaschueller: I tried to come up with a clearer description, see the new commit! (It's arguably clearer than those for the other convergence criteria now, but I don't want to start working on those as well :smile:)

davidscn:
Thanks a lot for your valuable contribution!

PR #1775, opened 2023-08-15
Prevent unintended extrapolation in B-splines

BenjaminRodenberg:
## Main changes of this PR

Adds assertion to prevent unintended extrapolation.

Additional security check to prevent unintended B-spline extrapolation. This might happen, if the storage is not cleared properly due to an incorrect implementation.

Some tests fail due to this new assertion (e.g. `Integration/Serial/Whitebox`). This is a hint for a possible bug in the implementation.

## Author's checklist

## Reviewers' checklist


fsimonis:
Running all tests in isolation, only `Integration/Serial/Whitebox/TestExplicitWithDataScaling` failed.

PR #1787, opened 2023-08-31
Rename events to profiling

### LINKED ISSUE #1764: Consider renaming precice-events

We currently profile sections of the code, which we internally call `Events`.
These are saved in a folder called `precice-events` by default, which can be configured using the `<profiling>` tag.

To process these files, we need to use the `precice-events` tool.
All of this is documented as “Performance analysis” in the user docs, as well as "Timings in preCICE" in dev docs.

We use a very inconsistent naming across preCICE.
Technically we record events and save them as profiling data.
Using the same naming for the tool and the directory doesn't make sense though.

We should decide on a naming before 3.0.0.

I propose to use `profiling` as a main name:
* tool: `precice-profiling`
* directory: `./precice-profiling/`

* `precice-perf` and `precice-perf-data` from Linux `perf`
* `precice-trace` and `precice-tracing` from `strace`

fsimonis:
## Main changes of this PR

This PR renames events to profiling in many locations.

Closes #1764

## Author's checklist


fsimonis:
I added the documentation changes to the precice-v3 branch of the website.

Issue #1712, opened 2023-07-05
Mixing tolerances for time-windows in compositional coupling schemes


fsimonis:
**Describe your setup**

preCICE Version: develop @ https://github.com/precice/precice/commit/3bfc9d9015c15e4ef8967269ccc67e92a9884086

**Describe the problem**

```xml
  <coupling-scheme:serial-explicit>
    <participants first="A" second="B" />
    <time-window-size value="1" valid-digits="10" />
  </coupling-scheme:serial-explicit>

  <coupling-scheme:serial-explicit>
    <participants first="B" second="C" />
    <time-window-size value="1" valid-digits="8" />
  </coupling-scheme:serial-explicit>
```

If B calls `advance(1.0-10e-9)`, then the first coupling scheme has completed, but the second hasn't despite equal time-windows and points in time.

The tolerance should be equal for all coupling schemes in a composition.

**Step To Reproduce**
1. Use the configuration above
2. In B, call `advance(1.0-10e-9)`

**Expected behaviour**
This should result in a configuration error.

uekerman:
Alternative "fix": Move the `valid-digits` somewhere global.

fsimonis:
Conclusion:
- checking valid digits doesn't make sense
- use min time step size instead
- make a global setting

Issue #1776, opened 2023-08-16
End of time window assertion fails when a large number of substeps are used


NiklasKotarsky:
**Describe your setup**

Operating system (e.g. Linux distribution and version): Ubuntu Ubuntu 22.10
preCICE Version: develop f7289a6

**Describe the problem**
If one uses more than 500 equally spaced substeps in the time window then the assertion `PRECICE_ASSERT(math::equals(timesAscending(nTimeSteps - 1), time::Storage::WINDOW_END))` and `PRECICE_ASSERT(math::equals(data->stamples().back().timestamp, time::Storage::WINDOW_END))` are triggered by the numerical accuracy inside BaseCouplingScheme.cpp and SerializedStamples.cpp. 

**Step To Reproduce**
1.The code to reproduce the error is in the repository
 https://github.com/NiklasKotarsky/precice/tree/Err-assertion-end-of-timeWindow
2. The failing test: Integration/Serial/Time/Implicit/SerialCoupling/ManySubstepsReadMapping

**Expected behaviour**
The assertion should not be triggered from only using more than 500 substeps.

**Additional context**
Attach relevant logs and files to this issue.
Format your text (and especially any output/error messages) (see the [GitHub Markdown guide](https://guides.github.com/features/mastering-markdown/)

NiklasKotarsky:
Yes, it seems to me that it is the same root cause. I solved it on the solver side side by adding `currentDt = std::abs(dt-maxDt)<1e-12 ? maxDt : dt;`. Ideally I should be able to use a tolerance of about 1e-14 here or numerical accuracy. The root cause of the issue might be mismatched tolerances in PreCICE. As I understand it in baseCouplingScheme.cpp the end of a time window is calculated with an accuracy of about 1e-10 in line 436 with `const bool isAtWindowEnd = math::equals(getNormalizedWindowTime(), time::Storage::WINDOW_END, _eps);`. The end of the time window is later checked to an accuracy of 1e-14 in both storage.cpp and data.cpp with `PRECICE_ASSERT(math::equals(timesAscending(nTimeSteps - 1), time::Storage::WINDOW_END))`.

uekerman:
also connected to #1712

BenjaminRodenberg:
@NiklasKotarsky fyi: I assigned you here and linked this issue to #1788. I hope that's correct and ok for you.

Issue #1751, opened 2023-07-27
Adaptive solver time steps emit preCICE assertions


davidscn:
**Describe your setup**

I'm using a fixed time-window size and adaptive time-steps within one of the solvers. I'm pretty sure that the origin of the assertion is within the adapter (rounding issues), but it emits one of the following assertions:

I'm using a recent develop snapshot of preCICE: ef76221a3ea9d5b4ba52e4ca524905ba275cb23c

**Describe the problem**

The scenatio triggers certain (developer) assertions in preCICE. I think both of the following assertions are related and should be actual preCICE errors:

```bash
ASSERTION FAILED
Location:   void precice::cplscheme::BaseCouplingScheme::sendData(const PtrM2N&, const DataMap&, bool)
File:       /home/schneidd/precice/src/cplscheme/BaseCouplingScheme.cpp:130
Expression: math::equals(timesAscending(nTimeSteps - 1), time::Storage::WINDOW_END)
Rank:       0
Arguments:  
  0: timesAscending(nTimeSteps - 1) == 0.9999999999999898
  1: nTimeSteps == 3
```

or

```bash
ASSERTION FAILED
Location:   void precice::time::Storage::setSampleAtTime(double, precice::time::Sample)
File:       ~/precice/src/time/Storage.cpp:33
Expression: math::smaller(maxStoredNormalizedDt(), time)
Rank:       0
Arguments:  
  0: maxStoredNormalizedDt() == 1
  1: time == 0.5000000000000001
  2: "Trying to write sample with a time that is too small. Please use clear(), if you want to write new samples to the storage." == Trying to write sample with a time that is too small. Please use clear(), if you want to write new samples to the storage.
```

**Step To Reproduce**
Was created with the flow-over-heated-plate tutorial and the two OpenFOAM participants. Adapter version: tba

**Expected behaviour**
Hit a preCICE error instead of the exception.

BenjaminRodenberg:
Yes. This should raise an error. I think this is again a good case where we would need a test for the "unhappy" execution path that leads to an error. Related to #280

BenjaminRodenberg:
@NiklasKotarsky fyi: Assigned you here and linked to #1788 for a better overview.

Issue #816, opened 2020-07-16
Clearly define and enforce allowed states and transitions in SolverInterfaceImpl


BenjaminRodenberg:
**Please describe the problem you are trying to solve.**

Our user interface currently offers many possible ways how the user can interact with it. Flexibility is, of course, nice, but some of the currently allowed workflows do not have an obvious use-case (e.g. calling `finalize()` without `initialize()`). If we limit the scope of allowed workflows this will help us to limit the number of things that can possibly go wrong and it will help our users to not run into dead-ends.

**Describe the solution you propose.**

`SolverInterfaceImpl` provides different [states](https://github.com/precice/precice/blob/4aa6a302def46bb2fbc85bcd7263d4c08fe16183/src/precice/impl/SolverInterfaceImpl.hpp#L553-L558) that we already use to check for correct transition between states. #806 added some more checks. I would propose to

1. clearly define which states and transitions are allowed. (read/write before initialize? finalize without initialize?...)
2. add `PRECICE_CHECKS` to avoid invalid transitions and (if necessary) update the tests correspondingly.

Note that especially the second point might cause a lot of work. Example: If we decide to forbid `State::Construction -> State::Finalized`, we will also have to update [this test](https://github.com/precice/precice/blob/4aa6a302def46bb2fbc85bcd7263d4c08fe16183/src/precice/tests/SerialTests.cpp#L161-L171).

**Additional context**

Note that we will end up with a more restrictive execution model for preCICE. Before restricting allowed usage of preCICE, we should carefully evaluate the restrictions that we introduce. We don't want to removing actual functionality, but only execution paths without a real use-case.

See also [this figure](https://github.com/precice/precice/pull/806#pullrequestreview-449720594).

uekerman:
Not breaking in my opinion if we don't restrict real use cases.

uekerman:
connected to #225

fsimonis:
This was mostly resolved by the resource cleanup #772

BenjaminRodenberg:
@fsimonis Can we close this issue? If not, let's move it to 3.x.x, because it does not sound critical for 3.0.0.

fsimonis:
This was resolved by #806 and updated in #1350

PR #1788, opened 2023-09-01
Change valid-digits to min-timestep

NiklasKotarsky:
## Main changes of this PR

Changes the tag valid-digits tag to a min-timestep tag. Also ensures that the same tolerance is used to check for the end of the time window in Storage, Bspline and BaseCouplingScheme.

Resolves: #1712,  #1751,  #1776

## Author's checklist

## Reviewers' checklist


BenjaminRodenberg (changes requested):
Nice that this PR resolves so many problems :+1: What's missing in my opinion:

* Can you write a test that reproduces the error you describe in https://github.com/precice/precice/issues/1776? If I understand everything correctly, the test should trigger the assertion mentioned in #1776, if you use develop and it should succeed, if you use this branch. This also helps us to not run into the same problem again in the future.
* Please add corresponding documentation (changelog entry and [porting guide](https://precice.org/couple-your-code-porting-overview.html))

NiklasKotarsky:
Yes, I also have to check that the errors raised in 1751 are actual precice errors with error messages. I should also at some point also add some documentation with the recomended value for min-timestep tag. For time adaptive solvers one would want to exchange the line "currentDt = dt > maxDt ? maxDt : dt" with "currentDt = dt > maxDt-minTimestep ? maxDt : dt" to avoid that the last time window is smaller than minTimestep.

NiklasKotarsky:
I also checked now that the added test ReadWriteScalarDataFirstParticipantHighTolerance fails on develop e256a11 with the same error message as #1776 if the min-timestep tag is exchanged to validDigits=1.

uekerman:
@NiklasKotarsky Status-quo here is still that you trigger me once I should have a look, right? Or should I have a look now already?

NiklasKotarsky:
> @NiklasKotarsky Status-quo here is still that you trigger me once I should have a look, right? Or should I have a look now already?

It might be ready for a first review. It is in a state where the bugs in #1751 and #1776 now emit proper preCICE errors with a recommendation of how to change the solvers to aviod them. I am still trying to figure out where to place the tag min-time-step-size in the config file, since it should only occur once in the whole file. 

The implementation also currently forces all solvers to hit the end of the time window within numerical accuracy. The problem is that if a solver stops before reaching the end of the time window then the Bspline algorithm would need to extrapolate the remaining part of the time window, which is currently not supported. The simple fix would be to only return the last value, but fixing this here feels a little bit out of scope for this pull request.

src/cplscheme/SerialCouplingScheme.hpp
```diff
@@ -30,7 +30,7 @@ class SerialCouplingScheme : public BiCouplingScheme {
  * @param[in] maxTime Simulation time limit, or UNDEFINED_MAX_TIME.
  * @param[in] maxTimeWindows Simulation time windows limit, or UNDEFINED_TIME_WINDOWS.
  * @param[in] timeWindowSize Simulation time window size.
- * @param[in] validDigits valid digits for computation of the remainder of a time window
+ * @param[in] minTimeStepSize Smallest time step size allowed by PreCICE
```
uekerman, line 33:
```suggestion
 * @param[in] minTimeStepSize Minimum time step size.
```

uekerman (changes requested):
Implementation looks good, also the configuration. Appreciated :+1: 
But, the implemented behavior now is different than what we had previously and 100% what we need I am afraid of.

- Previously: Participant can do do smaller time step sizes than eps, but preCICE does never return a next max time step size smaller than eps. If the remainder in the window is smaller than eps, preCICE simply rounds up to the window end (hence the term valid digits we used before).
- Currently in this PR: Participant is simply not allowed to hand over a time step size smaller than eps. I am not sure if this is what we want. The motivation is to prevent participants from doing to small time step sizes (which could lead to stability problems sometimes). The current behavior one could also implement directly in the solver. 

Fixing this should not be too complicated.

- [ ] Could you then please add an integration test that uses a `min-time-step-size` larger than 1e-14?

NiklasKotarsky:
> Implementation looks good, also the configuration. Appreciated 👍 (and sorry for my slow reply) But, the implemented behavior now is different than what we had previously and not 100% what we need I am afraid of.
> 
>     * Previously: Participant can do do smaller time step sizes than eps, but preCICE does never return a next max time step size smaller than eps. If the remainder in the window is smaller than eps, preCICE simply rounds up to the window end (hence the term valid digits we used before).
>
>     * Currently in this PR: Participant is simply not allowed to hand over a time step size smaller than eps. I am not sure if this is what we want. The motivation is to prevent participants from doing to small time step sizes (which could lead to stability problems sometimes). The current behavior one could also implement directly in the solver.

I can change it such that participant can do do smaller time step sizes than eps
 whithout preCICE throwing an error. The reason for removing the rounding up of the time window causes problems with the waveform iteration. The problem boils down to that we either have to set the time in the last sample to the end of the time window manually or we have to allow for extrapolation in the Bsplines interpolation, since the other solver otherwise could sample a time behind the last timestep. I have not yet found a nice way of making the old functionality consistent with the interpolation. There seems to be also something wrong with the Open-foam adapter related to waveform iterations and valid-digits, see  https://github.com/precice/openfoam-adapter/pull/305#issuecomment-1705341375.

I also had a very hard time coming up with a situation where one would need to use eps and not fixing it on the solver side. The only valid use case I could come up with is some solvers use 16-bit floating point arithmetic and some solvers use 32 bit floating point arithmetic.  I have no idea if this is a realistic use case?

uekerman:
> The problem boils down to that we either have to set the time in the last sample to the end of the time window manually

This should be the way to go in my opinion.

> I also had a very hard time coming up with a situation where one would need to use eps and not fixing it on the solver side.

Imagine you have a solver that has issues with too small time step sizes. Say you don't want a dt smaller than 1e-10. Now, it could happen that preCICE tells you that the last time step within a time window needs to be smaller than 1e-12. What do you do then. Of course, you could throw an error on the side of the solver, but that is no solution.
Instead, with this "valid digits / min time step" functionality, we can define in preCICE that the minimum should be 1e-10. Then, preCICE does no tell "do one more with 1e-12", but treats the last one the solver did as the window end already (i.e. constant extrapolation).
Understandable?

fsimonis:
> Then, preCICE does no tell "do one more with 1e-12", but treats the last one the solver did as the window end already (i.e. constant extrapolation).

I don't think that we correctly handle this case in the coupling logic.
The $\epsilon$ derived by the valid-digits is used to check if $|\text{timeWindowSize}-\text{computedTime}| < \epsilon$.

https://github.com/precice/precice/blob/e1e20f9d6cfb64f1965e27a959a7553ceaea7cdc/src/cplscheme/BaseCouplingScheme.cpp#L411-L414


We advance `timeWindowStart` by the `computedTime` instead of `timeWindowSize`:

https://github.com/precice/precice/blob/e1e20f9d6cfb64f1965e27a959a7553ceaea7cdc/src/cplscheme/BaseCouplingScheme.cpp#L343
https://github.com/precice/precice/blob/e1e20f9d6cfb64f1965e27a959a7553ceaea7cdc/src/cplscheme/BaseCouplingScheme.cpp#L360

This leads to a desync of the time across participants.
It also means that samples are not at the end of the time window.

BenjaminRodenberg:
> Imagine you have a solver that has issues with too small time step sizes. Say you don't want a dt smaller than 1e-10. Now, it could happen that preCICE tells you that the last time step within a time window needs to be smaller than 1e-12. What do you do then. Of course, you could throw an error on the side of the solver, but that is no solution.
> Instead, with this "valid digits / min time step" functionality, we can define in preCICE that the minimum should be 1e-10. Then, preCICE does no tell "do one more with 1e-12", but treats the last one the solver did as the window end already (i.e. constant extrapolation).
> Understandable?

Couldn't we have a test representing this case? I think the use-case of the feature is still hard to understand and if I search for `valid-digits` in our code base, there are only two configs actually using this option. One is `PrimaryRankSockets.xml` (does not sound like the main purpose of this test is to test this feature) and the other is `MultiCoupling.xml` (as far as I understand already a bi-coupling scheme should be sufficient).

Writing a test also helps to more clearly define what we actually want here.

NiklasKotarsky:
Nice catch @fsimonis, it should be possible to fix the coupling logic by adjusting the time in BaseCouplingScheme and moving the last sample to the end of the time window. However as far as I under stand it I do not have write access to the samples and stamples directly in baseCouplingScheme, which means I should not be able to move the samples directly. Therefore I would propose to just adding the last sample to the end of the time window as well, which should ensure that we do not have to extrapolate. The concern I have with this solution is that it introduces a time integration error, since preCICE is moving time steps around, which is not apparent or easily understandable from the user side. The proper way to avoid this integration error is to adjust the last time step that the solver takes and force it to hit the end of the time window  exactly meaning to make the last timestep 10e-12 bigger, but I can also see the argument that this feature would be a nice convenience when that is difficult to achieve.

uekerman:
> Couldn't we have a test representing this case?

Yes, that's what I am suggesting above.

> The proper way to avoid this integration error is to adjust the last time step that the solver takes and force it to hit the end of the time window exactly meaning to make the last timestep 10e-12 bigger,

We cannot do this as this is already in the past. The solver only tells preCICE about the time step size it used **after** it did the time step.

NiklasKotarsky:
I have implemented a test case now documenting the effects of a large min time step on the waveform iteration https://github.com/precice/precice/pull/1788/files#diff-0a12f7d580510ad191bb2b2d985a1ddc12bae9ef5f7b87c6187afeb6b1cd5164. I am still struggling a bit with accuracy of the interpolation, since it seems to me that the interpolation no longer conserves polynomials, but this might be a coding error in the feature.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -724,7 +724,20 @@ void BaseCouplingScheme::advanceTXTWriters()
...
+      auto   timeStepStorage = data->timeStepsStorage();
+      auto   values          = timeStepStorage.getTimesAndValues();
+      auto   ts              = values.first;
+      double endOfTimeWindow = getTime() - _computedTimeWindowPart + _timeWindowSize;
+      if (ts(ts.size() - 1) < endOfTimeWindow - math::NUMERICAL_ZERO_DIFFERENCE) {
+        data->setSampleAtTime(getTime() - _computedTimeWindowPart + _timeWindowSize, timeStepStorage.getSampleAtOrAfter(ts(ts.size() - 1)));
+      }
+      _computedTimeWindowPart = _timeWindowSize;
+    }
+  }
```
uekerman, line 738: A non-ideal place for this logic as it is in the function "did I reach the end?", where one would not expect such an alteration of the state. Better to put this in a separate function and call at the right place.

@BenjaminRodenberg Could you please double-check the logic here?

uekerman (approved):
Last comment, otherwise good to go from my side.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -724,7 +724,20 @@ void BaseCouplingScheme::advanceTXTWriters()
 
 bool BaseCouplingScheme::reachedEndOfTimeWindow()
 {
-  return math::equals(getNextTimeStepMaxSize(), 0.0, _eps) || not hasTimeWindowSize();
+  if (math::equals(getNextTimeStepMaxSize(), 0.0, _minTimeStepSize) && getNextTimeStepMaxSize() > math::NUMERICAL_ZERO_DIFFERENCE && hasTimeWindowSize()) {
+    for (auto &data : _allData | boost::adaptors::map_values) {
+      auto   timeStepStorage = data->timeStepsStorage();
+      auto   values          = timeStepStorage.getTimesAndValues();
```
BenjaminRodenberg, line 730: Please use `Storage::getTimes()` here instead to improve readability.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -724,7 +724,20 @@ void BaseCouplingScheme::advanceTXTWriters()
...
+      auto   ts              = values.first;
+      double endOfTimeWindow = getTime() - _computedTimeWindowPart + _timeWindowSize;
+      if (ts(ts.size() - 1) < endOfTimeWindow - math::NUMERICAL_ZERO_DIFFERENCE) {
```
BenjaminRodenberg, line 733: Instead of `ts(ts.size() - 1)` something like `lastT` would be much more readable. I thing you could even use `Storage::maxStoredTime()` here.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -724,7 +724,20 @@ void BaseCouplingScheme::advanceTXTWriters()
 
 bool BaseCouplingScheme::reachedEndOfTimeWindow()
 {
-  return math::equals(getNextTimeStepMaxSize(), 0.0, _eps) || not hasTimeWindowSize();
+  if (math::equals(getNextTimeStepMaxSize(), 0.0, _minTimeStepSize) && getNextTimeStepMaxSize() > math::NUMERICAL_ZERO_DIFFERENCE && hasTimeWindowSize()) {
+    for (auto &data : _allData | boost::adaptors::map_values) {
+      auto   timeStepStorage = data->timeStepsStorage();
+      auto   values          = timeStepStorage.getTimesAndValues();
+      auto   ts              = values.first;
+      double endOfTimeWindow = getTime() - _computedTimeWindowPart + _timeWindowSize;
```
BenjaminRodenberg, line 732:
```suggestion
      double endOfTimeWindow = getEndOfTimeWindow();
```

Please implement a dedicated function here. This is hard to understand otherwise (and you are also duplicating this statement below). Did something similar in #1822, but this is not merged yet. See `BaseCouplingScheme::getWindowEndTime()`.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -724,7 +724,20 @@ void BaseCouplingScheme::advanceTXTWriters()
...
+      if (ts(ts.size() - 1) < endOfTimeWindow - math::NUMERICAL_ZERO_DIFFERENCE) {
+        data->setSampleAtTime(getTime() - _computedTimeWindowPart + _timeWindowSize, timeStepStorage.getSampleAtOrAfter(ts(ts.size() - 1)));
```
BenjaminRodenberg, line 734: Is this the sample you are duplicating? I think a comment would be helpful. Why do you duplicate the sample? Why not shift the timestamp of the sample at `ts(ts.size() - 1)` to `endOfTimeWindow` instead? (This is just an alternative, but if I think about higher order interpolants maybe a valid one? No need to implement this alternative, but a comment would help)

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: > A non-ideal place for this logic as it is in the function "did I reach the end?", where one would not expect such an alteration of the state. Better to put this in a separate function and call at the right place.

Yes. Additionally, I would also claim that this function should be a `const` function. The way it is implemented currently this is not the case anymore.
 
> @BenjaminRodenberg Could you please double-check the logic here?

Logic is quite straight-forward. Looks ok. But I think the readability could be improved a lot and then this would also shrink down to a few lines of code. I added some comments.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -724,7 +724,20 @@ void BaseCouplingScheme::advanceTXTWriters()
 
 bool BaseCouplingScheme::reachedEndOfTimeWindow()
```
BenjaminRodenberg, line 725:
```suggestion
bool BaseCouplingScheme::reachedEndOfTimeWindow() const
```
Actually, this should be a const function. I would suggest to do all the non-const stuff that modifies `_computedTimeWindowPart` or `data` via `data->setSampleAtTime` at a different place.

BenjaminRodenberg (changes requested):

BenjaminRodenberg (commented):
Maybe another comment leading into the direction of a possible implementation:

What's generally happening inside `reachedEndOfTimeWindow()` looks reasonable, but I think it should be shifted somewhere else for better readability. How about the following place:

```
void BaseCouplingScheme::firstExchange()
{
  PRECICE_TRACE(_timeWindows, getTime());
  checkCompletenessRequiredActions();
  PRECICE_ASSERT(_isInitialized, "Before calling advance() coupling scheme has to be initialized via initialize().");
  if(math::smallerEquals(getNextTimeStepMaxSize(), _minTimeStepSize) {  // I think this is a more explicit way of saying "Is the next required time step size smaller than what is actually requested by the user?"
    // Then we have to do something about it. Basically, what's currently implemented inside reachedEndOfTimeWindow()
  }
  _hasDataBeenReceived  = false;
  _isTimeWindowComplete = false;
  PRECICE_ASSERT(_couplingMode != Undefined);
  if (reachedEndOfTimeWindow()) {  // TODO: do all the const things here
    _timeWindows += 1; // increment window counter. If not converged, will be decremented again later.
    exchangeFirstData();
  }
}
```

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: Sorry for making things more complicated, but I think the `const` was really missing in the first place. I directly added it on `develop`. See 6ee29277f892e6d08a514fa734f1274aa8b59cea. Please don't forget to merge again.

src/cplscheme/BaseCouplingScheme.cpp
NiklasKotarsky: I did not find a way to return a non const reference to the last sample, which means that I should not be able to modify the last sample directly if I have understood the cpp syntax correctly. Otherwise I think that moving the sample would have been a more elegant solution.

BenjaminRodenberg: Correct. There is currently no possibility to update samples or their timestamp. Like I said: Just an alternative idea and if there is no obvious benefit, let's go for the "duplication solution" as this is possible with what we already have.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -323,6 +328,11 @@ void BaseCouplingScheme::secondExchange()
 
   if (reachedEndOfTimeWindow()) {
 
+    // If preCICE has stopped before the end of the time window we have to duplicate the last available sample and put it at the end of the time window.
+    if (getNextTimeStepMaxSize() > math::NUMERICAL_ZERO_DIFFERENCE && hasTimeWindowSize()) {
+      addTimeStepAtWindowEnd();
+    }
```
BenjaminRodenberg, line 334:
```suggestion
```
I don't think you need to do this twice. Doing this once (in `firstExchange` or `secondExchange`) should be sufficient.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -302,6 +302,11 @@ void BaseCouplingScheme::firstExchange()
 
     _timeWindows += 1; // increment window counter. If not converged, will be decremented again later.
 
+    //If preCICE has stopped before the end of the time window we have to duplicate the last available sample and put it at the end of the time window.
+    if (getNextTimeStepMaxSize() > math::NUMERICAL_ZERO_DIFFERENCE && hasTimeWindowSize()) {
+      addTimeStepAtWindowEnd();
+    }
```
BenjaminRodenberg, line 308:
```suggestion
    //If preCICE has stopped before the end of the time window we have to duplicate the last available sample and put it at the end of the time window.
    if (getNextTimeStepMaxSize() > math::NUMERICAL_ZERO_DIFFERENCE) {
      addTimeStepAtWindowEnd();
    }
```
Is this additional condition really needed? Why?

Besides that: Readability has improved a lot :+1:

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -813,4 +823,25 @@ void BaseCouplingScheme::receiveConvergence(const m2n::PtrM2N &m2n)
...
+  for (auto &data : _allData | boost::adaptors::map_values) {
+    auto   timeStepStorage = data->timeStepsStorage();
+    double lastSampleT     = timeStepStorage.maxStoredTime();
+    double endOfTimeWindow = getWindowEndTime();
+    //If there does not exist a sample at the window end duplicate the last available sample and put it at the end of the time window.
+    if (lastSampleT < endOfTimeWindow - math::NUMERICAL_ZERO_DIFFERENCE) {
+      data->setSampleAtTime(endOfTimeWindow, timeStepStorage.getSampleAtOrAfter(lastSampleT));
+    }
+    //This keeps the time within preCICE synchronized
+    _computedTimeWindowPart = _timeWindowSize;
```
BenjaminRodenberg, line 838: This line should be outside of the for-loop (does not make sense to call it several times). Why not even put it outside of `BaseCouplingScheme::addTimeStepAtWindowEnd`? Does not really belong here in my opinion.

src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -560,6 +560,13 @@ class BaseCouplingScheme : public CouplingScheme {
    * @return true, if any CouplingData in dataMap requires initialization
    */
   bool anyDataRequiresInitialization(DataMap &dataMap) const;
+
+  /**
+   * @brief Goes through _allData and duplicate the last available sample and puts it at the end of the time window if there does not exist a sample at the window end.
+   */
+  void addTimeStepAtWindowEnd();
+
+  double getWindowEndTime() const;
```
BenjaminRodenberg, line 569: Please add some documentation for this function.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -409,7 +419,7 @@ bool BaseCouplingScheme::addComputedTime(
                 timeToAdd, _timeWindowSize - _computedTimeWindowPart + timeToAdd);
 
   if (hasTimeWindowSize()) {
```
BenjaminRodenberg, line 421:
```suggestion
  //@todo: if-else branch is to some degree a duplicate of BaseCouplingScheme::reachedEndOfTimeWindow. Try to remove duplicate code.
  if (hasTimeWindowSize()) {
```
Let's take care of this after directly on develop and not in this PR, because it's not related. But please add the todo in the meantime that we don't forget it.

BenjaminRodenberg (changes requested):
Some minor todos are left, but this should be the last iteration.

src/cplscheme/BaseCouplingScheme.cpp
NiklasKotarsky: Yes thanks, I had misunderstood what first and second exchange do.

src/cplscheme/BaseCouplingScheme.cpp
NiklasKotarsky: hasTimeWindowSize() is needed there, since if the participant does not have a time window size we can't put a sample at the end of the time window. I will add a comment clarifying this.

BenjaminRodenberg: True. Did not think about this point :+1:

fsimonis:
I have a strong feeling that we closed the discussion of #1712 prematurely in July, without a strong grasp of the full picture.
I realize that this is _very_ late in the process of this PR, but I would like to ask a fundamental question before we continue with this PR.

@uekerman @BenjaminRodenberg I got a mix of different signals here from what we want and need, so I would appreciate some feedback from you.

@NiklasKotarsky sorry if this back and forth is a bit frustrating. :sweat_smile: 

### Is all of this necessary?

Our API provides the exact maximum time step size via `getMaxTimeStepSize()` to the adapter.
Inside the rest of the time window, it can step as it pleases. There are no limitations imposed from other solvers.
If the adapter knows of a minimum viable time step size, then it can use that information in choosing an appropriate time step and if it is unknown, then it can also not be configured.
Hence, the solver/adapter has __all__ required information to decide on an appropriate time step sizes for the rest of the time window.

If it decides to make a step of `getMaxTimeStepSize() - x`, then it needs to deal with the fact that the next time step can be at most `x`.
I don't see how preCICE can provide any meaningful help here. Documentation seems to be key here.

This reduces the use-cases of the global "minimum time step size" to adjusting the numerical tolerance when checking if a time window end has been reached, which (to my understanding) was the original point of the `valid-digits` attribute.
The user may need to adjust the absolute tolerance of a time step aligning with the end of the time window, so why don't we just call it `time-window-end-atol="1e-16"`. (Yes, this is essentially `min-time-step-size` with a different coat of paint.)

This then gets rid of problems where the sample needs to go. We put it at the end of the time window, which may be shifted at most by `time-window-end-atol` and should hopefully be negligible. It can also be clearly documented.

### Simpler suggestion

Wouldn't it be simpler to change the behaviour of preCICE to the following:

We define a global absolute tolerance for reaching the end of a time window
```xml
<precice-configuration time-window-end-atol="1e-13">
```

In normal coupling schemes, we advance to the exact time window end instead of adding timeToAdd. This ensures that all time windows actually are at the time stamps $\Delta t_n=n * \text{timeWindowSize}, n \in (1,2,3,\ldots)$.

```cpp
bool BaseCouplingScheme::addComputedTime(double timeToAdd)
{
  if (!hasTimeWindowSize()) {
    // add time interval that has been computed in the solver to get the correct time remainder
    _computedTimeWindowPart += timeToAdd;
    return true;
  }

  const double maxTimeStepSize = getNextTimeStepMaxSize();

  // Check validness
  bool valid = math::smallerEquals(timeToAdd, maxTimeStepSize, _timeWindowEndAtol);
  PRECICE_CHECK(valid, ...);

  bool atTimeWindowEnd = math::equals(timeToAdd, maxTimeStepSize, _timeWindowEndAtol);
  if (atTimeWindowEnd) {
    // Step to the exact end of the time window
    _computedTimeWindowPart = _timeWindowSize;
  } else {
    // Subcycling
    _computedTimeWindowPart += timeToAdd;
  }
  return atTimeWindowEnd;
}
```

For compositional schemes, we get the remaining time step to the next time window end $\delta t_{max} = min_{c \in \text{activeCouplingSchemes}}(\text{c.getNextTimeStepMaxSize()})$ and check if $\text{timeToAdd} \approx \delta t_{max}$. If that is the case, then we use $\delta t_{max}$ as the next `timeToAdd`, otherwise we advance by `timeToAdd`.
This keeps all coupling schemes in sync.

```cpp
bool CompositionalCouplingScheme::addComputedTime(double timeToAdd)
{
  const double maxTimeStepSize = getNextTimeStepMaxSize(); // min over all active schemes

  bool reachesATimeWindowEnd = math::equals(timeToAdd, maxTimeStepSize, _timeWindowEndAtol);
  if (reachesATimeWindowEnd) {
    // align schemes with the next time window
    // call addComputedTime(maxTimeStepSize) for each active scheme
  } else {
    // Subcycling
    // call addComputedTime(timetoAdd) for each active scheme
  }
  return reachesATimeWindowEnd;
}
```

Then the samples will be always at the end of a time window. Meaning that their time stamp will be shifted at most by `time-window-end-atol`, which is something we can document directly in the attribute.

### Slightly off-topic: `min-time-window-size`

The only edge case where I could see a minimal time window size, is the prescription of time windows using `<time-window-size method="first-participant"/>` in serial coupling schemes. This directly limits the maximum time step size of the `second` participant. Here it may be useful to allow specifying a required minimum in the configuration file `<time-window-size method="first-participant" minimum="1e-7"/>` for cases where the second participant requires at least a given time window size to deal with. This would be an (easy to implement) error if the "first" solver calls `advance(1e-9)`.

BenjaminRodenberg:
> This then gets rid of problems where the sample needs to go. We put it at the end of the time window, which may be shifted at most by `time-window-end-atol` and should **hopefully** be negligible. It can also be clearly documented.

I think this is at the moment the important bit which should also decide whether we want to invest more time and effort in this PR (more reviewing, more refactoring...) or just merge (maybe with some imperfections).

Is the fact that we potentially duplicate the sample really a problem? I think we just don't know at the moment. The only piece of evidence that we have is that it seems to work for Niklas' use case. Before there was some problem according to #1776. This not complete, but still significantly more testing of this feature than we did before. For me this is good enough.

If we invest more time and effort to come up with a different strategy is it really better? Hopefully we find something better, but again we don't know. If we want to be sure, we need to come up with some test cases that are sensitive to the potential problems we are making theories about. From my perspective it's not really worth the effort and I again would argue that if the feature works for Niklas' use case it is good enough.

We might need to change things in the future, of course. Two possible reasons:

* If more use cases that are not covered by the current implementation pop up in the future.
* There are integration problems of this feature and the compositional coupling scheme (I cannot judge whether there is already a concrete problem today or not).

uekerman:
I would also vote for merging here if there are no concrete problems with the compositional coupling scheme. Rest seems high effort low gain in my opinion. In any case, `min-time-step-size` will not be a frequently used feature.

fsimonis:
> Is the fact that we potentially duplicate the sample really a problem? I think we just don't know at the moment.

Sure, the impact on the interpolant is theoretical at this point.
I added sample duplication with a shift of `1e-13` to the BSpline tests as a starting ground to test this. You can find the code at #1831. 

What's very real are the following factors:
* User transparency: It is not clear why there is suddenly one additional sample than time step in the output (and eventually export/ASTE). There is now a possible disconnect between `writeData` and (write-data) samples.
* Mapping cost: Additional samples require more mapping evaluations, which can be trivial (NN or very expensive (global RBG)
* Storage and communication: Additional samples require more storage and need to be communicated.

> If we invest more time and effort to come up with a different strategy is it really better?

The problem is that the handling of the samples is a non-trivial solution to a problem, which we don't know if it exists, to fix a use-case we don't know even matters.

> I would also vote for merging here if there are no concrete problems with the compositional coupling scheme.

Given a participant who writes data to a mesh, and this data is exchange-sent to 2 other participants via 2 coupling schemes with equal time-window-sizes.
If this participant makes a time step that falls into this above edge case, then the two coupling scheme tries to duplicate the last data sample.

Handling data in the coupling schemes is always tricky.

NiklasKotarsky:
> > Is the fact that we potentially duplicate the sample really a problem? I think we just don't know at the moment.
> 
> Sure, the impact on the interpolant is theoretical at this point. I added sample duplication with a shift of `1e-13` to the BSpline tests as a starting ground to test this. You can find the code at #1831.
> 
>     * User transparency: It is not clear why there is suddenly one additional sample than time step in the output (and eventually export/ASTE). There is now a possible disconnect between `writeData` and (write-data) samples.
 
>     * Mapping cost: Additional samples require more mapping evaluations, which can be trivial (NN or very expensive (global RBG)
> 
>     * Storage and communication: Additional samples require more storage and need to be communicated.


Yes, I am planning to write something for the homepage to explain how this feature works and a recommendation of how to use it. No matter which implementation we end up in we will introduce a coupling error that has to be clearly communicated to the user, since the solvers now can stop before the end of the time window. We also need to clearly explain the fix for this from the solver side without introducing a coupling error. However, min-time-step feature could be useful for removing rounding errors that in the end do not effect the simulation much. 

 The reason for choosing to duplicate samples is a mostly a practical one, the current api from the waveform iterations does not give us write access to the samples from the coupling class. Changing the api to allow the coupling to directly change the samples to support a little bit of a niche feature does not feel like a worth while tradeoff in my opinion.
 
> Given a participant who writes data to a mesh, and this data is exchange-sent to 2 other participants via 2 coupling schemes with equal time-window-sizes. If this participant makes a time step that falls into this above edge case, then the two coupling scheme tries to duplicate the last data sample.

 If I have done my coding correctly preCICE should only duplicate the sample once per data object, meaning that as long as the underlying data object is the same in both cases the sender should only duplicate the last sample once. The recievers should not be able to duplicate the last sample, since the sender already has duplicated the last sample in the sent data object.

BenjaminRodenberg (approved):
Only a minor todo is left. Feel free to merge when this is done. (Please don't forget to squash & merge)

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -808,4 +819,23 @@ void BaseCouplingScheme::receiveConvergence(const m2n::PtrM2N &m2n)
...
+{
+
+  for (auto &data : _allData | boost::adaptors::map_values) {
+    auto   timeStepStorage = data->timeStepsStorage();
+    double lastSampleT     = timeStepStorage.maxStoredTime();
+    double endOfTimeWindow = getWindowEndTime();
+    //If there does not exist a sample at the window end duplicate the last available sample and put it at the end of the time window.
+    if (lastSampleT < endOfTimeWindow - math::NUMERICAL_ZERO_DIFFERENCE) {
+      data->setSampleAtTime(endOfTimeWindow, timeStepStorage.getSampleAtOrAfter(lastSampleT));
+    }
```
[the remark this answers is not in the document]
NiklasKotarsky: Yes, we might also want to continue discussing the effect on the interpolation as well.

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -197,29 +196,22 @@ void CouplingSchemeConfiguration::xmlTagCallback(
...
+                    "Please check the <time-window-size value=\"{}\" method=\"{}\" /> "
                     "tag in the <coupling-scheme:...> of your precice-config.xml",
-                    _config.timeWindowSize, _config.validDigits, tag.getStringAttributeValue(ATTR_METHOD));
+                    _config.timeWindowSize, tag.getStringAttributeValue(ATTR_METHOD));
     }
-    PRECICE_CHECK((_config.validDigits >= 1) && (_config.validDigits < 17),
-                  "Valid digits of time window size has to be between 1 and 16. "
-                  "Please check the <time-window-size value=\"{}\" valid-digits=\"{}\" method=\"{}\" /> tag "
-                  "in the <coupling-scheme:...> of your precice-config.xml",
-                  _config.timeWindowSize, _config.validDigits, tag.getStringAttributeValue(ATTR_METHOD));
```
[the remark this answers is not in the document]
NiklasKotarsky: Thanks for catching this, I have updated the check for valid timeWindowSize such that it now checks minTimeStepSize < timeWindowSize.

fsimonis:
There is still a design issue that will bite us sooner or later:

We adjust the computed time inside `firstExchange()`, meaning that at least `getTime()` and `getNextTimeStepMaxSize()` are easy to use incorrectly between `addComputedTime()` and `firstExchange()`.

This means that we have even more state that changes across functions of our most complicated class hierarchy.
I strongly suggest keeping _at least the time_ consistent.

Also, we can handle the sample duplication in the `ParticipantImpl`, which would be in line with our endless efforts to get data-modifying logic out of the CouplingSchemes.  

Suggestion:
* In `ParticipantImpl::advance(computedTimeStepSize)`
  1. At the beginning, we get the unmodified time from the coupling scheme
  3. we add the computed time to the couplingscheme
  4. we write samples to the end of the time step given by the user: unmodified time + computedTimeStepSize
  5. if we reached the end of the time window:
    * we get the current time of the couplingscheme.
    * if it differs from the user-time-step-end, then we write another sample to the timewindow end.
  6. We execute write mappings and apply actions to added samples 
* In the coupling schemes:
  * in `addComputedTime()`: advance computedTime by either the given time step, or set it to the end of the time window
  * use `math::NUMERICAL_ZERO_DIFFERENCE` to compare timesteps instead of the `min-time-step-size`
  * We don't duplicate samples, but we can assert that an sample exists at the end of the time window.

That results in a way clearer flow inside the coupling scheme and keeps sample alterations in the ParticipantImpl.

Furthermore, the attribute `min-time-step-size` needs a more thorough explanation of what it's implication is. Note that as part of `<precice-configuration>` it is not hidden. It will show up at the very top of the XML reference and users will read its docs and inevitably use it.

The following is still missing:
* it prevents advancing by a smaller amount via advance.
* it extends time-steps to the end of the time-window, if the remaining time step would be smaller (which is not obvious),
* in which case, samples will be duplicated. The last sample will be at the end of the time window instead of stepped-to time. Only the duplicate is visible on the reading side, when using `substeps=off` (also not obvious).
* in case of multiple coupling schemes with mixed time window sizes, the duplication may occur at every time-window end of a participant, not coupling scheme.

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -197,29 +196,22 @@ void CouplingSchemeConfiguration::xmlTagCallback(
...
-    PRECICE_CHECK((_config.validDigits >= 1) && (_config.validDigits < 17), "Valid digits of time window size has to be between 1 and 16.");
     // Attribute does not exist for parallel coupling schemes as it is always fixed.
     _config.dtMethod = getTimesteppingMethod(tag.getStringAttributeValue(ATTR_METHOD, VALUE_FIXED));
     if (_config.dtMethod == constants::TimesteppingMethod::FIXED_TIME_WINDOW_SIZE) {
-      PRECICE_CHECK(_config.timeWindowSize > 0,
-                    "Time window size has to be larger than zero. "
-                    "Please check the <time-window-size value=\"{}\" valid-digits=\"{}\" method=\"{}\" /> tag "
-                    "in the <coupling-scheme:...> of your precice-config.xml",
-                    _config.timeWindowSize, _config.validDigits, tag.getStringAttributeValue(ATTR_METHOD));
+      PRECICE_CHECK(_config.timeWindowSize > _minTimeStepSize,
```
BenjaminRodenberg, line 202:
```suggestion
      PRECICE_CHECK(math::greaterEquals(_config.timeWindowSize, _minTimeStepSize),
```
Should be >=. A `_minTimeStepSize` that is identical to the `timeWindowSize` is on the borderline, but should be fine.

BenjaminRodenberg: `_minTimeStepSize` is guaranteed to be greater zero? Might still not harm to put a `PRECICE_ASSERT(math::greater(_minTimeStepSize, 0))` here to make this assumption explicit. I'm so picky here, because previously it was `PRECICE_CHECK(_config.timeWindowSize > 0, ...)` and we don't want to run into the corner case that `timeWindowSize = 0` is allowed.

NiklasKotarsky: Yes, _minTimeStepSize has to be greater than 0, since we otherwise run into troubles. I would also avoid using math::greaterEquals here, since they contain a tolerance of NUMERICAL_ZERO_DIFFERENCE which might be troublesome.

NiklasKotarsky:
I have updated the documentation of min-time-step-size to better reflect what the method does and link to the the issue #1832, which should contain all side effects of this implementation. Technically speaking it should be relative straight forward to move 'addTimeStepAtWindowEnd()' to the 'advance(double computedTimeStepSize)' funtion in ParticipantImpl.cpp, but I have to ensure that this does not break something else. Therefore I would suggest merging here and doing this refactoring later when it is more clear if this is a useful feature or if one should change it or even remove it entirely.

fsimonis:
Regarding https://github.com/precice/precice/pull/1788#issuecomment-1750736101

The problem is not that the CouplingScheme controls the time, it's how the time is controlled, when, and why.
Having periods of incorrect function returns essentially means that we need something like #816 but for coupling schemes.

I still strongly suggest to keep time consistent. There is also no reason not to return more information in `addComputedTime`. Example:

```cpp
struct TimeStepInfo {
  double userTimeStep;
  double adjustedTimeStep;
  bool   hasBeenAdjusted;
  bool   atEndOfTimeWindow;
};
```

> Therefore I would suggest fixing the documentation and then merging here and doing this refactoring later when it is more clear if this is a useful feature or if one should change it or even remove it entirely.

No matter the decision, we need to clean this up before the release of v3 along the remaining coupling-related topics.

fsimonis:
I would also like to point out that no one has answered the question "Is this even necessary?"  in my previous comment https://github.com/precice/precice/pull/1788#issuecomment-1746676864

NiklasKotarsky:
> I would also like to point out that no one has answered the question "Is this even necessary?" 

The only use case which I can come up with where it is necessary would be mixed precision, where one solver has a lower precision than the other. Other than that this would be more of a nice to have feature when the time step is difficult to change or if the solver itself does not really have a time step. For example a controller which has a fixed time step that interfaces to some FSI simulation. 

>I still strongly suggest to keep time consistent
>We adjust the computed time inside firstExchange(), meaning that at least getTime() and getNextTimeStepMaxSize() are easy to use incorrectly between addComputedTime() and firstExchange().

If I understand this correctly then we have to modify the time in addComputedTime() in order to keep the time consistent? 

I am a little bit unsure of how to proceed here. I could spend some time to move everything to advance in ParticipantIMPL and addComputedTime in order to make it more robust. The tricky part for me is that this feature will have unintended consequences especially for multicoupling with different time window lengths.

fsimonis:
> Other than that this would be more of a nice to have feature when the time step is difficult to change or if the solver itself does not really have a time step. For example a controller which has a fixed time step that interfaces to some FSI simulation.

&

> The whole discussion goes a bit into a fundamental direction: "What is black-box coupling?" Typically, solvers have methods like solverDt = computeTimeStepSize(). We don't know what the method does exactly. Underlying might be a hard stability criterion. Meaning, the solver cannot simply do a time step with solverDt + eps.

If it is a hard criterion or the time step cannot be changed, then why wouldn't the solver prescribe the time window size using a serial scheme?

The current implementation artificially extends the time step, which is completely intransparent for the adapter and coupled solver.
preCICE only tells the solver how long the next time step can be, there is no way of the solver finding out what timestep preCICE actually took. This leads to a shift in the time points of the solver and preCICE.

Let's say that this repeats every time window, then this time shift can build up.
Is this a problem? Do we need some way of telling this to the adapter?

NiklasKotarsky:
> If it is a hard criterion or the time step cannot be changed, then why wouldn't the solver prescribe the time window size using a serial scheme?

Using participant first might be a good idea for some systems and setups, but it only works in serial coupling. Meaning that we need to provide something else for the parallel and multicoupling case. The min-time-step tag is more meant to be used to avoid round up errors that could arise when doing integer time steps. Or as a quick fix when the time step in the solver is difficult to modify and the solver. 

> Let's say that this repeats every time window, then this time shift can build up. Is this a problem? Do we need some way of telling this to the adapter?

We are assuming that deltaCorrection << deltaT. Using this feature we are going to make an additional time integration error that scales with deltaCorrection, which in most cases will be smaller than the time integration error. Therefore we should be able to just ignore this error completely on both the user and coupling side. 

There might be a case for an api function in preCICE that returns the integrated time, since we might run into different round of errors in the solver and preCICE. This would also cover the problems this feature might cause for the solver, but this is out of scope for the current pull request.

fsimonis:
After the extensive discussion here, I think the rationale for the implementation is sufficiently documented.

@NiklasKotarsky are there any more changes you would like to do?
If not, then I would volunteer to squash and merge this PR.

NiklasKotarsky:
> After the extensive discussion here, I think the rationale for the implementation is sufficiently documented.
> 
> @NiklasKotarsky are there any more changes you would like to do? If not, then I would volunteer to squash and merge this PR.

No, this is ready to be merged.

fsimonis:
I'll deal with the conflicts and merge the PR.

Issue #1658, opened 2023-05-14
Resampling im time storage for non-matching meshes in time


BenjaminRodenberg:
**Please describe the problem you are trying to solve.**

In #1612 we assume that all time meshes used by actions are compatible. This assumption will break as soon as for a multi coupling scheme subcycling is used. Imagine a summation action that sums up DataA and DataB originating from SolverA and SolverB where A uses a time step size of 0.5 and B a time step size of 1/3. We cannot simply add DataA + DataB for all times here.

**Describe the solution you propose.**

A possible solution is to to a resampling of the input data onto a common mesh. If we stay with the given example of time step sizes 0.5 and 1/3, resulting in time steps at [0, 0.5, 1] and [0, 1/3, 2/3, 1] for a time window of size 1, we could use the mesh [0, 1/3, 0.5, 2/3, 1] and use strategies from timeinterpolation (BSplines) to create the additional samples.

**Describe alternatives you've considered**

Alternatively one could try to implement the actions in a more "functional" way. Meaning: We do not compute the sum of A and B inside advance. Instead, we follow a similar approach like with `readData`: We only evaluate DataA(t) + DataB(t) when the user asks for (DataA + DataB)(t). I already discussed this idea with @fsimonis.

uekerman:
Quick remark: I don't think that this is a problem specific to multi coupling. Should be more general.

PR #1794, opened 2023-09-04
Fix `acceleration:constant` for waveforms

BenjaminRodenberg:
## Main changes of this PR

* Provides proper implementation of constant underrelaxation when using subcycling.

Similar to mapping and actions, we need also a proper implementation for acceleration that supports waveforms (see also https://github.com/precice/precice/issues/1645).

Some todos:

- [x] Plan this for 3.0.0 or 3.x.x ? Technically the changes are not breaking. However, performance of Quasi-Newton acceleration will currently suffer if waveform relaxation is used. We still need to do some testing here.
- [x] Need to take care of previous iterations in an appropriate way (already deal with time grid changes from one iteration to the next here? Related to https://github.com/precice/precice/issues/1658)
- [x] Implement tests for all acceleration schemes (decided to only start with `acceleration:constant` for now)
- [x] Which variants of Quasi-Newton schemes do we want to implement for waveform relaxation? (decided to only start with `acceleration:constant` for now)

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/acceleration/Acceleration.cpp
+++ b/src/acceleration/Acceleration.cpp
@@ -17,16 +17,22 @@ void Acceleration::checkDataIDs(const DataMap &cplData) const
 void Acceleration::applyRelaxation(double omega, const DataMap &cplData) const
 {
   for (const DataMap::value_type &pair : cplData) {
-    const auto  couplingData = pair.second;
-    auto &      values       = couplingData->values();
-    const auto &oldValues    = couplingData->previousIteration();
-    values *= omega;
-    values += oldValues * (1 - omega);
-    if (couplingData->hasGradient()) {
-      auto &      gradients    = couplingData->gradients();
-      const auto &oldGradients = couplingData->previousIterationGradients();
-      gradients *= omega;
-      gradients += oldGradients * (1 - omega);
+    const auto couplingData = pair.second;
+
+    for (auto &stample : couplingData->stamples()) {
+      auto values            = stample.sample.values;
+      auto oldValues         = couplingData->getPreviousValuesAtTime(stample.timestamp); // IMPORTANT DETAIL: The interpolation that we use for resampling does not necessarily have to be the same interpolation as the interpolation the user accesses via read-data. (But probably it is easier to just use the same)
+      couplingData->values() = values * omega;
+      couplingData->values() += oldValues * (1 - omega);
+
+      if (couplingData->hasGradient()) {
+        auto       gradients      = stample.sample.gradients;
+        const auto oldGradients   = couplingData->getPreviousGradientsAtTime(stample.timestamp); // IMPORTANT DETAIL: The interpolation that we use for resampling does not necessarily have to be the same interpolation as the interpolation the user accesses via read-data. (But probably it is easier to just use the same)
+        couplingData->gradients() = gradients * omega;
+        couplingData->gradients() += oldGradients * (1 - omega);
+      }
+
+      couplingData->setSampleAtTime(stample.timestamp, couplingData->sample());
     }
   }
 }
--- a/src/cplscheme/CouplingData.cpp
+++ b/src/cplscheme/CouplingData.cpp
@@ -20,7 +20,7 @@ CouplingData::CouplingData(
       _timeStepsStoragePrevious()
 {
   PRECICE_ASSERT(_data != nullptr);
-  _timeStepsStoragePrevious.setInterpolationOrder(3); // @todo hard-coded for now, but we need to somehow link this to <read-data waveform-order="ORDER" />
+  _timeStepsStoragePrevious.setInterpolationDegree(3); // @todo hard-coded for now, but we need to somehow link this to <read-data waveform-order="ORDER" />
   _timeStepsStoragePrevious.setSampleAtTime(time::Storage::WINDOW_START, time::Sample{getDimensions(), Eigen::VectorXd::Zero(getSize())});
   _timeStepsStoragePrevious.setSampleAtTime(time::Storage::WINDOW_END, time::Sample{getDimensions(), Eigen::VectorXd::Zero(getSize())});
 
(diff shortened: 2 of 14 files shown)


NiklasKotarsky (commented):
I think this could be merged with develop now. The only remaining issue is that the _previousIteration always uses 3 order interpolation, which in theory could cause unexpected results for 1 and 2 order interpolation.

src/cplscheme/CouplingData.cpp
```diff
@@ -16,10 +16,14 @@ CouplingData::CouplingData(
     : requiresInitialization(requiresInitialization),
       _mesh(std::move(mesh)),
       _data(std::move(data)),
-      _previousIteration(_data->getDimensions(), Eigen::VectorXd::Zero(getSize())),
-      _exchangeSubsteps(exchangeSubsteps)
+      _exchangeSubsteps(exchangeSubsteps),
+      _timeStepsStoragePrevious()
 {
   PRECICE_ASSERT(_data != nullptr);
+  _timeStepsStoragePrevious.setInterpolationDegree(3); // @todo hard-coded for now, but we need to somehow link this to <read-data waveform-order="ORDER" />
```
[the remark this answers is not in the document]
BenjaminRodenberg: Ok. Just wanted to make sure nothing important is missing here w.r.t acceleration. Maybe I will also first take care of https://github.com/precice/precice/pull/1765 and the refactoring w.r.t Waveform and Storage. Then the issue with the degree here might also become a non-issue. We will see soon :)

src/acceleration/Acceleration.cpp
```diff
@@ -17,16 +17,22 @@ void Acceleration::checkDataIDs(const DataMap &cplData) const
...
+      couplingData->values() += oldValues * (1 - omega);
+
+      if (couplingData->hasGradient()) {
+        auto       gradients      = stample.sample.gradients;
+        const auto oldGradients   = couplingData->getPreviousGradientsAtTime(stample.timestamp); // IMPORTANT DETAIL: The interpolation that we use for resampling does not necessarily have to be the same interpolation as the interpolation the user accesses via read-data. (But probably it is easier to just use the same)
+        couplingData->gradients() = gradients * omega;
+        couplingData->gradients() += oldGradients * (1 - omega);
+      }
+
+      couplingData->setSampleAtTime(stample.timestamp, couplingData->sample());
```
uekerman, line 35: This now also happens if `substeps==false`, right? Or how is this case treated?

src/cplscheme/CouplingData.hpp
```diff
@@ -103,7 +108,7 @@ class CouplingData {
   mesh::PtrData _data;
 
   /// Sample values of previous iteration (end of time window).
-  time::Sample _previousIteration;
+  time::Storage _timeStepsStoragePrevious;
```
uekerman, line 111:
```suggestion
  time::Storage _previousTimeStepsStorage;
```
? Could be more consistent with the naming above

tests/serial/time/implicit/serial-coupling/WaveformSubcyclingWithConstantAcceleration.xml
```diff
@@ -0,0 +1,53 @@
...
+  <m2n:sockets acceptor="SolverOne" connector="SolverTwo" />
+
+  <coupling-scheme:serial-implicit>
+    <participants first="SolverOne" second="SolverTwo" />
+    <max-time-windows value="1" />
+    <time-window-size value="1" />
+    <max-iterations value="4" />
+    <min-iteration-convergence-measure min-iterations="4" data="DataOne" mesh="MeshOne" />
+    <exchange data="DataOne" mesh="MeshOne" from="SolverOne" to="SolverTwo" initialize="on" />
+    <exchange data="DataTwo" mesh="MeshOne" from="SolverTwo" to="SolverOne" initialize="on" />
```
uekerman, line 48: Why don't we need to switch `substeps` on here?

tests/serial/time/implicit/serial-coupling/WaveformSubcyclingWithConstantAcceleration.cpp
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  precice.finalize();
+}
+
+BOOST_AUTO_TEST_SUITE_END() // Integration
+BOOST_AUTO_TEST_SUITE_END() // Serial
+BOOST_AUTO_TEST_SUITE_END() // Time
+BOOST_AUTO_TEST_SUITE_END() // Explicit
+BOOST_AUTO_TEST_SUITE_END() // ParallelCoupling
```
uekerman, line 133:
```suggestion
BOOST_AUTO_TEST_SUITE_END() // Integration
BOOST_AUTO_TEST_SUITE_END() // Serial
BOOST_AUTO_TEST_SUITE_END() // Time
BOOST_AUTO_TEST_SUITE_END() // Implicit
BOOST_AUTO_TEST_SUITE_END() // SerialCoupling
```

tests/serial/time/implicit/serial-coupling/WaveformSubcyclingWithConstantAccelerationNoInit.cpp
```diff
@@ -0,0 +1,135 @@
...
+  }
+
+  precice.finalize();
+}
+
+BOOST_AUTO_TEST_SUITE_END() // Integration
+BOOST_AUTO_TEST_SUITE_END() // Serial
+BOOST_AUTO_TEST_SUITE_END() // Time
+BOOST_AUTO_TEST_SUITE_END() // Explicit
+BOOST_AUTO_TEST_SUITE_END() // ParallelCoupling
```
uekerman, line 133:
```suggestion
BOOST_AUTO_TEST_SUITE_END() // Implicit
BOOST_AUTO_TEST_SUITE_END() // SerialCoupling
```

uekerman (approved):
I lost a bit overview on how and where we treat the `substeps` option, see below. Otherwise, looks good. Be sure to document open TODOs beyond the code when merging.

src/acceleration/Acceleration.cpp
BenjaminRodenberg: Yes. But if `substeps==false`, the functions still return reasonable values (if `substeps==false`, just a single value at `WINDOW_END`).

tests/serial/time/implicit/serial-coupling/WaveformSubcyclingWithConstantAcceleration.xml
BenjaminRodenberg: Because at the moment the default is `substeps="true"` this will be changed via https://github.com/precice/precice/pull/1732. Then we will need to explicit `substeps="true"` or the test will fail.

src/cplscheme/CouplingData.hpp
BenjaminRodenberg: Done. [72c63a2](https://github.com/precice/precice/pull/1794/commits/72c63a2b096021f4db7ea6b9a2eb0c42d7f71336)

src/acceleration/Acceleration.cpp
uekerman: Because if `substeps==false`, we don't even collect anything in `stamples()`, correct?

Issue #1793, opened 2023-09-04
Master Thesis: Dependency Resolution for c++ projects (NuGet)


CodyGirl:
---
name: Discussion
about: Dependency Management
title: Dependency Management for C++ projects in preCICE
labels: discussion
assignees: @CodyGirl 

---

**Please describe the problem you are trying to solve.**
The GitHub dependency graph only recognizes the C++ dependencies [if they follow NuGet package management](https://docs.github.com/en/code-security/supply-chain-security/understanding-your-software-supply-chain/about-the-dependency-graph#supported-package-ecosystems). As a part of my master thesis, I would like to see if this works for this repository. Once we make GitHub dependency graph recognizes the dependencies, the Grafana dashboard (part of thesis solution) can also pick it up.

**Describe the solution you propose.**
The solution involves creating a dependency management file. This file is similar in functionality as we have for python repos (requirements.txt) only the file format is different. It is a packages.config file which lists all the dependencies along with its version and is then resolved automatically by the github dependency graph.
An example of packages.config file can be seen on [this link](https://digitalhouseblog.wordpress.com/2019/08/22/how-to-make-a-nuget-package-for-c/)

**Describe alternatives you've considered**
An alternative would be simply put a `requirements.txt` with all listed dependencies. However, that is not a standard solution for a C++ project (although it would work in exactly the same way for us behind the scenes).

**Expected Outcome**
Expected outcome is that after putting a packages.config file, we will see the external dependencies of preCICE in the graph.

MakisH:
This sounds quite interesting. Could you maybe start with a PR? We could then refine it in the review with the exact dependencies.

fsimonis:
This may give us green light here: https://www.bestpractices.dev/en/projects/3895?criteria_level=1#external_dependencies

PR #1798, opened 2023-09-14
#1793 NuGet package manager sample dependency file

CodyGirl:

This PR is corresponding to Issue #1793 
It has a sample packages.config file with a subset dependencies to check if they are detected by the GitHub dependency graphs.

### DIFF
--- a/packages.config
+++ b/packages.config
@@ -1,0 +1,7 @@
+<?xml version="1.0" encoding="utf-8"?>
+<packages>
+  <package id="Eigen3" version="3.4.0" />
+  <package id="Boost" version="1.71.0" />
+  <package id="PETSc" version="3.12" />
+  <package id="Ginkgo" version="1.5" />
+</packages>


MakisH:
Thanks a lot, @CodyGirl! As I am also very new to this, do you know if and where we can preview what Dependabot detects?

Regarding the dependency versions, I guess we still need:

- CMake 3.16.1 (build dependency only)
- libxml2 (any 2.*.* version)
- We need to somehow declare the dependency to MPI

The most up-do-date documentation page about dependencies should be on the `precice-v3` branch of the website: https://github.com/precice/precice.github.io/blob/precice-v3/pages/docs/installation/installation-source-dependencies.md If you see something outdated, please contribute! :smiley: (e.g., the Eigen 3.4.0 looks suspicious)

@davidscn should GinkGo also be in the dependencies page? We also list optional dependencies there.

davidscn:
> @davidscn should GinkGo also be in the dependencies page? We also list optional dependencies there.

I would say no, as we are not compatible with a released version.

MakisH:
> > @davidscn should GinkGo also be in the dependencies page? We also list optional dependencies there.
> 
> I would say no, as we are not compatible with a released version.

Where is the dependency documented, then? I am afraid we may forget to document the feature at all. If we currently don't have documentation, I would open an issue in the website repo about this.

MakisH:
But my original question remains: Is there a way I can test this on GitHub before merging?

fsimonis:
Theoretically, we also depend on Python (libpython) and fmtlib.

CodyGirl:
@MakisH https://github.com/CodyGirl/precice/network/dependencies?page=2
Last 4 are the ones which were mentioned in the packages.config file.

MakisH:
I also understand that the dependencies are detected automatically from various sources. Some of them are detected by the list of GitHub Actions, some of them are detected by the NuGet list of packages.

If I export the SBOM (software bill of materials) SPDX (software package data exchange) file, I get, for example, these two packages:

```
{"SPDXID":"SPDXRef-nuget-PETSc-3.12","name":"nuget:PETSc","versionInfo":"3.12","downloadLocation":"NOASSERTION","filesAnalyzed":false,"supplier":"NOASSERTION","externalRefs":[{"referenceCategory":"PACKAGE-MANAGER","referenceLocator":"pkg:nuget/PETSc@3.12","referenceType":"purl"}]}

{"SPDXID":"SPDXRef-actions-actions-checkout-4","name":"actions:actions/checkout","versionInfo":"4","downloadLocation":"NOASSERTION","filesAnalyzed":false,"supplier":"NOASSERTION","externalRefs":[{"referenceCategory":"PACKAGE-MANAGER","referenceLocator":"pkg:githubactions/actions/checkout@4","referenceType":"purl"}]}
```

The URL is not part of the specification, at least not the one exported. GitHub must be adding that automatically in the rendered list.

MakisH:
> Theoretically, we also depend on Python (libpython) and fmtlib. The latter is tricky, as we use a checked-in version. Question is also how to declare fully integrated dependencies such as tcb span.

Since the question here is not about runtime dependencies of the code, but generally about third-party software that the project depends upon (apparently including GitHub Actions), then we should include fmtlib.

If we have the full list of dependencies (software bill of materials), then we can answer questions such as "is there a security issue or other bug for which we would need to update the dependency version?" or "are the licenses compatible?".

MakisH:
I think that having such dependency files for preCICE and for other adapters would be interesting in the context of formally describing the preCICE ecosystem. In that sense, this is related to https://github.com/precice/preeco-orga/issues/17

The ability to automatically generate SBOMs could potentially slightly help with integrating preCICE into industrial workflows.

The Dependabot support would generally be interesting, but I wonder if it actually helps in anything: we have a policy of supporting older versions of dependencies for convenience, and we also typically don't run anything online or with security implications. As such, I could also imagine the Dependabot integration to make us simply run after Dependabot notifications (as we already do on the website, for example, where this is needed a bit more).

Potentially interesting (not sure to whom) would be to add this to bindings, to automatically get notifications for adapters we might need to update.

PR #1805, opened 2023-09-19
Compute bounding box for pum clustering locally

davidscn:
## Main changes of this PR

Computes the bounding box used to estimate the cluster radius locally instead of using the pre-computed bounding box of the meshes.

While investigating parallel performance of the implementation I notices quite some load imbalances. Reason: the cluster-radius estimates are poor, leading to large clusters on some ranks, which are consequently more expensive to compute.

The miserable radius computation puzzled me a bit and after digging a bit deeper, it turned out that calling `mesh->getBoundingBox()` in `computeMapping` returns the global bounding box (of the sender) instead of the bounding box of the local partition (which was what I assumed). 

Example in the image: looking at the yellow colorized partition (on the `to` participant), I received the bounding box (indicated by the outline) and the clustering algorithm generated the samples (indicated in blue) out of it as shown. However, what's actually required for the algorithm to work properly is the bounding box of the local yellow partition. In the example, I used a single rank for the `from` participant and four ranks for the `to` participant.

image

I don't understand in the first place, where the `to` participant receives information about the global bounding box of the `from` participant. Also, I'm not sure whether this indicates an actual bug in the interplay of `BoundingBox` as used in our `Mesh` data structure. IIRC, we can't simply clear and recompute the bounding box in the mesh due to the direct mesh access, where the bounding box might be defined by a user. In any case: if the behavior is intended this way, we should probably rename the function to `mesh::getGlobalBoundingBox()`.

I will try to come up with a unit test to prevent future confusion in this regard, at least in the pum mapping.  

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/query/Index.cpp
+++ b/src/query/Index.cpp
@@ -349,6 +349,30 @@ ProjectionMatch Index::findTriangleProjection(const Eigen::VectorXd &location, i
   return *min;
 }
 
+mesh::BoundingBox Index::getRtreeBounds()
+{
+  PRECICE_TRACE();
+  // if the mesh is empty, we will most likely hit an assertion in the bounding box class
+  // therefore, we keep the assert here, but might want to return an empty bounding box in case
+  // we want to allow calling this function with empty meshes
+  PRECICE_ASSERT(_mesh->vertices().size() > 0);
+
+  auto            rtreeBox = _pimpl->getVertexRTree(*_mesh)->bounds();
+  int             dim      = _mesh->getDimensions();
+  Eigen::VectorXd min(dim), max(dim);
+
+  min[0] = rtreeBox.min_corner().get<0>();
+  min[1] = rtreeBox.min_corner().get<1>();
+  max[0] = rtreeBox.max_corner().get<0>();
+  max[1] = rtreeBox.max_corner().get<1>();
+
+  if (dim > 2) {
+    min[2] = rtreeBox.min_corner().get<2>();
+    max[2] = rtreeBox.max_corner().get<2>();
+  }
+  return mesh::BoundingBox{min, max};
+}
+
 void Index::clear()
 {
   _pimpl->clear();
--- a/docs/changelog/1805.md
+++ b/docs/changelog/1805.md
@@ -1,0 +1,1 @@
+- Replaced calls to `computeBoundingBox` by manual implementation in partition of unity mapping for better cluster radius computations in parallel.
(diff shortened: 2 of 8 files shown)


davidscn:
So the actual issue here is that `mesh::clear()` doesn't reset the bounding box in the mesh data structure. Right now, we call initially `computeMapping` during the tagging (still don't know why the whole mesh is then sent to each rank, but ok). Afterwards, the computed bounding box remains there forever. Reason: If the bounding box was set using the direct mesh access, the bounding box holds the user input and we cannot delete it. Overall, the interplay of all this is a bit unsatisfactory. The solution for the mapping class here is somewhat trivial, we just compute everything necessary in the mapping class (we even should do that as we could also combine the direct mesh access with mappings). However, maybe we need to revisit the bounding box - mesh handling once more and clear it up. The current behavior might also lead to unexpected behavior in combination with dynamic meshes.

Adding a test for the applied changes is, unfortunately, incredible complex as it applies to realistic cases with non-equal vertex distributions and the changes are readable in the logs (and measurable in the performance), but not directly visible in the API and return values. Since now the bounding box computation/clustering is anyway decoupled from the mesh class, I think we can move on without a test. The general functionality is of course still covered.

fsimonis:
I think the reason for `clear()` not resetting the BoundingBox has to do with the repartitioning and may even be part of #633. I am not 100% sure though.

src/query/Index.cpp
```diff
@@ -349,6 +349,30 @@ ProjectionMatch Index::findTriangleProjection(const Eigen::VectorXd &location, i
...
+  min[1] = rtreeBox.min_corner().get<1>();
+  max[0] = rtreeBox.max_corner().get<0>();
+  max[1] = rtreeBox.max_corner().get<1>();
+
+  if (dim > 2) {
+    min[2] = rtreeBox.min_corner().get<2>();
+    max[2] = rtreeBox.max_corner().get<2>();
+  }
+  return mesh::BoundingBox{min, max};
+}
```
uekerman, line 374: Couldn't we add unit tests for this function?

uekerman:
> So the actual issue here is that mesh::clear() doesn't reset the bounding box in the mesh data structure. Right now, we call initially computeMapping during the tagging (still don't know why the whole mesh is then sent to each rank, but ok). Afterwards, the computed bounding box remains there forever. Reason: If the bounding box was set using the direct mesh access, the bounding box holds the user input and we cannot delete it. Overall, the interplay of all this is a bit unsatisfactory. The solution for the mapping class here is somewhat trivial, we just compute everything necessary in the mapping class (we even should do that as we could also combine the direct mesh access with mappings). However, maybe we need to revisit the bounding box - mesh handling once more and clear it up. The current behavior might also lead to unexpected behavior in combination with dynamic meshes.

I agree with all of this.

> still don't know why the whole mesh is then sent to each rank

This is the "filter on secondary ranks" approach (previously broadcast-filter). As a one-level initialization strategy, it is more performant for most case than "filter on primary rank" (filter first and then send), mainly due the efficiency of the MPI broadcast. Of course, two-level init would be better.

uekerman (approved):

src/query/Index.cpp
davidscn: Done

PR #1813, opened 2023-09-21
Fix use of bindings CMake vars

fsimonis:
## Main changes of this PR

This PR fixes some CMake variables that haven't been migrated yet.
Solverdummies are now tested correctly again.

## Author's checklist


davidscn:
I'm a bit surprised that macOS doesn't run the fortran solverdummy tests. Related to #1744

fsimonis:
Looks like XCode doesn't ship a fortran compiler. So we have to install fortran via brew.

PR #1815, opened 2023-09-22
Correct the threshold value in truncation of SVD

### LINKED ISSUE #1809: The truncation in “SVD factorization of Jacobian” has improper threshold value
**Describe your setup**

Operating system: _Ubuntu 22.04_
preCICE Version: develop

**Describe the problem**
In the restart method for IQN-IMVJ acceleration with SVD factorization, a truncation of small singular values is required. This is done within the `update()` function in `SVDFactorization.cpp` with code:
``` C++
if (_sigma(i) < (int) _sigma(0) * _truncationEps) {
        _cols = i;
        waste = _sigma.size() - i;
        break;
      }
```
This would generally only truncate the negative values (if the first singular value `_sigma(0)` is positive), instead of the small singular values, and result in an incorrect Jacobian matrix.

**Expected behaviour**
The small positive singular values could also be correctly truncated.

**Additional context**
This could be solved by removing the `(int)` before the threshold value.

Fujikawas:
## Main changes of this PR
Correct the threshold value for truncation the singular values in SVD, which is used in one restart method of IQN-IMVJ acceleration. 

Closes #1809

## Author's checklist

* [ ] ~For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).~

## Reviewers' checklist


uekerman (approved):
Looks reasonable. No idea what the motivation of the `(int)` here was.

Would it make sense to assert also that `_sigma(0)` is positive?

Needs a changelog entry as it fixes a bug.

docs/changelog/1815.md
```diff
@@ -0,0 +1 @@
+Fixed the incorrect threshold value for truncation in SVD, added assertions to check the resulting matrix size and value from SVD-update.
```
uekerman, line 1:
```suggestion
Fixed the threshold value for truncation in SVD for IQN-IMVJ using SVD-restart.
```

src/acceleration/impl/SVDFactorization.hpp
```diff
@@ -6,6 +6,7 @@
  *      Author: Klaudius Scheufele
  */
 
+#include <iostream>
```
fsimonis, line 9: Signs of debugging :wink: 

```suggestion
```

PR #1816, opened 2023-09-22
Use python interpreter from env

davidscn:
## Main changes of this PR

..which should be more robust. Especially for HPC clusters, where python might be loaded via modules, using the hard-coded python path is very likely to give the wrong interpreter. Using env should result in the correct one used by the OS.


fsimonis:
Sure, as long as lintian is happy with the changes.

PR #1820, opened 2023-09-25
Provide initial values in `tests/serial/mapping-volume/helpers.cpp`

BenjaminRodenberg:
## Main changes of this PR

Provides initial values in `tests/serial/mapping-volume/helpers.cpp` to make sure that constant function is used.

Without this fix test fails in #1746 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/tests/serial/mapping-volume/helpers.cpp
+++ b/tests/serial/mapping-volume/helpers.cpp
@@ -40,12 +40,13 @@ void testMappingVolumeOneTriangle(const std::string configFile, const TestContex
     BOOST_TEST(equals(mesh.triangles()[0].getArea(), 0.5), "Triangle area must be 0.5");
 
     // Initialize, write data, advance and finalize
+    std::vector<double> values{1.0, 100.0, 10.0};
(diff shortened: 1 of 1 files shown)


BenjaminRodenberg:
From my perspective not necessary to do a PR for this change anymore. I will close this one.

If we want to explicitly initialize the values in the test, we can do this by cherry picking directly to develop. I think generally explicitly initializing is better, but we can also decide to not change the test an keep things as they are. @uekerman Opinion?

uekerman:
I am not sure I have all the information to properly answer here.
What I understand is that this is an integration test on data mapping mimicking what a user of preCICE does. It is allowed to not initialize data, so a test can also cover this behavior. Refactoring all tests to initialize data makes no sense in my opinion. But maybe I am overlooking sth for this test specifically.

BenjaminRodenberg:
> Refactoring all tests to initialize data makes no sense in my opinion.

@uekerman This pretty much answers the question :grin: Ideally we would, of course test all cases, but let's not get lost in unnecessary testing. Coverage here should be good enough from what I know at the moment.

Issue #1672, opened 2023-05-30
Migrate mapping tests to samples


fsimonis:
**Please describe the problem you are trying to solve.**

The simplification of the data mappings #1660 makes them easier to test.
Prior behaviour required to create meshes, add data, setup data, then call `Mapping::map()`, and check the data.
Testing multiple inputs and or dimensions was tedious as it required to override the data etc.

**Describe the solution you propose.**

A mapping test can now work as follows:
1. Create input mesh
2. Create output mesh
3. Initialize the mapping
4. For each input data to test:
   1. Create an input sample
   2. Map the input to an output
   3. Check the output

This means that we can write helpers (at least for testing the in and out) in the form of:

```cpp
struct DirectMappingTest { time::Sample input, VectorXd expected };
struct TransientMappingTest { time::Sample input, VectorXd initialGuess, VectorXd expected };

void testNNMapping(inputMesh, outputMesh, std::vector<DirectMappingTest> tests);
void testPetRBFMapping(inputMesh, outputMesh, std::vector<TrasientMappingTest> tests);
```

fsimonis:
For context, the current mapping tests work as follows:

- Create input mesh
- Create input Data on input mesh
- Allocate data values of input mesh
- Create output mesh
- Create output data on output mesh
- Allocate data values of output mesh
- Initialize the mapping
- Specify values of output Data
- Initialize the mapping
- For each input data to test:
    - Set values of input Data
    - Wipe values of output Data 
    - Map the input Data to an output Data
    - Check the output data

Note that this method required different Data definitions on a mesh to test scalar and vectorial data.

fsimonis:
The current state is that we have moved most occurrence of the old map, but some have been creeping into the code base again.

* NearestProjection
* LinearCellInterpolation
* RadialBasisFctHelper
* PartitionOfUnityMapping

@davidscn Can you migrate your RBF tests to mapping samples?
I'll have a look at the NP and LCI tests.

Let's get this chapter closed.

PR #1824, opened 2023-09-29
Samplize Radial Basis Fct Mapping tests

Durganshu:
## Main changes of this PR

This PR migrates the Radial Basis function mapping tests to use samples.

See #1672 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mapping/tests/RadialBasisFctMappingTest.cpp
+++ b/src/mapping/tests/RadialBasisFctMappingTest.cpp
@@ -58,40 +58,42 @@ ReferenceSpecification format:
 struct MeshSpecification {
   std::vector<VertexSpecification> vertices;
   int                              meshDimension;
-  mesh::PtrMesh                    meshName;
+  std::string                      meshName;
 };
 
 /// Contains which values are expected on which rank: rank -> vector of data.
 using ReferenceSpecification = std::vector<std::pair<int, std::vector<double>>>;
 
-void getDistributedMesh(const TestContext &context,
-                        MeshSpecification &meshSpec,
-                        int                globalIndexOffset = 0,
-                        bool               meshIsSmaller     = false)
+mesh::PtrMesh getDistributedMesh(const TestContext &context,
+                                 MeshSpecification &meshSpec,
+                                 int                globalIndexOffset = 0,
+                                 bool               meshIsSmaller     = false)
 {
-  int i = 0;
+  mesh::PtrMesh distributedMesh(new mesh::Mesh(meshSpec.meshName, meshSpec.meshDimension, testing::nextMeshID()));
+  int           i = 0;
   for (auto &vertex : meshSpec.vertices) {
     if (vertex.rank == context.rank or vertex.rank == -1) {
       if (vertex.position.size() == 3) // 3-dimensional
-        meshSpec.meshName->createVertex(Eigen::Vector3d(vertex.position.data()));
+        distributedMesh->createVertex(Eigen::Vector3d(vertex.position.data()));
       else if (vertex.position.size() == 2) // 2-dimensional
-        meshSpec.meshName->createVertex(Eigen::Vector2d(vertex.position.data()));
+        distributedMesh->createVertex(Eigen::Vector2d(vertex.position.data()));
 
       if (vertex.owner == context.rank)
-        meshSpec.meshName->vertices().back().setOwner(true);
+        distributedMesh->vertices().back().setOwner(true);
       else
-        meshSpec.meshName->vertices().back().setOwner(false);
+        distributedMesh->vertices().back().setOwner(false);
 
       i++;
     }
   }
-  addGlobalIndex(meshSpec.meshName, globalIndexOffset);
+  addGlobalIndex(distributedMesh, globalIndexOffset);
   // All tests use eight vertices
   if (meshIsSmaller) {
-    meshSpec.meshName->setGlobalNumberOfVertices(7);
+    distributedMesh->setGlobalNumberOfVertices(7);
   } else {
-    meshSpec.meshName->setGlobalNumberOfVertices(8);
+    distributedMesh->setGlobalNumberOfVertices(8);
   }
+  return distributedMesh;
 }
 
 Eigen::VectorXd getDistributedData(const TestContext &      context,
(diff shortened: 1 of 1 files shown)


fsimonis (commented):
Is the old version of `getDistributedMesh` now dead code?

src/mapping/tests/RadialBasisFctMappingTest.cpp
```diff
@@ -103,6 +103,47 @@ void getDistributedMesh(const TestContext &      context,
   data->values() = d;
 }
 
+Eigen::VectorXd getDistributedMesh(const TestContext &      context,
+                                   MeshSpecification const &vertices,
+                                   mesh::PtrMesh &          mesh,
+                                   int                      globalIndexOffset = 0,
+                                   bool                     meshIsSmaller     = false)
```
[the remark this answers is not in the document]
Durganshu: Yes, this function can be split into two. 

Also, I didn't remove the older version of `getDistributedMesh` for future usage or to allow any potential backward compatibility. If that's not needed, I can remove it.

src/mapping/tests/RadialBasisFctMappingTest.cpp
```diff
@@ -55,77 +55,54 @@ ReferenceSpecification format:
 - -1 on rank means all ranks
 - v is the expected value of n-th vertex on that particular rank
 */
-using MeshSpecification = std::vector<VertexSpecification>;
+struct MeshSpecification {
+  std::vector<VertexSpecification> vertices;
+  int                              meshDimension;
+  mesh::PtrMesh                    meshName;
```
fsimonis, line 61: This should be the name of the mesh, not the mesh itself

src/mapping/tests/RadialBasisFctMappingTest.cpp
```diff
@@ -152,15 +122,16 @@ void testDistributed(const TestContext &    context,
...
                      bool                   meshIsSmaller       = false)
 {
-  int meshDimension  = inMeshSpec.at(0).position.size();
-  int valueDimension = inMeshSpec.at(0).value.size();
+  int valueDimension = inMeshSpec.vertices.at(0).value.size();
 
-  mesh::PtrMesh   inMesh(new mesh::Mesh("InMesh", meshDimension, testing::nextMeshID()));
-  Eigen::VectorXd inValues = getDistributedMesh(context, inMeshSpec, inMesh, inGlobalIndexOffset);
+  inMeshSpec.meshName = mesh::PtrMesh(new mesh::Mesh("InMesh", inMeshSpec.meshDimension, testing::nextMeshID()));
+  getDistributedMesh(context, inMeshSpec, inGlobalIndexOffset);
```
fsimonis, line 128: This is how I would like the final version to look like:

```suggestion
  MeshPtr inMesh = getDistributedMesh(context, inMeshSpec, inGlobalIndexOffset);
```

So, `getDistributedMesh` builds a `Mesh` and returns in as a `MeshPtr`.

src/mapping/tests/RadialBasisFctMappingTest.cpp
```diff
@@ -187,25 +158,35 @@ BOOST_AUTO_TEST_CASE(DistributedConsistent2DV1)
...
+      {-1, 1, {1, 0}, {3}},
+      {-1, 1, {1, 1}, {4}},
+      {-1, 2, {2, 0}, {5}},
+      {-1, 2, {2, 1}, {6}},
+      {-1, 3, {3, 0}, {7}},
+      {-1, 3, {3, 1}, {8}}};
+
+  MeshSpecification in = {
+      inVertexList,
+      static_cast<int>(static_cast<int>(inVertexList.at(0).position.size())),
```
fsimonis, line 173: Why do you cast this to int twice?

Also, it would be cleaner if you would deduce the dimensionality of the problem at the very top of the function and define it as a constant.

src/mapping/tests/RadialBasisFctMappingTest.cpp
```diff
@@ -187,25 +158,35 @@ BOOST_AUTO_TEST_CASE(DistributedConsistent2DV1)
...
+      nullptr};
+
+  std::vector<VertexSpecification> outVertexList = {
```
fsimonis, line 176: There is no reason not to directly construct the list:
```suggestion
  std::vector<VertexSpecification> outVertexList{
```

src/mapping/tests/RadialBasisFctMappingTest.cpp
```diff
@@ -187,25 +158,35 @@ BOOST_AUTO_TEST_CASE(DistributedConsistent2DV1)
...
+      {0, -1, {0, 0}, {0}},
+      {0, -1, {0, 1}, {0}},
+      {1, -1, {1, 0}, {0}},
+      {1, -1, {1, 1}, {0}},
+      {2, -1, {2, 0}, {0}},
+      {2, -1, {2, 1}, {0}},
+      {3, -1, {3, 0}, {0}},
+      {3, -1, {3, 1}, {0}}};
+
+  MeshSpecification out = {
```
fsimonis, line 186: Same here, you can directly construct the MeshSpecification:

```suggestion
  MeshSpecification out{
```

src/mapping/tests/RadialBasisFctMappingTest.cpp
```diff
@@ -187,25 +158,35 @@ BOOST_AUTO_TEST_CASE(DistributedConsistent2DV1)
...
+      outVertexList,
+      static_cast<int>(static_cast<int>(outVertexList.at(0).position.size())),
+      nullptr};
```
fsimonis, line 189: I would like this to be the name of the mesh in the final version.

fsimonis (changes requested):
We are slowly getting there, but there are some more things to do.

src/mapping/tests/RadialBasisFctMappingTest.cpp
Durganshu: I think the double casting was because of a mistake in Find and Replace. It'll be fixed in the latest commit.

src/mapping/tests/RadialBasisFctMappingTest.cpp
Durganshu: Do you mean instead of having `mesh::PtrMesh meshName` as a member, it should have `std::string meshName` as a member?

src/mapping/tests/RadialBasisFctMappingTest.cpp
fsimonis: It should be the name according to https://github.com/precice/precice/pull/1824#discussion_r1350331535

PR #1826, opened 2023-09-30
Add intel CI image to build and test workflow

fsimonis:
## Main changes of this PR

This PR adds the new `precice/ci-intel:latest` image to the build-and-test workflow.
It is based on the [Intel oneAPI HPC Toolkit](https://www.intel.com/content/www/us/en/developer/tools/oneapi/hpc-toolkit.html) and uses Intel C,C++,Fortran compilers as well as Intel MPI.
The image also contains a compatible PETSc version.

We finally need to test Intel stuff and the oneAPI makes this possible.

The GitHub runners only off 14GB of storage. So this may fail until we get custom runners.

## Author's checklist

## Reviewers' checklist


fsimonis:
The image is too large for the GitHub runners.

This also means that we can not use GitHub runners to shave off some unnecessary features.

A solution would be to build it from scratch:
https://github.com/intel/oneapi-containers/blob/master/images/docker/hpckit/Dockerfile.ubuntu-22.04

However, the GPG key used by Intel has expired. So we have to wait: https://github.com/oneapi-src/oneapi-ci/issues/77

fsimonis:
I managed to reduce the size of intel image to 8.4 GB.
This should be fine for the GitHub runners.

fsimonis:
The release build needs to use `ffp-model=precise` or `-fhonor-infinities -fhonor-nans` to work. [Read more](https://www.intel.com/content/www/us/en/docs/dpcpp-cpp-compiler/developer-guide-reference/2023-1/fp-model-fp.html)

There seems to be some problem with the following MPIPorts tests:

```
CommunicationTests/MPIPorts/Inter/SendReceiveFourProcesses
CommunicationTests/MPIPorts/Server/SendReceiveFour
CommunicationTests/MPIPorts/Server/SendReceiveFourV2
```

They sometimes fail with such an error:

```
Abort(2176143) on node 0 (rank 0 in comm 0): Fatal error in PMPI_Comm_accept: Other MPI error, error stack:
PMPI_Comm_accept(121)..........: MPI_Comm_accept(port="tag#0$connentry#02009F11AC1100020000000000000000$", MPI_INFO_NULL, root=0, MPI_COMM_SELF, newcomm=0x7ffd45ee2378) failed
MPID_Comm_accept(251)..........:
MPIDI_OFI_mpi_comm_accept(955).:
dynproc_exchange_map(531)......:
MPIDI_OFI_handle_cq_error(1105): OFI poll failed (ofi_events.c:1105:MPIDI_OFI_handle_cq_error:Truncation error)
```

fsimonis:
After disabling the mpi ports tests for intel mpi, the tests complete for the debug build.
The release build leads to failing tests in the `TimeHandler`, I try setting the fp-model to precise.

Issue #1463, opened 2022-11-08
Clean up IMVJ acceleration


uekerman:
## Problem

The IMVJ implementation has many different variants, for historic reasons (we developed the method in multiple steps) and for research reasons (we wanted to compare different variants). However, quite naturally, this led to the situation that:

- Not all variants are used, some are even discouraged.
- [x] We have a [bad test coverage](https://app.codecov.io/gh/precice/precice/blob/develop/src/acceleration/MVQNAcceleration.cpp).
- Documentation is not ideal, see also #1064.

Other issues:

- [x] We use `MVQN` in the code, which is inconsistent with `IQNILS` and with the configuration (where we use `IQN-IMVJ` as name already, see [config ref](https://precice.org/configuration-xml-reference.html#accelerationiqn-imvj)).
- [ ] ~~The "cyclic communication" we use in `ParallelMatrixOperations` is hacky and gives problems regularly. IIRC, this functionality is only needed for a sub-optimal variant of IMVJ.~~

https://github.com/precice/precice/blob/bee47b164e28e6e5936e9b46de39d81cd64c42fd/src/acceleration/impl/ParallelMatrixOperations.hpp#L310-L314

## Solution

Carefully study which variants we want to keep. Test and document these and remove the rest.

## Connected issues

- #885

Fujikawas:
> Regarding the cyclic communication, this is handled by the `ParallelMatrixOperations`.
> 
> The IQNIMVJAcceleration uses this here:
> 
> https://github.com/precice/precice/blob/a601de032919df316745085ac318212892494d2f/src/acceleration/IQNIMVJAcceleration.cpp#L84

We use the `_parMatrixOps->multiply()` in IQN-IMVJ method with or without restart for multiple times. And we also use the same function in the `SVDFactorization` class for `SVD-Restart` method. Based on these I think it's not only for a specific sub-optimal option in IMVJ method. 
So we either need to find another function to replace its `multiply()`, since it's the only operation we actually use, or we'd better keep it.

uekerman:
Mmh, @Fujikawas can you mathematically follow why we need the squared (not tall-and-skinny) matrix multiplication in these places? I did not yet look into details. Maybe, you are right and we also need them there :/

PR #1829, opened 2023-10-02
Add tests for IMVJ with restart methods

Fujikawas:
## Main changes of this PR
Added integration tests which employed IMVJ acceleration method with various restart methods.

To solve the low coverage problem described in https://github.com/precice/precice/issues/1463

## Author's checklist

## Reviewers' checklist


Fujikawas:
Current `IQNIMVJAcceleration.cpp`:  `Coverage% = 86%`

tests/quasi-newton/parallel/TestQN6.cpp
```diff
@@ -0,0 +1,22 @@
...
+#include <precice/precice.hpp>
+#include "../helpers.hpp"
+
+BOOST_AUTO_TEST_SUITE(Integration)
+BOOST_AUTO_TEST_SUITE(QuasiNewton)
+BOOST_AUTO_TEST_SUITE(Parallel)
+BOOST_AUTO_TEST_CASE(TestQN6)
+{
+  PRECICE_TEST("SolverOne"_on(2_ranks), "SolverTwo"_on(2_ranks));
+  // serial coupling, IQN-IMVJ (which is identical to IQN-ILS as only first timestep is considered), relaxed QR2 filter
```
uekerman, line 14: Is this comment actually correct? The config looks like this test does 5 time windows?

uekerman (approved):
Very good additions :+1:

Unit tests that actually test whether we do the correct thing could also be helpful, but that's for another PR.

@Fujikawas If you have a good idea, we could try renaming "QN7" etc to sth more readable. But not a must.

tests/quasi-newton/parallel/TestQN6.cpp
Fujikawas: Yes, it's 5 time windows. I'll improve them. 
And I would like to add the goals of relevant tests here in the comments so we don't need to do a lot of renaming and still provide some insight.

PR #1830, opened 2023-10-04
Optimize building the communication map

davidscn:
## Main changes of this PR

Optimizes the function [`buildCommunicationMap`](https://github.com/precice/precice/blob/7bdf93d6e060bad08cbef26946ea0536478251cd/src/m2n/PointToPointCommunication.cpp#L255) significantly.

While investigating PUM mappings, running larger cases on more cores became infeasible. After digging a bit deeper into the reason, it turned out that creating the 'lookup table' in `buildCommunicationMap` takes 99% of the time. So, I spent some effort to optimize the current way we compute the communication map. I'm not 100% sure that the proposed changes here are actually legal, as it performs a sorting on the vertex indices. However, I _could_ recover less, but still significant speed-up without doing so.

The changes are mostly relevant for a large number of global vertices. Here are some relevant performance measurements for one of the aste standard cases (one time-step, one mapping, serial-explicit schemes) on a 600k vertices input mesh and corresponding speedups on SuperMUC-NG

On 2 x 2 cores (input x output)

|   | `buildCommMap` | global run time |
| ------------- | ------------- | ------------- |
| `develop` | 2,081,519 |16,892,952 (~17secs) |
| `opt-bbmap` | 3,360  | 14,845,686 (~15secs) |
| speed up | 619 | 1.14 |

On 24 x 24 cores (input x output)

|   | `buildCommMap` | global run time |
| ------------- | ------------- | ------------- |
| `develop` |  839,366,952 |858,955,792 (~14min) |
| `opt-bbmap` | 2,358  | 4,554,192 (~4.5secs)|
| speed up | 355,965  | 188 |

On 240 x 240 cores (input x output)

|   | `buildCommMap` | global run time |
| ------------- | ------------- | ------------- |
| `develop` | 4,645,940,719 |4,688,943,494 (~77min)|
| `opt-bbmap` |  2,235 | 10,057,928 (~10 secs) |
| speed up |  2,078,720 | 466 |

Note that the issue is a general preCICE issue. However, using a NN mapping (as opposed to the PUM mapping used here) with a filter on the primary rank reduces the vertex count relevant for the function significantly (mapping fine to coarse), such that the issue would be observable only later. Also note that the global run time is now dominated by other events instead of the BB map creation. The larger the input mesh gets, the more significant the changes here would be (and vice versa).

EDIT: With the first version of this PR, cases with 1.35mio (bbMap in 3,300us) ran on 240 cores in <1min as well.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/m2n/PointToPointCommunication.cpp
+++ b/src/m2n/PointToPointCommunication.cpp
@@ -236,21 +236,42 @@ void printLocalIndexCountStats(std::map<int, std::vector<int>> const &m)
   }
 }
 
+namespace {
+/**
+   * @brief This function is by by and large the same as std::set_intersection().
+   * The only difference is that we don't return the intersection set itself, but
+   * we return the indices of elements in \p InputIt1, which appear in both sets
+   * ( \p InputIt1 and \p InputIt2 )
+   * The implementation was taken from
+   * https://en.cppreference.com/w/cpp/algorithm/set_intersection#Version_1 with the
+   * only difference that we compute and store std::distance() in the output iterator.
+   * Similar to the std function, this function operates on sorted ranges.
+   *
+   * @param ref1 The reference iterator, to which we compute the distance/indices.
+   * @param first1 The begin of the first range we want to compute the intersection with
+   * @param last1 The end of the first range we want to compute the intersection with
+   * @param first1 The begin of the second range we want to compute the intersection with
+   * @param last1 The end of the second range we want to compute the intersection with
+   * @param d_first Beginning of the output range
+   * @return OutputIt Iterator past the end of the constructed range.
+   */
 template <class InputIt1, class InputIt2, class OutputIt>
-OutputIt set_intersection(InputIt1 ref, InputIt1 first1, InputIt1 last1,
-                          InputIt2 first2, InputIt2 last2, OutputIt d_first)
+OutputIt set_intersection_indices(InputIt1 ref1, InputIt1 first1, InputIt1 last1,
+                                  InputIt2 first2, InputIt2 last2, OutputIt d_first)
 {
   while (first1 != last1 && first2 != last2) {
     if (*first1 < *first2)
       ++first1;
     else {
       if (!(*first2 < *first1))
-        *d_first++ = std::distance(ref, first1++); // *first1 and *first2 are equivalent.
+        *d_first++ = std::distance(ref1, first1++); // *first1 and *first2 are equivalent.
       ++first2;
     }
   }
   return d_first;
 }
+} // namespace
+
 /** builds the communication map for a local distribution given the global distribution.
  *
  *
@@ -260,12 +281,19 @@ OutputIt set_intersection(InputIt1 ref, InputIt1 first1, InputIt1 last1,
  *
  * @returns the resulting communication map for rank thisRank
  *
- * The approximate complexity of this function is:
- * \f$ \mathcal{O}(n \log(n) + m \log(n)) \f$
+ * The worst case complexity of the function is:
+ * \f$ \mathcal{O}(p n \log(n) + p 2 (2 n)) \f$
  *
- * * n is the total number of data indices for all ranks in `otherVertexDistribution'
- * * m is the number of local data indices for the current rank in `thisVertexDistribution`
+ * which is composed of the initial std::sort for each vector and the subsequent
+ * computation of the intersection.
  *
+ * * n is the number of data indices for each vector in `otherVertexDistribution'
+ * * p number of ranks
+ * * Note that n becomes smaller, if we have more ranks.
+ *
+ * However, in case our indices are already sorted (which is typically the case),
+ * the complexity boils down to linear complexity
+ * \f$ \mathcal{O}(p n) \f$
  */
 std::map<int, std::vector<int>> buildCommunicationMap(
     // `thisVertexDistribution' is input vertex distribution from this participant.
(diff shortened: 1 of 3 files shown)


src/m2n/PointToPointCommunication.cpp
```diff
@@ -245,42 +281,65 @@ void printLocalIndexCountStats(std::map<int, std::vector<int>> const &m)
...
-    }
-    return lookupIndexRank;
-  }();
+  std::map<int, std::vector<int>> communicationMap;
+
+  // take advantage that these data structures are in most cases sorted by construction,
+  // i.e., we perform here mostly a safety check and don't perform an actual sorting
+  if (!std::is_sorted(iterator->second.begin(), iterator->second.end())) {
+    std::sort(iterator->second.begin(), iterator->second.end());
+  }
```
fsimonis, line 315: I am pretty sure this should always be the case and would suggest asserting this instead (also below).

Then you should also be able to revert the parameter back to const ref.
```suggestion
  PRECICE_ASSERT(std::is_sorted(iterator->second.begin(), iterator->second.end()));
```

src/m2n/PointToPointCommunication.cpp
```diff
@@ -245,42 +281,65 @@ void printLocalIndexCountStats(std::map<int, std::vector<int>> const &m)
...
 
-  auto const &indices = iterator->second;
+  // now we iterate over all other vertex distributions to compute the intersection
+  for (auto &other : otherVertexDistribution) {
```
fsimonis, line 318: We have C++17, so this could be a good opportunity to clean up the below code with structured bindings:

```suggestion
  for (const auto & [rank, vertices] : otherVertexDistribution) {
```

src/m2n/PointToPointCommunication.cpp
```diff
@@ -236,6 +236,42 @@ void printLocalIndexCountStats(std::map<int, std::vector<int>> const &m)
...
+   * @param d_first Beginning of the output range
+   * @return OutputIt Iterator past the end of the constructed range.
+   */
+template <class InputIt1, class InputIt2, class OutputIt>
+OutputIt set_intersection_indices(InputIt1 ref1, InputIt1 first1, InputIt1 last1,
+                                  InputIt2 first2, InputIt2 last2, OutputIt d_first)
+{
+  while (first1 != last1 && first2 != last2) {
+    if (*first1 < *first2)
+      ++first1;
```
fsimonis, line 264: Not sure if this helps, but it could be interesting to see if changing this "step first1 once" to "advance first1 until" results in a performance uplift. Definitely low prio though. 

```suggestion
    if (*first1 < *first2) {
      while(first1 != last1 && *first1 < *first2) {
        ++first1;
      }
    }
```

fsimonis (commented):
Very nice to see that the single level initialization is getting some TLC again!

The lookup table was a low-effort high-gain improvement on the original version. This looks really good!

I put in a bunch of tweaks for readability or for potentially further performance gains. Really looking forward to seeing the final form of this PR impacts the repartitioning!

src/m2n/PointToPointCommunication.cpp
```diff
@@ -245,42 +281,65 @@ void printLocalIndexCountStats(std::map<int, std::vector<int>> const &m)
...
+      // in this case there is nothing to be done
+      continue;
+    } else {
+      // we have an intersection, let's compute it
+      std::vector<int> inters;
+      // the actual worker function, which gives us the indices of intersecting elements
+      // have a look at the documentation of the function for more details
+      precice::m2n::set_intersection_indices(iterator->second.begin(), iterator->second.begin(), iterator->second.end(),
+                                             other.second.begin(), other.second.end(),
+                                             std::back_inserter(inters));
```
fsimonis, line 339: I think that this will be the biggest bottleneck after your changes.

Suggestion:
We move the `inters` outside the for loop and `clear()` it here instead.
The capacity of the vector will not shrink by doing this, which essentially is a form of a preallocation heuristic.

The below, you insert a copy of `inters` into the communication map.
This will copy the content and result in a vector with `capacity() == size()`, which should be more memory efficient.

src/m2n/PointToPointCommunication.cpp
```diff
@@ -245,42 +281,65 @@ void printLocalIndexCountStats(std::map<int, std::vector<int>> const &m)
...
-      communicationMap[iter->second].push_back(index);
+    // before starting to compute an actual intersection, we first check if elements can
+    // possibly be in both data sets by comparing upper and lower index bounds of both
+    // data sets. For typical partitioning schemes, each rank only exchanges data with
+    // a few neighbors such that this check already filters out a significant amount of
+    // computations
+    if (iterator->second.empty() || other.second.empty() || (other.second.back() < iterator->second.at(0)) || (other.second.at(0) > iterator->second.back())) {
+      // in this case there is nothing to be done
+      continue;
+    } else {
```
fsimonis, line 332: This else isn't necessary, as `continue` already skips the rest here. Reducing indentation is always a good thing :wink:

davidscn: Yes, belongs to the series of 'being to deep into coding' :sweat_smile:

fsimonis:
The following works on develop. This would be a good addition to this PR to prevent strange things from happening in the future:

```cpp
void setVertexDistribution(VertexDistribution vd)
{
  PRECICE_ASSERT(std::all_of(vd.begin(), vd.end(), [](const auto& p){return std::is_sorted(p.second.begin(),p.second.end());}));
  _vertexDistribution = std::move(vd);
}
```

src/m2n/PointToPointCommunication.cpp
davidscn: > I think that this will be the biggest bottleneck after your changes.

Well, it is the function doing the work, so in principle yes, but the function as a whole is now part of the bottle body rather than the neck. There are other functions which are now much more relevant in terms of performance and I exceeded my performance targets already by a lot and don't want to spend more effort. Still, the proposed changes are easy to test (ignoring the

> could impact further processing of the `communicationMap`

which I cannot really test, I think a different data type for the `communicationMap` should be the way to go anyway).

The proposed version is in fact slightly slower than the current version on both, my local machine as well as on SuperMUC-NG (tested the 240 cores version). On my local machine, we have around 306 us on the primary rank and 267 us on the most expensive secondary rank (the mesh is slightly smaller), using your proposed changes we end up with 434us on the primary rank and 283 us on the most expensive secondary rank.
On SuperMUC-NG we have around 2.200 us for the ccurrent version (as stated in the initial comment) and around 2.480 us for the proposed changes.

However, the scale on which these changes are observable is actually negligible, which brings me back to the first point.

src/m2n/PointToPointCommunication.cpp
davidscn: I would expect that the final behavior depends on the structure of the underlying data, so let's don't investigate this (unless you want to), see also comment below.

src/m2n/PointToPointCommunication.cpp
davidscn: That would be nice, let me check what we can get from this. I don't know, whether this would still hold if we have dynamic meshes and similar things, but IIRC we anyway reinitialize everything and the assert should be enough for the future.

davidscn:
Assuming that the processed index ranges are sorted (I only tested the most costly case with 240 cores) gives an _additional_ speed-up of 4 - 7 (required 288 to 500 us). Actually, the timings are now more diverse since the total cost depends on how many intersections we have to go through on each rank whereas before (ie with the `std::is_sorted`) we had anyway a linear complexity on each rank. I'm sure with less ranks, the speed-up will be less though.
However, using this version we could directly remove the whole event, as it looks somewhat negligible.

src/m2n/PointToPointCommunication.cpp
davidscn: Done

fsimonis:
> I'm sure with less ranks, the speed-up will be less though.
However, using this version we could directly remove the whole event, as it looks somewhat negligible.

This should be easy to check, as we still have the event. Given that this used to be a major bottleneck, I would keep the event. This allows to directly compare preCICE versions.

If we ever need to improve this further, then the next step is possibly to use ranges instead of individual ids.
So instead of storing $(0,...,100,200,...,10000)$, we store first+last $(0,100,200,10000)$ or first+count $(0,100,200,9800)$, making intersection handling etc way faster.

src/m2n/PointToPointCommunication.cpp
fsimonis: Thanks for testing! The results really surprise me. However, it's not worth investing more time here for such minuscule gains.

fsimonis (approved):
Nice and this will definitely be user-observable, so a changelog entry would be nice.

PR #1831, opened 2023-10-04
Add sample duplication tests to bspline tests

fsimonis:
## Main changes of this PR

This PR adds sample duplication tests to the BSpline tests to help deciding how to proceed in #1788

See #1788

## Author's checklist

### DIFF
--- a/src/math/tests/BSplineTest.cpp
+++ b/src/math/tests/BSplineTest.cpp
@@ -3,21 +3,38 @@
 #include "testing/Testing.hpp"
 
 #include <Eigen/Core>
+#include <boost/test/data/test_case.hpp>
 
 using namespace precice::testing;
 
 BOOST_AUTO_TEST_SUITE(MathTests)
 BOOST_AUTO_TEST_SUITE(BSpline)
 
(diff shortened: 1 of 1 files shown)


fsimonis:
@NiklasKotarsky @BenjaminRodenberg Do you plan on integrating these tests or can the PR be closed?

BenjaminRodenberg:
Hard for me to judge whether these tests are needed or not. @NiklasKotarsky could you review and merge, if you think these tests are useful?

NiklasKotarsky:
I am not sure if this is the correct place for such a test, since the underlying waveform and spline changes when we duplicate the sample at the end. It is also not obvious to me which feature breaks and in what sense when the duplication tests fail. This is also tested on a higher level in #1788 together with the min-time-step tag making the connection more clear. Therefore I would suggest closing this pull request in favor of the test in #1788.

fsimonis:
@NiklasKotarsky OK. I'll close this PR.
If you make assumptions on b-splines of a given shape, then testing them in the bsplinetests is the correct way to go though. This makes problems easier to find as setups are easier to understand and it cuts out all other problem sources.

Issue #1689, opened 2023-06-22
Quasi Newton and waveform iterations


NiklasKotarsky:
**Please describe the problem you are trying to solve.**
Quasi newton acceleration is only applied to the last timestep of the waveform iteration currently. This yields slow convergence for higher order waveform iterations.

**Describe the solution you propose.**
WIP

**Describe alternatives you've considered**
WIP

**Additional context**

BenjaminRodenberg:
Is one of the tests in https://github.com/precice/precice/pull/1681 testing for this problem? 

Important note on this topic: Since #1664 and #1671 are not merged yet the problem you are describing above might not be observable on the current state of develop ([1180edb](https://github.com/precice/precice/commit/1180edb58bacce28dd066b852edfc2eb6623135a)): If only the last time step is exchanged and accelerated there are no other time steps that are skipped. But on #1671 we should definitely be able to reproduce the problem with a simple test.

BenjaminRodenberg:
#1696 is related to this PR as a first step into the direction of more advanced acceleration schemes.
#1671 will be merged soon, but does offer a general solution for quasi-Newton + subcycling.

BenjaminRodenberg:
Might be useful for working on this issue: I worked on updating the partitioned heat equation tutorial to version 3 today. You need the following two PRs:

* FEniCS adapter: https://github.com/precice/fenics-adapter/pull/153
* tutorial: https://github.com/precice/tutorials/pull/333

You can easily apply subcycling here by, for example, reducing `fenics_dt = .1` to `fenics_dt = .01`. In the configuration file you can use different acceleration schemes like:

```
<acceleration:constant>
  <relaxation value="0.5"/>
</acceleration:constant>
```

or

```
<acceleration:IQN-ILS>
  <data name="Temperature" mesh="Neumann-Mesh" />
  <initial-relaxation value="0.1" />
  <max-used-iterations value="10" />
  <time-windows-reused value="5" />
  <filter type="QR2" limit="1e-3" />
</acceleration:IQN-ILS>
```

additionally, you can use `substeps="true/false"` in the `<exchange ... />`. For #1696, `acceleration:constant` works nicely with both options. `acceleration:IQN-ILS` only works for `substeps="false"`. Note: There is a bugfix #1719 that you will also need.

fsimonis:
Closed by #2005

Issue #1825, opened 2023-09-30
Support all acceleration schemes if `substeps="true"`


BenjaminRodenberg:
Subcycling and waveform iteration require changes to our existing acceleration schemes. This issue provides an overview and is meant for reference. Detailed discussions about individual schemes should go into dedicated issues.

Here is a list of the current state with related issues / PRs:
- [x] `acceleration:constant` (working) #1794 
- [x] `acceleration:IQN-ILS` (planned after 3.0.0) https://github.com/precice/precice/pull/1698 :arrow_forward: #2005 
- [x] `acceleration:aitken` (@Fujikawas), https://github.com/precice/precice/pull/2163 ?

BenjaminRodenberg:
Can we close this issue when #2005 is merged? IIRC Aitken would be still missing. Will schedule this issue for the next release. We do not necessarily have to close it, but a rough perspective would be helpful.

BenjaminRodenberg:
Can we close this issue? https://github.com/precice/precice/pull/2163 is merged.

uekerman:
I think, yes.
IQN-IMVJ was also covered in #2005, correct?

fsimonis:
Everything implemented. Aitken only supports the reduced method atm. #2197

PR #1698, opened 2023-06-27
Fix acceleration for waveform iterations

NiklasKotarsky:
## Main changes of this PR

Draft to solve issue  #1696 built upon #1523 . 
Ensures that the acceleration is applied to every timestep in the waveform for both QN acceleration and relaxation. 

## Motivation and additional information

Adds support for time adaptive waveform iterations by moving the interpolation to the time storage with fixed waveform order 1. 
Removes  the constant specifier for the variable extrapolationOrder in the time storage class in order to make the instance copyable.

## Author's checklist

* [ ] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [ ] I added a test to cover the proposed changes in our test suite.
* [ ] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).
* [ ] I sticked to C++17 features.
* [ ] I sticked to CMake version 3.16.3.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

PR #1834, opened 2023-10-11
Add reduced Quasi Newton acceleration (rQNWR) for waveform iterations

NiklasKotarsky:
## Main changes of this PR

Adds support for reduced Quasi Newton acceleration for waveform iterations. 

Closes  issue #1689 and improves  #1825. Contains a updated version of pull request  #1698.

## Author's checklist

## Reviewers' checklist


NiklasKotarsky:
There is still some work left to be done with the tests and refactoring, but it would be nice to get some input on the structure of the QN classes.  I have implemented it such that the original QN is unchanged from the version without rQN. Right now this is done via if statements, but there might be a better way of implementing this?

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -289,7 +309,7 @@ void BaseQNAcceleration::performAcceleration(
   // assume data structures associated with the LS system can be updated easily.
 
   // scale data values (and secondary data values)
-  concatenateCouplingData(cplData, _dataIDs, _values, _oldValues);
+  Acceleration::concatenateCouplingData(cplData, _dataIDs, _values, _oldValues);
```
uekerman, line 312: To make things more explicit? I think we don't do it like this most of the times, but not sure whether this is good practice.

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -394,22 +444,13 @@ void BaseQNAcceleration::performAcceleration(
...
       }
     }
-
-    if (std::isnan(utils::IntraComm::l2norm(xUpdate))) {
-      PRECICE_ERROR("The quasi-Newton update contains NaN values. This means that the quasi-Newton acceleration failed to converge. "
-                    "When writing your own adapter this could indicate that you give wrong information to preCICE, such as identical "
-                    "data in succeeding iterations. Or you do not properly save and reload checkpoints. "
-                    "If you give the correct data this could also mean that the coupled problem is too hard to solve. Try to use a QR "
-                    "filter or increase its threshold (larger epsilon).");
-    }
```
uekerman, line 409: why removing the check?

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -394,22 +444,13 @@ void BaseQNAcceleration::performAcceleration(
...
   }
-
-  splitCouplingData(cplData);
```
uekerman, line 412: why moving this call?

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -367,13 +422,7 @@ void BaseQNAcceleration::performAcceleration(
      * PRECONDITION: All objects are unscaled, except the matrices within the QR-dec of V.
      *               Thus, the pseudo inverse needs to be reverted before using it.
      */
-    Eigen::VectorXd xUpdate = Eigen::VectorXd::Zero(_residuals.size());
-    computeQNUpdate(cplData, xUpdate);
-
-    /**
-     * apply quasiNewton update
-     */
-    _values = _oldValues + xUpdate + _residuals; // = x^k + delta_x + r^k - q^k
```
uekerman, line 376: Where did this line go?

uekerman (commented):
Sry for the very late reply.
Using the switch is a meaningful temporary solution. Let's reevaluate in the end.

I think that the there could be a more elegant overall solution. The main thing we need to do is to transform the waveform into a long vector, discretizing in time somehow. Couldn't we do this directly in `BaseQNAcceleration`? Then, it should directly work for all QN variants. Maybe I don't yet understand some of the difficulties?

NiklasKotarsky:
> Sry for the very late reply. Using the switch is a meaningful temporary solution. Let's reevaluate in the end.
> 
> I think that the there could be a more elegant overall solution. The main thing we need to do is to transform the waveform into a long vector, discretizing in time somehow. Couldn't we do this directly in `BaseQNAcceleration`? Then, it should directly work for all QN variants. Maybe I don't yet understand some of the difficulties?

Thanks for the review! The main difficulty is that preCICE does not assume anything regarding the time grids, which currently might change in every iteration. This is the case in my application with time adaptive sub solvers. This makes it difficult to to transform the waveform into a long vector, since the number and position of the time step might change in every iteration. My solution to this problem has been to store the wave forms and then interpolate everything to the current time grid. 

The downside of this solution is that it forces one to duplicate a lot of code in all QN classes that support both. I have since refactored the IQNILS class and replaced the W matrix with the waveform version, since this works for both cases.

src/acceleration/BaseQNAcceleration.cpp
NiklasKotarsky: Yes and also to make things easier to find for me, since I initially would search for such a function in the QN classes. I can remove it again if this does not improve the readability?

uekerman:
Did you ever think about using a fixed uniform user-defined time grid within each time window for the QN matrices (e.g. 5 slices) and interpolate the new data always on this grid?
This could make the implementation easier and more efficient as well. Do you see numerical pitfalls?

src/acceleration/BaseQNAcceleration.cpp
BenjaminRodenberg: I also just stumbled over the line `_values = _oldValues + xUpdate + _residuals; // = x^k + delta_x + r^k - q^k` and think it is pretty hard to follow how things are being updated here (btw. what is `q^k`?). Something like `x_kp1 = xUpdate + _values` or just `_values += xUpdate` would be much clearer from my point of view. I just tried it out and according to our tests it seems to work (see https://github.com/precice/precice/pull/1979).

NiklasKotarsky:
> Sry for the very late reply. Using the switch is a meaningful temporary solution. Let's reevaluate in the end.
> 
> I think that the there could be a more elegant overall solution. The main thing we need to do is to transform the waveform into a long vector, discretizing in time somehow. Couldn't we do this directly in `BaseQNAcceleration`? Then, it should directly work for all QN variants. Maybe I don't yet understand some of the difficulties?

I have looked a little bit into this, but there are a number of different ways one could do this. Going from only using predefined time grid within preCICE for everything to not making any assumptions on the time grid at all. 

Using resampling or interpolation to a coarser time grid in the QN schemes would not affect the overall computational accuracy, since convergence is checked before QN accelration is applied. However, it will affect the convergence behaviour and in extreme cases lead to divergence. From my experience there are edge cases where rQN converges much slower than full QN. Thus, in my opinion there is value in having both a much slower and robust algorithm (full QN) as well as having faster variants such as rQN.

NiklasKotarsky:
I have now added minimal support for waveform to IMVJ in order to refactor the base QN acceleration class. Adding full waveform support to IMVJ seems to me to be out of scope for this pull request, which already is quite big. 

I have also moved the update of the _values/cplData to the function computeQNupdate in IQNILS and IMVJ class. The reasoning behind this is:

1. Having a unified update of the _values/cplData in the base class is problematic if not all of the schemes have full waveform support. This would also make it a little bit more difficult to add new QN variants that do not support waveforms yet. 

2. It feels a little bit clunky to create a waveform version of xUpdate with the correction $\Delta x^k$ of every time step in the QNILSA and IMVJ class and then sending that to the base class in order to apply the updated there to cplData. 

But, I could also go for with the solution in 2. if this works better from a class structure perspective.

uekerman (changes requested):
Hi @NiklasKotarsky 

I had a longer look at this PR and the quasi-Newton implementation in general. This PR is really hard to review, but this is not your fault. It is due to the complexity we already had before. Maybe we could have done a better job here in the past.

I now understand and agree that rQNWI for IMVJ is not straight-forward. Also not math-wise as it is unclear how to treat secondary data for IMVJ, see #1996. So, the overall design goal has shifted for me: Move things as much as possible into the IQNILS subclass and keep changes in IMVJ and base as minimal as possible.

Two things here introduce additional complexity that we should try to avoid if possible:
- `std::map<int, std::vector<precice::time::Storage>> _waveformW;` instead of `Eigen::MatrixXd _matrixW;`
- Remove the returning `xUpdate` from `virtual void computeQNUpdate(const DataMap &cplData, Eigen::VectorXd &xUpdate)` (we should actually make it a return value)

Two suggestions:
- [ ] Let's do a refactoring or clean up PR before this one. You already fix small details, typos, wrong documentation. There is more, I also spotted quite a few mistakes. Goal is to keep the actual rQNWI as minimal as possible. Could you open a PR with your changes and then I add mine on top?
- [ ] Could we implement the new functionality by treating the intermediate time steps similar to secondary data? Only creating waveforms in the "rescaling" (we need a better name). I have not thought this completely through. But maybe we could then avoid the two complexities above, not touch the IMVJ subclass and move all waveform things into the IQNILS subclass?

Misc:
- We should unit test new functionality such as `rescaleWaveformInTime`.
- Please check the compiler warnings reported by the CI.
- Could we add a unit test that actually checks that all substeps are accelerated the same way?
- Needs integration tests for parallel-implicit coupling and also combination with preconditioner.

NiklasKotarsky:
Thanks for the review @uekerman. While the quasi-Newton implementation is quite complex machinery in preCICE with a lot of moving parts, I have the feeling that my idea for rQNWR and full QNWR algorithm might be a little bit to complex. I am currently investigating if it would be possible to only use 1 time grid inside QNWR method. There are some mathematical downsides resulting in worse convergence, but from  a programming perspective it should simplify the implementation by a couple of orders of magnitude making it a worthwhile tradeoff. 



>     * [ ]  Let's do a refactoring or clean up PR before this one. You already fix small details, typos, wrong documentation. There is more, I also spotted quite a few mistakes. Goal is to keep the actual rQNWI as minimal as possible. Could you open a PR with your changes and then I add mine on top?

Yes, I have opened a second pull request #2001 and moved the small things I found over to it.

NiklasKotarsky:
Replaced by  #2005, since this rQNWR algorithm is to complex from a programming perspective to maintain. The pull request in #2005 only use 1 time grid inside QNWR method and is thus much simpler. There are some mathematical downsides resulting in worse convergence, but from a programming perspective it should simplify the implementation by a couple of orders of magnitude making it a worthwhile tradeoff.

BenjaminRodenberg:
I tested the partitioned heat equation (https://github.com/precice/tutorials/commit/15338dea52f3f512f081eed69eb36fed833a2b78) with the quasi-Newton implementation provided here and it did not converge. The following changes are applied to `partitioned-heat-conduction/precice-config.xml` compared to https://github.com/precice/tutorials/commit/15338dea52f3f512f081eed69eb36fed833a2b78:

```
     <relative-convergence-measure data="Heat-Flux" mesh="Dirichlet-Mesh" limit="1e-10" />
     <relative-convergence-measure data="Temperature" mesh="Neumann-Mesh" limit="1e-10" />
-    <!-- <acceleration:IQN-ILS>
+    <acceleration:IQN-ILS>
       <data name="Temperature" mesh="Neumann-Mesh" />
       <max-used-iterations value="10" />
-      <time-windows-reused value="5" />
+      <!-- <time-windows-reused value="5" /> -->
       <filter type="QR2" limit="1e-3" />
-    </acceleration:IQN-ILS> -->
-    <acceleration:constant>
+    </acceleration:IQN-ILS>
+    <!-- <acceleration:constant>
       <relaxation value="0.5" />
-    </acceleration:constant>
+    </acceleration:constant> -->
 </precice-configuration>
```

Remark: With `substeps="false"` and `acceleration:IQN-ILS` it also does not converge. So there is actually a regression compared to preCICE v3.1.1 (924849475104090287b935d813e8e8ec435f19ba). This is a gap in our tests. Not sure how hard it is to close it with a unit or integration test. The simple way how we could do it is with a system test that runs the complete partitioned heat equation with quasi-Newton and `substeps="false"`. But this comes with some overhead.

I will now migrate to #2005 and repeat my tests there.

NiklasKotarsky:
Thanks, I also tried to run the heat conduction benchmark with this code and it only worked for 1 time window. There is also no current test with filtering and multiple time windows for QN-ILS, which probably explains the convergence issue. I also noticed that rQNWR struggled to converge when the spatial grid is coarse, which certainly is the case for the heat equation.

PR #1837, opened 2023-10-16
Do not interpolate data for existing timestamps

### LINKED ISSUE #1752: Do not interpolate data for existing timestamps

NiklasKotarsky:
## Main changes of this PR
Changes storage so that it returns the existing values instead of interpolating them. 

Fixes  #1752

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/time/Storage.cpp
+++ b/src/time/Storage.cpp
@@ -163,17 +163,15 @@ Eigen::VectorXd Storage::sample(double time) const
 
   PRECICE_ASSERT(usedDegree >= 1);
 
-  auto [times, values] = getTimesAndValues();
-
   //Return the sample corresponding to time if it exists
-  for (int i = 0; i < times.size(); i++) {
-    if (math::equals(times[i], time)) {
-      return values.col(i);
-    }
+  const int i = findTimeId(time);
+  if (i > -1) {                              // _stampleStorage contains sample at given time
+    return _stampleStorage[i].sample.values; // don't use getTimesAndValues, because this would iterate over the complete _stampleStorage.
   }
 
   //Create a new bspline if _bspline does not already contain a spline
   if (!_bspline.has_value()) {
+    auto [times, values] = getTimesAndValues();
     _bspline.emplace(times, values, usedDegree);
   }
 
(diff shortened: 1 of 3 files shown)


src/time/Storage.cpp
```diff
@@ -163,10 +163,18 @@ Eigen::VectorXd Storage::sample(double time) const
...
 
-  auto data = getTimesAndValues();
+  auto [times, values] = getTimesAndValues();
+
+  //Return the sample corresponding to time if it exists
+  for (int i = 0; i < times.size(); i++) {
+    if (math::equals(times[i], time)) {
+      return values.col(i);
+    }
+  }
```
BenjaminRodenberg, line 173: This could be implemented more efficiently: `times` are ascending. Therefore, I would suggest something in this direction:
```
int i = 0;
while(math::smallerEquals(times[i], time) {
    if(math::equals(times[i], time)) {
        return values.col(i);
    }
    i++;
}
```
Maybe even put this into a private function of `Storage`:
```
int findTimeId(time) const {
    int i = 0;
    while(math::smallerEquals(_stampleStorage[i].timestamp, time) {
        if(math::equals(_stampleStorage[i].timestamp, time)) {
        return return i;
        }
    i++;
    }
    return -1;  // time not found in times
}
```

BenjaminRodenberg: If we start with efficiency here: You can move `getTimesAndValues()` inside the `if(!_bspline.has_value())` branch, because you don't need it outside. This results in

```
Eigen::VectorXd Storage::sample(double time) const {
    ...
    const int i = findTimeId(time);
    if(i > -1) {  // _stampleStorage contains sample at given time
        return _stampleStorage[i].sample;  // don't use getTimesAndValues, because this would iterate over the complete _stampleStorage.
    }

    if(!_bspline.has_value()) {
        auto [times, values] = getTimesAndValues();
        _bspline.emplace(times, values, usedDegree);
    }
    
    return _bspline.value().interpolateAt(time);
}
```

BenjaminRodenberg (changes requested):
Generally looks good. Minor comments, but nothing critical.

src/time/Storage.cpp
fsimonis: Both changes together make a lot of sense :+1:.

We don't need a sentinel value, now that we have optional though `std::optional<int> findTime(double time) const;`

fsimonis (approved):

Issue #1838, opened 2023-10-16
Assertion failure for time step size <= 10e-6


BenjaminRodenberg:
**Describe your setup**

preCICE Version: https://github.com/precice/precice/commit/d1d4097fa5992fbc4d63092cf3c3020e8e702816

**Describe the problem**

Small time step sizes lead to a failing assertion, if we are reading data. Originally observed by @IshaanDesai in a more complicated setup.

**Step To Reproduce**

See tests https://github.com/precice/precice/pull/1839

**Expected behaviour**

No assertion triggered.

BenjaminRodenberg:
This bug was fixed as a side-effect via #1837 (integration level). We will still merge #1839 soon, because it avoids the underlying problem on the level of `math::Bspline` (unit testable).

PR #1839, opened 2023-10-16
Fix roundoff problem if close to window boundary in `math::Bspline`

BenjaminRodenberg:
## Main changes of this PR

Compare `t` to `_tsMin` and `_tsMax` within floating-point accuracy. If `t` equals `_tsMin` or `_tsMax`, set `t` to this value. This avoids round-off errors in later steps that can trigger assertions.

See #1838. Especially important for adaptive time stepping where time window sizes might become small. @NiklasKotarsky 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/math/Bspline.cpp
+++ b/src/math/Bspline.cpp
@@ -57,6 +57,12 @@ Bspline::Bspline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree
 
 Eigen::VectorXd Bspline::interpolateAt(double t) const
 {
+  if (math::equals(t, _tsMax)) {
+    t = _tsMax; // set t exactly to _tsMax, to avoid artifacts originating from floating-point arithmetic
+  } else if (math::equals(t, _tsMin)) {
+    t = _tsMin; // set t exactly to _tsMin, to avoid artifacts originating from floating-point arithmetic
+  }
+
   // transform t to the relative interval [0; 1]
   const double tRelative = (t - _tsMin) / (_tsMax - _tsMin);
   PRECICE_ASSERT(math::greaterEquals(tRelative, 0.0), "t is before the first sample!", tRelative, 0.0, _tsMin); // Only allowed to use BSpline for interpolation, not extrapolation.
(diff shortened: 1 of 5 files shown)


BenjaminRodenberg:
@NiklasKotarsky Any idea how to fix this? I already have a possible work around in mind, but since you are currently deep inside the floating-point accuracy rabbit hole, I think you might know a proper solution. Feel free to directly edit on this branch.

BenjaminRodenberg:
@NiklasKotarsky I think I found a nice fix. A review would still be helpful :)

src/math/Bspline.cpp
```diff
@@ -57,6 +57,12 @@ Bspline::Bspline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree
 
 Eigen::VectorXd Bspline::interpolateAt(double t) const
 {
+  if (math::equals(t, _tsMax)) {
+    t = _tsMax; // set t exactly to _tsMax, to avoid artifacts originating from floating-point arithmetic
```
NiklasKotarsky, line 61: This line might become unreachable when we merge  #1837. Due to that storage always will return the sample when math::equals(t,_tsMax) returns true. I still think that these lines are worth keeping, since they add a safeguard if we change the implementation in storage for when we return the existing sample and when we interpolate.

NiklasKotarsky (approved):
Looks good! The test also work on the current version of pull request  #1837, but we might want to change the tolerance when we return a existing sample there in the future.

src/math/Bspline.cpp
BenjaminRodenberg: Ok. Then let's merge this PR. Then you can integrate it into #1837 and we can discuss possible redundancies / safeguards over there. I agree with your argument about making the correct functionality of the Bspline class as independent as possible from the implementation of the Storage class.

BenjaminRodenberg: @NiklasKotarsky Added a unit test via 7bae79cf78033ab843d53fdfc39ef11ee0d27ef8. Can you also try this one on your branch? I guess this one should fail, because you are dealing with the issue on the level of the `Storage` and not on the `Bspline`.

BenjaminRodenberg: I just did this on my own, since #1837 was already merged into develop. No bad surprises. I'm just waiting for the test pipeline here to finish and will merge then.

NiklasKotarsky: Yes, this test fails on my branch.

PR #1841, opened 2023-10-17
Convert min-iterations-convergence to tag

### LINKED ISSUE #800: Convert min-iterations convergence measure to plain tag
Currently the min-iterations convergence measure needs to be associated to some data even though there is no real connection.
```xml
<min-iteration-convergence-measure data="CrossSectionLength" mesh="Structure_Nodes" min-iterations="2"/>
```

This cannot be fixed "directly" on the configuration level as convergence measures enforce having data:

https://github.com/precice/precice/blob/6437f8e43710818d6ff96074d16e93fca8604088/src/cplscheme/impl/ConvergenceMeasure.hpp#L36-L38

I suggest to convert the convergence measure to a simple tag of `coupling-scheme`, similar to `max-iterations`
```xml
      <participants first="FLUID" second="STRUCTURE"/>
      ...
      <max-iterations value="40"/>
      <min-iterations value="5"/>
      ...
```

Drawback of this is that then "min iterations" can no longer be combined with `suffices="true"` to any other convergence measure. In my opinion this loss, however, is acceptable. I am not sure if anybody uses that feature anyway.

Should we do this?

fsimonis:
## Main changes of this PR

This PR replaces the min-iterations-convergence measure to a simple tag and adds support for no convergence measures.

* reaching min-iterations is a logical and with "convergence"
* No convergence measures means "never converge".
* So min-iterations without convergence measures has no function.
* Max iterations default to infinite, which leads to an error, when used with no-convergence measure.

Closes #800 

## Author's checklist

## Reviewers' checklist


docs/changelog/1841.md
```diff
@@ -0,0 +1,3 @@
+- Removed `min-iteration-convergence-measure`, use the new `min-iterations` tag or `max-iterations` instead.
+- Added support for implicit coupling schemes without convergence measures, which will always iterates `max-iterations` times per time window.
```
BenjaminRodenberg, line 2:
```suggestion
- Added support for implicit coupling schemes without convergence measures, which will always iterate `max-iterations` times per time window.
```

docs/changelog/1841.md
```diff
@@ -0,0 +1,3 @@
...
+- Added tag `min-iterations` to coupling schemes, which allows defining a minimal amount of iterations before convergence measures are taken into account.
```
BenjaminRodenberg, line 3: What happens, if there are no `max-iterations`? I would expect a brief comment here.

examples/solverdummies/precice-config.xml
```diff
@@ -51,8 +51,8 @@
     <participants first="SolverOne" second="SolverTwo" />
     <max-time-windows value="2" />
     <time-window-size value="1.0" />
+    <min-iterations value="2" />
     <max-iterations value="2" />
-    <min-iteration-convergence-measure min-iterations="5" data="Data-One" mesh="SolverOne-Mesh" />
```
BenjaminRodenberg, line 55: Why was this 5 min-iterations before? Looks like it was a bit buggy before. Intuitively, I would expect only min iterations in this simple case where we want to say "do exactly X iterations". But this contradicts a bit what we already discussed, right?

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -63,14 +65,19 @@ BaseCouplingScheme::BaseCouplingScheme(
                    "Time window size has to be given when the fixed time window size method is used.");
   }
 
-  PRECICE_ASSERT((maxIterations > 0) || (maxIterations == UNDEFINED_MAX_ITERATIONS),
-                 "Maximal iteration limit has to be larger than zero.");
-
   if (isExplicitCouplingScheme()) {
+    PRECICE_ASSERT(minIterations == UNDEFINED_MAX_ITERATIONS);
```
BenjaminRodenberg, line 69:
```suggestion
    PRECICE_ASSERT(minIterations == UNDEFINED_MIN_ITERATIONS);
```
?

BenjaminRodenberg (commented):
Just a very quick review with some first impressions. Will probably do a more in-depth review tomorrow.

examples/solverdummies/precice-config.xml
fsimonis: Maybe we really should implement some actual super simple functionality here.

docs/changelog/1841.md
```diff
@@ -0,0 +1,3 @@
+- Removed `min-iteration-convergence-measure`, use the new `min-iterations` tag or `max-iterations` instead.
+- Added support for implicit coupling schemes without convergence measures, which will always iterate `max-iterations` times per time window. Not defining convergence measures now requires `max-iterations` to be defined.
```
BenjaminRodenberg, line 2: Open for discussion, see #800

examples/solverdummies/precice-config.xml
```diff
@@ -51,8 +51,8 @@
     <participants first="SolverOne" second="SolverTwo" />
     <max-time-windows value="2" />
     <time-window-size value="1.0" />
+    <min-iterations value="2" />
     <max-iterations value="2" />
```
BenjaminRodenberg, line 55: Open for discussion, see #800

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -621,14 +628,24 @@ bool BaseCouplingScheme::measureConvergence()
...
-  bool oneStrict    = false; // at least one convergence measure is strict and did not converge
-  PRECICE_ASSERT(_convergenceMeasures.size() > 0);
+  const bool reachedMinIterations = _iterations >= _minIterations;
   if (not utils::IntraComm::isSecondary()) {
     _convergenceWriter->writeData("TimeWindow", _timeWindows - 1);
     _convergenceWriter->writeData("Iteration", _iterations);
+    _convergenceWriter->writeData("ReachedMinIter", reachedMinIterations ? 1 : 0);
+  }
+
+  // If no convergence measures are define, we never converge
```
BenjaminRodenberg, line 638:
```suggestion
  // If no convergence measures are defined, we never converge
```

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -621,14 +628,24 @@ bool BaseCouplingScheme::measureConvergence()
...
   PRECICE_ASSERT(not doesFirstStep());
-  bool allConverged = true;
-  bool oneSuffices  = false; // at least one convergence measure suffices and did converge
-  bool oneStrict    = false; // at least one convergence measure is strict and did not converge
-  PRECICE_ASSERT(_convergenceMeasures.size() > 0);
+  const bool reachedMinIterations = _iterations >= _minIterations;
   if (not utils::IntraComm::isSecondary()) {
     _convergenceWriter->writeData("TimeWindow", _timeWindows - 1);
     _convergenceWriter->writeData("Iteration", _iterations);
+    _convergenceWriter->writeData("ReachedMinIter", reachedMinIterations ? 1 : 0);
```
BenjaminRodenberg, line 635: Do we really need this additional information in the output? I think this part is somehow trivial.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -681,6 +702,7 @@ void BaseCouplingScheme::initializeTXTWriters()
     if (not doesFirstStep()) {
       _convergenceWriter->addData("TimeWindow", io::TXTTableWriter::INT);
       _convergenceWriter->addData("Iteration", io::TXTTableWriter::INT);
+      _convergenceWriter->addData("ReachedMinIter", io::TXTTableWriter::INT);
```
BenjaminRodenberg, line 705: See above. Additionally, there seems to be quite some code duplication in this part.

src/cplscheme/CouplingScheme.cpp
```diff
@@ -9,6 +9,7 @@ const int CouplingScheme::UNDEFINED_TIME_WINDOWS = -1;
 const double CouplingScheme::UNDEFINED_TIME_WINDOW_SIZE = -1.0;
 
 const int CouplingScheme::UNDEFINED_MAX_ITERATIONS = -1;
```
BenjaminRodenberg, line 11: Isn't the key for `UNDEFINED_MAX_ITERATIONS` now identical with the key for an infinite number of iterations? I think it would be good to define another specific `const int` for this purpose.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -63,14 +65,19 @@ BaseCouplingScheme::BaseCouplingScheme(
...
     PRECICE_ASSERT(maxIterations == UNDEFINED_MAX_ITERATIONS);
   } else {
     PRECICE_ASSERT(isImplicitCouplingScheme());
-    PRECICE_ASSERT(maxIterations >= 1);
+    PRECICE_ASSERT(minIterations > 0,
+                   minIterations,
+                   "Minimal iteration limit has to be larger than zero.");
+    PRECICE_ASSERT((maxIterations == UNDEFINED_MAX_ITERATIONS) || (maxIterations > 0),
+                   maxIterations,
+                   "Maximal iteration limit has to be larger than zero or -1 (unlimited).");
```
BenjaminRodenberg, line 78: Isn't `-1` already occupied by `UNDEFINED_MAX_ITERATIONS`?

src/cplscheme/SerialCouplingScheme.hpp
```diff
@@ -50,7 +50,20 @@ class SerialCouplingScheme : public BiCouplingScheme {
       m2n::PtrM2N                   m2n,
       constants::TimesteppingMethod dtMethod,
       CouplingMode                  cplMode,
-      int                           maxIterations = UNDEFINED_MAX_ITERATIONS);
+      int                           minIterations,
+      int                           maxIterations);
```
BenjaminRodenberg, line 54: Why do we not use default values here anymore?

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -279,11 +269,28 @@ void CouplingSchemeConfiguration::xmlTagCallback(
     _meshConfig->addNeededMesh(nameParticipantFrom, nameMesh);
     _meshConfig->addNeededMesh(nameParticipantTo, nameMesh);
     _config.exchanges.emplace_back(std::move(newExchange));
+  } else if (tag.getName() == TAG_MIN_ITERATIONS) {
+    PRECICE_ASSERT(_config.type == VALUE_SERIAL_IMPLICIT || _config.type == VALUE_PARALLEL_IMPLICIT || _config.type == VALUE_MULTI);
+    _config.minIterations = tag.getIntAttributeValue(ATTR_VALUE);
+    PRECICE_CHECK(_config.minIterations > 0,
+                  "Maximum iteration limit has to be larger than zero. Please check the <min-iterations value = \"{}\" /> subtag in the <coupling-scheme:... /> of your precice-config.xml.",
```
BenjaminRodenberg, line 276:
```suggestion
                  "Minimum iteration limit has to be larger than zero. Please check the <min-iterations value = \"{}\" /> subtag in the <coupling-scheme:... /> of your precice-config.xml.",
```

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -599,11 +591,24 @@ void CouplingSchemeConfiguration::addBaseAttributesTagConvergenceMeasure(
   tag.addAttribute(attrStrict);
 }
 
+void CouplingSchemeConfiguration::addTagMinIterations(
+    xml::XMLTag &tag)
+{
+  using namespace xml;
+  XMLTag tagMinIterations(*this, TAG_MIN_ITERATIONS, XMLTag::OCCUR_NOT_OR_ONCE);
+  tagMinIterations.setDocumentation("Allows to specify a minimum amount of iterations per time window at from which point convergence measures are considered. "
+                                    "Note that not defining convergence measures is interpreted as \"no convergence\".");
```
BenjaminRodenberg, line 600:
```suggestion
  tagMinIterations.setDocumentation("Allows to specify a minimum amount of iterations that must be performed per time window.");
```

BenjaminRodenberg: I think mixing what `min-iterations` and convergence measures here is not really necessary and overly complicated. If I do not define a convergence measure then I should also not have to worry about it, no?

The information how convergence measures and min/max iterations interact should come somewhere else.

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -916,6 +894,20 @@ CouplingSchemeConfiguration::getTimesteppingMethod(
   }
 }
 
+void CouplingSchemeConfiguration::checkIterationLimits() const
+{
+  if (_config.convergenceMeasureDefinitions.empty()) {
+    PRECICE_CHECK(_config.maxIterations != -1,
+                  "No convergence measures without a maximum iteration limit leads to endless iterations. "
+                  "Please define a convergence measure or set a maximum iteration limit using <max-iterations value=\"...\" />.");
```
BenjaminRodenberg, line 902:
```suggestion
                  "Not defining convergence measures without providing a maximum iteration limit is forbidden."
                  "Please define a convergence measure or set a maximum iteration limit using <max-iterations value=\"...\" />.");
```
I think giving the reason here is a bit an overkill.

src/cplscheme/config/CouplingSchemeConfiguration.hpp
```diff
@@ -156,7 +155,8 @@ class CouplingSchemeConfiguration : public xml::XMLTag::Listener {
     };
     std::vector<Exchange>                    exchanges;
     std::vector<ConvergenceMeasureDefintion> convergenceMeasureDefinitions;
-    int                                      maxIterations = -1;
+    int                                      maxIterations = -1; // infinite by default
```
BenjaminRodenberg, line 158: See above. I think using a `const int INFINITE_MAX_ITERATIONS = -2` would be good here. Currently also overlaps with `UNDEFINED_MAX_ITERATIONS`.

src/cplscheme/tests/ParallelImplicitCouplingSchemeTest.cpp
```diff
@@ -14,7 +14,7 @@
 #include "cplscheme/ParallelCouplingScheme.hpp"
 #include "cplscheme/SharedPointer.hpp"
 #include "cplscheme/config/CouplingSchemeConfiguration.hpp"
-#include "cplscheme/impl/MinIterationConvergenceMeasure.hpp"
+#include "cplscheme/impl/AbsoluteConvergenceMeasure.hpp"
```
BenjaminRodenberg, line 17: Where does this come from? Does not really look like it belongs to this PR. Maybe consider just adding `#include "cplscheme/impl/AbsoluteConvergenceMeasure.hpp"` directly on `develop` to clean up the diff a bit.

src/cplscheme/tests/ParallelImplicitCouplingSchemeTest.cpp
```diff
@@ -115,7 +115,7 @@ BOOST_AUTO_TEST_CASE(testInitializeData)
   // Create the coupling scheme object
   ParallelCouplingScheme cplScheme(
       maxTime, maxTimeWindows, timeWindowSize, 16, nameParticipant0, nameParticipant1,
-      context.name, m2n, constants::FIXED_TIME_WINDOW_SIZE, BaseCouplingScheme::Implicit, 100);
+      context.name, m2n, constants::FIXED_TIME_WINDOW_SIZE, BaseCouplingScheme::Implicit, 1, 3);
```
BenjaminRodenberg, line 118: I generally prefer the following, because it's easier to read and to find-and-replace without introducing errors:

```
  const int minIterations = 1;
  const int maxIterations= 3;
  ...
  ParallelCouplingScheme cplScheme(
      maxTime, maxTimeWindows, timeWindowSize, 16, nameParticipant0, nameParticipant1,
      context.name, m2n, constants::FIXED_TIME_WINDOW_SIZE, BaseCouplingScheme::Implicit, minIterations, maxIterations);
```

But probably arguable.

tests/serial/convergence-measures/testConvergenceMeasures1.xml
```diff
@@ -45,7 +45,6 @@
     <exchange data="Data1" mesh="MeshOne" from="SolverOne" to="SolverTwo" />
     <exchange data="Data2" mesh="MeshOne" from="SolverTwo" to="SolverOne" />
     <max-iterations value="100" />
-    <absolute-convergence-measure limit="1e-1" data="Data2" mesh="MeshOne" suffices="true" />
-    <min-iteration-convergence-measure min-iterations="3" data="Data1" mesh="MeshOne" />
+    <absolute-convergence-measure limit="1e-1" data="Data2" mesh="MeshOne" />
```
BenjaminRodenberg, line 48:
```suggestion
    <absolute-convergence-measure limit="1e-1" data="Data2" mesh="MeshOne" />
    <min-iterations value="3" />
```

tests/serial/convergence-measures/testConvergenceMeasures2.xml
```diff
@@ -44,12 +44,7 @@
...
     <exchange data="Data2" mesh="MeshOne" from="SolverTwo" to="SolverOne" />
-    <max-iterations value="100" />
-    <absolute-convergence-measure limit="1e-1" data="Data2" mesh="MeshOne" />
-    <min-iteration-convergence-measure
-      min-iterations="3"
-      data="Data1"
-      mesh="MeshOne"
-      suffices="true" />
+    <max-iterations value="3" />
+    <absolute-convergence-measure limit="1e-3" data="Data2" mesh="MeshOne" />
```
BenjaminRodenberg, line 48:
```suggestion
    <absolute-convergence-measure limit="1e-3" data="Data2" mesh="MeshOne" />
    <min-iterations value="3" />
```

tests/serial/convergence-measures/testConvergenceMeasures3.xml
```diff
@@ -44,12 +44,7 @@
     <time-window-size value="1.0" />
     <exchange data="Data1" mesh="MeshOne" from="SolverOne" to="SolverTwo" />
     <exchange data="Data2" mesh="MeshOne" from="SolverTwo" to="SolverOne" />
-    <max-iterations value="100" />
+    <max-iterations value="4" />
```
BenjaminRodenberg, line 47:
```suggestion
    <max-iterations value="100" />
    <min-iterations value="2" />
```
Why `max-iterations value="4" />`? I wouldn't change the test here, if it is not necessary.

tests/serial/initialize-data/Implicit.xml
```diff
@@ -43,7 +43,6 @@
     <max-time-windows value="10" />
     <time-window-size value="1.0" />
     <max-iterations value="3" />
-    <min-iteration-convergence-measure min-iterations="5" data="Forces" mesh="MeshOne" />
```
BenjaminRodenberg, line 46: Also add `min-iterations value="5"`

tests/serial/initialize-data/ImplicitBoth.xml
```diff
@@ -42,8 +42,8 @@
     <participants first="SolverOne" second="SolverTwo" />
     <max-time-windows value="10" />
     <time-window-size value="1.0" />
+    <min-iterations value="3" />
```
BenjaminRodenberg, line 45:
```suggestion
    <min-iterations value="5" />
```

BenjaminRodenberg: What happened in this test before? Did we use min or max iterations? Maybe also something we should consider cleaning up on `develop` first before touching it here.

tests/serial/three-solvers/ThreeSolversExplicitImplicit.xml
```diff
@@ -62,8 +62,7 @@
     <participants first="SolverOne" second="SolverThree" />
     <max-time-windows value="10" />
     <time-window-size value="1.0" />
-    <max-iterations value="10" />
-    <min-iteration-convergence-measure min-iterations="3" data="Data2" mesh="MeshA" />
+    <max-iterations value="3" />
```
BenjaminRodenberg, line 65:
```suggestion
    <min-iterations value="3" />
    <max-iterations value="10" />
```

tests/serial/three-solvers/ThreeSolversImplicitExplicit.xml
```diff
@@ -54,8 +54,7 @@
     <participants first="SolverOne" second="SolverTwo" />
     <max-time-windows value="10" />
     <time-window-size value="1.0" />
-    <max-iterations value="10" />
-    <min-iteration-convergence-measure min-iterations="3" data="Data1" mesh="MeshA" />
+    <max-iterations value="3" />
```
BenjaminRodenberg, line 57:
```suggestion
    <min-iterations value="3" />
    <min-iterations value="10" />
```

BenjaminRodenberg (changes requested):
I think we should really conclude in #800. The updates in some of the configs are a bit hard to follow and I think a part of the problem for me is that the way how we deal with "no convergence measure" and how we then enforce "do exactly X iterations" is not really intuitive.

From the technical side most of the changes look good. Removing the convergence measure is definitely a good step.

BenjaminRodenberg:
I fixed the inconsistency in the solverdummies config directly via 0d36758ebb45ae1a3b7dc03368c10104d85ffdbe, because it is just making discussions more complicated.

src/cplscheme/BaseCouplingScheme.cpp
fsimonis: Trivial yes, but I would say also say it's useful for the reason of missing context.

Without `ReachedMinIter`, the file may state that everything has converged, yet the scheme continues to iterate. So, one needs to have a look at the configuration file to get the full picture.

Also, from the perspective of support, the convergence file now contains the information that the user configured a minimum of n iterations. So, the user only needs to provide one file.

An alternative to the boolean `ReachedMinIter` would be to simply add the `Min` value as a constant column. So the first columns of the file would look like this:

```
0 1 3
0 2 3
0 3 3
0 4 3
1 1 3
1 2 3
1 3 3
2 1 3
...
```

src/cplscheme/CouplingScheme.cpp
fsimonis: No iterations given is interpreted as infinite iterations. So, conceptually nothing changed. It's only interpreted differently.

src/cplscheme/BaseCouplingScheme.cpp
fsimonis: See https://github.com/precice/precice/pull/1841#discussion_r1372878125

src/cplscheme/SerialCouplingScheme.hpp
fsimonis: Two defaults are super easy to misuse. So I added a second constructor for explicit cases which doesn't take iterations at all.

src/cplscheme/tests/ParallelImplicitCouplingSchemeTest.cpp
fsimonis: Oh. I used a nonsense absolute measure before adding support for no conv measures.

tests/serial/initialize-data/Implicit.xml
fsimonis: min > max doesn't make sense :wink:

tests/serial/three-solvers/ThreeSolversExplicitImplicit.xml
fsimonis: After the conclusion in #800, max=3 is correct.

tests/serial/three-solvers/ThreeSolversImplicitExplicit.xml
fsimonis: Same as https://github.com/precice/precice/pull/1841#discussion_r1372892542

tests/serial/convergence-measures/testConvergenceMeasures1.xml
fsimonis: The absolute measure `sufficed` for convergence, so the `min-iteration-convergence` was  completely pointless.
Maybe testing exactly that was the point of the test. I don't know, which why I opened #1846

tests/serial/convergence-measures/testConvergenceMeasures2.xml
fsimonis: The min-iter-c-m `suffices`, so this always only iterates to 3 iterations.
Again maybe that was the point here. #1846

examples/solverdummies/precice-config.xml
fsimonis: After the conclusion in #800, the `min-itertations` should go.

src/cplscheme/BaseCouplingScheme.cpp
fsimonis: See https://github.com/precice/precice/pull/1841#discussion_r1372875827

src/cplscheme/config/CouplingSchemeConfiguration.hpp
fsimonis: Done, I used `UNDEFINDED_MXX_ITERATIONS` for both. For implicit cases, I convert `UNDEFINED_MAX_ITERATIONS` to `INFINITE_MAX_ITERATIONS`

tests/serial/initialize-data/ImplicitBoth.xml
fsimonis: Tests implicit coupling with initializeData on both exchanges. Iterations don't seem to matter here.

tests/serial/convergence-measures/testConvergenceMeasures3.xml
fsimonis: The test will SEGFAULT with more than 4 iterations. Given that the test will just run until the end, this should lead to the same result.

fsimonis:
I addressed the majority of comments.


#1846 is really necessary at some point. There are many tests, which don't really make sense anymore. Also, the point of the entire convergence measure tests are oblivious to me.

src/cplscheme/BaseCouplingScheme.cpp
fsimonis: After @uekerman voiced the same concerns, I removed the column.

PR #1847, opened 2023-10-23
Fix deadlock in two-level initialization

davidscn:
## Main changes of this PR
Fixes a deadlock in the two-level initialization, which occurs if ranks share vertices (or vertices are considered 'shared' when creating the owner information), but only one of the ranks actually needs the vertices to compute the mapping. In such a case, the rank which does not need the vertex never tells the partner rank the empty size. By initializing all coupling partner pairs with a zero vector, we can ensure that zero sized vectors are exchanged in any case.

It's somewhat surprising to me, that this error has never shown up before. However, I added an integration test to reproduce, showcase and safeguard against it for future references. 

I also converted the (somewhat erroneous) warning about filtered vertices into a debug log and changed the referenced number of filtered vertices from global to local. Raising a warning here doesn't make sense in my opinion and users would only be able to see information (by default of course) from the primary rank, which is again misleading.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/1847.md
+++ b/docs/changelog/1847.md
@@ -1,0 +1,1 @@
+- Fixed deadlock in two-level initialization, which occurs if ranks share vertices, but only one of the ranks actually needs the vertices (or any) to compute the mapping.
--- a/src/partition/ReceivedPartition.cpp
+++ b/src/partition/ReceivedPartition.cpp
@@ -743,10 +743,8 @@ void ReceivedPartition::createOwnerInformation()
     }
 
     setOwnerInformation(tags);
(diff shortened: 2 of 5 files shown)


davidscn:
Hm, macOS seems to be unhappy about my changes as the pipeline occasionally reports failure of the one of the PETSc mapping tests. However,

1. I don't know whether this is specific to the changes proposed here
2. I can't reproduce it on my machine
3. I can't figure out the failing test (which is more of a general problem in our setup)

So my options to investigate are limited.

uekerman (approved):
Makes sense, thanks for the fix. The feature was simply not tested well enough.

PR #1849, opened 2023-10-23
Speed-up two-level initialization owner decision for shared vertices

davidscn:
## Main changes of this PR
More of a performance bug. Speeds-up the computation of the owner rank for two-level initialization for shared vertices between ranks.

The `48x48` PUM mapping run on ca 8mio vertices took the two-level initialization more than 3 hours due to the implementation touched here. With the performance fixes here, we can run the same in $\mathcal{O}(10\text{ms})$. I don't want to investigate the exact speed-up for more cases, but the speed-up is significant.

The changes I made in #1830 are nice and run fast, but for larger cases, I was not able to bypass the memory bottlenecks, i.e., typically ran out of RAM on the primary rank.

Open points here:

- I used the nice set intersection function again. In principle, I like to keep it in an anonymous namespace, but duplicating it seems like a bad idea. Any idea where to put it (it is now used in different packages. maybe a new utils file @fsimonis  ?)
- what to do with the commented code? It was already there before

I am a bit afraid of the test coverage in the whole two-level initialization, but I ran a PETSc case, which succeeded. I would assume that the ownership is for PETSc important and errors in the ownership would lead to errors in the PETSc mapping.

For now, I only ran the (2LI) cases on a moderate number of cores and I really hope that the somewhat mediocre looking computation of the bounding box intersection in the beginning of the two-level initialization doesn't pose an issue for larger core counts. 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/m2n/PointToPointCommunication.cpp
+++ b/src/m2n/PointToPointCommunication.cpp
@@ -22,6 +22,7 @@
 #include "precice/impl/Types.hpp"
 #include "profiling/Event.hpp"
 #include "utils/IntraComm.hpp"
+#include "utils/algorithm.hpp"
 #include "utils/assertion.hpp"
 
 using precice::profiling::Event;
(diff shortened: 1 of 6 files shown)


fsimonis:
> Any idea where to put it (it is now used in different packages. maybe a new utils file @fsimonis ?)

Given that it is pretty generic, you could add it to the collection in `src/utils/algorithm.hpp`.

> what to do with the commented code? It was already there before

My gut feeling says to remove it. What about yours @uekerman

uekerman (approved):
Really nice job with the speed-up. Large vertex counts on a moderate amount of cores is a scenario that was not tested much before. Amin typically looked at large core counts.

> I am a bit afraid of the test coverage in the whole two-level initialization

I would anyway only trust those cases to work that are also covered in the tests (as you already observed as well). So, don't be afraid :smile: :see_no_evil: 

> My gut feeling says to remove it. What about yours @uekerman

Yes, remove.

Don't forget to add a changelog.

davidscn:
I added a very basic integration test for 2LI and PUM which gave the same result. Not sure how representative it is for complex scenarios, but still better to have something in place.

Issue #1777, opened 2023-08-16
`precice-events` misclassifies performance across ranks


davidscn:
**Please describe the problem you are trying to solve.**
Using our `precice-events` tools, I get the following log

```bash
Selection contains the primary rank 0, the cheapest secondary rank 1, and the most expensive secondary rank 3.
                                                 eid |     R0:sum R0:count    R0:mean     R0:min     R0:max |     R1:sum R1:count    R1:mean     R1:min     R1:max |     R3:sum R3:count    R3:mean     R3:min     R3:max
                                             _GLOBAL | 21358583.0        1 21358583.0 21358583.0 21358583.0 | 21306222.0        1 21306222.0 21306222.0 21306222.0 | 21306194.0        1 21306194.0 21306194.0 21306194.0
```

However, it's not clear why rank 3 is here the most expensive rank (rank 0 has globally the largest time consumption). I'm not even sure whether this classification makes sense. Looking (as an example) for the mapping entry of the same event-log, I get

```bash
initialize/map.rbf.computeMapping.FromA-MeshToB-Mesh | 20920872.0        1 20920872.0 20920872.0 20920872.0 |     2304.0        1     2304.0     2304.0     2304.0 |     4004.0        1     4004.0     4004.0     4004.0
```

which is even more wrong, as still rank 3 (last column) is supposed to be the 'most expensive rank'. Main problem is that I use a script to get the max timing of the 'most expensive' rank, which leads to wrong results.

The question is, what to expect from the 'most expensive' rank: is it globally the rank taking most of the time or would this metric anyway make more sense on a per-event basis?

fsimonis:
When @uekerman and I discussed which secondary ranks to pick for the output of analyze, we converged to using the overall time spend in `advance`, so ranks that spend least and most time in preCICE.

Of course this choice is debatable, especially as this currently only addresses the preCICE perspective and ignores initialize.

Some potential questions:
* Should the total cost include the solver? (`advance` + `solver.advance`)
* Should the total cost include `initalization` time? (`advance` + `initialize` ( + `solver.advance` + `solver.initialize` ) )
* Should this be configurable? If so, what should the options be? `precice-events analyze --secondary-order=precice|solver|both FluidSolver`?

> However, it's not clear why rank 3 is here the most expensive rank (rank 0 has globally the largest time consumption).

Note that the tool shows:
1. the primary rank 0
2. the least expensive secondary rank !=0
3. the most expensive secondary rank !=0

[...]

PR #1855, opened 2023-10-26
Parametrize event in profiling script

davidscn:
## Main changes of this PR
 Adds the event name for the analyze command to give more freedom to the user for parallel runs.

Remotely related to https://github.com/precice/precice/issues/1777


fsimonis (approved):
Minimally invasive and useful

fsimonis:
@davidscn Could you add a note or example to the website?

Issue #1850, opened 2023-10-23
Conflict between `no-restart` mode and `always-build-jacobian` in acceleration options


Fujikawas:
**Describe your setup**
preCICE Version: [develop](https://github.com/precice/precice/tree/develop)

**Describe the problem**
According to error message in https://github.com/precice/precice/blob/7cc055289a56dd4f7057abb4baebe3c8e6d9b40d/src/acceleration/config/AccelerationConfiguration.cpp#L204C1-L207C1, the `always-build-jacobian` option could not be used when there exists `restart` method.
But when the `imvj-restart-mode` is set to `no-restart`, it still runs into this error, which I suppose should not happen.

**Step To Reproduce**
1. Run any case with acceleration option like following combination of `always-build-jacobian = "True"` and `(imvj-restart-mode) type="no-restart"` :
```xml
<acceleration:IQN-IMVJ  always-build-jacobian="True">
     ...
      <imvj-restart-mode
        truncation-threshold="0.0001"
        chunk-size="2"
        reused-time-windows-at-restart="2"
        type="no-restart" />
 </acceleration:IQN-IMVJ>
```

**Expected behaviour**
Set stricter condition for giving this error, so that the `no-restart` mode could be excluded.

PR #1856, opened 2023-10-26
Fix too strict error for IMVJ no-restart and always-build-jacobian

Fujikawas:
## Main changes of this PR
Change the condition leading to the error information, so that people could have `no-restart` for `imvj-restart-mode` and `always-build-jacobian="True"`  at the same time.

closes https://github.com/precice/precice/issues/1850

## Author's checklist

* [x] I added a changelog file with make changelog if there are user-observable changes since the last release.
* [x] ~~I added a test to cover the proposed changes in our test suite.~~ I tested with a little change to existing test suite, but didn't make it to a new test.

## Reviewers' checklist


src/acceleration/config/AccelerationConfiguration.cpp
```diff
@@ -201,13 +201,12 @@ void AccelerationConfiguration::xmlTagCallback(
...
-    if (_config.alwaysBuildJacobian)
-      PRECICE_ERROR("IMVJ cannot be in restart mode while parameter always-build-jacobian is set to true. "
-                    "Please remove 'always-build-jacobian' from the configuration file or do not run in restart mode.");
-
 #ifndef PRECICE_NO_MPI
     _config.imvjChunkSize = callingTag.getIntAttributeValue(ATTR_IMVJCHUNKSIZE);
     const auto &f         = callingTag.getStringAttributeValue(ATTR_TYPE);
+    if (f != VALUE_NO_RESTART && _config.alwaysBuildJacobian)
+      PRECICE_ERROR("IMVJ cannot be in restart mode while parameter always-build-jacobian is set to true. "
+                    "Please remove 'always-build-jacobian' from the configuration file or do not run in restart mode.");
```
fsimonis, line 209: Could you please rewrite this as a `PRECICE_CHECK(condition, message...)`?

uekerman (approved):

uekerman:
@Fujikawas Let's add a changelog entry as this is user-facing.

PR #1858, opened 2023-10-27
Improve compile times of Testing.hpp and fmt

fsimonis:
## Main changes of this PR

This PR attempts to reduce compile times of components that are used a lot in the code-base.
* It links fmtlib as a static library and links it against the preciceCore
* It adjusts parameter types in formatting tools to avoid conversions etc. This includes `Logger.hpp` and `assertions.hpp`.
* It reduces templates and inline functions in `Testing.hpp` by making them explicit and moving them to `Testing.cpp`. This eliminates a row of transitive headers.
* It removes unused features from Testing

This reduces the total compile time (excluding linkage) from 1696 to 1320 seconds.

Replacing iostream in assertions with `fmt::printf(stderr, ...)` leads to a higher compile time.

Going forward, Eigen seems to be the next major burden. It's expensive to include and used in every test. The most likely next step would be to build testprecice as a unity build, which requires disambiguation across cpp files.

Thankfully this is super easy to setup with CMake and was introduced in 3.16.

## Author's checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -212,7 +212,6 @@ precice_validate_libxml2()
 
 # libfmt
 add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/thirdparty/fmt)
-precice_validate_fmtlib()
 
 #
 # Find Configurable Dependencies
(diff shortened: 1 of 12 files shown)


fsimonis:
Bare builds are still about the same, but MPI and MPI+PETSc builds take around 4 minutes less to compile.

PR #1861, opened 2023-11-02
Improve test logging and CI output

davidscn:
## Main changes of this PR

Adds the test name to the context description and adds dedicated error reports into the CI pipeline.

Tracing down bugs/errors in our CI is somewhat intricate. The current `context` description simply reveals the mode (serial vs parallel) and the rank the test is being performed on. In addition, we get now the test name, which is helpful if preCICE crashed (not the test fails), becaue the test context will give the context description before failing (so at least the failing test is known). As of know, the trace-like test-logging was ambiguous: it might happen that another test was already reported as 'entered' although the actual fail was related to another test.

In addition, our BOOST_TEST_LOG_LEVEL=all is very clumsy to read. However, we can create a dedicated error logger for the test suits which will create a nice error report for each test suite in a HRF. I added a printer for the error logs, such that inspecting the logs should be much easier.

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -119,7 +119,7 @@ jobs:
           Ginkgo: ${{ contains(matrix.CONFIG, 'Ginkgo') }}
         run: |
           cmake --version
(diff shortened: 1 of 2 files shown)


fsimonis:
Specifying a single logger overwrites the default one. So, we should specify two: `HRF,error,error.log:HRF,all,stdout`.

Wouldn't it make sense to log only errors to stdout and everything to the file?
Then the `run tests`-step in shows only errors in-line, which should be vastly easier to read.

We could then upload the full logs as a build artefact.
Currently, if an integration test fails, one often needs to download the raw test logs and search them. A build artefact would make that step way easier.

fsimonis:
FYI, I extracted the TestContext addition to a separate PR and changed it to display the full name of the test unit.

fsimonis:
Closing this in favour of #2118 which isolates tests and allows to always output the full log.

PR #1863, opened 2023-11-02
Give warning info other than error by zero-update of `V` in QN-update

Fujikawas:
## Main changes of this PR
Change the `PRECICE_CHECK` to `PRECICE_WARN` when a zero-column is going to be added to `V` in QN-acceleration. The relevant columns would be added to `V` and `W` as usual and possibly filtered out later.
 
Closes https://github.com/precice/tutorials/issues/385, https://github.com/precice/precice/issues/1496

## Author's checklist

## Reviewers' checklist


src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -238,7 +238,7 @@ void BaseQNAcceleration::updateDifferenceMatrices(
         // insert column deltaR = _residuals - _oldResiduals at pos. 0 (front) into the
         // QR decomposition and update decomposition
 
-        //apply scaling here
+        // apply scaling here
```
BenjaminRodenberg, line 241:
```suggestion
        //apply scaling here
```
Would not add these kind of changes here. This just makes the diff unnecessary large. @fsimonis Can we add this to our linter to have one rule that is enforced consistently?

BenjaminRodenberg (changes requested):
Raising a warning makes sense for my use case. I think we should also provide a unit or integration test here to make sure the feature works and we do not run into the same problem at some point in the future again.

@uekerman in https://github.com/precice/tutorials/issues/385 you suggested something about "do not add the column". If I understand the current implementation right, now we "add the column, but (probably) filter it out". Is this good enough? For my use case it works, but I don't know how this will behave in a more general setting.

BenjaminRodenberg:
#1107 is already closed. I think this makes the description of this PR a bit confusing. What about https://github.com/precice/precice/issues/1496? @uekerman also pointed at this issue in https://github.com/precice/tutorials/issues/385

BenjaminRodenberg:
I also added a new milestone https://github.com/precice/precice/milestone/21, because I would really like to see this bugfix in a bugfix release. (still up to discussion whether we actually want to do such a release)

BenjaminRodenberg:
What's also related to this topic:

* https://github.com/precice/precice/pull/1510 is unmerged, but proposes a very similar fix.
* https://github.com/precice/precice/issues/1558 also belongs to this topic.

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -223,10 +223,10 @@ void BaseQNAcceleration::updateDifferenceMatrices(
...
       }
 
-      PRECICE_CHECK(not math::equals(residualMagnitude, 0.0),
-                    "Attempting to add a zero vector to the quasi-Newton V matrix. This means that the residuals "
-                    "in two consecutive iterations are identical. If a relative convergence limit was selected, "
-                    "consider increasing the convergence threshold.");
+      if (math::equals(residualMagnitude, 0.0)) {
+        PRECICE_WARN("Add a zero vector to the quasi-Newton V matrix. This means that the residuals "
+                     "in two consecutive iterations are identical.");
+      }
```
uekerman, line 229:
```suggestion
      if (math::equals(residualMagnitude, 0.0)) {
        PRECICE_WARN("Adding a vector with a two-norm of {} to the quasi-Newton V matrix, which will lead to "
                     "ill-conditioning. A filter might delete the column again. Still, this could mean that you are " 
                     "converging too tightly, that you reached steady-state, or that you are giving by mistake identical "
                     "data to preCICE in two consecutive iterations.", residualMagnitude);
      }
```

having the exact norm value could help debugging ... is it 1e-15 or really 0.

uekerman (approved):
Let's merge this.
But before we should double-check whether we really solve all mentioned issues.

> @uekerman in https://github.com/precice/tutorials/issues/385 you suggested something about "do not add the column". If I understand the current implementation right, now we "add the column, but (probably) filter it out". Is this good enough? For my use case it works, but I don't know how this will behave in a more general setting.

Good enough for me.

> I think we should also provide a unit or integration test here

We cannot yet test whether preCICE throws a warning or not. #280

MakisH:
In develop, the tutorial partitioned-heat-conduction with OpenFOAM-OpenFOAM fails with the usual error:

```
---[precice]  Revision info: v2.5.0-525-gf5124bb2
...
Time = 0.1

DICPCG:  Solving for T, Initial residual = 0.0371246, Final residual = 9.69608e-13, No Iterations 129
ExecutionTime = 0.44 s  ClockTime = 3 s

---[precice]  relative convergence measure: relative two-norm diff of data "Heat-Flux" = 5.99e-13, limit = 1.00e-05, normalization = 2.00e+01, conv = true
---[precice]  relative convergence measure: relative two-norm diff of data "Temperature" = inf, limit = 1.00e-05, normalization = 0.00e+00, conv = false
---[precice]  Mapping "Heat-Flux" for t=0 from "Dirichlet-Mesh" to "Neumann-Mesh"
---[precice]  Mapping "Heat-Flux" for t=0.1 from "Dirichlet-Mesh" to "Neumann-Mesh"
---[precice]  iteration: 10 of 100 (min 1), time-window: 1, time: 0 of 1, time-window-size: 0.1, max-time-step-size: 0.1, ongoing: yes, time-window-complete: no, read-iteration-checkpoint 
Time = 0.1

DICPCG:  Solving for T, Initial residual = 0.0371246, Final residual = 9.69608e-13, No Iterations 129
ExecutionTime = 0.46 s  ClockTime = 3 s

---[precice]  relative convergence measure: relative two-norm diff of data "Heat-Flux" = 7.06e-14, limit = 1.00e-05, normalization = 2.00e+01, conv = true
---[precice]  relative convergence measure: relative two-norm diff of data "Temperature" = inf, limit = 1.00e-05, normalization = 0.00e+00, conv = false
---[precice] ERROR:  Attempting to add a zero vector to the quasi-Newton V matrix. This means that the residuals in two consecutive iterations are identical. If a relative convergence limit was selected, consider increasing the convergence threshold.
```

With this PR, the case completes normally, but I don't get the warning:

```
---[precice]  Revision info: v2.5.0-518-g900b8c6c
...
---[precice]  relative convergence measure: relative two-norm diff of data "Heat-Flux" = 2.45e-04, limit = 1.00e-05, normalization = 2.00e+01, conv = false
---[precice]  relative convergence measure: relative two-norm diff of data "Temperature" = 5.94e-06, limit = 1.00e-05, normalization = 3.25e+01, conv = true
---[precice]  Mapping "Heat-Flux" for t=0 from "Dirichlet-Mesh" to "Neumann-Mesh"
---[precice]  Mapping "Heat-Flux" for t=0.1 from "Dirichlet-Mesh" to "Neumann-Mesh"
---[precice]  iteration: 5 of 100 (min 1), time-window: 1, time: 0 of 1, time-window-size: 0.1, max-time-step-size: 0.1, ongoing: yes, time-window-complete: no, read-iteration-checkpoint 
Time = 0.1

DICPCG:  Solving for T, Initial residual = 0.0371246, Final residual = 9.69608e-13, No Iterations 129
ExecutionTime = 0.32 s  ClockTime = 2 s

---[precice]  relative convergence measure: relative two-norm diff of data "Heat-Flux" = 1.48e-04, limit = 1.00e-05, normalization = 2.00e+01, conv = false
---[precice]  relative convergence measure: relative two-norm diff of data "Temperature" = 3.74e-10, limit = 1.00e-05, normalization = 3.25e+01, conv = true
---[precice]  Mapping "Heat-Flux" for t=0 from "Dirichlet-Mesh" to "Neumann-Mesh"
---[precice]  Mapping "Heat-Flux" for t=0.1 from "Dirichlet-Mesh" to "Neumann-Mesh"
---[precice]  iteration: 6 of 100 (min 1), time-window: 1, time: 0 of 1, time-window-size: 0.1, max-time-step-size: 0.1, ongoing: yes, time-window-complete: no, read-iteration-checkpoint 
Time = 0.1

DICPCG:  Solving for T, Initial residual = 0.0371246, Final residual = 9.69608e-13, No Iterations 129
ExecutionTime = 0.34 s  ClockTime = 2 s

---[precice]  relative convergence measure: relative two-norm diff of data "Heat-Flux" = 1.06e-08, limit = 1.00e-05, normalization = 2.00e+01, conv = true
---[precice]  relative convergence measure: relative two-norm diff of data "Temperature" = 2.05e-10, limit = 1.00e-05, normalization = 3.25e+01, conv = true
---[precice]  All converged
---[precice]  Time window completed
---[precice]  Mapping "Heat-Flux" for t=0.1 from "Dirichlet-Mesh" to "Neumann-Mesh"
---[precice]  Mapping "Heat-Flux" for t=0.2 from "Dirichlet-Mesh" to "Neumann-Mesh"
---[precice]  iteration: 1 of 100 (min 1), time-window: 2, time: 0.1 of 1, time-window-size: 0.1, max-time-step-size: 0.1, ongoing: yes, time-window-complete: yes, write-iteration-checkpoint 
---[preciceAdapter] The coupling timestep completed. Writing the updated results.
Time = 0.2
```

The first coupling time window actually only takes 6 iterations instead of 10: I am confused.

uekerman:
> The first coupling time window actually only takes 6 iterations instead of 10: I am confused.

Could you please attach the convergence files of both runs?

MakisH:
> > The first coupling time window actually only takes 6 iterations instead of 10: I am confused.
> 
> Could you please attach the convergence files of both runs?

Here are the logs for a good and a bad commit: [debug-iterations.tar.gz](https://github.com/precice/precice/files/13414528/debug-iterations.tar.gz)

I ran a git bisect, and the first bad commit seems to be 451d90022e60d9bd3a8f0f1147f6bbac645e6d1a

MakisH:
If I merge develop into this branch, I get 100 iterations:

[precice-Neumann-convergence.log](https://github.com/precice/precice/files/13415036/precice-Neumann-convergence.log)
[precice-Neumann-iterations.log](https://github.com/precice/precice/files/13415038/precice-Neumann-iterations.log)

fsimonis:
@MakisH Could you try the following fix https://github.com/precice/precice/pull/1885 ?

fsimonis:
We don't to include this in the hotfix as the fixed error handler was added post 2.5.

PR #1867, opened 2023-11-13
Add warning for very small time step sizes

### LINKED ISSUE #1866: Subcycling leads to time drift  leading to desync
**Describe your setup**

Operating system (e.g. Linux distribution and version): _Ubuntu 00.00_
preCICE Version: develop at 05541eb8c43e063df59d7181876fc6de773ae98d

**Describe the problem**

Using subcycling leads to a time drift which can result in a desync of participants and a crash once only one participant decides that `isCouplingOngoing() == false`.

Problem by @BenjaminRodenberg: 
```py
max_time   = 1.0
precice_dt = 0.2
solver_dt  = precice_dt / 640 # = 0.0003125

t = 0
NUMERICAL_ZERO_DIFFERENCE = 1e-14

while abs(t - max_time) > NUMERICAL_ZERO_DIFFERENCE:
    i = 0
    t_loc = 0
    while abs(t_loc - precice_dt) > 10e-14:
        i+=1
        t_loc += solver_dt
    t += t_loc
    print(f"Reached {t} after {i} time steps")
    if(t > max_time):
        break
```

We are using the computed time window to advance `timeWindowStart` in the `BaseCouplingScheme`. This should always be moved by `_timeWindowSize` if available.
This leads to a drift which builds up over time.

**Step To Reproduce**
1. max-time = 1.0, time-window-size=0.2, P1 does 1 timestep, P2 does 640 timesteps of size 0.2/640.
2. Run the simulation
3. P1 finalizes while P2 crashes as it is waiting for data, but the communication was closed.

**Expected behaviour**

No drift, no crash.

**Additional context**

This is related to but not caused by #1788, which advances the time window size correctly.
This used to simply hang with the participant wait in finalize, which is now disabled by default #1600.

There are 3 relevant cases at the end of a time window:
1. the next time step is too small, so we move to the real end #1788
2. the next time step size is numerically 0 (this doesn't move to the next time window correctly)
3. there is no time window size (nothing to correct)

**Possible solution**

Always "snap" the provided time of the last timestep to the end of the time window. This also requires correcting the time stamps of the data samples.

https://github.com/precice/precice/blob/05541eb8c43e063df59d7181876fc6de773ae98d/src/cplscheme/BaseCouplingScheme.cpp#L298-L325

```cpp
if (reachedEndOfTimeWindow()) {

    _timeWindows += 1; // increment window counter. If not converged, will be decremented again later.

    if (hasTimeWindowSize()) {

      //If preCICE has stopped before the end of the time window we have to duplicate the last available sample and put it at the end of the time window.
      // We have to exclude the case where coupling scheme does not have a time window size, since this will cause problem with the interpolation later on
      if (getNextTimeStepMaxSize() > math::NUMERICAL_ZERO_DIFFERENCE) {

        addTimeStepAtWindowEnd();

        // Update the _computedTimeWindowPart in order to keep the time within preCICE synchronised
        // Has to be done before the second exchange, since the serial coupling scheme moves to the new time window before updating _timeWindowStartTime
        _computedTimeWindowPart = _timeWindowSize;
      } else {
        // snap final data samples to the end
        snapDataToTimeWindowEnd();
        _computedTimeWindowPart = _timeWindowSize;
      }
    }

    exchangeFirstData();
  }
```

BenjaminRodenberg:
## Main changes of this PR

~~Closes #1866~~

* ~~Tracks drift correction in `BaseCouplingScheme`~~
* ~~Raises warning, if total drift correction exceeds numerical limit~~

This PR implements a warning, if preCICE returns a time step size that is considered small ($\epsilon$). Then a user can pick a better strategy for determining the time step size, if performing a small time step at the end of the window is unwanted.

See #1866 

The focus of this PR has shifted a bit after #1897 resolved the problem with the drifting. However, there is still the possibility that round-off errors can lead to strange effects with the time step size when using subcycling:

$n$ substeps of size $\delta t$ can lead to a time that slightly misses the end of the time window, i.e. $\Delta t - n \delta t = \epsilon$. If a user picks the wrong strategy for picking the time step size, this can lead to $n$ steps of size $\delta t$ and one very small time step of size $\epsilon$.

## Author's checklist

* [x] Merge #1882 

## Reviewers' checklist


BenjaminRodenberg:
Note: A good candidate for an integration test is the scenario mentioned in #1866.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -360,6 +362,10 @@ void BaseCouplingScheme::secondExchange()
         _isTimeWindowComplete = true;
         _timeWindowStartTime += _computedTimeWindowPart;
         _computedTimeWindowPart = 0.0; // reset window
+        _totalDriftCorrection += _driftCorrection;
+        if (abs(_totalDriftCorrection) > math::NUMERICAL_ZERO_DIFFERENCE) {
+          PRECICE_WARN("preCICE has corrected the time by a total of {}. This is often necessary, if you are using very many substeps per time window over multiple time windows.", _totalDriftCorrection);
+        }
```
fsimonis, line 368: Isn't this pretty much inevitable when dealing with many time steps?

BenjaminRodenberg: Sure. But raising a warning still makes sense, because we should tell our users that they are entering slightly unsafe ground. I think keeping track of the `_totalDriftCorrection` and telling the user if it becomes significant and also how large it is also makes sense.

fsimonis: Wouldn't this duplicate though in case of multiple coupling schemes.

Also, the correction happens in a single coupling scheme, so compositional schemes with different window sizes will drift independently.

What is the expected behaviour here?

BenjaminRodenberg: > Wouldn't this message duplicate though in case of multiple coupling schemes. We could instead always mention the drift in the coupling scheme info log.

I would not show the drift per default. Currently, it is an extreme exception that most of our users will not have to worry about. This might change in the future, but this is hard to estimate currently.

However, I think we should mention the drift prominently, if it occurs. This will also lead to users coming back to us and we will learn more about the specific situations where we have to be careful.

I guess we should do some testing with this warning and our tutorial cases. This warning should not show up for our tutorial cases and other cases without subcycling, but I did not test this so I cannot promise. For in-depth testing I currently do not have time, but it might also be sufficient to replace the warning with an error and look for failing unit and integration tests.

> More importantly, the correction happens in a single coupling scheme, so compositional schemes with different window sizes will drift independently.
> 
> What is the expected behaviour here?

Since every coupling scheme has its own way of tracking time I expect that different drifts might occur. Again, this is a very exceptional case. I think we cannot really factor this out as long as every coupling scheme allows its own time window size.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -360,6 +362,10 @@ void BaseCouplingScheme::secondExchange()
         _isTimeWindowComplete = true;
         _timeWindowStartTime += _computedTimeWindowPart;
         _computedTimeWindowPart = 0.0; // reset window
+        _totalDriftCorrection += _driftCorrection;
+        if (abs(_totalDriftCorrection) > math::NUMERICAL_ZERO_DIFFERENCE) {
+          PRECICE_WARN("preCICE has corrected the time by a total of {}. This is often necessary, if you are using very many substeps per time window over multiple time windows.", _totalDriftCorrection);
```
NiklasKotarsky, line 367: I do not know if it makes sense add: preCICE has corrected the time by a total of {}. This leads to an additional time integration error which might effect the accuracy of the solution. 

For me the connection between this error and moving the time step is far from trivial. This might also lead to difficult to debug errors when the time integration error is small.

NiklasKotarsky (approved):

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg, line 367:
```suggestion
          PRECICE_WARN("preCICE has corrected the time by a total of {}. This is often necessary, if you are using very many substeps per time window over multiple time windows. Please note that this can leads to an additional time integration error which might affect the accuracy of the solution.", _totalDriftCorrection);
```

@NiklasKotarsky you mean adding something along these lines?

BenjaminRodenberg: > For me the connection between this extra error and moving the time step is far from trivial. This might also lead to difficult to debug errors when the time integration error is small.

Absolutely. What we are doing here will probably have side effects, but our current approach is even more problematic (see #1866). I think we will learn more about the situation in the future and ideally this will lead to a fix that generally works without errors or the necessity to raise warnings.

The ideal solution from my perspective would be some additional synchronization between the preCICE time and the solver time. However, this requires considerable API changes.

NiklasKotarsky: Yes, exactly.

fsimonis: > This leads to an additional time integration error which might effect the accuracy of the solution.

A solution could be:

> In case you are computing the length of the final timestep of the window, try to directly use the return value of `getMaxTimeStepSize()` instead.

BenjaminRodenberg: > In case you are computing the length of the final timestep of the window, try to directly use the return value of `getMaxTimeStepSize()` instead.

That's a very good idea. I directly implemented it. We now don't have to implement any additional functionality with unclear outcome. Instead we raise an error, if something looks wrong and the user has to fix it. Fixing it is relatively easy and the user also knows what's going on. See https://github.com/BenjaminRodenberg/oscillator-example/commit/6741dcf5edf7c7c51532ab67fc94df47e720b925.

fsimonis: My intent was to append it to the warning, but if the total drift is a valid enough criterium to bail out, then even better.

I would avoid referring to a GitHub issue though. Let's use a full text instead. Given that it is an error now, message length doesn't matter :wink:

BenjaminRodenberg: Well... this is something we should actually explain in the documentation with a code example. Maybe a bit too much for an error message. But also touching the website... Not sure that I'm currently in the right mood to open another box :laughing:

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -341,6 +340,11 @@ void BaseCouplingScheme::secondExchange()
 
   if (reachedEndOfTimeWindow()) {
 
+    if (hasTimeWindowSize()) {
+      _totalTimeDrift += abs(_computedTimeWindowPart - _timeWindowSize);
+      PRECICE_CHECK(abs(_totalTimeDrift) < math::NUMERICAL_ZERO_DIFFERENCE, "preCICE has detected a difference between its internal time and the time of this participant. This can happen, if you are using very many substeps per time window over multiple time windows. Please refer to https://github.com/precice/precice/issues/1866 for strategies to avoid this problem.");
```
fsimonis, line 345: `_totalTimeDrift` is always positive.

```suggestion
      PRECICE_CHECK(_totalTimeDrift < math::NUMERICAL_ZERO_DIFFERENCE, "preCICE has detected a difference between its internal time and the time of this participant. This can happen, if you are using very many substeps per time window over multiple time windows. Please refer to https://github.com/precice/precice/issues/1866 for strategies to avoid this problem.");
```

fsimonis:
Regarding the title: this affects coupling time, not participant time. There can be multiple coupling schemes per participant

BenjaminRodenberg:
> Regarding the title: this affects coupling time, not participant time. There can be multiple coupling schemes per participant

The participant tells the coupling scheme about its internal time by providing `dt` to `advance(dt)`. So preCICE basically knows about the local time of the participant by adding up all the `dt` that `advance` has received. The coupling scheme keeps track of this in `_computedTimeWindowPart`.

Our terminology is extremly fuzzy here and the `min-timestep` feature makes this worse, because we basically overwrite the `_computedTimeWindowPart` (participant time) with the time window size (preCICE time?/coupling time?), if a certain criterion is met.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -309,17 +309,21 @@ void BaseCouplingScheme::firstExchange()
...
       addTimeStepAtWindowEnd();
 
       // Update the _computedTimeWindowPart in order to keep the time within preCICE synchronised
       // Has to be done before the second exchange, since the serial coupling scheme moves to the new time window before updating _timeWindowStartTime
       _computedTimeWindowPart = _timeWindowSize;
     }
 
+    if (hasTimeWindowSize()) {
+      // Store time between actually _computedTimeWindowPart and _timeWindowSize in buffer to keep track of total time drift in _totalTimeDrift.
+      _currentWindowTimeDrift = abs(_computedTimeWindowPart - _timeWindowSize);
```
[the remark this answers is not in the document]
BenjaminRodenberg: Not necessarily. We have the following condition:

```
{
  return math::equals(_timeWindowSize - _computedTimeWindowPart, 0.0, _minTimeStepSize) || not hasTimeWindowSize();
}
```

This means we consider a window to be complete, if `_computedTimeWindowPart` is close to `_timeWindowSize` by `_minTimeStepSize` in positive or negative direction.

In the end using `abs` leads to a worst-case estimate which makes sense in any case here.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -364,6 +368,8 @@ void BaseCouplingScheme::secondExchange()
           PRECICE_DEBUG("Setting require create checkpoint");
           requireAction(CouplingScheme::Action::WriteCheckpoint);
         }
+        _totalTimeDrift += _currentWindowTimeDrift;
+        PRECICE_CHECK(abs(_totalTimeDrift) < math::NUMERICAL_ZERO_DIFFERENCE, "preCICE has detected a difference between its internal time and the time of this participant. This can happen, if you are using very many substeps per time window over multiple time windows. Please refer to https://github.com/precice/precice/issues/1866 for strategies to avoid this problem.");
```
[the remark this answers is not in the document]
BenjaminRodenberg: Yes. We can remove the `abs` here and use the `unsigned double _totalTimeDrift`.

I would say we should error, because everything else will stay unnoticed and if the time drift becomes too large we should exit. Otherwise we might end up with situations like the one described in https://github.com/precice/precice/issues/1866.

On documentation: We definitely need to explain this better, but I first would like to conclude on this issue and on #1882 before investing time into documentation. There is also a potential refactoring for `_computedTimeWindowPart` on the way, because the whole situation here is very fuzzy as well.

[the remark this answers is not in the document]
BenjaminRodenberg: I added some corresponding documentation here: https://github.com/precice/precice.github.io/pull/257/files#diff-d807d6cfae4162be6a479b8c0280b99775fafb443f95d6fee3fb19619a6391dc

Currently, this is a draft, but we can first merge this PR here and then take care of the documentation later. I definitely have a more general review of all documentation regarding subcycling and time stepping in mind.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -380,6 +380,10 @@ void BaseCouplingScheme::secondExchange()
     // Update internal time tracking
     if (_isTimeWindowComplete) {
       _timeWindowStartTime += _timeWindowSize;
+      // Store time between actually _computedTimeWindowPart and _timeWindowSize in buffer to keep track of total time drift in _totalTimeDrift.
```
BenjaminRodenberg, line 383:
```suggestion
      // Compute difference between actually computed _computedTimeWindowPart and required _timeWindowSize to keep track of total time drift in _totalTimeDrift.
```

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -380,6 +380,10 @@ void BaseCouplingScheme::secondExchange()
     // Update internal time tracking
     if (_isTimeWindowComplete) {
       _timeWindowStartTime += _timeWindowSize;
+      // Compute difference between actually computed _computedTimeWindowPart and required _timeWindowSize to keep track of total time drift in _totalTimeDrift.
+      PRECICE_ASSERT(hasTimeWindowSize())
+      _totalTimeDrift += abs(_computedTimeWindowPart - _timeWindowSize);
+      PRECICE_CHECK(abs(_totalTimeDrift) < math::NUMERICAL_ZERO_DIFFERENCE, "preCICE has detected a difference between its internal time and the time of this participant. This can happen, if you are using very many substeps per time window over multiple time windows. Please refer to https://precice.org/couple-your-code-time-step-sizes.html for strategies to avoid this problem.");
```
uekerman, line 386:
```suggestion
      PRECICE_CHECK(abs(_totalTimeDrift) < math::NUMERICAL_ZERO_DIFFERENCE, "preCICE has detected a difference between its internal time and the time of this participant. This can happen, if you are using many substeps per time window over multiple time windows due to added-up differences of machine precision.");
```
The docs does not give information on this problem. But I think that is also not needed.

I think it would be clearer to use `math::equals` here.

tests/serial/time/explicit/parallel-coupling/ReadWriteScalarDataWithSubcyclingSmallSteps.cpp
```diff
@@ -0,0 +1,90 @@
...
+
+  while (precice.isCouplingOngoing()) {
+    double preciceDt = precice.getMaxTimeStepSize();
+
+    // // Wrong strategy to compute solver dt: Leads to PRECICE_ERROR after some time windows. As soon as https://github.com/precice/precice/issues/280 is merged, we should check for the error.
+    // double currentDt = solverDt > preciceDt ? preciceDt : solverDt;
+
+    // Correct strategy to compute solver dt that users should apply to avoid PRECICE_ERROR
+    double currentDt;
+    if (preciceDt - solverDt < 10e-14) {
```
uekerman, line 70:
```suggestion
    if (preciceDt - solverDt < 1e-14) {
```
?

uekerman (approved):
Minor comments

BenjaminRodenberg:
I think bringing this PR to a good ending is becoming tricky. See my latest comment in #1866. I managed to come up with some way to detect the time drift, but I think it's making things intransparent and complicated, while it is a bit unclear that we can really give a reasonable error.

BenjaminRodenberg:
Closing this PR. See https://github.com/precice/precice/issues/1866 for details.

uekerman:
We should still merge here, no?

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -550,7 +550,13 @@ double ParticipantImpl::getMaxTimeStepSize() const
 {
   PRECICE_CHECK(_state != State::Finalized, "getMaxTimeStepSize() cannot be called after finalize().");
   PRECICE_CHECK(_state == State::Initialized, "initialize() has to be called before getMaxTimeStepSize() can be evaluated.");
-  return _couplingScheme->getNextTimeStepMaxSize();
+  const double nextTimeStepSize = _couplingScheme->getNextTimeStepMaxSize();
+  PRECICE_ASSERT(!math::equals(nextTimeStepSize, 0.0));
+  PRECICE_ASSERT(math::greater(nextTimeStepSize, 0.0));
+  if (not math::greater(nextTimeStepSize, 0.0, 100 * math::NUMERICAL_ZERO_DIFFERENCE)) {
+    PRECICE_WARN("preCICE just returned a maximum time step size of {}. This can happen, if you are using many substeps per time window over multiple time windows due to added-up differences of machine precision. Please check your criterion for setting the time step size in your adapter.", nextTimeStepSize);
```
BenjaminRodenberg, line 557: I removed the `bug` tag, because this issue is a usability problem only. With the current implementation the user is responsible for picking the time step size with the correct strategy. If we want, we can still raise the warning above. I think such a small time step size is still worth a warning.

tests/serial/time/explicit/parallel-coupling/ReadWriteScalarDataWithSubcycling6400Steps.cpp
```diff
@@ -25,6 +25,7 @@ BOOST_AUTO_TEST_CASE(ReadWriteScalarDataWithSubcycling6400Steps)
   PRECICE_TEST("SolverOne"_on(1_rank), "SolverTwo"_on(1_rank));
 
   bool useAdvancedDtStrategy = true;
+  // bool useAdvancedDtStrategy = false;  // Triggers warning, because preciceDt becomes very small. Test also fails for second window with 6401 substeps.
```
BenjaminRodenberg, line 28: If we set `false` here, this simulates a user applying the wrong strategy for this problem.

uekerman (approved):

PR #1871, opened 2023-11-15
Add Wendland C8 RBF

davidscn:
## Main changes of this PR

..which would make our selection of Wendland RBFs complete. I'm not sure whether we really want to add it, but I will open it as a placeholder.

Adding it rather late here as finding a definition of it is not trivial.

## Author's checklist

### DIFF
--- a/docs/changelog/1871.md
+++ b/docs/changelog/1871.md
@@ -1,0 +1,1 @@
+- Added compactly supported Wendland C8 basis function.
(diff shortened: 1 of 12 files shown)


uekerman:
> I'm not sure whether we really want to add it

Why not, is a valid option and doesn't hurt.

davidscn:
Adding this to v3 then as this might come in handy for future experiments.

davidscn:
Some (a bit out of context) reference results  to show that the overall behavior of the newly added C8 RBF is consistent with the other Wendland functions

[result-error.pdf](https://github.com/precice/precice/files/13442372/result-error.pdf)

PR #1873, opened 2023-11-15
Rewrite error message in RBF solver

davidscn:
## Main changes of this PR

Rewrite error message in RBF solver.

The current error message is misleading. dead-axis can only be a reason for divergence if somebody configures `polynomial=on` in the rbf mapping by hand, which nobody does. In most of the cases, there are either duplicated nodes in the mesh (I would say that's most likely the case) or people configured the RBF function too flat such that the meshes are fine but the resulting matrix is not. Not sure how understandable the latter point is, but this would be the other error message user might face in the context of PUM (or RBFs more in general), so loosely related to #1869 .

### DIFF
--- a/src/mapping/RadialBasisFctSolver.hpp
+++ b/src/mapping/RadialBasisFctSolver.hpp
@@ -246,8 +246,8 @@ RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::RadialBasisFctSolver(RADIAL_BASIS
   PRECICE_CHECK(decompositionSuccessful,
                 "The interpolation matrix of the RBF mapping from mesh \"{}\" to mesh \"{}\" is not invertable. "
                 "This means that the mapping problem is not well-posed. "
-                "Please check if your coupling meshes are correct. Maybe you need to fix axis-aligned mapping setups "
-                "by marking perpendicular axes as dead?",
+                "Please check if your coupling meshes are correct (e.g. no vertices are duplicated) or reconfigure "
+                "your basis-function (e.g. reduce the support-radius).",
                 inputMesh.getName(), outputMesh.getName());
 
   // Second, assemble evaluation matrix


MakisH:
Have you checked for 2D cases? I think this used to be the main trigger for this (or similar) errors.

MakisH:
So, if I remember correctly, back in the day, we had cases that were coupling a pseudo-2D OpenFOAM case with a proper 2D solver. Both front and back planes of the OpenFOAM mesh were used to construct the RBF, which was flat in one axis and was then triggering this error. This is where, I think, the dead-axis suggestion originates from.

While we now have proper 2D modes in most adapters, and we don't need this setting in the tutorials, this error can still be triggered due to a not configured dead-axis, right? If this is the case, then maybe the error message should still point to that possibility as well.

I have probably missed some updates in the meantime, so please correct me.

davidscn:
> were coupling a pseudo-2D OpenFOAM case with a proper 2D solver. Both front and back planes of the OpenFOAM mesh were used to construct the RBF, which was flat in one axis and was then triggering this error.

I'm not sure I get the setup here correct: 

1. You are using preCICE in 2D, a 2D solver and a 3D case in OpenFOAM? In this case, dead axis won't make a difference anyhow as preCICE will only see two dimensions of the OpenFOAM mesh, i.e., the OpenFOAM mesh seems like it contains duplicated vertices

2. You are using preCICE in 3D, a 2D solver and a 3D case in OpenFOAM and set one dimension dead? For the error message here, setting the axis to dead doesn't make any difference and both situations would be completely fine. One might face this error https://github.com/precice/precice/blob/f5124bb23624aa1b6f1d37f7229f1d5fdbd642e0/src/mapping/PetRadialBasisFctMapping.hpp#L622 when using PETSc, which is then different again (we could actually consider to not throw an error here, I would assume that the solution at 'max iterations' is still sufficient). However, in preCICE v3 changing from PETSc to any other solver is a different tag (`rbf-global-iterative` vs `rbf-global-direct`) such that a different behavior between these two could is more obvious.

MakisH:
I think the setup was, e.g., perpendicular flap with OpenFOAM in 3D and CalculiX in 3D: https://github.com/precice/tutorials/blob/precice-v1.6/FSI/flap_perp/OpenFOAM-CalculiX/precice-config_serial.xml

MakisH (approved):
The "no vertices are duplicated" makes a lot of sense. :+1:

PR #1875, opened 2023-11-17
Cleanup Parallel class

### LINKED ISSUE #828: MPI initialization does not work for multiple SolverInterfaces
When creating two SolverInterfaces from one Python script, one runs into this problem: 

```c++
ASSERTION FAILED
  Location:          static void precice::utils::Parallel::initializeManagedMPI(int*, char***)
  File:              /home/jaustar/software/compilescripts/precice/precice-82cf90a/src/utils/Parallel.cpp:196
  Rank:              0
  Failed expression: !_isInitialized
  Argument 0: A managed MPI session already exists.
```

Could also be a general problem, not sure if this is related to the Python bindings. 
Reported by @ajaust, I have not yet tried to reproduce it.

fsimonis:
## Main changes of this PR

This PR cleans up the Parallel utility-class.

It is now more strict when detection reconstruction of the Participant which would require reinitalization of MPI, which is forbidden by the standard.

It also allows to reconstruct a Participant sessions, where MPI was initialized outside of preCICE, but no communicator was passed to the Participant constructor.
This direclty resolves https://github.com/precice/precice/issues/828 , assuming that mpi4py initializes MPI before the Participant is initialized.

Closes #828

Removes code kept for backwards compatibility with the "old" test system.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/Tooling.cpp
+++ b/src/precice/Tooling.cpp
@@ -51,7 +51,7 @@ void checkConfiguration(const std::string &filename, const std::string &particip
   logging::setMPIRank(0);
   const auto wasInitialized = utils::Parallel::isMPIInitialized();
   if (!wasInitialized) {
-    utils::Parallel::initializeMPI(nullptr, nullptr);
+    utils::Parallel::initializeTestingMPI(nullptr, nullptr);
   }
   xml::ConfigurationContext context{
       participant,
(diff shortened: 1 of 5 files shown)


uekerman (approved):
I did not check all details, but seems much cleaner.

PR #1878, opened 2023-11-17
Remove time from performAction

### LINKED ISSUE #1864: Remove time from Actions::performAction

Actions are currently performed at some point in time using `Action::performAction(double time)`.

With the integration of time interpolation, we now deal with samples at given times, or stamples for short.

For post write data actions, this wasn't a problem, as there is one sample generated at the beginning of advance, which is when that action would also be called. This however changed with #1788 as we can now conceptually have 2 new samples per call to advance.

For post read data actions, the main issue is that we handle all received stamples. This should use the timestamps of the samples instead of the time given by ParticipantImpl.

Remove the time parameter from `performAction` and use the timestamps of stamples instead.
```cpp
Action::performAction();
```

None. This is a refactoring that still needs to be done after introducing stamples.

fsimonis:
## Main changes of this PR

This PR removes time from `Action::performAction()` as the time is now given by the processed Stample.

Closes #1864

## Author's checklist

### DIFF
--- a/src/action/Action.hpp
+++ b/src/action/Action.hpp
@@ -3,8 +3,7 @@
 #include "mapping/Mapping.hpp"
 #include "mesh/SharedPointer.hpp"
 
-namespace precice {
-namespace action {
+namespace precice::action {
 
 /**
  * @brief Abstract base class for configurable actions on data and/or meshes.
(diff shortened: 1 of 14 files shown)


BenjaminRodenberg (approved):
I'm not so sure what will happen to the python actions, but rest looks like a reasonable simplification.

uekerman:
Checking whether https://precice.org/configuration-action.html#python-callback-interface still works good be a good test.

fsimonis:
> Checking whether the example on https://precice.org/configuration-action.html#python-callback-interface still works good be a good test.

This example is strange as source and target are the same. So the action receives 2 numpy arrays that point to the same data and both are writable.

PR #1882, opened 2023-11-20
Remove `min-time-step-size`

BenjaminRodenberg:
## Main changes of this PR

Removes the feature that allows setting the minimum time step size supported by the coupling scheme via `min-time-step-size`.

Offering this feature makes the coupling scheme and its time update complicated. Currently, it is unclear what's the right approach to deal with situations where a solver only supports time steps up to a certain minimal size. Additionally, this feature conflicts to some degree with #1867.

## Author's checklist

## Reviewers' checklist


tests/serial/time/implicit/serial-coupling/ReadWriteScalarDataFirstParticipantHighTolerance.cpp
```diff
@@ -88,30 +89,36 @@ BOOST_AUTO_TEST_CASE(ReadWriteScalarDataFirstParticipantHighTolerance)
...
     }
 
-    dt = solverDt > preciceDt ? preciceDt : solverDt;
+    preciceDt = precice.getMaxTimeStepSize();
+
+    if (abs(preciceDt - solverDt) < minTimeStepSize) {
+      dt = preciceDt;
+    } else {
+      dt = solverDt > preciceDt ? preciceDt : solverDt;
+    }
```
BenjaminRodenberg, line 98: This is the main thing our users would have to add in their adapter codes, if they have a time step restriction. Main argument here why removing the feature makes sense: preCICE cannot really know about a suitable strategy, a user would apply, if `preciceDt` of the next time step becomes smaller than the minimum allowed `solverDt`. Therefore, let the user detect the problematic situation and deal with the problem.

BenjaminRodenberg:
Decided together with @fsimonis and @uekerman: We will remove this feature, because the maintainment effort outweights the value of the feature and there is also one crucial point where the design of the feature conflicts with our implementation of waveform iteration.

@NiklasKotarsky: Sorry for going back and forth so often with this feature. Currently, we assume that generally you can avoid having to use time steps that are too small by adding some additional logic in the adapter (documentation will follow soon). If you observe problems on your side, please contact us and we will find a solution. 

The main purpose of this feature is that if a participant only allows for time steps $\delta t$ that are greater than `min-time-step-size`, preCICE will automatically move to the next window or iteration, if the remaining time inside the current time window is smaller than `min-time-step-size`. 

### Example: 

We use a time window size of $\Delta t = 1$. Participant A sets `min-time-step-size` to 0.2 and uses $\delta t = 0.3$, this would result in participant A performing 3 time steps of size 0.3. At $T = 3 \delta t = 0.9$ preCICE would move to the next window (because $\Delta t - 0.9 < 0.2$). The next time window now has a size of $\Delta t + 0.1$, because $0.1$ are still missing from the previous window (Important: that actual implementation did not follow this design, but this is considered a bug).

This leads to a conflict with the current implementation of the Waveform storage: Inside of the second window we need to sample the waveform $w(t)$ at $t \in [0.9, 2]$, but the storage currently only covers the active window time $t\in[1,2]$.

tests/serial/time/implicit/serial-coupling/ReadWriteScalarDataFirstParticipantHighTolerance.cpp
NiklasKotarsky: I think we could technically delete this test as well, since it does not test a specific preCICE feature anymore and the features it tests are tested better elsewhere. It might however make sense to add this check to the other test cases as whit minTimeStepSize=1e-14 to safeguard against time drift and tripping the error in  #1867.

docs/changelog/1882.md
```diff
@@ -0,0 +1 @@
+- Remove optional `min-time-step-size` option in `<coupling-scheme:...>`.
```
uekerman, line 1:
```suggestion
- Removed `min-time-step-size` option in `<precice-configuration:...>`.
```

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -206,12 +206,11 @@ void CouplingSchemeConfiguration::xmlTagCallback(
     _config.dtMethod = getTimesteppingMethod(tag.getStringAttributeValue(ATTR_METHOD, VALUE_FIXED));
     if (_config.dtMethod == constants::TimesteppingMethod::FIXED_TIME_WINDOW_SIZE) {
 
-      PRECICE_ASSERT(_minTimeStepSize > 0);
-      PRECICE_CHECK(_config.timeWindowSize >= _minTimeStepSize,
-                    "Time window size has to be larger or equal to the minimal time step used by preCICE. "
+      PRECICE_CHECK(_config.timeWindowSize >= math::NUMERICAL_ZERO_DIFFERENCE,
+                    "Time window size has to be larger or equal to the minimal time step supported by preCICE. "
```
uekerman, line 210: We should talk about the numerical zero difference here and we could also give the value.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -440,7 +425,7 @@ bool BaseCouplingScheme::addComputedTime(
   _computedTimeWindowPart += timeToAdd;
 
   // Check validness
-  bool valid = math::greaterEquals(getNextTimeStepMaxSize(), 0.0, _minTimeStepSize);
+  bool valid = math::greaterEquals(getNextTimeStepMaxSize(), 0.0, math::NUMERICAL_ZERO_DIFFERENCE);
```
uekerman, line 428:
```suggestion
  bool valid = math::greaterEquals(getNextTimeStepMaxSize(), 0.0);
```
should be the default, right?

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -456,7 +441,7 @@ bool BaseCouplingScheme::willDataBeExchanged(
 {
   PRECICE_TRACE(lastSolverTimeStepSize);
   double remainder = getNextTimeStepMaxSize() - lastSolverTimeStepSize;
-  return not math::greater(remainder, 0.0, _minTimeStepSize);
+  return not math::greater(remainder, 0.0, math::NUMERICAL_ZERO_DIFFERENCE);
```
uekerman, line 444:
```suggestion
  return not math::greater(remainder, 0.0);
```

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -520,7 +505,7 @@ double BaseCouplingScheme::getNextTimeStepMaxSize() const
 
 bool BaseCouplingScheme::isCouplingOngoing() const
 {
-  bool timeLeft      = math::greater(_maxTime, getTime(), _minTimeStepSize) || math::equals(_maxTime, UNDEFINED_MAX_TIME);
+  bool timeLeft      = math::greater(_maxTime, getTime(), math::NUMERICAL_ZERO_DIFFERENCE) || math::equals(_maxTime, UNDEFINED_MAX_TIME);
```
uekerman, line 508:
```suggestion
  bool timeLeft      = math::greater(_maxTime, getTime()) || math::equals(_maxTime, UNDEFINED_MAX_TIME);
```

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -781,7 +766,7 @@ void BaseCouplingScheme::advanceTXTWriters()
 
 bool BaseCouplingScheme::reachedEndOfTimeWindow() const
 {
-  return math::equals(_timeWindowSize - _computedTimeWindowPart, 0.0, _minTimeStepSize) || not hasTimeWindowSize();
+  return math::equals(_timeWindowSize - _computedTimeWindowPart, 0.0, math::NUMERICAL_ZERO_DIFFERENCE) || not hasTimeWindowSize();
```
uekerman, line 769:
```suggestion
  return math::equals(_timeWindowSize - _computedTimeWindowPart, 0.0) || not hasTimeWindowSize();
```

uekerman (approved):
A few more minor tweaks

src/cplscheme/config/CouplingSchemeConfiguration.cpp
BenjaminRodenberg: Done: [d631956](https://github.com/precice/precice/pull/1882/commits/d6319565367c0f566901f7212217ddc659b99389). This check was broken anyway :see_no_evil:

PR #1885, opened 2023-11-20
Fix zero write data bug

fsimonis:
## Main changes of this PR

Fixed a regression introduced by #1705.

After converting the written data to a sample, I zeroed the write buffer and mistakenly zeroed the `values()`.

Shame

## Author's checklist

## Reviewers' checklist


MakisH (approved):
This indeed reduces the iterations, but I still have an issue with the OpenFOAM adapter, even when not subcycling: https://github.com/precice/tutorials/pull/406

I think this is independent of this PR, though.

PR #1892, opened 2023-11-22
Fix const relax acceleration

fsimonis:
## Main changes of this PR

This PR attempts to solve the problem discovered by https://github.com/precice/tutorials/pull/406

Problem is that we still need to update `Data::values()` alongside the time storage.

## Author's checklist

### DIFF
--- a/src/acceleration/test/AccelerationSerialTest.cpp
+++ b/src/acceleration/test/AccelerationSerialTest.cpp
@@ -515,6 +515,8 @@ BOOST_AUTO_TEST_CASE(testConstantUnderrelaxationWithoutSubsteps)
 
   acc.performAcceleration(data);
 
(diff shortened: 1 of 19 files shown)


src/acceleration/test/AccelerationSerialTest.cpp
```diff
@@ -524,6 +526,8 @@ BOOST_AUTO_TEST_CASE(testConstantUnderrelaxationWithoutSubsteps)
   BOOST_TEST(data.at(1)->values()(2) == 0.16);
   BOOST_TEST(data.at(1)->values()(3) == 0.16);
 
+#if 0
+
```
MakisH, line 530: Is this meant to remain here?

MakisH (approved):
This does solve the bug I encountered in https://github.com/precice/tutorials/pull/406

Issue #1896, opened 2023-11-22
Some errors trigger a std::bad_optional_access exception


fsimonis:
**Describe your setup**

Operating system (e.g. Linux distribution and version): Arch
preCICE Version: latest dev of v3

**Describe the problem**

Some errors trigger `std::bad_optional_access` exceptions, which shouldn't be the case.
They are also visually very anoying.
These seem to origin mainly from configuration.

**Expected behaviour**

Only an error, no exception

PR #1898, opened 2023-11-23
Prevent bad optional access in events

davidscn:
## Main changes of this PR

Closes https://github.com/precice/precice/issues/1896

Which is triggered during the configuration stage as the events are not initialized yet.


fsimonis:
Nicely spotted.
So the actual bug is that it is possible to stop the backend without having started it.
This should be checked in the destructor. I'll add some safeguards in another PR.

fsimonis:
Closed in favour of #1901

PR #1899, opened 2023-11-23
Map only if necessary

### LINKED ISSUE #1707: Avoid unnecessary mappings in `DataContext`
The [mapping is applied to the complete storage of `Stamples`](https://github.com/precice/precice/pull/1612/files#diff-8d2f34adf551a7a24672179250b2b510723c2cec784a5680ff00d0d7ce41e711R87-R102). This is especially important for follow-up PRs that improve subcycling (#1664, #1671).

The relevant code part is the following:

https://github.com/precice/precice/blob/394d6483effeee43c4d23cff709fafe7106ca566/src/precice/impl/DataContext.cpp#L101-L119

One mapping operation that we don't need to perform is the mapping of data at `WINDOW_START`, because this data does not change anymore. Currently, we perform a mapping for the data at `WINDOW_START` in every iteration even though this is not needed. There is a possibility for improving performance here, because we might be able to save half of the mapping that are performed (no subcycling, only data at `WINDOW_START` and `WINDOW_END`, only mapping at `WINDOW_END` is actually needed).

Note: during initialization we need to map also the data at `WINDOW_START`, which would require a special treatment.

fsimonis:
## Main changes of this PR

This PR changes mappings to only map stamples that are missing.

This comes with some significant implications:

* preCICE will **never** map a sample multiple times
* all mappings are now steered by clearing timestorages
* Actions will not work with mixed-time-window sizes as samples can persist multiple read/write mappings (unchecked)
* write-data samples will not be generated when subcycling is turned off for all exchanges. **enabling substeps on a single exchange enables samples from subcycling for the participant, which leads to the evaluation of all write mappings** 

Before we merge this PR, I would prefer to get this tested using a few tutorials under real conditions.

**Open questions:**

* Did I overlook a cleaning schenario?
* Do the exports still work?
* Does acceleration still work? 

Closes #1707

## Author's checklist

## Reviewers' checklist

* [ ] Can you imagine any situations in which this could backfire?

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.hpp
+++ b/src/cplscheme/BaseCouplingScheme.hpp
@@ -396,6 +396,12 @@ protected:
    */
   bool reachedEndOfTimeWindow() const;
 
+  /// @copydoc cplscheme::CouplingScheme::requiresSubsteps()
+  bool requiresSubsteps() const override final;
+
+  /// @copydoc cplscheme::CouplingScheme::implicitDataToReceive()
+  ImplicitData implicitDataToReceive() const override;
+
 private:
   /// Coupling mode used by coupling scheme.
   CouplingMode _couplingMode = Undefined;
--- a/src/cplscheme/CompositionalCouplingScheme.hpp
+++ b/src/cplscheme/CompositionalCouplingScheme.hpp
@@ -204,6 +204,12 @@ public:
   /// True if the implicit scheme has converged or no implicit scheme is defined
   bool hasConverged() const final;
 
+  /// @copydoc cplscheme::CouplingScheme::requiresSubsteps()
+  bool requiresSubsteps() const override final;
+
+  /// @copydoc cplscheme::CouplingScheme::implicitDataToReceive()
+  ImplicitData implicitDataToReceive() const override final;
+
 private:
   mutable logging::Logger _log{"cplscheme::CompositionalCouplingScheme"};
 
--- a/src/cplscheme/CouplingScheme.hpp
+++ b/src/cplscheme/CouplingScheme.hpp
@@ -5,6 +5,7 @@
 #include <vector>
 
 #include "com/SharedPointer.hpp"
+#include "cplscheme/ImplicitData.hpp"
 #include "precice/impl/Types.hpp"
 
 namespace precice {
@@ -226,6 +227,12 @@ public:
 
   /// Returns false if the scheme is implicit and hasn't converged
   virtual bool hasConverged() const = 0;
+
+  /// Returns true if any send data of the scheme requires substeps
+  virtual bool requiresSubsteps() const = 0;
+
+  /// Returns a vector of implicit data to receive in the next advance
+  virtual ImplicitData implicitDataToReceive() const = 0;
 };
 
 } // namespace cplscheme
--- a/src/precice/impl/DataContext.cpp
+++ b/src/precice/impl/DataContext.cpp
@@ -1,6 +1,8 @@
-#include "precice/impl/DataContext.hpp"
+#include <iterator>
 #include <memory>
 #include <utility>
+
+#include "precice/impl/DataContext.hpp"
 #include "utils/EigenHelperFunctions.hpp"
 
 namespace precice::impl {
@@ -85,20 +87,41 @@ bool DataContext::hasMapping() const
   return hasReadMapping() || hasWriteMapping();
 }
 
-void DataContext::mapData()
+int DataContext::mapData(std::optional<double> after)
 {
+  PRECICE_TRACE(getMeshName(), getDataName());
   PRECICE_ASSERT(hasMapping());
-  // Execute the mapping
+
+  int executedMappings{0};
+
+  // Execute the mappings
   for (auto &context : _mappingContexts) {
-    // Reset the toData before mapping any samples
-    context.clearToDataStorage(); // @todo needs optimization: We don't need to map the data at the beginning of the window, because it should be known from the last window where it was the data from the window end. Exception: Data initialization. Related to https://github.com/precice/precice/issues/1707
-    PRECICE_ASSERT(context.fromData->stamples().size() > 0);
+    PRECICE_ASSERT(!context.fromData->stamples().empty(),
+                   "There must be samples at this point!");
+
+    // linear lookup should be sufficient here
+    const auto timestampExists = [times = context.toData->timeStepsStorage().getTimes()](double lookup) -> bool {
+      return std::any_of(times.data(), std::next(times.data(), times.size()), [lookup](double time) {
+        return math::equals(time, lookup);
+      });
+    };
 
     auto &mapping = *context.mapping;
 
     const auto dataDims = context.fromData->getDimensions();
 
     for (const auto &stample : context.fromData->stamples()) {
+      // skip stamples before given time
+      if (after && math::smallerEquals(stample.timestamp, *after)) {
+        PRECICE_DEBUG("Skipping stample t={} (not after {})", stample.timestamp, *after);
+        continue;
+      }
+      // skip existing stamples
+      if (timestampExists(stample.timestamp)) {
+        PRECICE_DEBUG("Skipping stample t={} (exists)", stample.timestamp);
+        continue;
+      }
+
       PRECICE_INFO("Mapping \"{}\" for t={} from \"{}\" to \"{}\"",
                    getDataName(), stample.timestamp, mapping.getInputMesh()->getName(), mapping.getOutputMesh()->getName());
       time::Sample outSample{
@@ -116,8 +139,10 @@ void DataContext::mapData()
 
       // Store data from mapping buffer in storage
       context.toData->setSampleAtTime(stample.timestamp, std::move(outSample));
+      ++executedMappings;
     }
   }
+  return executedMappings;
 }
 
 bool DataContext::hasReadMapping() const
--- a/src/precice/impl/ParticipantImpl.cpp
+++ b/src/precice/impl/ParticipantImpl.cpp
@@ -376,6 +376,7 @@ void ParticipantImpl::advance(
   // Update the coupling scheme time state. Necessary to get correct remainder.
   const bool   isAtWindowEnd = _couplingScheme->addComputedTime(computedTimeStepSize);
   const double timeSteppedTo = _couplingScheme->getTime();
+  const auto   dataToReceive = _couplingScheme->implicitDataToReceive();
 
   handleDataBeforeAdvance(isAtWindowEnd, timeSteppedTo);
 
@@ -385,9 +386,12 @@ void ParticipantImpl::advance(
   const double timeAfterAdvance   = _couplingScheme->getTime();
   const bool   timeWindowComplete = _couplingScheme->isTimeWindowComplete();
 
-  handleDataAfterAdvance(isAtWindowEnd, timeWindowComplete, timeSteppedTo, timeAfterAdvance);
+  handleDataAfterAdvance(isAtWindowEnd, timeWindowComplete, timeSteppedTo, timeAfterAdvance, dataToReceive);
 
   PRECICE_INFO(_couplingScheme->printCouplingState());
+
+  PRECICE_DEBUG("Mapped {} samples in write mappings and {} samples in read mappings",
+                _executedWriteMappings, _executedReadMappings);
 
   _meshLock.lockAll();
 
@@ -398,37 +402,48 @@ void ParticipantImpl::advance(
 
 void ParticipantImpl::handleDataBeforeAdvance(bool reachedTimeWindowEnd, double timeSteppedTo)
 {
-  samplizeWriteData(timeSteppedTo);
+  if (reachedTimeWindowEnd || _couplingScheme->requiresSubsteps()) {
+    samplizeWriteData(timeSteppedTo);
+  }
+  resetWrittenData();
+
+  // Reset mapping counters here to cover subcycling
+  _executedReadMappings  = 0;
+  _executedWriteMappings = 0;
 
   if (reachedTimeWindowEnd) {
-    mapWrittenData();
+    mapWrittenData(_couplingScheme->getTimeWindowStart());
     performDataActions({action::Action::WRITE_MAPPING_POST});
   }
 }
 
-void ParticipantImpl::handleDataAfterAdvance(bool reachedTimeWindowEnd, bool isTimeWindowComplete, double timeSteppedTo, double timeAfterAdvance)
+void ParticipantImpl::handleDataAfterAdvance(bool reachedTimeWindowEnd, bool isTimeWindowComplete, double timeSteppedTo, double timeAfterAdvance, const cplscheme::ImplicitData &receivedData)
 {
   if (!reachedTimeWindowEnd) {
     // We are subcycling
     return;
   }
 
-  if (reachedTimeWindowEnd) {
-    mapReadData();
-    // FIXME: the actions timing for read mappings doesn't make sense after
-    performDataActions({action::Action::READ_MAPPING_POST});
-  }
-
-  handleExports();
-
   if (isTimeWindowComplete) {
     // Move to next time window
     PRECICE_ASSERT(math::greaterEquals(timeAfterAdvance, timeSteppedTo), "We must have stayed or moved forwards in time (min-time-step-size).");
 
     // As we move forward, there may now be old samples lying around
-    // We know that timeAfterAdvance is the start time of the time window
-    trimOldDataBefore(timeAfterAdvance);
+    trimOldDataBefore(_couplingScheme->getTimeWindowStart());
+  } else {
+    // We are iterating
+    PRECICE_ASSERT(math::greater(timeSteppedTo, timeAfterAdvance), "We must have moved back in time!");
 
+    trimSendDataAfter(timeAfterAdvance);
+  }
+
+  if (reachedTimeWindowEnd) {
+    trimReadMappedData(timeAfterAdvance, isTimeWindowComplete, receivedData);
+    mapReadData();
+    performDataActions({action::Action::READ_MAPPING_POST});
+  }
+
+  if (isTimeWindowComplete) {
     // Reset initial guesses for iterative mappings
     for (auto &context : _accessor->readDataContexts()) {
       context.resetInitialGuesses();
@@ -436,13 +451,9 @@ void ParticipantImpl::handleDataAfterAdvance(bool reachedTimeWindowEnd, bool isT
     for (auto &context : _accessor->writeDataContexts()) {
       context.resetInitialGuesses();
     }
-    return;
   }
 
-  // We are iterating
-  PRECICE_ASSERT(math::greater(timeSteppedTo, timeAfterAdvance), "We must have moved back in time!");
-
-  trimSendDataAfter(timeAfterAdvance);
+  handleExports();
 }
 
 void ParticipantImpl::samplizeWriteData(double time)
@@ -451,7 +462,6 @@ void ParticipantImpl::samplizeWriteData(double time)
   for (auto &context : _accessor->writeDataContexts()) {
     context.storeBufferedData(time);
   }
-  resetWrittenData();
 }
 
 void ParticipantImpl::trimOldDataBefore(double time)
@@ -1385,14 +1395,31 @@ void ParticipantImpl::computeMappings(std::vector<MappingContext> &contexts, con
   }
 }
 
-void ParticipantImpl::mapWrittenData()
+void ParticipantImpl::mapWrittenData(std::optional<double> after)
 {
   PRECICE_TRACE();
   computeMappings(_accessor->writeMappingContexts(), "write");
   for (auto &context : _accessor->writeDataContexts()) {
     if (context.hasMapping()) {
       PRECICE_DEBUG("Map write data \"{}\" from mesh \"{}\"", context.getDataName(), context.getMeshName());
-      context.mapData();
+      _executedWriteMappings += context.mapData(after);
+    }
+  }
+}
+
+void ParticipantImpl::trimReadMappedData(double startOfTimeWindow, bool isTimeWindowComplete, const cplscheme::ImplicitData &fromData)
+{
+  PRECICE_TRACE();
+  for (auto &context : _accessor->readDataContexts()) {
+    if (context.hasMapping()) {
+      if (isTimeWindowComplete) {
+        // For serial implicit second, we need to discard everything before startOfTimeWindow to preserve the time window start
+        // For serial implicit first, we need to discard everything as everything is new
+        // For parallel implicit, we need to discard everything as everything is new
+        context.clearToDataFor(fromData);
+      } else {
+        context.trimToDataAfterFor(fromData, startOfTimeWindow);
+      }
     }
   }
 }
@@ -1404,7 +1431,8 @@ void ParticipantImpl::mapReadData()
   for (auto &context : _accessor->readDataContexts()) {
     if (context.hasMapping()) {
       PRECICE_DEBUG("Map read data \"{}\" to mesh \"{}\"", context.getDataName(), context.getMeshName());
-      context.mapData();
+      // We always ensure that all read data was mapped
+      _executedReadMappings += context.mapData();
     }
   }
 }
@@ -1534,6 +1562,14 @@ const mesh::Mesh &ParticipantImpl::mesh(const std::string &meshName) const
 {
   PRECICE_TRACE(meshName);
   return *_accessor->usedMeshContext(meshName).mesh;
+}
+
+ParticipantImpl::MappedSamples ParticipantImpl::mappedSamples() const
+{
+  MappedSamples res;
+  res.read  = _executedReadMappings;
+  res.write = _executedWriteMappings;
+  return res;
 }
 
 } // namespace precice::impl
--- a/src/precice/impl/ParticipantImpl.hpp
+++ b/src/precice/impl/ParticipantImpl.hpp
@@ -267,6 +267,13 @@ public:
   /// @todo try to remove or make private. See https://github.com/precice/precice/issues/1269
   const mesh::Mesh &mesh(const std::string &meshName) const;
 
+  struct MappedSamples {
+    int write, read;
+  };
+
+  /// Returns the amount of mapped read and write samples in the last call to advance.
+  MappedSamples mappedSamples() const;
+
   /// Disable copy construction
   ParticipantImpl(ParticipantImpl const &) = delete;
 
@@ -327,6 +334,12 @@ private:
   /// Counts calls to advance for plotting.
   long int _numberAdvanceCalls = 0;
 
+  /// Counts the amount of samples mapped in write mappings executed in the latest advance
+  int _executedWriteMappings = 0;
+
+  /// Counts the amount of samples mapped in read mappings executed in the latest advance
+  int _executedReadMappings = 0;
+
   /**
    * @brief Configures the coupling interface from the given xml file.
    *
@@ -363,11 +376,18 @@ private:
   /// Helper for mapWrittenData and mapReadData
   void computeMappings(std::vector<MappingContext> &contexts, const std::string &mappingType);
 
-  /// Computes, performs, and resets all suitable write mappings.
-  void mapWrittenData();
+  /// Computes, and performs suitable write mappings either entirely or after given time
+  void mapWrittenData(std::optional<double> after = std::nullopt);
 
-  /// Computes, performs, and resets all suitable read mappings.
+  // Computes, and performs read mappings
   void mapReadData();
+
+  /**
+   * @brief Removes samples in mapped to data connected to received data via a mapping.
+   *
+   * This prevents old samples from blocking remappings.
+   */
+  void trimReadMappedData(double timeAfterAdvance, bool isTimeWindowComplete, const cplscheme::ImplicitData &fromData);
 
   /**
    * @brief Performs all data actions with given timing.
@@ -410,7 +430,7 @@ private:
   void handleDataBeforeAdvance(bool reachedTimeWindowEnd, double timeSteppedTo);
 
   /// Completes everything data-related after advancing the coupling scheme
-  void handleDataAfterAdvance(bool reachedTimeWindowEnd, bool isTimeWindowComplete, double timeSteppedTo, double timeAfterAdvance);
+  void handleDataAfterAdvance(bool reachedTimeWindowEnd, bool isTimeWindowComplete, double timeSteppedTo, double timeAfterAdvance, const cplscheme::ImplicitData &receivedData);
 
   /// Creates a Stample at the given time for each write Data and zeros the buffers
   void samplizeWriteData(double time);
--- a/src/time/Storage.cpp
+++ b/src/time/Storage.cpp
@@ -116,6 +116,17 @@ void Storage::clear()
   _bspline.reset();
 }
 
+void Storage::clearExceptLast()
+{
+  if (_stampleStorage.empty()) {
+    return;
+  }
+  _stampleStorage.erase(_stampleStorage.begin(), --_stampleStorage.end());
+
+  // The spline has to be recomputed, since the underlying data has changed
+  _bspline.reset();
+}
+
 void Storage::trimBefore(double time)
 {
   auto beforeTime = [time](const auto &s) { return math::smaller(s.timestamp, time); };
--- a/tests/serial/map-if-necessary/three-solvers/helper.hpp
+++ b/tests/serial/map-if-necessary/three-solvers/helper.hpp
@@ -1,0 +1,78 @@
+#pragma once
+
+#include <vector>
+
+#include "precice/impl/ParticipantImpl.hpp"
+#include "testing/Testing.hpp"
+
+// Coupling is A - C - B, which can be MultiCoupling or Compositional. A-C may be explicitly coupled.
+// A and B always go first.
+// A never sends substeps, B always sends substeps except in Multi Coupling, which should be reflected in the read mappings of C.
+// The tests choose various scenarios for C to send and receive substeps. Some with subteps, some without and some with mixed usage.
+// Checkpoints are ignored for the sake of simplicity and time window sizes are considered to be equal
+
+inline void runMultipleSolversMappingCount(precice::testing::TestContext &context, std::vector<int> readMappings, std::vector<int> writeMappings)
+{
+  BOOST_REQUIRE(readMappings.size() == writeMappings.size());
+  BOOST_REQUIRE(readMappings.size() > 1);
+
+  precice::Participant participant(context.name, context.config(), context.rank, context.size);
+
+  std::vector<double>            coords{0, 0, 1, 1};
+  std::vector<precice::VertexID> vertexIDs(2);
+  std::string                    meshName = "Mesh" + context.name;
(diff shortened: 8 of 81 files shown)


src/precice/impl/DataContext.cpp
```diff
@@ -85,20 +87,40 @@ bool DataContext::hasMapping() const
   return hasReadMapping() || hasWriteMapping();
 }
 
-void DataContext::mapData()
+void DataContext::mapData(std::optional<double> after)
 {
+  PRECICE_TRACE(getMeshName(), getDataName(), after.value_or(-666));
   PRECICE_ASSERT(hasMapping());
-  // Execute the mapping
+  PRECICE_ASSERT(after.value_or(0) >= 0.0, "Invalid time");
```
BenjaminRodenberg, line 94: Does this assertion really help? We are not dealing with relative times anymore and we do not really know a lot about the valid times to expect (before it was $t\in[0,1]$). Wouldn't it make more sense to just use `unsigned double` here? (btw: I think there is also some potential for refactoring in `BaseCouplingScheme` etc. to use `unsigned double` for all the members tracking absolute time)

fsimonis: The trouble with `unsigned double` is that one needs to implement this in the communication backends too.

But you are right. The assertion is not really necessary.

src/precice/impl/DataContext.cpp
```diff
@@ -85,20 +87,40 @@ bool DataContext::hasMapping() const
...
+    if (after) {
+      context.toData->timeStepsStorage().trimAfter(*after);
+    } else {
+      context.clearToDataStorage();
+    }
+
+    const auto existing = [&] {
+      auto times = context.toData->timeStepsStorage().getTimes();
+      return std::set<double>{times.data(), times.data() + times.size()};
+    }();
```
uekerman, line 111: Could you add a comment what this does any why?
size (int) is added to data (double) ?

src/precice/impl/DataContext.cpp
```diff
@@ -85,20 +87,40 @@ bool DataContext::hasMapping() const
...
 
     auto &mapping = *context.mapping;
 
     const auto dataDims = context.fromData->getDimensions();
 
     for (const auto &stample : context.fromData->stamples()) {
+      // skip existing stamples
+      if (existing.count(stample.timestamp) > 0) {
```
uekerman, line 119: Does this use a `math::equals`?

uekerman (commented):
Looks like an elegant solution, but indeed needs intensive testing.
Please also try to find good integration tests and really test whether the expected amount of `map` calls is done.
I don't see a backfire situation yet.

src/precice/impl/DataContext.cpp
fsimonis: It computes the end pointer. I think I need to change this to address your bottom comment.

src/precice/impl/DataContext.cpp
fsimonis: No, but it should. I replaced it with a linear lookup and math::equals.

src/precice/impl/DataContext.cpp
fsimonis: Removed it

BenjaminRodenberg (commented):
Generally looks good and nice that there are no bigger changes necessary.

src/precice/impl/DataContext.cpp
```diff
@@ -85,20 +87,40 @@ bool DataContext::hasMapping() const
...
+                   "There must be samples at this point!");
+
     // Reset the toData before mapping any samples
-    context.clearToDataStorage(); // @todo needs optimization: We don't need to map the data at the beginning of the window, because it should be known from the last window where it was the data from the window end. Exception: Data initialization. Related to https://github.com/precice/precice/issues/1707
-    PRECICE_ASSERT(context.fromData->stamples().size() > 0);
+    if (after) {
+      context.toData->timeStepsStorage().trimAfter(*after);
+    } else {
+      context.clearToDataStorage();
+    }
```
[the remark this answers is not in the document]
fsimonis: I removed the clear and replaced it with an assertion.
The timestorage will always be empty.

src/precice/impl/DataContext.cpp
```diff
@@ -85,20 +87,40 @@ bool DataContext::hasMapping() const
...
 
     auto &mapping = *context.mapping;
 
     const auto dataDims = context.fromData->getDimensions();
 
     for (const auto &stample : context.fromData->stamples()) {
+      // skip existing stamples
+      if (existing.count(stample.timestamp) > 0) {
+        PRECICE_DEBUG("Skipping stample t={}", stample.timestamp);
+        continue;
```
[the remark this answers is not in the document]
fsimonis: Yes, indentation is evil.

I changed this to the following which is way easier to understand:
```cpp
      // skip existing stamples
      if (timestampExists(stample.timestamp)) {
        PRECICE_DEBUG("Skipping stample t={}", stample.timestamp);
        continue;
      }
```

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -401,7 +401,7 @@ void ParticipantImpl::handleDataBeforeAdvance(bool reachedTimeWindowEnd, double
   samplizeWriteData(timeSteppedTo);
 
   if (reachedTimeWindowEnd) {
-    mapWrittenData();
+    mapWrittenData(timeSteppedTo);
```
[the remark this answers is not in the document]
fsimonis: It's required for the implicit scheme when using mixed time-windows.

An explicit scheme may still require the samples before the time-window-start of the implicit scheme. This only remove the write-data samples provided by the implicit scheme itself.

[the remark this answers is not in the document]
fsimonis: You are correct. This exists out of mere necessity as both read and write data call `DataContext::mapData`. I refactored the trimming out of `mapData`. `mapReadData()` now performs this trim if a timestamp is given.

fsimonis:
@uekerman 

> Please also try to find good integration tests and really test whether the expected amount of map calls is done.

preCICE will not map more than necessary, so we are safe from wasting time.
If it maps less than necessary, then tests which check data should fail. And we have many tests like that.

The exact amount of mapping invocations could be accessible via an accessor fixture, but this sounds pretty invasive.

uekerman:
> preCICE will not map more than necessary

only if your implementation does what it promises :D how can we be sure without testing :wink: 

> The exact amount of mapping invocations could be accessible via an accessor fixture, but this sounds pretty invasive.

invasive yes, but could be pretty strong tests, no?

Anyway, I do not insist on such tests, but the correct amount of mapping operations should be really a point of emphasis when testing the release.

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -436,13 +437,19 @@ void ParticipantImpl::handleDataAfterAdvance(bool reachedTimeWindowEnd, bool isT
...
+
+    trimSendDataAfter(timeAfterAdvance);
   }
 
-  // We are iterating
-  PRECICE_ASSERT(math::greater(timeSteppedTo, timeAfterAdvance), "We must have moved back in time!");
+  if (reachedTimeWindowEnd) {
+    mapReadData(timeAfterAdvance);
+    performDataActions({action::Action::READ_MAPPING_POST});
+  }
```
uekerman, line 450: I am wondering and did not yet find a good answer: Does it make a difference that you now moved the "map" behind the `redetInitalGuesses`?
What's the motivation for changing the order here?

tests/serial/map-if-necessary/three-solvers/mixed-substeps/Multi.cpp
```diff
@@ -0,0 +1,31 @@
...
+BOOST_AUTO_TEST_SUITE(Serial)
+BOOST_AUTO_TEST_SUITE(MapIfNecessary)
+BOOST_AUTO_TEST_SUITE(ThreeSolvers)
+BOOST_AUTO_TEST_SUITE(MixedSubsteps)
+
+BOOST_AUTO_TEST_CASE(Multi)
+{
+  PRECICE_TEST("A"_on(1_rank), "B"_on(1_rank), "C"_on(1_rank));
+
+  std::vector<int> readMappings{2, 0, 2, 0, 2, 0};
```
uekerman, line 19: I am not sure I understand the zeros here. Could you comment?

uekerman (commented):
Really nice tests! Did not yet look through everything.

src/precice/impl/ParticipantImpl.cpp
fsimonis: First and foremost, the ordering wrt the reset was an oversight.

That said, doesn't it make sense to map the received data after convergence with a zero initial guess? After all, it belongs to the next time window.

Accidental correctness?

tests/serial/map-if-necessary/three-solvers/helper.hpp
```diff
@@ -0,0 +1,78 @@
+#pragma once
+
+#include <vector>
+
+#include "precice/impl/ParticipantImpl.hpp"
+#include "testing/Testing.hpp"
+
+// Coupling is A - C - B, which can be MultiCoupling or Compositional. A-C may be explicitly coupled.
+// A and B always go first.
+// A never sends substeps, B always sends substeps, which should be reflected in the read mappings of C.
```
uekerman, line 10: "B always sends substeps" is not correct for the "multi" case, for example, or did I get this wrong?

tests/serial/map-if-necessary/three-solvers/mixed-substeps/Multi.cpp
uekerman: I think the zeros are wrong. There should be new data from A and B also in the second iteration.
```suggestion
  std::vector<int> readMappings{2, 2, 2, 2, 2, 2};
```

tests/serial/map-if-necessary/three-solvers/mixed-substeps/ParallelExplicit.cpp
```diff
@@ -0,0 +1,31 @@
...
+BOOST_AUTO_TEST_SUITE(Serial)
+BOOST_AUTO_TEST_SUITE(MapIfNecessary)
+BOOST_AUTO_TEST_SUITE(ThreeSolvers)
+BOOST_AUTO_TEST_SUITE(MixedSubsteps)
+
+BOOST_AUTO_TEST_CASE(ParallelExplicit)
+{
+  PRECICE_TEST("A"_on(1_rank), "B"_on(1_rank), "C"_on(1_rank));
+
+  std::vector<int> readMappings{2, 2, 2};
```
uekerman, line 19:
```suggestion
  std::vector<int> readMappings{3, 3, 3};
```
1 from A, 2 from B?

tests/serial/map-if-necessary/three-solvers/mixed-substeps/ParallelImplicit.cpp
```diff
@@ -0,0 +1,34 @@
...
+BOOST_AUTO_TEST_SUITE(Serial)
+BOOST_AUTO_TEST_SUITE(MapIfNecessary)
+BOOST_AUTO_TEST_SUITE(ThreeSolvers)
+BOOST_AUTO_TEST_SUITE(MixedSubsteps)
+
+BOOST_AUTO_TEST_CASE(ParallelImplicit)
+{
+  PRECICE_TEST("A"_on(1_rank), "B"_on(1_rank), "C"_on(1_rank));
+
+  std::vector<int> readMappings{3, 0, 3, 0, 3, 0};
```
uekerman, line 19:
```suggestion
  std::vector<int> readMappings{3, 1, 3, 1, 3, 1};
```
The new data from B still needs to be mapped??

uekerman (changes requested):
I did not yet look through all tests, but I have some doubts with the first few.

tests/serial/map-if-necessary/three-solvers/mixed-substeps/ParallelExplicit.cpp
fsimonis: Parallel explicit doesn't send substeps as they would never be reachable.

So 2, 2, 2 is correct.

fsimonis:
Looks like we still need information on what data will be received.

Problem is that the coupling scheme receives new data into the receive data.
It however has no idea about read mappings, so it cannot remove existing samples in there.

When iterating and before before executing the read mappings, the Participant cannot simply remove all received toData. The received data from explicit schemes hasn't changed and remapping it would be a waste of resources.

We need to implement `CouplingScheme::receivedData()` to be able to check what data was received and then clean `toData` with the matching dataid.

An alternative would be to keep some `new` flags in stamples, but that feels very error-prone.

src/precice/impl/ParticipantImpl.cpp
fsimonis: Restored

fsimonis:
I am pretty confident, that the tests are now testing the readMappings correctly.

Especially the serial tests with three solvers are really _really_ mindbending.
In iterations, the implicit scheme doesn't receive start data, while the explicit scheme already moves on in the first iteration, providing the whole window. Then at the time window end, the implicit scheme provides the full time window.

At least parallel schemes are somewhat straight forward.

src/cplscheme/CouplingScheme.hpp
```diff
@@ -226,6 +226,12 @@ class CouplingScheme {
 
   /// Returns false if the scheme is implicit and hasn't converged
   virtual bool hasConverged() const = 0;
+
+  /// Returns true if any send data of the scheme requires substeps
+  virtual bool requiresSubsteps() const = 0;
+
+  /// Returns a vector of implicit data to receive in the next advance
+  virtual std::vector<DataID> implicitDataToReceive() const = 0;
```
BenjaminRodenberg, line 234:
```suggestion
  /// @brief Returns true if any send data of the scheme requires substeps
  virtual bool requiresSubsteps() const = 0;

  /// @brief Returns a vector of implicit data to receive in the next advance
  virtual std::vector<DataID> implicitDataToReceive() const = 0;
```

src/cplscheme/BaseCouplingScheme.hpp
```diff
@@ -396,6 +396,10 @@ class BaseCouplingScheme : public CouplingScheme {
    */
   bool reachedEndOfTimeWindow() const;
 
+  bool requiresSubsteps() const override final;
+
+  std::vector<DataID> implicitDataToReceive() const override final;
```
BenjaminRodenberg, line 401:
```suggestion
  /// @copydoc cplscheme::CouplingScheme::requiresSubsteps()
  bool requiresSubsteps() const override final;

  /// @copydoc cplscheme::CouplingScheme::implicitDataToReceive()
  std::vector<DataID> implicitDataToReceive() const override final;
```

src/cplscheme/CompositionalCouplingScheme.hpp
```diff
@@ -204,6 +204,10 @@ class CompositionalCouplingScheme final : public CouplingScheme {
   /// True if the implicit scheme has converged or no implicit scheme is defined
   bool hasConverged() const final;
 
+  bool requiresSubsteps() const final override;
+
+  std::vector<DataID> implicitDataToReceive() const override final;
```
BenjaminRodenberg, line 209:
```suggestion
  /// @copydoc cplscheme::CouplingScheme::requiresSubsteps()
  bool requiresSubsteps() const override final;

  /// @copydoc cplscheme::CouplingScheme::implicitDataToReceive()
  std::vector<DataID> implicitDataToReceive() const override final;
```

Do we prefer `final override` or `override final`? In the end it does not really matter, but would be good to do this consistently and maybe even enforce it with the linter.

src/precice/impl/ParticipantImpl.hpp
```diff
@@ -327,6 +334,12 @@ class ParticipantImpl {
   /// Counts calls to advance for plotting.
   long int _numberAdvanceCalls = 0;
 
+  /// Counts the amount of samples mapped in write mappings executed in the latest advance
+  int _executedWriteMappings = 0;
+
+  /// Counts the amount of samples mapped in read mappings executed in the latest advance
+  int _executedReadMappings = 0;
```
BenjaminRodenberg, line 341: Is there a better way to deal with these "only for testing and debugging"-variables? It just looks a bit strange to me to have them on the same level as variables that are needed for the functionality of the code. I guess what we also do not want is that developers in the future start using these variables for implementing actual functionality. Maybe adding comment here already serves the purpose of avoiding this kind of wrong usage.
```suggestion
  /// Counts the amount of samples mapped in write mappings executed in the latest advance. Should only be used for testing.
  int _executedWriteMappings = 0;

  /// Counts the amount of samples mapped in read mappings executed in the latest advance. Should only be used for testing.
  int _executedReadMappings = 0;
```

src/precice/impl/ParticipantImpl.hpp
```diff
@@ -363,12 +376,14 @@ class ParticipantImpl {
   /// Helper for mapWrittenData and mapReadData
   void computeMappings(std::vector<MappingContext> &contexts, const std::string &mappingType);
 
-  /// Computes, performs, and resets all suitable write mappings.
-  void mapWrittenData();
+  /// Computes, and performs suitable write mappings either entirely or after given time
+  void mapWrittenData(std::optional<double> after = std::nullopt);
 
-  /// Computes, performs, and resets all suitable read mappings.
+  // Computes, and performs read mappings optionally removes samples in toData after the optional time
```
BenjaminRodenberg, line 382:
```suggestion
  // Computes and performs read mappings. Optionally removes samples in toData after the optional time.
```

Is the second part of the comment true? Which optional time do you mean?

src/precice/impl/ParticipantImpl.hpp
```diff
@@ -363,12 +376,14 @@ class ParticipantImpl {
...
   void mapReadData();
 
+  void trimReadMappedData(double timeAfterAdvance, bool isTimeWindowComplete, std::vector<DataID> fromData);
```
BenjaminRodenberg, line 385: Please add documentation

src/time/Storage.cpp
```diff
@@ -134,6 +134,15 @@ void Storage::trimAfter(double time)
   _bspline.reset();
 }
 
+void Storage::trimAfterIncluding(double time)
```
BenjaminRodenberg, line 137: Can you add a unit test for this function in `StorageTest.cpp`?

BenjaminRodenberg (approved):
Only minor comments.

* The update of the tests is something we can also do later when we need it, because I assume this is quite a lot of work with relatively low benefit.
* Documentation should be simple

src/cplscheme/CouplingScheme.hpp
fsimonis: The first line defaults to brief :wink:

src/precice/impl/ParticipantImpl.hpp
fsimonis: These were my exact thoughts too, which is why I added a debug statement, which prints this information after the coupling state.

After all, it may be interesting to see as a short summary of what has been mapped.

tests/serial/map-if-necessary/two-solvers/mixed-substeps/ParallelExplicit.cpp
```diff
@@ -0,0 +1,33 @@
...
+BOOST_AUTO_TEST_CASE(ParallelExplicit)
+{
+  PRECICE_TEST("One"_on(1_rank), "Two"_on(1_rank));
+
+  // We receive the new start sample
+  std::vector<int> readMappings{1, 1, 1};
+  // We write the two samples from the two time steps
+  std::vector<int> writeMappings{2, 2, 2};
+
+  runTwoSolversMappingCountExplicit(context, readMappings, writeMappings);
```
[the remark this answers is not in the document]
fsimonis: These tests are designed to have one particiapant that does things and others that only feed it samples. The others will always map 0 and they also test that.

tests/serial/map-if-necessary/two-solvers/helper.hpp
```diff
@@ -0,0 +1,86 @@
...
+
+      if (context.isNamed("One")) {
+        BOOST_REQUIRE(timeWindow < writeMappings.size());
+        BOOST_TEST(mappedTimeWindow.write == writeMappings.at(timeWindow));
+        BOOST_TEST(mappedTimeWindow.read == readMappings.at(timeWindow));
+      } else {
+        // Solver Two has no mappings
+        BOOST_TEST(mappedTimeWindow.write == 0);
+        BOOST_TEST(mappedTimeWindow.read == 0);
+      }
```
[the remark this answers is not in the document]
fsimonis: But then, we need to provide data for a solver that by design never does anything.

src/time/Storage.cpp
fsimonis: Ah, I forgot to remove the function.

I realized at some point that `clear()` is all we need.

tests/serial/map-if-necessary/three-solvers/helper.hpp
fsimonis: image

Correct

src/cplscheme/CompositionalCouplingScheme.hpp
fsimonis: I think `final override` makes visually more sense.
No real point trying to enforce this unless clang-format can deal with this automatically.

tests/serial/map-if-necessary/two-solvers/helper.hpp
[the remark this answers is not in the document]
uekerman: Agree with @fsimonis

tests/serial/map-if-necessary/two-solvers/mixed-substeps/SerialImplicit.cpp
```diff
@@ -0,0 +1,34 @@
...
+BOOST_AUTO_TEST_SUITE(TwoSolvers)
+BOOST_AUTO_TEST_SUITE(MixedSubsteps)
+
+BOOST_AUTO_TEST_CASE(SerialImplicit)
+{
+  PRECICE_TEST("One"_on(1_rank), "Two"_on(1_rank));
+
+  // 1: iterating: new end
+  // 2: converged: new start and end
+  std::vector<int> readMappings{1, 2, 1, 2, 1, 1};
```
uekerman, line 21:
```suggestion
  // 1: iterating: new end
  // 2: converged: new start and end
  std::vector<int> readMappings{1, 1, 1, 1, 1, 0};
```
but shouldn't the "new start" be the "end" of the previous time window? Hence the data was already mapped before? What am I not getting?

uekerman (changes requested):
Last doubt

tests/serial/map-if-necessary/two-solvers/mixed-substeps/SerialImplicit.cpp
fsimonis: Hmm. Right. The same holds for the other SerialImplicit cases.

tests/serial/map-if-necessary/two-solvers/with-substeps/SerialImplicit.cpp
```diff
@@ -0,0 +1,34 @@
...
+BOOST_AUTO_TEST_SUITE(TwoSolvers)
+BOOST_AUTO_TEST_SUITE(WithSubsteps)
+
+BOOST_AUTO_TEST_CASE(SerialImplicit)
+{
+  PRECICE_TEST("One"_on(1_rank), "Two"_on(1_rank));
+
+  // 2: iterating: new mid + end
+  // 3: new time window: new start + mid + end
+  std::vector<int> readMappings{2, 3, 2, 3, 2, 1};
```
fsimonis, line 21: @uekerman 

```suggestion
  // iterating: new mid + end
  // new time window: mid + end
  std::vector<int> readMappings{2, 2, 2, 2, 2, 1};
```

tests/serial/map-if-necessary/two-solvers/without-substeps/SerialImplicit.cpp
```diff
@@ -0,0 +1,34 @@
...
+BOOST_AUTO_TEST_SUITE(TwoSolvers)
+BOOST_AUTO_TEST_SUITE(WithoutSubsteps)
+
+BOOST_AUTO_TEST_CASE(SerialImplicit)
+{
+  PRECICE_TEST("One"_on(1_rank), "Two"_on(1_rank));
+
+  // 1: new end only (iterating)
+  // 2: new start and end
+  std::vector<int> readMappings{1, 2, 1, 2, 1, 1};
```
fsimonis, line 21:
```suggestion
  // iterating: new end sample of last iteration
  // converging: new end sample of 1st iteration of this time window
  std::vector<int> readMappings{1, 2, 1, 2, 1, 1};
```

tests/serial/map-if-necessary/two-solvers/mixed-substeps/ParallelImplicit.cpp
```diff
@@ -0,0 +1,31 @@
...
+BOOST_AUTO_TEST_SUITE(Serial)
+BOOST_AUTO_TEST_SUITE(MapIfNecessary)
+BOOST_AUTO_TEST_SUITE(TwoSolvers)
+BOOST_AUTO_TEST_SUITE(MixedSubsteps)
+
+BOOST_AUTO_TEST_CASE(ParallelImplicit)
+{
+  PRECICE_TEST("One"_on(1_rank), "Two"_on(1_rank));
+
+  std::vector<int> readMappings{1, 1, 1, 1, 1, 1};
```
fsimonis, line 19: When moving to the next time window, there is no sample to map
```suggestion
  std::vector<int> readMappings{1, 0, 1, 0, 1, 0};
```

tests/serial/map-if-necessary/two-solvers/with-substeps/ParallelImplicit.cpp
```diff
@@ -0,0 +1,34 @@
...
+BOOST_AUTO_TEST_SUITE(TwoSolvers)
+BOOST_AUTO_TEST_SUITE(WithSubsteps)
+
+BOOST_AUTO_TEST_CASE(ParallelImplicit)
+{
+  PRECICE_TEST("One"_on(1_rank), "Two"_on(1_rank));
+
+  // 2: we fail to converge and receive new samples for the substep and end of the time window
+  // 0: we succeed to converge and keep the sample of the end of the time window
+  std::vector<int> readMappings{2, 1, 2, 1, 2, 1};
```
fsimonis, line 21: When moving to the next time-window, there is no sample to map
```suggestion
  std::vector<int> readMappings{2, 0, 2, 0, 2, 0};
```

tests/serial/map-if-necessary/two-solvers/without-substeps/ParallelImplicit.cpp
```diff
@@ -0,0 +1,34 @@
...
+BOOST_AUTO_TEST_SUITE(TwoSolvers)
+BOOST_AUTO_TEST_SUITE(WithoutSubsteps)
+
+BOOST_AUTO_TEST_CASE(ParallelImplicit)
+{
+  PRECICE_TEST("One"_on(1_rank), "Two"_on(1_rank));
+
+  // 1: we fail to converge and receive a new sample for the end of the time window
+  // 0: we succeed to converge and keep the sample of the end of the time window
+  std::vector<int> readMappings{1, 1, 1, 1, 1, 1};
```
fsimonis, line 21: When moving to the next time window, there is no sample to map
```suggestion
  std::vector<int> readMappings{1, 0, 1, 0, 1, 0};
```

tests/serial/map-if-necessary/two-solvers/mixed-substeps/ParallelImplicit.cpp
uekerman: I am not sure I get this one. The end of the window should be new when moving to the next time window, no?

fsimonis:
Turns out that this became a bit more complicated than anticipated.

Data received on the second participant in a serial implicit coupling needs special treatment.

uekerman (approved):
Rather confident that we fixed all problems now, but would be good that somebody else also looks at the tests again. Also simply to build up some knowledge.
I probably looked at the test too often already.

tests/serial/map-if-necessary/three-solvers/mixed-substeps/ParallelExplicit.cpp
```diff
@@ -0,0 +1,31 @@
...
+  std::vector<int> writeMappings{4, 4, 4};
```
davidscn, line 20: Why do we still have four write mappings? The substeps shouldn't be mapped either in an explicit coupling scheme, no?

tests/serial/map-if-necessary/three-solvers/without-substeps/SerialExplicit.xml
```diff
@@ -0,0 +1,66 @@
...
+    <time-window-size value="1.0" />
+    <exchange data="DataA" mesh="MeshA" from="A" to="C" substeps="false" />
+    <exchange data="DataC" mesh="MeshA" from="C" to="A" substeps="false" />
+  </coupling-scheme:serial-explicit>
+
+  <coupling-scheme:serial-explicit>
+    <participants first="B" second="C" />
+    <max-time-windows value="3" />
+    <time-window-size value="1.0" />
+    <exchange data="DataB" mesh="MeshB" from="B" to="C" substeps="true" />
```
davidscn, line 63: Why is the test directory called 'without substeps', when the option is actually enabled here?

tests/serial/map-if-necessary/two-solvers/without-substeps/SerialExplicit.cpp
```diff
@@ -0,0 +1,33 @@
...
+BOOST_AUTO_TEST_SUITE(MapIfNecessary)
+BOOST_AUTO_TEST_SUITE(TwoSolvers)
+BOOST_AUTO_TEST_SUITE(WithoutSubsteps)
+
+BOOST_AUTO_TEST_CASE(SerialExplicit)
+{
+  PRECICE_TEST("One"_on(1_rank), "Two"_on(1_rank));
+
+  // We receive the new start sample
+  std::vector<int> readMappings{1, 1, 1};
```
davidscn, line 20: Are the mappings here 'initial data (0), 1 and 2' or the the last time step (3) actually mapped. From what I understand the third timestep shouldn't be mapped.

tests/serial/map-if-necessary/two-solvers/without-substeps/ParallelExplicit.cpp
```diff
@@ -0,0 +1,33 @@
...
+BOOST_AUTO_TEST_SUITE(WithoutSubsteps)
+
+BOOST_AUTO_TEST_CASE(ParallelExplicit)
+{
+  PRECICE_TEST("One"_on(1_rank), "Two"_on(1_rank));
+
+  // We receive the new start sample
+  std::vector<int> readMappings{1, 1, 1};
+  // We write the sample from the end of the time window
+  std::vector<int> writeMappings{1, 1, 1};
```
davidscn, line 22: Similar here, the last write mapping, is never exchanged to the other participant, is it?

davidscn (commented):
Overall, I can grasp the core logic in the `DataContext::mapData`. However, the implementation makes it incredible hard to understand at which time which data is accessible/available or erased. Shouldn't be considered a denying review, but without further help of documentation or some refactoring in the near future (unlikely), it will be difficult for newcomers and also for us.

The tests make sense.

> enabling substeps on a single exchange enables samples from subcycling for the participant, which leads to the evaluation of all write mappings

Is it also true for parallel coupling schemes? Here, reading sampled values doesn't make sense as we are already in the next time window, right?
I also don't fully understand, how we deal with the final timestep. Example: in serial implicit coupling schemes, we drop the last mapping coming from the first participant (makes sense). In serial explicit coupling schemes, we should drop the last mapping from the second participant, no? Maybe that's also yet another corner case which makes the implementation more complex for no real reason.

tests/serial/map-if-necessary/three-solvers/mixed-substeps/ParallelExplicit.cpp
davidscn: Probably explained in the PR description, right?

tests/serial/map-if-necessary/two-solvers/without-substeps/SerialExplicit.cpp
fsimonis: They are `{after first time window, after second time window, after third time window}` and are checked on the second participant of the serial scheme.

Starting with the mappings of the initial data would probably be very useful, especially for the given case of a `second` participant.

You are right though, there shouldn't be a mapping after the last time window.

tests/serial/map-if-necessary/two-solvers/without-substeps/ParallelExplicit.cpp
fsimonis: The last time window is still exchanged as usual. This allows to read the data after `isCouplingOngoing() == false`.

tests/serial/map-if-necessary/three-solvers/without-substeps/SerialExplicit.xml
fsimonis: With/Mixed/Without is from the perspective of the tested solver sending data. In case of the three-solver tests, that's the central one.

With exchanges substeps to both, mixed to one participant and without to none.

By design, the central solver receives no substeps from A and one substep from B.

I'll put this information into a readme.

fsimonis:
> Overall, I can grasp the core logic in the `DataContext::mapData`. However, the implementation makes it incredible hard to understand at which time which data is accessible/available or erased. Shouldn't be considered a denying review, but without further help of documentation or some refactoring in the near future (unlikely), it will be difficult for newcomers and also for us.

I totally agree that the logic of when, what is mapped and reset is very complicated and obscure.

However, I have to revisit all of this again to get dynamic meshes to work. So, it is in my own interest to clean this up in the upcoming months.

> 
> > enabling substeps on a single exchange enables samples from subcycling for the participant, which leads to the evaluation of all write mappings
> 
> Is it also true for parallel coupling schemes? Here, reading sampled values doesn't make sense as we are already in the next time window, right?

For parallel-implicit this is correct. For parallel-explicit, we currently throw an error if the user configures `substeps=true`, which doesn't make sense. 

> I also don't fully understand, how we deal with the final timestep. Example: in serial implicit coupling schemes, we drop the last mapping coming from the first participant (makes sense). In serial explicit coupling schemes, we should drop the last mapping from the second participant, no? Maybe that's also yet another corner case which makes the implementation more complex for no real reason.

Every participant has the latest final data at the end of the coupling scheme.
Serial explicit is essentially serial implicit with a convergence measure that directly converges. The mapping logic is identical and sending the result of the final time step from second to first is required for allowing the adapter to read the final data.

tests/serial/map-if-necessary/two-solvers/without-substeps/SerialExplicit.cpp
fsimonis: You have a good eye for details! First and second are swapped in the configuration.

I'll make them consistent.

tests/serial/map-if-necessary/three-solvers/without-substeps/SerialExplicit.xml
fsimonis: Added the info to both helpers

PR #1379, opened 2022-07-21
Extend Aitken acceleration to handle multiple data sets

davidscn:
## Main changes of this PR
Extends the Aitken underrelaxation to handle multiple data sets (resolves #1360 and resolves #1184), adds a preconditioner option to our Aitken acceleration and add a proper unit test for the Aitken acceleration.

## Motivation and additional information
Depends on #1373 

## Author's checklist

* [x] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [x] I ran `make format` to ensure everything is formatted correctly.
* [x] I sticked to C++14 features.
* [x] I sticked to CMake version 3.16.3.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [x] Does the changelog entry make sense? Is it formatted correctly?
* [x] Do you understand the code changes?

PR #1900, opened 2023-11-23
Add test for Aitken with preconditioner

davidscn:
## Main changes of this PR

... to complete https://github.com/precice/precice/pull/1379/

## Author's checklist

### DIFF
--- a/src/acceleration/Acceleration.cpp
+++ b/src/acceleration/Acceleration.cpp
@@ -22,7 +22,8 @@ void Acceleration::applyRelaxation(double omega, DataMap &cplData)
     if (couplingData.timeStepsStorage().empty()) {
       continue;
     }
-
+    // @todo: This will apply the scaling to the sample at t=0 and t=1 when
+    // calling performAcceleration the first time. Is that intended?
     for (auto &stample : couplingData.timeStepsStorage().stamples()) {
       auto &values    = stample.sample.values;
       auto  oldValues = couplingData.getPreviousValuesAtTime(stample.timestamp); // IMPORTANT DETAIL: The interpolation that we use for resampling does not necessarily have to be the same interpolation as the interpolation the user accesses via read-data. (But probably it is easier to just use the same)
--- a/src/acceleration/Acceleration.cpp
+++ b/src/acceleration/Acceleration.cpp
@@ -23,7 +23,8 @@ void Acceleration::applyRelaxation(double omega, DataMap &cplData)
       continue;
     }
     // @todo: This will apply the scaling to the sample at t=0 and t=1 when
-    // calling performAcceleration the first time. Is that intended?
+    // calling performAcceleration the first time. Computations at the
+    // previousValuesAtTime/oldValues don't change anything and are unneeded
     for (auto &stample : couplingData.timeStepsStorage().stamples()) {
       auto &values    = stample.sample.values;
       auto  oldValues = couplingData.getPreviousValuesAtTime(stample.timestamp); // IMPORTANT DETAIL: The interpolation that we use for resampling does not necessarily have to be the same interpolation as the interpolation the user accesses via read-data. (But probably it is easier to just use the same)
--- a/src/acceleration/AitkenAcceleration.cpp
+++ b/src/acceleration/AitkenAcceleration.cpp
@@ -80,7 +80,6 @@ void AitkenAcceleration::performAcceleration(
     // preconditioner not necessary
     _aitkenFactor = math::sign(_aitkenFactor) * std::min(_initialRelaxation, std::abs(_aitkenFactor));
   } else {
-    // Todo: make this conditional for dataIDs > 1
     _preconditioner->update(false, _values, residuals);
     _preconditioner->apply(residualDeltas);
     _preconditioner->apply(_oldResiduals);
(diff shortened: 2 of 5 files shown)


src/acceleration/Acceleration.cpp
```diff
@@ -22,7 +22,8 @@ void Acceleration::applyRelaxation(double omega, DataMap &cplData)
     if (couplingData.timeStepsStorage().empty()) {
       continue;
     }
-
+    // @todo: This will apply the scaling to the sample at t=0 and t=1 when
+    // calling performAcceleration the first time. Is that intended?
```
davidscn, line 26: @BenjaminRodenberg is this correct?

src/acceleration/AitkenAcceleration.cpp
```diff
@@ -80,7 +80,6 @@ void AitkenAcceleration::performAcceleration(
     // preconditioner not necessary
     _aitkenFactor = math::sign(_aitkenFactor) * std::min(_initialRelaxation, std::abs(_aitkenFactor));
   } else {
-    // Todo: make this conditional for dataIDs > 1
```
davidscn, line 83: @uekerman this leftover TODO doesn't make sense to me, what do you think?

Also, we apply the scaling ('preconditioning') to data used for the computation of the `_aitkenFactor`) and not for the coupling data itself. Is that correct? I think we discussed that in the past, but I'm asking once more to be sure.

src/acceleration/Acceleration.cpp
BenjaminRodenberg: At `t=0` the value of the sample should be stationary from one iteration to the next, so the acceleration should not have a visible effect. But applying the acceleration in the first place also looks unnecessary to me.

davidscn: > With extraplation, these configuration files should also trigger some error. If not, then we have a bug in preCICE. Is this the case?

No, it has no effect

> But applying the acceleration in the first place also looks unnecessary to me.

Yes, how do you handle this in other places?

uekerman: Other accelerations do not yet support waveforms. I would actually try to avoid including the window start. Could lead to singular systems.

src/acceleration/AitkenAcceleration.cpp
uekerman: You only need the "preconditioner" if you have more than one data field.
But does neither hurt to apply it.

> Also, we apply the scaling ('preconditioning') to data used for the computation of the _aitkenFactor) and not for the coupling data itself. Is that correct?

Exactly

src/acceleration/Acceleration.cpp
BenjaminRodenberg: I think you can copy-paste the "Don't use samples at the window start"-strategy pretty much from https://github.com/precice/precice/pull/1899. If this introduces too many changes here, let's try to merge https://github.com/precice/precice/pull/1899 first.

src/acceleration/AitkenAcceleration.cpp
davidscn: I made this now conditional and (please check) applied the scaling also in the first iteration, i.e., put the preconditioner out of the if statement. I think that's required for residual(sum) preconditioners which incorporate iteration history data.

src/acceleration/Acceleration.cpp
davidscn: I just took a closer look.

> Other accelerations do not yet support waveforms

That's only kind of true, as constant acceleration is supported and it also uses exactly the same functionality. So, this code section is actually part of https://github.com/precice/precice/pull/1794

> Could lead to singular systems.

Could you give an example of where this could happen? A cannot think of such an example as the `oldValue` will just be the same as before 

https://github.com/precice/precice/blob/586ac794e5e23bdc35aae623b919d9f260e5e474/src/acceleration/Acceleration.cpp#L29

> I think you can copy-paste the "Don't use samples at the window start"-strategy pretty much from https://github.com/precice/precice/pull/1899. If this introduces too many changes here, let's try to merge https://github.com/precice/precice/pull/1899 first.

The change here are not related to #1899 and I don't see a reason to connect these two things here.

davidscn: > Didn't you ask for a way how to not apply the acceleration to the sample at t=0

I was mainly wondering how you handle this elsewhere and the answer is 'we don't handle this until now'. I'm not familiar with the whole storage class, but I don't really understand why introducing a dedicated iterator range for all, but the first storage sample is not possible. We actually do this for the communication https://github.com/precice/precice/blob/586ac794e5e23bdc35aae623b919d9f260e5e474/src/utils/IntraComm.hpp#L37

Still, the PR here adds only test for a fix about the Aitken acceleration.

uekerman: > Could you give an example of where this could happen? A cannot think of such an example as the oldValue will just be the same as before

This can give problems for quasi-Newton, but not Aitken. Was only mention this as an answer to

> Yes, how do you handle this in other places?

src/acceleration/AitkenAcceleration.cpp
uekerman: In the first iteration `residualDeltas` and `_oldResiduals` are not used, so I don't think we need it here.
But shouldn't hurt either.

davidscn: It's not used in the `AitkenAcceleration` (is it what you mean?), but in the preconditioner class, e.g.,

uekerman: OK, I think I now understand what you mean. The `update` should indeed happen in every iteration. The `apply` only when needed.

davidscn: Yes, I thought it's not a big deal, moved it now back.

PR #1901, opened 2023-11-24
Fix EventRegistry stopping backend without starting it.

### LINKED ISSUE #1896: Some errors trigger a std::bad_optional_access exception

fsimonis:
## Main changes of this PR

This PR fixes the management of the back-end of the EventRegistry.

If an error occurs during configuration, then the registry was finalized before the back-end was started, resulting in a `std::bad_optional_access` exception.

Closes #1896
Superseeds #1898

Special thanks to @davidscn for finding the origin of the problem.

## Author's checklist

### DIFF
--- a/src/profiling/EventUtils.cpp
+++ b/src/profiling/EventUtils.cpp
@@ -69,6 +69,9 @@ void EventRegistry::initialize(std::string_view applicationName, int rank, int s
 
   _initialized = true;
   _finalized   = false;
+  if (_isBackendRunning) {
+    stopBackend();
+  }
 }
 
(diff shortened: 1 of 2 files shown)


davidscn:
Should have gone into #1898

Issue #1905, opened 2023-11-29
Forbid `readData` before `initialize` and `writeData` after last `advance`


fsimonis:
**Describe your setup**

preCICE Version: dev

**Describe the problem**

preCICE currently allows calling:
* `readData()` before calling `initialize()`
* `writeData()` after the last call to `advance()` after which `isCouplingOngoing() == false`

Both don't make sense.

**Expected behaviour**

These cases should be forbidden and lead to an error message.

**Additional context**

Somewhat related to https://github.com/precice/precice/issues/1904

fsimonis:
Solved by #1906

PR #1906, opened 2023-11-29
Make API more restrictive w.r.t correct context for calling `readData` and `writeData`

### LINKED ISSUE #1904: Don't allow calling `getMaxTimeWindowSize` and `readData` after last `advance`

Calling `readData` or `getMaxTimeWindowSize` after the last `advance` call (i.e. when `isCouplingOngoing` returns `false`) does not make sense, but is currently allowed.

Add security checks to disallow calling these API functions in the wrong context.

Leave things as they are. But this also requires us to be able to deal with this situation and we cannot make certain assumptions (e.g. `getMaxTimeWindowSize` *always* returns a `dt > 0`)

**Additional context**

Tests in #1867 failed due to `PRECICE_ASSERT(math::greater(nextTimeStepSize, 0.0), nextTimeStepSize)` in `double ParticipantImpl::getMaxTimeStepSize()`. This could be fixed by rearranging the API calls in the tests such that there is no `getMaxTimeStepSize` call after `advance`.

BenjaminRodenberg:
## Main changes of this PR

* Return `dt = 0.0`, if `dt = getMaxTimeStepSize()` is called after final `advance(dt)`
* Only allow relative read time `0.0`, if `readData(...)` is called after final `advance(dt)` (i.e. `not isCouplingOngoing()`)
* Don't allow calling `readData(...)` before `initialize()`
* Don't allow calling `writeData(...)` after final `advance(dt)` (i.e. `not isCouplingOngoing()`)
* Update tests correspondingly

Closes #1904 and #1905

Results in a clearer API with less potential for wrong usage.
Allows developers to make additional assumptions about state of coupling when calling `readData`, `writeData`, and `getMaxTimeStepSize`

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -484,6 +484,10 @@ int BaseCouplingScheme::getTimeWindows() const
 
 double BaseCouplingScheme::getNextTimeStepMaxSize() const
 {
+  if (!isCouplingOngoing()) {
+    return 0.0; // if coupling is not ongoing (i.e. coupling scheme reached end of window) the maximum time step size is zero
+  }
+
   if (hasTimeWindowSize()) {
     return _timeWindowStartTime + _timeWindowSize - _time;
   } else {
--- a/src/precice/impl/ParticipantImpl.cpp
+++ b/src/precice/impl/ParticipantImpl.cpp
@@ -1052,7 +1052,7 @@ void ParticipantImpl::readData(
   PRECICE_CHECK(_state != State::Finalized, "readData(...) cannot be called after finalize().");
   PRECICE_CHECK(math::smallerEquals(relativeReadTime, _couplingScheme->getNextTimeStepMaxSize()), "readData(...) cannot sample data outside of current time window.");
   PRECICE_CHECK(relativeReadTime >= 0, "readData(...) cannot sample data before the current time.");
-  PRECICE_CHECK(isCouplingOngoing() || relativeReadTime == 0, "Calling readData(...) with relativeReadTime = {} is forbidden if coupling is not ongoing. If coupling finished, only relativeReadTime = 0 is available. Please always use precice.getMaxTimeStepSize() to obtain the maximum allowed relativeReadTime.", relativeReadTime);
+  PRECICE_CHECK(isCouplingOngoing() || math::equals(relativeReadTime, 0.0), "Calling readData(...) with relativeReadTime = {} is forbidden if coupling is not ongoing. If coupling finished, only relativeReadTime = 0 is allowed. Please always use precice.getMaxTimeStepSize() to obtain the maximum allowed relativeReadTime.", relativeReadTime);
 
   PRECICE_REQUIRE_DATA_READ(meshName, dataName);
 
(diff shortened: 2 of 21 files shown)


BenjaminRodenberg:
I think the set of restrictions we now have is also nicely pushing our users into the direction to make the simulation loop follow the pattern 

```
dt -> read -> solve -> write -> advance
```
not

```
solve -> write -> advance -> dt -> read
```

This is also visible in the updates for the tests. The other approach also works, but the more natural approach is the first one.

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -1047,9 +1048,11 @@ void ParticipantImpl::readData(
     ::precice::span<double>         values) const
 {
   PRECICE_TRACE(meshName, dataName, vertices.size(), relativeReadTime);
+  PRECICE_CHECK(_state != State::Constructed, "readData(...) cannot be called before initialize().");
   PRECICE_CHECK(_state != State::Finalized, "readData(...) cannot be called after finalize().");
   PRECICE_CHECK(math::smallerEquals(relativeReadTime, _couplingScheme->getNextTimeStepMaxSize()), "readData(...) cannot sample data outside of current time window.");
   PRECICE_CHECK(relativeReadTime >= 0, "readData(...) cannot sample data before the current time.");
+  PRECICE_CHECK(isCouplingOngoing() || relativeReadTime == 0, "Calling readData(...) with relativeReadTime = {} is forbidden if coupling is not ongoing. If coupling finished, only relativeReadTime = 0 is allowed. Please always use precice.getMaxTimeStepSize() to obtain the maximum allowed relativeReadTime.", relativeReadTime);
```
davidscn, line 1055:
```suggestion
  PRECICE_CHECK(isCouplingOngoing() || relativeReadTime == 0, "Calling readData(...) with relativeReadTime = {} is forbidden if coupling is not ongoing. If coupling finished, only relativeReadTime = 0 is available. Please always use precice.getMaxTimeStepSize() to obtain the maximum allowed relativeReadTime.", relativeReadTime);
```

It's not something we came up with, but it's also that non-zero relativeReadTime is simply not available/ we are not able to compute it without extrapolating, right? It sounds at least for me quite different, whether we write 'we forbid' or 'what you are trying to do doesn't make sense/ we don't have the data you are requesting'.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -484,6 +484,10 @@ int BaseCouplingScheme::getTimeWindows() const
 
 double BaseCouplingScheme::getNextTimeStepMaxSize() const
 {
+  if (!isCouplingOngoing()) {
+    return 0.0; // if coupling is not ongoing (i.e. coupling scheme reached end of window) the maximum time step size is zero
+  }
```
davidscn, line 489: Shouldn't this be the result of `_maxTime-time` anyway?

src/precice/impl/ParticipantImpl.cpp
uekerman, line 1055: Better use `math::equals` here

uekerman (approved):
Makes sense to me

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: `_maxTime` is not necessarily given. You may not provide a `_maxTime`, if you provide `_maxTimeWindows`. Then `_maxTime = UNDEFINED_MAX_TIME`.

fsimonis (approved):
Nice and clean!

PR #1910, opened 2023-12-01
Prevent Floating-point exception during scaling

davidscn:
## Main changes of this PR

Prevents division by zero in case the coupling data is always zero or doesn't change over time.

Related to #1558


fsimonis:
@uekerman Should we include this in v3?

fsimonis (approved):

PR #1913, opened 2023-12-05
Simplify PETSc initialization

fsimonis:
## Main changes of this PR

This PR simplifies the PETSc initialization and fixes a bug in the RBF Configuration tests, which prevented from running all tests in one go (`mpirun -n 4 ./testprecice -t MappingTests` failed)

Remove weird code complexity

## Author's checklist


src/mapping/tests/MappingConfigurationTest.cpp
```diff
@@ -47,7 +47,7 @@ BOOST_AUTO_TEST_CASE(Configuration)
 
 BOOST_AUTO_TEST_CASE(RBFDirectConfiguration)
 {
-  PRECICE_TEST(1_rank);
+  PRECICE_TEST(1_rank, Require::PETSc);
```
davidscn, line 50: I don't see where this configuration test requires PETSc. If it is required for practical purposes, please add documentation here why it should be required.

fsimonis: Good catch! This slipped into the commit.

FYI I added PETSc to the alias tag test, as this may choose PETSc in the future.

PR #1916, opened 2023-12-08
Tidy the API documentation

fsimonis:
## Main changes of this PR

This PR aims to tidy the API documentation of Participant and add more context.

I used plantuml to generate some [activity diagrams](https://plantuml.com/activity-diagram-beta) in order to clarify the intent of how the API is supposed to be used.
Plantuml may still need to be installed in the CI of the website.

I didn't invest time into the gradient API as it is still experimental.
Mesh access remains unchanged.

We always say that the doxygen is for adapter developers that need to go deep.
Now it contains have more complete information.

## Screenshots of Sections in order

image

image

image

image

image
image

image

## Author's checklist

* [x] I added plantuml to the CI in the website and check that the path is setup

### DIFF
--- a/docs/documents/span.dox
+++ b/docs/documents/span.dox
@@ -10,7 +10,7 @@
  * A simple mental model is an object that holds a pointer and size, so it knows how many elements it is allowed to access after the initial pointer.
  * You can imagine a `span<double>` to look like this:
  *
- * @code cpp
+ * @code{cpp}
  * class span {
  *   double* begin;
  *   size_t size;
@@ -19,7 +19,7 @@
  *
  * This type can also be read-only aka `const`. You can imagine a `span<const double>` to look like this:
  *
- * @code cpp
+ * @code{cpp}
  * class span {
  *   double const* begin;
  *   size_t size;
--- a/docs/documents/span.dox
+++ b/docs/documents/span.dox
@@ -2,9 +2,36 @@
  *
  * @copyright Copyright Tristan Brindle 2018.
  *
- * @brief A C++ 11 implementation of the non-owning C++20 std::span type.
+ * @brief A C++ 11 implementation of the non-owning C++20 \ref std::span type.
  *
- * A span represents a non-owning view into a contiguous block of memory of a known size.
+ * A \ref span represents a non-owning view into a contiguous block of memory of a known size.
+ * This view can be mutable or const, depending on the type.
  *
- * This version contains an additional constructor for `const char *`-style C strings, allowing the span to mimic C++17 std::string_view.
+ * A simple mental model is an object that holds a pointer and size, so it knows how many elements it is allowed to access after the initial pointer.
+ * You can imagine a `span<double>` to look like this:
+ *
+ * @code cpp
+ * class span {
+ *   double* begin;
+ *   size_t size;
+ * }
+ * @endcode
+ *
+ * This type can also be read-only aka `const`. You can imagine a `span<const double>` to look like this:
+ *
+ * @code cpp
+ * class span {
+ *   double const* begin;
+ *   size_t size;
+ * }
+ * @endcode
+ *
+ * It can be constructed from:
+ * - A pointer and size
+ * - A pointer to first element and pointer to one past the last element
+ * - A C-style array `char[10]`
+ * - A C++ STL arrays `std::array<char,10>`
+ * - A contiguous range with size `end(R) - begin(R)`, such as a `std::vector`
+ *
+ * @note This version contains an additional constructor for `const char *`-style C strings, allowing the span to mimic C++17 \ref std::string_view.
  */
--- a/src/precice/Participant.cpp
+++ b/src/precice/Participant.cpp
@@ -133,10 +133,10 @@ void Participant::setMeshVertices(
 
 void Participant::setMeshEdge(
     ::precice::string_view meshName,
-    VertexID               firstVertexID,
-    VertexID               secondVertexID)
+    VertexID               first,
+    VertexID               second)
 {
-  _impl->setMeshEdge(toSV(meshName), firstVertexID, secondVertexID);
+  _impl->setMeshEdge(toSV(meshName), first, second);
 }
 
 void Participant::setMeshEdges(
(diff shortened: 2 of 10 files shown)


davidscn:
This looks very nice. As this is already targeting the API doxygen docs:

During some porting, I realized that we don't use consistent variable names across API functions. This would, however, help a lot.
Example:

```cpp
  void setMeshVertices(
      ::precice::string_view        meshName,
      ::precice::span<const double> positions,
      ::precice::span<VertexID>     ids);
```

```cpp
  void writeData(
      ::precice::string_view          meshName,
      ::precice::string_view          dataName,
      ::precice::span<const VertexID> vertices,
      ::precice::span<const double>   values);
```
Here, `ids` and `vertices` are from the user perspective the same data/variable. Naming them in the same way, would help to immediately recognize this relation as such.

carme-hp:
Is there a reason to use both `positions` and `coordinates`? I think it would be clearer to write `coordinates` everywhere.  

Other than that having `vertices` next to `coordinates` seems okayish to me and the diagrams are very useful.

docs/documents/span.dox
```diff
@@ -2,9 +2,36 @@
...
+ *   size_t size;
+ * }
+ * @endcode
+ *
+ * This type can also be read-only aka `const`. You can imagine a `span<const double>` to look like this:
+ *
+ * @code cpp
+ * class span {
+ *   double const* begin;
+ *   size_t size;
```
carme-hp, line 25: it would also be `size_t const size` right?

fsimonis: I had a similar conversation with @IshaanDesai lately.
While the viewed data cannot be changed in this example, the view itself can.
The pointed-to data is constant, not the pointer itself.

For example, one can still move the start of it.

src/precice/Participant.hpp
```diff
@@ -22,27 +22,180 @@ struct WhiteboxAccessor;
...
+ *
+ * Or you can pass something string-like:
+ *
+ * @code{.cpp}
+ * // one of
+ * const char*      meshName = "MyMesh";
+ * std::string      meshName = "MyMesh";
+ * std::string_view meshName = "MyMesh";
+ * // followed by
+ * getMeshDimensions(meshName);
```
davidscn, line 60: I'm not sure we should encourage people to use `string_view` or `char*` in their code. While all variants here work, I think the `std::string` (or the literals) are the way to go in CPP, as the user code typically also owns the data.

src/precice/Participant.hpp
```diff
@@ -22,27 +22,180 @@ struct WhiteboxAccessor;
...
+ * @endcode
+ *
+ * For more complex settings, you can pass pointer + size:
```
davidscn, line 63: What are those scenarios? This looks more like: "you can use the following constructor", but I could imagine that it confuses more than it helps.

src/precice/Participant.hpp
```diff
@@ -22,27 +22,180 @@ struct WhiteboxAccessor;
...
+   *
+   * The API of preCICE is accessible via the \ref Participant class.
+   * A constructed \ref Participant directly corresponds to a participant in the configuration file.
+   *
+   * Constructors require defining the parallel context of the Participant by providing
+   * index and size of the current process, which are equivalent to rank and size in the MPI terminology.
+   *
+   * If preCICE is compiled with MPI, then there are multiple ways for it to be used:
+   * 1. if the participant is configured to use an intra-comm using sockets, preCICE ignores MPI
+   * 2. if a custom communicator is provided, preCICE uses it
```
davidscn, line 193: Point one and two might be contradicting, what happens in the case where I provide a custom MPI comm and configure sockets?

src/precice/Participant.hpp
```diff
@@ -364,20 +569,20 @@ class PRECICE_API Participant {
...
-   * @pre position.size() == getMeshDimensions(meshName) * ids.size()
+   * @pre \p coordinates.size() == getMeshDimensions(meshName) * ids.size()
    *
    * @see getDimensions()
    */
   void setMeshVertices(
       ::precice::string_view        meshName,
-      ::precice::span<const double> positions,
+      ::precice::span<const double> coordinates,
       ::precice::span<VertexID>     ids);
```
davidscn, line 586: It's a bit arbitrary that we have here `VertexID`, but the single-call version `setMeshVertex` has the return type `int`. A brief explanation of the VertexID type would be necessary and also helpful to better understand what these IDs are and why they are needed.

davidscn (commented):
Some comments below

src/precice/Participant.hpp
fsimonis: I was thinking of such C-style functions:

```cpp
char meshName[MAX_MESH_NAME];
int length = getMeshNameFromObscurePlace(meshName);
getMeshDimensions(precice::string_view{meshName, length});
```

This is how MPI provides access to opened ports work for example.

src/precice/Participant.hpp
fsimonis: I looked this up again and realised that we always initialize MPI if we define a socket intracomm...

Fishy.

src/precice/Participant.hpp
fsimonis: Very good point!

src/precice/Participant.cpp
```diff
@@ -132,96 +125,96 @@ int Participant::getMeshVertexSize(
...
 }
 
 void Participant::setMeshTriangle(
     ::precice::string_view meshName,
-    int                    firstVertexID,
-    int                    secondVertexID,
-    int                    thirdVertexID)
+    VertexID               firstVertexID,
+    VertexID               secondVertexID,
+    VertexID               thirdVertexID)
```
fsimonis, line 153: Should we keep the parameter names this way?
Looks very redundant.

src/precice/Participant.hpp
fsimonis: Fixed

src/precice/Participant.hpp
davidscn: >  then we can also remove it.

That's no strong opinion from my side. For C codes, we anyway have a different interface.

src/precice/Participant.hpp
davidscn: > The point is to show that string_view can also be passed to the preCICE API.

Should be clear from the function signature, no? What you describe here is nice to know, but actually expected user knowledge in case they care about it I would say.

src/precice/Participant.cpp
davidscn: I t

src/precice/Participant.hpp
fsimonis: Sure, I can remove it. Less is sometimes more.

src/precice/Participant.cpp
MakisH: `first`, `second`, `third` sounds good, especially since we don't want to have much camelCase in the API.

It is indeed redundant now that the type is `VertexID` and having a short arguments list is usually easier to read. I don't think we lose any value by dropping the `...VertexID`.

fsimonis:
Many eyes and many changes. I'll merge this now into develop. We can iterate using another PR if necessary.

PR #1933, opened 2024-01-19
Fix time accumulation and max time

fsimonis:
## Main changes of this PR

This PR 
- replaces the time window start with a Kahan aggregator
- constrains the max dt by max-time

time-window-size = 0.01 and max-time = 5 leads to 501 time windows due to rounding errors.
This should solve the problem.

## Author's checklist

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -269,7 +269,8 @@ void BaseCouplingScheme::initialize(double startTime, int startTimeWindow)
   PRECICE_ASSERT(not isInitialized());
   PRECICE_ASSERT(math::greaterEquals(startTime, 0.0), startTime);
   PRECICE_ASSERT(startTimeWindow >= 0, startTimeWindow);
-  _timeWindowStartTime = startTime;
+  _timeWindowStartTime = KahanAccumulator{}; // reset the accumulator
+  _timeWindowStartTime(startTime);
   _timeWindows         = startTimeWindow;
   _hasDataBeenReceived = false;
 
@@ -345,7 +346,7 @@ void BaseCouplingScheme::secondExchange()
         // The computed time window part equals the time window size, since the
         // time window remainder is zero. Subtract the time window size and do another
         // coupling iteration.
-        PRECICE_ASSERT(math::greater(_time, _timeWindowStartTime));
+        PRECICE_ASSERT(math::greater(_time, getWindowStartTime()));
         _timeWindows -= 1;
         _isTimeWindowComplete = false;
       } else { // write output, prepare for next window
@@ -376,9 +377,9 @@ void BaseCouplingScheme::secondExchange()
 
     // Update internal time tracking
     if (_isTimeWindowComplete) {
-      _timeWindowStartTime += _timeWindowSize;
+      _timeWindowStartTime(_timeWindowSize);
     }
-    _time           = _timeWindowStartTime;
+    _time           = getWindowStartTime();
     _timeWindowSize = _nextTimeWindowSize;
   }
 }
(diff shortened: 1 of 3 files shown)


src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -489,7 +490,13 @@ double BaseCouplingScheme::getNextTimeStepMaxSize() const
   }
 
   if (hasTimeWindowSize()) {
-    return _timeWindowStartTime + _timeWindowSize - _time;
+    double maxDt = getWindowStartTime() + _timeWindowSize - _time;
+    if (math::equals(_maxTime, UNDEFINED_MAX_TIME)) {
+      return maxDt;
+    } else {
+      double leftover = _maxTime - _time;
+      return std::min(maxDt, leftover);
```
davidscn, line 498: Do we really want to forward here an arbitrary small dt? I would say that the current behavior is numerically still more robust. Not sure that this PR is the right place to add it into the library.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -376,9 +377,9 @@ void BaseCouplingScheme::secondExchange()
 
     // Update internal time tracking
     if (_isTimeWindowComplete) {
-      _timeWindowStartTime += _timeWindowSize;
+      _timeWindowStartTime(_timeWindowSize);
     }
-    _time           = _timeWindowStartTime;
+    _time           = getWindowStartTime();
```
davidscn, line 382: For consistency reasons, I would consider to make the (intermediate buffer) `_time` also such an accumulator, as we store the subcycling substeps in it.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -489,7 +490,13 @@ double BaseCouplingScheme::getNextTimeStepMaxSize() const
   }
 
   if (hasTimeWindowSize()) {
-    return _timeWindowStartTime + _timeWindowSize - _time;
+    double maxDt = getWindowStartTime() + _timeWindowSize - _time;
```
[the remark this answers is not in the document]
fsimonis: Once we change _time to a Kahan aggregate too, then this should be no problem.

src/cplscheme/BaseCouplingScheme.cpp
davidscn: With participant-first methods and adaptive time-stepping (e.g. in OpenFOAM) this can still lead to issues. In the adapter, we set the end time to infinity such that preCICE steers it. Hence, the solver is not aware of the maxTime limit and this return value can lead o arbitrary small dt causing numerical issues then.

davidscn: I just realized that this only applies to `hasTimeWindowSize()`. What happens to the second participant in a `participant-first` time-stepping. Do we then perform different time-step length for different participants?

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: I think we will only be able to solve the problem that triggered this whole discussion, if we turn `_time` into such a buffer. `_timeWindowStartTime` is a nice-to-have (probably similar problem on a different level), but `_time` is the actual problem. But let's first add a test and then we will know.

src/cplscheme/BaseCouplingScheme.cpp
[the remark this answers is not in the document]
fsimonis: Will be fine after https://github.com/precice/precice/pull/1934

src/cplscheme/BaseCouplingScheme.cpp
fsimonis: For `participant-first`, the `getNextTimeStepMaxSize()` is maxTime if provided and inf otherwise.

PR #1939, opened 2024-01-29
Improve time handling

fsimonis:
## Main changes of this PR

This PR further improves the time handling in the BaseCouplingScheme.

Using absolute times for both, the time window start time and the current time required to compute the current time window progress.
This was numerically problematic and lead to cancellation issues in computing the remainder of the time window.

We now store:
* the start time of the time window
* the progress in the current time window
* the remaining simulation time (essentially `max time - time window start time`)

This should eliminate all critical parts of cancellation.

Remaining are:
* `double maxDt = _timeWindowSize - getTimeWindowProgress()` to get the remaining time inside the time window
* `double leftoverTime = getRemainingTime() - getTimeWindowProgress()`, to get the remaining time of the simulation relative from the current time step

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -205,14 +205,13 @@ void BaseCouplingScheme::receiveData(const m2n::PtrM2N &m2n, const DataMap &rece
 void BaseCouplingScheme::receiveDataForWindowEnd(const m2n::PtrM2N &m2n, const DataMap &receiveData)
 {
   // buffer current time
-  const double oldProgress = getTimeWindowProgress();
+  auto oldProgress = _timeWindowProgress;
   // set _time state to point to end of this window such that _time in receiveData is at end of window
   _timeWindowProgress(_nextTimeWindowSize);
   // receive data for end of window
   this->receiveData(m2n, receiveData);
   // reset time state;
-  _timeWindowProgress = KahanAccumulator{};
-  _timeWindowProgress(oldProgress); // This should be 0, right?
+  _timeWindowProgress = oldProgress;
 }
 
 void BaseCouplingScheme::initializeWithZeroInitialData(const DataMap &receiveData)
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -205,14 +205,14 @@ void BaseCouplingScheme::receiveData(const m2n::PtrM2N &m2n, const DataMap &rece
 void BaseCouplingScheme::receiveDataForWindowEnd(const m2n::PtrM2N &m2n, const DataMap &receiveData)
 {
   // buffer current time
-  const double oldTime = getTime();
+  const double oldProgress = getTimeWindowProgress();
   // set _time state to point to end of this window such that _time in receiveData is at end of window
-  _time(_nextTimeWindowSize);
+  _timeWindowProgress(_nextTimeWindowSize);
   // receive data for end of window
   this->receiveData(m2n, receiveData);
   // reset time state;
-  _time = KahanAccumulator{};
-  _time(oldTime);
+  _timeWindowProgress = KahanAccumulator{};
+  _timeWindowProgress(oldProgress); // This should be 0, right?
 }
 
 void BaseCouplingScheme::initializeWithZeroInitialData(const DataMap &receiveData)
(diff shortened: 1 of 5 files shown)


src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -205,14 +205,14 @@ void BaseCouplingScheme::receiveData(const m2n::PtrM2N &m2n, const DataMap &rece
...
   // set _time state to point to end of this window such that _time in receiveData is at end of window
-  _time(_nextTimeWindowSize);
+  _timeWindowProgress(_nextTimeWindowSize);
   // receive data for end of window
   this->receiveData(m2n, receiveData);
   // reset time state;
-  _time = KahanAccumulator{};
-  _time(oldTime);
+  _timeWindowProgress = KahanAccumulator{};
+  _timeWindowProgress(oldProgress); // This should be 0, right?
```
BenjaminRodenberg, line 215: Yes. Otherwise, I think it would also be better to not initialize the Kahan accumulator with `double oldProgress`, but instead recover the actual `KahanAccumulator _timeWindowProgress` from above.

Maybe add an assertion to be sure before overwriting `_timeWindowProgress`: ``PRECICE_ASSERT(math::equals(getTimeWindowProgress(), 0.0));`

tests/time-handling/SimpleMaxTime.cpp
BenjaminRodenberg, line 1: Did you also try this test with the `develop` version of the code? Is it failing? I think my example was slightly different with `<max-time value="0.03" />`.

BenjaminRodenberg (commented):
I'll try this version out with my code. W.r.t readability, I think it would be good, if somebody who is not so familiar with this part of the code. There are some tricky and non-obvious parts.

tests/time-handling/SimpleMaxTime.cpp
fsimonis: I increased the max-time to 0.04 as it first shows a shorter timestep followed by a longer timestep.

Just to make sure both are fine.

src/cplscheme/BaseCouplingScheme.cpp
fsimonis: This assertion is triggered straight away.

tests/time-handling/SimpleMaxTime.cpp
```diff
@@ -0,0 +1,45 @@
...
+  p.advance(p.getMaxTimeStepSize());
+
+  BOOST_TEST(p.getMaxTimeStepSize() == expected);
+  BOOST_TEST(p.isCouplingOngoing());
+  p.advance(p.getMaxTimeStepSize());
+
+  BOOST_REQUIRE(p.isCouplingOngoing());
+  // BUG was: return value 0.009999999999999998
+  BOOST_TEST(p.getMaxTimeStepSize() == expected);
+  p.advance(p.getMaxTimeStepSize());
```
BenjaminRodenberg, line 31: I think whether this high accuracy is an overkill or not depends a lot on what we expect from preCICE here. If I tell preCICE that `<time-window-size value="0.01" />`, is there a good reason to accept preCICE returning something else than `0.01`?

fsimonis:
@BenjaminRodenberg I am not happy with the implementation part of this PR, however, the tests may be interesting to define expected behaviour.

@uekerman could you comment on what you would expect preCICE to return in these tests?

uekerman (commented):
Couldn't we do a test that does more timesteps and where we collect an error significantly larger than machine precision?
Currently `boost::unit_test::tolerance(1e-18)` looks odd to me.

tests/time-handling/SimpleMaxTime.xml
BenjaminRodenberg, line 1: Does this test work with the current version of develop?

* If yes: Why not directly merge this test into develop? This would indeed render this PR a pure refactoring.
* If no: This PR is more than a refactoring, if this test actually describes a valid use case.

So one question we should answer: Is this a relevant test case?

fsimonis:
When I change the test to run time-windows of `0.01` until t=100, then preCICE @ develop runs into an assertion at `t=64.02`.

```
ASSERTION FAILED
Location:   void precice::impl::ParticipantImpl::handleDataAfterAdvance(bool, bool, double, double, const cplscheme::ImplicitData &)
File:       /home/simonis/Development/precice/third/src/precice/impl/ParticipantImpl.cpp:429
Expression: math::greaterEquals(timeAfterAdvance, timeSteppedTo)
Rank:       0
Arguments:
  0: "We must have stayed or moved forwards in time (min-time-step-size)." == We must have ed or moved forwards in time (min-time-step-size).
  1: timeAfterAdvance == 64.02
  2: timeSteppedTo    == 64.02000000000001

```

fsimonis:
Superseeded by #1954

PR #1942, opened 2024-01-30
Skip zero samples when mapping initial data

fsimonis:
## Main changes of this PR

This PR adds the functionality to skip samples containing only 0 when mapping initial data.

This prevents possibly expensive data mappings of samples where the user provided no data / a zero initial condition.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/impl/DataContext.cpp
+++ b/src/precice/impl/DataContext.cpp
@@ -87,7 +87,7 @@ bool DataContext::hasMapping() const
   return hasReadMapping() || hasWriteMapping();
 }
 
-int DataContext::mapData(std::optional<double> after)
+int DataContext::mapData(std::optional<double> after, bool skipZero)
 {
   PRECICE_TRACE(getMeshName(), getDataName());
   PRECICE_ASSERT(hasMapping());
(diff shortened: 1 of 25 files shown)


src/precice/impl/DataContext.cpp
```diff
@@ -122,21 +122,25 @@ int DataContext::mapData(std::optional<double> after)
...
+          mapping.map(stample.sample, outSample.values);
+        }
+        PRECICE_DEBUG("Mapped values (t={}) = {}", stample.timestamp, utils::previewRange(3, outSample.values));
       }
 
-      PRECICE_DEBUG("Mapped values (t={}) = {}", stample.timestamp, utils::previewRange(3, outSample.values));
-
       // Store data from mapping buffer in storage
       context.toData->setSampleAtTime(stample.timestamp, std::move(outSample));
       ++executedMappings;
```
fsimonis, line 146: This still needs to be moved into the if

davidscn (commented):
ASTE turbine tutorial (where this showed up) works. Changes look clean and sane. Having a test to check ensure this behavior seems useful (and rather trivial in the context of map-if-necessary) though.

src/precice/impl/DataContext.cpp
fsimonis: Done

fsimonis:
I extended the `map-if-necessary` tests to ensure non-initialized data is skipped.

I also added tests for non-zero and zero initialized data mappings for both Serial and Parallel cases as well as Read and Write mappings.

tests/serial/map-initial-data/zero-data/SerialWrite.xml
```diff
@@ -0,0 +1,37 @@
...
+    <read-data name="Data" mesh="MeshTwo" />
+  </participant>
+
+  <m2n:sockets acceptor="One" connector="Two" />
+
+  <coupling-scheme:serial-explicit>
+    <participants first="One" second="Two" />
+    <max-time-windows value="1" />
+    <time-window-size value="1.0" />
+    <exchange data="Data" mesh="MeshTwo" from="One" to="Two" initialize="true" substeps="false" />
```
davidscn, line 35: Could it be valuable to configure at least one of the `zero-data` cases using `initialize=false`. Should have the same effect, right?

davidscn:
Other than that looks sane and also very nice testing coverage :+1:

tests/serial/map-initial-data/zero-data/SerialWrite.xml
fsimonis: These are covered by the `map-if-necessary` cases, which always use `initialize=false` and never write data, thus producing only 0 data samples.

PR #1943, opened 2024-02-01
Add Ubuntu 24.04LTS to CI

fsimonis:
## Main changes of this PR

This PR adds the prerelease image of Ubuntu 24.04 LTS to the CI.

The point of this PR is to test if this is even working at all before adding it to the test suite on develop.

Let's start testing early rather than late.

## Author's checklist


fsimonis:
Needs some cleanup, but apart from that ready to go.

Issue #1919, opened 2023-12-12
Simplify Mesh usage


fsimonis:
**Please describe the problem you are trying to solve.**

Using meshes and its vertices can be very cumbersome. There are many indirections necessary for the most simple tasks.

This is especially bad in `Mapping`s, where `input()` and `output()` return pointers.

**Describe the solution you propose.**

Simplify repeated use-cases:

## Size of a mesh

Currently: `mesh.vertices().size()`
Proposed: `mesh.nvertices()` or `mesh.size()`

## Access a vertex

Currently: `mesh.vertices()[index]`
Proposed: `mesh.vertex(index)`
Benefit: we can use checked access `.at(index)` in debug builds and unchecked access `[index]` in release builds.

## Access a vertex coordinate

Currently: `mesh.vertices()[index].getRawCoords()[coord]`
Proposed: `mesh.vertices()[index].coord(coord)` (with above `mesh.vertex(index).coord(coord)`)
Benefit: We loose efficiency due to calling the allocating `getCoords()` instead of `getRawCoords()` and we can check the index using dims.

## Access an input and output mesh in Mapping

Currently: `mesh::PtrMesh input() const` returns an owning pointer, which may be null.
Proposed: `mesh::Mesh& input() const` asserts that a mesh is set and returns a reference.
Benefit: Easier to use in `Mapping` implementations.

**Additional context**

These points are popping up every time we handle Mappings.
This should also simplify exporters.

MakisH:
> ## Size of a mesh
> 
> Currently: `mesh.vertices().size()` Proposed: `mesh.nvertices()` or `mesh.size()`

I find the `mesh.nvertices()` easier to understand and more difficult to misuse than `mesh.size()`. The latter does not communicate that the mesh contains vertices (it could be serialized, and `.size()` could be dimensions * nvertices). With `nvertices()`, I don't need to check the documentation when reading code (but I do need to check it when writing code).

> ## Access a vertex
> 
> Currently: `mesh.vertices()[index]` Proposed: `mesh.vertex(index)` Benefit: we can use checked access `.at(index)` in debug builds and unchecked access `[index]` in release builds.

I like this.

> ## Access a vertex coordinate
> 
[...]

PR #1946, opened 2024-02-12
Refactor Mesh and Vertex usage

fsimonis:
## Main changes of this PR

This PR adds
* `Mesh::empty()` to replace `mesh.vertices().empty()`
* `Mesh::nVertices()` to replace `mesh.vertices().size()`
* `Mesh::vertex(vid)` to replace `mesh.vertices()[vid]` and `mesh.vertices().at(vid)`
* `Vertex::coord(i)` to replace `vertex.getCoords()[i]` (which is allocating) and `vertex.rawCoords()[i]` as long as it doesn't require the 3rd coord of a 2D vertex.

Now `mesh.vertices()` is only required to iterate over the vertices.

This removes a lot of boilerplate from Mappings and makes the code more readable in many places, especially tests.

The pretty common `mesh->vertices().at(i).rawCoords()[j]` now becomes `mesh->vertex(i).coord(j)`.

Addresses most of #1919

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mesh/Mesh.hpp
+++ b/src/mesh/Mesh.hpp
@@ -69,11 +69,20 @@ public:
       int         dimensions,
       MeshID      id);
 
+  /// Mutable access to a vertex by VertexID
+  Vertex &vertex(VertexID id);
+
+  /// Const access to a vertex by VertexID
+  const Vertex &vertex(VertexID id) const;
+
   /// Returns modifieable container holding all vertices.
   VertexContainer &vertices();
 
   /// Returns const container holding all vertices.
   const VertexContainer &vertices() const;
+
+  /// Returns the number of vertices
+  std::size_t nVertices() const;
 
   /// Returns modifiable container holding all edges.
   EdgeContainer &edges();
--- a/src/mesh/Mesh.hpp
+++ b/src/mesh/Mesh.hpp
@@ -84,6 +84,12 @@ public:
   /// Returns the number of vertices
   std::size_t nVertices() const;
 
+  /// Does the mesh contain any vertices?
+  bool empty() const
+  {
+    return _vertices.empty();
+  }
+
   /// Returns modifiable container holding all edges.
   EdgeContainer &edges();
 
--- a/src/mesh/Vertex.hpp
+++ b/src/mesh/Vertex.hpp
@@ -53,6 +53,9 @@ public:
 
   void tag();
 
+  /// Returns a coordinate of a vertex
+  inline double coord(int index) const;
+
   inline bool operator==(const Vertex &rhs) const;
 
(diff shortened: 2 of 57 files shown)


src/mesh/Vertex.hpp
```diff
@@ -53,6 +53,9 @@ class Vertex {
 
   void tag();
 
+  /// Returns a coordinate of a vertex
+  inline double coord(int index) const;
```
MakisH, line 57: Do we want to allow for negatives? Would an `std::size_t` fit better?
Also in further places (since these are new methods).
It probably does not matter at all in this context.

I guess a strong type (e.g., `CoordID`) would just be painful without reason.

src/mesh/Mesh.hpp
```diff
@@ -69,12 +69,27 @@ class Mesh {
       int         dimensions,
       MeshID      id);
 
+  /// Mutable access to a vertex by VertexID
+  Vertex &vertex(VertexID id);
+
+  /// Const access to a vertex by VertexID
+  const Vertex &vertex(VertexID id) const;
```
MakisH, line 76: Does it make sense to delete the `operator[]`, to enforce that people use the checked variant?

MakisH (approved):
The API changes are beautiful! I wrote more details in the issue: https://github.com/precice/precice/issues/1919#issuecomment-1973287528

I did not spot any issues with the replacements.

Just brainstorming on two minor points, but ready to merge if you prefer.

src/mesh/Mesh.hpp
fsimonis: The operator[] is part of the std::deque so deleting is only an option if we want to hide vertices() somehow. Possible, but that's a bigger refactoring.

src/mesh/Vertex.hpp
fsimonis: This should actually be VertexID instead of bare int.

This is a pretty general issue we have in precice. VertexID is an int, Eigen uses int by default, stl styles use size_t and we normally use int for loops.

Moving everything to size_t may be a solution, but I am afraid that would be a v4 topic.

fsimonis: This is essentially `0`, `1`, or `2` for 3D.

`std::size_t` is a bit overkill for something that small. `unsigned char` would probably be the right choice here. Let's solve this in another PR though.

fsimonis: Opened an issue https://github.com/precice/precice/issues/1970

src/mesh/Mesh.hpp
fsimonis: Opened an issue https://github.com/precice/precice/issues/1971

PR #1949, opened 2024-02-14
Restructure docker releases

### LINKED ISSUE #1947: Docker release images are inconsistent
**Describe your setup**

preCICE Version: general

**Describe the problem**

The docker releases of preCICE are inconsistent:

* Version tags `:X.Y.Z` are based on `ubuntu:XX` and install the debian package attached to the release. [Dockerfile](https://github.com/precice/precice/blob/develop/tools/releasing/packaging/docker/Dockerfile
)
* Developer version `:develop` is based on the [`ci-ubuntu-XX`](https://github.com/precice/ci-images/blob/master/ci-ubuntu-2204.dockerfile
) image, which installs plenty more tools and also the additional user `precice` for running tests. The image builds and installs the `develop` branch of preCICE. [Dockerfile](https://github.com/precice/precice/blob/develop/tools/releasing/packaging/develop/Dockerfile)

These mixed tags are confusing and error-prone.

**Expected behaviour**

Images should be consistent.

**Proposed solution**

Make `precice/precice:develop` consistent with the `precice/precice:3.0.0` release tags.
Build and install preCICE based on the `ubuntu:XX` base image.

### LINKED ISSUE #1948: Rename nightly release image from develop to nightly

The current name of the nightly release image of preCICE is called `precice/precice:develop`.

This naming is problematic as it suggests that

* it may be a debug build
* it may include additional developer tools
* it may only be targetted at developers of preCICE

We use common terminology for the release. It is a nightly release, so we should just call it like that:

`precice/precice:nightly`

**Additional context**

Mentioned here https://github.com/precice/python-bindings/issues/191#issuecomment-1942379222

fsimonis:
## Main changes of this PR

This PR 
- changes develop dockerimage to base on blank ubuntu:22.04 instead of `precice/ci-ubuntu-2004`
- restructures the dockerimages to only reside in `tools/releasing/docker/`
- renames `precice/precice:develop` to `precice/precice:nightly`
- fixes job dependency in release pipeline

Both dockerfiles build locally.

Closes #1947
Closes #1948
Related to discussion in https://github.com/precice/python-bindings/issues/191

## Author's checklist


.github/workflows/release.yml
```diff
@@ -61,8 +61,8 @@ jobs:
           tag: ${{ github.ref }}
 
   dockerfile:
```
BenjaminRodenberg, line 63: This looks a bit redundant to what's happening in `.github/workflows/request-docker-release.yml`, but I guess there is a good reason?

tools/releasing/packaging/docker/nightly.dockerfile
```diff
@@ -0,0 +1,20 @@
+# Dockerfile to build a ubuntu image containing the installed develop version of preCICE
+
+FROM ubuntu:22.04
```
BenjaminRodenberg, line 3: I like the fact that we now use a standard image instead of a customized one :+1:

BenjaminRodenberg (approved):
Looks good to me. I think `nightly` is much more clear than `develop`. The python bindings and fenics adapter should also be able to deal with this docker image. I cannot really see problems right now. If I'm missing something we will see it sooner or later and it's probably good, if we then try to fix it downstream.

IshaanDesai (approved):
*nightly* reflects what the state of the image is :+1:

tools/releasing/packaging/docker/nightly.dockerfile
fsimonis: > I think if we always stick to the baseline version of the preCICE release this is a simple rule to follow and also more clear from the outside perspective. Our baseline for preCICE v3 is ubuntu 22.04 (?).

We support the latest 2 Ubuntu LTS versions, meaning our baseline is 20.04 and in April we will switch to 22.04.

> When do we change ubuntu:22.04 to something newer?

Every 2 years with the LTS releases, aka when baselines shift.

> Is a non-LTS also reasonable for a nightly build?

I would say no as this would unnecessarily restrict the support of past releases.

We again have conflicting interests.

Using the baseline Ubuntu to build release images makes sense from a testing perspective as you make sure that the baseline works.
Use the latest LTS to build release images makes sense from a production perspective as you make sure that the newest packages are available.

I'll open an issue about this.

fsimonis: There you go https://github.com/precice/precice/issues/1950

.github/workflows/release.yml
fsimonis: Redundant yes. One could maybe use a workflow dispatch, but this currently works fine and keeps all the information in the same place.

Maybe we need to refactor at some point by creating custom actions.

tools/releasing/packaging/docker/nightly.dockerfile
fsimonis: I'll mark this as resolved, as this PR currently uses the same version we already used for the releases.

BenjaminRodenberg: Sounds good. Which ubuntu we use in the end is a side-track that we can resolve in #1950 let's focus on the main issue here.

.github/workflows/docker-nightly.yml
MakisH, line 1: Since we are restructuring, does it make sense to also make the filenames `docker-nightly.yml` and `request-docker-release.yml` more consistent to each other?

Example: `publish-docker-nightly.yml` and `publish-docker-release.yml`.

Is the `request-` only due to the manual triggering?

Can we even make the `nightly` or `latest` parameters for the same workflow?

MakisH: Then no need to merge.

PR #1954, opened 2024-02-21
Refactor time processing

fsimonis:
## Main changes of this PR

This PR refactors the time processing into a separate class.

* the `DoubleAggregator` is an implementation of a Kahan Babushka Neumaier Aggregator with a bunch of quality of life improvments over our boost variant.
* the `TimeHandler` tracks the current time window

## TODOs

* [x] compute `nextTimeStepSize` in the TimeHandler
* [x] handle `maxTime` in the TimeHandler
* [ ] Add tests for double aggregator 
* [x] Add tests for TimeHandler

This allows us to test the time handling of very big cases.
This is too expensive to be run as integration tests.

Example:
* `time-window-size=1e-7 max-time=1`
* `time-window-size=0.01 max-time=1000`

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -79,6 +79,12 @@ BaseCouplingScheme::BaseCouplingScheme(
     PRECICE_ASSERT((maxIterations == INFINITE_MAX_ITERATIONS) || (minIterations <= maxIterations),
                    "Minimal iteration limit has to be smaller equal compared to the maximal iteration limit.");
   }
+
+  if (math::equals(maxTime, UNDEFINED_MAX_TIME)) {
+    _time = impl::TimeHandler();
+  } else {
+    _time = impl::TimeHandler(maxTime);
+  }
 }
 
 bool BaseCouplingScheme::isImplicitCouplingScheme() const
@@ -376,20 +382,10 @@ void BaseCouplingScheme::secondExchange()
 
     // Update internal time tracking
     if (_isTimeWindowComplete) {
-      // We move to the next time window
-      if (math::equals(_time.windowProgress(), _timeWindowSize)) {
-        _time.completeTimeWindowExact(_timeWindowSize);
-      } else {
-        // This only happens when the final time window is truncated due
-        // time window size not being a divider of max-time.
-        _time.completeTimeWindow();
-        PRECICE_ASSERT(!math::equals(_maxTime, UNDEFINED_MAX_TIME));
-        PRECICE_ASSERT(math::equals(_maxTime, getTime()));
-      }
+      _time.completeTimeWindow(_timeWindowSize);
+    } else {
+      _time.resetProgress();
     }
-    // We move the _time to the start of the updated time window.
-    // This can be a "reset" in case of an iteration, or the start of the next time window.
-    _time.resetProgress();
     _timeWindowSize = _nextTimeWindowSize;
   }
 }
@@ -429,17 +425,17 @@ bool BaseCouplingScheme::addComputedTime(
   PRECICE_TRACE(timeToAdd, getTime());
   PRECICE_ASSERT(isCouplingOngoing(), "Invalid call of addComputedTime() after simulation end.");
 
-  // add time interval that has been computed in the solver to get the correct time remainder
-  _time.progressBy(timeToAdd);
-
   // Check validness
-  bool valid = math::greaterEquals(getNextTimeStepMaxSize(), 0.0);
+  bool valid = math::greaterEquals(getNextTimeStepMaxSize(), timeToAdd);
   PRECICE_CHECK(valid,
                 "The time step size given to preCICE in \"advance\" {} exceeds the maximum allowed time step size {} "
                 "in the remaining of this time window. "
                 "Did you restrict your time step size, \"dt = min(preciceDt, solverDt)\"? "
                 "For more information, consult the adapter example in the preCICE documentation.",
-                timeToAdd, getWindowStartTime() + _timeWindowSize - getTime() + timeToAdd);
+                timeToAdd, getNextTimeWindowSize());
+
+  // add time interval that has been computed in the solver to get the correct time remainder
+  _time.progressBy(timeToAdd);
 
   return reachedEndOfTimeWindow();
 }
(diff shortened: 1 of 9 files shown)


src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -205,14 +211,13 @@ void BaseCouplingScheme::receiveData(const m2n::PtrM2N &m2n, const DataMap &rece
...
   // set _time state to point to end of this window such that _time in receiveData is at end of window
-  _time(_nextTimeWindowSize);
+  _time.progressBy(_nextTimeWindowSize);
   // receive data for end of window
   this->receiveData(m2n, receiveData);
   // reset time state;
-  _time = KahanAccumulator{};
-  _time(oldTime);
+  _time = oldTime;
 }
```
BenjaminRodenberg, line 221: This part with `oldTime` still looks strange, but is necessary due to the way how `receiveData` works, if we set `substeps="False"`. Maybe put a comment here that `oldTime` is something temporary. An issue on this topic would be helpful such that I don't have to write this again in every PR review (I know: I should do this and can take care of it).

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -433,17 +425,17 @@ bool BaseCouplingScheme::addComputedTime(
...
   PRECICE_CHECK(valid,
                 "The time step size given to preCICE in \"advance\" {} exceeds the maximum allowed time step size {} "
                 "in the remaining of this time window. "
                 "Did you restrict your time step size, \"dt = min(preciceDt, solverDt)\"? "
                 "For more information, consult the adapter example in the preCICE documentation.",
-                timeToAdd, getWindowStartTime() + _timeWindowSize - getTime() + timeToAdd);
+                timeToAdd, getNextTimeWindowSize());
+
+  // add time interval that has been computed in the solver to get the correct time remainder
+  _time.progressBy(timeToAdd);
```
BenjaminRodenberg, line 438: This change looks like it is more than a refactoring, but it's definitely a step into the right direction. Are you sure that you might not accidentally introduce something fishy here?

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -504,27 +496,20 @@ double BaseCouplingScheme::getNextTimeStepMaxSize() const
...
 
 bool BaseCouplingScheme::isCouplingOngoing() const
 {
-  bool timeLeft      = math::greater(_maxTime, getTime()) || math::equals(_maxTime, UNDEFINED_MAX_TIME);
   bool timestepsLeft = (_maxTimeWindows >= _timeWindows) || (_maxTimeWindows == UNDEFINED_TIME_WINDOWS);
-  return timeLeft && timestepsLeft;
+  bool timeLeft      = !_time.reachedEnd();
+  PRECICE_WARN("tsl:{} tl:{} t:{:e} mt:{:e} umt:{:e} is0:{} isabs0:{}", timestepsLeft, timeLeft, _time.time(), _maxTime, _time.untilTime(_maxTime),
+               math::equals(_time.untilTime(_maxTime), 0.0),
+               math::equals(std::abs(_time.untilTime(_maxTime)), 0.0));
```
BenjaminRodenberg, line 511:
```suggestion
```
Debugging statement?

src/cplscheme/impl/TimeHandler.hpp
BenjaminRodenberg, line 1: Did not look into this class in detail, but I guess that's also the good thing about abstraction. I think we need some unit-tests for this class.

src/utils/DoubleAggregator.hpp
BenjaminRodenberg, line 1: Why not use the boost version here? I know: The user interface of the boost version is not the best one, but this is something we could hide inside `TimeHandler.hpp`. If we write our own class, there is always the risk of doing something wrong.

BenjaminRodenberg (changes requested):
Generally good approach. Simplifies the (very complicated) `BaseCouplingScheme` and should improve maintainability. We, however, still need some tests (maybe not a must right now).

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: I tried to summarize the big picture here: https://github.com/precice/precice/issues/1957
@fsimonis would you mind adding a comment with a pointer to https://github.com/precice/precice/issues/1957 in this PR? I hope the issue is somehow understandable. Unfortunately, the situation is confusing. But fixing it needs some effort with low benefit (from the short term perspective).

fsimonis: Done

src/cplscheme/BaseCouplingScheme.cpp
fsimonis: Definitely more than a refactoring.

Doesn't this make more sense though?
Check if the step size is permitted and then apply.

src/cplscheme/impl/TimeHandler.hpp
fsimonis: Added some

src/utils/DoubleAggregator.hpp
fsimonis: Actually, this is so nicely abstracted away that I quickly swapped over to the boost::accumulator.
It is actually more accurate than custom implementation, so I migrated to it.
Thanks for the hint!

fsimonis:
This PR pretty much ready to go but has grown out of a simple refactoring.

The tests that failed in https://github.com/precice/precice/pull/1939#issuecomment-1956588970 do succeed on this branch.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: Of course it makes more sense and is simpler to read! I just wanted to say that I do not really know all the implications this might have. But if this part is sufficiently tested: Good to go.

src/utils/DoubleAggregator.hpp
```diff
@@ -0,0 +1,72 @@
+#include <boost/accumulators/accumulators.hpp>
+#include <boost/accumulators/statistics.hpp>
+#include <boost/accumulators/statistics/sum_kahan.hpp>
+
+namespace precice::utils {
+
+/// An accurate aggregator for doubles with usability in mind
+class DoubleAggregator {
+private:
+  using Acc = boost::accumulators::accumulator_set<double, boost::accumulators::stats<boost::accumulators::tag::sum_kahan>>;
```
BenjaminRodenberg, line 10:
```suggestion
  /// This class is mainly a wrapper for boost's Kahan accumulator which implements the Kahan summation algorithm (see https://en.wikipedia.org/wiki/Kahan_summation_algorithm)
  using Acc = boost::accumulators::accumulator_set<double, boost::accumulators::stats<boost::accumulators::tag::sum_kahan>>;
```
I think it would be nice to also mention the keyword "wrapper" here and provide a brief link to what we actually do. Both are implementation details. Therefore, I did not put it into the class description above. Not sure that this is 100% the right approach. Maybe you have a better idea @fsimonis?

src/utils/DoubleAggregator.hpp
BenjaminRodenberg, line 1: Looks very nice now!

src/cplscheme/impl/TimeHandler.hpp
```diff
@@ -0,0 +1,172 @@
...
+  /// @name Time differences
+  /// @{
+
+  /// Returns the time distance to the possibly truncated end of the current time window
+  double untilWindowEnd(double timeWindowSize) const
+  {
+    double maxDt = untilProgress(timeWindowSize);
+    if (!_maxTime) {
+      return maxDt;
+    }
```
BenjaminRodenberg, line 109:
```suggestion
```
Is this needed? `untilEnd()` returns `std::numeric_limits<double>::max()` if `!_maxTime`.

src/cplscheme/impl/TimeHandler.hpp
```diff
@@ -0,0 +1,172 @@
...
+
+  /** Has the end of the overall time been reached?
+   * This is always false if no max time has been defined.
+   */
+  bool reachedEnd() const
+  {
+    if (!_maxTime) {
+      return false;
+    }
+    return math::equals(untilTime(*_maxTime), 0.0);
```
BenjaminRodenberg, line 96:
```suggestion
    return math::equals(untilEnd(), 0.0);
```

BenjaminRodenberg (changes requested):
I think there are some minor simplifications. But maybe I'm also just introducing bugs because I don't know it better. Might also be a good motivation for some simple additional tests, if the corner cases I'm touching here are not tested yet.

src/utils/DoubleAggregator.hpp
fsimonis: Good idea.

src/cplscheme/BaseCouplingScheme.cpp
fsimonis: The message is actually buggy. :see_no_evil:

src/cplscheme/impl/TimeHandler.hpp
fsimonis: This is on purpose and should be documented. The reason is two-fold:

1. I try to reduce computations and comparisons of time points as much as possible to avoid accuracy issues. Here we know that it will always be maxDt if there is no maxTime set.
2. This makes the handling of the maxTime easier to spot in the code and makes the flow easier to understand.

src/cplscheme/impl/TimeHandler.hpp
fsimonis: Same reasoning as above. Instead of risking any issues here, we return early with a clear message. (needs a comment too)

uekerman (approved):
Looks good to me. Nice job!

BenjaminRodenberg (approved):

Issue #1961, opened 2024-02-26
Baseline upgrade to Ubuntu 22.04 LTS


fsimonis:
**Baseline change**

With the release of Ubuntu 24.04, we lift the baseline version from Ubuntu 20.04 LTS to Ubuntu 22.04 LTS. Which brings a range of dependency changes.

I try to summarize impacting changes below.

**Dependency version changes** 

Dependency | Old | New 
--- | --- | ---
CMake | 2.16.3 | 3.22.1
GCC/stdlibc++ | 9.3.0 | 11.2.0
Clang | 10.0 | 14.0
Eigen | 3.3.7 | 3.4.0
Boost | 1.71.0 | 1.74.0
PETSc | 3.12 | 3.15
Python | 3.8.2  | 3.10.6
Numpy | 1.17.4 | 1.21.5

_Note: we already required Eigen 3.4.0 when enabling Ginkgo_

**Available C++ Version**

g++ 11.2 and libstdc++ 11.2 only partially supports C++20.
Despite some big features missing, the version bump may still be justified by the following features:

**High impact features**

* [source_location](https://en.cppreference.com/w/cpp/utility/source_location) allows us to replace log macros with function calls.
* [Designated initializers](https://en.cppreference.com/w/cpp/language/aggregate_initialization#Designated_initializers) should make many type constructions easier to read.
* [std::span](https://en.cppreference.com/w/cpp/container/span), so we can convert from `precice::span` to `std::span` in `ParticipantImpl`, just like we do with `string_view`.
* [using enum](https://en.cppreference.com/w/cpp/language/enum#Using-enum-declaration), very useful to make checks in configuration and mappings more readable.

**Medium impact features**

* consistent container erasure for [vector](https://en.cppreference.com/w/cpp/container/vector/erase2) and map, simplifies especially tagging and repartitioning.
* [concepts](https://en.cppreference.com/w/cpp/concepts) should make some custom algorithms more robust to use.
* [init statement for ranged for](https://en.cppreference.com/w/cpp/language/range-for#Syntax) simplifies CompositionalCouplingScheme 
* `std::map::contains` instead of `map.count() > 0`. Replaces `utils::contains`

**Low impact features**

* Attributes [likely and unlikely](https://en.cppreference.com/w/cpp/language/attributes/likely), could be used to annotate all errors paths as unlikely.
* [Three-way comparison](https://en.cppreference.com/w/cpp/language/operator_comparison#Three-way_comparison)
[...]

PR #1962, opened 2024-02-26
Upgrade baseline to Ubuntu 22.04 LTS

fsimonis:
## Main changes of this PR

**On hold until the release of Ubuntu 24.04 LTS**

This PR upgrades the baseline versions from Ubuntu 20.04 LTS to 22.04 LTS in CMake and the workflows.

Part of #1961

Untouched are C++ version and internal changes.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -28,7 +28,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        TARGET: ["ubuntu:20.04", "ubuntu:22.04", "debian:11", "debian:12"]
+        TARGET: ["ubuntu:22.04", "ubuntu:24.04", "debian:11", "debian:12"]
     steps:
       - uses: actions/checkout@v4
       - name: Generate build directory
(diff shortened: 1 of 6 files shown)


.github/workflows/release.yml
```diff
@@ -28,7 +28,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        TARGET: ["ubuntu:20.04", "ubuntu:22.04", "debian:11", "debian:12"]
+        TARGET: ["ubuntu:22.04", "ubuntu:24.04", "debian:11", "debian:12"]
```
fsimonis, line 31:
```suggestion
        TARGET: ["ubuntu:22.04", "ubuntu:24.04", "debian:12"]
```

We loose support for Debian 11 as its Eigen version is still 3.3.9 and we now require 3.4.0

MakisH (approved):
I cannot think of anything else to update in this repository.
The website is, of course, another story. Same for Spack.

PR #1964, opened 2024-02-27
Fix mapping type bug

uekerman:
## Main changes of this PR

Reproduces bug #1965 

## Author's checklist

## Reviewers' checklist


fsimonis (approved):
Straight-forward fix

PR #1969, opened 2024-02-29
Improve CI speed and portability

fsimonis:
## Main changes of this PR

This PR attempts to speed up the CI by using `mold` as a linker and improves performance portability by using ninja, using all cores by default.
We now make sure to always use all available resources, no matter what runner a jobs runs on.

With the addition of Ubuntu 22.04 the CI becomes even slower.
It's time to speed up with some low-hanging fruit.

## Author's checklist


MakisH:
Which part of the diff specifies to use `mold`? I don't see anything related at the moment.

fsimonis:
@MakisH I added it, compared it to other runs and removed it again.

I documented conclusions in the OP.

fsimonis:
I salvaged the lessons learned from this PR and opened #1973

Turns out that getting ccache to work is not trivial and Ninja destroys the 20.04 builds.

Issue #1926, opened 2024-01-10
Port ConfigParser to string_view


fsimonis:
**Please describe the problem you are trying to solve.**

The ConfigParser bridges the call-back-based C-library libxml2 to the C++ XMLTag and XMLAttribute of preCICE.

This requires a lot of non-owning C-string to `std::string` transformations.
Performance aside, this is difficult to read and code analysers complain.

**Describe the solution you propose.**

Using `std::string_view` for everything non-owning would be way easier on the eye.
Less casting, less noise, fewer copies.

**Describe alternatives you've considered**
Ignore the linters

Durganshu:
I'm picking up on this issue.

fsimonis:
@Durganshu be aware that this is not simple replace job. It's about figuring out which components own data and which ones only read it.

This is especially important on the interface between libxml2 and the xml::ConfigParser::CTag generation.

Durganshu:
Thanks for the heads up!

fsimonis:
This has been largely done in #1978

PR #1978, opened 2024-03-07
Porting ConfigParser to use std::string_view wherever possible

Durganshu:
## Main changes of this PR

The ConfigParser bridges the call-back-based C-library libxml2 to the C++ XMLTag and XMLAttribute of preCICE. This PR ports ConfigParser to use std::string_view instead of std::string where the components don't own data. This leads to less casting, fewer copies, and potentially better performance.

See #1926.
## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/xml/ConfigParser.cpp
+++ b/src/xml/ConfigParser.cpp
@@ -64,7 +64,7 @@ void OnStartElementNs(
 
   auto pParser = static_cast<ConfigParser *>(ctx);
 
-  std::string sPrefix(prefix == nullptr ? "" : reinterpret_cast<const char *>(prefix));
+  std::string_view sPrefix(prefix == nullptr ? "" : reinterpret_cast<const char *>(prefix));
 
   pParser->OnStartElement(reinterpret_cast<const char *>(localname), sPrefix, attributesMap);
 }
@@ -210,7 +210,7 @@ auto gatherCandidates(const std::vector<std::shared_ptr<XMLTag>> &DefTags, std::
 
 void ConfigParser::connectTags(const ConfigurationContext &context, std::vector<std::shared_ptr<XMLTag>> &DefTags, CTagPtrVec &SubTags)
 {
-  std::unordered_set<std::string> usedTags;
+  std::unordered_set<std::string_view> usedTags;
 
   for (auto &subtag : SubTags) {
     std::string expectedName = (subtag->m_Prefix.length() ? subtag->m_Prefix + ":" : "") + subtag->m_Name;
@@ -259,8 +259,8 @@ void ConfigParser::connectTags(const ConfigurationContext &context, std::vector<
 }
 
 void ConfigParser::OnStartElement(
-    std::string         localname,
-    std::string         prefix,
+    std::string_view    localname,
+    std::string_view    prefix,
     CTag::AttributePair attributes)
 {
   auto pTag = std::make_shared<CTag>();
--- a/src/xml/ConfigParser.cpp
+++ b/src/xml/ConfigParser.cpp
@@ -228,7 +228,7 @@ void ConfigParser::connectTags(const ConfigurationContext &context, std::vector<
       auto matches = utils::computeMatches(expectedName, names);
       if (!matches.empty() && matches.front().distance < 3) {
         matches.erase(std::remove_if(matches.begin(), matches.end(), [](auto &m) { return m.distance > 2; }), matches.end());
-        std::vector<std::string> stringMatches;
+        std::vector<std::string_view> stringMatches;
         std::transform(matches.begin(), matches.end(), std::back_inserter(stringMatches), [](auto &m) { return m.name; });
         PRECICE_ERROR("The configuration contains an unknown tag <{}>. Did you mean <{}>?", expectedName, fmt::join(stringMatches, ">,<"));
       } else {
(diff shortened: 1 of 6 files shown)


src/xml/ConfigParser.cpp
```diff
@@ -59,14 +59,14 @@ void OnStartElementNs(
     auto        valueEnd   = reinterpret_cast<const char *>(attributes[index + 4]);
     std::string value(valueBegin, valueEnd);
 
-    attributesMap[attributeName] = decodeXML(value);
+    attributesMap[attributeName] = decodeXML(std::move(value));
```
fsimonis, line 62: This is a strong indicator that `decodeXML` should take a `string_view` and return a decoded owning `std::string`.

src/xml/ConfigParser.cpp
```diff
@@ -59,14 +59,14 @@ void OnStartElementNs(
...
   }
 
   auto pParser = static_cast<ConfigParser *>(ctx);
 
   std::string sPrefix(prefix == nullptr ? "" : reinterpret_cast<const char *>(prefix));
```
fsimonis, line 67: `sPrefix` is either empty or the prefix. This can also be a `string_view`.

src/xml/ConfigParser.cpp
```diff
@@ -59,14 +59,14 @@ void OnStartElementNs(
...
 
-  pParser->OnStartElement(reinterpret_cast<const char *>(localname), sPrefix, attributesMap);
+  pParser->OnStartElement(std::move(reinterpret_cast<const char *>(localname)), std::move(sPrefix), attributesMap);
```
fsimonis, line 69: Both can be non-owning, so we change OnStartElement to take two string_views

src/xml/ConfigParser.cpp
```diff
@@ -117,10 +117,10 @@ void OnFatalErrorFunc(void *userData, const char *error, ...)
 
 precice::logging::Logger ConfigParser::_log("xml::XMLParser");
 
-ConfigParser::ConfigParser(const std::string &filePath, const ConfigurationContext &context, std::shared_ptr<precice::xml::XMLTag> pXmlTag)
+ConfigParser::ConfigParser(std::string_view filePath, const ConfigurationContext &context, std::shared_ptr<precice::xml::XMLTag> pXmlTag)
     : m_pXmlTag(std::move(pXmlTag))
 {
-  readXmlFile(filePath);
+  readXmlFile(std::string(filePath));
```
fsimonis, line 123: Why does readXmlFile need an owning string in the interface?

Same below

src/xml/ConfigParser.cpp
```diff
@@ -210,7 +210,7 @@ auto gatherCandidates(const std::vector<std::shared_ptr<XMLTag>> &DefTags, std::
 
 void ConfigParser::connectTags(const ConfigurationContext &context, std::vector<std::shared_ptr<XMLTag>> &DefTags, CTagPtrVec &SubTags)
 {
-  std::unordered_set<std::string> usedTags;
+  std::unordered_set<std::string_view> usedTags;
```
fsimonis, line 213: This works only under the assumption that the tags don't change. Let's keep this `std::string` to make this more robust to future changes.

src/xml/Printer.cpp
```diff
@@ -317,7 +317,7 @@ std::ostream &printMD(std::ostream &out, const XMLTag &tag, int level, std::map<
         out << "* [" << heading << "](" << link << ") `" << subtag->getOccurrenceString(subtag->getOccurrence()) << "`\n";
       } else {
         auto &tags = groupedTags[ns];
-        tags.emplace_back("[" + subtag->getName() + "](" + link + ") `" + subtag->getOccurrenceString(subtag->getOccurrence()) + "`");
+        tags.emplace_back("[" + subtag->getName() + "](" + link + ") `" + std::string(subtag->getOccurrenceString(subtag->getOccurrence())) + "`");
```
fsimonis, line 320: Way cleaner alternative would be to using `fmt::format` here.

fsimonis (commented):
Hi there,

I did a first iteration on your PR and already learned some things. :+1:

src/xml/ConfigParser.cpp
Durganshu: Yes, your point is reasonable. 

The reason I was doing an explicit call at the interface was to make it compatible with std::ifstream here. 

https://github.com/precice/precice/blob/a8d2b233c640bdd86f67f036313bed5aeda4038e/src/xml/ConfigParser.cpp#L172

But it is not a clean way of passing it to the interface like this.

src/xml/ConfigParser.cpp
```diff
@@ -169,7 +171,7 @@ int ConfigParser::readXmlFile(std::string const &filePath)
   SAXHandler.error          = OnErrorFunc;
   SAXHandler.fatalError     = OnFatalErrorFunc;
 
-  std::ifstream ifs(filePath);
+  std::ifstream ifs{std::string(filePath).c_str()};
```
fsimonis, line 174: I don't think this is worth it. Let's change the `filePath` back to `std::string`.

The `ifstream` accepts `std::string`, as it guarantees to be zero-terminated, which the system C-APIs require. `string_view` doesn't fullfill this guarantee.

src/xml/Printer.cpp
```diff
@@ -314,10 +315,16 @@ std::ostream &printMD(std::ostream &out, const XMLTag &tag, int level, std::map<
 
       const auto ns = subtag->getNamespace();
       if (ns.empty()) {
-        out << "* [" << heading << "](" << link << ") `" << subtag->getOccurrenceString(subtag->getOccurrence()) << "`\n";
+        out << fmt::format("[* {}]({}) `{}`",
+                           heading,
+                           link,
+                           std::string(subtag->getOccurrenceString(subtag->getOccurrence())));
```
fsimonis, line 321: This should work without conversion.
```suggestion
                           subtag->getOccurrenceString(subtag->getOccurrence()));
```

fsimonis (approved):
This looks ready now. Thanks for the contribution!

PR #1979, opened 2024-03-08
Simplify QN update

BenjaminRodenberg:
## Main changes of this PR

Simplify code.

Improve readability and prepare for https://github.com/precice/precice/pull/1834.

## Author's checklist

* [x] ~~I added a changelog file with `make changelog` if there are user-observable changes since the last release.~~
* [x] ~~I added a test to cover the proposed changes in our test suite.~~
* [x] ~~For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).~~

## Reviewers' checklist

### DIFF
--- a/src/acceleration/BaseQNAcceleration.cpp
+++ b/src/acceleration/BaseQNAcceleration.cpp
@@ -371,7 +371,7 @@ void BaseQNAcceleration::performAcceleration(
     /**
      * apply quasiNewton update
      */
-    _values = _oldValues + xUpdate + _residuals; // = x^k + delta_x + r^k - q^k
+    _values += xUpdate;
 
     // pending deletion: delete old V, W matrices if timeWindowsReused = 0
     // those were only needed for the first iteration (instead of underrelax.)


BenjaminRodenberg:
Usually, this small change would not require a PR, but I think it might be a first step into the direction of a bigger refactoring. It also potentially interferes with #1834 and the work of @NiklasKotarsky. Some goals I have in mind are:

1. Simplify QN in general. Don't modify `_values` inside QN update.
2. Use a more explicit interface for the whole QN/Acceleration class hierarchy. Not `Acceleration::performAcceleration(DataMap &cplData)` but `Acceleration::accelerate(const DataMap &cplData, Eigen::VectorXd &acceleratedData)` (compare to https://github.com/precice/precice/pull/1660)

I don't think we have to solve everything in a single PR and the interface I'm sketching in 2. is probably also not the one we want in the end, but it should help to understand the idea.

uekerman (approved):
Seems correct, but maybe some history-digging on why we had it like this could be good. Could be that this was due to the "q^k" (manifold mapping), which we removed eventually.

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -371,7 +371,7 @@ void BaseQNAcceleration::performAcceleration(
     /**
      * apply quasiNewton update
      */
-    _values = _oldValues + xUpdate + _residuals; // = x^k + delta_x + r^k - q^k
```
NiklasKotarsky, line 374: These lines are mathematically and programming wise equivalent as far I can tell. I checked the pullrequest #703 where the manifold-mapping was removed and this update formula seems to come from there. As far as I can understand manifold-mapping feature was some sort of "multigrid method" with a fine and a coarse model. The q^k referenced in the comment was used to get the correct optimization target for the coarse model. 

I would be in favor for chainging this, since "_values += xUpdate;" is a lot closer to what I am doing in #1834 and would allow a easier comparison between rQNWR and preCICE current QN acceleration.

NiklasKotarsky (approved):

PR #1999, opened 2024-04-11
Fix skipping initial mappings in parallel

### LINKED ISSUE #1998: Using `initialize="true"` leads to inconsistent decision for skipping zero sample
**Describe your setup**

preCICE Version: 3.1.0

**Describe the problem**
My testcase (doesn't depend on the number of participants) requires to use `<exchange... intialize="true"/>`. I run in parallel and not all participants are part of the coupling interface. In our current implementation, we call `mapData` with `skipZero=true` and use a logical `&&` to connect it to data being zero or not

https://github.com/precice/precice/blob/4c3f798231b131583106b882b36af175b6b99538/src/precice/impl/DataContext.cpp#L129

The other participant nicely provides initial data and the condition `stample.sample.values.isZero()` evaluates to `false` for all participants, who are part of the coupling interface. However, for all participants not part of the interface,  `stample.sample.values.size() == 0`, and therefore `stample.sample.values.isZero()` evaluates to `true`.

This inconsistencies leads to decisions on individual ranks to enter `mapData` or skip it. However, we have mappings relying on collective MPI operations, such as the Eigen RBF or (didn't check, but should apply) our PETSc based mappings, such that we run into a deadlock for these cases.

The decision to skip the zero sample or not should be consistent across ranks. Not all mapping variants rely on collective MPI operations, but from a logical perspective, it would make sense to synchronize the decision (in my case, the logging on the primary rank was also wrong, as it didn't hold any interface fraction).

**Step To Reproduce**
1. Setup a participant and run it at least on two ranks. One of the ranks should not be part of the interface. Configure `initialize="true"` and initialize non-zero data.
2. Configure one of our global RBF mapping variants.
3. Run into a deadlock.

davidscn:
## Main changes of this PR

Fixes the decision for skipping initial data containing only zero data in parallel runs by making the decision consistent across ranks. 

Provided in #1998 and resolves #1998

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/impl/DataContext.cpp
+++ b/src/precice/impl/DataContext.cpp
@@ -4,7 +4,7 @@
 
 #include "precice/impl/DataContext.hpp"
 #include "utils/EigenHelperFunctions.hpp"
-
+#include "utils/IntraComm.hpp"
 namespace precice::impl {
 
 logging::Logger DataContext::_log{"impl::DataContext"};
@@ -126,7 +126,7 @@ int DataContext::mapData(std::optional<double> after, bool skipZero)
           dataDims,
           Eigen::VectorXd::Zero(dataDims * mapping.getOutputMesh()->nVertices())};
 
-      bool skipMapping = skipZero && stample.sample.values.isZero();
+      bool skipMapping = skipZero && (utils::IntraComm::l2norm(stample.sample.values) < math::NUMERICAL_ZERO_DIFFERENCE);
 
       PRECICE_INFO("Mapping \"{}\" for t={} from \"{}\" to \"{}\"{}",
                    getDataName(), stample.timestamp, mapping.getInputMesh()->getName(), mapping.getOutputMesh()->getName(),
--- a/src/precice/impl/DataContext.cpp
+++ b/src/precice/impl/DataContext.cpp
@@ -126,6 +126,7 @@ int DataContext::mapData(std::optional<double> after, bool skipZero)
           dataDims,
           Eigen::VectorXd::Zero(dataDims * mapping.getOutputMesh()->nVertices())};
 
+      // Note that the l2norm is only computed during initialization due to short-circuit evaluation in C++
       bool skipMapping = skipZero && (utils::IntraComm::l2norm(stample.sample.values) < math::NUMERICAL_ZERO_DIFFERENCE);
 
       PRECICE_INFO("Mapping \"{}\" for t={} from \"{}\" to \"{}\"{}",
--- a/docs/changelog/1999.md
+++ b/docs/changelog/1999.md
@@ -1,0 +1,1 @@
+- Fixed skipping initial mapping when data contains only zeros in parallel.
--- a/tests/parallel/map-initial-data/NonZeroData.cpp
+++ b/tests/parallel/map-initial-data/NonZeroData.cpp
@@ -1,0 +1,20 @@
+#ifndef PRECICE_NO_MPI
+
+#include "helper.hpp"
+#include "testing/Testing.hpp"
+
+BOOST_AUTO_TEST_SUITE(Integration)
+BOOST_AUTO_TEST_SUITE(Parallel)
+BOOST_AUTO_TEST_SUITE(MapInitialData)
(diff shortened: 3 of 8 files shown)


src/precice/impl/DataContext.cpp
```diff
@@ -126,7 +126,7 @@ int DataContext::mapData(std::optional<double> after, bool skipZero)
           dataDims,
           Eigen::VectorXd::Zero(dataDims * mapping.getOutputMesh()->nVertices())};
 
-      bool skipMapping = skipZero && stample.sample.values.isZero();
+      bool skipMapping = skipZero && (utils::IntraComm::l2norm(stample.sample.values) < math::NUMERICAL_ZERO_DIFFERENCE);
```
fsimonis, line 129: This is now an unconditional collective for every mapped data, which sounds pretty expensive to me.

An elegant and easy way to prevent this could be:
```suggestion
      bool skipMapping = skipZero && mapping.usesCollectives() && (utils::IntraComm::l2norm(stample.sample.values) < math::NUMERICAL_ZERO_DIFFERENCE);
```

This prevents the `l2norm` collective when `mapping.usesCollectives() == false`.

This could be as simple as
```cpp
// in Mapping.hpp
virtual bool usesCollectives() const = 0;
// for RBF
bool usesCollectives() const override { return true };
// for the NN, NNG and NP (as long as scaled-consistent is locked to serial participants)
bool usesCollectives() const override { return false };
```

davidscn: > This is now an unconditional collective for every mapped data

Actually no, according to the C++ standard, the logical AND (&&) and logical OR (||) operators perform short-circuit evaluation. The boolean is only `true` in the 'initial' skip. In all other cases, the boolean is `false` and the `l2norm` is not evaluated.

uekerman: @fsimonis But your solution no longer checks whether the data is actually zero in case of non-collective mappings?

fsimonis: > short-circuit evaluation

:man_facepalming: _true_

So currently, we have a collective for every data-sample when we initialize data.
While not tragic, that still hurts performance for NN/NP etc.

> But your solution no longer checks whether the data is actually zero in case of non-collective mappings?

This should do the trick:

```suggestion
      bool skipMapping = skipZero && ( mapping.usesCollective() ? (utils::IntraComm::l2norm(stample.sample.values) < math::NUMERICAL_ZERO_DIFFERENCE) : stample.sample.values.isZero());
```

davidscn: > While not tragic, that still hurts performance for NN/NP

Let's not worry about this at this point and introduce complexity. It's an initialization cost and for really large cases, establishing the connection and computing the communication maps takes significantly more time (I measured this back then on Supermuc for my mapping experiments). If we observe issues in the future, we can still fix it in the future.

fsimonis: Low gain at a low added complexity. I don't have a strong opinion here.
@uekerman your call.

davidscn: > Low gain at a low added complexity. I don't have a strong opinion here.

Well, your logic is incomplete. The PUM doesn't use collectives, whereas the other RBFs do. Also, the code changes might appear trivial, but introduce a dependency of the mapping package to the data context. Modifying one of the mapping to incorporate MPI calls would require changes here as well. If we add GPU to the PUM, we might need such collective MPI calls, how do we handle this then? So we would need separate tests for all cases, which is then yet another todo.

davidscn: I will add a comment to make the short-circuit evaluation more apparent.

davidscn, line 129:
```suggestion
      // Note that the l2norm is only computed during initialization due to short-circuit evaluation in C++
      bool skipMapping = skipZero && (utils::IntraComm::l2norm(stample.sample.values) < math::NUMERICAL_ZERO_DIFFERENCE);
```

fsimonis (approved):
Looks good!
Please squash and merge.

Issue #1996, opened 2024-04-09
Clean up secondary data in quasi-Newton acceleration


uekerman:
In quasi-Newton acceleration, we distinguish primary data (data that is used to compute the acceleration and accelerated) and secondary data (data that is *only* accelerated). For IQN-ILS, we store W matrices for each secondary data and use the coefficient vector `c` from the primary data:

https://github.com/precice/precice/blob/efd7fb6703c117a518501e4192e9d26823c11ed2/src/acceleration/IQNILSAcceleration.cpp#L209-L219

**What should we do with secondary data in IQN-IMVJ?** Seems like we silently don't do anything. We underrelax in the first iteration, but we don't do anything in `IQNIMVJAcceleration::computeQNUpdate`. Looking at the math, I am also not sure what we could do, but I did not yet think a lot about it.

## Two options

1. Find a proper treatment of secondary data in IQN-IMVJ and implement it.
2. Forbid (implicit) configuration of secondary data for IQN-IMVJ, clean up (e.g. `IQNIMVJAcceleration::computeUnderrelaxationSecondaryData`), and document.

In both cases, we need to improve test coverage. There are currently no tests for secondary data at all.

## Context

I stumbled over this while reviewing #1834. We are currently anyway improving documentation concerning secondary data in https://github.com/precice/precice.github.io/pull/366.

Fujikawas:
A possible solution according to Miriam: 
Given that the current IMVJ is 
$$x^p_{k+1} = \tilde{x}\_k^p + [ (J^p_{\text{prev}})^{-1} + \underbrace{(W^p_k - (J^p_{\text{prev}})^{-1} V_k)}\_{=: \tilde{W}^p\_{k}} V\_k^\dagger ] (-r\_k)$$ 
with  
$V_k = (\Delta r_i)\_{i=1,2,\ldots}$: Matrix with residual differences (in primary data)
$W^p_k = (\Delta \tilde{x}^p_i)_{i=1,2,\ldots}$: Matrix with differences in primary data
$V_k^\dagger$ from solving $V_k^\dagger \cdot Q \cdot R = I \mbox{ with } V_k = Q \cdot R$

We could do the same to secondary data separately:
[...]

PR #2005, opened 2024-04-18
Extend QN acceleration to waveform iterations

NiklasKotarsky:
## Main changes of this PR

Adds support for Quasi Newton acceleration for waveform iterations.

Implements the algorithm described https://onlinelibrary.wiley.com/doi/10.1002/nme.6443 for fixed time grids. For moving time grids everything will be interpolated to the first time grid from the first time window inside the QNWR class instead of using the waveforms inside the QN method as done in #1834. 

For IMVJ variant the same fix for all time steps as described in issue  #1996 will be used. 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/acceleration/BaseQNAcceleration.cpp
+++ b/src/acceleration/BaseQNAcceleration.cpp
@@ -14,6 +14,7 @@
 #include "mesh/Mesh.hpp"
 #include "mesh/SharedPointer.hpp"
 #include "profiling/Event.hpp"
+#include "time/TimeGrids.hpp"
 #include "utils/EigenHelperFunctions.hpp"
 #include "utils/Helpers.hpp"
 #include "utils/IntraComm.hpp"
@@ -38,13 +39,15 @@ BaseQNAcceleration::BaseQNAcceleration(
     int                     filter,
     double                  singularityLimit,
     std::vector<int>        dataIDs,
-    impl::PtrPreconditioner preconditioner)
+    impl::PtrPreconditioner preconditioner,
+    bool                    reducedTimeGrid)
     : _preconditioner(std::move(preconditioner)),
       _initialRelaxation(initialRelaxation),
       _maxIterationsUsed(maxIterationsUsed),
       _timeWindowsReused(timeWindowsReused),
       _primaryDataIDs(std::move(dataIDs)),
       _forceInitialRelaxation(forceInitialRelaxation),
+      _reducedTimeGrid(reducedTimeGrid),
       _qrV(filter),
       _filter(filter),
       _singularityLimit(singularityLimit),
@@ -95,17 +98,10 @@ void BaseQNAcceleration::initialize(
     _dataIDs.push_back(pair.first);
   }
 
-  for (const auto &data : cplData | boost::adaptors::map_values) {
-    PRECICE_CHECK(!data->exchangeSubsteps(),
-                  "Quasi-Newton acceleration does not yet support using data from all substeps. Please set substeps=\"false\" in the exchange tag of data \"{}\".", data->getDataName());
-  }
-
   _matrixCols.clear();
   _matrixCols.push_front(0);
   _firstIteration  = true;
   _firstTimeWindow = true;
-
-  initializeVectorsAndPreconditioner(cplData);
 }
 
 /** ---------------------------------------------------------------------------------------------
@@ -220,15 +216,25 @@ void BaseQNAcceleration::performAcceleration(
 
   profiling::Event e("cpl.computeQuasiNewtonUpdate", profiling::Synchronize);
 
+  // We can only initialize here since we need to know the time grid already.
+  if (_firstTimeWindow and _firstIteration) {
+    initializeVectorsAndPreconditioner(cplData);
+  }
+
   PRECICE_ASSERT(_oldPrimaryResiduals.size() == _oldPrimaryXTilde.size(), _oldPrimaryResiduals.size(), _oldPrimaryXTilde.size());
   PRECICE_ASSERT(_primaryValues.size() == _oldPrimaryXTilde.size(), _primaryValues.size(), _oldPrimaryXTilde.size());
   PRECICE_ASSERT(_oldPrimaryValues.size() == _oldPrimaryXTilde.size(), _oldPrimaryValues.size(), _oldPrimaryXTilde.size());
   PRECICE_ASSERT(_primaryResiduals.size() == _oldPrimaryXTilde.size(), _primaryResiduals.size(), _oldPrimaryXTilde.size());
 
-  // assume data structures associated with the LS system can be updated easily.
+  // Needs to be called in the first iteration of each time window.
+  if (_firstIteration) {
+    _timeGrids->moveTimeGridToNewWindow(cplData);
+    _primaryTimeGrids->moveTimeGridToNewWindow(cplData);
+  }
 
-  concatenateCouplingData(cplData, _primaryDataIDs, _primaryValues, _oldPrimaryValues);
-  concatenateCouplingData(cplData, _dataIDs, _values, _oldValues);
+  /// Sample all the data to the corresponding time grid in _timeGrids and concatenate everything into a long vector
+  concatenateCouplingData(_values, _oldValues, cplData, _dataIDs, _timeGrids);
+  concatenateCouplingData(_primaryValues, _oldPrimaryValues, cplData, _primaryDataIDs, _primaryTimeGrids);
 
   /** update the difference matrices V,W  includes:
    * scaling of values
@@ -276,8 +282,8 @@ void BaseQNAcceleration::performAcceleration(
      * The preconditioner is only applied to the matrix V and the columns that are inserted into the
      * QR-decomposition of V.
      */
-
     _preconditioner->update(false, _primaryValues, _primaryResiduals);
+
     // apply scaling to V, V' := P * V (only needed to reset the QR-dec of V)
     _preconditioner->apply(_matrixV);
 
@@ -309,9 +315,7 @@ void BaseQNAcceleration::performAcceleration(
     Eigen::VectorXd xUpdate = Eigen::VectorXd::Zero(_values.size());
     computeQNUpdate(xUpdate);
 
-    /**
-     * apply quasiNewton update
-     */
+    // Apply the quasi-Newton update
     _values += xUpdate;
 
     // pending deletion: delete old V, W matrices if timeWindowsReused = 0
@@ -348,7 +352,9 @@ void BaseQNAcceleration::performAcceleration(
         "filter or increase its threshold (larger epsilon).");
   }
 
-  splitCouplingData(cplData); // split the primary and secondary coupling data back into the individual data objects
+  // updates the waveform and values in coupling data by splitting the primary and secondary data back into the individual data objects
+  updateCouplingData(cplData);
+
   // number of iterations (usually equals number of columns in LS-system)
   its++;
   _firstIteration = false;
@@ -376,19 +382,32 @@ void BaseQNAcceleration::applyFilter()
   }
 }
 
-void BaseQNAcceleration::splitCouplingData(
+void BaseQNAcceleration::updateCouplingData(
     const DataMap &cplData)
 {
-  PRECICE_TRACE();
 
-  int offset = 0;
+  PRECICE_TRACE();
+  // offset to keep track of the position in xUpdate
+  Eigen::Index offset = 0;
+
   for (int id : _dataIDs) {
-    int   size       = cplData.at(id)->getSize();
-    auto &valuesPart = cplData.at(id)->values();
-    for (int i = 0; i < size; i++) {
-      valuesPart(i) = _values(i + offset);
+
+    auto & couplingData = *cplData.at(id);
+    size_t dataSize     = couplingData.getSize();
+
+    Eigen::VectorXd timeGrid = _timeGrids->getTimeGrid(id);
+    couplingData.timeStepsStorage().clear();
+    for (int i = 0; i < timeGrid.size(); i++) {
+
+      Eigen::VectorXd temp = Eigen::VectorXd::Zero(dataSize);
+      for (size_t j = 0; j < dataSize; j++) {
+        temp(j) = _values(offset + j);
+      }
+      offset += dataSize;
+
+      couplingData.sample().values = temp;
+      couplingData.setSampleAtTime(timeGrid(i), couplingData.sample());
     }
-    offset += size;
   }
 }
 
@@ -415,8 +434,20 @@ void BaseQNAcceleration::iterationsConverged(
   // the most recent differences for the V, W matrices have not been added so far
   // this has to be done in iterations converged, as PP won't be called any more if
   // convergence was achieved
-  concatenateCouplingData(cplData, _primaryDataIDs, _primaryValues, _oldPrimaryValues);
-  concatenateCouplingData(cplData, _dataIDs, _values, _oldValues);
+
+  // If, in the first time window, we converge already in the first iteration, we have not yet initialized. Then, we need to do it here.
+  if (_firstTimeWindow and _firstIteration) {
+    initializeVectorsAndPreconditioner(cplData);
+  }
+
+  // Needs to be called in the first iteration of each time window.
+  if (_firstIteration) {
+    _timeGrids->moveTimeGridToNewWindow(cplData);
+    _primaryTimeGrids->moveTimeGridToNewWindow(cplData);
+  }
+  /// Sample all the data to the corresponding time grid in _timeGrids and concatenate everything into a long vector
+  concatenateCouplingData(_values, _oldValues, cplData, _dataIDs, _timeGrids);
+  concatenateCouplingData(_primaryValues, _oldPrimaryValues, cplData, _primaryDataIDs, _primaryTimeGrids);
   updateDifferenceMatrices(cplData);
 
   if (not _matrixCols.empty() && _matrixCols.front() == 0) { // Did only one iteration
@@ -588,41 +619,67 @@ void BaseQNAcceleration::writeInfo(
   _infostringstream << std::flush;
 }
 
-void BaseQNAcceleration::concatenateCouplingData(
-    const DataMap &cplData, const std::vector<DataID> &dataIDs, Eigen::VectorXd &targetValues, Eigen::VectorXd &targetOldValues) const
+void BaseQNAcceleration::concatenateCouplingData(Eigen::VectorXd &data, Eigen::VectorXd &oldData, const DataMap &cplData, std::vector<int> dataIDs, std::unique_ptr<precice::time::TimeGrids> &timeGrids) const
 {
   Eigen::Index offset = 0;
-  for (auto id : dataIDs) {
-    Eigen::Index size      = cplData.at(id)->values().size();
-    auto &       values    = cplData.at(id)->values();
-    const auto & oldValues = cplData.at(id)->previousIteration();
-    PRECICE_ASSERT(targetValues.size() >= offset + size, "Target vector was not initialized.", targetValues.size(), offset + size);
-    PRECICE_ASSERT(targetOldValues.size() >= offset + size, "Target vector was not initialized.");
-    for (Eigen::Index i = 0; i < size; i++) {
-      targetValues(i + offset)    = values(i);
-      targetOldValues(i + offset) = oldValues(i);
+  for (int id : dataIDs) {
+    Eigen::Index    dataSize = cplData.at(id)->getSize();
+    Eigen::VectorXd timeGrid = timeGrids->getTimeGrid(id);
+
+    for (int i = 0; i < timeGrid.size(); i++) {
+
+      Eigen::VectorXd values    = cplData.at(id)->timeStepsStorage().sample(timeGrid(i));
+      Eigen::VectorXd oldValues = cplData.at(id)->getPreviousValuesAtTime(timeGrid(i));
+
+      PRECICE_ASSERT(data.size() >= offset + dataSize, "the values were not initialized correctly");
+      PRECICE_ASSERT(oldData.size() >= offset + dataSize, "the values were not initialized correctly");
+
+      for (Eigen::Index i = 0; i < dataSize; i++) {
+        data(i + offset)    = values(i);
+        oldData(i + offset) = oldValues(i);
+      }
+      offset += dataSize;
     }
-    offset += size;
   }
 }
 
 void BaseQNAcceleration::initializeVectorsAndPreconditioner(const DataMap &cplData)
 {
-  auto         addCplDataSize  = [&](size_t sum, int id) { return sum + cplData.at(id)->getSize(); };
-  const size_t primaryDataSize = std::accumulate(_primaryDataIDs.begin(), _primaryDataIDs.end(), (size_t) 0, addCplDataSize);
-  const size_t dataSize        = std::accumulate(_dataIDs.begin(), _dataIDs.end(), (size_t) 0, addCplDataSize);
 
-  _oldPrimaryXTilde    = Eigen::VectorXd::Zero(primaryDataSize);
-  _oldPrimaryResiduals = Eigen::VectorXd::Zero(primaryDataSize);
-  _primaryResiduals    = Eigen::VectorXd::Zero(primaryDataSize);
-  _primaryValues       = Eigen::VectorXd::Zero(primaryDataSize);
-  _oldPrimaryValues    = Eigen::VectorXd::Zero(primaryDataSize);
+  // If we are not subcycling then we only want to use the last time step in the QN system
+  bool subcycling = false;
+  for (const auto &data : cplData | boost::adaptors::map_values) {
+    if (data->exchangeSubsteps()) {
+      subcycling = true;
+    }
+  }
+
+  // Saves the time grid of each waveform in the data field to be used in the QN method
+  _timeGrids        = std::make_unique<time::TimeGrids>(cplData, _dataIDs, !subcycling);
+  _primaryTimeGrids = std::make_unique<time::TimeGrids>(cplData, _primaryDataIDs, !(subcycling and !_reducedTimeGrid));
+
+  // Helper function
+  auto addTimeSliceSize = [&](size_t sum, int id, std::unique_ptr<precice::time::TimeGrids> &timeGrids) { return sum + timeGrids->getTimeGrid(id).size() * cplData.at(id)->getSize(); };
+
+  // Size of primary data
+  const size_t primaryDataSize = std::accumulate(_primaryDataIDs.begin(), _primaryDataIDs.end(), (size_t) 0, [&](size_t sum, int id) { return addTimeSliceSize(sum, id, _primaryTimeGrids); });
+
+  // Size of values
+  const size_t dataSize = std::accumulate(_dataIDs.begin(), _dataIDs.end(), (size_t) 0, [&](size_t sum, int id) { return addTimeSliceSize(sum, id, _timeGrids); });
+
   _values              = Eigen::VectorXd::Zero(dataSize);
   _oldValues           = Eigen::VectorXd::Zero(dataSize);
   _oldXTilde           = Eigen::VectorXd::Zero(dataSize);
   _residuals           = Eigen::VectorXd::Zero(dataSize);
-  _matrixV.resize(0, 0);
+  _primaryValues       = Eigen::VectorXd::Zero(primaryDataSize);
+  _oldPrimaryValues    = Eigen::VectorXd::Zero(primaryDataSize);
+  _oldPrimaryXTilde    = Eigen::VectorXd::Zero(primaryDataSize);
+  _primaryResiduals    = Eigen::VectorXd::Zero(primaryDataSize);
+  _oldPrimaryResiduals = Eigen::VectorXd::Zero(primaryDataSize);
+
+  // Also clear the matrices, since the initialization phase will be called more than once when using remeshing
   _matrixW.resize(0, 0);
+  _matrixV.resize(0, 0);
 
   /**
    *  make dimensions public to all procs,
@@ -652,14 +709,17 @@ void BaseQNAcceleration::initializeVectorsAndPreconditioner(const DataMap &cplDa
 
       for (auto &elem : _dataIDs) {
         const auto &offsets = cplData.at(elem)->getVertexOffsets();
-        accumulatedNumberOfUnknowns += offsets[i] * cplData.at(elem)->getDimensions();
+
+        accumulatedNumberOfUnknowns += offsets[i] * cplData.at(elem)->getDimensions() * _timeGrids->getTimeGrid(elem).size();
+
         if (utils::contained(elem, _primaryDataIDs)) {
-          accumulatedNumberOfPrimaryUnknowns += offsets[i] * cplData.at(elem)->getDimensions();
+          accumulatedNumberOfPrimaryUnknowns += offsets[i] * cplData.at(elem)->getDimensions() * _primaryTimeGrids->getTimeGrid(elem).size();
         }
       }
       _dimOffsets[i + 1]        = accumulatedNumberOfUnknowns;
       _dimOffsetsPrimary[i + 1] = accumulatedNumberOfPrimaryUnknowns;
     }
+
     PRECICE_DEBUG("Number of unknowns at the interface (global): {}", _dimOffsets.back());
     if (utils::IntraComm::isPrimary()) {
       _infostringstream << fmt::format("\n--------\n DOFs (global): {}\n offsets: {}\n", _dimOffsets.back(), _dimOffsets);
@@ -679,7 +739,7 @@ void BaseQNAcceleration::initializeVectorsAndPreconditioner(const DataMap &cplDa
   _qrV.setGlobalRows(getPrimaryLSSystemRows());
 
   std::vector<size_t> subVectorSizes; // needed for preconditioner
-  std::transform(_primaryDataIDs.cbegin(), _primaryDataIDs.cend(), std::back_inserter(subVectorSizes), [&cplData](const auto &d) { return cplData.at(d)->getSize(); });
+  std::transform(_primaryDataIDs.cbegin(), _primaryDataIDs.cend(), std::back_inserter(subVectorSizes), [&cplData, this](const auto &d) { return _primaryTimeGrids->getTimeGrid(d).size() * cplData.at(d)->getSize(); });
   _preconditioner->initialize(subVectorSizes);
 
   specializedInitializeVectorsAndPreconditioner(cplData);
--- a/src/acceleration/BaseQNAcceleration.cpp
+++ b/src/acceleration/BaseQNAcceleration.cpp
@@ -228,13 +228,14 @@ void BaseQNAcceleration::performAcceleration(
 
   // Needs to be called in the first iteration of each time window.
   if (_firstIteration) {
-    _timeGrids->moveTimeGridToNewWindow(cplData);
-    _primaryTimeGrids->moveTimeGridToNewWindow(cplData);
+    _timeGrids.value().moveTimeGridToNewWindow(cplData);
+    _primaryTimeGrids.value().moveTimeGridToNewWindow(cplData);
   }
 
   /// Sample all the data to the corresponding time grid in _timeGrids and concatenate everything into a long vector
-  concatenateCouplingData(_values, _oldValues, cplData, _dataIDs, _timeGrids);
-  concatenateCouplingData(_primaryValues, _oldPrimaryValues, cplData, _primaryDataIDs, _primaryTimeGrids);
+  /// timeGrids are stored using std::opional, thus the .value() to get the actual object
+  concatenateCouplingData(_values, _oldValues, cplData, _dataIDs, _timeGrids.value());
+  concatenateCouplingData(_primaryValues, _oldPrimaryValues, cplData, _primaryDataIDs, _primaryTimeGrids.value());
 
   /** update the difference matrices V,W  includes:
    * scaling of values
@@ -446,8 +447,9 @@ void BaseQNAcceleration::iterationsConverged(
     _primaryTimeGrids->moveTimeGridToNewWindow(cplData);
   }
   /// Sample all the data to the corresponding time grid in _timeGrids and concatenate everything into a long vector
-  concatenateCouplingData(_values, _oldValues, cplData, _dataIDs, _timeGrids);
-  concatenateCouplingData(_primaryValues, _oldPrimaryValues, cplData, _primaryDataIDs, _primaryTimeGrids);
+  /// timeGrids are stored using std::opional, thus the .value() to get the actual object
+  concatenateCouplingData(_values, _oldValues, cplData, _dataIDs, _timeGrids.value());
+  concatenateCouplingData(_primaryValues, _oldPrimaryValues, cplData, _primaryDataIDs, _primaryTimeGrids.value());
   updateDifferenceMatrices(cplData);
 
   if (not _matrixCols.empty() && _matrixCols.front() == 0) { // Did only one iteration
@@ -619,12 +621,12 @@ void BaseQNAcceleration::writeInfo(
   _infostringstream << std::flush;
 }
 
-void BaseQNAcceleration::concatenateCouplingData(Eigen::VectorXd &data, Eigen::VectorXd &oldData, const DataMap &cplData, std::vector<int> dataIDs, std::unique_ptr<precice::time::TimeGrids> &timeGrids) const
+void BaseQNAcceleration::concatenateCouplingData(Eigen::VectorXd &data, Eigen::VectorXd &oldData, const DataMap &cplData, std::vector<int> dataIDs, precice::time::TimeGrids timeGrids) const
 {
   Eigen::Index offset = 0;
   for (int id : dataIDs) {
     Eigen::Index    dataSize = cplData.at(id)->getSize();
-    Eigen::VectorXd timeGrid = timeGrids->getTimeGrid(id);
+    Eigen::VectorXd timeGrid = timeGrids.getTimeGrid(id);
 
     for (int i = 0; i < timeGrid.size(); i++) {
 
@@ -655,17 +657,17 @@ void BaseQNAcceleration::initializeVectorsAndPreconditioner(const DataMap &cplDa
   }
 
   // Saves the time grid of each waveform in the data field to be used in the QN method
-  _timeGrids        = std::make_unique<time::TimeGrids>(cplData, _dataIDs, !subcycling);
-  _primaryTimeGrids = std::make_unique<time::TimeGrids>(cplData, _primaryDataIDs, !(subcycling and !_reducedTimeGrid));
+  _timeGrids.emplace(cplData, _dataIDs, !subcycling);
+  _primaryTimeGrids.emplace(cplData, _primaryDataIDs, !(subcycling and !_reducedTimeGrid));
 
   // Helper function
-  auto addTimeSliceSize = [&](size_t sum, int id, std::unique_ptr<precice::time::TimeGrids> &timeGrids) { return sum + timeGrids->getTimeGrid(id).size() * cplData.at(id)->getSize(); };
+  auto addTimeSliceSize = [&](size_t sum, int id, precice::time::TimeGrids timeGrids) { return sum + timeGrids.getTimeGrid(id).size() * cplData.at(id)->getSize(); };
 
   // Size of primary data
-  const size_t primaryDataSize = std::accumulate(_primaryDataIDs.begin(), _primaryDataIDs.end(), (size_t) 0, [&](size_t sum, int id) { return addTimeSliceSize(sum, id, _primaryTimeGrids); });
+  const size_t primaryDataSize = std::accumulate(_primaryDataIDs.begin(), _primaryDataIDs.end(), (size_t) 0, [&](size_t sum, int id) { return addTimeSliceSize(sum, id, _primaryTimeGrids.value()); });
 
   // Size of values
-  const size_t dataSize = std::accumulate(_dataIDs.begin(), _dataIDs.end(), (size_t) 0, [&](size_t sum, int id) { return addTimeSliceSize(sum, id, _timeGrids); });
+  const size_t dataSize = std::accumulate(_dataIDs.begin(), _dataIDs.end(), (size_t) 0, [&](size_t sum, int id) { return addTimeSliceSize(sum, id, _timeGrids.value()); });
 
   _values              = Eigen::VectorXd::Zero(dataSize);
   _oldValues           = Eigen::VectorXd::Zero(dataSize);
@@ -710,10 +712,10 @@ void BaseQNAcceleration::initializeVectorsAndPreconditioner(const DataMap &cplDa
       for (auto &elem : _dataIDs) {
         const auto &offsets = cplData.at(elem)->getVertexOffsets();
 
-        accumulatedNumberOfUnknowns += offsets[i] * cplData.at(elem)->getDimensions() * _timeGrids->getTimeGrid(elem).size();
+        accumulatedNumberOfUnknowns += offsets[i] * cplData.at(elem)->getDimensions() * _timeGrids.value().getTimeGrid(elem).size();
 
         if (utils::contained(elem, _primaryDataIDs)) {
-          accumulatedNumberOfPrimaryUnknowns += offsets[i] * cplData.at(elem)->getDimensions() * _primaryTimeGrids->getTimeGrid(elem).size();
+          accumulatedNumberOfPrimaryUnknowns += offsets[i] * cplData.at(elem)->getDimensions() * _primaryTimeGrids.value().getTimeGrid(elem).size();
         }
       }
       _dimOffsets[i + 1]        = accumulatedNumberOfUnknowns;
@@ -739,7 +741,7 @@ void BaseQNAcceleration::initializeVectorsAndPreconditioner(const DataMap &cplDa
   _qrV.setGlobalRows(getPrimaryLSSystemRows());
 
   std::vector<size_t> subVectorSizes; // needed for preconditioner
-  std::transform(_primaryDataIDs.cbegin(), _primaryDataIDs.cend(), std::back_inserter(subVectorSizes), [&cplData, this](const auto &d) { return _primaryTimeGrids->getTimeGrid(d).size() * cplData.at(d)->getSize(); });
+  std::transform(_primaryDataIDs.cbegin(), _primaryDataIDs.cend(), std::back_inserter(subVectorSizes), [&cplData, this](const auto &d) { return _primaryTimeGrids.value().getTimeGrid(d).size() * cplData.at(d)->getSize(); });
   _preconditioner->initialize(subVectorSizes);
 
   specializedInitializeVectorsAndPreconditioner(cplData);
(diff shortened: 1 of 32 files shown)


BenjaminRodenberg:
@NiklasKotarsky currently this PR does not build on my system and also not in our CI pipeline. Does it build on your system? Did you maybe forget to push some file?

NiklasKotarsky:
> @NiklasKotarsky currently this PR does not build on my system and also not in our CI pipeline. Does it build on your system? Did you maybe forget to push some file?

Yes, it compiles on my system and some of the systems in the CI pipline. This is probably due to different versions of Eigen having different feature support.

BenjaminRodenberg:
> > @NiklasKotarsky currently this PR does not build on my system and also not in our CI pipeline. Does it build on your system? Did you maybe forget to push some file?
> 
> Yes, it compiles on my system and some of the systems in the CI pipline. This is probably due to different versions of Eigen having different feature support.

If it is an issue with different eigen version: I once had a similar problem with some features not being supported by the Eigen version used by the CI pipeline and added the following code to resolve this:

https://github.com/precice/precice/blob/0d43d140341a2c0c24dcda65e49eaa48e9c38cf0/src/math/Bspline.cpp#L18-L22

Which Eigen version are you using on your system?

uekerman:
Now that #1996 is done, we could reformulate this PR using the new "secondary" implementation.

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -666,5 +581,241 @@ void BaseQNAcceleration::writeInfo(
...
+  size_t              primaryDataSize = 0;
+  std::vector<size_t> subVectorSizes; // needed for preconditioner
+
+  for (auto id : primaryDataIDs) {
+    if (_reduced) {
+      primaryDataSize += cplData.at(id)->values().size();
+      subVectorSizes.push_back(cplData.at(id)->getSize());
+    } else {
+      primaryDataSize += cplData.at(id)->timeStepsStorage().nTimes() * cplData.at(id)->values().size();
+      subVectorSizes.push_back(cplData.at(id)->timeStepsStorage().nTimes() * cplData.at(id)->getSize());
```
BenjaminRodenberg, line 684:
```suggestion
      primaryDataSize += cplData.at(id)->getSize();
      subVectorSizes.push_back(cplData.at(id)->getSize());
    } else {
      primaryDataSize += cplData.at(id)->timeStepsStorage().nTimes() * cplData.at(id)->getSize();
      subVectorSizes.push_back(cplData.at(id)->timeStepsStorage().nTimes() * cplData.at(id)->getSize());
```
`values().size()` and getSize()` is identical. I would avoid using `values().size()` (see https://en.wikipedia.org/wiki/Law_of_Demeter).

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -666,5 +581,241 @@ void BaseQNAcceleration::writeInfo(
...
+
+  size_t dataSize = 0;
+  for (auto id : dataIDs) {
+    dataSize += cplData.at(id)->timeStepsStorage().nTimes() * cplData.at(id)->values().size();
+  }
+
+  size_t              primaryDataSize = 0;
+  std::vector<size_t> subVectorSizes; // needed for preconditioner
+
+  for (auto id : primaryDataIDs) {
```
BenjaminRodenberg, line 678:
```suggestion
  for (auto id : primaryDataIDs) {
    auto primaryCplData = cplData.at(id);
```
And then use `primaryCplData` below? I think this would also be helpful.

Btw: Why `primaryDataIDs` and not `_primaryDataIDs`? Is it allowed to provide `primaryDataIDs != _primaryDataIDs` to this function?

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -666,5 +581,241 @@ void BaseQNAcceleration::writeInfo(
...
+void BaseQNAcceleration::saveTimeGrid(const DataMap &cplData)
+{
+  for (auto &pair : cplData) {
+    auto            dataID   = pair.first;
+    Eigen::VectorXd timeGrid = pair.second->timeStepsStorage().getTimes();
+    _timeGrids.insert(std::pair<int, Eigen::VectorXd>(dataID, timeGrid));
+  }
+}
+
+void BaseQNAcceleration::initializeVectorsAndPreconditioner(const DataMap &cplData, const std::vector<DataID> &dataIDs, const std::vector<DataID> &primaryDataIDs)
```
BenjaminRodenberg, line 667:
```suggestion
void BaseQNAcceleration::initializeVectorsAndPreconditioner(const DataMap &cplData)
```
I think you could simplify this function by using `_dataIDs` and `_primaryDataIDs`. No?

BenjaminRodenberg (commented):
I found some minor points regarding clean code. I also think there are some candidates for `const` in `BaseQNAcceleration` (e.g. `_primaryDataIDs`).

BenjaminRodenberg:
> I also think there are some candidates for `const` in `BaseQNAcceleration` (e.g. `_primaryDataIDs`).

I added a corresponding commit on `develop`: https://github.com/precice/precice/commit/3323020c159a48a9f083677c7372f62128f8ae9a

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -666,5 +581,241 @@ void BaseQNAcceleration::writeInfo(
...
+      PRECICE_ASSERT(primaryValues.size() >= offset + dataSize, "the primaryValues were not initialized correctly");
+      PRECICE_ASSERT(oldPrimaryValues.size() >= offset + dataSize, "the primaryValues were not initialized correctly");
+
+      for (Eigen::Index i = 0; i < dataSize; i++) {
+        primaryValues(i + offset)    = primaryData(i);
+        oldPrimaryValues(i + offset) = oldPrimaryData(i);
+      }
+      offset += dataSize;
+    }
+  }
```
BenjaminRodenberg, line 633: I'm not so happy with the amount of code duplication here and at other points where you are branching depending on `_reduced`. Why not do the following:

```
Eigen::Index offset = 0;
for (int id : _primaryDataIDs) {
  Eigen::VectorXd timeGrid;
  if (!_reduced) {
    timeGrid = _timeGrids.at(id);
  } else {
    timeGrid = ...; // reduced time grid only containing _timeGrids.at(id).last();
  }

  // do everything like implemented in the "if (!_reduced)"-part; should be possible to generalize the "else" part correspondingly.
}
```

This also has the side effect that you would use `waveform.sample(timeGrid(i));` instead of `cplData.at(id)->values()`. Sampling from `waveform` is in my opinion preferrable since `values()` corresponds to the raw data and is a bit more low-level.

Please also use `CouplingData::getSize()` and not `values().size()`. Reason: at some point it would be nice to completely get rid off `values()` and only access the `CouplingData::timeStepsStorage()` (i.e. the waveform).

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -666,5 +581,241 @@ void BaseQNAcceleration::writeInfo(
   }
   _infostringstream << std::flush;
 }
+
+void BaseQNAcceleration::concatenateCouplingData(
+    const DataMap &cplData, const std::vector<DataID> &dataIDs, const std::vector<DataID> &primaryDataIDs, Eigen::VectorXd &values,
+    Eigen::VectorXd &oldValues, Eigen::VectorXd &primaryValues, Eigen::VectorXd &oldPrimaryValues)
```
BenjaminRodenberg, line 587:
```suggestion
void BaseQNAcceleration::concatenateCouplingData(const DataMap &cplData)
```
Could be simplified by using the private members `_dataIDs, _primaryDataIDs, _values...` below.

BenjaminRodenberg: Changes provided via [ee153fa](https://github.com/precice/precice/pull/2005/commits/ee153faa08530c21618bdb63b9569e877cce666f)

BenjaminRodenberg:
Accidentally pushed [ee153fa](https://github.com/precice/precice/pull/2005/commits/ee153faa08530c21618bdb63b9569e877cce666f) and the merge to `develop` [675554b](https://github.com/precice/precice/pull/2005/commits/675554bab4a289c36cbc8e94047ff2b01f5ee0f3) to this branch. Originally wanted to push it on my own copy of this branch (https://github.com/BenjaminRodenberg/precice/tree/new-QNWR). I hope this is no problem. We can, of course, revert my changes.

NiklasKotarsky:
Thanks, for the comments @BenjaminRodenberg ! I have tried to reduce the code duplication now as well as making it a little bit easier to read. I think that this is ready for proper review now.

src/acceleration/BaseQNAcceleration.hpp
```diff
@@ -203,6 +204,9 @@ class BaseQNAcceleration : public Acceleration {
   /// @brief Stores x tilde deltas, where x tilde are values computed by solvers.
   Eigen::MatrixXd _matrixW;
 
+  /// @brief  Toggle to switch between using all of the substeps when constructing V_k or only the last time step of the time window.
+  bool _reduced = true;
```
BenjaminRodenberg, line 208:
```suggestion
  /// @brief Toggle to switch between using all of the substeps when constructing V_k or only the last time step of the time window.
  bool _reduced = true;
```
Are you referring to https://onlinelibrary.wiley.com/doi/epdf/10.1002/nme.6443 somewhere? I think otherwise the term `_reduced` might sound a bit cryptic. You could even consider writing " if `_reduced = false` uses the full QN-WI and if `_reduced = true` uses rQN-WI form paper ..."

src/acceleration/IQNILSAcceleration.cpp
```diff
@@ -52,7 +53,7 @@ void IQNILSAcceleration::updateDifferenceMatrices(
   BaseQNAcceleration::updateDifferenceMatrices(cplData);
 }
 
-void IQNILSAcceleration::computeQNUpdate(const DataMap &cplData, Eigen::VectorXd &xUpdate)
+void IQNILSAcceleration::computeQNUpdate(Eigen::VectorXd &xUpdate)
```
BenjaminRodenberg, line 56: Maybe directly merge some of these refactorings into `develop` to make this PR here a bit less  crowded?

Pulling out the refactorings is a bit tedious but I think it really helps us to concentrate on the important parts here. We can also do this together on Thursday.

src/acceleration/BaseQNAcceleration.hpp
```diff
@@ -256,23 +260,51 @@ class BaseQNAcceleration : public Acceleration {
...
 
   /// Computes the quasi-Newton update using the specified pp scheme (IQNIMVJ, IQNILS)
-  virtual void computeQNUpdate(const DataMap &cplData, Eigen::VectorXd &xUpdate) = 0;
+  virtual void computeQNUpdate(Eigen::VectorXd &xUpdate) = 0;
 
   /// Removes one iteration from V,W matrices and adapts _matrixCols.
   virtual void removeMatrixColumn(int columnIndex);
 
-  /// Wwrites info to the _infostream (also in parallel)
+  /// Writes info to the _infostream (also in parallel)
```
BenjaminRodenberg, line 274: Again: Candidate for direct merge into `develop`

src/acceleration/IQNIMVJAcceleration.cpp
```diff
@@ -79,45 +80,6 @@ void IQNIMVJAcceleration::initialize(
...
-    _matrixCols_RSLS.push_front(0);
-    _matrixV_RSLS = Eigen::MatrixXd::Zero(entries, 0);
-    _matrixW_RSLS = Eigen::MatrixXd::Zero(cplDataEntries, 0);
-  }
-  _Wtil = Eigen::MatrixXd::Zero(cplDataEntries, 0);
-
-  if (utils::IntraComm::isPrimary() || !utils::IntraComm::isParallel()) {
-    _infostringstream << " IMVJ restart mode: " << _imvjRestart << "\n chunk size: " << _chunkSize << "\n trunc eps: " << _svdJ.getThreshold() << "\n R_RS: " << _RSLSreusedTimeWindows << "\n--------\n"
-                      << '\n';
-  }
```
BenjaminRodenberg, line 120: Can you explain why all this code could be removed? Did we decide to delete an (unused) feature?

Edit: Just saw that you moved it into `specializedInitializeVectorsAndPreconditioner`. Again: Is this maybe a refactoring we could directly merge into develop?

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -90,101 +92,18 @@ void BaseQNAcceleration::initialize(
...
-    size_t primaryUnknowns = _dimOffsetsPrimary[utils::IntraComm::getRank() + 1] - _dimOffsetsPrimary[utils::IntraComm::getRank()];
-    PRECICE_ASSERT(primaryDataSize == primaryUnknowns, primaryDataSize, primaryUnknowns);
-  } else {
-    _infostringstream << fmt::format("\n--------\n DOFs (global): {}\n", dataSize);
-  }
-
-  // set the number of global rows in the QRFactorization.
-  _qrV.setGlobalRows(getPrimaryLSSystemRows());
-
-  _preconditioner->initialize(subVectorSizes);
```
BenjaminRodenberg, line 187: Again looks to me like a refactoring that we could take care of separately to make the main part more clear.

BenjaminRodenberg (changes requested):
I think we can split this PR into a refactoring part and an actual QN-WI acceleration part. The goal is to move as much as possible into the refactoring part to make the actual QN-WI acceleration part easier to understand. Generally, I have the impression that everything we need is here, it is just a bit hard to follow all the changes.

I have quite some experience with "pull refactorings into a different PR" from my gigantic waveform PRs in the past. So we could also do this together on Thursday if you want.

I will do some additional testing with my cases and the reduced QN from this PR during this week.

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -666,5 +580,223 @@ void BaseQNAcceleration::writeInfo(
...
+      PRECICE_ASSERT(_values.size() >= offset + dataSize, "the values were not initialized correctly");
+
+      for (Eigen::Index i = 0; i < dataSize; i++) {
+        _values(i + offset)    = data(i);
+        _oldValues(i + offset) = oldData(i);
+      }
+      offset += dataSize;
+    }
+  }
+}
```
BenjaminRodenberg, line 635: The blocks `for (int id : _primaryDataIDs)` and `for (int id : _dataIDs)` are almost identical. I think you could remove some duplication by introducing a function:
```suggestion
void BaseQNAcceleration::concatenateCouplingData(const DataMap &cplData)
{
  // Needs to be called every iteration, since the time window size can vary with participant first
  moveTimeGridToNewWindow(cplData);

    /// If not reduced quasi-Newton then sample the residual of data in dataIDs to the corresponding time grid in _timeGrids and concatenate everything into a long vector  
  doConcatenate(_primaryValues, _oldPrimaryValues, _primaryDataIDs, _primaryTimeGrids);
  
  /// Sample all the data to the corresponding time grid in _timeGrids and concatenate everything into a long vector
  doConcatenate(_values, _oldValues, _dataIDs, _timeGrids);
}
```
I think this also makes sense since you forgot the assertion `PRECICE_ASSERT(_oldValues.size() >= offset + dataSize, "the primaryValues were not initialized correctly");` in the second block. This cannot happen with a function that easily.

src/acceleration/IQNIMVJAcceleration.cpp
NiklasKotarsky: Yes, I had to move this code into specializedInitializeVectorsAndPreconditioner. The reason behind this change is that the vector and preconditioner initialization needs to know the time grid of the Quasi-Newton method, which is determined after the first waveform iteration. 

It probably makes sense to refactor this change in interpolation in a separate pull request. It would also be nice if there was a way of making this more robust for future changes and easier to follow.

NiklasKotarsky:
Following @BenjaminRodenberg great advice I have now split out the time grids in the Quasi-Newton method into their own classes. The reasons for this change are that this allows us to test the behavior of the time grids directly. Allowing us to ensure the correct behavior with participant first indirectly. Secondly, to support waveform iterations in Aitken we would need to reuse these parts anyway. It is also debatable weather the time grids methods belong to the Quasi-Newton or acceleration classes content wise.

BenjaminRodenberg:
Hi! What's the status for this PR? I'm planning to perform some experiments using the partitioned heat equation + this PR and compare to the results from https://onlinelibrary.wiley.com/doi/epdf/10.1002/nme.6443 next week. I think it would be great if we could get this PR merged before our next coding days (Nov 11-13) since IIRC this would be the last chance to get it into the 3.2 feature release (@uekerman please correct me, if I'm wrong).

@NiklasKotarsky do you think this is a somewhat realistic plan? Anything you would need from my side?

BenjaminRodenberg (commented):
I merged `develop` into this branch. There were some conflicts with the changes we introduced in #2073 that I had to resolve manually.

BenjaminRodenberg:
I introduced a bug via [f66061a](https://github.com/precice/precice/pull/2005/commits/f66061a511bc740066f71dc80084ce1b9174ac09):

```diff
- const size_t primaryDataSize = std::accumulate(_primaryDataIDs.begin(), _primaryDataIDs.end(), (size_t) 0, [&](size_t val, int id) { return val + _primaryTimeGrids.getTimeGrid(id).size() * cplData.at(id)->getSize(); });
- const size_t dataSize        = std::accumulate(_dataIDs.begin(), _dataIDs.end(), (size_t) 0, [&](size_t val, int id) { return val + _timeGrids.getTimeGrid(id).size() * cplData.at(id)->getSize(); });
+ auto         addTimeSliceSize = [&](size_t sum, int id) { return sum + _timeGrids.getTimeGrid(id).size() * cplData.at(id)->getSize(); };
+ const size_t primaryDataSize  = std::accumulate(_primaryDataIDs.begin(), _primaryDataIDs.end(), (size_t) 0, addTimeSliceSize);
+ const size_t dataSize         = std::accumulate(_dataIDs.begin(), _dataIDs.end(), (size_t) 0, addTimeSliceSize);
```

After some confusion I realized that I'm using `_timeGrids.getTimeGrid(id)` at both places. But this is not the same as `_primaryTimeGrids.getTimeGrid(id)`. I don't know how to solve this better. Any idea @NiklasKotarsky?

For now I will revert these changes to remove the bug.

BenjaminRodenberg:
I performed the experiments from the [QNWI paper](https://onlinelibrary.wiley.com/doi/epdf/10.1002/nme.6443) again using the code from this PR and got the following results:

image

Some comments:

* I skipped QN-WI. IIRC we are currently focusing on rQN-WI.
* Number of iterations is generally similar to the results from QNWI paper.
* The number of coupling iterations is almost independent from the multirate setup. Stands out even more clearly than in the QNWI paper.
* There are only minimal differences between rel-WI and rel-SC.
* Often the iteration numbers from the QNWI paper are slightly larger than the numbers observed here. Possible explanation: In this PR we only measure the convergence w.r.t the last sample in each window. In the QNWI paper, we applied more strict convergence checks and compared all the available samples[^1]

[^1]: not possible here due to adaptivity and changing time grids from one iteration to the next)

NiklasKotarsky:
> I introduced a bug via [f66061a](https://github.com/precice/precice/pull/2005/commits/f66061a511bc740066f71dc80084ce1b9174ac09):
> 
> ```diff
> - const size_t primaryDataSize = std::accumulate(_primaryDataIDs.begin(), _primaryDataIDs.end(), (size_t) 0, [&](size_t val, int id) { return val + _primaryTimeGrids.getTimeGrid(id).size() * cplData.at(id)->getSize(); });
> - const size_t dataSize        = std::accumulate(_dataIDs.begin(), _dataIDs.end(), (size_t) 0, [&](size_t val, int id) { return val + _timeGrids.getTimeGrid(id).size() * cplData.at(id)->getSize(); });
> + auto         addTimeSliceSize = [&](size_t sum, int id) { return sum + _timeGrids.getTimeGrid(id).size() * cplData.at(id)->getSize(); };
> + const size_t primaryDataSize  = std::accumulate(_primaryDataIDs.begin(), _primaryDataIDs.end(), (size_t) 0, addTimeSliceSize);
> + const size_t dataSize         = std::accumulate(_dataIDs.begin(), _dataIDs.end(), (size_t) 0, addTimeSliceSize);
> ```
> 
> After some confusion I realized that I'm using `_timeGrids.getTimeGrid(id)` at both places. But this is not the same as `_primaryTimeGrids.getTimeGrid(id)`. I don't know how to solve this better. Any idea @NiklasKotarsky?
> 
> For now I will revert these changes to remove the bug.
I use '_primaryTimeGrids' to toggle between the reduced variant and the non reduced variant as well. For the reduced variant '_primaryTimeGrids' only contains the last time step compared to '_timeGrids' which contains all time steps. The only solution I have is to use two nested lambda functions like this: 

``` 
// Helper function
  auto addTimeSliceSize = [&](size_t sum, int id, impl::WaveformTimeGrids timeGrids) { return sum + timeGrids.getTimeGrid(id).size() * cplData.at(id)->getSize(); };

  // Size of primary data
  const size_t primaryDataSize = std::accumulate(_primaryDataIDs.begin(), _primaryDataIDs.end(), (size_t) 0, [&](size_t sum, int id) { return addTimeSliceSize(sum, id, _primaryTimeGrids); });

  // Size of values
  const size_t dataSize = std::accumulate(_dataIDs.begin(), _dataIDs.end(), (size_t) 0, [&](size_t sum, int id) { return addTimeSliceSize(sum, id, _timeGrids); });
```

NiklasKotarsky:
> Hi! What's the status for this PR? I'm planning to perform some experiments using the partitioned heat equation + this PR and compare to the results from https://onlinelibrary.wiley.com/doi/epdf/10.1002/nme.6443 next week. I think it would be great if we could get this PR merged before our next coding days (Nov 11-13) since IIRC this would be the last chance to get it into the 3.2 feature release (@uekerman please correct me, if I'm wrong).
> 
> @NiklasKotarsky do you think this is a somewhat realistic plan? Anything you would need from my side?

This sounds reasonable to me. I will try to split this into two smaller pull requests with the waveform functionality to make it easier to review this or next week.

BenjaminRodenberg (commented):
From my perspective the code looks good now. I still think it would be good, if @uekerman and maybe also @fsimonis take a look at the PR. I'm also performing some tests with this PR this and next week. So I would suggest: Let's review this PR approx. one more week and then try to merge.

BenjaminRodenberg:
> I will try to split this into two smaller pull requests with the waveform functionality to make it easier to review this or next week.

I currently do not completely understand where you want to split. Can you explain in a bit more detail? What goes into PR1 and what goes into PR2? (or if you are already working on this I'm also fine waiting for the PRs and looking at the content)

src/acceleration/impl/WaveformTimeGrids.hpp
```diff
@@ -0,0 +1,56 @@
...
+  using DataMap = std::map<int, cplscheme::PtrCouplingData>;
+
+  /**
+   * @brief sets the time grid
+   * @param cplData
+   * @param dataIDs
+   * @param reduced if true then only the last timestep is saved in the timeGrid otherwise all timesteps are saved in the time grid
+   * This is used in the Quasi-Newton method to toggle between the full and reduced variant detailed here
+   */
+  void setTimeGrid(const DataMap &cplData, std::vector<int> dataIDs, bool reduced);
```
uekerman, line 36: From a setter, I would expect that I give the time grid as an argument. Could you extend the comment what the function exactly does. Maybe choose a different name?
And shouldn't it be plural as we pass multiple dataIDs?

src/acceleration/impl/WaveformTimeGrids.hpp
```diff
@@ -0,0 +1,56 @@
...
+
+/**
+ * @brief Interface for storing the time grids in the Quasi-Newton and Aitken methods
+ *
+ *
+ * initialize saves the time grids of the waveforms
+ * moveTimeGridToNewWindow Moves the time grid to the new time window. This is done by translating and scaling the QN time grid such that its startpoint and end point is the same as the waveforms
+ *
+ */
+class WaveformTimeGrids {
```
uekerman, line 24: Does the "waveform" in the name really make sense?

src/acceleration/impl/WaveformTimeGrids.hpp
```diff
@@ -0,0 +1,56 @@
...
+private:
+  logging::Logger _log{"acceleration::WaveformTimeGrids"};
+
+  /// @brief List of the time grid to which all the data will be interpolated to
+  /// Stored in a map, since each data entry has its own time grid
+  std::map<int, Eigen::VectorXd> _timeGrids;
+
+  //These two are needed for the reduced variants, since then the _primaryTimeGrids only contain one timestep
+  std::map<int, double> _timeGridsStartTime;
+  std::map<int, double> _timeGridsEndTime;
```
uekerman, line 51:
```suggestion
  std::map<int, double> _timeGridStartTimes;
  std::map<int, double> _timeGridEndTimes;
```

src/acceleration/config/AccelerationConfiguration.cpp
```diff
@@ -101,6 +102,11 @@ void AccelerationConfiguration::connectTags(xml::XMLTag &parent)
   {
     XMLTag tag(*this, VALUE_IQNILS, occ, TAG);
     tag.setDocumentation("Accelerates coupling data with the interface quasi-Newton inverse least-squares method.");
+
+    auto reducedQN = makeXMLAttribute(ATTR_REDUCEDQN, true)
+                         .setDocumentation("If set to true, QN method will only use the last timestep when constructing the Jacobian.");
```
uekerman, line 107:
```suggestion
                         .setDocumentation("Whether only the last time step of each time window is used to construct the Jacobian.");
```

src/acceleration/config/AccelerationConfiguration.cpp
```diff
@@ -114,6 +120,10 @@ void AccelerationConfiguration::connectTags(xml::XMLTag &parent)
                                                     " the Jacobian is only build in the last iteration and the updates are computed using (relatively) cheap MATVEC products.");
     tag.addAttribute(alwaybuildJacobian);
 
+    auto reducedQN = makeXMLAttribute(ATTR_REDUCEDQN, true)
+                         .setDocumentation("If set to true, QN method will only use the last timestep when constructing the Jacobian.");
```
uekerman, line 124:
```suggestion
                         .setDocumentation("Whether only the last time step of each time window is used to construct the Jacobian.");
```

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -130,7 +126,6 @@ void BaseQNAcceleration::updateDifferenceMatrices(
                   "providing new data first or you do not use available read data. "
                   "Or you just converge much further than actually necessary.");
 
-  // if (_firstIteration && (_firstTimeWindow || (_matrixCols.size() < 2))) {
   if (_firstIteration && (_firstTimeWindow || _forceInitialRelaxation)) {
```
uekerman, line 129: Why this change?

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -218,15 +213,24 @@ void BaseQNAcceleration::performAcceleration(
 
   profiling::Event e("cpl.computeQuasiNewtonUpdate", profiling::Synchronize);
 
+  // If its the first iteration of the first time window we want to save the timegrid that the quasi-Newton method uses.
```
uekerman, line 216:
```suggestion
  // We can only initialize here since we need to know the time grid already.
```
?
(don't describe the what, but the why)

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -413,8 +401,21 @@ void BaseQNAcceleration::iterationsConverged(
   // the most recent differences for the V, W matrices have not been added so far
   // this has to be done in iterations converged, as PP won't be called any more if
   // convergence was achieved
-  concatenateCouplingData(cplData, _primaryDataIDs, _primaryValues, _oldPrimaryValues);
-  concatenateCouplingData(cplData, _dataIDs, _values, _oldValues);
+
+  // If its the first iteration of the first time window we want to save the timegrid that the quasi-Newton method uses.
+  // This might happen if we converge in the first iteration
```
uekerman, line 406:
```suggestion
  // If, in the first time window, we converge already in the first iteration, we have not yet initialized. Then, we need to do it here.
  ```

**Review by uekerman — CHANGES_REQUESTED** [review]
Looks pretty clean now. Mainly cosmetic comments below. No must to merge this before the coding days IMO, but let's try it. It is close to being done now. And sorry for the late reply.

- [ ] needs a changelog entry
- [ ] Currently, only IQN-ILS is covered in the integration tests, IQN-IMVJ should already work no?

@BenjaminRodenberg 

> I skipped QN-WI. IIRC we are currently focusing on rQN-WI.

How I understand this PR is that we now have both variants. Please correct me if I am wrong. Could you please test the non-reduced variant as well?

File: src/acceleration/BaseQNAcceleration.hpp [hunk 20]
```diff
@@ -70,7 +71,8 @@ class BaseQNAcceleration : public Acceleration {
       int                     filter,
       double                  singularityLimit,
       std::vector<int>        dataIDs,
-      impl::PtrPreconditioner preconditioner);
+      impl::PtrPreconditioner preconditioner,
+      bool                    reduced = true);
```
BenjaminRodenberg, line 75: Currently the configuration is hard-coded to `reduced = true`. If we want to test the reduced variant we need to make it configurable first and we also need corresponding tests. I would actually prefer sticking to the reduced variant for now. According to [our paper](https://onlinelibrary.wiley.com/doi/epdf/10.1002/nme.6443) the non-reduced variant has little value (at least for the cases that we have studied). The way that the QN scheme is implemented we can activate the non-reduced variant at a later point in time if needed. (would also be a nice topic for a thesis to find a proper use case)

uekerman: I think it is configurable already? See the integration tests. I mainly meant for validating the implementation now.

BenjaminRodenberg: True. It is configurable. I would rather remove this option from this PR, get this PR merged, and add it the non-reduced variant later if needed. From my perspective rQN is sufficient and the non-reduced variant has no clear benefit. (sure: maybe there are examples that I do not have on the radar at the moment) If we also want to offer the non-reduced variant this requires additional validation and documentation for which I do not really have time at the moment. Factoring out the non-reduced QN into an independent PR should not be a big problem.

I'm not sure on what we agreed originally but having the non-reduced variant as a feature that we add later if needed sounds to me like the right approach to get rQN merged soon. If we also want to have the non-reduced variant now this makes this PR more complex than necessary.

NiklasKotarsky: I can create a new pull request with the reduced variant. Before I do that I am going to try to figure out what broke when I merged develop with the re-meshing features into this branch. Ideally I would still keep the non reduced variant, since it has a lot nicer mathematical properties. 

For example if the coupled system is linear then the full QNWR will converge in Nd iteration, where d corresponds to the number of grid points in the interface and N to the number of time steps. If enough information from previous time steps is included this can be reduced to 2 iterations. Both of these statements do not hold for the reduced variant. I can also design linear systems for which the reduced variant converges very slowly. I have however not been able to find a linear system except for the non trivial ones where it does not converge to the correct solution.

BenjaminRodenberg: > For example if the coupled system is linear then the full QNWR will converge in Nd iteration, where d corresponds to the number of grid points in the interface and N to the number of time steps. If enough information from previous time steps is included this can be reduced to 2 iterations. Both of these statements do not hold for the reduced variant. I can also design linear systems for which the reduced variant converges very slowly. I have however not been able to find a linear system except for the non trivial ones where it does not converge to the correct solution.

Sounds to me like we should be able to show this for the heat equation? If this is possible, then I think looking at this question in another PR would be interesting.

Don't get me wrong: If the reduced variant has advantages over the non-reduced variant it we can definitely keep it. What you are mentioning sounds like we could use the heat equation as an end-to-end test for QNWI similar to how I used the heat equation with the manufactured solution to develop the time stepping. But I'm currently trying to close some boxes and not open new ones. That's the main reason I'm trying to get this PR merged rather sooner than later.

NiklasKotarsky:
Thanks for the review @uekerman. I will try to fix the feedback as soon as I figured out whats happening with the remeshing tests. Currently the tests in the test suite Integration/Remeshing/IQNILS fail. This is due to that the  timestep storage goes from containing a sample at the start and the end of the time window to only containing a sample at the end of the time window after the remeshing. This breaks the following assumption that I made inside the interpolation in QNWR: Namely, that each time window contains at least one sample at the start of the time window and one at the end of the time window. I am a little bit unsure if this is a bug or if I should rework my interpolation to allow for this. Another option would be to wait until #2122 is resolved, which is not ideal.

src/acceleration/BaseQNAcceleration.hpp
uekerman: IMO moving the non-reduced variant into a separate PR sounds unnecessarily complicated.
We have the implementation and we have integration tests here.

The only missing piece that we could do on top would be to repeat (some of) the tests of https://github.com/precice/precice/pull/2005#issuecomment-2425100108
@BenjaminRodenberg I assume you have this scripted. This should only be one config option to change and then start, no?

fsimonis:
> This is due to that the timestep storage goes from containing a sample at the start and the end of the time window to only containing a sample at the end of the time window after the remeshing.

The accelerations are all rest on remesh.

https://github.com/precice/precice/blob/70a2362a6d1e30d9842c589405a79a54c41b11b2/src/cplscheme/BaseCouplingScheme.cpp#L315

My current way of handling this in the CouplingData is to set all samples of the previous timewindow to zero of the correct new size.
The mesh data contains only one stample of the end of the time window.

When remeshing, I currently loose the start sample of the time window. I could try to preserve that by getting the start time from the couplingscheme.

I guess an upcoming issue for you is that all samples are suddenly 0.

NiklasKotarsky:
Thanks for the answer! My problem is actually: 
> When remeshing, I currently loose the start sample of the time window. I could try to preserve that by getting the start time from the couplingscheme.
When designing the interpolation and time grids inside QN method I assumed that the coupling data always contains a sample at the start of the time window and at the end of the time window. Adding a sample at the start of the time window should fix this problem. 

I have also started to work on a fix for this for the QN, since I assume that this assumption might become a problem in the future. This requires some small redesigning. If it is an easy fix to add a sample to the start of the time window it will probably be faster to do that and then merge this pull request. I can add the fix in the QN later as a refactoring, which makes the QN method more robust.

src/acceleration/BaseQNAcceleration.hpp
BenjaminRodenberg: > IMO moving the non-reduced variant into a separate PR sounds unnecessarily complicated.

The additional effort for factoring out the non-reduced QNWI should be negligible. My point here is that the non-reduced variant does not seem like a must-have to me and generally smaller PRs are easier to integrate, review and test than big ones. (This was also my motivation for opening https://github.com/precice/precice/pull/2107 and https://github.com/precice/precice/pull/2108)

At the moment there seems to be an integration problem with the current state of develop. Manual testing before this problem is fixed looks to me like a waste of time. We can discuss the right approach at the coding days but my opinion here is still that it makes more sense to reduce the feature set of this PR to get it merged soon. With all the things going on on develop at the moment regarding dynamic meshes we will get more work the longer we wait.

The scripts for running the tests are here: https://github.com/BenjaminRodenberg/tutorials/blob/dissertation-partitioned-heat-conduction-v3/partitioned-heat-conduction/doAccelerationStudy.py. Extending the scripts and running the tests is simple, but still requires one or two hours of work. I can also do this at the coding days. But as long as the PR is not final, I don't want to spend time here.

BenjaminRodenberg:
It seems like some errors were caused by a merge conflict that was not resolved correctly.

The run https://github.com/precice/precice/actions/runs/11631163365/job/32391622895 based on [493ffb4](https://github.com/precice/precice/pull/2005/commits/493ffb482da9b4fe5613befea0833ce65442db90) gives

```
The following tests FAILED:
Errors while running CTest
	  1 - precice.acceleration (Failed)
	 19 - precice.integration.QuasiNewton (Failed)
	 20 - precice.integration.Remeshing (Failed)
	 21 - precice.integration.Serial (Failed)
```

Using [b1dbaf9](https://github.com/precice/precice/pull/2005/commits/b1dbaf9ce69145f0f1064dc8b57e1fcf0b5c226f) the workflow https://github.com/precice/precice/actions/runs/11781717041/job/32815135262 gives:

```
Errors while running CTest
The following tests FAILED:
	 21 - precice.integration.Remeshing (Failed)
```

Sorry that the history is a bit messed up. Here is (the relevant part of) `git diff 493ffb4 b1dbaf9`:

```diff
diff --git a/src/acceleration/BaseQNAcceleration.cpp b/src/acceleration/BaseQNAcceleration.cpp
index a495fc729..c97c52237 100644
--- a/src/acceleration/BaseQNAcceleration.cpp
+++ b/src/acceleration/BaseQNAcceleration.cpp
@@ -634,10 +634,15 @@ void BaseQNAcceleration::initializeVectorsAndPreconditioner(const DataMap &cplDa
   PRECICE_ASSERT(_primaryValues.size() == 0, "Values already initialized, something is wrong with the QN method");
   PRECICE_ASSERT(_oldPrimaryValues.size() == 0, "Values already initialized, something is wrong with the QN method");
 
-  _values    = Eigen::VectorXd::Zero(dataSize);
-  _oldValues = Eigen::VectorXd::Zero(dataSize);
-  _oldXTilde = Eigen::VectorXd::Zero(dataSize);
-  _residuals = Eigen::VectorXd::Zero(dataSize);
+  _oldPrimaryXTilde    = Eigen::VectorXd::Zero(primaryDataSize);
+  _oldPrimaryResiduals = Eigen::VectorXd::Zero(primaryDataSize);
+  _primaryResiduals    = Eigen::VectorXd::Zero(primaryDataSize);
+  _primaryValues       = Eigen::VectorXd::Zero(primaryDataSize);
+  _oldPrimaryValues    = Eigen::VectorXd::Zero(primaryDataSize);
+  _values              = Eigen::VectorXd::Zero(dataSize);
+  _oldValues           = Eigen::VectorXd::Zero(dataSize);
+  _oldXTilde           = Eigen::VectorXd::Zero(dataSize);
+  _residuals           = Eigen::VectorXd::Zero(dataSize);
```

@NiklasKotarsky does this look ok to you?

BenjaminRodenberg:
The following tests are still failing:
```
tests/remeshing/parallel-implicit/acceleration/IQN-ILS/RemeshBothSerial.cpp
tests/remeshing/parallel-implicit/acceleration/IQN-ILS/RemeshFirstSerial.cpp
tests/remeshing/parallel-implicit/acceleration/IQN-ILS/RemeshSecondSerial.cpp
tests/remeshing/parallel-implicit/acceleration/IQN-IMVJ/RemeshBothSerial.cpp
tests/remeshing/parallel-implicit/acceleration/IQN-IMVJ/RemeshFirstSerial.cpp
tests/remeshing/parallel-implicit/acceleration/IQN-IMVJ/RemeshSecondSerial.cpp
```

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -629,11 +629,6 @@ void BaseQNAcceleration::initializeVectorsAndPreconditioner(const DataMap &cplDa
   // Size of values
   const size_t dataSize = std::accumulate(_dataIDs.begin(), _dataIDs.end(), (size_t) 0, [&](size_t sum, int id) { return addTimeSliceSize(sum, id, _timeGrids); });
 
-  PRECICE_ASSERT(_values.size() == 0, "Values already initialized, something is wrong with the QN method");
-  PRECICE_ASSERT(_oldValues.size() == 0, "Values already initialized, something is wrong with the QN method");
-  PRECICE_ASSERT(_primaryValues.size() == 0, "Values already initialized, something is wrong with the QN method");
-  PRECICE_ASSERT(_oldPrimaryValues.size() == 0, "Values already initialized, something is wrong with the QN method");
-
```
fsimonis, line 614: The remeshing is now a valid case where initialize can be called again, which in turn should (re)set all required fields.

Thus, these assertions aren't needed anymore.

BenjaminRodenberg:
image

Did some experiments comparing the reduced and non-reduced variant. $\mu$ describes the relationship of substeps of first and second participant. I.e., $\mu = 1/5$ means first participant uses $1$ substep per window and the second uses $5$ substeps.

The highlighted lines are identical. This should be due to the serial coupling scheme that we use (data of second participant is used for acceleration and second participant uses only a single substep). There is no clear winner but rQN mostly uses slightly fewer iterations.

NiklasKotarsky:
> `tests/remeshing/parallel-implicit/acceleration/IQN-ILS/RemeshBothSerial.cpp tests/remeshing/parallel-implicit/acceleration/IQN-ILS/RemeshFirstSerial.cpp tests/remeshing/parallel-implicit/acceleration/IQN-ILS/RemeshSecondSerial.cpp tests/remeshing/parallel-implicit/acceleration/IQN-IMVJ/RemeshBothSerial.cpp tests/remeshing/parallel-implicit/acceleration/IQN-IMVJ/RemeshFirstSerial.cpp tests/remeshing/parallel-implicit/acceleration/IQN-IMVJ/RemeshSecondSerial.cpp`

Yes, I have fixed these test now by making sure that the time grids only contain the last time step when were not subcycling.

src/acceleration/BaseQNAcceleration.hpp
```diff
@@ -203,6 +205,9 @@ class BaseQNAcceleration : public Acceleration {
   /// @brief Stores x tilde deltas, where x tilde are values computed by solvers.
   Eigen::MatrixXd _matrixW;
 
+  /// @brief  if _reduced = false uses the full QN-WI and if _reduced = true uses rQN-WI form the paper https://onlinelibrary.wiley.com/doi/10.1002/nme.6443
+  const bool _reduced;
```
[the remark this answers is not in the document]
NiklasKotarsky: reducedTimeGrid sounds like a good name.

src/acceleration/BaseQNAcceleration.hpp
```diff
@@ -256,7 +261,7 @@ class BaseQNAcceleration : public Acceleration {
   virtual void updateDifferenceMatrices(const DataMap &cplData);
 
   /// Splits up QN system vector back into the coupling data
-  virtual void splitCouplingData(const DataMap &cplData);
+  virtual void applyQNValuesToCouplingData(const DataMap &cplData);
```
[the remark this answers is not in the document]
NiklasKotarsky: This renaming was to differentiate it from splitCouplingData, but this is no longer necessary since it is the only place we use splitCouplingData.

src/acceleration/BaseQNAcceleration.hpp
```diff
@@ -287,6 +289,17 @@ class BaseQNAcceleration : public Acceleration {
    */
   virtual void specializedInitializeVectorsAndPreconditioner(const DataMap &cplData) = 0;
 
+  /// @brief Samples and concatenates the data and old data in cplData into a long vector
+  void concatenateCouplingDataWaveform(Eigen::VectorXd &data, Eigen::VectorXd &oldData, const DataMap &cplData, std::vector<int> dataIDs, impl::WaveformTimeGrids timeGrids) const;
```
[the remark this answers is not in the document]
NiklasKotarsky: This was done to differentiate between the normal concatenateCouplingData that lives the Aitken acceleration, but I can remove the addition to the name since they are in two different files.

src/acceleration/BaseQNAcceleration.hpp
```diff
@@ -287,6 +289,17 @@ class BaseQNAcceleration : public Acceleration {
...
+
+  /// @brief List of the time grid to which all the data will be interpolated to
+  /// Stored in a map, since each data entry has its own time grid
+  impl::WaveformTimeGrids _timeGrids;
+
+  /// @brief List of the time grid to which all the primary data used in the QN system will be interpolated to
+  /// Stored in a map, since each data entry has its own time grid
+  impl::WaveformTimeGrids _primaryTimeGrids;
```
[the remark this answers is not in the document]
NiklasKotarsky: The naming follows the same naming convention we have for _values and _primaryValues. For rQNWR the matrix V_k and the residual use the last time step. We still need the full time grid of both primary and secondary values to construct the W_k matrix and values. The big reason for the two time grids is to be able to toggle between rQNWR and QNWR easily.

NiklasKotarsky:
> @BenjaminRodenberg Thanks for the results in [#2005 (comment)](https://github.com/precice/precice/pull/2005#issuecomment-2469017852)
> 
> How do those compare to the original QNWI paper?

I compared these with the results from table 1 in https://onlinelibrary.wiley.com/doi/epdf/10.1002/nme.6443 and the number of iterations are close with the results being less than 1 iteration apart in all cases. However, the full QNWR from this pull request requires consistently fewer iterations to converge. This discrepancy is best explained by the change in termination condition as discussed in comment https://github.com/precice/precice/pull/2005#issuecomment-2425100108 .

fsimonis:
@NiklasKotarsky I trust that the reviews from @BenjaminRodenberg and @uekerman ironed out the functional problems by now. So, I'll do a technical review.

fsimonis (commented):
My first impression is that this is pretty much ready. I have some tweaks and questions in some places, but no major changes.

The test naming is … very traditional :wink:
No issue though. This can be tidied up later.

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -588,41 +590,68 @@ void BaseQNAcceleration::writeInfo(
...
-    PRECICE_ASSERT(targetValues.size() >= offset + size, "Target vector was not initialized.", targetValues.size(), offset + size);
-    PRECICE_ASSERT(targetOldValues.size() >= offset + size, "Target vector was not initialized.");
-    for (Eigen::Index i = 0; i < size; i++) {
-      targetValues(i + offset)    = values(i);
-      targetOldValues(i + offset) = oldValues(i);
+  for (int id : dataIDs) {
+    Eigen::Index    dataSize = cplData.at(id)->getSize();
+    Eigen::VectorXd timeGrid = timeGrids.getTimeGrid(id);
+
+    // Skip the first sample, since it contains the initial data
```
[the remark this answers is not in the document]
NiklasKotarsky: Nice, catch it was easier to not do that.

src/acceleration/BaseQNAcceleration.hpp
```diff
@@ -70,7 +71,8 @@ class BaseQNAcceleration : public Acceleration {
       int                     filter,
       double                  singularityLimit,
       std::vector<int>        dataIDs,
-      impl::PtrPreconditioner preconditioner);
+      impl::PtrPreconditioner preconditioner,
+      bool                    reducedTimeGrid = true);
```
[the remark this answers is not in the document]
NiklasKotarsky: Thanks for the hint, I noticed that I had messed up one of the acceleration test when I removed the default.

src/time/TimeGrids.cpp
```diff
@@ -0,0 +1,57 @@
...
+      // Only way to access the first time stamp is through the whole vector
+      Eigen::VectorXd newtimeGrid = cplData.at(dataID)->timeStepsStorage().getTimes();
+      double          newTimesMin = newtimeGrid(0);
+      double          newTimesMax = newtimeGrid(newtimeGrid.size() - 1);
+
+      Eigen::VectorXd timeGrid    = pair.second;
+      double          oldTimesMin = _timeGrids.at(dataID)(0);
+      double          oldTimesMax = _timeGrids.at(dataID)(timeGrid.size() - 1);
+
+      // transform the time to the new time grid
```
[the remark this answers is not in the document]
NiklasKotarsky: I have added an explanation here of what this does. I decided to add the reason of why we need it to the moveTimeGridToNewWindow, since this is the main thing that this function does.

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -348,7 +352,9 @@ void BaseQNAcceleration::performAcceleration(
         "filter or increase its threshold (larger epsilon).");
   }
 
-  splitCouplingData(cplData); // split the primary and secondary coupling data back into the individual data objects
+  // Apply the quasi-Newton update to the waveform
+  splitCouplingData(cplData);
```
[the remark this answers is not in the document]
NiklasKotarsky: For me the purpose has not changed, since we still need to split the _values back into the coupling data. The difference is that we now also update the waveform in the coupling data as well. I have updated the comment to something better reflect whats actually going on here. 
 
After some thought I changed the name to updateCouplingData. This sounds a little bit better in my opinion, since I am technically also changing the time grid of the waveforms in cplData. I am however, open for better suggestions, since this is still not very descriptive.

src/time/TimeGrids.cpp
```diff
@@ -0,0 +1,57 @@
...
+#include <Eigen/Core>
+#include <numeric>
+#include <vector>
+
+#include "TimeGrids.hpp"
+#include "utils/assertion.hpp"
+
+namespace precice::time {
+
+void TimeGrids::saveTimeGridOfCouplingData(const DataMap &cplData, std::vector<int> dataIDs, bool reducedTimeGrid)
```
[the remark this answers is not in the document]
NiklasKotarsky: The key of the cplData includes both primary and secondary data. I added the dataIDs parameter to be able to only save the time grids in primary data in _primaryTimeGrids instance, but I can remove this if it simplifes the code.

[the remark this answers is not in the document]
NiklasKotarsky: I have the saveTimeGridOfCouplingData to become the constructor of this class. To do this I also changed the _timeGrids and _primaryTimeGrids to unique pointers. It would be nice to get some feedback on this, since I am not that comfortable with unique pointers.

NiklasKotarsky:
Thanks for the reviews @uekerman and @fsimonis! I have gone through the pull request now and fixed the comments. If you do not have anything more that needs to be fixed we should be able to merge soon.

src/acceleration/BaseQNAcceleration.cpp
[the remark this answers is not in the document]
fsimonis: In this case vaguer is more correct. So it's a decent choice. :+1:

fsimonis:
@NiklasKotarsky don't worry about the macOS workflow. Something is going wrong since they updated the runner image.

fsimonis:
@NiklasKotarsky Can you reply to https://github.com/precice/precice/pull/2005#discussion_r1854055119 and trigger me afterwards?

Once that is done, I'll merge. :+1:

src/acceleration/BaseQNAcceleration.hpp
```diff
@@ -287,6 +289,16 @@ class BaseQNAcceleration : public Acceleration {
...
 
+  /// @brief Samples and concatenates the data and old data in cplData into a long vector
+  void concatenateCouplingData(Eigen::VectorXd &data, Eigen::VectorXd &oldData, const DataMap &cplData, std::vector<int> dataIDs, std::unique_ptr<precice::time::TimeGrids> &timeGrids) const;
+
+  /// @brief Stores the time grids to which the primary and secondary data involved in the QN system will be interpolated to.
+  std::unique_ptr<time::TimeGrids> _timeGrids;
+
+  /// @brief Stores the time grids to which the primary data involved in the QN system will be interpolated to.
+  /// If _reducedTimeGrids is true then this will only contain the last time stamp of the time window, see https://doi.org/10.1002/nme.6443
+  std::unique_ptr<time::TimeGrids> _primaryTimeGrids;
```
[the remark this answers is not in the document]
NiklasKotarsky: Thanks for the suggestion. I decided to switch it to std::optional. In my opinion this reflects my use case a little bit better, since the timeGrid objects are reinitialized when we remesh.

fsimonis:
@NiklasKotarsky Thanks for the contribution :tada:

PR #2006, opened 2024-04-18
Add secondary data to exsting QN tests

Fujikawas:
## Main changes of this PR
One data set is added to the integrated QN tests and it is supposed to be accelerated as secondary data. 

Among the Quasi-Newton integration tests, those with `ILS` acceleration could pass and those with `IMVJ` acceleration would fail since there is no proper acceleration to the secondary data.

To increase test coverage of secondary data in QN methods as mentioned in issue #1996.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/tests/quasi-newton/helpers.cpp
+++ b/tests/quasi-newton/helpers.cpp
@@ -8,17 +8,21 @@
 /// tests for different QN settings if correct fixed point is reached
 void runTestQN(std::string const &config, TestContext const &context)
 {
-  std::string meshName, writeDataName, readDataName;
+  std::string meshName, writeDataName1, writeDataName2, readDataName1, readDataName2;
 
   if (context.isNamed("SolverOne")) {
-    meshName      = "MeshOne";
-    writeDataName = "Data1";
-    readDataName  = "Data2";
+    meshName       = "MeshOne";
+    writeDataName1 = "Data11";
+    writeDataName2 = "Data12";
+    readDataName1  = "Data21";
+    readDataName2  = "Data22";
   } else {
     BOOST_REQUIRE(context.isNamed("SolverTwo"));
-    meshName      = "MeshTwo";
-    writeDataName = "Data2";
-    readDataName  = "Data1";
+    meshName       = "MeshTwo";
+    writeDataName1 = "Data21";
+    writeDataName2 = "Data22";
+    readDataName1  = "Data11";
+    readDataName2  = "Data12";
   }
 
   precice::Participant interface(context.name, config, context.rank, context.size);
@@ -45,8 +49,10 @@ void runTestQN(std::string const &config, TestContext const &context)
   }
 
   interface.initialize();
-  double inValues[4]  = {0.0, 0.0, 0.0, 0.0};
-  double outValues[4] = {0.0, 0.0, 0.0, 0.0};
+  double inValues1[4]  = {0.0, 0.0, 0.0, 0.0};
+  double inValues2[4]  = {0.0, 0.0, 0.0, 0.0};
+  double outValues1[4] = {0.0, 0.0, 0.0, 0.0};
+  double outValues2[4] = {0.0, 0.0, 0.0, 0.0};
 
   int iterations = 0;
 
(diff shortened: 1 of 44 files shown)


tests/quasi-newton/parallel/TestQN2.xml
```diff
@@ -42,14 +52,17 @@
     <participants first="SolverOne" second="SolverTwo" />
     <max-time-windows value="1" />
     <time-window-size value="1.0" />
-    <exchange data="Data1" mesh="MeshOne" from="SolverOne" to="SolverTwo" substeps="false" />
-    <exchange data="Data2" mesh="MeshOne" from="SolverTwo" to="SolverOne" substeps="false" />
+    <exchange data="Data11" mesh="MeshOne" from="SolverOne" to="SolverTwo" />
+    <exchange data="Data12" mesh="MeshOne" from="SolverOne" to="SolverTwo" />
+    <exchange data="Data21" mesh="MeshOne" from="SolverTwo" to="SolverOne" substeps="false" />
+    <exchange data="Data22" mesh="MeshOne" from="SolverTwo" to="SolverOne" substeps="false" />
```
uekerman, line 58:
```suggestion
    <exchange data="Data11" mesh="MeshOne" from="SolverOne" to="SolverTwo" substeps="false" />
    <exchange data="Data12" mesh="MeshOne" from="SolverOne" to="SolverTwo" substeps="false" />
    <exchange data="Data21" mesh="MeshOne" from="SolverTwo" to="SolverOne" substeps="false" />
    <exchange data="Data22" mesh="MeshOne" from="SolverTwo" to="SolverOne" substeps="false" />
```
There might be more problems like this. It defaults to `false`, so no real problem, but for readability, we should aim for consistency here.

tests/quasi-newton/helpers.cpp
```diff
@@ -8,17 +8,21 @@
...
 
   if (context.isNamed("SolverOne")) {
-    meshName      = "MeshOne";
-    writeDataName = "Data1";
-    readDataName  = "Data2";
+    meshName       = "MeshOne";
+    writeDataName1 = "Data11";
+    writeDataName2 = "Data12";
+    readDataName1  = "Data21";
+    readDataName2  = "Data22";
```
uekerman, line 18:
```suggestion
    writePrimaryDataName = "PrimaryData1";
    writeSecondaryDataName = "SecondaryData1";
    readPrimaryDataName  = "PrimaryData2";
    readSecondaryName = "SecondaryData2";
```
I think such renaming could makes things easier to grasp.

tests/quasi-newton/helpers.cpp
```diff
@@ -45,8 +49,10 @@ void runTestQN(std::string const &config, TestContext const &context)
   }
 
   interface.initialize();
-  double inValues[4]  = {0.0, 0.0, 0.0, 0.0};
-  double outValues[4] = {0.0, 0.0, 0.0, 0.0};
+  double inValues1[4]  = {0.0, 0.0, 0.0, 0.0};
+  double inValues2[4]  = {0.0, 0.0, 0.0, 0.0};
+  double outValues1[4] = {0.0, 0.0, 0.0, 0.0};
+  double outValues2[4] = {0.0, 0.0, 0.0, 0.0};
```
uekerman, line 55:
```suggestion
  double inPrimaryValues[4]  = {0.0, 0.0, 0.0, 0.0};
  double inSecondaryValues[4]  = {0.0, 0.0, 0.0, 0.0};
  double outPrimaryValues[4] = {0.0, 0.0, 0.0, 0.0};
  double outSecondaryValues[4] = {0.0, 0.0, 0.0, 0.0};
```

tests/quasi-newton/helpers.cpp
```diff
@@ -70,16 +77,22 @@ void runTestQN(std::string const &config, TestContext const &context)
```
uekerman, line 77: Could you please add here above some documentation on how primary and secondary data here is handled.

uekerman (changes requested):
Looks good in general. Thanks for the push :+1: 

We no longer have any tests that have no secondary data now. But we could add use all data as primary. You could copy from how things are done here: https://github.com/precice/precice/blob/develop/tests/serial/mapping-nearest-projection/MappingNearestProjectionEdges.cpp

Fujikawas:
> We no longer have any tests that have no secondary data now. 

The `parallel/TestQNXemptyPartition` tests are still (mostly) with only primary data.

tests/quasi-newton/helpers.cpp
Fujikawas: In most of the tests we use serial coupling, so the data written by `SolverOne` are neither primary nor secondary. Should we still use this renaming or just describe them in the comment?

uekerman (approved):

Issue #2004, opened 2024-04-18
Required preCICE version in the preCICE configuration file


MakisH:
At the moment, when a user tries to use a preCICE v2 configuration file with preCICE v3 (or the reverse), they will just get errors regarding unknown configuration tags.

Since we keep the config compatible in the same preCICE major version, but then also often change it with each major version, I think it would be very helpful to add a `<precice-version major="3">` tag, which we could then check and throw a proper error.

We could even extend that to something like `<precice-version major="3" minor="5">` to signify that a feature mentioned in this config requires preCICE v3, but at least v3.5.

Retrospectively, we could implement a simple error message that looks for old keywords and complains that `This configuration file was probably made for preCICE v1, but you are using preCICE v3.2.0`.

Questions:

- Would something like this be helpful?
- Would adding such a tag be feasible?
- Would adding such a heuristic for older versions be feasible?

MakisH:
We now have such a check at least for major versions. Using a v2 config file with v3 leads to:

```
ERROR: This configuration contains the tag <solver-interface>, meaning it was created for a preCICE version prior to version 3. Are you using the correct version of your simulation case? Has this simulation case been updated to this version of preCICE?
```

At the same time, looking at the headers in OpenFOAM configuration files and how they develop over time, specifying a version in a config file might end up being quite confusing for users. All this provided that users still write config files manually.

After we start auto-generating config files in practice, this might be more useful.

ankitv06:
Hi @MakisH, I'm interested in contributing to this. Given the heuristic detection already in place for major versions using tags identification like `<solver-interface>`, do you still see value in implementing the `<precice-version>` tag, or would you prefer to wait until config auto-generation is in place? Happy to work on either this or an alternative approach if you have something in mind.

PR #2007, opened 2024-04-18
Add simple version check

fsimonis:
## Main changes of this PR

This PR adds a simple version check to detect configurations for preCICE v1 and v2.
The parser fails with the following message if it encounters a tag called `<solver-interface`:

```console
$ precice-tools check v2/solverdummy.xml
Checking v2/solverdummy.xml for syntax and basic setup issues...
precice-tools: ERROR: This configuration contains the tag <solver-interface>, meaning it was written for a preCICE version 1 or 2. Please check that you are using the correct case or preCICE version.
```

v1 and v2 configurations fail in the ConfigParser as it encounters an unknown tag `solver-interface`.
This is a very obscure error though and doesn't highlight the actual problem that the config is for an older verison.

This change is a naive implementation that should be useful enough.

Related to #2004

## Author's checklist

### DIFF
--- a/src/xml/ConfigParser.cpp
+++ b/src/xml/ConfigParser.cpp
@@ -215,7 +215,7 @@ void ConfigParser::connectTags(const ConfigurationContext &context, std::vector<
   for (auto &subtag : SubTags) {
     std::string expectedName = (subtag->m_Prefix.length() ? subtag->m_Prefix + ":" : "") + subtag->m_Name;
     PRECICE_CHECK(expectedName != "solver-interface",
-                  "This configuration contains the tag <solver-interface>, meaning it was written for a preCICE version 1 or 2. "
+                  "This configuration contains the tag <solver-interface>, meaning it was created for a preCICE version prior to version 3. "
                   "Please check that you are using the correct case or preCICE version.");
     const auto tagPosition = std::find_if(
         DefTags.begin(),
--- a/src/xml/ConfigParser.cpp
+++ b/src/xml/ConfigParser.cpp
@@ -216,7 +216,7 @@ void ConfigParser::connectTags(const ConfigurationContext &context, std::vector<
     std::string expectedName = (subtag->m_Prefix.length() ? subtag->m_Prefix + ":" : "") + subtag->m_Name;
     PRECICE_CHECK(expectedName != "solver-interface",
                   "This configuration contains the tag <solver-interface>, meaning it was created for a preCICE version prior to version 3. "
-                  "Please check that you are using the correct case or preCICE version.");
+                  "Are you using the correct version of your simulation case? Has this simulation case been updated to this version of preCICE?");
     const auto tagPosition = std::find_if(
         DefTags.begin(),
         DefTags.end(),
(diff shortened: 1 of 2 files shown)


src/xml/ConfigParser.cpp
```diff
@@ -214,7 +214,10 @@ void ConfigParser::connectTags(const ConfigurationContext &context, std::vector<
 
   for (auto &subtag : SubTags) {
     std::string expectedName = (subtag->m_Prefix.length() ? subtag->m_Prefix + ":" : "") + subtag->m_Name;
-    const auto  tagPosition  = std::find_if(
+    PRECICE_CHECK(expectedName != "solver-interface",
+                  "This configuration contains the tag <solver-interface>, meaning it was written for a preCICE version 1 or 2. "
```
MakisH, line 218: I think that "preCICE version 1" is clear that it includes all of 1.x. The "a preCICE version 1 or 2" sounds strange to me.

```suggestion
                  "This configuration contains the tag <solver-interface>, meaning it was written for preCICE version 1 or 2. "
```

MakisH (approved):
Thank you for the quick reaction! This should cover most of the issues already, and showcase how we can add similar checks in the future.

Let's include it in the next bugfix or feature release.

I have not tested it, but I trust you did.

src/xml/ConfigParser.cpp
```diff
@@ -214,7 +214,10 @@ void ConfigParser::connectTags(const ConfigurationContext &context, std::vector<
...
+                  "Please check that you are using the correct case or preCICE version.");
```
MakisH, line 219:
```suggestion
                  "Please check that you are using the intended configuration file and preCICE version.");
```

src/xml/ConfigParser.cpp
fsimonis: Alternative phrasing could be "this configuration compatible with versions prior to version 3"

MakisH: Both are fine and clear.

fsimonis:
```suggestion
                  "This configuration contains the tag <solver-interface>, meaning it was created for a preCICE version prior to version 3. "
```

src/xml/ConfigParser.cpp
fsimonis: Let's tailor this part explicitly to a new user, who mistakenly uses the incorrect version of the tutorials / quickstart.

* Please check that the expected version of your simulation case matches this version of preCICE.

* Please ensure that the used simulation case was made for this version of preCICE.

* Please ensure that the used simulation case is compatible with this version of preCICE.

* Are you using the correct version of your simulation case? Has this simulation case been updated to this version of preCICE?

fsimonis:
```suggestion
                  "Are you using the correct version of your simulation case? Has this simulation case been updated to this version of preCICE?");
```

MakisH (approved):
I like how the error message asks questions!

PR #2010, opened 2024-04-21
Use combined W matrix for primary and secondary data in IQN

Fujikawas:
## Main changes of this PR
The secondary data is handled together with primary data in the `W` matrix both in IQN-ILS and IQN-IMVJ method.

Relating to #1996 

## To do

* [x] Check if the `secondaryDataIDs` and `_values` with only primary data are still needed; -> Yes, for computing `V`, warn zero update of `V`;
* [x] The preconditioning of `Wtil` should still be done, currently commented out;
* [x] With the tests with secondary data in #2006, this method could work. Still the QNTest9EmptyPartition which is to test SVD-restart failed. Other tests with SVD could work.

## Author's checklist

## Reviewers' checklist


src/acceleration/BaseQNAcceleration.hpp
```diff
@@ -161,6 +161,9 @@ class BaseQNAcceleration : public Acceleration {
   /// Data IDs of data not involved in IQN coefficient computation.
   std::vector<int> _secondaryDataIDs;
 
+  /// Data IDs of all coupling data.
+  std::vector<int> _cplDataIDs;
+
```
uekerman, line 166: Is `_dataIDs` then only primary data?
If yes, I would suggest to rename: `_dataIDs` -> `_primaryDataIDs` and `cplDataIDs` -> `_dataIDs`.

src/acceleration/BaseQNAcceleration.hpp
```diff
@@ -274,6 +277,12 @@ class BaseQNAcceleration : public Acceleration {
   /// @brief Concatenation of all (old) coupling data involved in the QN system.
   Eigen::VectorXd _oldValues;
 
+  /// @brief Concatenation of all primary and secondary data involved in the QN system.
+  Eigen::VectorXd _cplValues;
+
+  /// @brief Concatenation of all old primary and secondary data involved in the QN system.
+  Eigen::VectorXd _oldCplValues;
+
```
uekerman, line 285: We could also rename here to make things more explicit, see above.

uekerman (commented):
Looks promising!
I guess the SVD changes could go to a separate refactoring PR.

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -95,25 +95,41 @@ void BaseQNAcceleration::initialize(
                   "Quasi-Newton acceleration does not yet support using data from all substeps. Please set substeps=\"false\" in the exchange tag of data \"{}\".", data->getDataName());
   }
 
-  size_t              entries = 0;
+  size_t              entries     = 0;
+  size_t              cplDataSize = 0;
```
uekerman, line 99:
```suggestion
  size_t              primaryDataSize = 0;
  size_t              dataSize = 0;
```

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -149,24 +166,15 @@ void BaseQNAcceleration::initialize(
     }
 
     // test that the computed number of unknown per proc equals the number of entries actually present on that proc
-    size_t unknowns = _dimOffsets[utils::IntraComm::getRank() + 1] - _dimOffsets[utils::IntraComm::getRank()];
-    PRECICE_ASSERT(entries == unknowns, entries, unknowns);
+    // size_t unknowns = _dimOffsets[utils::IntraComm::getRank() + 1] - _dimOffsets[utils::IntraComm::getRank()];
+    // PRECICE_ASSERT(entries == unknowns, entries, unknowns); //?
```
uekerman, line 170: This is a tricky point. Should `_dimOffsets` include secondary data or not? Probably makes a difference for IMVJ. Let's try to also add a test with secondary data that runs in parallel.

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -275,18 +286,19 @@ void BaseQNAcceleration::updateDifferenceMatrices(
...
-  PRECICE_ASSERT(_oldValues.size() == _oldXTilde.size(), _oldValues.size(), _oldXTilde.size());
-  PRECICE_ASSERT(_residuals.size() == _oldXTilde.size(), _residuals.size(), _oldXTilde.size());
+  PRECICE_ASSERT(_oldPrimaryResiduals.size() == _oldPrimaryXTilde.size(), _oldPrimaryResiduals.size(), _oldPrimaryXTilde.size());
+  PRECICE_ASSERT(_primaryValues.size() == _oldPrimaryXTilde.size(), _primaryValues.size(), _oldPrimaryXTilde.size());
+  PRECICE_ASSERT(_oldPrimaryValues.size() == _oldPrimaryXTilde.size(), _oldPrimaryValues.size(), _oldPrimaryXTilde.size());
+  PRECICE_ASSERT(_primaryResiduals.size() == _oldPrimaryXTilde.size(), _primaryResiduals.size(), _oldPrimaryXTilde.size());
 
   // assume data structures associated with the LS system can be updated easily.
 
   // scale data values (and secondary data values)
```
uekerman, line 300: Is this comment correct? I don't think that there is any scaling in there.

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -298,16 +310,15 @@ void BaseQNAcceleration::performAcceleration(
...
+    _oldPrimaryXTilde    = _primaryValues;    // Store x tilde of primary data
+    _oldXTilde           = _values;           // Store x tilde of primary and secondary data
+    _oldPrimaryResiduals = _primaryResiduals; // Store current residual of primary data
 
     // Perform constant relaxation
     // with residual: x_new = x_old + omega * res
     _residuals *= _initialRelaxation;
-    _residuals += _oldValues;
-    _values = _residuals;
+    _values = _oldValues + _residuals;
```
uekerman, line 320: Is this correct? Now `_residuals` has a different value.

src/acceleration/impl/ParallelMatrixOperations.hpp
```diff
@@ -136,7 +136,6 @@ class ParallelMatrixOperations {
 
     PRECICE_ASSERT(_needCyclicComm);
     PRECICE_ASSERT(leftMatrix.cols() == q, leftMatrix.cols(), q);
-    PRECICE_ASSERT(leftMatrix.rows() == rightMatrix.cols(), leftMatrix.rows(), rightMatrix.cols());
```
uekerman, line 139: Removing this doesn't feel good.

src/acceleration/impl/Preconditioner.hpp
```diff
@@ -58,16 +58,22 @@ class Preconditioner {
   {
     PRECICE_TRACE();
     if (transpose) {
-      PRECICE_ASSERT(M.cols() == (int) _weights.size(), M.cols(), _weights.size());
-      for (int i = 0; i < M.cols(); i++) {
+      PRECICE_WARN_IF(M.cols() < (int) _weights.size(), "Matrix and weights size mismatch, only part of the preconditioning weights will be used.");
+      PRECICE_WARN_IF(M.cols() > (int) _weights.size(), "Matrix and weights size mismatch, only part of the matrix will be preconditioned.");
```
uekerman, line 62: Let's find a better solution here. Such warnings will confuse 99% of users.

src/acceleration/impl/Preconditioner.hpp
uekerman, line 1: Could we add unit tests to cover these changes?

src/acceleration/impl/SVDFactorization.hpp
uekerman, line 1: I still think it would be better to do these changes in a separate PR before this one. And add unit tests for the new functionality.

uekerman (changes requested):
Nice to see how these changes reduce code duplication :+1:

Still a bit of work to do, especially for parallel runs I guess, see below.

Requires a changelog as we now have support of secondary data for IMVJ

src/acceleration/BaseQNAcceleration.cpp
Fujikawas: From the test results I didn't see difference. If you mean that the `_residuals` is not updated here, I see that the `_residuals` would be calculated from `_values` and `_oldValues` by updating differences and is not stored like `_values`, so the step here still works.

src/acceleration/BaseQNAcceleration.cpp
Fujikawas: Only concatenate happens until the next comment

src/acceleration/BaseQNAcceleration.hpp
```diff
@@ -90,7 +90,7 @@ class BaseQNAcceleration : public Acceleration {
    */
   virtual std::vector<int> getDataIDs() const
   {
-    return _dataIDs;
+    return _primaryDataIDs;
   }
```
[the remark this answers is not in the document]
Fujikawas: For the whole acceleration method? Other under-relaxation methods also use this function

src/acceleration/impl/ParallelMatrixOperations.hpp
Fujikawas: In the comment it says that ` For multiplication W_til * Z = J, p = r = n_global, q = m`, which is not the case now. I'm not so sure why this assertion is necessary for the operation unless these two variables are mixed used. Need to check.

src/acceleration/BaseQNAcceleration.cpp
Fujikawas: Since it's used to initialize the size of `Wtil`, I would say that the `_dimOffsets` should include the secondary data. But I'm a little confused with this function now: https://github.com/precice/precice/blob/8458db5189c7f4c9c41551cac55eb957ce46af80/src/acceleration/BaseQNAcceleration.cpp#L626-L632
In the extended tests, the parallel tests with IMVJ include also secondary data. They could work with this implementation here(secondary data included in the `_dimOffsets`). This assert should be modified a bit then.

src/acceleration/BaseQNAcceleration.hpp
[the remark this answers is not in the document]
uekerman: For Aitken, we in principle also have the primary vs secondary story.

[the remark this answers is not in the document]
uekerman: Why not `getPrimaryDataIDs()`? Otherwise, this might be confusing. And makes sense for all acceleration methods, no?

src/acceleration/BaseQNAcceleration.cpp
uekerman: But then the comment is misleading, no?

src/acceleration/test/ParallelMatrixOperationsTest.cpp
uekerman, line 1: Could it make sense to add more tests here?

uekerman (commented):
I did not check all implementation details, but I have a good feeling in general.
Would it make sense to also add unit tests?

src/acceleration/BaseQNAcceleration.cpp
uekerman: now, it says only `scale data values`, but there is still no scaling happening, no? :)

src/acceleration/test/ParallelMatrixOperationsTest.cpp
uekerman: looks good! please add a quick comment how you created the test

uekerman (approved):

PR #2014, opened 2024-04-29
Extend SVD to cover input multiplicators with different sizes

Fujikawas:
## Main changes of this PR
The SVD factorization could deal with two input vectors with different sizes. Tests are modified to test it.

It serves the #2010.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/acceleration/IQNIMVJAcceleration.cpp
+++ b/src/acceleration/IQNIMVJAcceleration.cpp
@@ -80,11 +80,6 @@ void IQNIMVJAcceleration::initialize(
   if (_imvjRestartType > 0)
     _imvjRestart = true;
 
-  // initialize parallel matrix-matrix operation module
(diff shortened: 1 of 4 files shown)


Fujikawas:
Do we want to extend the initialization that when only one size is given, same number are given to the two input sizes by default?

uekerman (approved):
Are the getters `rowsA`, `rowsB` actually needed? They seem to be not used in any test.

> Do we want to extend the initialization that when only one size is given, same number are given to the two input sizes by default?

As long as we have no use-case, I would say no. We don't want to implement a general-purpose SVD, but only what we need :blush:

Fujikawas:
> Are the getters `rowsA`, `rowsB` actually needed? They seem to be not used in any test.

They are not used for the current implementation. Since I see little chance to set a derived class for SVD, they could be removed.

PR #2019, opened 2024-05-08
Add sync events to measure time spend synchronizing

fsimonis:
## Main changes of this PR

This PR adds profiling events that measure the time waiting when using syncMode

This allows to correct events with synchronizing sub events and make the trace easier to follow.

[Example trace](https://github.com/precice/precice/files/15248241/trace.json.txt)

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/2019.md
+++ b/docs/changelog/2019.md
@@ -1,0 +1,1 @@
+- Add the synchronization of ranks to the profiling data if `<profiling synchronize="true" />`
--- a/src/profiling/Event.cpp
+++ b/src/profiling/Event.cpp
@@ -8,8 +8,12 @@ namespace precice::profiling {
 Event::Event(std::string_view eventName, Options options)
     : _fundamental(options.fundamental), _synchronize(options.synchronized)
 {
-  auto &er = EventRegistry::instance();
-  _eid     = er.nameToID(std::string(EventRegistry::instance().prefix).append(eventName));
+  auto &er   = EventRegistry::instance();
+  auto  name = std::string(EventRegistry::instance().prefix).append(eventName);
+  _eid       = er.nameToID(name);
+  if (_synchronize) {
+    _sid = er.nameToID(name + ".sync");
+  }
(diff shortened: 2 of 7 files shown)


davidscn (approved):
Looks good, a test would be good here as well. Only considerations:

- should we also report the wait time for events which are synchronized by construction, for example the advance call?
- in the example event you provided: what is the primary rank doing while every other rank is waiting for sendData.sync?
- is we expect users to make use of this as well: is `sync` really the correct expression? Maybe wait-for or idle?

fsimonis:
> should we also report the wait time for events which are synchronized by construction, for example the advance call?

Not sure I understand what you mean with "synchronized by construction".

> in the example event you provided: what is the primary rank doing while every other rank is waiting for sendData.sync?

Good question. I don't have an answer to that one. Definitely worth investigating.

> is we expect users to make use of this as well: is sync really the correct expression? Maybe wait-for or idle?

wait would be the better term, but sync is consistent with the configuration option.

davidscn:
> Not sure I understand what you mean with "synchronized by construction".

For example the `advance` call is synchronized across ranks, essentially a waitall.

fsimonis:
https://github.com/precice/precice/blob/2a8c421edfe8edc53945b656651df5a276b2f29a/src/precice/impl/ParticipantImpl.cpp#L353

Marks this event to be synchronized in case of `<profiling synchronize="1" />`.

https://github.com/precice/precice/blob/2a8c421edfe8edc53945b656651df5a276b2f29a/src/precice/impl/ParticipantImpl.cpp#L365-L370

Only synchronizes in debug builds.

After that, data communication and mappings decide what when to sync.

fsimonis:
> in the example event you provided: what is the primary rank doing while every other rank is waiting for sendData.sync?

This is an m2n event. So the data is sent from one participant to another. The sync is the time until all ranks are ready to send.

Merging now, we can still tweak later.

PR #2020, opened 2024-05-08
Relax assertions in preconditioner for ummatched weight and target sizes

Fujikawas:
## Main changes of this PR
The assertions in the preconditioner are relaxed to `debug` information to allow non-equal sizes of the weight and target to be scaled. This is the case for `Wtil` in #2010.  
Tests are added for this change.

Part of #2010.

## Author's checklist
* [ ] I sticked to CMake version 3.22.1.

## Reviewers' checklist

### DIFF
--- a/src/acceleration/impl/Preconditioner.hpp
+++ b/src/acceleration/impl/Preconditioner.hpp
@@ -58,7 +58,7 @@ public:
   {
     PRECICE_TRACE();
     if (transpose) {
-      PRECICE_DEBUG_IF(_weights.size() != M.cols(), "The number of columns of the matrix {} and weights size {} mismatched.", M.rows(), _weights.size());
+      PRECICE_DEBUG_IF((int) _weights.size() != M.cols(), "The number of columns of the matrix {} and weights size {} mismatched.", M.rows(), _weights.size());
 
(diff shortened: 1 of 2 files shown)


src/acceleration/impl/Preconditioner.hpp
```diff
@@ -58,16 +58,20 @@ class Preconditioner {
   {
     PRECICE_TRACE();
     if (transpose) {
-      PRECICE_ASSERT(M.cols() == (int) _weights.size(), M.cols(), _weights.size());
-      for (int i = 0; i < M.cols(); i++) {
+      PRECICE_DEBUG_IF((int) _weights.size() != M.cols(), "The number of columns of the matrix {} and weights size {} mismatched.", M.rows(), _weights.size());
```
davidscn, line 61:
```suggestion
      PRECICE_DEBUG_IF(static_cast<Eigen::Index>(_weights.size()) != M.cols(), "The number of columns of the matrix {} and weights size {} mismatched.", M.rows(), _weights.size());
```
We should avoid the C-style casting. I think clang-tidy should complain as well

uekerman (approved):

Issue #2013, opened 2024-04-23
Missing API Profiling


fsimonis:
**Please describe the problem you are trying to solve.**

Currently, only a subset of the API functions show up in the profiling, namely:

* construction
* initialize
* advance
* finalize

We further annotate the time spent in the solver.

* solver.initialize
* solver.advance

However, this time still includes preCICE API calls, which can be interesting for the user to know the runtime of.

* setMeshVertex / setMeshVertices
* setMesh<Connectivity> (Singular + Plural)
* readData (which may perform interpolation)
* writeData
* writeGradientData
* getVertexIDsAndCoordinates

Downside is that this will make already inefficient singular versions even slower. (setMeshVertex, single-vertex reads and writes)

**Describe the solution you propose.**

Add events to the API calls (after rough input validation).
Big upside of using names in the API calls is that we can now directly use them to name the events:

Example for `readData`: `Event e{fmt::format("readData.{}_{}", meshName, dataName)};`

**Describe alternatives you've considered**

Only profile internals. The API layer is mostly pretty cheap and non-allocating.

For example: `readData` can be a copy if the t exists, a copy if we extrapolate, or an expensive BSpline interpolation with possible computation. The interpolation and the computation are more interesting measurements than the API itself.

**Additional context**

The user can also be the author of bindings, who may want to profile extra functionality e.g. in readData.
I had to manually add these events for https://github.com/precice/python-bindings/issues/202

The attached trace-v3.json of [this discourse post](https://precice.discourse.group/t/pyprecice-slow-at-setting-up-secondary-communications/1889/12?) contains some of these events.

PR #2021, opened 2024-05-13
Improve the computational efficiency of the Bspline interpolation

NiklasKotarsky:
## Main changes of this PR

Change the matrix A which contains the time step information in the Bspline algorithm to sparse.

The current implementation of the Bspline algorithm uses a dense matrix as A containing the time steps, which results in a significant slow down a large number of time steps. This change utilizes the sparse nature of the matrix A and thus speeds up the considerably for a large number of time steps. 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/2021.md
+++ b/docs/changelog/2021.md
@@ -1,0 +1,1 @@
+- Improved the runtime of the creation of the Bspline interpolation, by switching the matrix A inside the Bspline class to a sparse matrix.
--- a/src/math/Bspline.cpp
+++ b/src/math/Bspline.cpp
@@ -3,6 +3,7 @@
 #include "utils/assertion.hpp"
 
 #include <Eigen/Core>
+#include <Eigen/Sparse>
 #include <algorithm>
 #include <cstdlib>
 #include <unsupported/Eigen/Splines>
@@ -28,27 +29,31 @@ Bspline::Bspline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree
   //The code for computing the knots and the control points is copied from Eigens bspline interpolation with minor modifications, https://gitlab.com/libeigen/eigen/-/blob/master/unsupported/Eigen/src/Splines/SplineFitting.h
 
   Eigen::KnotAveraging(ts, splineDegree, _knots);
-  Eigen::DenseIndex n = xs.cols();
-  Eigen::MatrixXd   A = Eigen::MatrixXd::Zero(n, n);
+  Eigen::DenseIndex                   n = xs.cols();
+  std::vector<Eigen::Triplet<double>> matrixEntries;
+  matrixEntries.reserve(n * splineDegree + 2);
 
   for (Eigen::DenseIndex i = 1; i < n - 1; ++i) {
-    //Attempt at hack... the spline dimension is not used here explicitly
-    const Eigen::DenseIndex span = Eigen::Spline<double, 1>::Span(ts[i], splineDegree, _knots);
+    const Eigen::DenseIndex span      = Eigen::Spline<double, 1>::Span(ts[i], splineDegree, _knots);
+    auto                    basisFunc = Eigen::Spline<double, 1>::BasisFunctions(ts[i], splineDegree, _knots);
 
-    // The segment call should somehow be told the spline order at compile time.
-    A.row(i).segment(span - splineDegree, splineDegree + 1) = Eigen::Spline<double, 1>::BasisFunctions(ts[i], splineDegree, _knots);
+    for (Eigen::DenseIndex j = 0; j < splineDegree + 1; ++j) {
+      matrixEntries.emplace_back(i, span - splineDegree + j, basisFunc(j));
+    }
   }
-  A(0, 0)         = 1.0;
-  A(n - 1, n - 1) = 1.0;
 
-  auto qr = A.householderQr();
+  matrixEntries.emplace_back(0, 0, 1.0);
+  matrixEntries.emplace_back(n - 1, n - 1, 1.0);
 
-  Eigen::MatrixXd controls = Eigen::MatrixXd::Zero(n, _ndofs);
+  Eigen::SparseMatrix<double> A(n, n);
+  A.setFromTriplets(matrixEntries.begin(), matrixEntries.end());
+  A.makeCompressed();
 
-  for (int i = 0; i < _ndofs; i++) {
-    controls.col(i) = qr.solve(xs.row(i).transpose());
-  }
-  _ctrls = std::move(controls);
+  Eigen::SparseQR<Eigen::SparseMatrix<double>, Eigen::COLAMDOrdering<int>> qr;
+  qr.analyzePattern(A);
+  qr.factorize(A);
+
+  _ctrls = qr.solve(xs.transpose());
 }
 
 Eigen::VectorXd Bspline::interpolateAt(double t) const


uekerman:
Looks interesting. Could you report on example performance data? How does it behave for only a few time steps per window?

BenjaminRodenberg (approved):
Changes look reasonable to me. @fsimonis do you have any readability advice? Looks ok to me, but maybe you have some nice ideas.

I also compared the old (`develop`) and new (#2021) implementation with the tutorial case I'm contributing via https://github.com/precice/tutorials/pull/500. If I use a large time window + strict tolerances for the time stepper, I get already observable overhead with 370 substeps for the `develop` version, with 1720 substeps the overhead gets pretty much unbearable for such a simple experiment. Comparing to the #2021 version I see a clear improvement. With 370 substeps there is no real overhead and with 1720 substeps the situation is much better.

I did not do a detailed profiling, but the positive effect of this PR is already quite visible for this simple case, if there are sufficiently many points.

NiklasKotarsky:
> Looks interesting. Could you report on example performance data? How does it behave for only a few time steps per window?

I have timed the old (develop) and new (https://github.com/precice/precice/pull/2021) implementation using the test ReadWriteScalarDataSubcyclingThird in test/serial/time/implicit/serial-coupling, which does 4 timesteps with a third order bspline interpolation. I changed the number of time windows to 100 to get a more representative picture and used the package chrono to time the while (precice.isCouplingOngoing()) loop. 

For the current version of preCICe develop I got a time of 26.6 seconds and my changes with a sparse matrix took 26.8 seconds. 

Changing the test to do 400 timesteps and 10 time windows instead the current version of preCICe develop gets a time of 3.6 seconds and my changes with a sparse matrix took 2.9 seconds.

src/math/Bspline.cpp
```diff
@@ -28,23 +29,29 @@ Bspline::Bspline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree
...
   for (Eigen::DenseIndex i = 1; i < n - 1; ++i) {
-    //Attempt at hack... the spline dimension is not used here explicitly
-    const Eigen::DenseIndex span = Eigen::Spline<double, 1>::Span(ts[i], splineDegree, _knots);
+    const Eigen::DenseIndex span      = Eigen::Spline<double, 1>::Span(ts[i], splineDegree, _knots);
+    auto                    basisFunc = Eigen::Spline<double, 1>::BasisFunctions(ts[i], splineDegree, _knots);
 
-    // The segment call should somehow be told the spline order at compile time.
-    A.row(i).segment(span - splineDegree, splineDegree + 1) = Eigen::Spline<double, 1>::BasisFunctions(ts[i], splineDegree, _knots);
+    for (Eigen::DenseIndex j = 0; j < splineDegree + 1; ++j) {
+      matrixEntries.emplace_back(Eigen::Triplet<double>(i, span - splineDegree + j, basisFunc(j)));
```
fsimonis, line 41: The constructor isn't _normally_ needed to emplace. This also applies for the elements below.

```cpp
      matrixEntries.emplace_back(i, span - splineDegree + j, basisFunc(j));
```

src/math/Bspline.cpp
```diff
@@ -28,23 +29,29 @@ Bspline::Bspline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree
   //The code for computing the knots and the control points is copied from Eigens bspline interpolation with minor modifications, https://gitlab.com/libeigen/eigen/-/blob/master/unsupported/Eigen/src/Splines/SplineFitting.h
 
   Eigen::KnotAveraging(ts, splineDegree, _knots);
-  Eigen::DenseIndex n = xs.cols();
-  Eigen::MatrixXd   A = Eigen::MatrixXd::Zero(n, n);
+  Eigen::DenseIndex           n = xs.cols();
+  Eigen::SparseMatrix<double> A(n, n);
```
fsimonis, line 33: Could you move this underneath the loop, to the place where you use A for the first time?

src/math/Bspline.cpp
```diff
@@ -28,23 +29,29 @@ Bspline::Bspline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree
...
 
+  std::vector<Eigen::Triplet<double>> matrixEntries;
```
fsimonis, line 35: Given that this is a big vector, and we know the amount of elements that will be added, I am positive that reserving space will make a significant performance difference as it prevent reallocations.

```cpp
matrixEntries.reserve( /* num elements from loop n*splineDegree? */);
```

src/math/Bspline.cpp
```diff
@@ -28,23 +29,29 @@ Bspline::Bspline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree
...
+    }
   }
-  A(0, 0)         = 1.0;
-  A(n - 1, n - 1) = 1.0;
 
-  auto qr = A.householderQr();
+  matrixEntries.emplace_back(Eigen::Triplet<double>(0, 0, 1.0));
+  matrixEntries.emplace_back(Eigen::Triplet<double>(n - 1, n - 1, 1.0));
+  A.setFromTriplets(matrixEntries.begin(), matrixEntries.end());
+  A.makeCompressed();
```
fsimonis, line 48: It would be interesting to see the cost-reward of this based on the amount of substeps and samplings one does.

fsimonis:
It would be interesting to use Events in the Bspline implementation and combine it with #2013.

This way, you could use the existing performance tools to analyze the cost.

src/math/Bspline.cpp
```diff
@@ -28,23 +29,29 @@ Bspline::Bspline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree
...
 
-  Eigen::MatrixXd controls = Eigen::MatrixXd::Zero(n, _ndofs);
+  Eigen::SparseQR<Eigen::SparseMatrix<double>, Eigen::COLAMDOrdering<int>> qr;
+  qr.analyzePattern(A);
+  qr.factorize(A);
 
+  Eigen::MatrixXd controls = Eigen::MatrixXd::Zero(n, _ndofs);
   for (int i = 0; i < _ndofs; i++) {
     controls.col(i) = qr.solve(xs.row(i).transpose());
```
fsimonis, line 56: Would it be possible to solve all `_ndofs` systems in one go by passing a matrix of systems to [solve](https://eigen.tuxfamily.org/dox/classEigen_1_1SparseQR.html#aa77f5cb833bbea615adb532139761971)?

Something like this:

```cpp
_ctrls = qr.solve(xs);
```

BenjaminRodenberg:
One thing I just realized: Please don't forget to add a changelog entry before merging. But I can also do another review loop when you are done with the remarks of @fsimonis.

fsimonis:
@NiklasKotarsky this looks pretty neat now!

Could you please compare what the final difference is between develop and this current version? :pray: :eyes:

docs/changelog/2021.md
```diff
@@ -0,0 +1 @@
+- Speed up the creation of the Bspline interpolation, by switching the matrix A inside the Bspline class to a sparse matrix
```
fsimonis, line 1: Could you please reword this to start with `- Improved `?
This results in a more digestible changelog.

src/math/Bspline.cpp
```diff
@@ -28,27 +29,31 @@ Bspline::Bspline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree
...
   for (Eigen::DenseIndex i = 1; i < n - 1; ++i) {
-    //Attempt at hack... the spline dimension is not used here explicitly
-    const Eigen::DenseIndex span = Eigen::Spline<double, 1>::Span(ts[i], splineDegree, _knots);
+    const Eigen::DenseIndex span      = Eigen::Spline<double, 1>::Span(ts[i], splineDegree, _knots);
+    auto                    basisFunc = Eigen::Spline<double, 1>::BasisFunctions(ts[i], splineDegree, _knots);
 
-    // The segment call should somehow be told the spline order at compile time.
-    A.row(i).segment(span - splineDegree, splineDegree + 1) = Eigen::Spline<double, 1>::BasisFunctions(ts[i], splineDegree, _knots);
+    for (Eigen::DenseIndex j = 0; j < splineDegree + 1; ++j) {
+      matrixEntries.insert(matrixEntries.end(), Eigen::Triplet<double>(i, span - splineDegree + j, basisFunc(j)));
```
fsimonis, line 41: My suggestion was the following:

```suggestion
      matrixEntries.emplace_back(i, span - splineDegree + j, basisFunc(j));
```

To clarify:

`emplace_back` constructs an element in-place using the given arguments. So, for a `vector` with element type `Eigen:Triplet<double>`, calling `emplace_back(x,y,z)` is like calling `push_back(Eigen::Triplet<double>(x,y,z))`.

In this case, this is mostly a readability improvement.

NiklasKotarsky:
> @NiklasKotarsky this looks pretty neat now!
> 
> Could you please compare what the final difference is between develop and this current version? 🙏 👀

Thanks for the review, I re-timed the new implementation (https://github.com/precice/precice/pull/2021) and for the 4 timesteps and 100 time windows I now get a runtime of 26.6 seconds, yielding similar performance to preCICe develop. For 1 time window and 2000 time steps it took 0.4 seconds compared to 3.3 seconds of preCICe develop.

src/math/Bspline.cpp
NiklasKotarsky: Thanks for the nice explanation. I originally had 'push_back((x,y,z))' without the 'Eigen::Triplet<double>', which triggered an error in the linter so I switched to 'emplace_back((x,y,z))', since this was suggested by the linter.

fsimonis (approved):
Let me know when you are done an I'll merge the PR. :smiley:

PR #2039, opened 2024-06-21
Remove `MAX_WAVEFORM_DEGREE`

### LINKED ISSUE #2038: Allow `MAX_WAVEFORM_DEGREE` > 10
We simply have to modify the following line:
https://github.com/precice/precice/blob/fc3fddf308652abb59799b5a949b91d7f9ec7c8e/src/time/Time.cpp#L9

Should we set an upper bound at all?

Keep it as it is and recompile preCICE, if I want to use a higher waveform degree.

**Additional context**
We decided to restrict the waveform degree when introducing the feature. For my experiments I'm regularly manually adjusting this option. So far, I have not seen any strange behavior when picking a very high waveform degree. Therefore, I think we could be less restrictive.

BenjaminRodenberg:
## Main changes of this PR

Allow arbitrary waveform degree by removing upper bound `MAX_WAVEFORM_DEGREE`

Closes #2038 

## Author's checklist

* [ ] ~~I added a test to cover the proposed changes in our test suite.~~
* [ ] ~~For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).~~

## Reviewers' checklist

### DIFF
--- a/docs/changelog/2039.md
+++ b/docs/changelog/2039.md
@@ -1,0 +1,1 @@
+- Allowed arbitraryly high waveform degree.
--- a/docs/changelog/2039.md
+++ b/docs/changelog/2039.md
@@ -1,1 +1,1 @@
-- Allowed arbitraryly high waveform degree.
+- Improved time-interpolation to allow arbitrary high waveform degree.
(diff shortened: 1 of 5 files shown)


docs/changelog/2039.md
```diff
@@ -0,0 +1 @@
+- Allowed arbitraryly high waveform degree.
```
fsimonis, line 1:
```suggestion
- Allowed arbitrary high waveform degree.
```

uekerman (approved):

fsimonis:
@BenjaminRodenberg Note that there is no test for higher degrees :wink:

PR #2044, opened 2024-06-27
Add error for missing samples when initializing coupling schemes

### LINKED ISSUE #1870: Exchanging data without write-data tag leads to an assertion
**Describe your setup**

Operating system (e.g. Linux distribution and version): _Arch_
preCICE Version: https://github.com/precice/precice/commit/15438f215adbcfdbb620cdb35dd9dc6712d2abc9

**Describe the problem**

Forgetting to specify a `<write-data name="D" mesh="M" />` with a coupling scheme exchanging data `D` triggers an assertion instead of an error.

```
Location:   void precice::cplscheme::BaseCouplingScheme::sendData(const m2n::PtrM2N &, const DataMap &)
File:       /home/simonis/Development/precice/third/src/cplscheme/BaseCouplingScheme.cpp:120
Expression: stamples.size() > 0
Rank:       1
Arguments:  none
```

**Step To Reproduce**
1. Omit `write-data` tag of an exchanged data
2. See the assertion triggered.

**Expected behaviour**

See an error during configuration or during the exchange.

**Additional context**

fsimonis:
## Main changes of this PR

This PR adds an error message when the user exchanges data which is not generated by the participant either via write-data tags or mappings.

Closes #1870

## Author's checklist

### DIFF
--- a/docs/changelog/2044.md
+++ b/docs/changelog/2044.md
@@ -1,1 +1,1 @@
-- Added a configuration check for data which is send, but neither written nor mapped.
+- Added a configuration check for data which is sent, but neither written nor mapped.
--- a/docs/changelog/2044.md
+++ b/docs/changelog/2044.md
@@ -1,0 +1,1 @@
+- Added a configuration check for data which is send, but neither written nor mapped.
--- a/docs/changelog/2044.md
+++ b/docs/changelog/2044.md
@@ -1,1 +1,1 @@
-- Added a configuration check for data which is sent, but neither written nor mapped.
+- Added a configuration check for missing data samples, which may be caused by missing mappings or `write-data` tags.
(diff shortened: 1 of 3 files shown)


BenjaminRodenberg (approved):
Adding a more helpful error message makes sense. I assume this error is triggered if the corresponding write-data tag is missing? Did you manually test this? Automatic testing is not possible. So this is one more item on the list for https://github.com/precice/precice/issues/280.

docs/changelog/2044.md
```diff
@@ -0,0 +1 @@
+- Added a configuration check for data which is send, but neither written nor mapped.
```
uekerman, line 1:
```suggestion
- Added a configuration check for data which is sent, but neither written nor mapped.
```

fsimonis:
@BenjaminRodenberg thanks for the nudge to test! While testing the changes, I realized that this can be generally tested in initialize of the BaseCouplingScheme.

If the coupling data is received and has no samples, then `initializeReceivedDataStorage` is buggy -> assertion.
If the coupling data is sent and has no samples, then the configuration has an issue -> Error.

I'll rework this PR.

PR #2046, opened 2024-07-03
Fix error messages for reading/writing data on undefined mesh

MakisH:
## Main changes of this PR

When trying to write data on a participant on a wrong mesh, I got the following message:

```
---[precice] ERROR:  Participant "Generator-Left" attempts to read data "Color" from an unknown mesh "Generator-Mesh". <mesh name="Generator-Mesh"> needs to be defined first.
```

Looking at the error messages, the read/write direction seems to be swapped. This PR exchanges the read<->write.

## Author's checklist

* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release. -> Do we need one?
* For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html). -> N/A
* I sticked to C++17 features. -> N/A
* I sticked to CMake version 3.22.1. -> N/A

## Reviewers' checklist

* Does the changelog entry make sense? Is it formatted correctly? -> N/A
* [x] Do you understand the code changes? @fsimonis


fsimonis (approved):
Good catch!

PR #2049, opened 2024-07-08
Use HIP as CMake language

davidscn:
## Main changes of this PR

Makes use of CMake's `enable_language` for HIP.

Since CMake version 3.21, [HIP is handled as a language](https://cmake.org/cmake/help/latest/command/enable_language.html) by CMake. Due to our baseline upgrade, we can make use of this upgrade.

This was contributed by @MarcelKoch

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -269,6 +269,11 @@ if (PRECICE_FEATURE_GINKGO_MAPPING)
 
   # 2. Check if Ginkgo was built with HIP and if HIP is available on the system
   if(PRECICE_WITH_HIP)
+    include(CheckLanguage)
+    check_language(HIP)
+    enable_language(HIP)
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -269,6 +269,11 @@ if (PRECICE_FEATURE_GINKGO_MAPPING)
 
   # 2. Check if Ginkgo was built with HIP and if HIP is available on the system
   if(PRECICE_WITH_HIP)
+    include(CheckLanguage)
```
fsimonis, line 272: Is it safe to include this multiple times?

fsimonis (approved):

CMakeLists.txt
davidscn: Despite the fact that executing both if branches seems like a very rare case, it is safe AFAIK.

PR #2051, opened 2024-07-09
Replace Ginkgo kernel launch by Kokkos extension

davidscn:
## Main changes of this PR

Replace Ginkgo kernel launch by Kokkos extension.

In our current setup, we relied on an unreleased/internal feature of Ginkgo to dispatch the RBF matrix assemblies onto the device. Since Ginkgo version 1.8.0 (released in June 2024), these kind of kernel dispatch mechanisms are supported through Kokkos, i.e., Ginkgo has interoperability with Kokkos algorithms. With these changes, the Ginkgo-featured mapping becomes compatible with Ginkgo v1.8.0 (onwards).

With the changes here/the way Ginkgo supports the dispatch now, Kokkos becomes a mandatory dependency of preCICE if the user wants to make use of our Ginkgo-featured data mappings. Note that Kokkos is a mandatory dependency for xSDK anyhow, as it is so widely used.

As opposed to Ginkgo, Kokkos is not _runtime_ configurable with respect to the target hardware, i.e., Kokkos only supports one hardware it was compiled for at a runtime. Therefore, Ginkgo inherits the supported hardware from Kokkos such that we lose the _runtime_ flexibility with this approach. In practice, that's not too much of a loss as typically there is one device meant for the computations which we are then tied to by the compiler.

We now have interoperability with `Kokkos::Serial` for the `Ginkgo::ReferenceExecutor` (meant for testing) and either OpenMP, Cuda or HIP for all existing cases.

The changes here were developed with the greatfully acknowledged help of @MarcelKoch in davidscn/precice#12

@MarcelKoch would you mind taking a look at the CMake config again? also, the following to remarks

- for my Cuda version, I had some trouble with `constexpr` in the kernel functions, I removed it for now (it was anyway only used for the kernel selection). Not sure if an extra compiler flag was required.
- I would compile the helper library as a static library. Otherwise we would need to install it along with preCICE and I don't see a benefit of doing so. What do you think?

## Author's checklist

* <del> For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html). </del>

## Post merge TODO

* [ ] update spack recipe

## Reviewers' checklist


MarcelKoch:
@davidscn the CMake config looks good to me.
Regarding the `constexpr` issue, for ginkgo, we usually add these flags to NVCC:
```
--expt-relaxed-constexpr
--extended-lambda
```
but I'm not sure if they are relevant to this case. 
Regarding the library, I'm not that familiar with shared or static library, but if using a static library works here, then I don't know of any issues that could appear. Just curious, would Ginkgo and Kokkos also need to be static libraries for this to work?

davidscn:
@MarcelKoch thanks for the feedback. I updated now a few more things and it should work now. Especially forwarding the `PRECICE_WITH_HIP` to the QR decomposition solved any related linker errors. 

> Just curious, would Ginkgo and Kokkos also need to be static libraries for this to work?

No, they (can) remain shared.

> Regarding the constexpr issue, for ginkgo, we usually add these flags to NVCC:

I am still a bit experimenting with this.

Also, I changed a few more things:

- I had to change the initialization of the device constant to make it work for `Kokkos:Serial` as well

```diff
-PRECICE_MEMORY_SPACE constexpr double NUMERICAL_ZERO_DIFFERENCE = 1.0e-14;
+#define NUMERICAL_ZERO_DIFFERENCE_DEVICE 1.0e-14
```

in the basis function implementation, I am not entirely sure if I should use `Kokkos::log` instead of the `std::log`. You also switched to `std::fma` in this regard. Any hint?

- The `std::array` causes some issues when compiling in debug mode, as the bound checking in the STL contains some host-only code. I replaced the access operator by a pointer access (no bound checking), but I think passing a compiler flag could solve the issue as well. Still figuring out the details.

MarcelKoch:
> in the basis function implementation, I am not entirely sure if I should use `Kokkos::log` instead of the `std::log`. You also switched to `std::fma` in this regard. Any hint?

I would prefer `Kokkos::log`. I don't remember why I used `std::fma`, maybe it's not available in Kokkos. If it is available, I would also prefer it.

>     * The `std::array` causes some issues when compiling in debug mode, as the bound checking in the STL contains some host-only code. I replaced the access operator by a pointer access (no bound checking), but I think passing a compiler flag could solve the issue as well. Still figuring out the details.

Kokkos also has a `Kokkos::Array` implementation, but somehow it doesn't show up in the documentation. Alternatively, you could use `Kokkos::View` with fixed bounds.

> Edit: It would be nice if we could inherit the device-specific flags from Ginkgo. I could imagine that these also change in the future.

I don't think this will happen. You can't interact directly with Ginkgo's device implementations, so there isn't any need for us to expose this.

davidscn:
Once the CI images are updated precice/ci-images#8 we are good to go, I tested all backends we support and the tests pass. Ready for review.

CMakeLists.txt
```diff
@@ -510,7 +507,7 @@ target_include_directories(precice PUBLIC
   $<INSTALL_INTERFACE:include>
   )
 
-target_link_libraries(precice PRIVATE preciceCore)
+target_link_libraries(precice PRIVATE preciceCore preciceExternal)
```
fsimonis, line 510: I still think it would be way simpler to extend the preciceCore directly instead of introducing multiple new targets. This way there is no risk for linking issues and the overall CMake code shrinks. Furthermore, symbol hiding (which doesn't exist in static libraries) and wider LTO scope should lead to more optimization potential.

src/mapping/GinkgoRadialBasisFctSolver.hpp
```diff
@@ -186,7 +158,9 @@ class GinkgoRadialBasisFctSolver {
 
   std::shared_ptr<gko::stop::Iteration::Factory> _iterationCriterion;
 
-  std::shared_ptr<gko::stop::ResidualNormReduction<>::Factory> _residualCriterion;
+  std::shared_ptr<gko::stop::ResidualNorm<>::Factory> _residualCriterion;
+
+  std::shared_ptr<gko::stop::ResidualNorm<>::Factory> _absoluteResidualCriterion;
```
fsimonis, line 163: Is the ownership really shared?

src/mapping/device/Ginkgo.cpp
```diff
@@ -0,0 +1,35 @@
...
+    Kokkos::initialize(Kokkos::InitializationSettings().set_num_threads(nThreads).set_device_id(deviceId).set_disable_warnings(true));
+  }
+}
+
+void Ginkgo::finalize()
+{
+  if (Kokkos::is_initialized()) {
+    Kokkos::finalize();
+  }
+}
```
fsimonis, line 31: Similar to MPI and PETSc, the user may have already initialized this, hence we need to keep track if we initialized kokkos and only then finalize it.

src/mapping/device/GinkgoRBFKernels.cpp
```diff
@@ -1,132 +1,188 @@
...
+    Kokkos::parallel_for(
+        "create_rbf_system_matrix_row_major",
+        Kokkos::MDRangePolicy<typename MemorySpace::execution_space, Kokkos::Rank<2>>{{0, 0}, {mtx->get_size()[0], mtx->get_size()[1]}},
+        KOKKOS_LAMBDA(const int &i, const int &j) {
+          double dist = 0;
+          if (addPolynomial) {
+            k_mtx(i, j) = 0; // Zero the matrix entry if polynomial terms are added
+          }
+          // We need to use a pointer here, because the bound checking of std::array
+          // contins some host-only code, which yields errors when compiling in
```
fsimonis, line 62:
```suggestion
          // contains some host-only code, which yields errors when compiling in
```

src/mapping/device/GinkgoRBFKernels.cpp
```diff
@@ -1,132 +1,188 @@
...
+                                   bool                                      addPolynomial,
+                                   unsigned int                              extraDims)
+{
+  auto k_mtx           = map_data<MemorySpace>(mtx.get());
+  auto k_supportPoints = map_data<MemorySpace>(supportPoints.get());
+  auto k_targetPoints  = map_data<MemorySpace>(targetPoints.get());
+
+  // the outcome of this if-condition is known at compile time. However, Cuda
+  // has issues with constexpr
+  if (std::is_same_v<MemorySpace, Kokkos::HostSpace>) {
```
fsimonis, line 51: Does it make sense to split the function up into two?

The general col-major version and then specialize it for MemorySpace `Kokkos::HostSpace`?

src/mapping/device/GinkgoRBFKernels.cpp
```diff
@@ -1,132 +1,188 @@
...
 
-template void create_rbf_system_matrix<double, precice::mapping::CompactPolynomialC2>(std::shared_ptr<const DefaultExecutor>, const std::size_t, const std::size_t, const std::size_t, const std::array<bool, 3>,
-                                                                                      double *, double *, double *, precice::mapping::CompactPolynomialC2, const RadialBasisParameters,
-                                                                                      const std::size_t, const std::size_t, const bool, const unsigned int);
+namespace kernel {
+
+template <typename MemorySpace, typename EvalFunctionType>
+void create_rbf_system_matrix_impl(std::shared_ptr<const gko::Executor>      exec,
+                                   gko::ptr_param<GinkgoMatrix>              mtx,
+                                   const std::array<bool, 3>                 activeAxis,
```
fsimonis, line 37: If you ever need more performance, turning this into a template argument could help.

src/mapping/device/GinkgoRBFKernels.cpp
```diff
@@ -1,132 +1,188 @@
...
+    create_rbf_system_matrix_impl<Kokkos::HostSpace>(exec, mtx, activeAxis, supportPoints, targetPoints, f, rbf_params, addPolynomial, extraDims);
+  } else {
+    create_rbf_system_matrix_impl<Kokkos::DefaultExecutionSpace::memory_space>(exec, mtx, activeAxis, supportPoints, targetPoints, f, rbf_params, addPolynomial, extraDims);
+  }
+}
 
-template void create_rbf_system_matrix<double, precice::mapping::CompactPolynomialC6>(std::shared_ptr<const DefaultExecutor>, const std::size_t, const std::size_t, const std::size_t, const std::array<bool, 3>,
-                                                                                      double *, double *, double *, precice::mapping::CompactPolynomialC6, const RadialBasisParameters,
-                                                                                      const std::size_t, const std::size_t, const bool, const unsigned int);
+#define INSTATIATE_CREATE_RBF_SYSTEM_MATRIX(_function_type)                                                                                     \
```
fsimonis, line 119: xSDK requires own macros to start with the project name. Also, typo :wink: 

```suggestion
#define PRECICE_INSTANTIATE_CREATE_RBF_SYSTEM_MATRIX(_function_type)                                                                                     \
```

src/mapping/device/GinkgoRBFKernels.cpp
```diff
@@ -1,132 +1,188 @@
...
+                                 const unsigned int                   dims)
+{
 
-template void create_rbf_system_matrix<double, precice::mapping::CompactPolynomialC8>(std::shared_ptr<const DefaultExecutor>, const std::size_t, const std::size_t, const std::size_t, const std::array<bool, 3>,
-                                                                                      double *, double *, double *, precice::mapping::CompactPolynomialC8, const RadialBasisParameters,
-                                                                                      const std::size_t, const std::size_t, const bool, const unsigned int);
+  auto k_mtx = map_data<MemorySpace>(mtx.get());
+  auto k_x   = map_data<MemorySpace>(x.get());
+
+  if (std::is_same_v<MemorySpace, Kokkos::HostSpace>) {
```
fsimonis, line 147: Same as above

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -3,30 +3,40 @@
...
-#define FMA fma
+#define FMA Kokkos::fma
+#define LOG Kokkos::log
 
 #else
 
 #define PRECICE_HOST_DEVICE
 #define PRECICE_MEMORY_SPACE
 #define FMA std::fma
+#define LOG std::log
```
fsimonis, line 31: Is there any reason for this aside convenience?

Macros need to be prefixed due to xSDK, which eliminates the length upside.

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -3,30 +3,40 @@
...
 
 #endif
 
-PRECICE_MEMORY_SPACE const double NUMERICAL_ZERO_DIFFERENCE = 1.0e-14;
+#define NUMERICAL_ZERO_DIFFERENCE_DEVICE 1.0e-14
```
fsimonis, line 35: `constexpr double` is the modern way of doing this.

src/mapping/tests/GinkgoRadialBasisFctSolverTest.cpp
```diff
@@ -50,72 +50,72 @@ BOOST_AUTO_TEST_SUITE(GinkgoRadialBasisFunctionSolver)
 #define TEST_FOR_ALL_RBFS(EXECUTOR, SOLVER)                                              \
```
fsimonis, line 50: You can do this directly in [boost test](https://www.boost.org/doc/libs/1_85_0/libs/test/doc/html/boost_test/tests_organization/test_cases/test_organization_templates.html#boost_test.tests_organization.test_cases.test_organization_templates.test_case_template_with_manual_r)
For another PR though

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -514,6 +515,10 @@ void ParticipantImpl::finalize()
 
   // Finalize PETSc and Events first
   utils::Petsc::finalize();
+// This will lead to issues if we call finalize afterwards again
+#ifndef PRECICE_NO_GINKGO
+  device::Ginkgo::finalize();
```
fsimonis, line 520: Consistent alternative here is to make this do nothing if precice isn't compiled with ginkgo.

src/testing/TestContext.cpp
```diff
@@ -213,6 +219,17 @@ void TestContext::initializePetsc()
   }
 }
 
+void TestContext::initializeGinkgo()
+{
+  if (!invalid && _kokkos) {
```
fsimonis, line 224: Kokkos or Ginkgo? :sweat_smile:

src/testing/TestContext.hpp
```diff
@@ -285,6 +287,9 @@ class TestContext {
   /// whether to initialize events
   bool _events = false;
 
+  /// whether to initialize PETSc
```
fsimonis, line 290: or ginkgo?
```suggestion
  /// whether to initialize Kokkos
```

CMakeLists.txt
```diff
@@ -57,7 +57,7 @@ option(PRECICE_FEATURE_MPI_COMMUNICATION "Enables MPI-based communication and ru
 option(PRECICE_FEATURE_PETSC_MAPPING "Enable use of the PETSc linear algebra library for global-iterative RBF data mappings." ON)
 option(PRECICE_FEATURE_PYTHON_ACTIONS "Enable Python support for preCICE actions." ON)
 option(PRECICE_CONFIGURE_PACKAGE_GENERATION "Configure package generation." ON)
-option(PRECICE_FEATURE_GINKGO_MAPPING "**Experimental** Enable use of Ginkgo (unreleased version) for data mapping." OFF)
+option(PRECICE_FEATURE_GINKGO_MAPPING "Enable use of Ginkgo for accelerated data mapping." OFF)
```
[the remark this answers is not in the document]
davidscn: Hm I would maybe leave it as is, but no strong opinion. We also use Kokkos types through Ginkgo.

CMakeLists.txt
```diff
@@ -424,26 +427,20 @@ else()
...
-  set_target_properties(preciceCore_cuda PROPERTIES POSITION_INDEPENDENT_CODE ON)
-  set_target_properties(preciceCore_reference PROPERTIES POSITION_INDEPENDENT_CODE ON)
-  set_target_properties(preciceCore_hip PROPERTIES POSITION_INDEPENDENT_CODE ON)
-  set_target_properties(preciceCore_omp PROPERTIES POSITION_INDEPENDENT_CODE ON)
-  set_target_properties(preciceCore_dpcpp PROPERTIES POSITION_INDEPENDENT_CODE ON)
+  add_library(preciceCore_ginkgo STATIC
+          src/mapping/device/GinkgoRBFKernels.cpp
+          src/mapping/device/Ginkgo.cpp)
+  target_link_libraries(preciceCore_ginkgo PRIVATE Ginkgo::ginkgo Kokkos::kokkos preciceCore)
+  target_compile_definitions(preciceCore_ginkgo PUBLIC EIGEN_NO_CUDA)
```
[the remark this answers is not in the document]
davidscn: Eigen 'only' provides templates, which are implicitly inline, so I don't see that this violates the ODR rule.

src/mapping/device/Ginkgo.cpp
davidscn: That's what we do, no?

src/mapping/device/GinkgoRBFKernels.cpp
davidscn: In principle, this states an alternative, yes. However, this would require more boilerplate code and I don't see immediate benefit or the upside.

src/mapping/device/GinkgoRBFKernels.cpp
davidscn: The `std::array`?

src/mapping/GinkgoRadialBasisFctSolver.hpp
davidscn: Yes, typically between host and device.

src/mapping/device/GinkgoRBFKernels.cpp
davidscn: Yes, agree, I would probably even undefine it directly at the end of the file.

src/mapping/impl/BasisFunctions.hpp
davidscn: Yes, the macro maps to different functions, depending on which compiler looks at the file. Either the host compiler, or the kokkos launcher which might select the device compiler. For device code, we make use of `Kokkos::log` and `Kokkos::fma`.

src/mapping/impl/BasisFunctions.hpp
davidscn: Yes, I did it this way because the reference executor with Kokkos didn't want to accept the `constexpr double` and whenever used, the function just returned `nan`. Using a `#define` was the only solution I could find at this point.

src/precice/impl/ParticipantImpl.cpp
davidscn: Depends a bit on how we treat the build pipeline: In the current pipeline, the kokkos compiler launcher takes a look at the corresponding source file, i.e., preciceCore only gets the link to `finalize`. If the library is then compiled without Ginkgo, the interface is not created and we end up with undefined references. We would then need to compile a dummy file.

src/testing/TestContext.cpp
davidscn: I think this is mainly about initializing the device. Hence, it doesn't really matter here.

CMakeLists.txt
davidscn: How would this then look like? Note that this would then lead to the whole preciceCore linking against Kokkos and Kokkos ships its own compiler launcher, such that preCICE inherits flags from Kokkos. Different source code files still need to be compiled by different compilers. In principle, we can of course also build the helper interfaces as dynamic libraries, but then we might need to install them.

CMakeLists.txt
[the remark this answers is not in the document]
fsimonis: Templates reduce to symbols, which still clash if their implementation differs. This also holds for the solver btw, which needs to be compiled with the same macros. It's essentially the same reason Jaust ran into strange behaviour due to conflicting destructors in Eigen.

src/mapping/device/Ginkgo.cpp
fsimonis: Currently, if the user initializes Kokkos before calling preCICE, then precice finalizes  Kokkos.

src/mapping/device/GinkgoRBFKernels.cpp
fsimonis: The xyz bools, yes.

src/mapping/device/GinkgoRBFKernels.cpp
fsimonis: Specializing doesn't require much boilerplate. It could actually be net 0 in this case.

src/mapping/impl/BasisFunctions.hpp
fsimonis: Given that function aliases are a pain, and selective overloading is too cumbersome, I would either create a wrapper `precice::mapping::detail::log(double d)` which internally calls the one or other, or prefix the macro.

src/precice/impl/ParticipantImpl.cpp
fsimonis: > We would then need to compile a dummy file.

True. This is essentially what we do for MPI and PETSc.

I am not saying one better than the other. Only upside in my eyes of the dummy is that calling code (especially ParticipantImpl) gets cleaner.

src/testing/TestContext.cpp
fsimonis: The names should be consistent tough. Requirement, option, function

CMakeLists.txt
fsimonis: > Note that this would then lead to the whole preciceCore linking against Kokkos and Kokkos ships its own compiler launcher, such that preCICE inherits flags from Kokkos. Different source code files still need to be compiled by different compilers.

If kokkos defines the compiler launcher for CUDA, then this shouldn't impact precice sources.

> How would this then look like?

I see two options:

1. we use `target_sources` to add all extra files straight to the preciceCore, link the core to ginkgo and Kokkos, and let CMake figure out the rest. If this works, then this would be the simplest solution.
2. we define an object library for device code, link that to ginkgo and Kokkos, and link that to the preciceCore. precice and testprecice may need to be manually linked to ginkgo and Kokkos for this to work.

https://cmake.org/cmake/help/latest/command/add_library.html#object-libraries

fsimonis:
It compiles using spack and openmp only.

CMakeLists.txt
davidscn: > If kokkos defines the compiler launcher for CUDA, then this shouldn't impact precice sources.

The launcher is used for all files. Kokkos can also be used for other optimizations (OpenMP, Serial, Vectorization etc..)

Option two is what we do right now, correct?

How would option one look like with the additional variables and language definitions?

src/mapping/device/GinkgoRBFKernels.cpp
davidscn: What makes you assume so? The array has already a compile-time constant size, I don't see templates helping here with performance.

src/mapping/impl/BasisFunctions.hpp
davidscn: > or prefix the macro.

..and undefine it

CMakeLists.txt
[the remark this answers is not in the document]
davidscn: I don't really understand how inline code can clash, but we can just take over the compile definition to preciceCore. I would assume that this doesn't make a difference anyhow, since the definition makes the functions then mostly callable and executable on a device, which is not what we compile anywhere.

src/mapping/device/GinkgoRBFKernels.cpp
davidscn:
```bash 
error: non-class, non-variable partial specialization ‘create_rbf_system_matrix_impl<Kokkos::HostSpace, EvalFunctionType>’ is not allowed
   48 | void create_rbf_system_matrix_impl<Kokkos::HostSpace, EvalFunctionType>
```

and the net
 1 file changed, 82 insertions(+), 57 deletions(-)

Repeating the signature is already significant. So I would leave it as is?

CMakeLists.txt
davidscn: Let me try this

davidscn: Tried something like this

```diff
 # Setup Ginkgo
 if (PRECICE_FEATURE_GINKGO_MAPPING)
   # Next we configure the common kernels for the matrix assembly
-  add_library(preciceCore_ginkgo STATIC
-          src/mapping/device/GinkgoRBFKernels.cpp
-          src/mapping/device/Ginkgo.cpp)
-  target_link_libraries(preciceCore_ginkgo PRIVATE Ginkgo::ginkgo Kokkos::kokkos preciceCore)
+
+  target_sources(preciceCore PRIVATE
+  src/mapping/device/GinkgoRBFKernels.cpp
+  src/mapping/device/Ginkgo.cpp
+  )
+
+  target_link_libraries(preciceCore PRIVATE Ginkgo::ginkgo Kokkos::kokkos)
   target_compile_definitions(preciceCore PUBLIC EIGEN_NO_CUDA)
-  target_compile_definitions(preciceCore_ginkgo PUBLIC EIGEN_NO_CUDA)
-  set_target_properties(preciceCore_ginkgo PROPERTIES POSITION_INDEPENDENT_CODE ON)
 
   # Disable compiler launcher since Kokkos ships its own nvcc wrapper
-  set_target_properties(preciceCore_ginkgo PROPERTIES CXX_COMPILER_LAUNCHER "")
+  set_target_properties(preciceCore PROPERTIES CXX_COMPILER_LAUNCHER "")
 
```

but this triggers all kinds of errors in the 'preCICE' TUs, especially Boost PP macros seem to be a problem here:

```bash
precice/src/mesh/Vertex.hpp(130): error: identifier "PRECICE_ASSERT_IMPL_BOOST_PP_IIF_BOOST_PP_BOOL_BOOST_PP_IIF_BOOST_PP_NOT_EQUAL_CHECK_BOOST_PP_NOT_EQUAL_BOOST_PP_VARIADIC_SIZE" is undefined

precice/src/mesh/Vertex.hpp(130): error: identifier "BOOST_PP_NIL" is undefined
```

src/mapping/device/GinkgoRBFKernels.cpp
fsimonis: I mean to turn the three bools into template arguments. So you end up with one function per valid combination, eliminating the inner loop.

davidscn: ah I see, noted!

src/mapping/device/GinkgoRBFKernels.cpp
davidscn: https://github.com/precice/precice/pull/2051#discussion_r1682767677

CMakeLists.txt
fsimonis: https://github.com/precice/precice/pull/2051#discussion_r1682815113 this looks correct and the errors look very strange.

Then we need option two.

> Option two is what we do right now, correct?

Currently, we build static libraries which are then linked against the shared library. This comes with some additional implications.

Using an object library means "please compile these sources in this different way, I'll use them later to link a library.". This is exactly what we need here.

davidscn: Can you give me some hint on what to do in this case? Changing `STATIC` to `OBJECT`, yes, but I also need to link Kokkos and other TPLs then against the interface, right?

```diff
-  add_library(preciceCore_ginkgo STATIC
+  add_library(preciceCore_ginkgo OBJECT
```
and
```diff
-  target_link_libraries(preciceExternal INTERFACE preciceCore_ginkgo)
-else()
+  target_link_libraries(preciceExternal INTERFACE $<TARGET_OBJECTS:preciceCore_ginkgo>)
+  target_link_libraries(preciceExternal INTERFACE Kokkos::kokkos)
```

davidscn:
For spack, the following build succeeded for me:
```
./spack install precice@develop+gpu~petsc ^ginkgo@develop+cuda+openmp cuda_arch=80 ^kokkos+cuda+wrapper+pic cuda_arch=80 cuda_lambda=true cuda_constexpr=true
```
with an external cuda `spack external find cuda`.

src/precice/impl/ParticipantImpl.cpp
davidscn: Considering that we have `Ginkgo.cpp` now as a separate Object library, I would consider the current approach easier to realize.

CMakeLists.txt
fsimonis: After changing to `OBJECT`, you should be able to directly link to the object lib. Older version of CMake didn't do this correctly, but 3.12 should work.

```diff
-  target_link_libraries(preciceExternal INTERFACE preciceCore_ginkgo)
+  target_link_libraries(preciceCore PUBLIC preciceCore_ginkgo)
+  target_link_libraries(preciceCore_ginkgo PUBLIC Kokkos::kokkos)
```

davidscn: If I do this I end up with the following:

```bash
CMake Error: The inter-target dependency graph contains the following strongly connected component (cycle):
  "preciceCore" of type OBJECT_LIBRARY
    depends on "preciceCore_ginkgo" (weak)
  "preciceCore_ginkgo" of type OBJECT_LIBRARY
    depends on "preciceCore" (weak)
At least one of these targets is not a STATIC_LIBRARY.  Cyclic dependencies are allowed only among static libraries.
```
so I should link the ginkgo Object against preCICE itself?

fsimonis: What could work is to remove the dependency of Core on Core_ginkgo and make precice and testprecice depend on both Core and Core_ginkgo.

The best solution would be https://github.com/precice/precice/pull/2051#discussion_r1682815113 though. Maybe its worth having a deeper look into why this doesn't work. The problem is the Kokkos compiler wrapper, correct?

davidscn: I made it to work (see the last two commits) and compiling succeeds for all backends we offer. The culprit of the aforementioned issues were actually, that nvcc has issues with variadic macros.  I enabled them now in all TUs in preciceCore `BOOST_PP_VARIADICS`. Also, for HIP and Cuda, I had to resolve the `::min` conflict in CompositionalCouplingSchemes. 

For testing, the OpenMP setup works flawless. However, for Cuda and Hip, I get the following runtime error:

```bash
 Product of tile dimensions exceed maximum limit: 0
ExecSpace Error: MDRange tile dims exceed maximum number of threads per block - choose smaller tile dims
Backtrace:
[0x7fc513d50ad5] Kokkos::Impl::save_stacktrace()
[0x7fc513d24a9c] Kokkos::Impl::host_abort(char const*)
```
which puzzles me a bit. The previous CMake setup works without this error. Any clues here?

davidscn: With Kokkos version 4.3.01, Cuda works and only with Kokkos develop ([eb11070](https://github.com/kokkos/kokkos/commit/eb11070f67565b2e660659f5207f0363bdf3b882)) also HIP works. I have no clue what's going on. With Cuda, version 4.3.00 doesn't work either.

davidscn:
Summarizing https://github.com/precice/precice/pull/2051#discussion_r1692793129:

The problem here is that Kokkos has problems with hidden visibility https://github.com/kokkos/kokkos/pull/7065
Luckily, this is fixed on Kokkos develop [a6beb5a](https://github.com/kokkos/kokkos/commit/a6beb5a568d5337e84fd6d2797a27a7bef0d65d1). With the next release, using the current state of the build tool chain should be possible. 

As a solution, we now disable the hidden visibility in preCICE conditionally, if the applied Kokkos version lies in between the minimum required version (4.2) and below 4.4. If the Kokkos version is 4.4 or higher, everything remains as it is. Once we require Kokkos version 4.4 in preCICE in general, we can then simply remove this piece of code from the CMakeLists

CMakeLists.txt
fsimonis: Good investigative development

fsimonis (approved):
That was more work than expected :sweat_smile: 

Looks good now :+1:

Issue #1991, opened 2024-04-05
Create Paraview series files for time-window exports


fsimonis:
**Please describe the problem you are trying to solve.**

preCICE exports are numbered either using the current time window or iteration.
For exported time windows, it can be tedious to visualize the results as they are decoupled from the actual time.

When comparing exports to other exports, this is not an issue as all exports follow the same numbering scheme, assuming no mixed time-window-sizes.

It can be tedious to synchronize exports of preCICE to exports of the solvers, as they may follow different time scales.

**Describe the solution you propose.**

As exports of time-window have a unique mapping from file to time, we can export this mapping in the shape of a `.series` file.

[Example of a Paraview series file](https://docs.paraview.org/en/latest/UsersGuide/dataIngestion.html#handling-temporal-file-series) `SolidMesh-Solid.vtk.series`:

```json
{
  "file-series-version" : "1.0",
  "files" : [
    { "name" : "SolidMesh-Solid.dt0.vtk", "time" : 0 },
    { "name" : "SolidMesh-Solid.dt1.vtk", "time" : 0.1 },
    { "name" : "SolidMesh-Solid.dt2.vtk", "time" : 0.2 }
  ]
}
```

This allows to directly integrate the visualization of the preCICE exports with the result of OpenFOAM `foamToVTK` script.

**Describe alternatives you've considered**

Instead of handling this in the Export, we could also create a generator-script.
Meshes, Participants, time-window-sizes and exports can be extracted from the configuration file.

fsimonis:
An alternative seem to be pvd files, which are in XML.

MakisH:
I have encountered `.pvd` files in the wild, but not `.series`, as far as I can remember. Definitely a nice addition!

fsimonis:
It seems that pvd is the perfect fit. It can be used to refer to any kind of data loadable by Paraview and can refer to various parts per timestep. So in theory it can be used for all exporters, even to load partitioned CSV.

```xml
<VTKFile type="Collection" version="0.1">
  <Collection>
    <DataSet file="mypiece_dt1.pvtu" timestep="0.001"/>
    <DataSet file="mypiece_dt2.pvtu" timestep="0.002"/>
  </Collection>
</VTKFile>
```

For formats with multiple parts:

```xml
<VTKFile type="Collection" version="0.1">
  <Collection>
    <DataSet file="mypiece_r0.csv" part="0" timestep="0.001"/>
[...]

Issue #1563, opened 2023-02-13
The initial and final exports contain 0 data


fsimonis:
**Describe your setup**

Operating system (e.g. Linux distribution and version): _Ubuntu 18.04_
preCICE Version: 2.5.0 + develop

**Describe the problem**

When using exports, the `inital` and `final` exports do contain 0 data.

For the `initial` export, this is the case even if data is initialized by the solver. This will was partially solved by #1350 as now the data is written first. Written data is still reset.

For the `final` export, written data is reset and thus it will contain 0 data.

**Expected behaviour**
Export data before it is reset.

**Additional context**

#1157 closes this issue.

uekerman:
Let's also consider whether removing the final vtk export completely. I don't see a real use case and it often confuses more than it helps.

PR #2053, opened 2024-07-15
Refactor mesh information into exporters

fsimonis:
## Main changes of this PR

This PR refactors the mesh, parallel context, and location into the exporters.
Hence, there is now one `Exporter` instance per mesh.

This simplifies the participant and the handling of filenames.
It also allows adding `series` files #1991.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/io/ExportCSV.cpp
+++ b/src/io/ExportCSV.cpp
@@ -68,11 +68,9 @@ void ExportCSV::doExport(int index, double time)
   }
   outfile /= filename;
 
-  const auto &mesh = *_mesh;
-
   // Prepare filestream
   std::ofstream outFile(outfile.string(), std::ios::trunc);
-  const bool    is3d = (mesh.getDimensions() == 3);
+  const bool    is3d = (_mesh->getDimensions() == 3);
 
   // write header
   outFile << "PosX;PosY";
@@ -80,10 +78,10 @@ void ExportCSV::doExport(int index, double time)
     outFile << ";PosZ";
   }
   outFile << ";Rank";
-  for (const auto &data : mesh.data()) {
+  for (const auto &data : _mesh->data()) {
     auto dataName = data->getName();
     auto dim      = data->getDimensions();
-    PRECICE_ASSERT(static_cast<std::size_t>(data->values().size()) == mesh.nVertices() * dim);
+    PRECICE_ASSERT(static_cast<std::size_t>(data->values().size()) == _mesh->nVertices() * dim);
     outFile << ';' << dataName;
     if (dim == 2) {
       outFile << "X;" << dataName << 'Y';
@@ -95,7 +93,7 @@ void ExportCSV::doExport(int index, double time)
 
   // Prepare writing data
   std::vector<StridedAccess> dataColumns;
-  for (const auto &data : mesh.data()) {
+  for (const auto &data : _mesh->data()) {
     auto    dim    = data->getDimensions();
     double *values = data->values().data();
     for (int i = 0; i < dim; ++i) {
@@ -105,9 +103,9 @@ void ExportCSV::doExport(int index, double time)
 
   // write vertex data
   const std::string rankCol = ";" + std::to_string(_rank);
-  const auto        size    = mesh.nVertices();
+  const auto        size    = _mesh->nVertices();
   for (std::size_t vid = 0; vid < size; ++vid) {
-    const auto &vertex = mesh.vertex(vid);
+    const auto &vertex = _mesh->vertex(vid);
     outFile << vertex.coord(0) << ';';
     outFile << vertex.coord(1);
     if (is3d) {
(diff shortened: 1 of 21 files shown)


src/io/ExportCSV.cpp
```diff
@@ -35,35 +35,41 @@ struct StridedAccess {
...
   namespace fs = std::filesystem;
-  fs::path outfile(location);
-  if (not location.empty()) {
+  fs::path outfile(_location);
+  if (not _location.empty()) {
     fs::create_directories(outfile);
   }
   outfile /= filename;
 
+  const auto &mesh = *_mesh;
```
davidscn, line 71: Why is this necessary? If you work with a pointer before, I would prefer a pointer in the following as well.

src/io/ExportVTK.cpp
```diff
@@ -17,28 +17,38 @@
...
 {
-  PRECICE_TRACE(name, location, mesh.getName());
-  PRECICE_ASSERT(name != std::string(""));
-  PRECICE_ASSERT(!utils::IntraComm::isParallel(), "ExportVTK only supports serial participants.");
+  PRECICE_TRACE(index, time, _mesh->getName());
+  PRECICE_ASSERT(index >= 0);
+  PRECICE_ASSERT(time >= 0.0);
+  PRECICE_ASSERT(!isParallel(), "ExportVTK only supports serial participants.");
+
+  auto filename = fmt::format("{}-{}.{}{}.vtk", _participantName, _mesh->getName(), kindPrefix(), index);
```
davidscn, line 37: This changes the current export names, right? Could affect the users and or the tests.

src/io/ExportXML.cpp
```diff
@@ -138,6 +157,9 @@ void ExportXML::writeSubFile(
   outSubFile << (utils::isMachineBigEndian() ? "BigEndian\">" : "LittleEndian\">") << '\n';
 
   outSubFile << "   <" << formatType << ">\n";
+
+  const auto &mesh = *_mesh;
```
davidscn, line 161: Same here

davidscn (approved):
Looks good, only minor things

src/io/ExportVTK.cpp
fsimonis: Actually, I think this is the only instance, where the filename is the same.

My idea was to make paraview detect all exports as series. I haven't though about backwards compatibility.

I'll revert the name changes.

src/io/ExportCSV.cpp
fsimonis: It prevents having to change the uses of mesh in the remaining code. I'll clean it up.

fsimonis:
@davidscn this may be a good opportunity to change the `inital` export to `dt0`. Would you consider this a bugfix or a breaking change?

davidscn:
> @davidscn this may be a good opportunity to change the `inital` export to `dt0`. Would you consider this a bugfix or a breaking change?

If anything then breaking. It was or is not a bug and should be mostly a debugging feature anyway. It might be, however, that people use the exports in scripts, we also use it in the replay mode for example (also in the systemtests @MakisH ?). Not sure what will happen if we rename it here without clear communication and or backwards compatibility.

fsimonis:
I doubt that the inital exports are used especially as they never contained data up to version 3 #1563. Are they used/loaded in ASTE?

davidscn:
https://github.com/precice/aste/blob/bdc3fbb45051a5f3721d56a390a798b399ca33c3/src/mesh.cpp#L384-L390

MakisH:
We currently do have `.init.vtk` files in the reference results of the tutorials (i.e., system tests), but we can update the reference results to remove those.

fsimonis:
OK, reverting to init

fsimonis:
Patched some edge cases. Merging now

MakisH:
> We currently do have `.init.vtk` files in the reference results of the tutorials (i.e., system tests), but we can update the reference results to remove those.

> OK, reverting to init

Did not react early enough to this: My comment was only an FYI. I was not opposing the `.init`. I think that eventually it would make sense to replace the `init` with `0`, as also in the training course, it appears as a duplicate entry in ParaView (which in that context definitely generates questions).

fsimonis:
> it appears as a duplicate entry in ParaView 

so dt0 == init?

MakisH:
No, I meant that you see the `init` and the group `1..N` separately. But this is why you started this discussion, I guess.

fsimonis:
Ah. That's part of the motivation for #2055. Now there is one file you can load which also exposes the init export as t=0.

MakisH:
We held back the `.init` to not break backwards compatibility, but it looks like we did break it by swapping the order of mesh and participant:

- previously: `Fluid-Mesh-Fluid.dt1.vtk`
- now: `Fluid-Fluid-Mesh.dt1.vtk`

While this makes sense, it now breaks the system tests making debugging a bit complicated (some bugs are actual bugs, but all tests fail because of the file renaming).

Could we revert this change until we can do the breaking in a more controlled and documented way?

PR #2055, opened 2024-07-16
Add series to exporter

### LINKED ISSUE #1991: Create Paraview series files for time-window exports

fsimonis:
## Main changes of this PR

This PR adds series files to exporters exporting time windows.

This allows opening series from the terminal using paraview.
It also assigns the correct timestamp to every export.

Series for iterations are not possible as paraview cannot handle repeated time stamps.

Closes #1991

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/2055.md
+++ b/docs/changelog/2055.md
@@ -1,0 +1,1 @@
+- Added Paraview series files to most exporters of tine windows. This makes loading exports easier and assigns the correct time stamps to them.
--- a/docs/changelog/2055.md
+++ b/docs/changelog/2055.md
@@ -1,1 +1,1 @@
-- Added Paraview series files to most exporters of tine windows. This makes loading exports easier and assigns the correct time stamps to them.
+- Added Paraview series files to most exporters of time windows. This makes loading exports easier and assigns the correct time stamps to them.
--- a/docs/changelog/2055.md
+++ b/docs/changelog/2055.md
@@ -1,1 +1,1 @@
-- Added Paraview series files to most exporters of time windows. This makes loading exports easier and assigns the correct time stamps to them.
+- Added Paraview series files to `vtk`, `vtu`, and `vtp` exporters when exporting time windows. This makes loading exports easier and assigns the correct time stamps to them.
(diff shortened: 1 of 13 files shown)


fsimonis:
I still need to test parallel exports.

fsimonis:
Tried with vtp exporter on a parallel participant. Loading the `.pvtp.series` file straight from the terminal works as expected.

docs/changelog/2055.md
```diff
@@ -0,0 +1 @@
+- Added Paraview series files to most exporters of tine windows. This makes loading exports easier and assigns the correct time stamps to them.
```
uekerman, line 1:
```suggestion
- Added Paraview series files to most exporters of time windows. This makes loading exports easier and assigns the correct time stamps to them.
```
Could you add more details on the "most" here?

fsimonis:
@uekerman our exporter tests only tests if things don't explode. I think a decent step forward would be to at least check for created files in the directory. In another PR though.
Checking the content would be better, but requires additional dependencies.

PR #2059, opened 2024-07-22
Implement RBF LOOCV error estimate

davidscn:
## Main changes of this PR

Implements the [Rippa scheme](https://doi.org/10.1023/a:1018975909870)  for LOOCV in our RBF Solver class. 

With this, we can give users information on how accurate the current RBF configuration actually is, which increases the usability of these methods significantly.

This is just the very first step, namely the core algorithm <del>(for positive-definite basis functions for now)</del>. There are still open questions of practical relevance:
- the biggest open question is when to compute and report the error metric. The advantage of the Rippa algorithm is that it takes the actual data into account, i.e., it tells you how good your current interpolant is. However, in many multi-physics simulations, the initial data is just zero or some constant and not representative for the actual simulation data
- do we want to make this option configurable? Somewhat related: I still have to evaluate the overhead the algorithm introduces, but I would assume that it makes a considerable difference. For in-production runs, it should be possible to turns this off. Maybe we can even add some related option into the `Mapping` class as we have similar accuracy metrics for NN, NP or LCI. However, these metrics only take the topology of the meshes into account and no preCICE configuration. We could, of course, also consider to evaluate the LOOCV initially for some dummy data or function.
- how do we handle the polynomial in this context? do we want to compute the estimate based on data with or without polynomial?
- for our PUM we have to think about a global aggregation method. I would consider to scale each error component by the cluster weight.

Some data regarding performance:

For the Cholesky decomposition, comparing the manual forward substitution against Eigen TRSM, I get the following, where the LOOCV is a subevent of mapData and the manual implementation is given by:

```cpp
    const Eigen::MatrixXd &L     = _decMatrixC.matrixL();
    Eigen::MatrixXd        L_inv = Eigen::MatrixXd::Identity(n, n);

    for (int j = 0; j < n; ++j) {
      L_inv(j, j) = 1.0 / L(j, j);
      for (int i = j + 1; i < n; ++i) {
        double sum = 0.0;
        for (int k = j; k < i; ++k) {
          sum += L(i, k) * L_inv(k, j);
        }
        L_inv(i, j) = -sum / L(i, i);
      }
    }
```

| **Method**       | **Vertices** | **computeLOOCV** | **mapData** | **LOOCV/mapData** |
|------------------|--------------|----------------------------------|------------------------------------------------|-------------------|
| **Manual**       | 438          | 13340.0                          | 14416.0                                        | 0.9254            |
| **Eigen(trsm)**  | 438          | 7339.0                           | 8145.0                                         | 0.9013            |
|                  |              |                                  |                                                |                   |
|                  |              |                                  |                                                |                   |
| **Manual**       | 3458         | 14571939.0                       | 14577048.0                                     | 0.9997            |
| **Eigen(trsm)**  | 3458         | 3082969.0                        | 3087632.0                                      | 0.9985            |

  
Hence, I opt for the Eigen-provided variant. Note that an offline-online split (similar to computeMapping and mapData) would be possible and we would be able to boil the computational cost significantly down in the online phase.

## Author's checklist

## Reviewers' checklist


src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -384,7 +384,7 @@ void RadialBasisFctMapping<SOLVER_T, Args...>::mapConsistent(const time::Sample
       for (int i = 0; i < this->input()->getGlobalNumberOfVertices(); i++) {
         in[i] = inputValues[i * valueDim + dim];
       }
-
+      PRECICE_INFO("LOOCV (no polynomial) for vector component {}: {}", dim, _rbfSolver->computeRippaLOOCVerror(in));
```
uekerman, line 387: Could we include the data name here?

uekerman (commented):
This is an important contribution, which should increase usability significantly :+1:

I did not yet look at the code details.

For the moment, you only consider the direct solver, right? The iterative one is a whole different story, maybe not worth investigating at all or what do you think?

> when to compute and report the error metric

I could imagine that we compute the error metric the first n map calls and indeed use and online-offline split. For example:

```xml
<mapping:rbf direction="read" from="A-Mesh" to="B-Mesh" constraint="consistent" loocv-first-n-mappings="5">
  <basis-function:thin-plate-splines />
</mapping:rbf>
```

Naturally, a value 0 would then switch off the feature. We would still need to discuss a default value. In my opinion, we should switch the feature on by default if the overhead is in the same order as the mapping itself.

> Maybe we can even add some related option into the `Mapping` class as we have similar accuracy metrics for NN, NP or LCI

Sounds too artificial in my opinion.

> how do we handle the polynomial in this context?

We should take it into consideration. We should replicate the actual mapping error as well as possible. Then, a user could potentially also test whether using a polynomial is a good idea.

> for our PUM we have to think about a global aggregation method. I would consider to scale each error component by the cluster weight.

You mean for each input vertex, you compute a weighted sum of the different contributions, right? This should work within one rank. What does this mean for distributed meshes?
An alternative could be to simply average all local mapping errors (or a RMS).
Let's look at the literature here as well or ask experts.

src/mapping/RadialBasisFctMapping.hpp
fsimonis: Mapping has no notion of the data anymore.

This info statement gives logs of each map invocation necessary context:
https://github.com/precice/precice/blob/11f3a544ac53f71adf0951df595f02b52ac58e75/src/precice/impl/DataContext.cpp#L135-L137

davidscn:
> For the moment, you only consider the direct solver, right? The iterative one is a whole different story, maybe not worth investigating at all or what do you think?

Maybe not in terms of the math, but we then would need to think about ways of getting the diagonal inverse from an iterative solver, which sounds a bit counter-intuitive.

I implemented now the offline-online split variant, where we compute the inverse diagonal entries in the offline phase and then evaluate the Rippa formula during the online phase. Although there might be room for further performance improvements, ere are some performance metrics:

We have now: the total compute mapping time, including the computation of the inverse diagonal entries, and the total map data time, including the evaluation of the Rippa formula 

[halton-run-computet.pdf](https://github.com/user-attachments/files/16557467/halton-run-computet.pdf)


[halton-run-mapt.pdf](https://github.com/user-attachments/files/16557471/halton-run-mapt.pdf)

and here are the ratios of the LOOCV/computeMappingTime, i.e., the additional overhead for the offline and online phase. I had to make it a log plot (though it is essentially between 0 and 1) to make the differences for the very small meshes visible. The experiment used here is the Halton points.

[halton-run-compute-mapping-ratio.pdf](https://github.com/user-attachments/files/16557490/halton-run-compute-mapping-ratio.pdf)

[halton-run-mapdata-ratio.pdf](https://github.com/user-attachments/files/16557500/halton-run-mapdata-ratio.pdf)

davidscn:
Will merge here, the feature is disabled in the code as the PU version is still an open question.

PR #2065, opened 2024-08-14
Add simple exceptions

### LINKED ISSUE #1189: Let preCICE only exit on critical errors

fsimonis:
## Main changes of this PR

This PR changes the preCICE C++ API to throw a `precice::Error` from `precice/exceptions.hpp` instead of calling `std::abort` on error.
The new header is also part of the `precice/precice.hpp` header.

The C and Fortran APIs internally catch this exception and abort, hence maintaining previous behaviour.

I added tests that construct `Participant` with a missing config, passing null as communicator, and with an incorrect size.
The latter was actually incorrectly checked and lead to an assertion before the check was performed, so I also included the fix.

To use:

```cpp
BOOST_CHECK_THROW(mythingtotest(), ::precice::Error);
```

To check the message too:

```cpp
BOOST_CHECK_EXCEPTION(mythingtotest(), ::precice::Error, [](auto&e){ return strstr(e.what(), "match this") != nullptr;  });
```

We can now finally test failing inputs. 

Part of #280
Replaces #1594
Closes #1189

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/exceptions.hpp
+++ b/src/precice/exceptions.hpp
@@ -1,0 +1,15 @@
+#pragma once
+
+#include <stdexcept>
+
+#include "precice/export.h"
+
+namespace precice {
+
+class PRECICE_API Error : public std::runtime_error {
+public:
+  Error(const std::string &what_arg)
+      : std::runtime_error(what_arg){};
+};
+
+} // namespace precice
--- a/extras/bindings/c/src/preciceC.cpp
+++ b/extras/bindings/c/src/preciceC.cpp
@@ -31,13 +31,15 @@ void precicec_createParticipant_withCommunicator(
     int         solverProcessIndex,
     int         solverProcessSize,
     void *      communicator)
-{
+try {
   PRECICE_CHECK(impl == nullptr, errormsgCreate);
   impl.reset(new precice::Participant(participantName,
                                       configFileName,
                                       solverProcessIndex,
                                       solverProcessSize,
                                       communicator));
+} catch (::precice::Error &e) {
+  std::abort();
 }
 
 void precicec_createParticipant(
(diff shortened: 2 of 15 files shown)


src/precice/exceptions.hpp
davidscn, line 1: Considering that this header is now part of the API, I guess we also need to install the header via CMake, right?

fsimonis: Yes, this is handled via a `PUBLIC_HEADER` property.

https://github.com/fsimonis/precice/blob/d61b1249a37eec5823300b56c483cf5ed65371d1/src/sources.cmake#L357

docs/changelog/2065.md
```diff
@@ -0,0 +1,3 @@
+- Added header `precice/exceptions.hpp` which provides the preCICE exception type `precice::Error`. It is included via `precice/precice.hpp`.
+- Changed the C++ API to throw a `precice::Error` instead of aborting if preCICE detects an error. This allows adapters to clean up on error and makes preCICE easier to use via interactive sessions.
+- Fixed mismatched MPI communicator sizes to trigger an assertion before their sanitization check.
```
BenjaminRodenberg, line 3: Does this belong into this PR? Sounds to me a bit off-topic.

BenjaminRodenberg (approved):
Starting with a very general exception sounds like a good approach. One of the first things we could do now is using this exception system in the python bindings (@NiklasVin).

I also remember that I pointed to the original issue #280 from many places where this exception system would be handy to write an actual test. We should make sure to not forget about this list when merging this PR and closing the corresponding issue.

docs/changelog/2065.md
fsimonis: I fixed this as part of the test that I wrote. Slightly off-topic yes.

fsimonis:
@BenjaminRodenberg IMHO this PR doesn't close #280 as it contains way more concerns like matching different kinds of exceptions in tests.

I would close #1189 though.

BenjaminRodenberg:
> @BenjaminRodenberg IMHO this PR doesn't close #280 as it contains way more concerns like matching different kinds of exceptions in tests.
> 
> I would close #1189 though.

Closing #1189 sounds ok to me. Then we still have #280 open and don't risk forgetting about any of the test cases that require the exception system.

fsimonis:
One change before I merge is to change the header to uppercase to be consistent with the rest.

PR #2066, opened 2024-08-16
Handle event scoping in post

### LINKED ISSUE #1673: Handle event scoping in post-processing
The profiling logic currently allows starting and ending named scopes.
https://github.com/precice/precice/blob/e64ee3c195e8ac07574145980d9dcb6ca7d3139a/src/precice/impl/ParticipantImpl.cpp#L360-L361

These scopes will define a prefix, which is then appended to the event name.
This manual scoping is tedious and error-prone as it changes global state of the EventRegistry.
It is also difficult to keep consistent and comes with a performance penalty as it leads to constantly changing state and more IO.  

The scoping is implied by the nesting of recorded events.
Hence, there is no need to scope manually.
It can be performed completely in post-processing.

We can also keep a stack of active events in the EventsRegistry and apply the scoping before writing the event names.

**Additional context**
#1568

fsimonis:
## Main changes of this PR

This PR removed event scoping from the preCICE library and handles it when post-processing the files in `precice-profiling merge`.

While most events keep the same name, some nested events now appear scoped in the output.

Some examples:

```
advance/map.np.computeMapping.FromA-MeshToB-Mesh/query.index.getVertexIndexTree.A-Mesh
initialize/map.bbm.tagMeshFirstRound.FromA-MeshToB-Mesh/map.np.computeMapping.FromA-MeshToB-Mesh/query.index.getVertexIndexTree.A-Mesh
initialize/m2n.requestSecondaryRanksConnection/m2n.buildCommunicationMap
initialize/m2n.requestSecondaryRanksConnection/m2n.broadcastVertexDistributions
```

While technically correct, this could lead to breaking scripts.

The event registry doesn't need to keep track of a global prefix for events anymore. This prevents bugs when a dev forgets to pop the scope. The name of an event now isn't changed, and name handling is generally easier.

Closes #1673

## Author's checklist

## Reviewers' checklist

* [ ] @davidscn does this interfere with you profiling scripts? What should be adjusted in preCICE?


fsimonis:
@davidscn and I discussed this and tested a real case.
ASTE is now compatible with this change https://github.com/precice/aste/pull/207.
The event names are getting too long though. This PR makes #2117 even more important.

PR #2068, opened 2024-08-16
Conservative unity builds

fsimonis:
## Main changes of this PR

This PR adds support for unity building all integration tests and src module like `mapping` or `com`.

Unit tests are mixing namespaces to an extent that they are a hassle to unit building them.
 
Locally observed build times (Debug) using 5900X:

Setup | user time | total
---|---|---
stock | 1631 | 1:25
unity safe integration tests | 1447 | 1:16
unity all integration tests | 824 | 0:52
unity integration + src modules (this PR) | 538 | 0:46

Roughly comparing to `develop`:

- Bare run 5m to 4m
- MPI run 14m to 8m
- Full run 13-16m to 7-9m
- Total CI 200m to 120m

This should reduce the compile time significantly at the cost of a higher peak memory consumption.

After some testing even for Release builds, peak memory is not significantly different to building before.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -64,6 +64,7 @@ option(PRECICE_ALWAYS_VALIDATE_LIBS "Validate libraries even after the validatat
 option(PRECICE_BINDINGS_C "Enable the native C bindings" ON)
 option(PRECICE_BINDINGS_FORTRAN "Enable the native Fortran bindings" ON)
 option(PRECICE_BUILD_TOOLS "Build the \"precice-tools\" executable" ON)
+option(PRECICE_BUILD_UNITY "Use unity builds in preCICE" ON)
 option(PRECICE_FEATURE_LIBBACKTRACE_STACKTRACES "Enable libbacktrace for stacktrace generation." OFF)
 
 option(CMAKE_INTERPROCEDURAL_OPTIMIZATION "Enable interprocedural optimization for all targets." OFF)
@@ -357,8 +358,6 @@ set_target_properties(preciceCore PROPERTIES
   POSITION_INDEPENDENT_CODE YES
   CXX_VISIBILITY_PRESET default
   VISIBILITY_INLINES_HIDDEN NO
-  UNITY_BUILD Yes
-  UNITY_BUILD_MODE GROUP
   )
 
 if(NOT ${Kokkos_VERSION} VERSION_LESS 4.4)
@@ -523,13 +522,19 @@ configure_file("${PROJECT_SOURCE_DIR}/src/precice/impl/versions.hpp.in" "${PROJE
 configure_file("${PROJECT_SOURCE_DIR}/src/testing/SourceLocation.hpp.in" "${PROJECT_BINARY_DIR}/src/testing/SourceLocation.hpp" @ONLY)
 configure_file("${PROJECT_SOURCE_DIR}/src/precice/Version.h.in" "${PROJECT_BINARY_DIR}/src/precice/Version.h" @ONLY)
 
-get_target_property(_precice_sources preciceCore SOURCES)
-foreach(SOURCE_FILE IN LISTS _precice_sources)
-  if(SOURCE_FILE MATCHES "^src/([a-z0-9]+)/")
-    set_source_files_properties(${SOURCE_FILE} PROPERTIES UNITY_GROUP "${CMAKE_MATCH_1}")
-  endif()
-endforeach()
-unset(_precice_sources)
+if(PRECICE_BUILD_UNITY)
+  set_target_properties(preciceCore PROPERTIES
+    UNITY_BUILD Yes
+    UNITY_BUILD_MODE GROUP
+  )
+  get_target_property(_precice_sources preciceCore SOURCES)
+  foreach(SOURCE_FILE IN LISTS _precice_sources)
+    if(SOURCE_FILE MATCHES "^src/([a-z0-9]+)/")
+      set_source_files_properties(${SOURCE_FILE} PROPERTIES UNITY_GROUP "${CMAKE_MATCH_1}")
+    endif()
+  endforeach()
+  unset(_precice_sources)
+endif()
 
 # Configure version file generation
 include(GenerateVersionInformation)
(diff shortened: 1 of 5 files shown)


fsimonis:
This compiles without issues and comes at a huge compile time advantage.
I'll merge this now and see how it works out in develop.

MakisH:
This is very helpful for building on my few cores laptop :pray: . I observe half the building time, slightly higher total memory usage (nothing to worry about), and a small drop in CPU usage (i.e., indeed less parallelization opportunity):

Checking with `cmake .. && /usr/bin/time make -j 8`

With a previous state (b743fec339746de4acf6581923f472ba4d98db5b):

```
4777.23user 426.20system 11:38.80elapsed 744%CPU (0avgtext+0avgdata 2044652maxresident)k
255176inputs+8235600outputs (582major+53330596minor)pagefaults 0swaps
```

With the latest develop, including this (3517cec7f0bda941c44a53f27d4e9963eaba053c):

```
1763.42user 132.59system 4:42.99elapsed 669%CPU (0avgtext+0avgdata 2255344maxresident)k
0inputs+4428624outputs (26major+16989824minor)pagefaults 0swaps
```

tests/parallel/map-initial-data/helper.hpp
```diff
@@ -1,4 +1,6 @@
 #pragma once
+#ifndef YEO
+#define YEO
```
MakisH, line 3: Probably left-over

fsimonis: 3517cec7f0bda941c44a53f27d4e9963eaba053c

CMakeLists.txt
```diff
@@ -64,6 +64,7 @@ option(PRECICE_ALWAYS_VALIDATE_LIBS "Validate libraries even after the validatat
 option(PRECICE_BINDINGS_C "Enable the native C bindings" ON)
 option(PRECICE_BINDINGS_FORTRAN "Enable the native Fortran bindings" ON)
 option(PRECICE_BUILD_TOOLS "Build the \"precice-tools\" executable" ON)
+option(PRECICE_BUILD_UNITY "Use unity builds in preCICE" ON)
```
davidscn, line 67: At the moment, this only applies to test and I think it would be good to reflect this in the variable description. Also, the name is now similar to BUILD_TOOLS. One might guess that unity is something build by this project, but it is a way to compile objects. So maybe a name like `PRECICE_UNITY_BUILD_TESTING` would be more expressive in my opinion

davidscn:
Introducing this has some implications on how to write code in our integration tests. I would appreciate documentation on this as one might run into unexpected behavior or errors compared to conventional builds.

davidscn:
Also, enabling this by default [is not recommended by CMake](https://cmake.org/cmake/help/latest/prop_tgt/UNITY_BUILD.html)

> Projects should not directly set the UNITY_BUILD property or its associated [CMAKE_UNITY_BUILD](https://cmake.org/cmake/help/latest/variable/CMAKE_UNITY_BUILD.html#variable:CMAKE_UNITY_BUILD) variable to true. Depending on the capabilities of the build machine and compiler used, it might or might not be appropriate to enable unity builds. Therefore, this feature should be under developer control, which would normally be through the developer choosing whether or not to set the [CMAKE_UNITY_BUILD](https://cmake.org/cmake/help/latest/variable/CMAKE_UNITY_BUILD.html#variable:CMAKE_UNITY_BUILD) variable on the [cmake(1)](https://cmake.org/cmake/help/latest/manual/cmake.1.html#manual:cmake(1)) command line or some other equivalent method.

CMakeLists.txt
fsimonis: Unit builds are applied to both, preCICE sources and integration tests.

fsimonis:
I understand that `should be under developer control` means `do not hardcode it to yes`, which it isn't. One can disable it if necessary using the option.

My way of thought for the default is the following:

* Unity builds are more restrictive than normal builds and prone to ODR violations and symbol collisions.
* Unity builds result in noticeably shorter build times.
* Thus developing using unity builds makes sense for both the compile time benefit and to catch ODR violations early in the build process, keeping unity builds functional, whilst preventing the code from failing in the CI.

I would prefer to use the safe and efficient setting until we have a reason to change the default. This is exactly what we can do now on `develop`.

davidscn:
> I understand that should be under developer control means do not hardcode it to yes, which it isn't. One can disable it if necessary using the option.

Well, people might not have the understanding to change this whereas the developer has, so I would consider the default more or less as the same as it hits unexperienced users.

CMakeLists.txt
davidscn: Ok then maybe `PRECICE_UNITY_BUILD` would express the difference I am referring to.

MakisH: Why not `PRECICE_CONFIGURE_UNITY_BUILD`?
I think the point about `PRECICE_BUILD_UNITY` been misunderstood as "build something called Unity" is valid.

MakisH:
What issues exactly can arise from unity builds?
I expected significantly higher memory usage, which does not seem to be the case.

The point about the usefulness of `compile_commands.json` is valid, but mostly for developers.

fsimonis:
> Well, people might not have the understanding to change this whereas the developer has, so I would consider the default more or less as the same as it hits unexperienced users.

Who would this hit though? We test quite a lot of compilers and common OS configurations. If someone has to build preCICE from source on an obscure system then they need to adjust settings anyway.

MinGW, AUR, nix are all handled by package maintainers that can disable unity building in the package.

The only problem I see ATM are the `compile_commands.json`.

davidscn:
> The symbol collisions and the ODR we already check in the CI.

It's true that the CI will complain if there the compiler detects ODR violations, since we use unity builds. However, the violations I am referring to occur due to the unity build, e.g., one uses `using namespace` somewhere and different namespaces contain the same symbol: In an incremental build, this wouldn't be a problem in the first place. I think we should document somewhere the restrictions this comes with.

CMakeLists.txt
fsimonis: I see what you mean now. It is also called `UNITY_BUILD` in CMake, so that would also be consistent.

MakisH:
We should definitely document the option, the default, and the restrictions, but [we should also not overuse `using namespace`](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rs-using). I understand that we mostly have `using namespace` in local scopes, so there should not be any problem ([search finds 215 files](https://github.com/search?q=repo%3Aprecice%2Fprecice+%22using+namespace%22&type=code)).

Or do unity builds somehow mix up the `using namespace` in local scopes?

fsimonis:
Lets move this discussion to https://github.com/precice/precice/issues/2071

PR #2070, opened 2024-08-21
Add support for remeshing of parallel coupling schemes

fsimonis:
## Main changes of this PR

This PR adds support for highly experimental support for remeshing.

As this feature **requires** additional synchronization points, it is off by default.

To enable it, one needs to activate experimental features and allow remeshing globally.

```xml
<precice-configuration experimental="1" allow-remeshing="1" >
```

This allows calling `resetMesh()` and disallows the use of serial coupling schemes as well as coupling scheme compositions and multi coupling schemes.

See #225

## Author's checklist

## TODO

* [x] Disable handshakes when subcycling
* [x] Port or reimplement integration tests from the previous version(2)
* [x] Add tests for parallel implicit coupling
* [ ] ~~Add tests for multi coupling~~
* [x] Cleanup history

### DIFF
--- a/src/acceleration/BaseQNAcceleration.cpp
+++ b/src/acceleration/BaseQNAcceleration.cpp
@@ -90,6 +90,7 @@ void BaseQNAcceleration::initialize(
 
   checkDataIDs(cplData);
   // store all data IDs in vector
+  _dataIDs.clear();
   for (const DataMap::value_type &pair : cplData) {
     _dataIDs.push_back(pair.first);
   }
@@ -99,6 +100,7 @@ void BaseQNAcceleration::initialize(
                   "Quasi-Newton acceleration does not yet support using data from all substeps. Please set substeps=\"false\" in the exchange tag of data \"{}\".", data->getDataName());
   }
 
+  _matrixCols.clear();
   _matrixCols.push_front(0);
   _firstIteration  = true;
   _firstTimeWindow = true;
@@ -610,8 +612,6 @@ void BaseQNAcceleration::initializeVectorsAndPreconditioner(const DataMap &cplDa
   const size_t primaryDataSize = std::accumulate(_primaryDataIDs.begin(), _primaryDataIDs.end(), (size_t) 0, addCplDataSize);
   const size_t dataSize        = std::accumulate(_dataIDs.begin(), _dataIDs.end(), (size_t) 0, addCplDataSize);
 
-  PRECICE_ASSERT(_oldPrimaryXTilde.size() == 0);
-  PRECICE_ASSERT(_oldPrimaryResiduals.size() == 0);
   _oldPrimaryXTilde    = Eigen::VectorXd::Zero(primaryDataSize);
   _oldPrimaryResiduals = Eigen::VectorXd::Zero(primaryDataSize);
   _primaryResiduals    = Eigen::VectorXd::Zero(primaryDataSize);
@@ -621,6 +621,8 @@ void BaseQNAcceleration::initializeVectorsAndPreconditioner(const DataMap &cplDa
   _oldValues           = Eigen::VectorXd::Zero(dataSize);
   _oldXTilde           = Eigen::VectorXd::Zero(dataSize);
   _residuals           = Eigen::VectorXd::Zero(dataSize);
+  _matrixV.resize(0, 0);
+  _matrixW.resize(0, 0);
 
   /**
    *  make dimensions public to all procs,
@@ -673,6 +675,7 @@ void BaseQNAcceleration::initializeVectorsAndPreconditioner(const DataMap &cplDa
   }
 
   // set the number of global rows in the QRFactorization.
+  _qrV.reset();
   _qrV.setGlobalRows(getPrimaryLSSystemRows());
 
   std::vector<size_t> subVectorSizes; // needed for preconditioner
--- a/src/acceleration/impl/Preconditioner.hpp
+++ b/src/acceleration/impl/Preconditioner.hpp
@@ -40,7 +40,6 @@ public:
   {
     PRECICE_TRACE();
 
-    PRECICE_ASSERT(_weights.empty());
     _subVectorSizes = svs;
 
     size_t N = std::accumulate(_subVectorSizes.begin(), _subVectorSizes.end(), static_cast<std::size_t>(0));
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -298,6 +298,27 @@ void BaseCouplingScheme::initialize()
   _isInitialized = true;
 }
 
+void BaseCouplingScheme::reinitialize()
+{
+  PRECICE_TRACE();
+  PRECICE_ASSERT(isInitialized());
+
+  if (isImplicitCouplingScheme()) {
+    // overwrite past iteration with new samples
+    for (const auto &data : _allData | boost::adaptors::map_values) {
+      // TODO: reset CouplingData of changed meshes only #2102
+      data->reinitialize();
+    }
+
+    if (not doesFirstStep()) {
+      if (_acceleration) {
+        _acceleration->initialize(getAccelerationData());
+      }
+    }
+    initializeTXTWriters();
+  }
+}
+
 bool BaseCouplingScheme::sendsInitializedData() const
 {
   return _sendsInitializedData;
--- a/src/cplscheme/config/CouplingSchemeConfiguration.cpp
+++ b/src/cplscheme/config/CouplingSchemeConfiguration.cpp
@@ -322,6 +322,7 @@ void CouplingSchemeConfiguration::xmlEndTagCallback(
   PRECICE_TRACE(tag.getFullName());
   if (tag.getNamespace() == TAG) {
     if (_config.type == VALUE_SERIAL_EXPLICIT) {
+      PRECICE_CHECK(!_allowRemeshing, "Remeshing is currently incompatible with serial coupling schemes. Try using a parallel or a multi coupling scheme instead.");
       std::string       accessor(_config.participants[0]);
       PtrCouplingScheme scheme = createSerialExplicitCouplingScheme(accessor);
       addCouplingScheme(scheme, accessor);
@@ -342,6 +343,7 @@ void CouplingSchemeConfiguration::xmlEndTagCallback(
       //_couplingSchemes[accessor] = scheme;
       _config = Config();
     } else if (_config.type == VALUE_SERIAL_IMPLICIT) {
+      PRECICE_CHECK(!_allowRemeshing, "Remeshing is currently incompatible with serial coupling schemes. Try using a parallel or a multi coupling scheme instead.");
       updateConfigForImplicitCoupling();
       std::string       accessor(_config.participants[0]);
       PtrCouplingScheme scheme = createSerialImplicitCouplingScheme(accessor);
@@ -353,6 +355,7 @@ void CouplingSchemeConfiguration::xmlEndTagCallback(
       //_couplingSchemes[accessor] = scheme;
       _config = Config();
     } else if (_config.type == VALUE_PARALLEL_IMPLICIT) {
+      PRECICE_INFO_IF(_allowRemeshing, "Remeshing for implicit coupling schemes is in development. Currently, the acceleration data is deleted on remeshing.");
       updateConfigForImplicitCoupling();
       std::string       accessor(_config.participants[0]);
       PtrCouplingScheme scheme = createParallelImplicitCouplingScheme(accessor);
@@ -362,6 +365,9 @@ void CouplingSchemeConfiguration::xmlEndTagCallback(
       addCouplingScheme(scheme, accessor);
       _config = Config();
     } else if (_config.type == VALUE_MULTI) {
+      /// TODO test multi coupling scheme
+      PRECICE_CHECK(!_allowRemeshing, "Remeshing is currently incompatible with multi coupling schemes. Try using a parallel coupling scheme instead.");
+      PRECICE_INFO_IF(_allowRemeshing, "Remeshing for implicit coupling schemes is in development. Currently, the acceleration data is deleted on remeshing.");
       updateConfigForImplicitCoupling();
       PRECICE_CHECK(_config.setController,
                     "One controller per MultiCoupling needs to be defined. "
@@ -390,6 +396,8 @@ void CouplingSchemeConfiguration::addCouplingScheme(
     return;
   }
   PRECICE_ASSERT(_couplingSchemes.count(participantName) > 0);
+
+  PRECICE_CHECK(!_allowRemeshing, "Remeshing is currently incompatible with compositional coupling schemes. If you need remeshing, try using a multi coupling scheme to compose your participants.");
 
   // Create a composition to add the new cplScheme to
   if (!utils::contained(participantName, _couplingSchemeCompositions)) {
@@ -1198,6 +1206,12 @@ void CouplingSchemeConfiguration::setParallelAcceleration(
         "Consider switching to a serial-implicit coupling scheme or changing the acceleration method.",
         participant);
   }
+}
+
+void CouplingSchemeConfiguration::setRemeshing(
+    bool allowed)
+{
+  _allowRemeshing = allowed;
 }
 
 } // namespace precice::cplscheme
--- a/src/mesh/Mesh.cpp
+++ b/src/mesh/Mesh.cpp
@@ -263,6 +263,7 @@ void Mesh::clear()
 
   for (mesh::PtrData &data : _data) {
     data->values().resize(0);
+    data->timeStepsStorage().clear();
   }
 }
 
--- a/src/partition/ReceivedPartition.cpp
+++ b/src/partition/ReceivedPartition.cpp
@@ -85,6 +85,8 @@ void ReceivedPartition::communicate()
     PRECICE_ASSERT(globalNumberOfVertices >= 0);
     _mesh->setGlobalNumberOfVertices(globalNumberOfVertices);
   }
+
+  PRECICE_ASSERT(_mesh->getGlobalNumberOfVertices() >= 0);
 }
 
 void ReceivedPartition::compute()
@@ -209,7 +211,6 @@ void ReceivedPartition::compute()
       PRECICE_DEBUG("Send partition feedback to primary rank");
       utils::IntraComm::getCommunication()->sendRange(vertexIDs, 0);
     } else { // Primary
-
       mesh::Mesh::VertexDistribution vertexDistribution;
       int                            numberOfVertices = _mesh->nVertices();
       std::vector<VertexID>          vertexIDs(numberOfVertices, -1);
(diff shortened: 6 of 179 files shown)


src/partition/ReceivedPartition.cpp
```diff
@@ -262,6 +263,13 @@ void ReceivedPartition::compute()
     PRECICE_ASSERT(_mesh->getVertexOffsets().empty());
     _mesh->setVertexOffsets(std::move(vertexOffsets));
   }
+
+  PRECICE_ASSERT(!_mesh->getVertexOffsets().empty());
+  /* TODO this triggers even though it shouldn't. Asserting in the branch above after setting it does not trigger
+  if (m2n().usesTwoLevelInitialization() && utils::IntraComm::isPrimary()) {
+    PRECICE_ASSERT(!_mesh->getVertexDistribution().empty());
+  }
```
fsimonis, line 271: This is concerning.

This assertion works fine after l227. So either there is some UB leading so nonsense or my condition evaluates to something else.

src/testing/QuickTest.hpp
```diff
@@ -0,0 +1,86 @@
...
+  {
+    auto                n = vertexIDs.size() * interface.getDataDimensions(meshName, dataName);
+    std::vector<double> result(n, -0.0);
+    interface.readData(meshName, dataName, vertexIDs, interface.getMaxTimeStepSize(), result);
+    return result;
+  }
+
+  Participant &         interface;
+  std::string           meshName;
+  std::string           dataName;
```
fsimonis, line 72: For implicit coupling I need to send and receive different data, so I need to distinguish between read and write data here.

fsimonis: Done and migrated everything

fsimonis:
State of this PR: coupling and repartitioning is tested for parallel implicit and explicit schemes.

Untested are:
* convergence measures
* acceleration
* multi-coupling schemes

tests/remeshing/ResetWhileIterating.cpp
```diff
@@ -0,0 +1,57 @@
...
+  }();
+
+  qt.setVertices({0.0, y, 1.0, y})
+      .initialize()
+      // TW 1 It 1 currently not allowed
+      .writeCheckpoint();
+  BOOST_CHECK_THROW(qt.resetMesh(), ::precice::Error);
+  qt.readCheckpoint()
+      .advance()
+      // TW 1 It
```
uekerman, line 30:
```suggestion
      // TW 1 It 2
```

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -369,7 +378,15 @@ void ParticipantImpl::advance(
...
   // Update the coupling scheme time state. Necessary to get correct remainder.
-  const bool   isAtWindowEnd = _couplingScheme->addComputedTime(computedTimeStepSize);
+  const bool isAtWindowEnd = _couplingScheme->addComputedTime(computedTimeStepSize);
+
+  if (_allowsRemeshing && isAtWindowEnd) {
+    int totalMeshesChanges = getTotalMeshChanges();
+    if (reinitHandshake(totalMeshesChanges)) {
+      reinitialize();
+    }
+  }
```
uekerman, line 388: We could add an assertion here, right?

```suggestion
  if (_allowsRemeshing) {
    int totalMeshesChanges = getTotalMeshChanges();
    if (isAtWindowEnd){
      if (reinitHandshake(totalMeshesChanges)) {
        reinitialize();
      }
    else {
      assert(totalMeshesChanges == 0);
    }
  }
```

Maybe overkill, but could also help in understanding how things work.

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -644,8 +661,12 @@ void ParticipantImpl::resetMesh(
     std::string_view meshName)
 {
   PRECICE_EXPERIMENTAL_API();
+  PRECICE_CHECK(_allowsRemeshing, "Cannot reset meshes. This feature needs to be enabled using <precice-configuration experimental=\"1\" allow-remeshing=\"1\">.");
+  PRECICE_CHECK(_state == State::Initialized, "initialize() has to be called before resetMesh().");
   PRECICE_TRACE(meshName);
   PRECICE_VALIDATE_MESH_NAME(meshName);
+  PRECICE_CHECK(_couplingScheme->isCouplingOngoing(), "Cannot remesh while subcycling or iterating. Remeshing is only allowed when the time window is completed.");
```
uekerman, line 668: Wrong error message

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -1618,4 +1643,67 @@ ParticipantImpl::MappedSamples ParticipantImpl::mappedSamples() const
   return res;
 }
 
+// Reinitialization
+
+int ParticipantImpl::getTotalMeshChanges() const
+{
+  PRECICE_TRACE();
+  PRECICE_ASSERT(_allowsRemeshing);
+  Event e("intra-handshake", profiling::Synchronize);
```
uekerman, line 1652:
```suggestion
  Event e("reinit-intra-handshake", profiling::Synchronize);
```

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -1618,4 +1643,67 @@ ParticipantImpl::MappedSamples ParticipantImpl::mappedSamples() const
...
+  utils::IntraComm::allreduceSum(localMeshesChanges, totalMeshesChanges);
+  PRECICE_DEBUG("Total Mesh Changes: {}", totalMeshesChanges);
+  return totalMeshesChanges;
+}
+
+bool ParticipantImpl::reinitHandshake(bool requestReinit) const
+{
+  PRECICE_TRACE();
+  PRECICE_ASSERT(_allowsRemeshing);
+  Event e("inter-handshake", profiling::Synchronize);
```
uekerman, line 1666:
```suggestion
  Event e("reinit-inter-handshake", profiling::Synchronize);
```

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -1618,4 +1643,67 @@ ParticipantImpl::MappedSamples ParticipantImpl::mappedSamples() const
...
+  int   localMeshesChanges = _meshLock.countUnlocked();
+  PRECICE_DEBUG("Local Mesh Changes: {}", localMeshesChanges);
+
+  int totalMeshesChanges = 0;
+  utils::IntraComm::allreduceSum(localMeshesChanges, totalMeshesChanges);
+  PRECICE_DEBUG("Total Mesh Changes: {}", totalMeshesChanges);
+  return totalMeshesChanges;
+}
+
+bool ParticipantImpl::reinitHandshake(bool requestReinit) const
```
uekerman, line 1662: We currently still reinit all m2ns, right? Not only those, which participants share a reset mesh?

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -1618,4 +1643,67 @@ ParticipantImpl::MappedSamples ParticipantImpl::mappedSamples() const
...
+      if (iter.second.isRequesting) {
+        comm.send(requestReinit, 0);
+        comm.receive(received, 0);
+      } else {
+        comm.receive(received, 0);
+        comm.send(requestReinit, 0);
+      }
+      swarmReinitRequired |= received;
+    }
+    PRECICE_DEBUG("Result of Phase 1 -{} reinit required.", (swarmReinitRequired ? "" : " no"));
```
uekerman, line 1686: If this is Phase 1, what is Phase 2?

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -321,6 +321,7 @@ void CouplingSchemeConfiguration::xmlEndTagCallback(
   PRECICE_TRACE(tag.getFullName());
   if (tag.getNamespace() == TAG) {
     if (_config.type == VALUE_SERIAL_EXPLICIT) {
+      PRECICE_CHECK(!_allowRemeshing, "Remeshing is currently incompatible with serial coupling schemes. Try using a parallel or multi couplingscheme instead.");
```
uekerman, line 324:
```suggestion
      PRECICE_CHECK(!_allowRemeshing, "Remeshing is currently incompatible with serial coupling schemes. Try using a parallel or a multi coupling scheme instead.");
```

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -341,6 +342,7 @@ void CouplingSchemeConfiguration::xmlEndTagCallback(
       //_couplingSchemes[accessor] = scheme;
       _config = Config();
     } else if (_config.type == VALUE_SERIAL_IMPLICIT) {
+      PRECICE_CHECK(!_allowRemeshing, "Remeshing is currently incompatible with serial coupling schemes. Try using a parallel or multi couplingscheme instead.");
```
uekerman, line 345:
```suggestion
      PRECICE_CHECK(!_allowRemeshing, "Remeshing is currently incompatible with serial coupling schemes. Try using a parallel or a multi coupling scheme instead.");
```

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -390,6 +394,8 @@ void CouplingSchemeConfiguration::addCouplingScheme(
   }
   PRECICE_ASSERT(_couplingSchemes.count(participantName) > 0);
 
+  PRECICE_CHECK(!_allowRemeshing, "Remeshing is currently incompatible with compositional coupling schemes. If you need remeshing, try using a multi couplingscheme to compose your participants.");
```
uekerman, line 397:
```suggestion
  PRECICE_CHECK(!_allowRemeshing, "Remeshing is currently incompatible with compositional coupling schemes. If you need remeshing, try using a multi coupling scheme to compose your participants.");
```

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -297,6 +297,25 @@ void BaseCouplingScheme::initialize()
...
+    if (not doesFirstStep()) {
+      // reserve memory and initialize data with zero
+      if (_acceleration) {
+        _acceleration->initialize(getAccelerationData());
+      }
+    }
+    initializeTXTWriters();
+  }
+
+  // exchangeInitialData();
```
uekerman, line 316:
```suggestion
```
This should be clear, no?

src/testing/QuickTest.hpp
uekerman, line 1: Probably good to make adding this more visible. Maybe in a separate PR?
I think it would also be good to have some brief docs here: https://precice.org/dev-docs-dev-testing.html

uekerman (commented):
Looks pretty clean already :ok_hand: 
I mainly looked at the implementation. I only skimmed through the tests. You are currently "only" testing 2 participants with one mesh each, correct?

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -311,32 +313,39 @@ void ParticipantImpl::initialize()
     watchIntegral->initialize();
   }
 
-  _meshLock.lockAll();
+  if (reinitializing) {
+    PRECICE_DEBUG("Reinitialize coupling schemes");
+    _couplingScheme->reinitialize();
+  } else {
+    _meshLock.lockAll();
```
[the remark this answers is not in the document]
fsimonis: The end of advance locks all meshes.

I would like to keep the locking information around until we figure out SerialCouplingSchemes.

src/precice/impl/ParticipantImpl.cpp
fsimonis: These were placeholder names and should be changed.

What about
```suggestion
  Event e("coordinate-local-mesh-changes", profiling::Synchronize);
```

src/precice/impl/ParticipantImpl.cpp
fsimonis:
```suggestion
  Event e("remesh.exchangeRemoteMeshChanges", profiling::Synchronize);
```

src/precice/impl/ParticipantImpl.cpp
fsimonis: Yes, this is currently restricted in the configuration though.

src/precice/impl/ParticipantImpl.cpp
fsimonis: Fixed. This is the "no remesh after final time window"-check.

src/precice/impl/ParticipantImpl.cpp
fsimonis: Clarified

src/precice/impl/ParticipantImpl.hpp
```diff
@@ -453,6 +456,10 @@ class ParticipantImpl {
   /// Discards send (currently write) data of a participant after a given time when another iteration is required
   void trimSendDataAfter(double time);
 
+  int  getTotalMeshChanges() const;
+  bool reinitHandshake(bool requestReinit) const;
+  void reinitialize();
```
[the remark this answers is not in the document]
fsimonis: Done

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -242,8 +243,9 @@ void ParticipantImpl::initialize()
 {
   PRECICE_TRACE();
   PRECICE_CHECK(_state != State::Finalized, "initialize() cannot be called after finalize().");
-  PRECICE_CHECK(_state != State::Initialized, "initialize() may only be called once.");
-  PRECICE_ASSERT(not _couplingScheme->isInitialized());
+  // PRECICE_CHECK(_state != State::Initialized, "initialize() may only be called once.");
+  // PRECICE_ASSERT(not _couplingScheme->isInitialized());
```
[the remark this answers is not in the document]
fsimonis: I refactored common initialization parts out into a separate function, which allows enabling these calls.

src/precice/impl/ParticipantImpl.cpp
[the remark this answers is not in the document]
fsimonis: This is easier to follow after the refactoring

fsimonis:
Another use-case that came up at the workshop was combining `resetMesh` with direct-access to change the area of interest at runtime.

This should work, but is currently impossible as `resetMesh` can only be called for provided meshes.

fsimonis:
Convergence measures work now.

This currently doesn't support subcycling.

Problem is that on remeshing, we know that the samples from the last time window including the last sample represent the old mesh. So they need to be reset to at least be 0 of the correct shape.
The issue is that the reinitialization occurs before data is exchanged and before the written sample is written to the data (WriteDataContext).

Based on the new size of the mesh and the data dimensions, I can create a zero sample of correct shape and overwrite the current samples as well as the samples of the last iteration, aka timewindow. The latter seems a bit weird though.

fsimonis:
Acceleration schemes work too. I tested them all manually for now. I'll add tests next week.

tests/remeshing/parallel/implicit/acceleration/helper.hpp
```diff
@@ -16,17 +16,18 @@ using testing::operator""_write;
 
 namespace acceleration {
 
-inline void runResetA(testing::TestContext &context)
+// The data written in the acceleration tests are always 40, 30, 15
+// first and second are read data from the first and second iteration
+// The end of a timewindow and the end of the simulation are never accelerated and thus 15
```
fsimonis, line 21: @uekerman what are your thoughts on this testing method?

uekerman: Should have a certain test coverage already, yes. In a single time window IQN-ILS and IQN-IMVJ are identical (as you also observed). To really test IQN-IMVJ, you could only remesh every second time window.

fsimonis: Added a second time window before and after the reinit.
Despite the same data, the IQNs result in different data as the initial data is once 0 and once the result of the 2nd time window.

fsimonis:
Acceleration tests added.
Doing some passes regarding documentation.

src/partition/ReceivedPartition.cpp
fsimonis: Wrong condition. This only happens without 2L1.

fsimonis:
Untested are:

* Multiple meshes where only one is remeshed. The implementation currently throws all data away.
* Subcycling. This may not work as remeshing is only allowed in the first iteration but the remeshing is performed in the last iteration. The information about which mesh was reset/unlocked needs to be sticky until the end of the time window.

uekerman (commented):
Looked at the implementation again. Looks good :+1:
Only minor remarks below.

I still want to look at the tests asap.

uekerman (approved):
Tests look good. Impossible to review everything, but I think I understood the overall layout.

`tests/remeshing/parallel/...` -> the parallel here stands for parallel coupling, right? This is a bit inconsistent with other integration tests. Let's better use `parallel-coupling` or similar.

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -369,7 +378,15 @@ void ParticipantImpl::advance(
 #endif
 
   // Update the coupling scheme time state. Necessary to get correct remainder.
-  const bool   isAtWindowEnd = _couplingScheme->addComputedTime(computedTimeStepSize);
+  const bool isAtWindowEnd = _couplingScheme->addComputedTime(computedTimeStepSize);
```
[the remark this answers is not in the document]
fsimonis: Done https://github.com/precice/precice/issues/2122

src/mesh/Mesh.cpp
```diff
@@ -263,6 +263,13 @@ void Mesh::clear()
 
   for (mesh::PtrData &data : _data) {
     data->values().resize(0);
+
+    auto &tts = data->timeStepsStorage();
+    if (!tts.empty()) {
+      auto tl = tts.last().timestamp;
+      tts.clear();
+      tts.setSampleAtTime(tl, data->sample());
```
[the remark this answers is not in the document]
fsimonis: Agreed and done.

src/acceleration/impl/Preconditioner.hpp
```diff
@@ -40,7 +40,6 @@ class Preconditioner {
   {
     PRECICE_TRACE();
 
-    PRECICE_ASSERT(_weights.empty());
```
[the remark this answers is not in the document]
fsimonis: The only purpose of this assert is preventing initialize from being called again as the weights are reset below.

So removing it isn't an issue.

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -115,12 +115,11 @@ void BaseQNAcceleration::initialize(
     }
   }
 
+  _matrixCols.clear();
```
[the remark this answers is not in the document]
fsimonis: If I move these calls in some reset function, then the initialization is lacking the context of which vector is actually empty or not.
While nice in theory, this will lead to a lot of jumping back and forth, especially as the QN base has so many members.

fsimonis:
> Let's better use parallel-coupling or similar.

I merged the `Parallel` and `Implicit`/`Explicit` suites into `ParallelExplicit` and `ParallelImplicit`. That should be clear enough, right?

src/acceleration/BaseQNAcceleration.cpp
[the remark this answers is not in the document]
fsimonis: We concluded to clean this up when we do the next refactoring pass of the Acceleration hierarchy.

src/precice/impl/ParticipantImpl.cpp
fsimonis: Done

src/precice/impl/ParticipantImpl.cpp
fsimonis: Done

fsimonis:
I'll compress the history now. Then we are ready to merge.

PR #2072, opened 2024-08-29
Allow multiple direct mesh accesses per mesh

davidscn:
## Main changes of this PR

Allows multiple direct mesh accesses per mesh.

In #1033 we restricted the call of `setMeshAccessRegion` to a single call per participant per rank. I don't remember the motivation for this restriction, but it is highly inconsistent with the rest of our configuration. I don't see a clear reason for this restriction and there might be very well use cases for this, which would look like

```xml
    <receive-mesh name="MeshTwo" from="SolverTwo" direct-access="true" />
    <receive-mesh name="MeshTwoB" from="SolverTwo" direct-access="true" />
```

i.e., a participant wants to access different meshes in a direct way. I have a test case using this for the indirect mesh access. The PR here is essentially an isolated fix, which was fixed as part of the (yet to come) indirect mesh access.

## Author's checklist

* [x] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits. -> formatting is broken on Ubuntu 24.04
* [x] I added a test to cover the proposed changes in our test suite. -> coming with indirect access

## Reviewers' checklist

### DIFF
--- a/src/precice/impl/ParticipantImpl.cpp
+++ b/src/precice/impl/ParticipantImpl.cpp
@@ -1153,10 +1153,11 @@ void ParticipantImpl::setMeshAccessRegion(
   PRECICE_REQUIRE_MESH_USE(meshName);
   PRECICE_CHECK(_state != State::Finalized, "setMeshAccessRegion() cannot be called after finalize().");
   PRECICE_CHECK(_state != State::Initialized, "setMeshAccessRegion() needs to be called before initialize().");
-  PRECICE_CHECK(!_accessRegionDefined, "setMeshAccessRegion may only be called once.");
 
   // Get the related mesh
   MeshContext & context = _accessor->meshContext(meshName);
+
+  PRECICE_CHECK(!context.accessRegionDefined, "setMeshAccessRegion was already defined for mesh \"{}\" and may only be called once per mesh.", context.mesh->getName());
   mesh::PtrMesh mesh(context.mesh);
   int           dim = mesh->getDimensions();
   PRECICE_CHECK(boundingBox.size() == static_cast<unsigned long>(dim) * 2,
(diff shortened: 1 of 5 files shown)


src/precice/impl/ParticipantImpl.cpp
```diff
@@ -1153,10 +1153,11 @@ void ParticipantImpl::setMeshAccessRegion(
   PRECICE_REQUIRE_MESH_USE(meshName);
   PRECICE_CHECK(_state != State::Finalized, "setMeshAccessRegion() cannot be called after finalize().");
   PRECICE_CHECK(_state != State::Initialized, "setMeshAccessRegion() needs to be called before initialize().");
-  PRECICE_CHECK(!_accessRegionDefined, "setMeshAccessRegion may only be called once.");
 
   // Get the related mesh
-  MeshContext & context = _accessor->meshContext(meshName);
+  MeshContext &context = _accessor->meshContext(meshName);
+
+  PRECICE_CHECK(!context.accessRegionDefined, "setMeshAccessRegion was already defined for mesh \"{}\" and may only be called once per mesh.", context.mesh->getName());
```
uekerman, line 1160:
```suggestion
  PRECICE_CHECK(!context.accessRegionDefined, "A mesh access region was already defined for mesh \"{}\". setMeshAccessRegion may only be called once per mesh.", context.mesh->getName());
```

uekerman (approved):
Makes sense to me.
Anything we need to update in the documentation?

davidscn:
>  Anything we need to update in the documentation?

I couldn't spot anything else.

fsimonis:
Maybe a mention that version 3.2.0 allows this for multiple meshes per participant would be nice in the docs https://precice.org/couple-your-code-direct-access

PR #2073, opened 2024-09-05
Refactor Quasi-Newton to make it more compatible with Waveform iterations

NiklasKotarsky:
## Main changes of this PR

Moves concatenate coupling data into Aitken and BaseQNAcceleration to make it easier to first only support waveform iterations in the Quasi-Newton methods

Moves the initialization of vectors, matrices and preconditioners into their own function. The reason behind this change is that the vector and preconditioner initialization needs to know the time grid of the Quasi-Newton method, which is determined after the first waveform iteration.

Fixes some small errors

This pull request refactors Quasi-Newton methods to make them more compatible with Waveform iterations and to make #2005 easier to review.

## Author's checklist

* [ ] I sticked to CMake version 3.22.1.

## Reviewers' checklist

### DIFF
--- a/src/acceleration/Acceleration.hpp
+++ b/src/acceleration/Acceleration.hpp
@@ -64,9 +64,6 @@ protected:
   /// Checks if all dataIDs are contained in cplData
   void checkDataIDs(const DataMap &cplData) const;
 
-  /// Concatenates all coupling data involved into a single vector
-  void concatenateCouplingData(const DataMap &cplData, const std::vector<DataID> &dataIDs, Eigen::VectorXd &targetValues, Eigen::VectorXd &targetOldValues) const;
-
   /// performs a relaxation given a relaxation factor omega
   static void applyRelaxation(double omega, DataMap &cplData);
 };
--- a/src/acceleration/Acceleration.hpp
+++ b/src/acceleration/Acceleration.hpp
@@ -66,6 +66,10 @@ protected:
 
   /// performs a relaxation given a relaxation factor omega
   static void applyRelaxation(double omega, DataMap &cplData);
+
+  /// @brief Concatenates the data and old data in cplData into two long vectors
+  virtual void concatenateCouplingData(
+      const DataMap &cplData, const std::vector<DataID> &dataIDs, Eigen::VectorXd &targetValues, Eigen::VectorXd &targetOldValues) const = 0;
 };
 } // namespace acceleration
 } // namespace precice
--- a/src/acceleration/AitkenAcceleration.hpp
+++ b/src/acceleration/AitkenAcceleration.hpp
@@ -37,7 +37,7 @@ public:
       const DataMap &cpldata);
 
 protected:
-  /// @brief Concatenates the data and old data in cplData into two long vectors
+  /// @copydoc acceleration::Acceleration::concatenateCouplingData
   void concatenateCouplingData(
       const DataMap &cplData, const std::vector<DataID> &dataIDs, Eigen::VectorXd &targetValues, Eigen::VectorXd &targetOldValues) const override final;
 
--- a/src/acceleration/AitkenAcceleration.hpp
+++ b/src/acceleration/AitkenAcceleration.hpp
@@ -36,6 +36,11 @@ public:
   virtual void iterationsConverged(
       const DataMap &cpldata);
 
+protected:
+  /// @brief Concatenates the data and old data in cplData into two long vectors
+  void concatenateCouplingData(
+      const DataMap &cplData, const std::vector<DataID> &dataIDs, Eigen::VectorXd &targetValues, Eigen::VectorXd &targetOldValues) const;
+
 private:
   logging::Logger _log{"acceleration::AitkenAcceleration"};
 
--- a/src/acceleration/AitkenAcceleration.hpp
+++ b/src/acceleration/AitkenAcceleration.hpp
@@ -39,7 +39,7 @@ public:
 protected:
   /// @brief Concatenates the data and old data in cplData into two long vectors
   void concatenateCouplingData(
-      const DataMap &cplData, const std::vector<DataID> &dataIDs, Eigen::VectorXd &targetValues, Eigen::VectorXd &targetOldValues) const;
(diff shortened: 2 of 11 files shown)


src/acceleration/Acceleration.hpp
```diff
@@ -64,9 +64,6 @@ class Acceleration {
   /// Checks if all dataIDs are contained in cplData
   void checkDataIDs(const DataMap &cplData) const;
 
-  /// Concatenates all coupling data involved into a single vector
-  void concatenateCouplingData(const DataMap &cplData, const std::vector<DataID> &dataIDs, Eigen::VectorXd &targetValues, Eigen::VectorXd &targetOldValues) const;
```
BenjaminRodenberg, line 68: I think we should keep this as `virtual void concatenateCouplingData(...) const = 0;` to make the relationship of `AitkenAcceleration::concatenateCouplingData` and `BaseQNAcceleration::concatenateCouplingData` clear. Compare to how some functions in the `BaseCouplingScheme` are implemented (e.g. [`hasAnySendData`](https://github.com/precice/precice/blob/15cdcbdf72fea5c06c996cc1c2e0f0b4e137916d/src/cplscheme/BaseCouplingScheme.hpp#L209))

src/acceleration/AitkenAcceleration.hpp
```diff
@@ -36,6 +36,11 @@ class AitkenAcceleration : public Acceleration {
   virtual void iterationsConverged(
       const DataMap &cpldata);
 
+protected:
+  /// @brief Concatenates the data and old data in cplData into two long vectors
+  void concatenateCouplingData(
+      const DataMap &cplData, const std::vector<DataID> &dataIDs, Eigen::VectorXd &targetValues, Eigen::VectorXd &targetOldValues) const;
```
BenjaminRodenberg, line 42:
```suggestion
  void concatenateCouplingData(
      const DataMap &cplData, const std::vector<DataID> &dataIDs, Eigen::VectorXd &targetValues, Eigen::VectorXd &targetOldValues) const override final;
```
Use `override final`. Compare to ([`hasAnySendData`](https://github.com/precice/precice/blob/15cdcbdf72fea5c06c996cc1c2e0f0b4e137916d/src/cplscheme/BiCouplingScheme.hpp#L65))

src/acceleration/BaseQNAcceleration.hpp
```diff
@@ -262,14 +262,18 @@ class BaseQNAcceleration : public Acceleration {
...
   virtual void removeMatrixColumn(int columnIndex);
 
-  /// Wwrites info to the _infostream (also in parallel)
+  /// Writes info to the _infostream (also in parallel)
   void writeInfo(const std::string &s, bool allProcs = false);
 
+  /// @brief Concatenates the data and old data in cplData into two long vectors
+  void concatenateCouplingData(
+      const DataMap &cplData, const std::vector<DataID> &dataIDs, Eigen::VectorXd &targetValues, Eigen::VectorXd &targetOldValues) const;
+
```
BenjaminRodenberg, line 276:
```suggestion
  void concatenateCouplingData(
      const DataMap &cplData, const std::vector<DataID> &dataIDs, Eigen::VectorXd &targetValues, Eigen::VectorXd &targetOldValues) const override final;

```
See above

src/acceleration/Acceleration.hpp
```diff
@@ -64,9 +64,6 @@ class Acceleration {
   /// Checks if all dataIDs are contained in cplData
   void checkDataIDs(const DataMap &cplData) const;
 
-  /// Concatenates all coupling data involved into a single vector
```
BenjaminRodenberg, line 67: Please keep the documentation here and use `@copydoc` in the child classes. Compare [here](https://github.com/precice/precice/blob/15cdcbdf72fea5c06c996cc1c2e0f0b4e137916d/src/cplscheme/BiCouplingScheme.hpp#L91)

BenjaminRodenberg (changes requested):
Just a minor thing regarding inheritance. Rest looks good should be good to merge if the tests work :+1:

src/acceleration/AitkenAcceleration.hpp
```diff
@@ -36,6 +36,11 @@ class AitkenAcceleration : public Acceleration {
   virtual void iterationsConverged(
       const DataMap &cpldata);
 
+protected:
```
uekerman, line 39: Could go to `private` here, no?

uekerman (approved):
Other a minor comment below and @BenjaminRodenberg' comments, the refactoring looks correct.

I assume that the motivation is that you later want to modify `BaseQNAcc::concatenateCouplingData`, right?

> Moves the initialization of vectors, matrices and preconditioners into their own function.

This you have not yet done or am I missing sth?

NiklasKotarsky:
> Other a minor comment below and @BenjaminRodenberg' comments, the refactoring looks correct.
> 
> I assume that the motivation is that you later want to modify `BaseQNAcc::concatenateCouplingData`, right?
> 
> > Moves the initialization of vectors, matrices and preconditioners into their own function.
> 
> This you have not yet done or am I missing sth?

Yes, I have not done this yet. The reason for this structure is that the arguments of concatenateCoupling data will be different in  the waveform version. I also do not know how to keep the structure with an abstract concatenate function in base class when the concatenateCoupling data has different input arguments in the base classes.

src/acceleration/AitkenAcceleration.hpp
NiklasKotarsky: Yes. However looking at Benni's structure I think it is nicer to keep this as protected, since it now overrides concatenateCouplingData in acceleration.hpp.

NiklasKotarsky:
This is ready to merge if the class structure in regards to concatenateCouplingData looks correct. 
 
> I assume that the motivation is that you later want to modify `BaseQNAcc::concatenateCouplingData`, right?

Yes, this is correct. The reason for this structure is that the arguments of concatenateCoupling data will be different in  the waveform version. I also do not know how to keep the structure with an abstract concatenate function in base class when the concatenateCoupling data has different input arguments in the base class.

BenjaminRodenberg (approved):
Looks good to me. Let's merge this PR.

uekerman:
> Yes, this is missing, but it probably makes more sense to do this change in conjunction with merging the large pull request.

Probably makes sense. Could you please update the PR description above? :pray:

PR #2080, opened 2024-09-19
Initialize devices only if required

davidscn:
## Main changes of this PR

Initializes the device(s) only if required in the corresponding solver class.

In the previous setup, we initialized the device in the MappingConfiguration. This initialized the device (1) on both participants, although the mapping is only computed on one of them and (2) on each MPI rank when running participants in parallel. I was thinking that this is not overly problematic, but the initialize call allocates memory on the device (about 500 MiB), and as a consequence we run out of memory for more ranks with the same problem size. I don't understand what this memory is allocated for and why this is happening in the first place. Maybe @MarcelKoch has a clue?

Kokkos has an `initialize` overload, which takes the `argc` and `argv`, but from what I understand from the [Kokkos documentation](https://kokkos.org/kokkos-core-wiki/API/core/initialize_finalize/initialize.html#kokkosinitialize), Kokkos checks in the `argv` for kokkos specific settings. 
They recommend to initialize after the MPI initialization, which might not happen in preCICE at all. Thus, this option is off the table I guess.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/2080.md
+++ b/docs/changelog/2080.md
@@ -1,0 +1,1 @@
+- Moved Kokkos initialization into solver class to prevent unnecessary initializations where they are not required.
--- a/src/mapping/GinkgoRadialBasisFctSolver.hpp
+++ b/src/mapping/GinkgoRadialBasisFctSolver.hpp
@@ -62,7 +62,6 @@ template <typename RADIAL_BASIS_FUNCTION_T>
 class GinkgoRadialBasisFctSolver {
 public:
   using BASIS_FUNCTION_T       = RADIAL_BASIS_FUNCTION_T;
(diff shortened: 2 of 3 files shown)


MarcelKoch:
Since `Ginkgo::initialize` only calls the kokkos initialization, I don't really know what's causing the allocations. My guess would be that kokkos is initializing cuda/hip libraries such as cublas, cusparse, etc. Ginkgo does the same, but only when the first cuda/hip executor is created.

davidscn:
I see, thanks for the feedback. I can imagine that some allocations are required to enable device access, but the amount seemed drastic in my opinion. However, good to know, I will add another comment in the code, thank you!

PR #2103, opened 2024-10-17
Prevent temporary allocations of Eigen Map in vertex creation

davidscn:
## Main changes of this PR

Makes the vertex definition non-allocating in the vertex loop.

While doing some memory investigations, there was a suspicious amount of temporary allocations when creating our mesh.
Well, passing Eigen expressions to functions that expect concrete types (const Eigen::VectorXd&) forces Eigen to evaluate the expressions and creates a temporary Eigen::VectorXd to hold the result.
Using a [Ref](https://eigen.tuxfamily.org/dox/classEigen_1_1Ref.html) prevents the unnecessary allocation.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mesh/Mesh.cpp
+++ b/src/mesh/Mesh.cpp
@@ -100,7 +100,7 @@ int Mesh::getDimensions() const
   return _dimensions;
 }
 
-Vertex &Mesh::createVertex(const Eigen::VectorXd &coords)
+Vertex &Mesh::createVertex(const Eigen::Ref<const Eigen::VectorXd> &coords)
 {
   PRECICE_ASSERT(coords.size() == _dimensions, coords.size(), _dimensions);
   auto nextID = _vertices.size();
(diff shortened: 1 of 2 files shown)


fsimonis (approved):
Nice!
I wasn't aware of `Ref`. This makes sense in _many_ locations in our code base.

davidscn:
For reference: the second place I won't fix here was stemming from

https://github.com/precice/precice/blob/a41d1a229c8f53bab2d80d122d2f2447dbf7d3de/src/mesh/Mesh.cpp#L242

https://github.com/precice/precice/blob/a41d1a229c8f53bab2d80d122d2f2447dbf7d3de/src/mesh/BoundingBox.cpp#L149

We could add a const and non-allocating version of `getCoords`, e.g., using a `Eigen::Map<const Eigen::VectorXd>`, which would be completely sufficient for the purpose of the mesh function.

PR #2107, opened 2024-10-18
Simplify Acceleration class by removing unnecessary functions

BenjaminRodenberg:
## Main changes of this PR

Simplifies the `Acceleration.hpp` class.

Reduce complexity of `Acceleration` class hierarchy.

Drawback: I need to update the implementation of the writer in `BaseCouplingScheme`. However, I think that the implementation before was also not great: writing `QNColumns`, `DeletedQNColumns`, and `DroppedQNColumns` for non-QN acceleration schemes does not really make sense. I still decided to keep this and hard-code the value to `0` (same behavior as before) to implement this in a non-breaking fashion.

We could, of course, also just remove the `else` branch. This would affect the file output. This would be breaking but also an improvement.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/2107.md
+++ b/docs/changelog/2107.md
@@ -1,1 +1,0 @@
-- Remove columns "QNColumns", "DeletedQNColumns", and "DroppedQNColumns" from `*-iterations.log` for non-quasi-Newton acceleration schemes.
--- a/docs/changelog/2107.md
+++ b/docs/changelog/2107.md
@@ -1,0 +1,1 @@
+- Remove columns "QNColumns", "DeletedQNColumns", and "DroppedQNColumns" from `*-iterations.log` for non-quasi-Newton acceleration schemes.
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -771,19 +771,12 @@ void BaseCouplingScheme::advanceTXTWriters()
     bool converged = _iterations >= _minIterations && (_maxIterations < 0 || (_iterations < _maxIterations));
     _iterationsWriter->writeData("Convergence", converged ? 1 : 0);
 
-    if (not doesFirstStep()) {
-      std::shared_ptr<precice::acceleration::BaseQNAcceleration> qnAcceleration = std::dynamic_pointer_cast<precice::acceleration::BaseQNAcceleration>(_acceleration);
-      if (qnAcceleration) {
-        // Only write values for additional columns, if using a QN-based acceleration scheme
-        _iterationsWriter->writeData("QNColumns", qnAcceleration->getLSSystemCols());
-        _iterationsWriter->writeData("DeletedQNColumns", qnAcceleration->getDeletedColumns());
-        _iterationsWriter->writeData("DroppedQNColumns", qnAcceleration->getDroppedColumns());
-      } else {
-        // non-breaking implementation uses "0" for these columns (delete columns in the future?)
-        _iterationsWriter->writeData("QNColumns", 0);
-        _iterationsWriter->writeData("DeletedQNColumns", 0);
-        _iterationsWriter->writeData("DroppedQNColumns", 0);
-      }
+    std::shared_ptr<precice::acceleration::BaseQNAcceleration> qnAcceleration = std::dynamic_pointer_cast<precice::acceleration::BaseQNAcceleration>(_acceleration);
+    if (qnAcceleration && not doesFirstStep()) {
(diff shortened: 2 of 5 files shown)


src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -770,10 +771,19 @@ void BaseCouplingScheme::advanceTXTWriters()
...
+        // Only write values for additional columns, if using a QN-based acceleration scheme
+        _iterationsWriter->writeData("QNColumns", qnAcceleration->getLSSystemCols());
+        _iterationsWriter->writeData("DeletedQNColumns", qnAcceleration->getDeletedColumns());
+        _iterationsWriter->writeData("DroppedQNColumns", qnAcceleration->getDroppedColumns());
+      } else {
+        // non-breaking implementation uses "0" for these columns (delete columns in the future?)
+        _iterationsWriter->writeData("QNColumns", 0);
+        _iterationsWriter->writeData("DeletedQNColumns", 0);
+        _iterationsWriter->writeData("DroppedQNColumns", 0);
+      }
```
BenjaminRodenberg, line 786:
```suggestion
    std::shared_ptr<precice::acceleration::BaseQNAcceleration> qnAcceleration = std::dynamic_pointer_cast<precice::acceleration::BaseQNAcceleration>(_acceleration);
    if (qnAcceleration && not doesFirstStep()) {
      // Only write values for additional columns, if using a QN-based acceleration scheme
      _iterationsWriter->writeData("QNColumns", qnAcceleration->getLSSystemCols());
      _iterationsWriter->writeData("DeletedQNColumns", qnAcceleration->getDeletedColumns());
      _iterationsWriter->writeData("DroppedQNColumns", qnAcceleration->getDroppedColumns());
```
We could also just drop the QN-related columns from non-QN acceleration schemes.

BenjaminRodenberg (commented):
This is the code version that will affect the user

BenjaminRodenberg:
I decided to use the (slightly) breaking version of the code. If this is breaking too much we can still cherry-pick [0cc4f01](https://github.com/precice/precice/pull/2107/commits/0cc4f01c9161841cae7646fa9cd138dc5a64d090) and turn this PR into a v4 PR.

fsimonis:
This breaks post-processing scripts like plotting if they depend on the column order or count.

I suggest moving the writing changes to an issue instead.

docs/changelog/2107.md
```diff
@@ -0,0 +1 @@
+- Remove columns "QNColumns", "DeletedQNColumns", and "DroppedQNColumns" from `*-iterations.log` for non-quasi-Newton acceleration schemes.
```
fsimonis, line 1: The current state doesn't remove the columns

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -770,10 +771,12 @@ void BaseCouplingScheme::advanceTXTWriters()
     bool converged = _iterations >= _minIterations && (_maxIterations < 0 || (_iterations < _maxIterations));
     _iterationsWriter->writeData("Convergence", converged ? 1 : 0);
 
-    if (not doesFirstStep() && _acceleration) {
-      _iterationsWriter->writeData("QNColumns", _acceleration->getLSSystemCols());
-      _iterationsWriter->writeData("DeletedQNColumns", _acceleration->getDeletedColumns());
-      _iterationsWriter->writeData("DroppedQNColumns", _acceleration->getDroppedColumns());
+    std::shared_ptr<precice::acceleration::BaseQNAcceleration> qnAcceleration = std::dynamic_pointer_cast<precice::acceleration::BaseQNAcceleration>(_acceleration);
+    if (qnAcceleration && not doesFirstStep()) {
```
fsimonis, line 775: Does this actually work? Not writing data should trigger an assertion.

fsimonis (changes requested):

docs/changelog/2107.md
BenjaminRodenberg: I guess because of this part?
```
if (_acceleration) {
  _iterationsWriter->addData("QNColumns", io::TXTTableWriter::INT);
  _iterationsWriter->addData("DeletedQNColumns", io::TXTTableWriter::INT);
  _iterationsWriter->addData("DroppedQNColumns", io::TXTTableWriter::INT);
}
```
Anyway: I will revert this PR to the non-breaking version and put the suggestions that would be breaking into an issue.

fsimonis (approved):

PR #2111, opened 2024-10-21
Add error when defining multiple accelerations

### LINKED ISSUE #2109: Configuration file allows multiple acceleration tags per coupling scheme
The following configuration is allowed:

```
    ...
    <acceleration:IQN-ILS>
      <data name="Temperature" mesh="Neumann-Mesh" />
      <initial-relaxation value="0.1" />
      <max-used-iterations value="100" />
      <time-windows-reused value="0" />
      <filter type="QR2" limit="1e-3" />
    </acceleration:IQN-ILS>
    <acceleration:constant>
      <relaxation value="0.5" />
    </acceleration:constant>
    ...
```

I believe the last acceleration scheme is the one actually applied.

**Describe your setup**

Operating system (e.g. Linux distribution and version): _Ubuntu 24.04_
preCICE Version: badff07c74088c36995ea7810783da44a161bbea based on 3.1.2

**Describe the problem**

The definition of multiple acceleration schemes should lead to an error. Since otherwise this can be very puzzling when performing studies.

**Step To Reproduce**
Should be reproducible for any case. I did a study with the partitioned heat conduction case and activated the QN-acceleration without deactivating the underrelaxation. This first confused me quite a lot. Then I discovered that I forgot to deactivate the underrelaxation.

**Expected behaviour**
There should be an error if more than one acceleration scheme exists in a coupling scheme

fsimonis:
## Main changes of this PR

Adds an error when defining two acceleration schemes in a coupling scheme.

```
ERROR: You are trying to define multiple acceleration schemes, which is not allowed. Please remove one of them.
```

Closes #2109

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/2111.md
+++ b/docs/changelog/2111.md
@@ -1,0 +1,1 @@
+- Added error message when defining multiple acceleration schemes per coupling-scheme.
--- a/src/acceleration/config/AccelerationConfiguration.cpp
+++ b/src/acceleration/config/AccelerationConfiguration.cpp
@@ -256,6 +256,7 @@ void AccelerationConfiguration::xmlEndTagCallback(
       }
     }
 
+    PRECICE_CHECK(!_acceleration, "You are trying to define multiple acceleration schemes, which is not allowed. Please remove one of them.");
     if (callingTag.getName() == VALUE_CONSTANT) {
       _acceleration = PtrAcceleration(
           new ConstantRelaxationAcceleration(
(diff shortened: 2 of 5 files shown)


BenjaminRodenberg (approved):
Looks good (cannot test it right now because I'm running something else).

How about adding a test here to make sure that the error is triggered? This should be possible with the new exceptions, no?

PR #2116, opened 2024-10-22
Add sunburst plotting to profiling

fsimonis:
## Main changes of this PR

This PR adds a command to plot events of a participant as sunburst plots.

The plotting will improve with #1673 as the scoping is currently manual and incomplete.

## Author's checklist

### DIFF
--- a/tools/profiling/precice-profiling
+++ b/tools/profiling/precice-profiling
@@ -592,6 +592,100 @@ def analyzeCommand(profilingfile, participant, event, outfile=None, unit="us"):
     return 0
 
 
+def sunburstCommand(profilingfile, participant, unit="us"):
+    requireModules(["polars", "plotly"])
+    import polars as pl
+
+    checkPolarsVersion()
+
+    run = Run(profilingfile)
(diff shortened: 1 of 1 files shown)


davidscn:
I think this would be useful, if the plot groups events of the same name together. Does it?

fsimonis:
Events are grouped together, and the sum is shown. Something seems to be off with the global event or the scaling though. This needs some investigation.

fsimonis:
It is possible to visualize the same data as an Icicle graph/Flamegraph. This is easier to visually see the cost of nested events, but makes it more difficult to read the event names.

Example of the Solid solver:

newplot(7)

MakisH:
I also find the sunburst graphs:

- easier to navigate (if you know what you are looking at)
- visually more accurate (easier to compare length/area of a rectangle than angle/area of a ring)
- easier to fit more event names

fsimonis:
Removing this from v3.2 as we need to find a consensus on #2231

fsimonis:
Moved to https://github.com/precice/profiling/pull/2

PR #2118, opened 2024-10-24
Automatically register tests in CTest

fsimonis:
## Main changes of this PR

This PR adds automatic test registration to CMake and CTest.

After building `testprecice`, CMake will generate a file containing all registered tests, based on `./testprecice --list_units`.

This opens the door for running tests in parallel using `ctest -j`, rerunning failed test using `ctest --rerun-failed`, rerunning n times, filtering by label/regex and more.

Note that this is more expensive the bundled method before, as calling `MPI_Init` is an expensive operation.

The tests now log everything by default, which makes `--output-on-failure` finally useful.

## Author's checklist

### DIFF
--- a/.github/workflows/build-and-test-mac.yml
+++ b/.github/workflows/build-and-test-mac.yml
@@ -82,7 +82,7 @@ jobs:
       env:
         CTEST_OUTPUT_ON_FAILURE: 1
       working-directory: build
-      run: ctest
+      run: ctest -j
     - uses: actions/upload-artifact@v4
       if: failure()
       with:
--- a/.github/workflows/build-and-test-mac.yml
+++ b/.github/workflows/build-and-test-mac.yml
@@ -81,6 +81,8 @@ jobs:
     - name: Tests
       env:
         CTEST_OUTPUT_ON_FAILURE: 1
+        OMPI_MCA_btl: self,vader
+        OMPI_MCA_pml: ob1
       working-directory: build
       run: ctest -j
     - uses: actions/upload-artifact@v4
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -152,7 +152,9 @@ jobs:
         env:
           CTEST_OUTPUT_ON_FAILURE: 1
           BOOST_TEST_LOG_LEVEL: all
-          OMPI_MCA_btl_tcp_if_include: lo # required for arch
+          OMPI_MCA_btl: self,vader
+          OMPI_MCA_pml: ob1
+          FI_PROVIDER: ${{ matrix.IMAGE == 'Intel' && 'tcp' || 'sockets' }} # use sockets on fedora and tcp on intel
           PSM3_HAL: loopback # required for fedora
         working-directory: build
         run: su -c ". ../venv/bin/activate && ctest -j" $PRECICE_USER
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -155,7 +155,7 @@ jobs:
           OMPI_MCA_btl_tcp_if_include: lo # required for arch
           PSM3_HAL: loopback # required for fedora
         working-directory: build
-        run: su -c ". ../venv/bin/activate && ctest" $PRECICE_USER
+        run: su -c ". ../venv/bin/activate && ctest -j" $PRECICE_USER
       - name: Coverage Report
         working-directory: build
         if: ${{ matrix.COVFLAGS }}
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -223,7 +223,9 @@ if (PRECICE_FEATURE_MPI_COMMUNICATION)
   set(MPI_DETERMINE_LIBRARY_VERSION ON) # Try to detect the library vendor
 
   find_package(MPI REQUIRED)
-  message(STATUS "MPI Version: ${MPI_CXX_LIBRARY_VERSION_STRING}")
+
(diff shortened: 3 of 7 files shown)


davidscn:
I agree that this is in principle the correct way to use `ctest`, as it (as mentioned above), finally enables to successfully identify failing units.

However, the runtime overhead is significant. Using `ctest -j` doesn't change anything for and the tests are still executed sequentially. I think we would need an intuitive way to only execute a subset of tests. `ctest -R` is great in principle, but our test naming convention is quite loose. Maybe we could or should also consider shortening some test names, as `ctest` expands the status according to the longest test name.

fsimonis:
> However, the runtime overhead is significant.

Sadly I am not sure if we can get rid of this. The vast majority of the overhead comes from the fact that `MPI_Init()` is really slow.

This is why I had to change the testing executable. Initializing MPI before printing the test list had an overhead of about 1 second and that was without using `mpirun`.

> Using ctest -j doesn't change anything for and the tests are still executed sequentially.

Using `-j` runs all tests in parallel on my machine :tm:.
The tests were locked to running in serial before. I removed this in this PR to see how things go.

> I think we would need an intuitive way to only execute a subset of tests. ctest -R is great in principle, but our test naming convention is quite loose. Maybe we could or should also consider shortening some test names, as ctest expands the status according to the longest test name.

One option is to always include the prefix `-R precice.MathTests/Blah/Blub`.
Another option is to auto assign more labels based on the name of the test. I currently do this here: 

https://github.com/precice/precice/pull/2118/files#diff-cf229386e7e1d6b7068d216ab217ef21ca81b2ed0ac74f5a2779fcdca85ac8d1R31-R38

fsimonis:
> Also, why does preCICE suddenly fail to compile on Fedora... I don't even see an error in the log. This doesn't look like it is the post build hook, even though nothing else changed...

Seems to be a problem with GNU Make 4.4.1 I don't get an error using Ninja.

davidscn:
We already have a requirement system for specific dependencies such as PETSc etc. Couldn't we just use the same for MPI?

fsimonis:
The requirements are setup inside the test function and not accessible outside. Maybe it could be moved into a decorator/fixture. This depends when these fixtures are called/used/applied though.

fsimonis:
I noticed `MPI_Init` hangs a while after all ranks fail to init the btl ofi using this program:

```cpp
// g++ init.cpp -lmpi -g -o init
#include <mpi.h>

int main(int argc, char**argv) {
  MPI_Init(&argc, &argv);
  MPI_Finalize();
  return 0;
}
```

```console
$ mpirun --mca btl_base_verbose 100 -n 4 ./init
...
[x:15969] select: initializing btl component self
[x:15969] select: init of component self returned success
[x:15969] select: initializing btl component ofi
[x:15969] select: init of component ofi returned failure
>>> hangs here for a while
```

If I disable ofi, the init times look way better. In fact, they are about the same as enabling only self and shared memory btls:

```console
$ hyperfine "mpirun -n 4 ./init" "mpirun -n 4 --mca btl ^ofi ./init" "mpirun -n 4 --mca btl self,sm ./init"
Benchmark 1: mpirun -n 4 ./init
  Time (mean ± σ):      1.391 s ±  0.033 s    [User: 0.119 s, System: 0.608 s]
  Range (min … max):    1.365 s …  1.478 s    10 runs

Benchmark 2: mpirun -n 4 --mca btl ^ofi ./init
  Time (mean ± σ):     348.3 ms ±   6.1 ms    [User: 125.7 ms, System: 551.1 ms]
  Range (min … max):   340.4 ms … 359.2 ms    10 runs

Benchmark 3: mpirun -n 4 --mca btl self,sm ./init
  Time (mean ± σ):     338.9 ms ±   4.7 ms    [User: 117.2 ms, System: 551.4 ms]
  Range (min … max):   333.3 ms … 348.8 ms    10 runs

Summary
  mpirun -n 4 --mca btl self,sm ./init ran
    1.03 ± 0.02 times faster than mpirun -n 4 --mca btl ^ofi ./init
    4.10 ± 0.11 times faster than mpirun -n 4 ./init
```

The using `sm` instead of `tcp` is slightly faster:

```console
$ hyperfine "mpirun -n 4 --mca btl self,tcp ./init" "mpirun -n 4 --mca btl self,sm ./init"
Benchmark 1: mpirun -n 4 --mca btl self,tcp ./init
  Time (mean ± σ):     366.1 ms ±  37.5 ms    [User: 113.3 ms, System: 599.5 ms]
  Range (min … max):   325.0 ms … 442.3 ms    10 runs

Benchmark 2: mpirun -n 4 --mca btl self,sm ./init
  Time (mean ± σ):     341.2 ms ±  10.2 ms    [User: 117.4 ms, System: 557.2 ms]
  Range (min … max):   328.9 ms … 359.7 ms    10 runs

Summary
  mpirun -n 4 --mca btl self,sm ./init ran
    1.07 ± 0.11 times faster than mpirun -n 4 --mca btl self,tcp ./init
```

fsimonis:
Looks like some integration tests use the default `m2n` exchange directory, being `..`.
All tests hammer this directory and some tests with identical participant names interfere with each other.

I added an `inner` directory inside each test to prevent this, which significantly reduces the overall runtime too. (at least locally)

fsimonis:
Rebased to pull in the fix from #2145, which will impact the overall test execution time.

fsimonis:
I added automatic label assignment to reconstruct the groups we had previously.
These are extracted from the first test suite which either contains the module in precice (xml, io,...) or `integration`.
I also converted all labels to lowercase for consistency.

davidscn (approved):
We had now enough (internal) iterations on these changes. Follow-up changes for improved test runtime are ahead.

fsimonis:
I'll cleanup the commits and merge.

PR #2121, opened 2024-10-30
More refined repartitioning for direct access

davidscn:
## Main changes of this PR

If we face compbinations of direct access + mappings on the same remote mesh, we now use a more fine-grained tagging.

Before, we used to tag 'accessRegion + mapping BB * safetyFactor'.

Now, we tag 'accessRegion' and make use of the tagging mechanisms of the mapping(s).

Related to and extracted from #2099. Where this becomes useful.


davidscn:
The intention for separating this was mostly to run the tests. Will be revisited in #2099 again.

PR #2124, opened 2024-10-31
Add error for swapped meshes in mappings

fsimonis:
## Main changes of this PR

This PR adds an error message when meshes in mappings are swapped, indicating a wrong read/write direction or swapped meshes

I ran into this issue a few time recently and the current error messages are confusing as they suggest receiving the provided mesh or vice versa.

Example:
```
precice-tools: ERROR: A write mapping of participant "SolverTwo" needs to map from a provided to a received mesh, but in this case they are swapped. Did you intent to map from mesh "SolverOne-Mesh" to mesh "SolverTwo-Mesh", or use a read mapping instead?
```

Closes https://github.com/precice/precice/pull/2128

## Author's checklist

### DIFF
--- a/docs/changelog/2124.md
+++ b/docs/changelog/2124.md
@@ -1,0 +1,2 @@
+- Added descriptive error when meshes are swapped in read/write mappings.
+- Fixed a crash when meshes in a mapping neither received nor provided by a participant.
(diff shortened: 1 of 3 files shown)


carme-hp:
I have modified the error message: 
The sentence  _"Did you intent to map from mesh \"{}\" to mesh \"{}\", or use a read mapping instead?"_ returned from mesh A to mesh B as stated by the (wrong) `precice-config.xml` instead of actually suggesting to swap names. 

Otherwise it looks good.

carme-hp (approved):

Issue #2060, opened 2024-07-24
Test output Exporter


fsimonis:
**Please describe the problem you are trying to solve.**

Exporter tests only test if they don't fail while exporting.
This still leaves many open questions:

* do all expected files exist?
* have empty piece files been correctly skipped? 
* is their content correct?

**Describe the solution you propose.**

The first two could be as simple as deleting files before the test and checking for existing files afterwards.

The final one is more tricky. Checking the syntax of the XML-based formats should be easy, as preCICE already relies on libxml2.
Checking the semantics could be done using a XSD schema file, but VTK doesn't seem to provide one.
Checking the content should be easy for XML-based files, but will be hacky for legacy VTK.
Other files like the CSV and the series (JSON) files require additional dependencies, which may not be worth the hassle.

**Additional context**

Follow-up of #2055

HenryCai-st:
Hi!
I would like to help solving this issue. Is this issue still available? If yes, that the last one seems more difficult as you have stated. Shall I first work on the first 2 points only? Which is testing existence of outputs after the additional exporters from [this pull request](https://github.com/precice/precice/pull/2055)? That there seems to be [tests for exporters](https://github.com/precice/precice/tree/develop/src/io/tests) already?

fsimonis:
Hi @HenryCai-st, sure, you can try to extend the tests.

For serial tests, this is pretty simple.

For parallel tests, it is important that each rank is responsible for its own piece file. Testing for a rank 2 file on rank 1 can sometimes work, or even often, but will not always work.

We already ended up with some of these ghosts in the test system.

HenryCai-st:
I draft a [PR](https://github.com/precice/precice/pull/2445). The basic functions are done. But there are still something I feel uncertain about(written in PR), which I would appreciate if it can be clarified.

PR #2052, opened 2024-07-10
Add system tests workflow

uekerman:
## Main changes of this PR

Adds a workflow to run the "release" testsuite, which gets triggered by the `trigger-system-tests` label.

## Motivation and additional information

We have the system tests now. Let's use them.

## Author's checklist

* [x] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [x] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Do you understand the code changes?

PR #2125, opened 2024-11-08
Revert swapped participant and mesh name in export filenames

fsimonis:
## Main changes of this PR

This PR reverts the change of exported file names from Participant-Mesh to Mesh-Participant.

Breaks the systemtests https://github.com/precice/precice/pull/2052#issuecomment-2464145569

@davidscn Shouldn't this have been noticeable in ASTE too?

Another reason for #2060 

Regression introduced in #2053

## Author's checklist

### DIFF
--- a/src/io/ExportCSV.cpp
+++ b/src/io/ExportCSV.cpp
@@ -57,11 +57,11 @@ void ExportCSV::doExport(int index, double time)
   // Construct filename
   std::string filename;
   if (isParallel()) {
-    // Participant-Mesh-r2.it2
-    filename = fmt::format("{}-{}.{}_{}.csv", _participantName, _mesh->getName(), _rank, formatIndex(index));
+    // Mesh-Participant.r2.it2
+    filename = fmt::format("{}-{}.{}_{}.csv", _mesh->getName(), _participantName, _rank, formatIndex(index));
   } else {
-    // Participant-Mesh.it2
-    filename = fmt::format("{}-{}.{}.csv", _participantName, _mesh->getName(), formatIndex(index));
+    // Mesh-Participant.it2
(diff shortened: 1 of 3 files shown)


davidscn:
> @davidscn Shouldn't this have been noticeable in ASTE too?

Yes, also tried to make this clear here https://github.com/precice/precice/pull/2053#issuecomment-2230260497

davidscn (commented):
If these changes revert another PR: did you consider to revert via GitHub? There is a button for this..
At least I would propose to put somewhere a 'Revert' into the title to make this more clear.

MakisH:
> > @davidscn Shouldn't this have been noticeable in ASTE too?
> 
> Yes, also tried to make this clear here [#2053 (comment)](https://github.com/precice/precice/pull/2053#issuecomment-2230260497)

We already discussed there about the `.init`, on which all the attention was focused. What slipped through the review was that the order `Mesh-Participant` was swapped. This was not discussed there.

> If these changes revert another PR: did you consider to revert via GitHub? There is a button for this.. At least I would propose to put somewhere a 'Revert' into the title to make this more clear.

The rest of the refactoring stays, only the order of the names changes.

I guess `Revert change of order in names of exported files` would be good as well.

fsimonis:
> Yes, also tried to make this clear here https://github.com/precice/precice/pull/2053#issuecomment-2230260497

I mean that the files have a different naming scheme now. Regarding the initial export, sure. But they also weren't used by ASTE IIRC.

> If these changes revert another PR

This was just a small unintentional change from when I refactored the exporters as a whole.

MakisH (approved):
Thank you for the quick fix, this seems to do what I asked for in https://github.com/precice/precice/pull/2053#issuecomment-2462906376.

fsimonis:
Ok merged. Thankfully this hasn't been released.

Issue #982, opened 2021-03-19
Clean multi-step configuration


fsimonis:
**Please describe the problem you are trying to solve.**
Currently the configuration of preCICE is directly tied into the parsing of the XML file.
This consists out of:
* LibXML2 which parses an XML file and uses callbacks for entering/leaving tags.
* The `XMLTag`s and `XMLAttribute`s, which are used to validate the XML and contain information on which `Configuration` callbacks should be dispatched to. The `XMLAttribute` knows how to validate and parse its value to the final type (`int`, `double`, `string`). Both of them also contain documentation.
* The ConfigParser, which uses libxml2 to parse an XML file and dispatches the callbacks from LibXML2 to `Listeners` based on the AST of `XMLTag`s given by the root tag.
* The `Configuration`s, which implement `xml::XMLTag::Listener`, receiving callbacks and build and configuring preCICE objects. They also build the AST of `XMLTag`s and `XMLAttribute`s and hold configured preCICE objects
* The `SolverInterface`, which created the base configuration, runs the ConfigParser on the Configuration (object and file), extracts the required configured objects from the Configuration and finalizes their configuration if required.

**Describe the solution you propose.**
Restructure the parsing to multiple stages:
1. Use LibXML2 to create an XML AST using the callbacks. This simply contains nested tags and attributes.
2. Transform the XML AST to a Configuration AST. This part checks tag occurrence and validates attributes.
3. Transform the Configuration AST to the required preCICE objects using contextual information (participant name, current rank, total ranks).

The benefits of this strategy is:
* Easier to test as stages can be checked separately
* Easier to debug as the output relies only on the input and both can be inspected separately.
* Allows to integrate location information (from the  XML file) for each object in the Configuration AST. This should allow to generate very detailed error messages.
* The configuration AST contains all information, hence the order in the XML will not matter any more.

The downsides are:
* A lot of work
* Spreads out information into the transformation functions and the objects.

Open Questions:
* Where to put the XML documentation?
* How to distribute the information over the packages (`com`, `m2n`, ...)

**Describe alternatives you've considered**
Leave it as it is.

uekerman:
Would this interfere with the localized configuration we have right now? Meaning that each package has its own local configuration? In my opinion this is a very neat feature and a low-entry barrier. I have seen other codes where all configuration routines are in one place -- a nightmare if you want to change sth.

>     The configuration AST contains all information, hence the order in the XML will not matter any more.

[...]

PR #2126, opened 2024-11-11
Fix unnecessary empty watchpoint and watchintegral files

### LINKED ISSUE #1989: Empty Watchpoint files in non-watching participants
**Describe your setup**

Operating system (e.g. Linux distribution and version): Yes
preCICE Version: develop (3.1.0 rc)

**Describe the problem**

When configuring a watchpoint in a participant, also other participants create the watchpoint log files.

This leaves empty watchpoint files around, which

* can lead to confusion of the user
* results in unnecessary files
* makes automatic data aggregation and plotting more difficult as files are not unique.
* can lead to problems when multiple participants run from the same directory as this is a filesystem race-condition. This is generally not the case though.

**Step To Reproduce**
1. define a watchpoint on a mesh in one participant (the elastic-tube-1d tutorial is an easy-to-run example)
2. run the case
3. find a watchpoint log file in every participant directory.

**Expected behaviour**

Only write log files from the participant that is being run.

**Additional context**

* Watchpoint/integral constructor opens the log file
* ParticipantConfiguration::finishParticipantConfiguration creates watchpoints/integrals for every configured participant.

We need #982

uekerman:
## Main changes of this PR

Closes #1989 

A simple, but not clean solution. #982 should make such things much cleaner.
I tested with the Elastic Tube 1D tutorial.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/2126.md
+++ b/docs/changelog/2126.md
@@ -1,0 +1,1 @@
+- Fixed unnecessary creation of empty watchpoint and watchintegral files at non-owning participants.
--- a/src/precice/config/ParticipantConfiguration.cpp
+++ b/src/precice/config/ParticipantConfiguration.cpp
@@ -658,38 +658,42 @@ void ParticipantConfiguration::finishParticipantConfiguration(
   _exportConfig->resetExports();
 
   // Create watch points
-  for (const WatchPointConfig &config : _watchPointConfigs) {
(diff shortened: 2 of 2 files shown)


fsimonis (approved):
Not pretty, but pretty pragmatic

PR #2127, opened 2024-11-11
Expand data in profiling export

### LINKED ISSUE #2123: Additional data in `precice-profiling`

Our events have the option to commit data, which is not a duration value.

https://github.com/precice/precice/blob/ffe0f992db5b28381ecb5f32ebeec3bd0f4afd52/src/profiling/Event.hpp#L94

However, accessing the generated data through our CLI is not really possible. There is apparently some opportunity so export the data to a csv data frame, but this option is not even documented somewhere.

The event data is useful if you want to connect an event duration to a certain data value, e.g., the size of your rank-local meshes certainly affects the time the mapping requires on this rank. We use this, e.g., in the RBF Solver to track the number of iterations:
https://github.com/precice/precice/blob/ffe0f992db5b28381ecb5f32ebeec3bd0f4afd52/src/mapping/PetRadialBasisFctMapping.hpp#L484

We should provide access to this data through the CLI we currently have. I am not sure how this data was treated in previous preCICE versions (or previous versions of our event tool).

We could consider to print something like an EventData table

```
precice-profiling analyze Fluid --show-data
```

or simply always give an additional table with the data. 

An alternative could be to integrate the table into our 'usual' event timings. The key would then be `eventName+key` and instead of a duration, the table then contains the data value. However, this becomes problematic if we change units in this table.

Overall, I think having the data more accessible would greatly enhance their application.

fsimonis:
## Main changes of this PR

This PR changed `precice-profiling export` to expand existing data as columns.
If an event doesn't have a data, its value is empty.
In importers of pandas and polars, empty fields will be mapped to NaN, which can be filtered on.

Related to #2123

## Author's checklist

### DIFF
--- a/tools/profiling/precice-profiling
+++ b/tools/profiling/precice-profiling
@@ -472,8 +472,26 @@ class Run:
 
         return {"traceEvents": metaEvents + mainEvents}
 
-    def toExportList(self, unit=None):
+    def allDataFields(self):
+        return list(
+            {
+                dname
+                for rank in self.iterRanks()
+                for e in rank.events
+                if "data" in e
+                for dname in e["data"].keys()
+            }
+        )
+
+    def toExportList(self, unit, dataNames):
         factor = ns_to_unit_factor(unit) * 1e3 if unit else 1
+
+        def makeData(e):
+            if "data" not in e:
+                return tuple("" for dname in dataNames)
+
+            return tuple(e["data"].get(dname, "") for dname in dataNames)
+
         for rank in self.iterRanks():
             for e in rank.events:
                 yield (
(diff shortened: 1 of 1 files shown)


davidscn (commented):
Hm, I'm not too sure about it. The added data was already visible before in the trace visualization (using hovering over the event bar).

Workflow: exporting the profiling data, opening the data in an external tool

It's at least not too much related to #2123 as there is no human-readable output from the command line.

Workflow: command -> human-readable output

Which data do I get here in parallel runs?

fsimonis:
Example of the elastic-tube-1d, where `Connections` get expanded to a separate column

```csv
participant,rank,size,event,timestamp,duration,Connections
Fluid,0,1,initialize/m2n.requestSecondaryRanksConnection/m2n.createCommunications,1731685322586466,1185.0,1
Solid,0,1,initialize/m2n.acceptSecondaryRanksConnection/m2n.createCommunications,1731685322633099,1142.0,1
```

fsimonis:
While this PR doesn't visualize the data, it makes it easier to use in post-processing scripts.
I'll merge and leave #2123 open.

PR #2130, opened 2024-11-12
Add events to API calls

### LINKED ISSUE #2013: Missing API Profiling

Fujikawas:
## Main changes of this PR
Added profiling Events to more API functions.

Close #2013 

## Author's checklist

## Reviewers' checklist


fsimonis:
Thanks! Could you still add a changelog?

fsimonis (approved):
Great! Let's merge and see how it behaves in large cases.

We may have to remove the event for setting an individual vertex for performance reasons.


```
Reading events file profiling.json
Output timing are in us.
name                                                             │       sum    count               mean       min       max
─────────────────────────────────────────────────────────────────┼──────────────────────────────────────────────────────────
total                                                            │ 2079587.0        1          2079587.0 2079587.0 2079587.0
  solver.advance                                                 │    4878.0      849  5.745583038869258       4.0      22.0
    readData.Solid-Nodes-Mesh_Pressure                           │     506.0      848 0.5966981132075472       0.0       2.0
    writeData.Solid-Nodes-Mesh_CrossSectionLength                │     330.0      848 0.3891509433962264       0.0       1.0
  solver.initialize                                              │      18.0        1               18.0      18.0      18.0
    setMeshVertices.Solid-Nodes-Mesh                             │       3.0        1                3.0       3.0       3.0
    writeData.Solid-Nodes-Mesh_CrossSectionLength                │       1.0        1                1.0       1.0       1.0
---- >8 ====
```

PR #2132, opened 2024-11-13
Skip non-existent data in exporter

fsimonis:
## Main changes of this PR

This PR skips non-existent data when exporting.

Non-exisiting data (received mesh, non-exchanged data) was exported as 0, even though it doesn't really exist.
It is now omitted from the export.

@MakisH this breaks reference results of the system tests.

## Author's checklist


uekerman:
Let's please add an integration test here.

fsimonis:
I just realized that the exporter tests don't define data at all. So it's all mesh-only.

uekerman:
But an integration should be easy. Clear what I mean? If not, I could also add one.

fsimonis:
@uekerman Added tests in #2147

PR #2133, opened 2024-11-13
Add multi coupling test

carme-hp:
## Main changes of this PR
Add integration tests for multi-coupling that include initialization of data.

## Motivation and additional information
Until now initialization of data in multi-coupling was not tested. I have included tests that consider the scenario of multi-coupling with 2 and 3 participants. All the tests pass:)

This was motivated by the issue in  https://github.com/precice/precice/issues/1512, however the issue is not yet resolved. 

## Author's checklist

* [ ] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [ ] I added a test to cover the proposed changes in our test suite.
* [ ] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).
* [ ] I stuck to C++17 features.
* [ ] I stuck to CMake version 3.22.1.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

PR #1911, opened 2023-12-01
Add additional cpl-scheme to compositional test

carme-hp:
## Main changes of this PR

This PR modifies the system test TwoActivatedMuscles to include an additional serial-explicit coupling scheme. 

## Motivation and additional information

This is the follow up of https://github.com/precice/precice/pull/1514#event-11121685856. 
The addition  of the proposed serial-explicit coupling scheme, is apparently not possible, as described in https://github.com/precice/precice/issues/1907.

## Author's checklist

* [x] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [ ] I added a test to cover the proposed changes in our test suite.
* [ ] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).
* [ ] I sticked to C++17 features.
* [ ] I sticked to CMake version 3.16.3.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

PR #1432, opened 2022-09-10
Remove Multi- and BiCouplingScheme

BenjaminRodenberg:
## Main changes of this PR

* Merge Multi- and ParallelCouplingScheme
* Remove BiCouplingScheme in favor of SerialCouplingScheme
* Simplifies class hierarchy and limits amount of duplicate code

## Motivation and additional information

Removes a lot of code and reduces the implementation effort necessary for #1414. On the long run I assume that the MultiCouplingScheme will also profit from this change, because it gets automatically tested by the ParallelCouplingScheme.

## Author's checklist

* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [ ] I added a test to cover the proposed changes in our test suite.
* [ ] I ran `make format` to ensure everything is formatted correctly.
* [ ] I sticked to C++17 features.
* [ ] I sticked to CMake version 3.16.3.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

PR #2144, opened 2024-11-20
Fix initial data in multi-coupling schemes

fsimonis:
## Main changes of this PR

This PR fixes initial data in multi-coupling schemes and enforces that data send to multiple participant is configured identically in terms of `initialize` and `substeps`.
The configuration limitation is something that never worked :tm:.

Tests are in #2134
Closes #2133

## Author's checklist

## Reviews

@BenjaminRodenberg can you have a general look on how this fits into your previous refactoring strategy?
@MakisH @carme-hp Can you verify that this works for your failing cases?

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -253,7 +253,7 @@ PtrCouplingData BaseCouplingScheme::addCouplingData(const mesh::PtrData &data, m
   PRECICE_CHECK(direction == CouplingData::Direction::Send,
                 "Data \"{0}\" cannot be received from multiple sources. Please remove either <exchange data=\"{0}\" ... /> tag", data->getName());
   PRECICE_CHECK(existing->exchangeSubsteps() == communicateSubsteps,
-                "Data \"{0}\" is configured with substeps enabled and disabled at the same time. Please make the configuration consistent in the <exchange data=\"{0}\" ... /> tags", data->getName());
+                "Data \"{0}\" is configured with substeps enabled and disabled at the same time. Please make the configuration consistent in the <exchange data=\"{0}\" ... substeps=\"True/False\" ... /> tags", data->getName());
   PRECICE_CHECK(existing->requiresInitialization == requiresInitialization,
                 "Data \"{0}\" is configured with data initialization enabled and disabled at the same time. Please make the configuration consistent in the <exchange data=\"{0}\" ... /> tags", data->getName());
 
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -237,16 +237,28 @@ void BaseCouplingScheme::initializeWithZeroInitialData(const DataMap &receiveDat
 
 PtrCouplingData BaseCouplingScheme::addCouplingData(const mesh::PtrData &data, mesh::PtrMesh mesh, bool requiresInitialization, bool communicateSubsteps, CouplingData::Direction direction)
 {
-  int             id = data->getID();
-  PtrCouplingData ptrCplData;
-  if (!utils::contained(id, _allData)) { // data is not used by this coupling scheme yet, create new CouplingData
-    ptrCplData = std::make_shared<CouplingData>(data, std::move(mesh), requiresInitialization, communicateSubsteps, direction);
-    _allData.emplace(id, ptrCplData);
-  } else { // data is already used by another exchange of this coupling scheme, use existing CouplingData
-    ptrCplData = _allData[id];
-    PRECICE_CHECK(ptrCplData->getDirection() == direction, "Data \"{0}\" cannot be added for sending and for receiving. Please remove either <exchange data=\"{0}\" ... /> tag", data->getName());
+  const DataID id = data->getID();
+
+  // new data
+  if (_allData.count(id) == 0) {
+    auto ptr = std::make_shared<CouplingData>(data, std::move(mesh), requiresInitialization, communicateSubsteps, direction);
+    _allData.emplace(id, ptr);
+    return ptr;
   }
-  return ptrCplData;
+
+  // data is already used by another exchange of this coupling scheme, use existing CouplingData
+  auto &existing = _allData[id];
+  PRECICE_CHECK(existing->getDirection() == direction,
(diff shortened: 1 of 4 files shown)


src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -237,16 +237,28 @@ void BaseCouplingScheme::initializeWithZeroInitialData(const DataMap &receiveDat
 
 PtrCouplingData BaseCouplingScheme::addCouplingData(const mesh::PtrData &data, mesh::PtrMesh mesh, bool requiresInitialization, bool communicateSubsteps, CouplingData::Direction direction)
 {
-  int             id = data->getID();
-  PtrCouplingData ptrCplData;
-  if (!utils::contained(id, _allData)) { // data is not used by this coupling scheme yet, create new CouplingData
```
BenjaminRodenberg, line 242: Do we still need `utils::contained`? Using `count` is definitely the better way to go. (no need to clean this up in this PR)

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -237,16 +237,28 @@ void BaseCouplingScheme::initializeWithZeroInitialData(const DataMap &receiveDat
...
+    return ptr;
+  }
+
+  // data is already used by another exchange of this coupling scheme, use existing CouplingData
+  auto &existing = _allData[id];
+  PRECICE_CHECK(existing->getDirection() == direction,
+                "Data \"{0}\" cannot be added for sending and for receiving. Please remove either <exchange data=\"{0}\" ... /> tag", data->getName());
+  PRECICE_CHECK(direction == CouplingData::Direction::Send,
+                "Data \"{0}\" cannot be received from multiple sources. Please remove either <exchange data=\"{0}\" ... /> tag", data->getName());
+
```
BenjaminRodenberg, line 255:
```suggestion
```

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -237,16 +237,28 @@ void BaseCouplingScheme::initializeWithZeroInitialData(const DataMap &receiveDat
...
+  PRECICE_CHECK(existing->exchangeSubsteps() == communicateSubsteps,
+                "Data \"{0}\" is configured with substeps enabled and disabled at the same time. Please make the configuration consistent in the <exchange data=\"{0}\" ... /> tags", data->getName());
```
BenjaminRodenberg, line 257:
```suggestion
                "Data \"{0}\" is configured with substeps enabled and disabled at the same time. Please make the configuration consistent in the <exchange data=\"{0}\" ... substeps=\"True/False\" ... /> tags", data->getName());
```

src/cplscheme/MultiCouplingScheme.cpp
```diff
@@ -82,41 +82,47 @@ void MultiCouplingScheme::exchangeInitialData()
   PRECICE_ASSERT(isImplicitCouplingScheme(), "MultiCouplingScheme is always Implicit.");
 
   if (_isController) {
-    if (receivesInitializedData()) {
```
BenjaminRodenberg, line 85: `receivesInitialData()` still needed?

src/cplscheme/MultiCouplingScheme.cpp
```diff
@@ -82,41 +82,47 @@ void MultiCouplingScheme::exchangeInitialData()
...
-      for (auto &receiveExchange : _receiveDataVector) {
-        initializeWithZeroInitialData(receiveExchange.second);
+    for (auto &[from, data] : _receiveDataVector) {
+      if (receivesInitializedDataFrom(from)) {
+        receiveData(_m2ns.at(from), data);
+      } else {
+        initializeWithZeroInitialData(data);
       }
     }
-    if (sendsInitializedData()) {
```
BenjaminRodenberg, line 95: `sendsInitializedData()` still needed?

src/cplscheme/MultiCouplingScheme.cpp
```diff
@@ -82,41 +82,47 @@ void MultiCouplingScheme::exchangeInitialData()
...
-      for (auto &receiveExchange : _receiveDataVector) {
-        initializeWithZeroInitialData(receiveExchange.second);
+    for (auto &[from, data] : _receiveDataVector) {
+      if (receivesInitializedDataFrom(from)) {
+        receiveData(_m2ns.at(from), data);
+      } else {
+        initializeWithZeroInitialData(data);
       }
     }
+    notifyDataHasBeenReceived();
```
BenjaminRodenberg, line 111: Does it make sense to explicitly call `notifyDataHasBeenReceived(from)` inside the `for`-loop and then do a `PRECICE_CHECK` at the end to make sure all the required data has been received? (no need to do this here. Just an idea.)

BenjaminRodenberg (approved):
Looks good. Just some general remarks. Nothing critical.

src/cplscheme/BaseCouplingScheme.cpp
fsimonis: [One day](https://en.cppreference.com/w/cpp/container/map/contains) :wink:

carme-hp (approved):
All the multi-coupling tests I had I already add them to https://github.com/precice/precice/pull/2134.

The compositional muscle tests are still failing (it hangs) but this is not connected to this (the problem appears without multi-coupling, see https://github.com/precice/precice/pull/1911

src/cplscheme/MultiCouplingScheme.cpp
fsimonis: I think this is a common theme of this PR. Some functionality should be moved from the `BaseCouplingScheme` to either the `BiCouplingScheme` for two participant cases, and `MultiCouplingScheme` for multi-participant cases.

A more general alternative would be to change everything in the `BaseCouplingScheme` to a multi-participant format and always pass the coupling partner.

BenjaminRodenberg: > A more general alternative would be to change everything in the `BaseCouplingScheme` to a multi-participant format and always pass the coupling partner.

You mean similar to https://github.com/precice/precice/pull/1432 ? At that point in time we decided against the "make everything a multi coupling scheme"-approach.

PR #2150, opened 2024-11-28
Launch tests in CTest with exact amount of ranks

fsimonis:
## Main changes of this PR

This PR adds the amount of requested ranks per test to the output of `./testprecice --list_units` and launches the tests either without MPI (if only 1 rank is requested) or with the exact amount of required MPI ranks.

* This allows using any amount of ranks in tests, removing the historic maximum of 4.
* Running tests via CTest are always correctly sized and don't need to handle unused ranks, resulting in more consistent runtimes.
* Tests using one rank should be able to run for builds with MPI disabled.

Less communication and initialization leads to a more consistent test runtime.
On my local machine:

```
Benchmark develop: ctest -j
  Time (mean ± σ):     28.418 s ±  4.859 s    [User: 317.758 s, System: 213.569 s]
  Range (min … max):   25.990 s … 38.546 s    10 runs

Benchmark this PR: ctest -j
  Time (mean ± σ):     20.106 s ±  0.828 s    [User: 208.488 s, System: 122.066 s]
  Range (min … max):   18.841 s … 21.008 s    10 runs

Summary
  this PR
    1.41 ± 0.25 times faster than develop
```

Further possible improvements:
* Don't initialize MPI running when using only 1 rank. This can lead to many problems though.
* ~~Use the amount of ranks as CTest processors to prevent overloading the system. I'll look into this.~~ Too conservative. 50% slower.

## Author's checklist

### DIFF
--- a/cmake/discover_tests.cmake
+++ b/cmake/discover_tests.cmake
@@ -15,7 +15,7 @@ function(add_command NAME TEST_NAME)
   set(script "${script}" PARENT_SCOPE)
 endfunction()
 
-function(add_test NAME)
+function(add_test NAME RANKS)
   if(NAME MATCHES "MPIPorts|MPISinglePorts" AND (PRECICE_MPI_VERSION MATCHES "Open MPI|Intel"))
     # Test is unsupported by this MPI implementation
     return()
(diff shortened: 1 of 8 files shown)


uekerman (approved):
Runs for me, `./testprecice --list_units` gives the correct output.

> This allows using any amount of ranks in tests, removing the historic maximum of 4.

I must admit I don't understand how these changes give this effect and currently it is not tested as we don't have such a test yet, right?

fsimonis:
> I must admit I don't understand how these changes give this effect and currently it is not tested as we don't have such a test yet, right?

Until now, CTest had to guess on how to many ranks to use to launch individual test suites. We picked 4 as it covered all test setups.

Now, CTest launches each test separately, meaning it doesn't need to satisfy the max ranks of an entire test suite.
Furthermore, CTest knows how many ranks are required by a test and launches it with the necessary amount. Hence, a test can request 6 ranks and CTest will launch it with 6.

PR #2131, opened 2024-11-12
Fix VTK exporter

fsimonis:
## Main changes of this PR

@MakisH

## Author's checklist

* [ ] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [ ] I added a test to cover the proposed changes in our test suite.
* [ ] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).
* [ ] I stuck to C++17 features.
* [ ] I stuck to CMake version 3.22.1.
* [ ] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

PR #2151, opened 2024-12-04
Fix exporter crash on non-existant data

fsimonis:
## Main changes of this PR

This PR fixes other exporters than VTK (#2131) crashing when exporting a mesh with non-existent data.

The tests now also check for existing output files, which uncovered that most tests weren't exporting anything, making the tests useless.

## Author's checklist


fsimonis:
Failing tests are due to ranks checking all files instead of only the ones they create.

PR #2153, opened 2024-12-05
Add LOOCV config option

davidscn:
## Main changes of this PR

Follow-up on #2059, which makes the change user-configurable and enables a consistent logging.

I left out the parallel aggregation for PUM, as it would introduce unnecessary communication. 

## Author's checklist

Please check the naming of the option, the logging and the default value.

## Reviewers' checklist

### DIFF
--- a/src/mapping/PartitionOfUnityMapping.hpp
+++ b/src/mapping/PartitionOfUnityMapping.hpp
@@ -287,7 +287,22 @@ void PartitionOfUnityMapping<RADIAL_BASIS_FUNCTION_T>::mapConsistent(const time:
   PRECICE_ASSERT(outData.isZero());
 
   // 2. Execute the actual mapping evaluation in all vertex clusters and accumulate the data
-  std::for_each(_clusters.begin(), _clusters.end(), [&](auto &clusters) { clusters.mapConsistent(inData, outData); });
+  // std::for_each(_clusters.begin(), _clusters.end(), [&](auto &clusters) { clusters.mapConsistent(inData, outData); });
+  auto maxErrors = std::array<double, 3>{-1, -1, -1};
+  for (auto &cluster : _clusters) {
+    auto errors = cluster.mapConsistent(inData, outData);
+    for (int i = 0; i < 3; ++i) {
+      maxErrors[i] = std::max(maxErrors[i], errors[i]);
+    }
+  }
+  std::vector<double> logErrors;
+  std::copy_n(maxErrors.begin(), inData.dataDims, std::back_inserter(logErrors));
+  // Do we want to log every iteration about disabled error metrics?
+  if (std::all_of(logErrors.begin(), logErrors.end(), [](auto value) { return value == -1; })) {
+    PRECICE_INFO("Evaluation of componen-wise cross-validation error (LOOCV) from \"{}\" to \"{}\" is disabled in the preCICE configuration.", input()->getName(), output()->getName());
+  } else {
+    PRECICE_INFO("Componen-wise cross-validation error (LOOCV) from \"{}\" to \"{}\": {:e}", input()->getName(), output()->getName(), logErrors);
+  }
 }
 
 template <typename RADIAL_BASIS_FUNCTION_T>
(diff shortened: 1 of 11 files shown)


src/mapping/config/MappingConfiguration.cpp
```diff
@@ -75,7 +75,7 @@ struct BackendSelector {
 // Specialization for the RBF Eigen backend
 template <typename RBF>
 struct BackendSelector<RBFBackend::Eigen, RBF> {
-  typedef mapping::RadialBasisFctMapping<RadialBasisFctSolver<RBF>> type;
+  typedef mapping::RadialBasisFctMapping<RadialBasisFctSolver<RBF>, bool> type;
```
uekerman, line 78: Brief docs what the bool is.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -217,6 +217,8 @@ MappingConfiguration::MappingConfiguration(
   auto attrPolynomial = makeXMLAttribute(ATTR_POLYNOMIAL, POLYNOMIAL_SEPARATE)
                             .setDocumentation("Toggles use of the global polynomial")
                             .setOptions({POLYNOMIAL_ON, POLYNOMIAL_OFF, POLYNOMIAL_SEPARATE});
+  auto attrCrossValidation = makeXMLAttribute(ATTR_CROSS_VALIDATION, false)
+                                 .setDocumentation("Computes and logs a cross-validation error using the coupling data and the mapping matrix. This can be useful for tuning the basis function. Note that this might become computationally expensive.");
```
uekerman, line 221:
```suggestion
                                 .setDocumentation("Computes and logs a cross-validation error using coupling data and the mapping matrix. This can be useful for tuning the basis function. Note that this might become computationally expensive.");
```

To make clearer that the mapping matrix is static and the data dynamic.

src/mapping/impl/SphericalVertexCluster.hpp
```diff
@@ -56,14 +56,15 @@ class SphericalVertexCluster {
...
+                         bool                    crossValidation,
                          mesh::PtrMesh           inputMesh,
                          mesh::PtrMesh           outputMesh);
 
   /// Evaluates a conservative mapping and agglomerates the result in the given output data
   void mapConservative(const time::Sample &inData, Eigen::VectorXd &outData) const;
 
   /// Evaluates a consistent mapping and agglomerates the result in the given output data
-  void mapConsistent(const time::Sample &inData, Eigen::VectorXd &outData) const;
+  std::array<double, 3> mapConsistent(const time::Sample &inData, Eigen::VectorXd &outData) const;
```
uekerman, line 67: The "3" is the number of components, right?
So, for scalar data the last two components will remain unused?

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -287,7 +292,22 @@ void PartitionOfUnityMapping<RADIAL_BASIS_FUNCTION_T>::mapConsistent(const time:
...
+  auto maxErrors = std::array<double, 3>{-1, -1, -1};
+  for (auto &cluster : _clusters) {
+    auto errors = cluster.mapConsistent(inData, outData);
+    for (int i = 0; i < 3; ++i) {
+      maxErrors[i] = std::max(maxErrors[i], errors[i]);
+    }
+  }
+  std::vector<double> logErrors;
+  std::copy_n(maxErrors.begin(), inData.dataDims, std::back_inserter(logErrors));
+  // Do we want to log every iteration about disabled error metrics?
```
uekerman, line 305: No, why should we? I would remove this completely.

src/mapping/config/MappingConfiguration.cpp
uekerman: Could we give an estimate on the "computationally expensive"? 5x higher than the mapping without cross validation or similar?

uekerman (commented):
Still a few hacks in here, but it is still a draft :smile: 

Could you please give an example output for some tutorial?

Currently, we can only switch it on or off and default is off. Maybe we want sth in between in the end(?)

PR #2155, opened 2024-12-16
Release preCICE using annotated tags

davidscn:
Annotated tags include metadata (message, author, timestamp) that tools like [Software Heritage](https://docs.softwareheritage.org/user/faq/index.html#i-cannot-find-all-my-releases-in-a-git-repository-archived-in-software-heritage-what-should-i-do) rely on for properly displaying and recognizing "releases". Software heritage is an additional archiving platform typically encouraged to use by Darus. preCICE is already there, but the releases are currently not picked-up, so it would be advantageous to have the preCICE releases there by default.


fsimonis:
I just read up on annotated tags.

It turns out that non-annotated tags are for local development only, which is why they aren't pushed nor fetched by default.

Annotated tags are intended for anything publicly shared such as releases.

We should use them for all released by default.
Sadly there is currently no way to enforce this using a ruleset.

MakisH (approved):
Good to learn about the common practices this. Yes, let's do annotated tags!

Advantages:

- More information (complete objects, instead of just pointers to a commit)
- Get automatically copied into mirror repositories
- I understand that they actually work as we thought tags work. The lightweight tags we use now seem to be pretty much like branches, as they can change where they point to.

PR #2079, opened 2024-09-17
Replace Ginkgo triangular solver by dense BLAS trsv

davidscn:
## Main changes of this PR

.. to replace the Ginkgo triangular solver (only relevant for the QR-decomposition).

## Motivation and additional information

More of a performance regression (loosely related to #2051)

The mapData timings
[comparison-mapt.pdf](https://github.com/user-attachments/files/17029800/comparison-mapt.pdf)

and triangular solve timings
[comparison-trsvt.pdf](https://github.com/user-attachments/files/17029795/comparison-trsvt.pdf)

I started a discussion in the Ginkgo project, let's see if I get some feedback there.

I still need to clean-up the current state. Also (FYI) the assembly timings:

[comparison-assemblyt.pdf](https://github.com/user-attachments/files/17029820/comparison-assemblyt.pdf)

## Author's checklist

* [x] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [x] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [x] I added a test to cover the proposed changes in our test suite.
* [x] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).
* [x] I sticked to C++17 features.
* [x] I sticked to CMake version 3.22.1.
* [x] I squashed / am about to squash all commits that should be seen as one.

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [ ] Do you understand the code changes?

PR #2161, opened 2025-01-08
Separate Ginkgo mapping compilation

fsimonis:
## Main changes of this PR

This RP builds Ginkgo mappings as a static library which is then linked into the preCICE core.
It also adds support for Clang Cuda, which is an alternative to the Nvidia Cuda compiler.

Linking against Kokkos with Cuda enabled, builds the entirety of preCICE as Cuda code.

This leads to compilation issues deep inside Boost for some Cuda and boost combinations.
The solution is to conceal all Kokkos functionality inside a single target which is then linked against the `preciceCore`.
For some reason, using CMake Object libraries doesn't work, so static libraries is the remaining option.

This change is required to reliably work with Spack.

## Author's checklist


src/mapping/impl/BasisFunctions.hpp
```diff
@@ -1,6 +1,18 @@
 #pragma once
 
-#if defined(__NVCC__)
+#if defined(__NVCC__) || (defined(__CUDA__) && defined(__CUDA_ARCH__))
```
fsimonis, line 3: The right side detects Clang Cuda.

CMakeLists.txt
```diff
@@ -422,46 +422,60 @@ endif()
...
   if(PRECICE_WITH_CUDA)
-    target_compile_definitions(preciceCore PUBLIC PRECICE_WITH_CUDA)
+    target_compile_definitions(ginkgoKernels PUBLIC PRECICE_WITH_CUDA)
     # Required to make nvcc happy
-    target_compile_definitions(preciceCore PRIVATE BOOST_PP_VARIADICS=1)
-    target_sources(preciceCore PRIVATE ${PROJECT_SOURCE_DIR}/src/mapping/device/CudaQRSolver.cuh ${PROJECT_SOURCE_DIR}/src/mapping/device/CudaQRSolver.cu)
-    target_link_libraries(preciceCore PRIVATE cusolver)
+    target_compile_definitions(ginkgoKernels PRIVATE BOOST_PP_VARIADICS=1)
+    target_sources(ginkgoKernels PRIVATE ${PROJECT_SOURCE_DIR}/src/mapping/device/CudaQRSolver.cuh ${PROJECT_SOURCE_DIR}/src/mapping/device/CudaQRSolver.cu)
+    target_link_libraries(ginkgoKernels PRIVATE cusolver)
```
davidscn, line 457: Can you rebase this on develop? I think `cublas` and `hipblas` are missing here. https://github.com/precice/precice/pull/2079 should give a conflict

davidscn:
> Linking against Kokkos with Cuda enabled, builds the entirety of preCICE as Cuda code.

The Kokkos wrapper will check preCICE yes, but nvcc only gets active for `cu` or `cuh` files.

> This leads to compilation issues deep inside Boost for some Cuda and boost combinations.

Can you document here which kind of issues. I remember that we had it working with spack in this state

https://github.com/precice/precice/pull/2051#issuecomment-2238557328

using CUDA and spack. However I don't remember if it relied on a system boost or on a spack boost.

> For some reason, using CMake Object libraries doesn't work, so static libraries is the remaining option.

How does this affect the visibility of the code elements after linking into 'hidden visibility' preciceCore.

CMakeLists.txt
fsimonis: Fixed

fsimonis:
> The Kokkos wrapper will check preCICE yes, but nvcc only gets active for cu or cuh files.

I need to check again, but IIRC this surprisingly turned out not to be the case. The clang Cuda non-wrapper compilation also failed in some way. I definitely need to recheck and document all of this though.

> How does this affect the visibility of the code elements after linking into 'hidden visibility' preciceCore.

Static libraries have no concept of visibility. This will leak all symbols contained in the static library.
Not sure if there is some post-processing possible after it gets integrated into the shared library though.

fsimonis:
> It is also worth mentioning that this a pure requirement of spack. Without spack, the current CMake setup works just fine.

It works fine on the systems we tested it so far. The question I fail to find an answer for is why. We build Kokkos from source both manually and via Spack and use it via CMake. I ran into issues with Gcc + wrapper using both the system Cuda and the Spack Cuda. I am missing the difference in both cases. Maybe Spack patches Kokkos.

> Does spack use the Kokkos wrapper to build downstream dependencies as well?

No, only preCICE.

fsimonis:
Building and testing with local installations of Kokkos and Ginkgo works flawlessly without the separate target.

Something is really strange on how spack handles kokkos.

fsimonis:
Closing as this wasn't necessary in preCICE and didn't solve our issues with spack

PR #2165, opened 2025-01-13
Force kebab style for XML Attributes and Namespaces

### LINKED ISSUE #2164: Automatically check style of precice-config files in tests
In https://github.com/precice/precice/pull/2159 fixed a recently introduced attribute to follow `kebab-style`. Note that such errors are affecting the user directly and we should therefore be careful to not introduce such mistakes.

I think it would be a good idea to apply some automatic linting to the `precice-config.xml` files in our tests. This should allow us to catch possible errors (assuming that newly introduced attributes come with a corresponding test).

Maybe we could also check in the cpp file where we define the tag using a string. But I have no idea how we can do this.

fsimonis:
## Main changes of this PR

This PR enforces kebab-style for XML attributes and XML Tag namespaces

Closes #2164

This is currently not possible for `XMLTag` names due to the `IQN-` variants using upper-case

## Author's checklist

### DIFF
--- a/src/utils/String.cpp
+++ b/src/utils/String.cpp
@@ -2,6 +2,7 @@
 #include <Eigen/Dense>
 #include <boost/algorithm/string/case_conv.hpp>
 #include <memory>
+#include <regex>
 #include <vector>
 #include "boost/algorithm/string/classification.hpp"
 #include "boost/algorithm/string/split.hpp"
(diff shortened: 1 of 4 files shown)


BenjaminRodenberg (commented):
See my comment in #2164. I like the idea. If we could even check this at compile time (without crazy code nobody understands) I think this would be the cherry on the top. But runtime is also totally fine and I would not invest too much effort here.

fsimonis:
Compile time would be ideal, but this requires baking the string it into the type system. A bit too cumbersome for a quick fix like this.

PR #2166, opened 2025-01-14
Detect Cuda compilation instead of Cuda compiler in BasisFunctions

fsimonis:
## Main changes of this PR

This PR detects Cuda compilation in the RBF Functions, which now also handles Clang Cuda.

Part of #2161

## Author's checklist

### DIFF
--- a/src/mapping/impl/BasisFunctions.hpp
+++ b/src/mapping/impl/BasisFunctions.hpp
@@ -1,6 +1,18 @@
 #pragma once
 
-#if defined(__NVCC__)
+#if defined(__NVCC__) || (defined(__CUDA__) && defined(__CUDA_ARCH__))
+#define PRECICE_CUDA_DEVICE
+#endif
+
+#if defined(__HIPCC__)
+#define PRECICE_HIP_DEVICE
+#endif
+
+#if defined(PRECICE_CUDA_DEVICE) || defined(PRECICE_HIP_DEVICE)
+#define PRECICE_DEVICE
+#endif
+
+#if defined(PRECICE_CUDA_DEVICE)
 
 #include <cuda_runtime.h>
 #include <ginkgo/extensions/kokkos.hpp>
@@ -12,7 +24,7 @@
 #define PRECICE_FMA Kokkos::fma
 #define PRECICE_LOG Kokkos::log
 
-#elif defined(__HIPCC__)
+#elif defined(PRECICE_HIP_DEVICE)
 
 #include <ginkgo/extensions/kokkos.hpp>
 #include <ginkgo/ginkgo.hpp>
@@ -34,7 +46,7 @@
 
 #define NUMERICAL_ZERO_DIFFERENCE_DEVICE 1.0e-14
 
-#if !defined(__NVCC__) || !defined(__HIPCC__)
+#ifndef PRECICE_DEVICE
 #include "logging/Logger.hpp"
 #endif
 #include "math/math.hpp"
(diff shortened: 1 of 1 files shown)


src/mapping/impl/BasisFunctions.hpp
```diff
@@ -1,6 +1,18 @@
 #pragma once
 
-#if defined(__NVCC__)
+#if defined(__NVCC__) || (defined(__CUDA__) && defined(__CUDA_ARCH__))
```
davidscn, line 3:
```suggestion
#if defined(__CUDACC__)
```
I think this should be CUDACC, not sure if the CUDA here is required at all.

Could also be that we are missing here an apple flag. See the Ginkgo version of it

https://github.com/ginkgo-project/ginkgo/blob/75b134aa73fac11e0b085a52cadbecb31af47c9a/accessor/utils.hpp#L15-L32

davidscn: Rather than distinguishing the vendor, we could also distinguish their intention: one (CUDACC) is: which compiler is looking at the code, the other is: do I compile for host or device?

fsimonis: I think this is actually missing the `__Clang__` define. https://releases.llvm.org/19.1.0/docs/CompileCudaWithLLVM.html#detecting-clang-vs-nvcc-from-code

Where and how is this compiled as device code?

davidscn: https://github.com/precice/precice/blob/dd0b7734433021339503ad4bace8309216791e6a/src/mapping/device/GinkgoRBFKernels.cpp#L74

The kernels are executed on the device and the basis functions are included here. 

> Both clang and nvcc define __CUDACC__ during CUDA compilation. You can detect NVCC specifically by looking for __NVCC__.

THere is no value for us in detecting `clang` specifically. As written above, `__NVCC__` is specific to nvcc and not compatible with clang. Both define `__CUDACC__` however.

fsimonis: So, for CUDA we should just check with `_CUDACC_` if this is device code.

davidscn (commented):
We didn't have that until now either, are you sure we don't need the `CUDA_ARCH`? For "host mode" we would still want the checks and logging information, right?

```cpp
#if defined(__clang__) && defined(__CUDA__) && !defined(__CUDA_ARCH__)
// clang compiling CUDA code, host mode.
#endif

#if defined(__clang__) && defined(__CUDA__) && defined(__CUDA_ARCH__)
// clang compiling CUDA code, device mode.
#endif
```

fsimonis:
> We didn't have that until now either, are you sure we don't need the CUDA_ARCH? For "host mode" we would still want the checks and logging information, right?

Using `__CUDACC__` is defined when compiling Cuda code in both Clang and nvcc.
The proposed code only removes the logging in device code (`__CUDACC__` or `__HIPCC__`). The host code remains untouched.

davidscn:
in host mode the current logging would be disabled, as written above. Only checking additionally that `CUDA_ARCH` is __not__ defined adjusts compilation.

```cpp
defined(__CUDA__) && !defined(__CUDA_ARCH__)
// compiling CUDA code, host mode.
```

fsimonis:
The preprocessor defines `PRECICE_DEVICE` when compiling Cuda device code. 
https://github.com/precice/precice/blob/563583dfee648f9962010cad864ed4d07be905ed/src/mapping/impl/BasisFunctions.hpp#L3-L5

This disables the logger include
https://github.com/precice/precice/blob/563583dfee648f9962010cad864ed4d07be905ed/src/mapping/impl/BasisFunctions.hpp#L41-L43

and disables the logging in the functions bodies
https://github.com/precice/precice/blob/563583dfee648f9962010cad864ed4d07be905ed/src/mapping/impl/BasisFunctions.hpp#L183-L187

`__CUDACC__` is undefined when compiling normal C++ for the host and logging is active.
This results in different code when this is included by a compilation unit to be compiled for the device, which shouldn't result in ODR violations.

davidscn:
> The preprocessor defines PRECICE_DEVICE when compiling Cuda device code.

Yes, because the macro now simply detects the cuda compiler as written in the documentation you linked above. Thus, the preprocessor define is now active when compiling Cuda device code and when compiling host code using the cuda compiler.

fsimonis:
I misunderstood the documentation. The macro only defines that the current compiler is treating this file as CUDA file. It doesn't distinguish between device and host.

I am closing this PR as I have a way easier idea on how to simplify all of this.

PR #2168, opened 2025-01-15
Simplify compilation of Basisfunctions for devices

fsimonis:
## Main changes of this PR

This PR simplifies the compilation of the Basisfunctions for Cuda and HIP devices by

- Replacing logging in constructors with exceptions
- Conditionally handling either a version for accelerators or a version for the host

Part of #2161
Replaces #2166

## Author's checklist


src/mapping/impl/BasisFunctions.hpp
```diff
@@ -1,22 +1,17 @@
 #pragma once
 
-#if defined(__NVCC__)
-
+#ifdef __CUDA_ARCH__
```
davidscn, line 3: I would still check for the definition of CUDACC here

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -1,22 +1,17 @@
...
-#include <ginkgo/extensions/kokkos.hpp>
-#include <ginkgo/ginkgo.hpp>
+#endif
 
-#define BOOST_PP_VARIADICS 1
-#define PRECICE_HOST_DEVICE __host__ __device__
-#define PRECICE_MEMORY_SPACE __device__
-#define PRECICE_FMA Kokkos::fma
-#define PRECICE_LOG Kokkos::log
+#ifdef __HIP_ARCH__
```
davidscn, line 7: Similar here for HIPCC

davidscn (commented):
From the developer perspective, I think the proposed solution looks actually more complex and artificial in terms of the splitting. 

As an alternative, I would maybe propose to remove the logger entirely from the basis function implementation and check (we actually do this I think) in the configuration. We could still use an assert (which doesn't require a logger IIRC) in the basis function (if cuda allows) or remove it completely.

Other concern of this approach is somewhat performance related, as the basis function implementation is rather sensitive wrt to the assembly performance. However, I guess the constructor is maybe not critical part and unity builds actually enable visibility again.

src/mapping/impl/BasisFunctions.hpp
fsimonis: Right. I wanted to revert that.

fsimonis:
> As an alternative, I would maybe propose to remove the logger entirely from the basis function implementation and check (we actually do this I think) in the configuration. We could still use an assert (which doesn't require a logger IIRC) in the basis function (if cuda allows) or remove it completely.

That's another more minimal option. The preCICE assertion uses quite a bit of Boost PP stuff, which made nvcc unhappy in the past. The simplest solution is probably to throw a https://en.cppreference.com/w/cpp/error/invalid_argument in the constructor.

> Other concern of this approach is somewhat performance related, as the basis function implementation is rather sensitive wrt to the assembly performance. However, I guess the constructor is maybe not critical part and unity builds actually enable visibility again.

We only construct these objects once in the Configuration. That should have no impact on performance.

davidscn (approved):
That's a nice solution :+1:

Issue #1907, opened 2023-11-30
Unable to add a further serial-explicit coupling scheme to compositional system test


carme-hp:
**Please describe the problem you are trying to solve.**
I want to extent the system test [TwoActivatedMuscles](https://github.com/precice/precice/blob/develop/tests/serial/compositional/TwoActivatedMuscles.xml) by adding a further serial-explicit coupling scheme.

Currently the scheme considers the following setup:
![current_test](https://github.com/precice/precice/assets/71499004/3cab731c-9509-4666-a94b-1de83f845e1e)
Note that in this setup there is already a circular loop, which is closed by sending the stretch from M2SM to M1.

I want to add another coupling scheme which sends stretch from M1SM to M2, which means closing an additional circular loop.
This is the final setup I would like to test:

![target_setup](https://github.com/precice/precice/assets/71499004/29933b03-d1b0-4d41-bc72-c3f97f10088f)

When I try to run the target setup the test hangs. I have also investigated a simpler case: Instead of sending stretch from M2SM to M1, I send stretch from M1SM to M2. I have observed that this test also hangs (it seems to hang in the call to `participant.advance()`. But, if I change the order of the participants in the parallel-implicit scheme, the test passes!

```diff
 <coupling-scheme:parallel-implicit>
-     <participants first="M1SM" second="M2SM" />
+    <participants second="M1SM" first="M2SM" />
```

If the order of the participants has to be different for each of the two loops, I don't see how can I have the two loops simultaneously. 

**Describe the solution you propose.**

**Describe alternatives you've considered**

**Additional context**

uekerman:
When building cycles with serial coupling schemes, one can easily run into a deadlock. Remedies are either parallel coupling schemes or flipping first and second.
I am afraid that the issue here is a wontfix. I modified the test in #1911 to make it pass.
I will try to summarize the situation in a new docs section.

PR #2175, opened 2025-01-21
Add five participants test

carme-hp:

This PR adds a compositional coupling test, in particular a test that uses 5 participants. The test is motivated by my muscle-OpenDiHu application case. The config file is ilustrated here:

5_participants

I have hidden data access, communicators, mapping and coupling schemes for better visibility. There are 
- 2 serial-explicit schemes (unidirectional)
- 2 serial-explicit schemes (bidirectional)
- 1 multi-coupling scheme (participants M1SM, M2SM and Tendon)

This test is related to https://github.com/precice/precice/issues/1907. When tests were restricted to 4 ranks, I adapted my use-case by removing the tendon: Instead of a multi-coupling scheme connecting M1SM and M2SM through the tendon, I had a parallel-implicit scheme connecting M1SM and M2SM directly. There I observed that I was not able to add the 1 of the 2 serial-explicit schemes (unidirectional) I have in this test, and that the one I could add depended on who was the first participant in the parallel-implicit scheme. 
However adding the second serial-explicit (unidirectional) scheme here did not raise any problem.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/tests/serial/compositional/FiveParticipants.cpp
+++ b/tests/serial/compositional/FiveParticipants.cpp
@@ -1,0 +1,135 @@
+#ifndef PRECICE_NO_MPI
+
+#include "testing/Testing.hpp"
+
+#include <boost/test/tools/detail/per_element_manip.hpp>
+#include <precice/Participant.hpp>
+#include <vector>
+
+BOOST_AUTO_TEST_SUITE(Integration)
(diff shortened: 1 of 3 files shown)


fsimonis (approved):
Tested locally and its looking good!

PR #2181, opened 2025-01-24
Add initial data in watch integrals and watch points

### LINKED ISSUE #2180: Export initial data in watchpoint

Watchpoints currently do not include initial data. There are two reasons why initial data is useful:

1) Plotting: especially when comparing watchpoints on different time grids. If the initial data is missing this looks odd.
2) Debugging: initial data has an influence on the final result and the effect of setting `initialize="true"/"false"` would be easier to spot if watchpoints include initial data.

I will add a corresponding PR in a moment. Important: Adding initial data to watchpoints would change the behavior of preCICE and therefore be a breaking change. For adding this feature in v3 we have to make it configurable and set the default to not writing initial data.

**Additional context**
While working on #2172 I observed a major influence of setting `initialize="true"/"false"`on the results (I will provide a detailed update in #2172 later). Having initial data in the watchpoint would be useful for debugging in this case.

BenjaminRodenberg:
## Main changes of this PR

Closes #2180 

See #2180 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/impl/ParticipantState.cpp
+++ b/src/precice/impl/ParticipantState.cpp
@@ -361,8 +361,6 @@ void ParticipantState::exportInitial()
     context.exporter->doExport(0, 0.0);
   }
 
-  // @todo make this conditional depending on config.
-  // Export watch point data
   for (const PtrWatchPoint &watchPoint : watchPoints()) {
     watchPoint->exportPointData(0.0);
   }
--- a/src/precice/impl/ParticipantState.cpp
+++ b/src/precice/impl/ParticipantState.cpp
@@ -366,10 +366,6 @@ void ParticipantState::exportInitial()
   for (const PtrWatchPoint &watchPoint : watchPoints()) {
     watchPoint->exportPointData(0.0);
   }
-
-  for (const PtrWatchIntegral &watchIntegral : watchIntegrals()) {
-    watchIntegral->exportIntegralData(0.0);
-  }
 }
 
 bool ParticipantState::hasExports() const
--- a/src/precice/impl/ParticipantState.cpp
+++ b/src/precice/impl/ParticipantState.cpp
@@ -364,6 +364,10 @@ void ParticipantState::exportInitial()
   for (const PtrWatchPoint &watchPoint : watchPoints()) {
     watchPoint->exportPointData(0.0);
   }
+
+  for (const PtrWatchIntegral &watchIntegral : watchIntegrals()) {
+    watchIntegral->exportIntegralData(0.0);
+  }
 }
 
 bool ParticipantState::hasExports() const
--- a/src/precice/impl/ParticipantState.cpp
+++ b/src/precice/impl/ParticipantState.cpp
@@ -360,6 +360,16 @@ void ParticipantState::exportInitial()
     PRECICE_DEBUG("Exporting initial mesh {} to location \"{}\"", context.meshName, context.location);
     context.exporter->doExport(0, 0.0);
   }
+
+  // @todo make this conditional depending on config.
+  // Export watch point data
+  for (const PtrWatchPoint &watchPoint : watchPoints()) {
+    watchPoint->exportPointData(0.0);
+  }
+
+  for (const PtrWatchIntegral &watchIntegral : watchIntegrals()) {
+    watchIntegral->exportIntegralData(0.0);
+  }
 }
 
 bool ParticipantState::hasExports() const
--- a/tests/serial/watch-integral/WatchIntegralScaleAndNoScale.cpp
+++ b/tests/serial/watch-integral/WatchIntegralScaleAndNoScale.cpp
@@ -98,6 +98,7 @@ BOOST_AUTO_TEST_CASE(WatchIntegralScaleAndNoScale)
       std::string fileName = "precice-SolverTwo-watchintegral-WatchIntegral.log";
       auto        result   = readDoublesFromTXTFile(fileName, 3);
       auto        expected = std::vector<double>{
+          0.0, 0.0, 0.0,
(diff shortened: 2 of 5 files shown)


fsimonis:
Is there any good reason to not always exclude the initial data?

I would consider this a bugfix. @uekerman what do you think?

BenjaminRodenberg:
Btw: Tests are failing because all watchpoints now include the initial data and this does not match the data the test expects.

BenjaminRodenberg:
I removed quite some code from the original PR:

* [0bd3a42](https://github.com/precice/precice/pull/2181/commits/0bd3a4200dff31fc27ab36cf45ec5a788269ca65) removes the code influencing the watch integrals. Then we do not have to update any tests.
* [2dd21ad](https://github.com/precice/precice/pull/2181/commits/2dd21adf49ca94d3279e3e26215143a8e652a9c9) removes the code for making it configurable. Assuming we consider this as a bugfix and non-breaking.

Let's try to fix the scope of this PR next. Then we can also revert these commits and add the required code again if necessary.

BenjaminRodenberg:
@MakisH added you as a reviewer after our discussion in the telco. As far as I understood we are putting this PR on hold until some other issues are resolved. Could you maybe provide pointers to these issues here and review/merge as soon as everything looks good from your perspective?

MakisH:
I agree that this is a nice addition. What I meant in the telco: let's wait before merging this (and https://github.com/precice/precice/pull/2132) until we conclude on the regression encountered in https://github.com/precice/precice/pull/2052 and update the reference results, to not run after multiple regressions at the same time.

I think there is not much left to do there, I will write in both PRs once we can go ahead.

src/precice/impl/ParticipantState.cpp
```diff
@@ -360,6 +360,10 @@ void ParticipantState::exportInitial()
     PRECICE_DEBUG("Exporting initial mesh {} to location \"{}\"", context.meshName, context.location);
     context.exporter->doExport(0, 0.0);
   }
+
+  for (const PtrWatchPoint &watchPoint : watchPoints()) {
+    watchPoint->exportPointData(0.0);
```
fsimonis, line 365: We should do the same for the WatchIntegrals for consistency

BenjaminRodenberg: Ok. Then I think I will also have to update the test accordingly. Will take care of it.

BenjaminRodenberg: Done.

tests/serial/watch-integral/WatchIntegralScaleAndNoScale.cpp
```diff
@@ -98,6 +98,7 @@ BOOST_AUTO_TEST_CASE(WatchIntegralScaleAndNoScale)
       std::string fileName = "precice-SolverTwo-watchintegral-WatchIntegral.log";
       auto        result   = readDoublesFromTXTFile(fileName, 3);
       auto        expected = std::vector<double>{
+          0.0, 6.5, 3.0,
```
BenjaminRodenberg, line 101: I would have expected `0.0, 0.0, 0.0` here. Not sure where the `6.5` comes from. @fsimonis any idea?

tests/serial/watch-integral/WatchIntegralScaleAndNoScale.cpp
```diff
@@ -115,6 +116,7 @@ BOOST_AUTO_TEST_CASE(WatchIntegralScaleAndNoScale)
       std::string fileName = "precice-SolverTwo-watchintegral-WatchIntegralNoScale.log";
       auto        result   = readDoublesFromTXTFile(fileName, 3);
       auto        expected = std::vector<double>{
+          0.0, 6.0, 3.0,
```
BenjaminRodenberg, line 119: See above. Again: I would have expected `0.0, 0.0, 0.0`.

BenjaminRodenberg (commented):
I added the watch integrals. I'm a bit surprised by the results in the test. As mentioned earlier: I do not work with watch integrals so there might be some misunderstanding. The `3.0` in the last column surprises me generally: I thought the last column is the are of the surface which should be `1.0` for the given triangle.

tests/serial/watch-integral/WatchIntegralScaleAndNoScale.cpp
BenjaminRodenberg: Checked this again: I think the last column is clear to me now. We consider two edges with a total length of `3.0`. The second column still looks dubious to me.

BenjaminRodenberg: If I change the coupling scheme in the test to `parallel-implicit` the results look like I would expect them. This also means that the output of the integrals depends on the coupling scheme. I will open a bug report in a moment since this kind of behavior looks wrong to me.

BenjaminRodenberg: See #2187 for more information.

BenjaminRodenberg:
At the moment we can either merge #2188 or #2181 first. After merging one of the two PRs we potentially need to update some tests in the other one.

BenjaminRodenberg:
Added the `help wanted` label to indicate that merging this PR currently depends on another one. We could also consider introducing a new label `stuck` or `waiting` for this purpose.

BenjaminRodenberg:
Not that https://github.com/precice/precice/pull/2052 is merged I will merge this PR.

MakisH:
FYI, in the end this did not conflict with the reference results already used by the system tests: https://github.com/precice/tutorials/actions/runs/13460800713

Issue #2185, opened 2025-01-27
NP mapping on macOS maps to more triangles that one other systems


fsimonis:
**Describe your setup**

Operating system (e.g. Linux distribution and version): GitHub runner  `macOS 	3 (M1) 	7 GB 	14 GB 	arm64 macos-latest`
preCICE Version: https://github.com/precice/precice/commit/4749919650e7fd84892f75da3bf0db6bbfafc261
ASTE Version: https://github.com/precice/aste/commit/cf248be176e44a5f54f9545bc8f8773118628b6a

**Describe the problem**

macOS runner maps on more triangles, leading different behaviour of the NP mapping.
This also impacts the average distance to projection, which is a bad sign.

### macOS

```
INFO : Mapping distance min:0 max:0.00950504 avg: 0.000611378 var: 1.42401e-06 cnt: 3458
INFO : Nearest-projections are 3369 triangles, 89 edges, and 0 vertices
```

### everyone and their 🐕 

```
Mapping distance min:0 max:0.00950504 avg: 0.000609515 var: 1.39599e-06 cnt: 3458
Nearest-projections are 3362 triangles, 96 edges, and 0 vertices
```

**Step To Reproduce**
1. Checkout https://github.com/precice/aste/pull/212
2. Remove the `-L mapping_tester_serial` argument to ctest from the mac workflow.
3. Observe the test fail https://github.com/precice/aste/actions/runs/12952056089/job/36222700005#step:9:211

**Expected behaviour**

The test should map to the same primitives.

**Additional context**

This is an image of the worst change in mapping accuracy. The surface is the from-mesh and the points is a selection of the to-mesh. The mesh is pretty bad there (given that it should be the round front of a wing).
![Image](https://github.com/user-attachments/assets/f78b6a05-f1eb-4827-8fc5-acd649253627)

fsimonis:
Ok, the problem is the following:

We select the nearest n=4 triangles as candidates.
3 are on the upper side of the blade but don't result in an interpolating projection and are thus discarded.

The 4th one is from the bottom side of the blade. Fundamentally, these shouldn't be picked as there are way better projections on edges from the upper side.

There are two triangles that meet at the same Vertex which is also the closest vertex to the search point. Both lead to AABBs that have the same distance to the search point. Triangle A results in an interpolation, while Triangle B doesn't.
On macOS, the 4th-nearest neighbor is A.
On other systems, its B.

[...]

PR #2186, opened 2025-01-30
Add collective projection

fsimonis:
## Main changes of this PR

This PR changes the nearestProjection to consider all candidates and selecting the one with the closest interpolating projection to the search point.

Test solution for #2185
There are two variants proposed here

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/query/Index.cpp
+++ b/src/query/Index.cpp
@@ -274,11 +274,55 @@ std::vector<TetrahedronID> Index::getEnclosingTetrahedra(const Eigen::VectorXd &
 
 ProjectionMatch Index::findNearestProjection(const Eigen::VectorXd &location, int n)
 {
+#if 0
+  std::vector<ProjectionMatch> candidates;
   if (_mesh->getDimensions() == 2) {
-    return findEdgeProjection(location, n, findVertexProjection(location));
+    candidates.reserve(1+n);
   } else {
-    return findTriangleProjection(location, n, findVertexProjection(location));
+    candidates.reserve(1+2*n);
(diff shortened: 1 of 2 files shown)


src/query/Index.cpp
```diff
@@ -274,11 +274,55 @@ std::vector<TetrahedronID> Index::getEnclosingTetrahedra(const Eigen::VectorXd &
...
+  if (_mesh->getDimensions() == 3) {
+    for (const auto &match : getClosestTriangles(location, n)) {
+      auto polation = mapping::Polation(location, _mesh->triangles()[match.index]);
+      if (polation.isInterpolation()) {
+        candidates.emplace_back(std::move(polation));
+      }
+    }
+  }
+
+  return *std::min_element(candidates.begin(), candidates.end());
```
fsimonis, line 304: This is the vector variant

src/query/Index.cpp
```diff
@@ -274,11 +274,55 @@ std::vector<TetrahedronID> Index::getEnclosingTetrahedra(const Eigen::VectorXd &
...
+
+  if (_mesh->getDimensions() == 3) {
+    for (const auto &match : getClosestTriangles(location, n)) {
+      auto polation = mapping::Polation(location, _mesh->triangles()[match.index]);
+      if (polation.isInterpolation() && polation < nearest) {
+        nearest = std::move(polation);
+      }
+    }
   }
+  return nearest;
```
fsimonis, line 324: This is the min variant

PR #2190, opened 2025-02-04
Avoid sample copies when not interpolating

fsimonis:
## Main changes of this PR

This PR prevents copying data values when not interpolating by creating a proxy-class called `SampleResult`.
Results are either

* non-owning, pointing to an existing data. This is common for reading at the beginning or end of the time window.
* owning, including the interpolated data. This is always the case  when interpolating data.

This prevents copies of the whole data sample when:

* reading data
* accelerating data

## Author's checklist

### DIFF
--- a/src/acceleration/BaseQNAcceleration.cpp
+++ b/src/acceleration/BaseQNAcceleration.cpp
@@ -639,15 +639,15 @@ void BaseQNAcceleration::concatenateCouplingData(Eigen::VectorXd &data, Eigen::V
 
     for (int i = 0; i < timeGrid.size(); i++) {
 
-      Eigen::VectorXd values    = cplData.at(id)->timeStepsStorage().sample(timeGrid(i));
-      Eigen::VectorXd oldValues = cplData.at(id)->getPreviousValuesAtTime(timeGrid(i));
+      auto current = cplData.at(id)->timeStepsStorage().sample(timeGrid(i));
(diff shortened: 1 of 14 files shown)


fsimonis:
Something is fishy with the tests.

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -630,15 +630,15 @@ void BaseQNAcceleration::concatenateCouplingData(Eigen::VectorXd &data, Eigen::V
...
+      auto old     = cplData.at(id)->getPreviousValuesAtTime(timeGrid(i));
 
       PRECICE_ASSERT(data.size() >= offset + dataSize, "the values were not initialized correctly");
       PRECICE_ASSERT(oldData.size() >= offset + dataSize, "the values were not initialized correctly");
 
       for (Eigen::Index i = 0; i < dataSize; i++) {
-        data(i + offset)    = values(i);
-        oldData(i + offset) = oldValues(i);
+        data(i + offset)    = current.values()(i);
+        oldData(i + offset) = old.values()(i);
```
BenjaminRodenberg, line 641:
```suggestion
        data(i + offset)    = current(i);
        oldData(i + offset) = old(i);
```

BenjaminRodenberg (approved):
Looks good :+1: Only minor comment left.

PR #2191, opened 2025-02-04
Fix finding existing stamples

### LINKED ISSUE #2160: Crashing without error message after thousands of time steps
**Edit:** There were some miscommunications when I first opened the issue. I have now edited the title and description. I previously thought the issue was due to the first/second in the coupling scheme and that it happened in the first time step. This is not the case.

**Describe your setup**

Operating system (e.g. Linux distribution and version): ? (system of @Ccaccia73)
preCICE Version: v3.1.2

**Describe the problem**

[This configuration file](https://gitlab.com/stilita/lockin-test/-/blob/ac2c4b2921a514d8c3a5e3c8d39d1c993698ea7b/05_Cylinder-OF-MDByn_coarse/precice-config.xml) defines an FSI simulation, where:

- a serial-implicit scheme is used
- as first participant, `Solid` is defined, and `Fluid` as second (the opposite of our tutorials)
- an IQN-ILS acceleration is defined, applied on the data `Forces0` of `Solid-Mesh` (provided by the **first** participant, and exchanged from the second to the first participant. This should be legal, according to our [documentation](https://precice.org/configuration-acceleration.html#coupling-data-and-primary-data)).
- A 3D mesh is defined for a pseudo-2D case. Coupling OpenFOAM and MBDyn.

This config leads to a floating-point exception in `readData` (unfortunately, I only have a stack trace without debug symbols for preCICE):

```
(after thousands of time steps)

forces forces write:
    Sum of forces
        Total    : (92.6607 6.44218e-06 -4.82986e-22) <-- Are the z values expected?
        Pressure : (88.9805 6.8383e-06 -5.04853e-25)
        Viscous  : (3.68025 -3.96123e-07 -4.82481e-22)
    Sum of moments
        Total    : (8.90077e-21 1.84687e-19 2.90562e-08)
        Pressure : (-1.38247e-21 -8.58675e-25 2.5865e-12)
        Viscous  : (1.02832e-20 1.84688e-19 2.90536e-08)
    writing force and moment files.

    functionObjects::CourantNo Co1 writing field: Co
---[precice] Mapping "Forces0" for t=68.47 from "Fluid-Mesh" to "Solid-Mesh"
---[precice] relative convergence measure: relative two-norm diff of data "Displacements0" = 9.71e-02, limit = 1.00e-04, normalization = 1.99e-05, conv = false
---[precice] relative convergence measure: relative two-norm diff of data "Forces0" = 1.81e-05, limit = 1.00e-04, normalization = 5.59e-02, conv = true
---[precice] Mapping "Displacements0" for t=68.47 from "Solid-Mesh" to "Fluid-Mesh"
---[precice] iteration: 2 of 50 (min 1), time-window: 6847, time: 68.46 of 1000, time-window-size: 0.01, max-time-step-size: 0.01, ongoing: yes, time-window-complete: no, read-iteration-checkpoint 
[5] [stack trace]
[5] =============
[5] #1  Foam::sigFpe::sigHandler(int) in /usr/lib/openfoam/openfoam2312/platforms/linux64GccDPInt32Opt/lib/libOpenFOAM.so
[5] #2  ? in /lib/x86_64-linux-gnu/libc.so.6
[5] #3  ? in /lib/x86_64-linux-gnu/libprecice.so.3
[5] #4  ? in /lib/x86_64-linux-gnu/libprecice.so.3
[5] #5  preciceAdapter::Interface::readCouplingData(double) in ~/OpenFOAM/claudio-v2312/platforms/linux64GccDPInt32Opt/lib/libpreciceAdapterFunctionObject.so
(...)

[5] =============
[HP-Z840:130105] *** Process received signal ***
[HP-Z840:130105] Signal: Floating point exception (8)
[HP-Z840:130105] Signal code:  (-6)
[HP-Z840:130105] Failing at address: 0x3e80001fc39
[HP-Z840:130105] [ 0] /lib/x86_64-linux-gnu/libc.so.6(+0x42520)[0x7b62aac42520]
(...)
```

**Step To Reproduce**

Not clear at the moment. I was only given the [case config](https://gitlab.com/stilita/lockin-test) and logs, which depends on the MBDyn and OpenFOAM adapters.

**Expected behaviour**

preCICE reports an error and exits.

**Additional context**

This issue was observed by @Ccaccia73. Please add more context here.

A seemingly similar issue (crashing after thousands of time steps) is also reported by another user in the forum: https://precice.discourse.group/t/problem-coupling-openfoam-and-python/2299/2

fsimonis:
## Main changes of this PR

This PR fixes a bug in the way we currently try to find stample.
In some situations, this can lead to out of bounds access.

Fixes #2160

## Author's checklist

### DIFF
--- a/docs/changelog/2191.md
+++ b/docs/changelog/2191.md
@@ -1,0 +1,1 @@
+- Fixes a bug handling timestamp in sample retrieval which leads to crashes for some combinations of time-window-sizes and time-windows.
--- a/src/time/Storage.cpp
+++ b/src/time/Storage.cpp
@@ -206,10 +206,14 @@ Eigen::VectorXd Storage::sample(double time) const
 
   PRECICE_ASSERT(usedDegree >= 1);
 
-  //Return the sample corresponding to time if it exists
-  const int i = findTimeId(time);
-  if (i > -1) {                              // _stampleStorage contains sample at given time
-    return _stampleStorage[i].sample.values; // don't use getTimesAndValues, because this would iterate over the complete _stampleStorage.
+  // Find existing samples
+  for (const auto &stample : _stampleStorage) {
+    if (math::equals(stample.timestamp, time)) {
+      return stample.sample.values;
+    }
+    if (math::greater(stample.timestamp, time)) {
+      break;
+    }
   }
 
   //Create a new bspline if _bspline does not already contain a spline
@@ -246,18 +250,6 @@ time::Sample Storage::getSampleAtBeginning()
 time::Sample Storage::getSampleAtEnd()
 {
   return _stampleStorage.back().sample;
-}
-
-int Storage::findTimeId(double time) const
-{
-  int i = 0;
-  while (math::smallerEquals(_stampleStorage[i].timestamp, time)) {
-    if (math::equals(_stampleStorage[i].timestamp, time)) {
-      return i;
-    }
-    i++;
-  }
-  return -1; // time not found in times
 }
 
 } // namespace precice::time
--- a/src/time/Storage.hpp
+++ b/src/time/Storage.hpp
@@ -173,8 +173,6 @@ private:
   time::Sample getSampleAtBeginning();
 
   time::Sample getSampleAtEnd();
-
-  int findTimeId(double time) const;
 };
 
 } // namespace precice::time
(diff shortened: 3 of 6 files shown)


BenjaminRodenberg:
Can you turn the configuration mentioned in https://precice.discourse.group/t/problem-coupling-openfoam-and-python/2299/18 into a test to make sure this does not happen again?

fsimonis:
@BenjaminRodenberg I was able to trigger the warning with Dt=10.01 and tmax=72 in 7 time windows. This is fine for a test.

fsimonis:
Added test which fails locally on develop.

PR #2192, opened 2025-02-05
Dynamic access region

### LINKED ISSUE #2094: Change direct-access region at runtime

At the workshop I was made aware that the remeshing feature could be nicely combined with the direct-access feature to change the region of interest at runtime.

While changes for the remeshing should allow this case nice, there is currently no way to trigger this.

Allow `resetMesh()` on the received mesh followed by `setMeshAccessRegion()`.

Allow `setMeshAccessRegion()` to be called again at similar timings than `resetMesh()`.

**Additional context**

#2070

davidscn:
## Main changes of this PR

Enables to redefine the access region at runtime to support direct-mesh access and just-in-time mapping with remeshing.

See and closes #2094

Depends on #2099 

## Author's checklist

## Reviewers' checklist


src/precice/impl/ParticipantImpl.cpp
```diff
@@ -775,6 +775,25 @@ void ParticipantImpl::resetMesh(
...
+                "This participant attempteded to reset the mesh access region (via \"resetMeshAccessRegion\") on mesh \"{0}\", "
+                "but mesh \"{0}\" is either not a received mesh or its api access was not enabled in the configuration. "
+                "resetMeshAccessRegion({0}) is only valid for (<receive-mesh name=\"{0}\" ... api-access=\"true\"/>).",
+                meshName);
+
+  MeshContext &context = _accessor->meshContext(meshName);
+  PRECICE_WARN_IF(!context.userDefinedAccessRegion, "This participant called \"resetMeshAccessRegion({0})\", but a mesh access region on mesh \"{0}\" was not defined.", meshName);
+  context.userDefinedAccessRegion.reset();
+  context.mesh->resetBoundingBox();
+  _meshLock.unlock(meshName);
```
davidscn, line 794: @fsimonis would you suggest to also reset the actual (received) mesh? I could even forward the call in this case to 'resetMesh'

fsimonis: It will be deleted during the repartitioning. So there is no need to clear the mesh itself as long as the user is aware that the vertex IDs and positions lose their validity after the call to advance.

davidscn:
Moved this one milestone further, due to unforeseen complexity https://github.com/precice/precice/issues/2094#issuecomment-2735604443

fsimonis:
Due to the complexity of this issue, I'll move it to the 3.x milestone.

PR #2193, opened 2025-02-05
Do not check exchangeSubsteps in QN acceleration

### LINKED ISSUE #2172: Regression for subcycling scenario with `substeps="false"`
**Describe your setup**

preCICE Version: `3.1.2`, `ad68bc0` (before https://github.com/precice/precice/pull/2005), `e43ba2f` (after https://github.com/precice/precice/pull/2005)

* preCICE configuration: `PRECICE_FEATURE_PETSC_MAPPING = "OFF"`
* `perpendicular-flap` @ https://github.com/BenjaminRodenberg/tutorials/commit/01f064c4689c153a0161af937de18f15a82943eb (uses [`initialize="true"`](https://github.com/BenjaminRodenberg/tutorials/blob/01f064c4689c153a0161af937de18f15a82943eb/perpendicular-flap/precice-config-template-v3.1.2.xml#L49-L50))
* Openfoam adapter 1.3.1
* FEniCS adapter 2.2.0

**Describe the problem**

The image below shows some data from a subcycling experiment that I presented at PDESoft (see https://mediatum.ub.tum.de/1748260). Here, the perpendicular flap case is compared for different time window sizes with fixed time step sizes for both participants. The exchange of substeps is either turned on or off. When performing this experiment with version `3.1.2` or `ad68bc0` of preCICE I get identical flap displacement to my presentation from PDESoft. For the latest preCICE version and starting at `e43ba2f` I get results that are clearly different.

Image
*edit:* Updated this plot because legend entries were misleading.

**Step To Reproduce**
1. Clone https://github.com/BenjaminRodenberg/tutorials/commit/01f064c4689c153a0161af937de18f15a82943eb, but should also work for other versions of this tutorial or the adapters.
2. Run case with configuration given [here](https://github.com/BenjaminRodenberg/tutorials/blob/01f064c4689c153a0161af937de18f15a82943eb/perpendicular-flap/precice-config-template-v3.1.2.xml). Use `waveform-degree = 1`, `substeps = "false"`, `max-time = "5"`, `time_window_size=1.0"` (again: Other configuration where subcycling without substeps is used should also work.
3. Compare watchpoints for different preCICE versions.

**Expected behaviour**
There should not be a difference between ad68bc0 and e43ba2f since https://github.com/precice/precice/pull/2005 provided an extension of the quasi-Newton acceleration when substeps are used. With `substeps="false"` no change should occur.

**Additional context**
I see two possibilities:

1. The newly implemented quasi-Newton scheme uses substeps even if the user sets `substeps="false"`. This would be a regression.
2. The newly implemented quasi-Newton scheme does not use substeps but still improves the accuracy of the solution. This would be surprising but interesting. (The blue markers are actually more accurate and much closer to the solution including substeps than the red markers)

BenjaminRodenberg:
:warning: This is still WIP and I need to investigate the issue more closely; opening this PR to use CI for running tests :warning: 

## Main changes of this PR

Removes the check of `data->exchangeSubsteps()` from BaseQNAcceleration.

* From my understanding especially the secondary data should be influenced by `substeps="true"/"false"`. If `substeps="false"` data at the beginning of the window will be ignored (or even deleted?).
* ~~Trying to fix #2172~~ (*edit:* was fixed by https://github.com/precice/precice/pull/2194; this turns this PR into a pure refactoring)

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/acceleration/BaseQNAcceleration.cpp
+++ b/src/acceleration/BaseQNAcceleration.cpp
@@ -647,18 +647,9 @@ void BaseQNAcceleration::concatenateCouplingData(Eigen::VectorXd &data, Eigen::V
 
 void BaseQNAcceleration::initializeVectorsAndPreconditioner(const DataMap &cplData)
 {
-
-  // If we are not subcycling then we only want to use the last time step in the QN system
-  bool subcycling = false;
-  for (const auto &data : cplData | boost::adaptors::map_values) {
-    if (data->exchangeSubsteps()) {
-      subcycling = true;
-    }
-  }
-
   // Saves the time grid of each waveform in the data field to be used in the QN method
-  _timeGrids.emplace(cplData, _dataIDs, !subcycling);
-  _primaryTimeGrids.emplace(cplData, _primaryDataIDs, !(subcycling and !_reducedTimeGrid));
+  _timeGrids.emplace(cplData, _dataIDs, false);
+  _primaryTimeGrids.emplace(cplData, _primaryDataIDs, _reducedTimeGrid);
 
   // Helper function
   auto addTimeSliceSize = [&](size_t sum, int id, precice::time::TimeGrids timeGrids) { return sum + timeGrids.getTimeGrid(id).size() * cplData.at(id)->getSize(); };


src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -647,18 +647,9 @@ void BaseQNAcceleration::concatenateCouplingData(Eigen::VectorXd &data, Eigen::V
...
-  for (const auto &data : cplData | boost::adaptors::map_values) {
-    if (data->exchangeSubsteps()) {
-      subcycling = true;
-    }
-  }
-
   // Saves the time grid of each waveform in the data field to be used in the QN method
-  _timeGrids.emplace(cplData, _dataIDs, !subcycling);
-  _primaryTimeGrids.emplace(cplData, _primaryDataIDs, !(subcycling and !_reducedTimeGrid));
+  _timeGrids.emplace(cplData, _dataIDs, false);
```
BenjaminRodenberg, line 651: @NiklasKotarsky From my understanding the secondary data should never be reduced?

BenjaminRodenberg:
For 1d8a096 the CI pipeline is successful. There are three possibilities:

1) The statement that I removed is unnecessary but has no effect.
2) The statement that I removed is needed but the effect is untested. @NiklasKotarsky are we missing a test for a feature?
3) The statement that I removed actually led to a bug. We need a test that will result in a failure for commits before 1d8a096 to avoid this bug in the future.

src/acceleration/BaseQNAcceleration.cpp
NiklasKotarsky: Yes, we need all the time steps in W_k, since we otherwise loose information. The reason for these complications when I was coding it: Re-meshing removes the first sample in the waveform causing the moving of the time grid to fail. This seems to no longer be a problem, since the test complete. However, I will take a closer look today or tomorrow to make sure that there is no issue.

BenjaminRodenberg: Sounds good. If there is an edge case that we are missing now we should have a test. We could also consider excluding that edge case for now in favor of fixing https://github.com/precice/precice/issues/2172 and add it as a feature later.

fsimonis: @NiklasKotarsky We have tests for remeshing with acceleration. https://github.com/precice/precice/tree/develop/tests/remeshing/parallel-implicit/acceleration/IQN-ILS

These should break, right?

```
ctest -R precice.Integration/Remeshing/ParallelImplicit/Acceleration
```

NiklasKotarsky: Yes, in theory. However, after a lot of digging I found that the function `moveTimeGridToNewWindow` https://github.com/precice/precice/blob/294cf77e80f3a3a723e6b111b1a8ea4e1a5c01a1/src/time/TimeGrids.cpp#L32
takes care of this special case without the checks. So my comment from before is really a non issue and everything seems to be working correctly!

NiklasKotarsky:
Thanks for finding and proposing a fix for this bug. I did not properly consider the case where we do not subcycle, but still interpolate between the first and last sample of the waveform iteration. The pull request fixes this: https://github.com/precice/precice/issues/2172#issuecomment-2636252030 by removing the unencessary check of data->exchangeSubsteps() from BaseQNAcceleration. I have checked the interoperability with re-meshing and this does not introduce a regression there, since the loss of the first sample is handled implicitly in the function 'moveTimeGridToNewWindow' https://github.com/precice/precice/blob/294cf77e80f3a3a723e6b111b1a8ea4e1a5c01a1/src/time/TimeGrids.cpp#L32.

BenjaminRodenberg:
@fsimonis Do you think we should merge this fix since it at least resolves a part of the regression in #2172 ? I think we should still add a test here. @NiklasKotarsky if you have an idea for a test it would be very helpful if you could add one. I'm still trying to find the problem mentioned in #2172 with `initialize="true"`.

NiklasKotarsky:
>  @NiklasKotarsky if you have an idea for a test it would be very helpful if you could add one. 

Yes, I can change the test to also cover this use case.

uekerman:
How does this fix relate to the other fix #2194 and the original problem #2172?

BenjaminRodenberg:
> How does this fix relate to the other fix #2194 and the original problem #2172?

The situation here was really a bit confusing. My current point of view on this issue:

* #2194 fixed #2172. #2172 is a critical bug we need to fix before we release.
* #2193 covers some edge cases of #2172 but is from my perspective not the actual fix. I would still suggest to merge it because it simplifies the code. But this is not critical for the release. (let's keep it on the milestone, but we can also shift it to the next release)
* https://github.com/precice/precice/pull/2195 provides potentially some ideas for testing #2193

Main todo:

- [ ] Merge #2194 before #2193.
- [ ] Add tests in #2193 to ensure we do not break certain edge cases. Here, the expertise of @NiklasKotarsky could be very valuable.

NiklasKotarsky (approved):
I have looked through the code one more time and I am sure that this pull request does not change the behavior of the QN methods. However, it simplifies the QN class by removing the exchangeSubsteps variable from the if statements in initializeVectorsAndPreconditioner and I would therefore still want to merge this as a refactoring.

fsimonis (approved):
From a code standpoint, this PR looks ... fine

I'll trust you Waveform guys that this works. :wink:

PR #2195, opened 2025-02-07
Fix QN tests to cover #2172

NiklasKotarsky:
## Main changes of this PR
Changes the VIQNPP test in Acceleration serial test to check that the samples in the waveform are correctly updated according to the expected behaviour in #2193.

Extends test coverage to cover #2172 

## Author's checklist

## Reviewers' checklist


tests/quasi-newton/helpers.cpp
```diff
@@ -295,6 +295,8 @@ void runTestQNWithWaveforms(std::string const &config, TestContext const &contex
 
     interface.readData(meshName, readDataName, {vertexIDs, 2}, dt, {inValues, 2});
 
+    interface.readData(meshName, readDataName, {vertexIDs, 2}, dt, {inValuesOff, 2});
```
BenjaminRodenberg, line 298: @NiklasKotarsky Currently, there seems to be a problem with this test. Can you add some comments to explain the modification?

tests/quasi-newton/serial/TestQN11.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8" ?>
 <precice-configuration experimental="true">
-  <data:scalar name="Data1" waveform-degree="2" />
-  <data:scalar name="Data2" waveform-degree="2" />
+  <data:scalar name="Data1" waveform-degree="1" />
+  <data:scalar name="Data2" waveform-degree="1" />
```
BenjaminRodenberg, line 4: Is this change necessary? I do not really see how this related to the regression.

BenjaminRodenberg (changes requested):

BenjaminRodenberg (commented):
I think now I understand the direction of the tests.

Why do you replace the old ones? You could also add new ones and keep the old ones. I think testing with and without initialization has a value. I would keep both versions.

Additionally: Underrelaxation with initialization was also a weak spot. We could use the same tests and apply underrelaxation instead of QN.

BenjaminRodenberg:
Be aware that there was a bug in the way how the some tests used subcycling. I fixed it in https://github.com/precice/precice/commit/8353de2758ed69cff64d7f1a4bc8a2b11a5a075b. This might also be helpful here.

NiklasKotarsky:
> I think now I understand the direction of the tests.
> 
> Why do you replace the old ones? You could also add new ones and keep the old ones. I think testing with and without initialization has a value. I would keep both versions.

In the old QN test we currently do not check the initial data at all. It would probably be a good idea to check this as well to avoid future pitfalls there as well. The old test are also completely covered by the new tests. 

> Additionally: Underrelaxation with initialization was also a weak spot. We could use the same tests and apply underrelaxation instead of QN.

Yes, I will se if a can configure a test case that fails in a reasonable amount of iterations.

NiklasKotarsky:
After some digging I think that the sample of the start of the time window is actually copied over in the waveform without the fix, since if it weren't preCICE would crash trying to sample outside the waveform. I have also checked that it does not accidentally copy over a later sample. However, I am not exactly sure how the responsible code functions: 

https://github.com/precice/precice/blob/8353de2758ed69cff64d7f1a4bc8a2b11a5a075b/src/cplscheme/BaseCouplingScheme.cpp#L907-L911

tests/quasi-newton/serial/TestQN15.cpp
```diff
@@ -0,0 +1,23 @@
...
+#include "testing/Testing.hpp"
+
+#include <precice/precice.hpp>
+#include "../helpers.hpp"
+
+BOOST_AUTO_TEST_SUITE(Integration)
+BOOST_AUTO_TEST_SUITE(QuasiNewton)
+BOOST_AUTO_TEST_SUITE(Serial)
+PRECICE_TEST_SETUP("SolverOne"_on(1_rank), "SolverTwo"_on(1_rank))
+BOOST_AUTO_TEST_CASE(TestQN11)
```
BenjaminRodenberg, line 12:
```suggestion
BOOST_AUTO_TEST_CASE(TestQN15)
```

tests/quasi-newton/helpers.cpp
```diff
@@ -273,18 +273,28 @@ void runTestQNWithWaveforms(std::string const &config, TestContext const &contex
     }
   }
 
-  int             nSubsteps = 5;             // perform subcycling on solvers. 5 steps happen in each window.
+  int             nSubsteps = 10;            // perform subcycling on solvers. 5 steps happen in each window.
```
BenjaminRodenberg, line 276:
```suggestion
  int             nSubsteps = 10;            // perform subcycling on solvers. 10 steps happen in each window.
```
?

tests/quasi-newton/serial/TestQN11.cpp
```diff
@@ -9,10 +9,10 @@ BOOST_AUTO_TEST_SUITE(Integration)
 BOOST_AUTO_TEST_SUITE(QuasiNewton)
 BOOST_AUTO_TEST_SUITE(Serial)
 PRECICE_TEST_SETUP("SolverOne"_on(1_rank), "SolverTwo"_on(1_rank))
-BOOST_AUTO_TEST_CASE(TestQN11)
+BOOST_AUTO_TEST_CASE(TestQN15)
```
BenjaminRodenberg, line 12: According to the file name this test should be `TestQN11`

NiklasKotarsky:
The QN15 test works on my machine, which means that the initial sample of the waveform iteration is copied over from somewhere. With this in mind I am still unsure what exactly causes the error in #2172 making it difficult to design good test. I have therefore decided to test the QN methods directly in Acceleration tests and making sure I see the behavior expected from pullrequest 2193 and 2194, which fix the issue in 2172.

BenjaminRodenberg:
Maybe this observation helps? https://github.com/precice/precice/issues/2172#issuecomment-2653652042

Generally, I think the amount of tests in #2194 is fine and we can merge that PR. Still: If you would like to contribute additional tests: They are always helpful.

NiklasKotarsky:
> Maybe this observation helps? [#2172 (comment)](https://github.com/precice/precice/issues/2172#issuecomment-2653652042)

Thanks, for the hint. This tells me that there is probably something going on in one of the solvers, which contributes to the instability that we observe. Thus, it will be difficult to design a simple test test for  #2194 and #2193.  From my perspective it still makes sense to merge the changes here to VIQNPP test in AccelerationSerialTest, since this checks that the QN methods exhibit a reasonable behavior in a waveform setting.

BenjaminRodenberg:
FYI: I just added this PR to the milestone for the next release. From my current perspective this PR still needs some work but would be nice to have it in the release. If not: We can still shift it to the next release.

NiklasKotarsky:
Yes, I am going to introduce proper names for the start of the time/end of the time window, similar to what commit c95c6f7 did in AccelerationIntracommTest. I am also going to try to make the sampling in the checks a little bit more easy to read. Should I also update the rest of the test in Acceleration/tests to also check the waveform instead of the values in coupling data?

BenjaminRodenberg:
> Should I also update the rest of the test in Acceleration/tests to also check the waveform instead of the values in coupling data?

Sure. Our goal is to use `couplingData->TimeStepsStorage().sample(...)` instead of values `CouplingData::values()` because this is more explicit.

src/acceleration/test/AccelerationSerialTest.cpp
```diff
@@ -226,6 +235,26 @@ void testVIQNPP(bool exchangeSubsteps)
...
+  BOOST_TEST(testing::equals(data.at(0)->timeStepsStorage().sample(1)(1), 6.10309919173607440257e-01));
+  BOOST_TEST(testing::equals(data.at(0)->timeStepsStorage().sample(1)(2), 1.78402117927690717636e+00));
+  BOOST_TEST(testing::equals(data.at(0)->timeStepsStorage().sample(1)(3), 2.95773243938020513610e+00));
+  BOOST_TEST(testing::equals(data.at(1)->timeStepsStorage().sample(1)(0), 8.28025852497733944046e-02));
+  BOOST_TEST(testing::equals(data.at(1)->timeStepsStorage().sample(1)(1), 8.28025852497733944046e-02));
+  BOOST_TEST(testing::equals(data.at(1)->timeStepsStorage().sample(1)(2), 8.28025852497733944046e-02));
+  BOOST_TEST(testing::equals(data.at(1)->timeStepsStorage().sample(1)(3), 8.28025852497733944046e-02));
+
+  // Check that the values stored in data are updated as well
   BOOST_TEST(testing::equals(data.at(0)->values()(0), -5.63401340929692295845e-01));
```
BenjaminRodenberg, line 258: These lines are duplicates of the ones above and should be removed. No?

BenjaminRodenberg (commented):
I think we have to be a bit careful with the scope of this PR: Currently it is partially 

* a refactoring: replacing `data.at(...)->values()(...)` with `data.at(...)->timeStepsStorage().sample(...)`)
* an extension of the existing test: Additionally check at `sample(0)`.

I think keeping the refactoring and the extension separate would simplify reviewing. But this is also not a must if this results in too much work.

NiklasKotarsky:
> I think we have to be a bit careful with the scope of this PR: Currently it is partially
> 
>     * a refactoring: replacing `data.at(...)->values()(...)` with `data.at(...)->timeStepsStorage().sample(...)`)
> 
>     * an extension of the existing test: Additionally check at `sample(0)`.
> 
> 
> I think keeping the refactoring and the extension separate would simplify reviewing. But this is also not a must if this results in too much work.

I have now moved the refactoring into its own pull request #2212, so that it can be merged independently to #2193.

BenjaminRodenberg:
As far as I understand https://github.com/precice/precice/pull/2212 replaces this PR. Therefore, can we close this PR? But I might also miss something.

NiklasKotarsky:
Pull request  #2193 does not change the behavior of the QN acceleration making it unnecessary to test for it specifically. The test covarge has already been extended to ensure that the first sample is not accelerated in pull request  #2194, which caused the issue in  #2172.

PR #2200, opened 2025-02-17
Update Docker base image to ubuntu:24.04

### LINKED ISSUE #1950: Base image for docker releases

Currently, the docker releases of preCICE use `ubuntu:22.04` as a base image, which is the latest Ubuntu LTS.

For testing, we may want a release of our baseline though, namely `ubuntu:20.04`.
This is especially important for the nightly build of preCICE, as it cannot easily be installed using a Debian package.

We have 3 options:
1. Provide docker releases based on our Ubuntu baseline `ubuntu:20.04`
2. Provide docker releases based on the latest Ubuntu LTS `ubuntu:22.04` (our current approach)
3. Provide docker releases for both, the baseline and the latest Ubuntu LTS version.

If we choose to provide both, then we need another image. I suggest to keep using `precice/precice` for the latest LTS, and `precice/precice-baseline` for the baseline LTS.

**Additional context**

Question raised here https://github.com/precice/precice/pull/1949

MakisH:
## Main changes of this PR

Bumps the base Docker image from `ubuntu:22.04` to `ubuntu:24.04`. Also updates the example version, because 2.1.1 looked weird.

Closes #1950.

I have checked that the images build locally.

Implements the decision in #1950 to use the latest Ubuntu LTS.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/tools/releasing/packaging/docker/README.md
+++ b/tools/releasing/packaging/docker/README.md
@@ -6,12 +6,12 @@ All images contain the user `precice`, allowing them to run executables using MP
 
 ## Release images `release.dockerfile`
 
-This Dockerfile uses named releases of preCICE such as `2.1.1` and installs the attached debian package in the container.
+This Dockerfile uses named releases of preCICE such as `3.1.2` and installs the attached debian package in the container.
 
-Use the following to build the `2.1.1` image locally:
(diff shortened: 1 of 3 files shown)


fsimonis:
Looks good! Thanks for testing locally

PR #2201, opened 2025-02-18
Add warning for configuration

### LINKED ISSUE #531: preCICE Config accepts <max-used-iterations> greater <max-iterations> if <timesteps-reused value="0"/>
preCICE Config accepts `<max-used-iterations>` greater `<max-iterations>` if `<timesteps-reused value="0"/>`. However, this configuration is invalid and we should instead throw an exception: If we do not reuse any previous timesteps (`<timesteps-reused value="0"/>`), we cannot reuse more iterations than we actually allow to be performed in one timestep (`<max-used-iterations>` greater `<max-iterations>`).

Be aware that `<max-used-iterations>` may be greater than `<max-iterations>`, if we reuse previous timesteps.

Fujikawas:
## Main changes of this PR
Add warning info when user set larger `max-used-iterations` than `max-iterations` when no previous time window is reused.

Close #531 

## Author's checklist

* [x] Tested manually with serial QN integration test.

## Reviewers' checklist


src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -302,6 +302,10 @@ void BaseCouplingScheme::initialize()
       // reserve memory and initialize data with zero
       if (_acceleration) {
         _acceleration->initialize(getAccelerationData());
+        std::shared_ptr<precice::acceleration::BaseQNAcceleration> qnAcceleration = std::dynamic_pointer_cast<precice::acceleration::BaseQNAcceleration>(_acceleration);
+        if (qnAcceleration) {
```
fsimonis, line 306: You can use auto and the if init statement to make this a bit more compact. 

```suggestion
        if (auto qnAcceleration = std::dynamic_pointer_cast<precice::acceleration::BaseQNAcceleration>(_acceleration);
            qnAcceleration) {
```

fsimonis (approved):
Minor simplifications. Looks good apart from that!

Issue #669, opened 2020-02-10
Validate preCICE Versions of Participants


fsimonis:
# Problem Description

When running a simulation with multiple participants, there is always the possibility that they may link to differing versions.

Currently, this will run, could work and may crash.

# Suggested Solution

Exchange version information directly after establishing each master connection.
Then assert that the remote version matches the local version.

MakisH:
I think we should implement this only as a warning: if one has some built solver (e.g. with preCICE as a static library) that just works, we should not prevent the user from keep using it. But we should definitely warn her.

fsimonis:
Decision at the coding days is to lock this to the full version and emit an error.

fsimonis:
Closed by #2202

PR #2202, opened 2025-02-18
Add an error if participants use different preCICE versions

carme-hp:
## Main changes of this PR

Directly after connecting the primary ranks, we check the preCICE versions (e.g., 3.1.2) of the requester and acceptor and add an error if they differ. 

```
PRECICE_ERROR("Participant {} uses preCICE version {} but participant {} uses preCICE version {}.", requesterName, requesterVersion, acceptorName, acceptorVersion);
```

This PR is motivated by #669 .

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/m2n/M2N.cpp
+++ b/src/m2n/M2N.cpp
@@ -57,9 +57,7 @@ void M2N::acceptPrimaryRankConnection(
     auto        requesterVersion = acceptorVersion;
     _interComm->send(acceptorVersion, 0);
     _interComm->receive(requesterVersion, 0);
-    if (requesterVersion != acceptorVersion) {
-      PRECICE_ERROR("Participant {} uses preCICE version {} but participant {} uses preCICE version {}.", requesterName, requesterVersion, acceptorName, acceptorVersion);
-    }
+    PRECICE_CHECK(requesterVersion != acceptorVersion, "Participant {} uses preCICE version {} but participant {} uses preCICE version {}.", requesterName, requesterVersion, acceptorName, acceptorVersion);
   }
 
   utils::IntraComm::broadcast(_isPrimaryRankConnected);
@@ -82,11 +80,9 @@ void M2N::requestPrimaryRankConnection(
     //check that local and remote have the same precice version
     std::string requesterVersion = PRECICE_VERSION;
     auto        acceptorVersion  = requesterVersion;
-    _interComm->send(requesterVersion, 0);
     _interComm->receive(acceptorVersion, 0);
-    if (acceptorVersion != requesterVersion) {
-      PRECICE_ERROR("Participant {} uses preCICE version {} but participant {} uses preCICE version {}.", acceptorName, acceptorVersion, requesterName, requesterVersion);
-    }
+    _interComm->send(requesterVersion, 0);
+    PRECICE_CHECK(requesterVersion != acceptorVersion, "Participant {} uses preCICE version {} but participant {} uses preCICE version {}.", requesterName, requesterVersion, acceptorName, acceptorVersion);
   }
 
   utils::IntraComm::broadcast(_isPrimaryRankConnected);
--- a/src/m2n/M2N.cpp
+++ b/src/m2n/M2N.cpp
@@ -4,15 +4,14 @@
 #include "DistributedCommunication.hpp"
 #include "M2N.hpp"
 #include "com/Communication.hpp"
+#include "logging/LogConfiguration.hpp"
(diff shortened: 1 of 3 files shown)


src/m2n/M2N.cpp
```diff
@@ -69,6 +78,15 @@ void M2N::requestPrimaryRankConnection(
     PRECICE_DEBUG("Request primary connection");
     _interComm->requestConnection(acceptorName, requesterName, "PRIMARYCOM", 0, 1);
     _isPrimaryRankConnected = _interComm->isConnected();
+
+    //check that local and remote have the same precice version
+    std::string requesterVersion = PRECICE_VERSION;
+    auto        acceptorVersion  = requesterVersion;
+    _interComm->send(requesterVersion, 0);
+    _interComm->receive(acceptorVersion, 0);
```
fsimonis, line 86: These two should be swapped as we are expecting send and receive to be blocking.

src/m2n/M2N.cpp
```diff
@@ -50,7 +52,14 @@ void M2N::acceptPrimaryRankConnection(
...
     _interComm->acceptConnection(acceptorName, requesterName, "PRIMARYCOM", utils::IntraComm::getRank());
-    _isPrimaryRankConnected = _interComm->isConnected();
+    _isPrimaryRankConnected      = _interComm->isConnected();
+    std::string acceptorVersion  = PRECICE_VERSION;
+    auto        requesterVersion = acceptorVersion;
+    _interComm->send(acceptorVersion, 0);
+    _interComm->receive(requesterVersion, 0);
+    if (requesterVersion != acceptorVersion) {
+      PRECICE_ERROR("Participant {} uses preCICE version {} but participant {} uses preCICE version {}.", requesterName, requesterVersion, acceptorName, acceptorVersion);
+    }
```
fsimonis, line 62: A `PRECICE_CHECK` would be a better fit here.

src/m2n/M2N.cpp
```diff
@@ -69,6 +78,15 @@ void M2N::requestPrimaryRankConnection(
...
     _isPrimaryRankConnected = _interComm->isConnected();
+
+    //check that local and remote have the same precice version
+    std::string requesterVersion = PRECICE_VERSION;
+    auto        acceptorVersion  = requesterVersion;
+    _interComm->send(requesterVersion, 0);
+    _interComm->receive(acceptorVersion, 0);
+    if (acceptorVersion != requesterVersion) {
+      PRECICE_ERROR("Participant {} uses preCICE version {} but participant {} uses preCICE version {}.", acceptorName, acceptorVersion, requesterName, requesterVersion);
+    }
```
fsimonis, line 89: Same as above

fsimonis (approved):

carme-hp:
I modified the error message. 
But worth to mention, the CI does no longer pass. After talking with @fsimonis it was decided to send the version as a vector.

src/m2n/M2N.cpp
```diff
@@ -50,7 +52,12 @@ void M2N::acceptPrimaryRankConnection(
     PRECICE_DEBUG("Accept primary connection");
     PRECICE_ASSERT(_interComm);
     _interComm->acceptConnection(acceptorName, requesterName, "PRIMARYCOM", utils::IntraComm::getRank());
-    _isPrimaryRankConnected = _interComm->isConnected();
+    _isPrimaryRankConnected      = _interComm->isConnected();
+    std::string acceptorVersion  = PRECICE_VERSION;
+    auto        requesterVersion = acceptorVersion;
+    _interComm->send(acceptorVersion, 0);
+    _interComm->receive(requesterVersion, 0);
+    PRECICE_CHECK(requesterVersion != acceptorVersion, "This participant {} uses preCICE version {} but the requester participant {} uses preCICE version {}.", acceptorName, acceptorVersion, requesterName, requesterVersion);
```
fsimonis, line 60: Have a look at that condition :wink:

carme-hp:
> Could we add a sentence like "Mixing preCICE versions can lead to undefined behavior and is, thus, forbidden."?

Done!

carme-hp:
It seems like the failed CI arised because I used the reverse logic, and now it passes without problems. 
@fsimonis does it still make sense to pass the version as an array or should we keep it as a string?

carme-hp:
@fsimonis can you check the changelog? I'm not sure if this is what you mean.

fsimonis:
@carme-hp Have a look at the author's checklist in the PR text. If it's unclear, then we should update it.

PR #2203, opened 2025-02-18
Remove unnecessary write access to `CouplingData::sample()` and make `const`

BenjaminRodenberg:
## Main changes of this PR

Make functions `CouplingData::sample()`, `CouplingData::values()`, `CouplingData::gradients()` read-only (`const`). Data in `CouplingData` should only be modified via the storage.

Storage becomes single source of coupling data, no necessity for synchronizing between storage and values.

Related to #1645 

## Author's checklist

* [x] Merge #2204
* [x] Merge https://github.com/precice/precice/pull/2207
* [x] Merge https://github.com/precice/precice/pull/2213

## Reviewers' checklist


src/acceleration/Acceleration.cpp
```diff
@@ -40,10 +40,8 @@ void Acceleration::applyRelaxation(double omega, DataMap &cplData, double window
         auto  oldGradients = couplingData.getPreviousGradientsAtTime(stample.timestamp); // IMPORTANT DETAIL: The interpolation that we use for resampling does not necessarily have to be the same interpolation as the interpolation the user accesses via read-data. (But probably it is easier to just use the same)
         gradients          = gradients * omega + oldGradients * (1.0 - omega);
       }
+      couplingData.setSampleAtTime(stample.timestamp, stample.sample); // @todo try to remove, currently needed to update CouplingData::_data::_sample. Should not be needed anymore when https://github.com/precice/precice/issues/1645 is resolved.
```
BenjaminRodenberg, line 43: We could also call this function outside of the loop for the time associated with the last stample. I think it is (a bit) cleaner this way. In any case it is a (hopefully temporary) hack..

fsimonis: Would it make sense to make this explicit with a `setGlobalSample()` until we can get rid of it?

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -138,14 +138,12 @@ void BaseCouplingScheme::sendData(const m2n::PtrM2N &m2n, const DataMap &sendDat
       }
     } else {
       if (data->hasGradient()) {
-        data->sample() = stamples.back().sample;
         // Data is only received on ranks with size>0, which is checked in the derived class implementation
-        m2n->send(data->values(), data->getMeshID(), data->getDimensions());
-        m2n->send(data->gradients(), data->getMeshID(), data->getDimensions() * data->meshDimensions());
+        m2n->send(data->stamples().back().sample.values, data->getMeshID(), data->getDimensions());
+        m2n->send(data->stamples().back().sample.gradients, data->getMeshID(), data->getDimensions() * data->meshDimensions());
```
fsimonis, line 143: Do we know the timestamp of this sample?, then we could avoid `back()`

src/cplscheme/CouplingData.cpp
```diff
@@ -48,24 +48,30 @@ int CouplingData::nVertices() const
...
 {
-  return sample().values;
+  return sample().gradients;
 }
 
-Eigen::MatrixXd &CouplingData::gradients()
+int CouplingData::gradientsRows() const
 {
-  return sample().gradients;
+  return sample().gradients.rows();
```
fsimonis, line 63: At the risk of self-humiliation

```suggestion
  return _mesh->nVertices() * _data->getDimensions();
```

src/cplscheme/CouplingData.cpp
```diff
@@ -48,24 +48,30 @@ int CouplingData::nVertices() const
...
 }
 
-const Eigen::MatrixXd &CouplingData::gradients() const
+int CouplingData::gradientsCols() const
 {
-  return sample().gradients;
+  return sample().gradients.cols();
```
fsimonis, line 68:
```suggestion
  return _mesh->getDimensions();
```

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: `getTime()` should work. It's behavior is a bit complicated since it depends on the context where `sendData/receiveData` is called. But I think it's still a step into the right direction.

I will also add some assertions to make the context-sensitivity more explicit.

src/cplscheme/CouplingData.cpp
BenjaminRodenberg: You took the risk and I think it's the other way around :smirk: 
But good point. And I will also add an assertion.

src/cplscheme/CouplingData.cpp
BenjaminRodenberg: See above.

src/cplscheme/CouplingData.cpp
BenjaminRodenberg: Looks like [f8d320a](https://github.com/precice/precice/pull/2203/commits/f8d320ab33ac5f3903028447e031017131b9d90d) broke some tests (e.g. `precice.CommunicationTests/SerializedStamples/DeserializeValuesAndGradients`).

src/acceleration/Acceleration.cpp
BenjaminRodenberg: Done.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -940,7 +923,11 @@ double BaseCouplingScheme::getWindowStartTime() const
 
 double BaseCouplingScheme::getWindowEndTime() const
 {
-  return getWindowStartTime() + getTimeWindowSize();
+  if (hasTimeWindowSize()) {
+    return _time.time() + _time.untilWindowEnd(_timeWindowSize);
+  } else {
+    return _time.time() + _time.untilEnd();
+  }
```
BenjaminRodenberg, line 930: The old implementation was problematic for, for example, `precice.Integration/Serial/Time/Explicit/SerialCoupling/DoNonfittingWindows`. I think we should factor this change out into a different PR.

src/cplscheme/CouplingData.cpp
BenjaminRodenberg: Using `getSize()` and `getMeshDimensions()` seems to do the trick. See 4f879e84ff80ae32fb677481b7ee9a1bdcad2e86.

src/cplscheme/BaseCouplingScheme.cpp
BenjaminRodenberg: Fixed in #2213

src/acceleration/Acceleration.cpp
```diff
@@ -40,10 +40,8 @@ void Acceleration::applyRelaxation(double omega, DataMap &cplData, double window
         auto  oldGradients = couplingData.getPreviousGradientsAtTime(stample.timestamp); // IMPORTANT DETAIL: The interpolation that we use for resampling does not necessarily have to be the same interpolation as the interpolation the user accesses via read-data. (But probably it is easier to just use the same)
         gradients          = gradients * omega + oldGradients * (1.0 - omega);
       }
+      couplingData.setGlobalSample(stample.sample); // @todo try to remove, currently we have to set CouplingData::_data::_sample to the final stample in the storage. Should not be needed anymore when https://github.com/precice/precice/issues/1645 is resolved.
     }
```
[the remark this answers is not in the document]
fsimonis: Should be fine. Its a temporary:tm: measure anyways.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -137,15 +136,14 @@ void BaseCouplingScheme::sendData(const m2n::PtrM2N &m2n, const DataMap &sendDat
         m2n->send(serialized.gradients(), data->getMeshID(), data->getDimensions() * data->meshDimensions() * serialized.nTimeSteps());
       }
     } else {
+      const auto sample = data->timeStepsStorage().getSampleAtOrAfter(this->getTime()); // this->getTime() returns getWindowStartTime() when exchanging initial data; returns getWindowEndTime() for data exchange at window end.
```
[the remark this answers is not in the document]
BenjaminRodenberg: Any idea how we could implement this instead? I added lots of assertions to ensure that `getTime()` has the correct value when calling `sendData`. Where do you imagine a problem? If we could turn this into a test this would help with the development. Currently, I find it hard to find a better solution.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -896,23 +894,7 @@ void BaseCouplingScheme::doImplicitStep()
...
-        PRECICE_ASSERT(stamples.size() > 0);
-        data->sample() = stamples.back().sample;
-      }
-
       _acceleration->performAcceleration(getAccelerationData(), getTimeWindowStart(), getWindowEndTime());
-
-      // Store from buffer
-      // @todo Currently only data at end of window is accelerated. Remaining data in storage stays as it is.
-      for (auto &data : getAccelerationData() | boost::adaptors::map_values) {
-        data->setSampleAtTime(getTime(), data->sample());
```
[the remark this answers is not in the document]
BenjaminRodenberg: No. We don't need this anymore because the acceleration directly operates on the storage.

src/com/SerializedStamples.hpp
```diff
@@ -85,14 +85,14 @@ class SerializedStamples {
    *
    * @param data the data that is serialized
    */
-  void serializeValues(const cplscheme::PtrCouplingData data);
+  void serializeValues(std::shared_ptr<const cplscheme::CouplingData> data);
```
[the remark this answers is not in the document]
BenjaminRodenberg: Using references would generally be a good idea in `SerializedStamples`. The main reason for the existence of `PtrCouplingData` is that we use it in the `DataMap` (where we cannot use references). I would suggest to keep things as they are here and do a follow up PR to use more references and fewer pointers in `SerializedStamples`. Doing the change here would introduce quite a lot of changes.

BenjaminRodenberg (commented):
I would honestly prefer to merge the PR as it is. If there is really a risk in using `data->timeStepsStorage().getSampleAtOrAfter(this->getTime())` we can fix it here. Right now I do not really see an issue (based on the tests we have).

src/cplscheme/BaseCouplingScheme.cpp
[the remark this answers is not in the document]
fsimonis:
```suggestion
      const auto sample = data->timeStepsStorage().getSampleAtEnd();
```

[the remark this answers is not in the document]
BenjaminRodenberg: Let's see if this explodes. Sometimes the serial coupling scheme + initialization is quite a beast :smirk:

src/time/Storage.cpp
```diff
@@ -242,14 +242,9 @@ int Storage::computeUsedDegree(int requestedDegree, int numberOfAvailableSamples
...
-time::Sample Storage::getSampleAtBeginning()
+Stample Storage::getStampleAtEnd()
 {
-  return _stampleStorage.front().sample;
-}
-
-time::Sample Storage::getSampleAtEnd()
-{
-  return _stampleStorage.back().sample;
+  return _stampleStorage.back();
```
fsimonis, line 247:
```suggestion
  PRECICE_ASSERT(!_stampleStorage.empty());
  return _stampleStorage.back();
```

src/time/Storage.cpp
```diff
@@ -242,14 +242,9 @@ int Storage::computeUsedDegree(int requestedDegree, int numberOfAvailableSamples
   return std::min(requestedDegree, numberOfAvailableSamples - 1);
 }
 
-time::Sample Storage::getSampleAtBeginning()
+Stample Storage::getStampleAtEnd()
```
fsimonis, line 245: To avoid the copies.

```suggestion
Stample& Storage::getStampleAtEnd()
```

BenjaminRodenberg: Done [1bf2cb7](https://github.com/precice/precice/pull/2203/commits/1bf2cb7083375b4ba6f318d81e6f08bd9872a505)

src/time/Storage.cpp
BenjaminRodenberg: Done [36830fa](https://github.com/precice/precice/pull/2203/commits/36830fa4767a76543530769b2c35d351729910da)

src/com/SerializedStamples.hpp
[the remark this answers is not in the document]
BenjaminRodenberg: I already started this change here: https://github.com/precice/precice/pull/2215

fsimonis (approved):
The murky waters are slowly clearing.

PR #2206, opened 2025-02-19
Provide time window end to perform acceleration

Fujikawas:
## Main changes of this PR
As the title

To synchronize with QN in Aitken, we want to get the value at a clear time point than getting the latest data by default while concatenating data. For Aitken, it's the end time of the time window.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/acceleration/AitkenAcceleration.cpp
+++ b/src/acceleration/AitkenAcceleration.cpp
@@ -125,14 +125,14 @@ void AitkenAcceleration::concatenateCouplingData(
   for (auto id : dataIDs) {
     Eigen::Index size = cplData.at(id)->values().size();
 
-    Eigen::VectorXd values    = cplData.at(id)->timeStepsStorage().sample(windowEnd).values();
-    Eigen::VectorXd oldValues = cplData.at(id)->getPreviousValuesAtTime(windowEnd).values();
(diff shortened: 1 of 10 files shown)


src/acceleration/AitkenAcceleration.cpp
```diff
@@ -117,13 +118,16 @@ void AitkenAcceleration::iterationsConverged(
...
 {
   Eigen::Index offset = 0;
+
   for (auto id : dataIDs) {
-    Eigen::Index size      = cplData.at(id)->values().size();
-    auto &       values    = cplData.at(id)->values();
-    const auto & oldValues = cplData.at(id)->previousIteration();
+    Eigen::Index size = cplData.at(id)->values().size();
+
+    Eigen::VectorXd values    = cplData.at(id)->timeStepsStorage().sample(windowEnd);
```
BenjaminRodenberg, line 128:
```suggestion
    Eigen::VectorXd values    = cplData.at(id)->timeStepsStorage().sample(windowEnd).values();
```
This was changed very recently via #2190

fsimonis: Actually, prefer to save the result of sample and then use `result.values()` in the computation. This can avoid copies.

fsimonis (approved):
Nice small and clean refactoring!

PR #2210, opened 2025-02-22
Decouple couplingData dims from samples

fsimonis:
## Main changes of this PR

This PR decouples the CouplingData dimension queries from the internal sample and uses mesh and data instead.

This is mostly a TODO note as this change should be trivial, but triggers errors in tests that use meshes but don't define vertices.

## Author's checklist

### DIFF
--- a/src/cplscheme/CouplingData.cpp
+++ b/src/cplscheme/CouplingData.cpp
@@ -36,16 +36,12 @@ int CouplingData::getDimensions() const
 
 int CouplingData::getSize() const
 {
-  // @todo this correct implementation breaks a ton of tests that don't define vertices of a test mesh
-  //return _mesh->nVertices() * getDimensions();
-  return sample().values.size();
+  return _mesh->nVertices() * getDimensions();
 }
 
 int CouplingData::nVertices() const
 {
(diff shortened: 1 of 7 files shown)


src/cplscheme/CouplingData.cpp
```diff
@@ -36,16 +36,12 @@ int CouplingData::getDimensions() const
 
 int CouplingData::getSize() const
 {
-  // @todo this correct implementation breaks a ton of tests that don't define vertices of a test mesh
-  //return _mesh->nVertices() * getDimensions();
-  return sample().values.size();
+  return _mesh->nVertices() * getDimensions();
```
BenjaminRodenberg, line 39:
```suggestion
  const int size = _mesh->nVertices() * getDimensions();
  PRECICE_ASSERT(sample().values.size() ==  size);
  return size;
```
Similar in the change below. Generally all this should be consistent, no?

fsimonis: When is the `sample()` created though?

fsimonis: Hmm. True. We should keep this until we remove the sample member.

fsimonis: Actually, `sample()` is a better place for this check.

PR #2217, opened 2025-02-27
Check consistent configs accross participants

### LINKED ISSUE #899: preCICE allows mixing different configuration files
It is possible to mix `implicit` and `explicit` couplings. 

**Describe your setup**

preCICE Version: Commit cdf38703e0ad7debbd76c6785fdc0d3d9d3d2dfc from this PR: https://github.com/precice/precice/pull/84 as debug build

preCICE is used via Python using the bindings v2.1.1.1

**Describe the problem**

One can mix `implicit` and `explicit` couplings. The solver using the `explicit` coupling will interpret the subiterations of the `implicit` solver as time windows and participate. The simulation will eventually hang

**Step To Reproduce**
1. Choose a `serial-implicit` coupling configuration of your solver
1. Choose a `serial-explicit` coupling configuration of your solver
1. Run the simulation
1. The `serial-implicit` solver will start the subiterations while the `serial-explicit` solver will interpret each subiteration as time windows.

I have only testet this with `serial` coupings.

**Expected behaviour**

I would expect an error message and termination of the solvers.

fsimonis:
## Main changes of this PR

This PR changes the ConfigParser to create a hash of the configuration file content and adds a cross-checking step when establishing primary connections between Participants.
The filepaths are not checked to allow for creative solutions to running in a containerized or distributed setting.

It reuses the same portable hashing we already use for the creation of communication files in `precice-run`.

Fixes #899

This problem pops up from time to time and even [our veterans](https://precice.discourse.group/t/adapter-advance-with-multiple-interfaces/2330/4) like waste time with this.

This is an extension of #2202

## Author's checklist

### DIFF
--- a/docs/changelog/2217.md
+++ b/docs/changelog/2217.md
@@ -1,0 +1,1 @@
+- Added a check to ensure coupled participants use configuration files with the same content.
--- a/src/cplscheme/tests/CompositionalCouplingSchemeTest.cpp
+++ b/src/cplscheme/tests/CompositionalCouplingSchemeTest.cpp
@@ -218,10 +218,10 @@ struct CompositionalCouplingSchemeFixture : m2n::WhiteboxAccessor {
     BOOST_TEST(not communication->isConnected());
     useOnlyPrimaryCom(communication) = true;
     if (participant0 == localParticipant) {
-      communication->requestPrimaryRankConnection(participant1, participant0);
+      communication->requestPrimaryRankConnection(participant1, participant0, "");
     } else {
       BOOST_TEST(participant1 == localParticipant);
-      communication->acceptPrimaryRankConnection(participant1, participant0);
+      communication->acceptPrimaryRankConnection(participant1, participant0, "");
     }
   }
 };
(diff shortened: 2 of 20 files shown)


MakisH:
This is the patch we apply before running system tests:

https://github.com/precice/tutorials/blob/962793d5b173d828e2309af017076d8965fcfe79/tools/tests/docker-compose.template.yaml#L15-L17

fsimonis:
Debatable is if this should be an error or a warning.

There may be some local patching required in the m2n section.

uekerman (approved):
That's a nice check, thanks for adding this!

My minor tendency would be "warning", allowing us to hack this ourselves if we ever need. But then, the nice non-happy-path integration test does not work. I am happy with either choice.

fsimonis:
An error will be the more helpful for the vast majority of people as a warning can still be overlooked.

Also, making this an error will trigger users that need this feature to reporting it.
Meaning, we can gather use-case and come up with non-hacky solutions for the user.

@carme-hp commented that a person that needs this kind of special tuning will not have any problems building their own preCICE.

PR #2228, opened 2025-03-17
Simplify write buffer resize

fsimonis:
## Main changes of this PR

This PR uses conservativeResize of Eigen to simplify the write buffer resize.

Less custom code to maintain

## Author's checklist

### DIFF
--- a/src/precice/impl/WriteDataContext.cpp
+++ b/src/precice/impl/WriteDataContext.cpp
@@ -1,7 +1,5 @@
 #include "WriteDataContext.hpp"
 
-#include "utils/EigenHelperFunctions.hpp"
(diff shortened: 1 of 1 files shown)


BenjaminRodenberg:
If you are already at it: Something very similar that would also profit from a refactoring is

https://github.com/precice/precice/blob/14363ca5b2e86812e706cb7ebd37e895f879451f/src/mesh/Data.cpp#L135-L171

fsimonis:
@BenjaminRodenberg Shouldn't `allocateData` be removed instead?

fsimonis:
Waiting for #2230 before merging this.

PR #2229, opened 2025-03-18
Add nightly system tests badge in README.md

MakisH:
## Main changes of this PR

Adds a dynamic badge to the system tests workflow:

[System tests (latest components)](https://github.com/precice/tutorials/actions/workflows/system-tests-latest-components.yml)

Clicking on the badge gives you also a manual workflow trigger, with no inputs (just takes the latest components). The fully customizable manual workflow is [this](https://github.com/precice/tutorials/actions/workflows/run_testsuite_manual.yml), which I will rename to be more easily discoverable.

The workflow should run nightly at 4am. It does not modify the commit build status and it is not connected to this repository.

Rendered as:

Screenshot from 2025-03-18 17-45-15

We need a way to easily check the status of these tests, and notifications are a bit tricky to set up (we need either an email account or integration with Matrix).

At the same time, we don't want them to be too intrusive and have failing system tests give the wrong impression for the library itself. Therefore, I added a category "Downstream Components" and moved the related badges there (system tests, distribution, website). I did not use "Ecosystem", as this might not be clear to first-time readers.

Also adds a changelog entry, which was missing in #2052.

## Author's checklist

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -1,13 +1,10 @@
 # preCICE #
 
 **Project Status**  
-[![preCICE website status](https://img.shields.io/website-up-down-green-red/https/precice.org.svg?label=website)](https://precice.org/)
+[![Build status](https://github.com/precice/precice/workflows/Build%20and%20Test/badge.svg)](https://github.com/precice/precice/actions?query=workflow%3A%22Build+and+Test%22+branch%3Adevelop)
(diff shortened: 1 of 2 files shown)


MakisH:
Scheduled run worked: https://github.com/precice/tutorials/actions/runs/13938867216

BenjaminRodenberg (approved):
Looks good :+1:

PR #2234, opened 2025-03-20
Invalidate write data buffer after repartitioning

davidscn:
## Main changes of this PR

invalidate the write buffer after repartitioning it.

The write buffer is closely connected to the mesh. If we change the mesh, the currently buffered data automatically becomes invalid. There is no connection between the currently stored data and the new layout.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/impl/ParticipantImpl.cpp
+++ b/src/precice/impl/ParticipantImpl.cpp
@@ -1665,10 +1665,11 @@ void ParticipantImpl::computePartitions()
 
     meshContext->mesh->allocateDataValues();
 
+    // Should be relevant for direct mesh access only
     const auto requiredSize = meshContext->mesh->nVertices();
     for (auto &context : _accessor->writeDataContexts()) {
       if (context.getMeshName() == meshContext->mesh->getName()) {
-        context.resizeBufferTo(requiredSize);
+        context.resizeBufferTo(requiredSize, true);
       }
     }
   }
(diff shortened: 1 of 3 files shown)


fsimonis:
What about keeping a `bool _valid;` flag, which tracks the invalidation separately? This way you can raise an error or throw an assertion if there is any attempt at writing data directly in `writeValuesIntoDataBuffer()`.

davidscn:
> What about keeping a bool _valid; flag, which tracks the invalidation separately? This way you can raise an error or throw an assertion if there is any attempt at writing data directly in writeValuesIntoDataBuffer().

This would be more readable, indeed. I would consider it an addition. The resizing in the marked impl touches all WriteDataContexts, but for all provided meshes, this is a noop I guess.

fsimonis:
I always find separate calls easier to read than boolean arguments.

```cpp
 if (context.getMeshName() == meshContext->mesh->getName()) {
        context.resizeBufferTo(requiredSize);
        if (!meshContext->provideMesh) {
          context.invalidate();
        }
      }
```

```cpp
 if (context.getMeshName() == meshContext->mesh->getName()) {
        context.resizeBufferTo(requiredSize, !meshContext->provideMesh);
      }
```

davidscn:
I don't really like these kind of nestings. Makes it in my opinion more difficult to read. Seems to be a matter of preference then.

PR #2236, opened 2025-03-20
Prevent write data with api access and remeshing

davidscn:
## Main changes of this PR

Prevent write data with api access and remeshing

See https://github.com/precice/precice/issues/2094

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/impl/ParticipantImpl.cpp
+++ b/src/precice/impl/ParticipantImpl.cpp
@@ -257,10 +257,6 @@ void ParticipantImpl::initialize()
                 "Initial data has to be written to preCICE before calling initialize(). "
                 "After defining your mesh, call requiresInitialData() to check if the participant is required to write initial data using the writeData() function.");
 
-  for (auto &context : _accessor->writeDataContexts()) {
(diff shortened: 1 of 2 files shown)


src/precice/impl/ParticipantImpl.cpp
```diff
@@ -268,6 +268,10 @@ void ParticipantImpl::initialize()
 
   _meshLock.lockAll();
 
+  for (auto &context : _accessor->writeDataContexts()) {
+    PRECICE_CHECK(_accessor->meshContext(context.getMeshName()).provideMesh || !(_accessor->isDirectAccessAllowed(context.getMeshName()) && _allowsRemeshing), "Writing data via API access (configuration <write-data ... mesh=\"{}\") is not (yet) supported with remeshing", context.getMeshName());
+  }
+
```
fsimonis, line 274: This check belongs into the `ParticipantConfiguration`, most likely into `finishParticipantConfiguration`. You can use `bool _remeshing` to check if remeshing is enabled.

davidscn: What's the problem here?

fsimonis: What can be checked in the configuration should be checked there to raise an error as early as possible.
This also makes this checkable by `precice-tools check`.

PR #2241, opened 2025-03-24
Rename develop preset to development

fsimonis:
## Main changes of this PR

This PR renames the `develop` CMake preset to `development`.

Currently, the preset is called the same as the develop branch.
It has caused some confusion with the term "develop version of preCICE", where a user build a release with the develop preset instead of building the develop branch.

This could break some CIs.

~User-documentation needs to be updated.~ User-documentation doesn't mention the develop preset.

## Author's checklist

### DIFF
--- a/CMakePresets.json
+++ b/CMakePresets.json
@@ -30,7 +30,7 @@
       }
     },
     {
-      "name": "develop",
+      "name": "development",
       "inherits": "base",
       "displayName": "Development",
       "description": "Development config with default settings",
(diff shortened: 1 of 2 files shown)


fsimonis:
> Do we regularly tell users to run cmake .. --preset develop or how do they come across these labels?

They are mentioned in https://precice.org/installation-source-configuration.html#using-presets and are shown in tools like VS Code/CLion or CMake GUI.

PR #2242, opened 2025-03-24
Fix type of param in `writeAndMapData` doc

IshaanDesai:
## Main changes of this PR

The parameter `values` in the API function `writeAndMapData` should be an input parameter.

## Author's checklist

## Reviewers' checklist


davidscn (approved):
Thanks!

PR #2244, opened 2025-03-25
Update pre-commit versions

davidscn:
## Main changes of this PR

Updates the versions and applies a few changes triggered.

## Author's checklist

## Reviewers' checklist


MakisH (approved):
This updates the black, precice-pre-commit-hooks, and codespell hooks. Does not update clang-format.

Nice that it caught a few spelling mistakes as well.

I stumbled for a moment on whether it would be a good idea to fix things in past changelog entries, but I don't see any good reason not to. The file itself is anyway constantly updated, and the fix simplifies our checks.

PR #2247, opened 2025-03-25
System test updates

MakisH:
## Main changes of this PR

Collecting some necessary updates to the system tests workflow trigger, as I am now polishing a few aspects that I encounter.

1. Renames two variables, see https://github.com/precice/tutorials/pull/636
2. Extends the job summary for the preparatory step, see https://github.com/precice/tutorials/pull/635

Since this is the first release that uses this workflow, I would like to have the workflow up-to-date before merging.

## Author's checklist


BenjaminRodenberg (approved):
Looks good :+1

PR #2250, opened 2025-03-25
Clang format version bump

fsimonis:
## Main changes of this PR

This PR updates clang-format to v20.1.0 and applies it to the codebase.

The format wasn't changed as this results in massive changes.

## Author's checklist


davidscn (changes requested):
We have (ancient) formatting scripts in `tools/formatting`, which we should either update or remove. Would also be worth a changelog

fsimonis:
Good point. I'll remove them, but in a separate commit.

Merging this behemoth now.

PR #2251, opened 2025-03-26
Track mesh size in profiling record

davidscn:
## Main changes of this PR

...which could be very helpful to interpret the performance data.

## Author's checklist

## Reviewers' checklist


fsimonis:
Could you add the same to `reinitialize()` to the event `Event e("reinitialize", profiling::Fundamental);`?

fsimonis (approved):
I marked this as an enhancement as this information allows developer and user to have a significantly better idea about load-balance, and expected mapping and communication costs.
All by looking at the default profiling information.

PR #2261, opened 2025-04-01
Update the release PR workflow

fsimonis:
## Main changes of this PR

This PR updates the release PR template with some lessons learned

I also wrote out the full git commands of many steps and what exactly to check.

## Author's checklist


IshaanDesai (commented):
Perhaps adding a point about updating the tested preCICE version in the [setup-precice-action](https://github.com/precice/setup-precice-action) would make sense.

fsimonis:
> Perhaps adding a point about updating the tested preCICE version in the [setup-precice-action](https://github.com/precice/setup-precice-action) would make sense.

I think this may not be necessary. This tests installing a packaged version from a git tag, which should be independent of the version itself.
The bump is only necessary if that version doesn't support `ubuntu-latest`.

PR #2272, opened 2025-04-07
Add option to keep export series updated

### LINKED ISSUE #2264: Option to generate exported series file at end of every time window instead of just at end?
The new series file capability implemented by #2055 is extremely useful for debugging quickly and just general visualization in ParaView. Unfortunately when simulations fail, this series file is not generated as it is only generated at the **end** of a coupled simulation. For debugging purposes, this would be helpful as test cases sometimes blow up in the process of implementing and verifying a precice adapter. For running large simulations at scale, job time limits on HPC clusters often mean that simulations do not run to end.

It would be a really nice feature if it's not too hard to implement to have the series file written/updated every specified number of time windows instead of just at the end.

Either by default have the series file generated + updated every export, or have an additional option that enables this.

None.

**Additional context**
None.

fsimonis:
## Main changes of this PR

This PR adds a configuration option `update-series` to exporter tags, which controls weather the series file of an exporter is updated after every export or only at the of the simulation.

The setting defaults to `false` to prevent repeatedly overwriting the same file, which could lead to issues with distributed file systems and result in extensive wear when running for a large amount of timesteps.

Closes #2264

## Author's checklist


MakisH (commented):
Toggling the option seems to have the expected effect, and it seems to resolve the original issue.

I checked in the elastic-tube-1d tutorial with:

```
tail -F precice-exports/Fluid-Nodes-Mesh-Solid.vtu.series
```

A test is needed, still.

fsimonis:
Added tests for both series that are created only at the end of the simulation and series updated after every time window end.

MakisH (approved):
The tests also make sense.

PR #2266, opened 2025-04-04
Fix compilation on MinGW

fsimonis:
## Main changes of this PR

This PR attempts to fix export issue on mingw

## Motivation and additional information

Based on #2265
Closes #2262
Related to #200 

## Author's checklist

* [x] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [ ] I added a test to cover the proposed changes in our test suite.
* [ ] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).
* [ ] I stuck to C++17 features.
* [ ] I stuck to CMake version 3.22.1.
* [x] I squashed / am about to squash all commits that should be seen as one.
* [ ] I ran the systemtests by adding the label `trigger-system-tests` (may be skipped if minor change)

PR #2273, opened 2025-04-08
Update build_win_mingw.yml

3rav:
Successful


fsimonis:
I fixed the CMakeLists to allow for correct compilation and linkage in #2266.
The next step is to get the tests running.

Thanks for the suggestion though.

PR #2300, opened 2025-05-19
Hardcode preCICE as shared library

### LINKED ISSUE #2299: Only support preCICE as a shared library

Supporting preCICE as a static library has multiple problems:
* support via CMake requires exporting the entire link graph of preCICE, which is tricky due to our use of internal object libraries. 
* some dependencies such as PETSc are not integrated into CMake. Using them via pkg-config makes correct dependency propagation difficult.
* exposes all symbols which allows easier access to using internals in user-code
* requires even more CI builds as shared libs use additional flags.

In addition to this, using preCICE with `BUILD_SHARED_LIBS=OFF` currently leads to CMake errors.

Drop support for static libraries, set the preCICE library type to `SHARED`, thus always building as shared.

Invest time to make preCICE buildable as static library.

**Additional context**

#973
#2262  
#2119

fsimonis:
## Main changes of this PR

This PR changes the preCICE CMake to always generate preCICE as a SHARED library.

Closes #2299

## Author's checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -53,12 +53,15 @@ if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
   set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS Debug Release RelWithDebInfo MinSizeRel)
 endif()
 
+if (DEFINED BUILD_SHARED_LIBS AND NOT BUILD_SHARED_LIBS)
+  message(WARNING "You disabled BUILD_SHARED_LIBS, but preCICE unconditionally builds as shared library.")
+endif()
+
 option(PRECICE_FEATURE_MPI_COMMUNICATION "Enables MPI-based communication and running coupling tests." ON)
 option(PRECICE_FEATURE_PETSC_MAPPING "Enable use of the PETSc linear algebra library for global-iterative RBF data mappings." ON)
 option(PRECICE_FEATURE_PYTHON_ACTIONS "Enable Python support for preCICE actions." ON)
 option(PRECICE_CONFIGURE_PACKAGE_GENERATION "Configure package generation." ON)
 option(PRECICE_FEATURE_GINKGO_MAPPING "Enable use of Ginkgo for accelerated data mapping." OFF)
-option(BUILD_SHARED_LIBS "Build shared instead of static libraries" ON)
 option(BUILD_TESTING "Build tests" ON)
 option(PRECICE_ALWAYS_VALIDATE_LIBS "Validate libraries even after the validatation succeeded." OFF)
 option(PRECICE_BINDINGS_C "Enable the native C bindings" ON)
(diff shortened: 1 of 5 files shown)


docs/changelog/2300.md
```diff
@@ -0,0 +1 @@
+- Removed support for configuring preCICE as static library.
```
MakisH, line 1: I think we should add a comment that static was anyway not working before, thus we bring this in a minor release.

cmake/PrintHelper.cmake
```diff
@@ -59,7 +59,6 @@ function(print_configuration)
   endif()
 
   print_variables( VARS
-    "BUILD_SHARED_LIBS;Build shared libraries"
```
MakisH, line 62: Is there an easy way to inform people that still use `-DBUILD_SHARED_LIBS=ON` that this is not needed anymore?

CMakeLists.txt
```diff
@@ -53,12 +53,15 @@ if(NOT CMAKE_CONFIGURATION_TYPES AND NOT CMAKE_BUILD_TYPE)
   set_property(CACHE CMAKE_BUILD_TYPE PROPERTY STRINGS Debug Release RelWithDebInfo MinSizeRel)
 endif()
 
+if (DEFINED BUILD_SHARED_LIBS AND NOT BUILD_SHARED_LIBS)
+  message(WARNING "You disabled BUILD_SHARED_LIBS, but preCICE unconditionally builds as shared library.")
```
MakisH, line 57:
```suggestion
  message(WARNING "You disabled BUILD_SHARED_LIBS, but preCICE can only being built as a shared library. This will still build preCICE as a shared library.")
```

MakisH (approved):

docs/changelog/2300.md
fsimonis: Static libs are broken since we use the object library for preciceCore, so 3.0.0.

```suggestion
- Removed support for configuring preCICE as static library, which has been broken since version 3.0.0.
```

cmake/PrintHelper.cmake
fsimonis: The option defaulted to building shared since bb83edc7a638947ccb476320dcdb8473c9098e2e or release 2.3.0 of preCICE.

There is no reason to explicitly set this ever since. I think this shouldn't really be a problem.

CMakeLists.txt
fsimonis: > can only being built as a shared library.

This implies that the user is doing something wrong. In the case of a subproject, this is completely fine.

Slightly more verbose version:

```suggestion
  message(WARNING "You disabled BUILD_SHARED_LIBS. Note that 3.2.0 was the last release to honor this option. Newer preCICE releases always build as a shared library.")
```

cmake/PrintHelper.cmake
MakisH: Not a problem, I was just thinking of an effective way to inform people to clean up their scripts. It's already ok as-is.

MakisH (approved):

PR #2301, opened 2025-05-20
Add coarse graining as just-in-mapping

davidscn:
## Main changes of this PR

Adds a coarse graining as described in [Weinhart et al.](https://doi.org/10.1063/1.4812809) as just-in-time mapping variant.

Coarse graining is particularly interesting for particle-mesh coupling to extract the continuum field from the particle data. I am not yet entirely sure whether the implementation should actually live in preCICE or rather in an adapter (via direct mesh access). The current approach is specific to spherical particles.

Coarse graining exhibits important conservation properties, since data mapping is performed using a basis function $\mathcal{W}(x)$ with the property

 $$\int_{\mathbb{R}^d}\mathcal{W}(x)dV = 1$$

such that spatial integration of the mapped variable recovers (theoretically) exact conservation. Any violation of conservative properties is due to numerical approximation. While the input variable is an extensive quantity (typically called _conservative_ in preCICE, e.g., volume or forces), the output variable is a scaled quantity (typically called _consistent_ in preCICE, e.g., solid-fraction, porosity or force density). Thus, the clear "constraint" definition we have in preCICE doesn't really fit. Luckily, the parallelization scheme still follows the one used for conservative mappings, such that we could offer the `write-conservative` (+just-in-time) mapping variant we need for the particle -> mesh coupling without adding another layer of complexity. 

The (very basic) implementation right now only supports just-in-time mappings in `write` direction using a `conservative` constraint and is mainly meant for experimenting with the method as such.

Using coarse-graining for the data mapping requires a proper treatment of boundary conditions. With the smoothing length $w$, particles which are $w = r/2$ close to the boundaries have to be mirrored and passed to preCICE as well. As an alternative approach to the mirroring, one could also enforce the conservation in a discrete manner, which then omits the mirroring. However, this would require access to quadrature rules of the receiving continuum solver, which contradicts somewhat the black-box coupling approach. We could, of course, transfer those quantities via preCICE as well and then use them via direct-mesh access.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/tests/serial/just-in-time-mapping/ExplicitWriteCG.cpp
+++ b/tests/serial/just-in-time-mapping/ExplicitWriteCG.cpp
@@ -1,0 +1,151 @@
+#ifndef PRECICE_NO_MPI
+
+#include "testing/Testing.hpp"
+
+#include <numeric>
+#include <precice/precice.hpp>
+#include <vector>
+
+BOOST_AUTO_TEST_SUITE(Integration)
+BOOST_AUTO_TEST_SUITE(Serial)
+BOOST_AUTO_TEST_SUITE(JustInTimeMapping)
+PRECICE_TEST_SETUP("SolverOne"_on(1_rank), "SolverTwo"_on(1_rank))
+
+// Test case for a just-in-time mapping on one participant to a mesh defined
+// by another participant. The region of interest is defined through a
+// boundingBox. The test case here is the most basic variant in order
+// use such a feature. SolverTwo defines the mesh whereas SolverOne writes
+// just-in-time to this mesh.
+// cg-conservative-write
+
+BOOST_AUTO_TEST_CASE(ExplicitWriteCG)
+{
+  PRECICE_TEST();
+
+  // Set up Participant
+  precice::Participant couplingInterface(context.name, context.config(), 0, 1);
+
+  constexpr int               dim         = 3;
+  std::array<double, dim * 2> boundingBox = {0.0, 1.0, 0.0, 1.0, 0.0, 1.0};
+
+  if (context.isNamed("SolverOne")) {
+    auto otherMeshName  = "MeshTwo";
+    auto vectorDataName = "Velocity";
+    BOOST_REQUIRE(couplingInterface.getMeshDimensions(otherMeshName) == 3);
+
+    int                 meshSize = 30;
+    std::vector<int>    ids(meshSize);
+    std::vector<double> positions;
+    positions.reserve(dim * meshSize);
+
+    // Adding 30 test positions
+    double offset = 0.05;         // small offset to ensure no overlap with grid points
+    for (int k = 0; k < 5; ++k) { // Three loops, but only 5 iterations each to add 10 new positions
+      for (int l = 0; l < 2; ++l) {
+        for (int m = 0; m < 3; ++m) {
+          positions.emplace_back(k * 0.2 + offset);  // Increment x by 0.2 each time, starting from 0.1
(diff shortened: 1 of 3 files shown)


src/mapping/config/MappingConfiguration.hpp
```diff
@@ -163,6 +165,10 @@ class MappingConfiguration : public xml::XMLTag::Listener {
   // For iterative RBFs
   const std::string ATTR_SOLVER_RTOL = "solver-rtol";
 
+  // For coarse graining
+  const std::string ATTR_GRAIN_DIMENSION = "grain-dim";
```
uekerman: What is the dimension of the grain exactly? How is it related to the mesh dimension?

src/mapping/config/MappingConfiguration.hpp
```diff
@@ -163,6 +165,10 @@ class MappingConfiguration : public xml::XMLTag::Listener {
...
+  const std::string ATTR_FUNCTION_RADIUS = "function-radius";
```
uekerman:
```suggestion
  const std::string ATTR_FUNCTION_RADIUS = "radius";
```
Clear like this as well?

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -226,6 +229,11 @@ MappingConfiguration::MappingConfiguration(
                                .setDocumentation("Toggles use a local (per cluster) polynomial")
                                .setOptions({POLYNOMIAL_OFF, POLYNOMIAL_SEPARATE});
 
+  auto attrGrainDimension = makeXMLAttribute<int>(ATTR_GRAIN_DIMENSION, 0)
+                                .setDocumentation("Dimension to use in the coarse-graining function.");
+  auto attrFunctionRadius = makeXMLAttribute<double>(ATTR_FUNCTION_RADIUS, 0.)
+                                .setDocumentation("Radius or range of the coarsening function (Lucy function).");
```
uekerman: Needs probably a bit more docs.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -254,6 +262,7 @@ MappingConfiguration::MappingConfiguration(
   addAttributes(rbfIterativeTags, {attrFromMesh, attrToMesh, attrDirection, attrConstraint, attrPolynomial, attrXDead, attrYDead, attrZDead, attrSolverRtol});
   addAttributes(pumDirectTags, {attrFromMesh, attrToMesh, attrDirection, attrConstraint, attrPumPolynomial, verticesPerCluster, relativeOverlap, projectToInput});
   addAttributes(rbfAliasTag, {attrFromMesh, attrToMesh, attrDirection, attrConstraint, attrXDead, attrYDead, attrZDead});
+  addAttributes(coarseGrainingTags, {attrFromMesh, attrToMesh, attrDirection, attrConstraint, attrGrainDimension, attrFunctionRadius});
```
uekerman: maybe we could directly remove `attrConstraint` here?

davidscn: Unfortunately not, because this also has some implications for internal data processing.

src/mapping/config/MappingConfiguration.hpp
davidscn: I removed this parameter. It is meant to configure the dimension of the particles. In practice, however, only 3D particles are relevant. Any other dimension is not physical.

src/mapping/config/MappingConfiguration.cpp
davidscn: I removed the first attribute, for the second one, I don't see what else to document.

Issue #2307, opened 2025-05-29
Segmentation fault in `precicec_writeAndMapData()`


r-walloner:
**Describe your setup**

Operating system (e.g. Linux distribution and version): _Fedora 42 (Workstation Edition)_
preCICE Version: 3.2.0

**Describe the problem**
Calling `precicec_writeAndMapData()` like this causes a segmentation fault.
```C
double *x = (double*)malloc(3 * sizeof(double));
double *values = (double*)malloc(3 * sizeof(double));
x[0] = x[1] = x[2] = 0.0;
precicec_mapAndReadData("Fluid-Mesh", "Velocity", 1, x, update->dt, values); // works
precicec_writeAndMapData("Fluid-Mesh", "DragForce", 1, x, values); // segfault here
free(x);
free(values);
```

So I built and ran preCICE in debug mode and figured out that the two first asserts in `WriteDataContext::writeAndMapValues()` fail because both  `mappingCache` and `justInTimeMapping` store nullptrs.
```C++
void WriteDataContext::writeAndMapValues(::precice::span<const double> coordinates, ::precice::span<const double> values)
{
  PRECICE_TRACE();
  PRECICE_ASSERT(mappingCache); // this fails
  PRECICE_ASSERT(justInTimeMapping); // this fails too, if we ignore the first assert above

  ...  

  justInTimeMapping->mapConservativeAt(coords, inputData, *mappingCache, localData);
}
```

When ignoring these asserts by uncommenting them, the segfault reappears while calling `mapConservativeAt`. We are trying to call `mapConservativeAt()` on a nullptr.
```C++
void WriteDataContext::writeAndMapValues(::precice::span<const double> coordinates, ::precice::span<const double> values)
{
  PRECICE_TRACE();
  // PRECICE_ASSERT(mappingCache);
  // PRECICE_ASSERT(justInTimeMapping);

  ...  

  justInTimeMapping->mapConservativeAt(coords, inputData, *mappingCache, localData); // segfault here
}
```

**Step To Reproduce**
Does not need any special setup to reproduce.

**Expected behaviour**
Write and map the data as described in the docs for `precicec_writeAndMapData()`.

**Additional context**
```
[FWRW:67018:0:67018] Caught signal 11 (Segmentation fault: address not mapped to object at address (nil))
==== backtrace (tid:  67018) ====
 0  /lib64/libucs.so.0(ucs_handle_error+0x2e4) [0x7fffef85adf4]
[...]

PR #2308, opened 2025-05-29
Add error message in writeAndMapValues in case of misconfiguration

r-walloner:
## Main changes of this PR
Adds helpful user facing warning to address #2307.

Without this message, this type is misconfiguration is somewhat hard to diagnose for the user.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/impl/WriteDataContext.cpp
+++ b/src/precice/impl/WriteDataContext.cpp
@@ -42,12 +42,12 @@ void WriteDataContext::completeJustInTimeMapping()
 void WriteDataContext::writeAndMapValues(::precice::span<const double> coordinates, ::precice::span<const double> values)
 {
   PRECICE_TRACE();
-  PRECICE_ASSERT(mappingCache);
   PRECICE_CHECK(justInTimeMapping,
                 "This participant attempted to write data to mesh \"{}\" using a just-in-time mapping, "
                 "but there is no write mapping configured for that mesh. "
                 "Perhaps you forgot to define a <mapping:... direction=\"write\" /> or want to use direct access with the API function \"writeData({0},...)\".",
                 getMeshName(), getDataName());
+  PRECICE_ASSERT(mappingCache);
   PRECICE_ASSERT((coordinates.size() / getSpatialDimensions()) * getDataDimensions() == values.size());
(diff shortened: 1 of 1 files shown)


src/precice/impl/WriteDataContext.cpp
```diff
@@ -43,7 +43,11 @@ void WriteDataContext::writeAndMapValues(::precice::span<const double> coordinat
 {
   PRECICE_TRACE();
   PRECICE_ASSERT(mappingCache);
-  PRECICE_ASSERT(justInTimeMapping);
+  PRECICE_CHECK(justInTimeMapping,
```
davidscn, line 45: As already hinted in your comment, the assert for the cache should then go after the `justInTimeMapping` check, otherwise, the check and its text will be shadowed by the assert above.

src/precice/impl/WriteDataContext.cpp
```diff
@@ -43,7 +43,11 @@ void WriteDataContext::writeAndMapValues(::precice::span<const double> coordinat
...
+                "This participant attempted to write data to mesh \"{}\", "
```
davidscn:
```suggestion
                "This participant attempted to write data to mesh \"{}\" using a just-in-time mapping, "
```

src/precice/impl/WriteDataContext.cpp
```diff
@@ -43,7 +43,11 @@ void WriteDataContext::writeAndMapValues(::precice::span<const double> coordinat
...
+                "but there is no write mapping configured for that mesh. "
+                "Perhaps you forgot to define a <mapping:... direction=\"write\" />.",
```
davidscn:
```suggestion
                "Perhaps you forgot to define a <mapping:... direction=\"write\" /> or want to use direct access with the API function \"writeData({0},...)\".",
```

davidscn (approved):
Thanks @r-walloner! Unfortunately, we have to check these kind of issues within the 'writeAndMap' data function since the configuration syntax for just-in-time mapping and direct mesh access really only differ by the additional mapping configuration. Although it would be nice to catch this during the configuration stage, it might be legal to not define a mapping (e.g. when using direct access).

Issue #2303, opened 2025-05-20
testprecice fails to link with Intel 2023.1.0 and Intel OpenMP


VictorEijkhout:
```
CMakeFiles/testprecice.dir/src/acceleration/test/ParallelMatrixOperationsTest.cpp.o: undefined reference to symb\
ol 'omp_get_max_threads@@VERSION'
/opt/apps/gcc/12.2.0/lib/gcc/x86_64-pc-linux-gnu/12.2.0/../../../../x86_64-pc-linux-gnu/bin/ld: /opt/intel/oneapi/compiler/2023.1.0/linux/compiler/lib/intel64_lin/libiomp5.so: error adding symbols: DS\
O missing from command line
icpx: error: linker command failed with exit code 1 (use -v to see invocation)
```

Any idea what's missing here? I've never seen the `@@` stuff in an OpenMP function reference.

Btw, this happens when linking the `testprecice` executable, so if you can give me a flag to disable testing I'll be happy enough.

fsimonis:
Hi,

Looks like some Intel openmp linker problem. I haven't come across this version thing either.

You can use `cmake -DBUILD_TESTING=OFF .` to disable the tests.

fsimonis:
@VictorEijkhout Could you let us know how you configured preCICE? Maybe even the complete `CMakeCache.txt` of the build could be useful here.

VictorEijkhout:
[CMakeCache.txt](https://github.com/user-attachments/files/20401235/CMakeCache.txt)

fsimonis:
I see that you don't use the Ginkgo mappings, meaning that the only dependency that uses OpenMP is PETSc. The CMakeCache shows that it forwards the `-qopenmp` flag to preCICE. 

Could you try rebuilding without support for PETSc mappings?
```
cmake -DPRECICE_FEATURE_PETSC_MAPPING=OFF .
```

davidscn:
I observe a similar problem, also with the Intel toolchain (but 2024.0.0). Compiling preCICE works fine but linking fails for the `testprecice` executable. It seems that not the actual linking is the problem, but rather the generation of the test list, so the underlying issue might be different though

```bash
[ 60%] Linking CXX executable testprecice
Build succeeded.
Generating list of tests
CMake Error at ~/precice/cmake/discover_tests.cmake:114 (message):
  Error running test executable.

    Path: '~/precice/build/testprecice'
    Working directory: '~/precice/build/TestOutput'
[...]

PR #2310, opened 2025-06-16
Move test list generation to separate target

fsimonis:
## Main changes of this PR

This PR moves the test list generation to a separate target which is built by default in `ALL` or if make is involved without arguments.

Before this PR, `testprecice` is executed after it is build to produce the test list. If this fails, the build is considered a failure and the `testprecice` executable is deleted. Therefore, only `make testprecice` is required to run `ctest`.

This PR moves the generation of the test list to a separate command and adds the target `precice-test-list`. Meaning that `make testprecice precice-test-list` is required to run `ctest`. `make testprecice` is not sufficient any more.
The upside is that if the generation of the test list fails, the `testprecice` executable is still present. It can then be debugged or ran separately to create said file.

Fixes https://github.com/precice/precice/issues/2303#issuecomment-2965715922

## Author's checklist

### DIFF
--- a/cmake/CTestConfig.cmake
+++ b/cmake/CTestConfig.cmake
@@ -183,7 +183,7 @@ else()
     VERBATIM)
 
   # Custom target that forces the test list to be updated
-  add_custom_target(test-list ALL
+  add_custom_target(precice-test-list ALL
     DEPENDS "${ctest_tests_file}"
   )
 endif()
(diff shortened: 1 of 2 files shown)


fsimonis:
> Do you have any idea why the generation of the test file would fail?

Looks like `MPI_Init` hangs. It could be that it loads the wrong MPI version or that some network backend leads to a hang.

You could try enabling debug of the MPI implementation via an environment variable.

davidscn:
I think having MPI as a requirement for the test list generation is not appropriate

fsimonis:
> I think having MPI as a requirement for the test list generation is not appropriate

You are right, so I checked the testing code, and we don't initialize MPI when `--list_units` is passed. There must be something else going on.

The CI uses intel 2025 https://github.com/precice/precice/actions/runs/15679615130/job/44167767421#step:6:29

PR #2324, opened 2025-07-10
Support Radius Optimization

FabioTucciarone:
## Main changes of this PR

As requested by @davidscn: a Pull Request with my current progress and on RBF-parameter optimization / automatic tuning.

* Adding a class interface and 2 specializations for optimizing an RBF support-radius.
* Goal: Find an optimal/good range for the support-radius
* `RBFParameterTunerSimple<RBF_T>`: Containing two simple iterative optimizers: 
   - `optimizeIterativeIncrease()`: Iterative sampling in exponential steps.
   - `optimizeBisection()`: Bisection method for convex optimization.
* `RBFParameterTunerBO<RBF_T>`: Uses the Bayes-Optimization algorithm provided by the [limbo](https://resibots.eu/limbo) library for a fixed number of iterations. Optional dependency: [libcmaes (CMA-ES optimizer for limbo)](https://github.com/CMA-ES/libcmaes).
* Support for consistent mapping and strictly positive definite RBFs.
* Support for the Gaussian RBF.

Finding a good shape parameter can be difficult for unexperienced users, but a good choice is critical for accurate results.

### Example runs

error-plot

#### Bayes Optimizer
The Bayes-optimizer is initialized with exponentially increasing samples.
Bayes-optimization is then performed on the logarithmically transformed found interval for 6 iterations. //TODO
```
Bayes-Initializer: rad=5.6346e-03, err=1.0339e+00, 1/cond=2.0258e-04
Bayes-Initializer: rad=5.6346e-02, err=2.5627e-03, 1/cond=2.4013e-08
Bayes-Initializer: rad=5.6346e-01, err=1.2135e-05, 1/cond=2.3648e-12
Bayes-Initializer: rad=5.6346e+00, err=nan, 1/cond=0.0000e+00

State function evaluation: f(5.0071e-01)=-8.0613e-05, [rad=1.7906e-01, err=8.3346e-05]
State function evaluation: f(6.0517e-01)=-1.5454e-05, [rad=3.6846e-01, err=1.5978e-05]
State function evaluation: f(3.9996e-01)=-6.7895e-04, [rad=8.9280e-02, err=7.0197e-04]
State function evaluation: f(7.3876e-01)=-1.0352e-05, [rad=9.2714e-01, err=1.0703e-05]
State function evaluation: f(2.5751e-01)=-1.4228e-02, [rad=3.3373e-02, err=1.4710e-02]
State function evaluation: f(5.5894e-01)=-2.6092e-05, [rad=2.6773e-01, err=2.6976e-05]

Best sample: rad=9.2714e-01, err=1.0703e-05 (LOOCV)

┆ relative-l2 ┆ median(abs) ┆ mapDataTime ┆ computeMappingTime ┆
┆ 0.000005    ┆ 2.5686e-7   ┆ 7.2555598e7 ┆ 4.286847e6         ┆
```
We use the "Expected Improvement" acquisition function with $\xi = 0.02$ (prefer exploitation over exploration).
We see that a good radius is found rather quickly, however, we have not found a sensible error-based termination criterion to stop earlier.

#### Bisection
termination criterion: `upperBound.pos > 1.5 * lowerBound.pos`
```
RBF tuner sample: rad=5.6346e-02, err=2.5627e-03, 1/cond=2.4013e-08
RBF tuner sample: rad=5.6346e-01, err=1.2135e-05, 1/cond=2.3648e-12
RBF tuner sample: rad=5.6346e+00, matrix decomposition failed

Current interval: [5.6346e-01, 5.6346e+00], Sample: rad=3.0990e+00, err=nan
Current interval: [5.6346e-01, 3.0990e+00], Sample: rad=1.8313e+00, err=nan
Current interval: [5.6346e-01, 1.8313e+00], Sample: rad=1.1974e+00, err=1.0420e-05
Current interval: [5.6346e-01, 1.1974e+00], Sample: rad=8.8041e-01, err=1.0795e-05

Best sample: rad=8.8041e-01, err=1.0795e-05 (LOOCV)

┆ relative-l2 ┆ median(abs) ┆ mapDataTime ┆ computeMappingTime ┆
┆ 0.000005    ┆ 2.5782e-7   ┆ 3.5178401e7 ┆ 3.932884e6         ┆

```
The Bisection method is a convex optimization method. Although the is not convex, a sufficiently good result can typically be found in relatively few iterations.
The acquisition function has had only a minor impact in preliminary experiments. 
For terminating early, i.e., setting “max-iterations” low, the EI with a small $\xi$ performs well. 

#### Iteration Method
Increase the sample-radius exponentially until the matrix decomposition fails or the condition number is too high ($> 10^{13}$).
Decrease the stepSize and start again from the last sample until the samples are too close to each other (stepSize > 1.5).
```
RBF tuner sample: rad=5.6346e-03, err=1.0339e+00, 1/cond=2.0258e-04
RBF tuner sample: rad=5.6346e-02, err=2.5627e-03, 1/cond=2.4013e-08
RBF tuner sample: rad=5.6346e-01, err=1.2135e-05, 1/cond=2.3648e-12
RBF tuner sample: rad=5.6346e+00, matrix decomposition failed
RBF tuner sample: rad=1.7818e+00, err=1.0221e-05, 1/cond=1.1102e-14
RBF tuner sample: rad=1.0020e+00, err=1.0593e-05, 1/cond=1.5824e-13
RBF tuner sample: rad=1.7818e+00, err=1.0221e-05, 1/cond=1.1102e-14

Best sample: rad=1.0020e+00, err=1.0593e-05

┆ relative-l2 ┆ median(abs) ┆ mapDataTime ┆ computeMappingTime ┆
┆ 0.000005    ┆ 2.5611e-7   ┆ 4.8298825e7 ┆ 3.903418e6         ┆
```
This method tries to take advantage of the typical appearance of the error graph of the problem, which decreases rapidly until a certain condition number is surpassed.

### Problems and future additions

* Vectorial data has not been tested
* Stopping criteria based on the error are difficult to find, as the optimal error depends on the mesh resolution and RBF.
* Conservative mapping is not yet supported.
* Configuration is not implemented optimally.
   - Added struct `MappingConfiguration::RBFOptional` for optional parameters.
   - Added parameter: `autotune-shape="true"` or `"false"`. Currently not usable without also specifying a "support-radius" or "shape-parameter"
* Only for SPD RBFs: Implementation currently relies on the Cholesky decomposition to compute the LOOCV-error.
* For PUM, it would be a good idea to transfer the knowledge/optimum from one cluster to another one. 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mapping/RadialBasisFctSolver.hpp
+++ b/src/mapping/RadialBasisFctSolver.hpp
@@ -284,16 +284,20 @@ RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::RadialBasisFctSolver(RADIAL_BASIS
 
   // First, assemble the interpolation matrix and check the invertability
   bool decompositionSuccessful = false;
-  if constexpr (RADIAL_BASIS_FUNCTION_T::isStrictlyPositiveDefinite()) {
-    if (_autotuneShape) {
-      _tuner = std::make_unique<RBFParameterTunerSimple<RADIAL_BASIS_FUNCTION_T>>(inputMesh, inputIDs, polynomial, activeAxis);
+  if (_autotuneShape) {
+    if (rbfConfig.optimizationType == MappingConfiguration::AutotuningParams::Type::BAYES_OPTIMIZATION) {
+      _tuner = std::make_unique<RBFParameterTunerBO<RADIAL_BASIS_FUNCTION_T>>(inputMesh, inputIDs, polynomial, activeAxis);
     } else {
-      _decMatrixC             = buildMatrixCLU(basisFunction, inputMesh, inputIDs, activeAxis, polynomial).llt();
-      decompositionSuccessful = _decMatrixC.info() == Eigen::ComputationInfo::Success;
+      _tuner = std::make_unique<RBFParameterTunerSimple<RADIAL_BASIS_FUNCTION_T>>(inputMesh, inputIDs, polynomial, activeAxis);
     }
   } else {
-    _decMatrixC             = buildMatrixCLU(basisFunction, inputMesh, inputIDs, activeAxis, polynomial).colPivHouseholderQr();
-    decompositionSuccessful = _decMatrixC.isInvertible();
+    if constexpr (RADIAL_BASIS_FUNCTION_T::isStrictlyPositiveDefinite()) {
+      _decMatrixC             = buildMatrixCLU(basisFunction, inputMesh, inputIDs, activeAxis, polynomial).llt();
+      decompositionSuccessful = _decMatrixC.info() == Eigen::ComputationInfo::Success;
+    } else {
+      _decMatrixC             = buildMatrixCLU(basisFunction, inputMesh, inputIDs, activeAxis, polynomial).colPivHouseholderQr();
+      decompositionSuccessful = _decMatrixC.isInvertible();
+    }
   }
 
   PRECICE_CHECK(_autotuneShape || decompositionSuccessful,
--- a/src/mapping/impl/RBFParameterTuner.hpp
+++ b/src/mapping/impl/RBFParameterTuner.hpp
@@ -111,9 +111,8 @@ typename RBFParameterTuner<RBF_T>::DecompositionType RBFParameterTuner<RBF_T>::b
     } else {
       return _kernelMatrix.colPivHouseholderQr();
     }
-
   }
-  return DecompositionType();
+  PRECICE_UNREACHABLE("RBF does not support a radius-initialization and was still used to instantiate an optimizer.");
 }
 
 template <typename RBF_T>


CMakeLists.txt
```diff
@@ -218,6 +218,22 @@ precice_validate_libxml2()
 # libfmt
 add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/thirdparty/fmt)
 
+# limbo
+add_subdirectory(${CMAKE_CURRENT_SOURCE_DIR}/thirdparty/limbo)
+# optional optimizer packages for limbo
+find_package(libcmaes QUIET)
```
davidscn, line 224: Limbo allows to use other optimizer packages as well, no?

CMakeLists.txt
```diff
@@ -218,6 +218,22 @@ precice_validate_libxml2()
...
+if(libcmaes_FOUND)
+  message(STATUS "Found libcmaes for limbo.")
+else()
+  message(STATUS "libcmaes (CMA-ES optimizer) for limbo not found.")
+  find_package(libnlopt QUIET)
+  if(libnlopt_FOUND)
+    message(STATUS "Found NLopt for limbo.")
```
davidscn, line 231: Did you try using nlopt?

src/mapping/MathHelper.hpp
```diff
@@ -66,4 +67,176 @@ inline Eigen::MatrixXd invertLowerTriangularBlockwise(const Eigen::MatrixXd &L)
...
+  Eigen::VectorXd inverseDiagonal = (L_inv.array().square().colwise().sum()).transpose();
+
+  return inverseDiagonal;
+}
+
+/**
+ * @brief For C = QR, compute the diagonal entries of the inverse kernel matrix.
+ * @param decMatrixC QR decomposition of the kernel matrix.
+ */
+inline Eigen::VectorXd computeInverseDiagonal(const Eigen::ColPivHouseholderQR<Eigen::MatrixXd> &decMatrixC)
```
davidscn, line 112: @FabioTucciarone you write that only spd kernels are supported at the moment, but we have the inverse diagonal for the QR as well, so cpd kernels should work as well, no?

src/mapping/MathHelper.hpp
```diff
@@ -66,4 +67,176 @@ inline Eigen::MatrixXd invertLowerTriangularBlockwise(const Eigen::MatrixXd &L)
   }
   return L_inv;
 }
+
+/**
+ * @brief For C = LL^T, compute the diagonal entries of the inverse kernel matrix.
+ * @param decMatrixC Cholesky decomposition of the kernel matrix.
+ */
+inline Eigen::VectorXd computeInverseDiagonal(const Eigen::LLT<Eigen::MatrixXd> &decMatrixC)
```
davidscn, line 75: Since Eigen is entirely template based, it could also be an interesting idea to try single precision for the first few shots (especially for Bayes). The condition number will suffer more though.

src/mapping/config/MappingConfiguration.hpp
```diff
@@ -98,6 +98,13 @@ class MappingConfiguration : public xml::XMLTag::Listener {
     bool                basisFunctionDefined = false;
   };
 
+  struct RBFOptional {
```
davidscn: Seems like a generic name. Why optional?

src/mapping/config/MappingConfiguration.hpp
```diff
@@ -98,6 +98,13 @@ class MappingConfiguration : public xml::XMLTag::Listener {
...
+    bool autotuneShape = false;
+
+    std::string executor = "executor"; // TODO: required for compilation but meaningless
+    std::string solver   = "solver";   // TODO: required for compilation but meaningless
```
davidscn: TODO

src/mapping/impl/BayesOptRBFParameterTuner.hpp
```diff
@@ -0,0 +1,301 @@
+#pragma once
+
+#include "mapping/impl/RBFParameterTuner.hpp"
+
+#include <libcmaes/cmaes.h>
```
davidscn: CMake states that this is a conditional dependency, but we include it here unconditionally. Does this compile without libcmaes?

davidscn (commented):
Thanks for now! I will take a closer look at the code at a later time. For now, we should conclude on

- how this is supposed to look like from the user perspective @uekerman 
- whether we would want to ship limbo with preCICE (I checked the licensing compatibility and from what I understand this should be possible) if we want to include the BO optimizer @fsimonis 
- whether we want both variants (bisection and BO) in the code base. Is there a clear use case for the 14k lines of codes added through limbo? We should find a scenario where this has an actual use cases, i.e., BO clearly outperformed bisection. Did bisection fail for any scenario or was the performance bad? Could also be something which only shows its benefit for a time dependent problem and something we cannot answer right now. What do you think @FabioTucciarone?

davidscn:
Discussing with @uekerman:

- from the user perspective, we would want the optimizer to work if no support-radius or shape-parameter was specified
- the last point from above: Is there a clear use case for the 14k lines of codes added through limbo?
- we should log the support radius the optimizer found
- it would also make sense to log the cross validation error throughout the simulation

If there is no clear use case for BO, we could merge the bisection in a first step only and then gather some experience with it.

CMakeLists.txt
FabioTucciarone: libcmaes, NLopt or the standard grid search.

CMakeLists.txt
FabioTucciarone: I have now: it should, however, be `NLopt_FOUND` and `find_package(NLopt QUIET)`. I fixed that.
I didn't see any big difference in quality or runtime in my experiments. libcmaes seems to be slightly slower and more accurate.

src/mapping/MathHelper.hpp
FabioTucciarone: Yes, my fist implementation only considered SPD kernels using the Cholesky decomposition.
My current version supports both.

src/mapping/config/MappingConfiguration.hpp
FabioTucciarone: I changed the name to `AutotuningParams` and changed how the optimization is configured:
It is now possible to specify `support-radius="auto"`.

src/mapping/config/MappingConfiguration.hpp
FabioTucciarone: This was required because I passed the parameters as an optional argument to `RadialBasisFctMapping`, which caused a compilation error in `RadialBasisFctMapping<SOLVER_T, Args...>::getName()`. I have changed that.

src/mapping/impl/BayesOptRBFParameterTuner.hpp
FabioTucciarone: It shouldn't. For some reason, the `#include` was still found when I tested the default grid search.

Issue #2018, opened 2024-05-08
Logging in compositional coupling schemes


davidscn:
**Please describe the problem you are trying to solve.**
The following code lines

https://github.com/precice/precice/blob/f34556269887329c2d6e2e58b0e3ac20a302864b/src/cplscheme/CompositionalCouplingScheme.cpp#L320-L335

lead to a linebreak in the user logging.

```bash
--- [precice] MuscleMechanics|0|impl::DataContext|l132|mapData|Mapping "Displacement" for t=0.38 from "TendonTopAMesh" to "MuscleMeshTopA"
--- [precice] MuscleMechanics|0|impl::DataContext|l132|mapData|Mapping "Velocity" for t=0.38 from "TendonTopAMesh" to "MuscleMeshTopA"
--- [precice] MuscleMechanics|0|impl::DataContext|l132|mapData|Mapping "Displacement" for t=0.38 from "TendonTopBMesh" to "MuscleMeshTopB"
--- [precice] MuscleMechanics|0|impl::DataContext|l132|mapData|Mapping "Velocity" for t=0.38 from "TendonTopBMesh" to "MuscleMeshTopB"
--- [precice] MuscleMechanics|0|impl::ParticipantImpl|l391|advance|MuscleFibers: iteration: 1, time-window: 39, time: 0.38 of 50, time-window-size: 0.01, max-time-step-size: 0.01, ongoing: yes, time-window-complete: yes, 
TendonBottom: iteration: 1 of 30 (min 1), time-window: 39, time: 0.38 of 50, time-window-size: 0.01, max-time-step-size: 0.01, ongoing: yes, time-window-complete: yes, write-iteration-checkpoint
```

This is semi-conformal with the logging API we provide the preCICE user. 
My format specification is `%TimeStamp(format="%H:%M:%S.%f")%|%Participant%|%Rank%|%Module%|l%Line%|%Function%|%Severity%%Message%`, such that the bottom line is rather dangling, I was not completely sure that it comes from preCICE in the first place. Actually, I don't even know what the bottom line means or what it is supposed to say.

**Describe the solution you propose.**
Find out the meaning and clarify. Forbid linebreaks in the log message.

fsimonis:
> This is semi-conformal with the logging API we provide the preCICE user.

This violates our "one line per log entry"-rule. So, this is definitely not good.

> Actually, I don't even know what the bottom line means or what it is supposed to say.

[...]

PR #2325, opened 2025-07-15
Compact iteration output message

fsimonis:
## Main changes of this PR

This PR implements the suggestions in https://github.com/precice/precice/issues/2018#issuecomment-2122169024 to reduce the overall verbosity of the coupling state output.
While a minor effect in simulations with a single coupling-scheme, this makes compositional schemes easier to understand.

* iteration -> it, time -> t, time-window-size -> Dt -> max-time-step-size -> max-dt
* iterations were removed from explicit schemes
* time-window complete was removed as it implied by the iteration
* write/read-iteration-checkpoint were removed as they are also implied by the iteration

Note that these examples below are up-to-date with the code. The conversation is most likely about an older version.

Implicit:

```
it 1 (min: 2, max: 2), time-window 1 (max: 2), t 0, Dt 1, max-dt 1
it 1 (min: 2, max: 2), time-window 1 (max: 2), t 0, Dt 1, max-dt 1
it 2 (min: 2, max: 2), time-window 1 (max: 2), t 0, Dt 1, max-dt 1
it 2 (min: 2, max: 2), time-window 1 (max: 2), t 0, Dt 1, max-dt 1
it 1 (min: 2, max: 2), time-window 2 (max: 2), t 1, Dt 1, max-dt 1
it 1 (min: 2, max: 2), time-window 2 (max: 2), t 1, Dt 1, max-dt 1
it 2 (min: 2, max: 2), time-window 2 (max: 2), t 1, Dt 1, max-dt 1
it 2 (min: 2, max: 2), time-window 2 (max: 2), t 1, Dt 1, max-dt 1
Reached end at: final time-window: 2, final time: 2
Reached end at: final time-window: 2, final time: 2
```

Explicit:

```
time-window 1 (max: 2), t 0, Dt 1, max-dt 1
time-window 1 (max: 2), t 0, Dt 1, max-dt 1
time-window 2 (max: 2), t 1, Dt 1, max-dt 1
time-window 2 (max: 2), t 1, Dt 1, max-dt 1
Reached end at: final time-window: 2, final time: 2
Reached end at: final time-window: 2, final time: 2
```

Helps to shrink the output after: #2353

## Author's checklist

### DIFF
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -615,12 +615,6 @@ std::string BaseCouplingScheme::printCouplingState() const
   if (hasTimeWindowSize()) {
     fmt::format_to(out, ", Dt {}, max-dt {}", _timeWindowSize, getNextTimeStepMaxSize());
   }
-  if (!_requiredActions.empty()) {
-    fmt::format_to(out, ", ");
-    for (auto action : _requiredActions) {
-      fmt::format_to(out, "{} ", toString(action));
-    }
-  }
   return str;
 }
 


uekerman:
In the example above, the iterating / converged does not fit the iteration counter. I have a feeling where this comes from, but might confuse many users.

Also "converged" might be confusing in such an example as it is not really "converged" in the mathematical sense, but "only" the minimum number of its is reached.

Alternative for both problems: Do we need this information in the first place? The convergence measures are logging this anyway.

fsimonis:
This is a direct tanslation of what was proposed in https://github.com/precice/precice/issues/2018#issuecomment-2122169024.

I aggree that this is confusing, and the code to deduce the various states is anything but easy to understand.

> Also "converged" might be confusing in such an example as it is not really "converged" in the mathematical sense, but "only" the minimum number of its is reached.

True

> In the example above, the iterating / converged does not fit the iteration counter. I have a feeling where this comes from, but might confuse many users.

That's right in the sense that the `time-window` part explains the result of the last iteration, while the iteration counter shows the current iteration.

> Do we need this information in the first place? The convergence measures are logging this anyway.

We could also remove `time-window-complete` as it is implied by the iteration and the time window count.

fsimonis:
I removed the `time-window-complete` section. The new output looks like this:

```
it 1 of 2 (min 2), time-window 1 of 2, t 0, Dt 1, max-dt 1, write-iteration-checkpoint 
it 1 of 2 (min 2), time-window 1 of 2, t 0, Dt 1, max-dt 1, write-iteration-checkpoint 
it 2 of 2 (min 2), time-window 1 of 2, t 0, Dt 1, max-dt 1, read-iteration-checkpoint 
it 2 of 2 (min 2), time-window 1 of 2, t 0, Dt 1, max-dt 1, read-iteration-checkpoint 
it 1 of 2 (min 2), time-window 2 of 2, t 1, Dt 1, max-dt 1, write-iteration-checkpoint 
it 1 of 2 (min 2), time-window 2 of 2, t 1, Dt 1, max-dt 1, write-iteration-checkpoint 
it 2 of 2 (min 2), time-window 2 of 2, t 1, Dt 1, max-dt 1, read-iteration-checkpoint 
it 2 of 2 (min 2), time-window 2 of 2, t 1, Dt 1, max-dt 1, read-iteration-checkpoint 
Reached end at: final time-window: 2, final time: 2
Reached end at: final time-window: 2, final time: 2
```

MakisH:
In the formulation `it 1 of 2 (min 2)`, the `of 2` refers to the configured `max-iterations` and the `(min 2)` refers to the `min-iterations`, right?

A clearer formulation could be `it 1 (min: 2, max: 5)`.

This would be inconsistent with the time windows, but I think this is fine: there will always be a `time-window 2 of 2`, but not always a `it 100 of 100`.

When there are no values given by the user, one of the two could be omitted (`it 1 (max: 5)`, `it 1 (min: 2)`), or the default values could be printed, instead (`it 1 (min: 1, max: none)`)*. 

The latter would be a bit more verbose, but make tooling easier.

**Note:** I think that this one does not yet close #2018, as the different schemes are not clearly annotated. It would be great to highlight when there are multiple coupling schemes and which output is from which one. They don't have names, but I imagine that even an ID could be useful when debugging a log of a composite scheme. Not sure what an easy solution here could be, since this output comes already from the base coupling scheme.

*Do we maybe need a default value for `max-iterations`?

fsimonis:
> This would be inconsistent with the time windows, but I think this is fine: there will always be a time-window 2 of 2, but not always a it 100 of 100.

The `of 2` in the time-window depends on `<max-time-windows` which is optional. The same counts for t and `max-time`.

Consistent would be:
```
it 2 (min:2, max:2), time-window 2 (max:2), t 1, Dt 1, max-dt 1, read-iteration-checkpoint 
```

Or when limiting time:
```
it 2 (min:2, max:2), time-window 2, t 1 (max:2), Dt 1, max-dt 1, read-iteration-checkpoint 
```


> I think that this one does not yet close https://github.com/precice/precice/issues/2018, as the different schemes are not clearly annotated.

Do you mean the type in the sense of serial vs parallel vs multi?

> It would be great to highlight when there are multiple coupling schemes and which output is from which one. They don't have names, but I imagine that even an ID could be useful when debugging a log of a composite scheme. Not sure what an easy solution here could be, since this output comes already from the base coupling scheme.

When multiple coupling schemes are defined, each output is prefixed by the name of the coupling partner(s).

```
it 2 (min:2, max:2), time-window 2, t 1 (max:2), Dt 1, max-dt 1, read-iteration-checkpoint 
```

Becomes:

```
SolverTwo: it 2 (min:2, max:2), time-window 2, t 1 (max:2), Dt 1, max-dt 1, read-iteration-checkpoint 
```

Only when a single coupling scheme is defined, then there is no partner in the log.

fsimonis:
This test is useful to inspect compositional output:

```
ctest -VV -R FiveParticipant
```

Example:

```
M1SM|0|impl::ParticipantImpl|l454|advance|M2: time-window 2 (max: 2), t 1, Dt 1, max-dt 1
M1: time-window 2 (max: 2), t 1, Dt 1, max-dt 1
Tendon: it 1 (min: 1, max: 1), time-window 2 (max: 2), t 1, Dt 1, max-dt 1, write-iteration-checkpoint 
```

This output violates out rule to have one line per log. The following would be better:

```
M1SM|0|impl::ParticipantImpl|l454|advance|M2: time-window 2 (max: 2), t 1, Dt 1, max-dt 1; M1: time-window 2 (max: 2), t 1, Dt 1, max-dt 1; Tendon: it 1 (min: 1, max: 1), time-window 2 (max: 2), t 1, Dt 1, max-dt 1, write-iteration-checkpoint 
```

There could also be multiple info statements:

```
M1SM|0|impl::ParticipantImpl|l454|advance|M2: time-window 2 (max: 2), t 1, Dt 1, max-dt 1
M1SM|0|impl::ParticipantImpl|l454|advance|M1: time-window 2 (max: 2), t 1, Dt 1, max-dt 1
M1SM|0|impl::ParticipantImpl|l454|advance|Tendon: it 1 (min: 1, max: 1), time-window 2 (max: 2), t 1, Dt 1, max-dt 1, write-iteration-checkpoint 
```

MakisH:
I think that forcing every scheme log into one line would make the log less readable.
The last variant seems to me easier to read and understand, but I would also understand each line as a separate event.

fsimonis:
@IshaanDesai and I came to the conclusion that similar to the `time-window-completed` tag, the actions are also redundant and can be deduced from the iteration and the time window.
Dropping those would make the implicit coupling output even clearer.

The value of `max-dt` is also questionable for the user, but it is useful for us.

fsimonis:
In my discussion with @carme-hp about this, we came to the conclusion that the `3 of 10` format is very easy to read even though it makes it difficult to incorporate the `min`. She also thinks that displaying an unset iteration minimum could be helpful `it 2 (min:None, max:10)`.
She is also in favor of using one line per participant in compositional schemes instead of merging it all into one line.

fsimonis:
After a lot of feedback we kind of converged. Let's merge here and iterate.

PR #2330, opened 2025-07-18
Fix applying safety factor during repartitioning

davidscn:
## Main changes of this PR

Applies the safety factor to the received mesh pieces rather then (potentially) repeatedly to the same bounding box used in the `ReceivedPartition`

## Author's checklist

## Reviewers' checklist


fsimonis (commented):
The implementation reapplied the safety factor, so this is obviously a bug fix which requires a changelog entry.

The proposed solution applies the safety factor to the BB of each mapping before merging. I agree that this makes the most sense, however, the comments and log statements in the ReceivedPartition state that the bounding boxes are first merged and the expanded. So they need adjusting.

PR #2331, opened 2025-07-21
Allow defining multiple access regions per rank

davidscn:
## Main changes of this PR

Allows defining multiple access regions per rank.

Allows calling `setMeshAccessRegion` multiple times before calling `initialize()`. preCICE then repartitions and operates on a bounding box collection rather than a single bounding box. This gives additional flexibility for describing the domain in terms of axis-aligned bounding boxes.

FYI @IshaanDesai 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mesh/Mesh.cpp
+++ b/src/mesh/Mesh.cpp
@@ -390,9 +390,14 @@ const BoundingBox &Mesh::getBoundingBox() const
   return _boundingBox;
 }
 
-void Mesh::expandBoundingBox(const BoundingBox &boundingBox)
+void Mesh::appendToAccessRegions(const BoundingBox &boundingBox)
 {
-  _boundingBox.expandBy(boundingBox);
+  _accessRegions.emplace_back(boundingBox);
+}
+
+const std::vector<BoundingBox> &Mesh::getAccessRegions() const
+{
+  return _accessRegions;
 }
 
 void Mesh::preprocess()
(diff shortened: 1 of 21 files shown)


davidscn:
> IMO the only thing this PR needs to a pass over the documentation in the Participant.hpp.

Probably also some documentation updates.

The only real drawback I can think of at the moment is that collections spanning the entire global domain, e.g., the sit in opposite corners, would lead to a rather expensive initialization for the two-level initialization approach: The corresponding secondary participant would connect to all ranks of the coupling partner. This could potentially lead to scaling issues. For the single-level initialization, that's not a problem. Changing this behavior for 2LI would require to rewire the entire bounding box exchange and more complicated. I think the approach here is a comparable low effort for a quite significant gain in terms of domain descriptions.

Waiting for @IshaanDesai, who wanted to try it out in one of his application examples.

fsimonis:
> The only real drawback I can think of at the moment is that collections spanning the entire global domain, e.g., the sit in opposite corners, would lead to a rather expensive initialization for the two-level initialization approach:

True. This case may finally justify changing the current single bounding-box processing to some sort of [bounding volume hierarchy](https://en.wikipedia.org/wiki/Bounding_volume_hierarchy#BVH_design_issues). This could also allow switching between bounding boxes and bounding spheres to allow for tighter volumes and thus lower communication costs.
Could be an interesting thesis topic.

PR #2333, opened 2025-07-24
Improve test discovery script and test without MPI

fsimonis:
## Main changes of this PR

This PR ports the test discovery script to python, adds the ability to run 1 rank tests without MPI, and adds the amount of ranks as a label.

The CMake version of the script is very tedious to maintain due to the amount of cumbersome string processing.

## Author's checklist


cmake/discover_tests.py
```diff
@@ -0,0 +1,125 @@
...
+
+    pargs = [
+        testname,
+        "PROPERTIES",
+        "ENVIRONMENT",
+        '"OMP_NUM_THREADS=2;OMP_PROC_BIND=false"',
+        "TIMEOUT",
+        '"30"',
+        "WORKING_DIRECTORY",
+        f'"{dir}"',
```
fsimonis:
```suggestion
        f'"{dir.as_posix()}"',
```

PR #2336, opened 2025-07-25
Add default intracom without MPI

### LINKED ISSUE #2026: Default intra-comm when MPI turned off

When building preCICE without MPI, each parallel participant needs to define an `<intra-comm:sockets/>`.
This is tedious and could be the default behaviour. 

When preCICE is build without MPI, default to a sockets-based intra-comm.

Keep the current behaviour, preventing surprises.

**Additional context**

This would streamline the use of preCICE without MPI.
A further simplification is #2025

fsimonis:
## Main changes of this PR

This PR adds a default intra-comm for the case that preCICE is not build with MPI.
It uses the default loopback as network and emits an info message on how to customize the comm by defining your own.

Closes #2026
Related to #2334

## Author's checklist

### DIFF
--- a/docs/changelog/2336.md
+++ b/docs/changelog/2336.md
@@ -1,0 +1,1 @@
+- Added implicit socket-based intra-comms for parallel participants in case preCICE is compiled without MPI.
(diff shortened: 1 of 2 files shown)


fsimonis:
We currently cannot test this feature as we need MPI to run parallel tests.

PR #2337, opened 2025-07-25
Ignore intra-comm when running in serial

### LINKED ISSUE #2025: Error defining an intra-comm tag and running in serial
**Describe your setup**

preCICE Version: 3.x

**Describe the problem**

Defining an `intra-comm` tag leads to an error when a participant runs in Serial.

**Step To Reproduce**

1. Use an existing case that may run in parallel
2. Add `<intra-comm:sockets />` to the parallel participant
3. Run the participant with 1 rank

**Expected behaviour**

No error when running in serial.
An info statement and at most a warning that the defined intra-comm is ignored as this is a serial participant.

fsimonis:
## Main changes of this PR

This PR fixed the error message that appears when configuring an `<intra-comm:x />` and running that participant in serial. 
preCICE now ignores the defined intra-communicator and emits the info:

```
Ignoring user-defined intra-comm for participant SolverTwo as it is running in serial.
```

Closes #2025

## Author's checklist


fsimonis:
~~Waiting for #2338 to test this~~
Which I can't as the check is performed in `ParticipantImpl::configure` for some reason.

PR #2340, opened 2025-07-27
Bisection Support Radius Optimization

FabioTucciarone:

Adds a bisection optimizer for RBF support-radius tuning (see pull request [#2342](https://github.com/precice/precice/pull/2324)).
This pull request fixes some problems mentioned in the discussion of [#2342](https://github.com/precice/precice/pull/2324) and removes the Bayesian optimizer for now. A use case may be found later for vectorial data.

**Configuration example:** `<basis-function:gaussian support-radius="auto"/>`

## Author's checklist

## Reviewers' checklist


src/mapping/MathHelper.hpp
```diff
@@ -66,4 +67,184 @@ inline Eigen::MatrixXd invertLowerTriangularBlockwise(const Eigen::MatrixXd &L)
...
+  if (qrDec.info() != Eigen::ComputationInfo::Success) {
+    return 0;
+  }
+  const Eigen::Index n = qrDec.matrixR().rows();
+
+  double max_r = std::numeric_limits<double>::min();
+  double min_r = std::numeric_limits<double>::max();
+
+  for (Eigen::Index i = 0; i < n; i++) {
+    const double rii = qrDec.matrixR()(i, i);
```
davidscn: Couldn't we merge this code with the LLT version using `if constexpr` and a template?

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -287,6 +287,11 @@ class Gaussian : public CompactSupportBase,
     return _params;
   };
 
+  static constexpr double transformRadiusToShape(const double radius)
```
davidscn:
```suggestion
  inline constexpr double transformRadiusToShape(const double radius)
```

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -287,6 +287,11 @@ class Gaussian : public CompactSupportBase,
...
+  {
+    return std::sqrt(-std::log(cutoffThreshold)) / radius;
```
davidscn: the linter complains, I guess the `std` functions here are not `constexpr`?

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,173 @@
...
+#include <mapping/MathHelper.hpp>
+#include <mapping/RadialBasisFctSolver.hpp>
+#include <mapping/config/MappingConfigurationTypes.hpp>
+#include <mapping/impl/BasisFunctions.hpp>
+
+namespace precice::mapping {
+
+// Forward declaration of function found in <mapping/RadialBasisFctSolver.hpp>
+template <typename RADIAL_BASIS_FUNCTION_T, typename IndexContainer>
+Eigen::MatrixXd buildMatrixCLU(RADIAL_BASIS_FUNCTION_T basisFunction, const mesh::Mesh &inputMesh, const IndexContainer &inputIDs, std::array<bool, 3> activeAxis, Polynomial polynomial);
```
davidscn: I think the `Mesh` include is missing here

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,173 @@
...
+  double h = 0;
+  for (int i = 0; i < sampleSize; i++) {
+    const mesh::Vertex xi = inputMesh.vertices().at(matches.at(i));
+    h += std::sqrt(utils::computeSquaredDifference(xi.rawCoords(), x0.rawCoords()));
+  }
+  return h / sampleSize;
+}
+
+template <typename RBF_T>
+double RBFParameterTuner<RBF_T>::getMinBoundSize(const mesh::Mesh &inputMesh)
```
davidscn: I think the `index` has this information already

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,173 @@
...
+    PRECICE_ASSERT(false, "Selected RBF does not support a radius.");
+    expr = [&](double x) { return x; };
+  }
+  return matrixExpr.unaryExpr(expr);
+}
+
+template <typename RBF_T>
+double RBFParameterTuner<RBF_T>::estimateMeshResolution(const mesh::Mesh &inputMesh)
+{
+  constexpr int sampleSize = 3;
```
davidscn:
```suggestion
  constexpr int sampleSize = 10;
```
?

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,173 @@
...
+  {
+  }
+};
+
+template <typename RBF_T>
+class RBFParameterTuner {
+
+protected:
+  mutable Eigen::MatrixXd _kernelMatrix;
+  Eigen::MatrixXd         _distanceMatrix;
```
davidscn: If you think it brings us performance benefits we can store the distance matrix temporarily. However, after an optimization was completed, we should deallocate it: the matrices are anyway memory heavy

src/mapping/impl/SimpleRBFParameterTuner.hpp
```diff
@@ -0,0 +1,167 @@
+#pragma once
+
+#include "mapping/impl/RBFParameterTuner.hpp"
+
+namespace precice::mapping {
+
+template <typename RBF_T>
+class RBFParameterTunerSimple : public RBFParameterTuner<RBF_T> {
```
davidscn: Why not calling it bisection?

src/mapping/impl/SimpleRBFParameterTuner.hpp
```diff
@@ -0,0 +1,167 @@
...
+  mutable logging::Logger _log{"mapping::RBFParameterTuner"};
+
+public:
+  template <typename IndexContainer>
+  RBFParameterTunerSimple(const mesh::Mesh &inputMesh, const IndexContainer &inputIDs, const Polynomial &polynomial, const std::array<bool, 3> &activeAxis);
+
+  double optimize(const Eigen::VectorXd &inputData) override;
+
+  Sample optimizeIterativeIncrease(const Eigen::VectorXd &inputData, double posTolerance, size_t maxSuccessfulSamples);
+  Sample optimizeBisection(const Eigen::VectorXd &inputData, double posTolerance, double errorTolerance, int maxIterations);
```
davidscn: Here, only the `override` function should be public, all others should be private

src/mapping/impl/SimpleRBFParameterTuner.hpp
```diff
@@ -0,0 +1,167 @@
...
+};
+
+// Implementation:
+
+template <typename RBF_T>
+template <typename IndexContainer>
+RBFParameterTunerSimple<RBF_T>::RBFParameterTunerSimple(const mesh::Mesh &inputMesh, const IndexContainer &inputIDs, const Polynomial &polynomial, const std::array<bool, 3> &activeAxis)
+    : RBFParameterTuner<RBF_T>(inputMesh, inputIDs, polynomial, activeAxis)
+{
+  PRECICE_ASSERT(this->rbfSupportsRadius(), "RBF is not supported by this optimizer, as it does not accept a support-radius."
```
davidscn: Or do you check this before already?

```suggestion
  PRECICE_CHECK(this->rbfSupportsRadius(), "RBF is not supported by this optimizer, as it does not accept a support-radius."
```

src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -449,6 +401,17 @@ Eigen::VectorXd RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::solveConsistent(E
     inputData -= (_matrixQ * polynomialContribution);
   }
 
+  double optimizedRadius = 0;
+
+  if (_autotuneShape) {
+    if constexpr (RADIAL_BASIS_FUNCTION_T::isStrictlyPositiveDefinite()) {
+      optimizedRadius = _tuner->optimize(inputData); // TODO: Optimization in every iteration is not ideal.
```
davidscn: No it's not. If we cannot come up with a decent strategy to determine when to recompute, then we should move this code to `computeMapping`

src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -449,6 +401,17 @@ Eigen::VectorXd RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::solveConsistent(E
...
+      _decMatrixC     = _tuner->buildKernelDecomposition(optimizedRadius);
```
davidscn: How many times do we store the matrix decomposition now? After the optimization completed, the decomposition should be available already. Why recomputing it?

src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -449,6 +401,17 @@ Eigen::VectorXd RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::solveConsistent(E
...
+    } else {
+      PRECICE_ASSERT(false, "Not supported.");
+    }
+  }
+
```
davidscn: Can we log the LOOCV error when getting new data?

FabioTucciarone:
> I am not convinced by the code design: the Solver class doesn't really solve now right, but solving happens inside the optimizer class?
> 
> Now, we build kernel matrices in the optimizer. These code paths would be duplicate if we implement another optimizer.
> 
> Separation of concerns if important. The Optimizer base should define the interface and potentially common routines. The bisection optimizer defines how we iterate through the parameter space. All RBF related operations should be provided by the RBFSolver class. This would also allow to potentially reuse the optimizer in other places.

The Optimizer only uses the LOOCV Error and condition number approximation from the MathHelper.h file and the matrix assembly method from the solver class. It also stores the kernel matrix and the distance matrix, which I could reconsider.
I'm not sure if I understand your concern correctly. Should I remove the kernel matrices and instead add methods to the Solver for building the decomposition and accessing the error?

> Maybe the Optimizer can be defined as class template with a `typename Operator` parameter, which is then the solver class. In the current design, it's intransparent where the solve part lives and where kernel matrices are stored.

Would you suggest instantiating a Solver within the Optimizer? The optimizer does not need the full mapping but only the interpolation coefficients for the LOOCV Error.

davidscn:
>  Should I remove the kernel matrices and instead add methods to the Solver for building the decomposition and accessing the error?

Exactly

> Would you suggest instantiating a Solver within the Optimizer? The optimizer does not need the full mapping but only the interpolation coefficients for the LOOCV Error.

There is not instantiation necessary. If you make it a class or function template, the implementation could look like

```cpp
<typename Operator>
Bisection<Operator>::optimize(Operator& solver, double tolerance)
{
    while(error > tolerance)
    {
      // compute new support radius in the parameter space...
     double error = solver.computeErrorEstimate(parameter);
    }
}
```

then the solver just needs to provide the functions for the Optimizer to work. Since you anyway have the BO optimizer as an alternative, the above approach should minimize code duplication. Adding the BO would really only calling the already available solver routines in the BO optimizer provided by limbo. I could imagine that maybe limbo already defines an interface you could consider for designing the Optimizer interface. What do you think?

Could you also check the failing tests/missing includes?

FabioTucciarone:
Thank you for the clarification, that seems sensible.

> Since you anyway have the BO optimizer as an alternative, the above approach should minimize code duplication. 
The implementation of the Bayes Optimization is already very minimal, most of the code consists of configuring the limbo optimizer.
> Adding the BO would really only calling the already available solver routines in the BO optimizer provided by limbo.

That is already mostly the case for the function calls to the optimizer.
The initialization code and state function definition is not completely "compatible"

> I could imagine that maybe limbo already defines an interface you could consider for designing the Optimizer interface. What do you think?

Your suggested change would in principle be similar to how limbo handles the state function.

src/mapping/impl/BasisFunctions.hpp
FabioTucciarone: The inline specifier is redundant in classes. "A function defined within a class definition is an inline function." 10.1.6 C++17 standard. 
Why not static?

davidscn: Does it make sense to have it as a member function or rather a free function, considering that you want it to be static? Dont we need here a function interface which applies to all basis functions?

FabioTucciarone:
I restructured the code. Without storing the `_distanceMatrix` I need access to the `IndexContainer` and the input mesh to evaluate the new `_kernelMatrix`. I therefore store a pointer to the input mesh (so that the standard constructor is still available) in the Solver class and copy the indices into an Eigen vector.
However, his is not optimal. I tried to pass the index container to every function that needed it, but this turned out to be very convoluted.
I would therefore recommend replacing the `IndexContainer` template type with an `std::variant`:
```c++
using IndexContainerVariant = std::variant<boost::container::flat_set<VertexID>, boost::integer_range<VertexID>>;
```
Which can then be used in combination with `std::visit`.

davidscn:
> However, his is not optimal. I tried to pass the index container to every function that needed it, but this turned out to be very convoluted.

Could be still ok. You want to replace it by a variant to make it a member variable. But I would not advice to duplicate the index set in the solver class. Changing the index set in the PU or RBF will become difficult otherwise. The solver should also keep its operator-like responsibility. Maybe you can introduce free functions to isolate functionality into other places? You could also use argument forwarding if the signature becomes a burden?

FabioTucciarone:
> Maybe you can introduce free functions to isolate functionality into other places? You could also use argument forwarding if the signature becomes a burden?

I'm not sure how this would help.

> You want to replace it by a variant to make it a member variable.

Otherwise, we would have to add it to the function signature of `RadialBasisFctMapping<SOLVER_T>::mapConsistent()`

davidscn:
> Otherwise, we would have to add it to the function signature of RadialBasisFctMapping<SOLVER_T>::mapConsistent()

Rather `RadialBasisFctSolver<T>::solveConsistent()`? Changing the signature of the mapping method doesnt seem useful

FabioTucciarone:
- I added a new configuration option to the basis function tag: `<basis-function:gaussian support-radius="auto" tuner-optimization-interval="1"/>`.
  This allows the user to configure how often the optimizer is called.
- `tuner-optimization-interval="n"`: every n `solveConsistent` calls. This will be changed, so that an optimization happens every n iterations (mapData calls), in the future, as soon as the optimizer is able to take care of vectorial data.
- `tuner-optimization-interval="once"` only optimizes the support radius once for non-zero input data. 
If the input data is zero, the optimizer stops early and is called again during the next iteration.

benchmarks/rbf-assembly-kernels.cpp
davidscn, line 1: What was the incentive for the changes in the file?

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -617,6 +617,38 @@ class CompactPolynomialC8 : public CompactSupportBase,
...
+
+  static constexpr bool isAvailable()
+  {
+    return RADIAL_BASIS_FUNCTION_T::hasCompactSupport() || definesCutoffThreshold<RADIAL_BASIS_FUNCTION_T>;
+  }
+
+  static double transformRadiusToShape(const double radius)
+  {
+    if constexpr (requiresConversion()) {
+      return std::sqrt(-std::log(RADIAL_BASIS_FUNCTION_T::cutoffThreshold)) / radius;
```
davidscn: This is specific to the Gauss kernel, right? Then the function name should reflect this

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -617,6 +617,38 @@ class CompactPolynomialC8 : public CompactSupportBase,
...
+    } else {
+      return radius;
+    }
+  }
+};
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+template <typename T>
+constexpr bool RadiusInitialization<RADIAL_BASIS_FUNCTION_T>::definesCutoffThreshold<T, std::void_t<decltype(T::cutoffThreshold)>> = true; // SFINAE specialization
```
davidscn: Why keeping it general here? The Gaussian is the only RBF we have that specifies a cutoff. As far as I can see, the radius initialization is used in the `MappingConfiguration`, then why keeping the code here? Let's move it there?

src/mapping/impl/BisectionRBFTuner.hpp
```diff
@@ -0,0 +1,118 @@
...
+template <typename Solver>
+template <typename IndexContainer>
+Sample BisectionRBFTuner<Solver>::optimizeBisection(const Solver &solver, const Eigen::VectorXd &inputData, const IndexContainer &inputIds, double posTolerance, double errorTolerance, int maxIterations)
+{
+  constexpr double increaseSize = 10;
+  double           sampleRadius = this->_lowerBound;
+
+  this->_lastSampleWasOptimum = false;
+
+  PRECICE_INFO("Start optimization with lower bound = {:.4e}, upper bound = {:.4e}", this->_lowerBound, this->_upperBound);
```
davidscn:
```suggestion
  PRECICE_INFO("Starting optimization with lower bound = {:.4e}, upper bound = {:.4e}", this->_lowerBound, this->_upperBound);
```

src/mapping/impl/BisectionRBFTuner.hpp
```diff
@@ -0,0 +1,118 @@
...
+  // collect samples with exponential growth and decrease growth rate until the support radius is "good enough"
+  while (std::isnan(this->_upperBound)) {
+    sampleRadius *= increaseSize;
+
+    auto [error, rcond] = solver.computeErrorEstimate(inputData, inputIds, sampleRadius);
+
+    PRECICE_INFO("RBF tuner sample: rad={:.4e}, err={:.4e}, 1/cond={:.4e}", sampleRadius, error, rcond);
+
+    if (rcond < 1e-12 || std::isinf(error)) {
+      PRECICE_CHECK(sampleRadius != this->_lowerBound, "Parameter tuning failed in first iteration using support-radius={}", sampleRadius);
```
davidscn: Is this something the user can configure? The initial guess? As a user: what to do when hitting this error message? Could you put this into the error message?

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,88 @@
...
+  template <typename IndexContainer>
+  std::tuple<double, double> optimize(const Solver &solver, const IndexContainer &inputIds, const Eigen::VectorXd &inputData);
+
+  bool lastSampleWasOptimum() const;
+};
+
+template <typename Solver>
+RBFParameterTuner<Solver>::RBFParameterTuner(const mesh::Mesh &inputMesh)
+{
+  constexpr bool radiusRBF = RadiusInitialization<typename Solver::BASIS_FUNCTION_T>::isAvailable();
```
davidscn: Isnt this exactly what you get with `hasCompactSupport()`. The Gaussian in preCICE is a compactly supported RBF. THe error message is good though, as this might not be obvious.

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,88 @@
...
+  PRECICE_ASSERT(radiusRBF, "RBF is not supported by this optimizer, as it does not accept a support-radius."
```
davidscn:
```suggestion
  PRECICE_CHECK(radiusRBF, "RBF is not supported by this optimizer, as it does not accept a support-radius."
```
unless we dont check for this already in the MappingCOnfiguration.

src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -94,27 +109,25 @@ class RadialBasisFctSolver {
 
   bool                computeCrossValidation = false;
   std::array<bool, 3> _localActiveAxis;
+
+  mutable Eigen::MatrixXd _kernelMatrix;
```
davidscn: Didnt we agree to _not_ store the decomposition and the kernel matrix. Those are heavy in memory. Also, the excessive use of `mutable` here makes the concept of `const` in this class a bit questionable. Isnt there any other solution? The `tuner` might be mutable, since the `solve` is const

src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -449,6 +451,30 @@ Eigen::VectorXd RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::solveConsistent(E
     inputData -= (_matrixQ * polynomialContribution);
   }
 
+  if (_tuningConfig.autotuneShape) {
+    // @todo: Data dimension not considered
```
davidscn: ?

tests/serial/rbf-parameter-tuning/helpers.cpp
```diff
@@ -0,0 +1,126 @@
...
+    interfaceA.finalize();
+
+  } else {
+    Participant interfaceB("SolverTwo", configFile, 0, 1);
+    auto [idsB, meshB] = generateMeshTwo(interfaceB, meshBID);
+    interfaceB.initialize();
+
+    std::array<double, n * dim> expectedValues = evaluateFunction(meshB, tA);
+
+    int it = 0;
```
davidscn:
```suggestion
```

tests/serial/rbf-parameter-tuning/GaussianSupportRadiusPolynomial.xml
```diff
@@ -0,0 +1,40 @@
...
+
+  <participant name="SolverOne">
+    <provide-mesh name="MeshOne" />
+    <write-data name="DataOne" mesh="MeshOne" />
+  </participant>
+
+  <participant name="SolverTwo">
+    <receive-mesh name="MeshOne" from="SolverOne" />
+    <provide-mesh name="MeshTwo" />
+    <mapping:rbf-global-direct
```
davidscn, line 21: The same solver backend is also involved in the `pu-rbf` solver, but the test doesnt cover this yet.

tests/serial/rbf-parameter-tuning/GaussianSupportRadiusPolynomial.xml
```diff
@@ -0,0 +1,40 @@
+<?xml version="1.0" encoding="UTF-8" ?>
+<precice-configuration>
+  <data:vector name="DataOne" />
```
davidscn, line 3: How about testing scalar data

tests/serial/rbf-parameter-tuning/GaussianShapeParameter.xml
```diff
@@ -0,0 +1,40 @@
...
+
+  <participant name="SolverTwo">
+    <receive-mesh name="MeshOne" from="SolverOne" />
+    <provide-mesh name="MeshTwo" />
+    <mapping:rbf-global-direct
+      direction="read"
+      from="MeshOne"
+      to="MeshTwo"
+      constraint="consistent"
+      polynomial="off">
```
davidscn: How about testing the optimization with polynomial?

davidscn (changes requested):

src/mapping/MathHelper.hpp
```diff
@@ -10,7 +12,8 @@ namespace precice::utils {
  * @param L the lower triangular matrix
  * @return Eigen::MatrixXd the inverse of L
  */
-inline Eigen::MatrixXd invertLowerTriangularBlockwise(const Eigen::MatrixXd &L)
+template <typename T>
+inline Eigen::Matrix<T, -1, -1> invertLowerTriangularBlockwise(const Eigen::Matrix<T, -1, -1> &L)
```
fsimonis: Did you maybe do experiments with mixed precision?
Currently, there is no reason to why this should be templatized.

src/mapping/MathHelper.hpp
```diff
@@ -66,4 +69,169 @@ inline Eigen::MatrixXd invertLowerTriangularBlockwise(const Eigen::MatrixXd &L)
...
+  // A^{-1} = R^{-1} Q^T
+  // A^{-1}_{ii} = sum_{k=1}^n R^{-1}_{ik} Q^{T}_{ki}
+
+  Eigen::VectorXd    inverseDiagonal;
+  const Eigen::Index n = decMatrixC.matrixR().cols();
+
+  // 1a: Compute the inverse of the lower triangular matrix L
+  // Solve R * Rinv = I
+  Eigen::MatrixXd R_inv = Eigen::MatrixXd::Identity(n, n);
+  decMatrixC.matrixR().template triangularView<Eigen::Upper>().solveInPlace(R_inv);
```
fsimonis, line 139: There is no dependency on another template. I think this isn't needed here. (untested)

```suggestion
  decMatrixC.matrixR().triangularView<Eigen::Upper>().solveInPlace(R_inv);
```

src/mapping/MathHelper.hpp
```diff
@@ -66,4 +69,169 @@ inline Eigen::MatrixXd invertLowerTriangularBlockwise(const Eigen::MatrixXd &L)
...
+inline double computeSquaredDifference(
+    const std::array<double, 3> &u,
+    std::array<double, 3>        v,
+    const std::array<bool, 3>   &activeAxis = {{true, true, true}})
+{
+  // Subtract the values and multiply out dead dimensions
+  for (unsigned int d = 0; d < v.size(); ++d) {
+    v[d] = (u[d] - v[d]) * static_cast<int>(activeAxis[d]);
+  }
+  // @todo: this can be replaced by std::hypot when moving to C++17
```
fsimonis: We have :wink:

src/mapping/MathHelper.hpp
```diff
@@ -66,4 +69,169 @@ inline Eigen::MatrixXd invertLowerTriangularBlockwise(const Eigen::MatrixXd &L)
...
+  return std::accumulate(v.begin(), v.end(), static_cast<double>(0.), [](auto &res, auto &val) { return res + val * val; });
```
fsimonis: No need for references here:

```suggestion
  return std::accumulate(v.begin(), v.end(), static_cast<double>(0.), [](auto res, auto val) { return res + val * val; });
```

src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -70,12 +79,18 @@ class RadialBasisFctSolver {
...
 private:
   mutable precice::logging::Logger _log{"mapping::RadialBasisFctSolver"};
 
+  template <typename IndexConatiner>
+  DecompositionType buildKernelDecomposition(const IndexConatiner &inputIds, double parameter) const;
+
   double evaluateRippaLOOCVerror(const Eigen::VectorXd &lambda) const;
   /// Decomposition of the interpolation matrix
-  DecompositionType _decMatrixC;
+  mutable DecompositionType _decMatrixC;
```
fsimonis: Using `mutable` is an indication that the constness of some functions are incorrect.

mutable isn't an option for something that is definitely part of the object state. Same for the other members below.

src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -437,9 +420,28 @@ Eigen::VectorXd RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::solveConservative
...
+    if constexpr (RadiusInitialization<RADIAL_BASIS_FUNCTION_T>::requiresConversion()) {
+      parameter = RadiusInitialization<RADIAL_BASIS_FUNCTION_T>::transformRadiusToShape(sampleRadius);
+    }
+    RADIAL_BASIS_FUNCTION_T kernel(parameter);
+    expr = [kernel](double x) { return kernel.evaluate(x); };
+  } else {
+    PRECICE_ASSERT(false, "Selected RBF does not support a radius.");
+    expr = [](double x) { return x; };
+  }
+  return matrixExpr.unaryExpr(expr);
```
fsimonis: Let's turn the entire else branch into a static_assert and get rid of `std::function<...> expr;`.

src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -449,6 +451,30 @@ Eigen::VectorXd RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::solveConsistent(E
...
+        auto [radius, error] = _tuner->optimize(*this, inputIDs, inputData);
+        _optimizedRadius     = radius;
+        _lastTuningError     = error;
+        if (!_tuner->lastSampleWasOptimum()) {
+          _decMatrixC = buildKernelDecomposition(inputIDs, _optimizedRadius);
+        }
+        PRECICE_INFO("Using: radius={}, LOOCV={}", _optimizedRadius, error);
+      } else {
+        PRECICE_ASSERT(false, "Not supported.");
+      }
```
fsimonis: This else branch should also be a `static_assert`

src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -467,6 +505,41 @@ Eigen::VectorXd RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::solveConsistent(E
...
+    RADIAL_BASIS_FUNCTION_T kernel(parameter);
+    fillKernelEntries(_kernelMatrix, kernel, _inputMesh, inputIds, _activeAxis);
+
+    if constexpr (RADIAL_BASIS_FUNCTION_T::isStrictlyPositiveDefinite()) {
+      return Eigen::LLT<Eigen::MatrixXd>(_kernelMatrix); // consider Eigen::LLT<Eigen::Ref<Eigen::MatrixXd>>
+    } else {
+      return Eigen::ColPivHouseholderQR<Eigen::MatrixXd>(_kernelMatrix); // consider Eigen::ColPivHouseholderQR<Eigen::Ref<Eigen::MatrixXd>>
+    }
+  }
+  PRECICE_UNREACHABLE("RBF does not support a radius-initialization and was still used to instantiate an optimizer.");
```
fsimonis: This should also be turned into a static_assert. Then you can remove the first condition of the previous branch.

src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -467,6 +505,41 @@ Eigen::VectorXd RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::solveConsistent(E
...
+}
+
+template <typename RADIAL_BASIS_FUNCTION_T>
+template <typename IndexConatiner>
+std::tuple<double, double> RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::computeErrorEstimate(const Eigen::VectorXd &inputData, const IndexConatiner &inputIds, double parameter) const
```
fsimonis: Let's use a struct as return type here. Returning pair or tuple is bad practice as it erases the information on what these values represent.

src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -467,6 +505,41 @@ Eigen::VectorXd RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::solveConsistent(E
...
+{
+  _decMatrixC = buildKernelDecomposition(inputIds, parameter);
+
+  if (_decMatrixC.info() != Eigen::ComputationInfo::Success) {
+    return {std::numeric_limits<double>::infinity(), 0};
```
fsimonis: returning infinity can be tricky as some compilers optimize infinity out.

If this is some kind of error condition, prefer to use a std:optional instead.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -232,6 +232,9 @@ MappingConfiguration::MappingConfiguration(
   // auto attrMaxIterations = makeXMLAttribute(ATTR_MAX_ITERATIONS, 1e6)
   //                              .setDocumentation("Maximum number of iterations of the solver");
 
+  auto attrTunerInterval = makeXMLAttribute(ATTR_TUNER_INTERVAL, "")
+                               .setDocumentation("Support radius optimization every N solver calls");
+
```
fsimonis: This should move down to attrSuportRadius

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -309,10 +312,10 @@ MappingConfiguration::MappingConfiguration(
       XMLTag{*this, RBF_CPOLYNOMIAL_C8, once, SUBTAG_BASIS_FUNCTION}.setDocumentation("Wendland C8 function"),
       XMLTag{*this, RBF_CTPS_C2, once, SUBTAG_BASIS_FUNCTION}.setDocumentation("Compact thin-plate-spline C2")};
 
-  auto attrSupportRadius = XMLAttribute<double>(ATTR_SUPPORT_RADIUS)
+  auto attrSupportRadius = XMLAttribute<std::string>(ATTR_SUPPORT_RADIUS)
```
fsimonis, line 348: Changing the type will make the documentation confusing, please update that too.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -336,8 +339,8 @@ MappingConfiguration::MappingConfiguration(
   std::list<XMLTag> GaussRBF{
       XMLTag{*this, RBF_GAUSSIAN, once, SUBTAG_BASIS_FUNCTION}.setDocumentation("Gaussian basis function accepting a support radius or a shape parameter.")};
   attrShapeParam.setDefaultValue(std::numeric_limits<double>::quiet_NaN());
-  attrSupportRadius.setDefaultValue(std::numeric_limits<double>::quiet_NaN());
-  addAttributes(GaussRBF, {attrShapeParam, attrSupportRadius});
+  attrSupportRadius.setDefaultValue("NaN");
```
fsimonis: The sentinel value should probably be `""` here and then use `.empty()` to figure out if an argument was passed or not.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -458,20 +461,55 @@ void MappingConfiguration::xmlTagCallback(
                                                             "from mesh \"{}\" to mesh \"{}\".",
                   _mappings.back().fromMesh->getName(), _mappings.back().toMesh->getName());
 
-    std::string basisFctName   = tag.getName();
-    double      supportRadius  = tag.getDoubleAttributeValue(ATTR_SUPPORT_RADIUS, 0.);
-    double      shapeParameter = tag.getDoubleAttributeValue(ATTR_SHAPE_PARAM, 0.);
+    std::string basisFctName     = tag.getName();
+    std::string supportRadiusTag = tag.getStringAttributeValue(ATTR_SUPPORT_RADIUS, "NaN");
```
fsimonis: Similar here, let's use `""` as sentinel value.

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -458,20 +461,55 @@ void MappingConfiguration::xmlTagCallback(
...
+          _autotunerConfig.iterationInterval = AutotuningParams::UNINITIALIZED;
+        }
+      }
+      PRECICE_CHECK(_autotunerConfig.iterationInterval > 0, "\"{}\" must be a positive integer or \"once\"", ATTR_TUNER_INTERVAL);
+    }
+
+    // Parse support radius if it should not be tuned. Otherwise, set it to some finite value to mark it as "set".
+    if (!_autotunerConfig.autotuneShape) {
+      try {
+        supportRadius = std::stod(supportRadiusTag);
```
fsimonis: This is not locale-safe, take a look at `xml/ValueParser.hpp` and reuse these functions please.

As an explanation: `1.2` will parse as `1` if your local system uses a locale that uses comma as decimal "point".

src/mapping/config/MappingConfiguration.hpp
```diff
@@ -98,6 +98,19 @@ class MappingConfiguration : public xml::XMLTag::Listener {
     bool                basisFunctionDefined = false;
   };
 
+  struct AutotuningParams {
+    enum class Type {
+      BISECTION,
+      BAYES_OPTIMIZATION
+    };
```
fsimonis: There currently is no bayes version, right?

src/mapping/config/MappingConfiguration.hpp
```diff
@@ -98,6 +98,19 @@ class MappingConfiguration : public xml::XMLTag::Listener {
...
+    static constexpr int OPTIMIZE_ONLY_ONCE = 0;
+    static constexpr int UNINITIALIZED      = -1;
+
+    Type optimizationType  = Type::BISECTION; // not yet used
```
fsimonis: Either the comment is outdated or this can be removed.

src/mapping/impl/RBFParameterTuner.hpp
fsimonis: This is available at compile time, right? So this could be a static assert.

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,88 @@
...
+
+  _lastSampleWasOptimum = false;
+}
+
+template <typename Solver>
+double RBFParameterTuner<Solver>::estimateMeshResolution(const mesh::Mesh &inputMesh)
+{
+  constexpr int sampleSize = 3;
+
+  const size_t       i0 = inputMesh.vertices().size() / 2;
```
fsimonis: Either use `std::size_t` or auto. Note that vector::size_type doesn't need to be size_t.

```suggestion
  const auto         i0 = inputMesh.vertices().size() / 2;
```

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,88 @@
...
+  const mesh::Vertex x0 = inputMesh.vertices().at(i0);
+
+  const std::vector<int> matches = inputMesh.index().getClosestVertices(x0.getCoords(), sampleSize);
```
fsimonis:
```suggestion
  const std::vector<VertexID> matches = inputMesh.index().getClosestVertices(x0.getCoords(), sampleSize);
```

src/query/Index.hpp
```diff
@@ -72,7 +72,7 @@ class Index {
   VertexMatch getClosestVertex(const Eigen::VectorXd &sourceCoord);
 
   /// Get n number of closest vertices to the given vertex
-  std::vector<VertexID> getClosestVertices(const Eigen::VectorXd &sourceCoord, int n);
+  std::vector<VertexID> getClosestVertices(const Eigen::VectorXd &sourceCoord, int n) const;
```
fsimonis: This is wrong. This can trigger the creation of the rtree.

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,88 @@
...
+  }
+
+  Sample()
+      : pos(std::numeric_limits<double>::quiet_NaN()), error(std::numeric_limits<double>::infinity())
+  {
+  }
+};
+
+template <typename Solver>
+class RBFParameterTuner {
```
fsimonis: This complete class doesn't need to be templatized.
The only thing that needs a template parameter is the check if the radiusinitialization is available. This doesn't depend on anything used here and should be moved out.

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,88 @@
...
+  Eigen::Index _inSize;
+
+  double _lowerBound;
+  double _upperBound;
+  bool   _lastSampleWasOptimum;
+
+  static double estimateMeshResolution(const mesh::Mesh &inputMesh);
+
+public:
+  virtual ~RBFParameterTuner() = default;
```
fsimonis: This doesn't look right in this context. Why make this base virtual?

src/mapping/impl/BisectionRBFTuner.hpp
```diff
@@ -0,0 +1,118 @@
...
+
+#include <mapping/MathHelper.hpp>
+#include <mapping/config/MappingConfigurationTypes.hpp>
+#include <mesh/Mesh.hpp>
+#include "mapping/impl/RBFParameterTuner.hpp"
+
+namespace precice::mapping {
+
+template <typename Solver>
+class BisectionRBFTuner : public RBFParameterTuner<Solver> {
```
fsimonis: I look at this code now 3 times and think that I get a vague grasp on what you are trying to achieve here. I have a feeling that you attempted CRTP here.

Assuming that there used to be another tuner, this still isn't CRTP. It's a refactoring of common classes into a base. This is fine, but the base shouldn't be templatized here.
In the current state, the RBFParameterTuner can just be merged into this class.

The templatization of this class is also not necessary. Some places require an RBF or a Solver. If you want to ensure consistent RBF usage, then you can use the RBF as a template parameter, then templatize the functions that need a solver type and ensure that it uses the same RBF.

In general the BisectionRBFTuner and RBFParameterTuner need simplification.

benchmarks/rbf-assembly-kernels.cpp
FabioTucciarone: I needed a pointer to the input mesh as a member variable in `RadialBasisFctSolver.hpp` and modified the `buildMatrixA()` method to take in a pointer. I could also revert this change and pass the mesh with `*inputMesh` in the Solver class.

davidscn: It's less about the general change, but rather the place: did you use and compile the benchmarks at all?

src/mapping/impl/BasisFunctions.hpp
FabioTucciarone: I thaught keeping it general might make the code in `RadialBasisFctSolver<RBF_T>::buildKernelDecomposition` and `RadialBasisFctSolver<RBF_T>::solveConsistent` more readable:

```c++
if constexpr (RadiusInitialization<RADIAL_BASIS_FUNCTION_T>::isAvailable()) { ... }
```

```c++
if constexpr (RadiusInitialization<RADIAL_BASIS_FUNCTION_T>::isAvailable()) {
    if constexpr (RadiusInitialization<RADIAL_BASIS_FUNCTION_T>::requiresConversion()) {
      parameter = RadiusInitialization<RADIAL_BASIS_FUNCTION_T>::transformRadiusToShape(parameter);
    }
    ...
}
```
In case you plan to add other conversions or new RBFs that don't support a radius.

But ultimately I wasn't sure what the best approach here would be.

src/mapping/impl/BisectionRBFTuner.hpp
FabioTucciarone: I will clarify that. In regard to the initial guess, should I add something like a parameter range e.g.: [0.1, 1] to the configuration such that the user can specify where the bisection algorithm starts searching?

src/mapping/impl/RBFParameterTuner.hpp
FabioTucciarone: As mentioned in a previous reply, I wasn't quite sure how to treat the RBFs ie. if i should make the tests specific to the Gaussian or mor general so as to be more readable/expandable.
This might have caused some confusions and inconsistencies in the code.

src/mapping/RadialBasisFctSolver.hpp
FabioTucciarone: - mutable: This is why I originally stored the matrix and calculated the error approximation in the tuner, which made this less confusing.
- _"Didnt we agree to not store the decomposition and the kernel matrix."_: We shortly discussed it, but i wasn't sure how to efficiently implement that, since I need to recalculate the coefficients and decomposition for each parameter choise.

src/mapping/MathHelper.hpp
FabioTucciarone: I intended to, but haven't. But as far as I remember, in this case, I just moved David's code to the `MathHelper.hpp`

src/mapping/RadialBasisFctSolver.hpp
FabioTucciarone: I wasn't aware of that. In some cases using infinity was useful for comparisons. I'll look in to that.

src/mapping/config/MappingConfiguration.hpp
FabioTucciarone: Yes, this should be removed

src/mapping/RadialBasisFctSolver.hpp
FabioTucciarone: I'm aware of this problem, which is why I originallly only marked the `tuner` as mutable, but David sugested that I put the error approximation method into the solver and since the optimization takes place in `solveConsistent` I couldn't figure out a better way to do that.
I'll discuss what approach might be sensible here with David.

benchmarks/rbf-assembly-kernels.cpp
FabioTucciarone: I compiled, but didn't use it. I did all my tests using aste and heaptrack.

tests/serial/rbf-parameter-tuning/GaussianSupportRadiusPolynomial.xml
FabioTucciarone: I only changed the Solver, so if it's correct for vectorial data it should also be correct for scalar data.
Since we thought about expandinng the tuner and solver to be able to handle vectorial data, I thought this would be the more general testing approach.

src/mapping/RadialBasisFctSolver.hpp
FabioTucciarone: Then the code doesn't compile, since the configuration of the basis functions and the tuner happen at runtime.

src/mapping/impl/BisectionRBFTuner.hpp
FabioTucciarone: Since we don't include the Bayes optimization, I have removed the base class for now.

src/mapping/RadialBasisFctSolver.hpp
FabioTucciarone: The optimizer can currently only be confgured  to run every n `solveConsistent` calls instead of every n `mapData` calls.
This is solved by not processing the components of vector data individuallly: https://github.com/precice/precice/pull/2389

src/mapping/MathHelper.hpp
FabioTucciarone: I can remove them. This is just code I moved to the MathHelper
https://github.com/precice/precice/blob/e6e1515237c5966556a4bc8a9313192640509fce/src/mapping/RadialBasisFctSolver.hpp#L111-L112

src/mapping/MathHelper.hpp
FabioTucciarone: You have replaced it by std::hypot? Or should I do that?
https://github.com/precice/precice/blob/e6e1515237c5966556a4bc8a9313192640509fce/src/mapping/RadialBasisFctSolver.hpp#L111-L112

src/mapping/RadialBasisFctSolver.hpp
FabioTucciarone: I removed the `mutable` keyword, and spoke with David.
He suggested either moving the problem / tuning to the `RadialBasisFctMapping` class or to use a `const_cast`.

Right now the code is:
https://github.com/FabioTucciarone/precice/blob/74e28cbf2993bd64f7ea33f8df18cf2de35e931a/src/mapping/RadialBasisFctSolver.hpp#L466-L467
Which I'm not sure about. 

What do you think @fsimonis?

src/mapping/MathHelper.hpp
davidscn: We have refers to C++17, but hypot cannot easily integrate the dead axis and you have to write down all three args as well, so I dont think this saves us a lot here

src/mapping/MathHelper.hpp
davidscn: I dont think it really matters here. Why removing it @fsimonis?

src/mapping/config/MappingConfiguration.cpp
```diff
@@ -336,8 +339,8 @@ MappingConfiguration::MappingConfiguration(
   std::list<XMLTag> GaussRBF{
       XMLTag{*this, RBF_GAUSSIAN, once, SUBTAG_BASIS_FUNCTION}.setDocumentation("Gaussian basis function accepting a support radius or a shape parameter.")};
   attrShapeParam.setDefaultValue(std::numeric_limits<double>::quiet_NaN());
-  attrSupportRadius.setDefaultValue(std::numeric_limits<double>::quiet_NaN());
-  addAttributes(GaussRBF, {attrShapeParam, attrSupportRadius});
+  attrSupportRadius.setDefaultValue(""); // TODO REPLACED "NaN"
```
davidscn: What does this TODO actually asks anyone to do?

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,196 @@
...
+  Sample _currentOptimum;
+
+  static double estimateMeshResolution(mesh::Mesh &inputMesh);
+
+public:
+  ~RBFParameterTuner() = default;
+  explicit RBFParameterTuner(mesh::Mesh &inputMesh);
+
+  template <typename IndexContainer, typename Solver>
+  Sample optimize(Solver &solver, const IndexContainer &inputIds, const Eigen::VectorXd &inputData);
```
davidscn: Doxygen docs missing

src/mapping/impl/RBFParameterTuner.hpp
davidscn, line 1: doxygen docs missing for this file

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,196 @@
+#pragma once
+
+#include <mapping/MathHelper.hpp>
+#include <mapping/impl/BasisFunctions.hpp>
+#include <mesh/Mesh.hpp>
+
+namespace precice::mapping {
+
+struct RBFErrorEstimate {
+  double error;
```
davidscn, line 15:
```suggestion
public:
    double error;
```

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,196 @@
...
+
+template <typename IndexContainer, typename Solver>
+Sample RBFParameterTuner::optimize(Solver &solver, const IndexContainer &inputIds, const Eigen::VectorXd &inputData)
+{
+  constexpr bool radiusRBF = Solver::BASIS_FUNCTION_T::hasCompactSupport();
+  static_assert(radiusRBF, "RBF is not supported by this optimizer, as it does not accept a support-radius."
+                           "Currently supported: Compactly supported RBFs and Gaussians.");
+
+  // TODO: maybe make dependent on domain extend
+  constexpr double increaseSize = 10;
```
davidscn:
```suggestion
  // TODO: maybe make dependent on domain extent
  constexpr double increaseSize = 10;
```
what's an increase size? needs docs.

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,196 @@
...
+  double           sampleRadius = _lowerBound;
+
+  _lastSampleWasOptimum = false;
+
+  PRECICE_INFO("Starting optimization with lower bound = {:.4e}, upper bound = {:.4e}", _lowerBound, _upperBound);
```
davidscn: This and other INFO prints will be confusing, if multiple mappings and participants are configured: to which mapping does it then refer?

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,196 @@
...
+    sampleRadius *= increaseSize;
+
+    solver.rebuildKernelDecomposition(inputIds, sampleRadius);
+    RBFErrorEstimate estimate = solver.computeErrorEstimate(inputData, inputIds);
+
+    PRECICE_INFO("RBF tuner sample: rad={:.4e}, err={:.4e}, 1/cond={:.4e}", sampleRadius, estimate.error, estimate.rcond);
+
+    if (estimate.rcond < RCOND_TOLERANCE || estimate.error >= std::numeric_limits<double>::max()) {
+      PRECICE_CHECK(sampleRadius != _lowerBound, "Parameter tuning failed in first iteration using support-radius={}."
+                                                 "Try using a different basis function, or select a smaller support radius manually.",
```
davidscn:
```suggestion
                                                 "Try using a different basis function, or manually select a smaller support radius.",
```
Smaller than what?

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,196 @@
...
+  while (std::isnan(_upperBound)) {
+    sampleRadius *= increaseSize;
+
+    solver.rebuildKernelDecomposition(inputIds, sampleRadius);
+    RBFErrorEstimate estimate = solver.computeErrorEstimate(inputData, inputIds);
+
+    PRECICE_INFO("RBF tuner sample: rad={:.4e}, err={:.4e}, 1/cond={:.4e}", sampleRadius, estimate.error, estimate.rcond);
+
+    if (estimate.rcond < RCOND_TOLERANCE || estimate.error >= std::numeric_limits<double>::max()) {
+      PRECICE_CHECK(sampleRadius != _lowerBound, "Parameter tuning failed in first iteration using support-radius={}."
```
davidscn:
```suggestion
      PRECICE_CHECK(sampleRadius != _lowerBound, "RBF parameter tuning diverged in first iteration using support-radius={}."
```
?

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,196 @@
...
+{
+  size_t sampleSize = std::min((size_t) 3, inputMesh.vertices().size());
+
+  const auto         i0 = inputMesh.vertices().size() / 2;
+  const mesh::Vertex x0 = inputMesh.vertices().at(i0);
+
+  const std::vector<VertexID> matches = inputMesh.index().getClosestVertices(x0.getCoords(), sampleSize);
+
+  double h = 0;
+  for (int i = 0; i < sampleSize; i++) {
```
davidscn: comparison of integer expressions of different signedness: ‘int’ and ‘size_t’ {aka ‘long unsigned int’} [-Wsign-compare]

src/mapping/impl/SphericalVertexCluster.hpp
```diff
@@ -177,8 +179,10 @@ SphericalVertexCluster<RADIAL_BASIS_FUNCTION_T>::SphericalVertexCluster(
   // mapping in this cluster as computed (mostly for debugging purpose)
   std::vector<bool>         deadAxis(inputMesh->getDimensions(), false);
   precice::profiling::Event e("map.pou.computeMapping.rbfSolver");
-  _rbfSolver          = RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>{function, *inputMesh.get(), _inputIDs, *outputMesh.get(), _outputIDs, deadAxis, _polynomial};
+  _rbfSolver          = RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>(function, inputMesh, _inputIDs, outputMesh, _outputIDs, deadAxis, _polynomial, rbfOptional);
   _hasComputedMapping = true;
+
+  // warning: use of function template name with no prior declaration in function call with explicit template arguments is a C++20 extension [-Wc++20-extensions]
```
davidscn: ?

src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -24,9 +27,9 @@ namespace precice::mapping {
 template <typename RADIAL_BASIS_FUNCTION_T>
 class RadialBasisFctSolver {
 public:
-  using DecompositionType = std::conditional_t<RADIAL_BASIS_FUNCTION_T::isStrictlyPositiveDefinite(), Eigen::LLT<Eigen::MatrixXd>, Eigen::ColPivHouseholderQR<Eigen::MatrixXd>>;
+  using DecompositionType = std::conditional_t<RADIAL_BASIS_FUNCTION_T::isStrictlyPositiveDefinite(), Eigen::LLT<Eigen::Ref<Eigen::MatrixXd>>, Eigen::ColPivHouseholderQR<Eigen::Ref<Eigen::MatrixXd>>>;
```
davidscn, line 30: Needs an extensive comment on the type and implications!

src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -37,16 +40,26 @@ class RadialBasisFctSolver {
...
-  RadialBasisFctSolver(RADIAL_BASIS_FUNCTION_T basisFunction, const mesh::Mesh &inputMesh, const IndexContainer &inputIDs,
-                       const mesh::Mesh &outputMesh, const IndexContainer &outputIDs, std::vector<bool> deadAxis, Polynomial polynomial);
+  RadialBasisFctSolver(RADIAL_BASIS_FUNCTION_T basisFunction, mesh::PtrMesh inputMesh, const IndexContainer &inputIDs,
+                       mesh::PtrMesh outputMesh, const IndexContainer &outputIDs, std::vector<bool> deadAxis, Polynomial polynomial, MappingConfiguration::AutotuningParams rbfConfig);
 
-  /// Maps the given input data
-  Eigen::VectorXd solveConsistent(Eigen::VectorXd &inputData, Polynomial polynomial) const;
+  template <typename IndexContainer>
+  RadialBasisFctSolver(RADIAL_BASIS_FUNCTION_T basisFunction, mesh::PtrMesh inputMesh, const IndexContainer &inputIDs,
+                       mesh::PtrMesh outputMesh, const IndexContainer &outputIDs, std::vector<bool> deadAxis, Polynomial polynomial); // TODO: refactor and adapt for PUM
```
davidscn: TODO?

src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -70,12 +83,21 @@ class RadialBasisFctSolver {
...
   mutable precice::logging::Logger _log{"mapping::RadialBasisFctSolver"};
 
   double evaluateRippaLOOCVerror(const Eigen::VectorXd &lambda) const;
-  /// Decomposition of the interpolation matrix
-  DecompositionType _decMatrixC;
+
+  /// Matrix used for storing the kernel coefficients and the modified coefficients used by the decomposition.
+  Eigen::MatrixXd _decMatrixCoefficients;
+  /// Eigen in-place decomposition type of the interpolation matrix _decMatrixCoefficients. std::unique_ptr since Eigen:Ref is not resizable.
+  std::unique_ptr<DecompositionType> _decMatrixC;
```
davidscn, line 99: points to the same memory as `_decMatrixCoefficients`
doxygen can also reference other variables. Please do this here to make this crystal clear and also give a comment on why we actually want this change

src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -94,27 +116,23 @@ class RadialBasisFctSolver {
...
+
+  MappingConfiguration::AutotuningParams     _tuningConfig;
+  mutable std::unique_ptr<RBFParameterTuner> _tuner;
+
+  mutable int _iterSinceOptimization = 1;
+
+  Polynomial _polynomial;
+
+  std::array<bool, 3> _activeAxis;
+  mesh::PtrMesh       _inputMesh;
```
davidscn: Also add a comment why we need ownership of the input mesh and the consequences this brings

src/mapping/impl/RBFParameterTuner.hpp
FabioTucciarone: why?

davidscn: just making explicit, what's been done here

src/mapping/impl/RBFParameterTuner.hpp
FabioTucciarone: Since I cannot access `solver.input()->getName()` from there, I'm not sure how to improve that.
Should I just move them to the Solver class?

src/mapping/impl/RBFParameterTuner.hpp
FabioTucciarone: Smaller than the support radius mentioned in the line above. Is this unclear?

src/mapping/RadialBasisFctSolver.hpp
FabioTucciarone: I replaced the `mesh::PtrMesh` with a `std::weak_ptr` and throw an error in `RadialBasisFctMapping.h` if a parallel mapping is used. T think this also makes sens from a user perspective regarding the repartitioning/tagging of the meshes.

src/mapping/MathHelper.hpp
fsimonis: The rule is: If an argument is not modified and fits into a register, pass it by value.
Passing by reference is significantly more tricky for compilers to optimize, as the lifetime spans across potentially multiple function calls. Each call could change the value or even take its address and use it in another way.

Apart from that, parameters for the binary operation of std::accumulate either need to be passed by const ref or value. This is defined in the docs of std::accumulate.

FabioTucciarone: I removed the references

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,228 @@
...
+  // collect samples with exponential growth and decrease growth rate until the support radius is "good enough"
+  while (std::isnan(_upperBound)) {
+    sampleRadius *= increaseSize;
+
+    solver.rebuildKernelDecomposition(inMesh, inputIds, sampleRadius);
+    RBFErrorEstimate estimate = solver.computeErrorEstimate(inputData, inputIds);
+
+    PRECICE_INFO("RBF tuner sample: rad={:.4e}, err={:.4e}, 1/cond={:.4e}", sampleRadius, estimate.error, estimate.rcond);
+
+    if (estimate.rcond < RCOND_TOLERANCE || estimate.error >= std::numeric_limits<double>::max()) {
```
fsimonis: What does `estimate.error >= std::numeric_limits<double>::max()` mean? :sweat_smile:

What is more than the representable maximum?

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,228 @@
...
+
+  // error is numerically insignificant: Either zero data or fully explained by the polynomial
+  if (estimate.error < 1e-15) {
+    _lastSampleWasOptimum = true;
+    _currentOptimum       = lowerBound;
+    return lowerBound;
+  }
+
+  // collect samples with exponential growth and decrease growth rate until the support radius is "good enough"
+  while (std::isnan(_upperBound)) {
```
fsimonis: `std::isnan` depends on compiler settings. We should not introduce more reliance on this (and inf) if we ever want to support fast math modes.

Can you use some other way to signal that this hasn't been computed yet? Either a `bool` or `std::optional<double>`?

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,228 @@
...
+
+  const std::vector<VertexID> matches = inputMesh.index().getClosestVertices(x0.getCoords(), sampleSize);
+
+  double h = 0;
+  for (size_t i = 0; i < sampleSize; i++) {
+    const mesh::Vertex xi = inputMesh.vertices().at(matches.at(i));
+    h += std::sqrt(utils::computeSquaredDifference(xi.rawCoords(), x0.rawCoords()));
+  }
+  return h / sampleSize;
+}
```
fsimonis: Could you move this function to `src/mesh/Utils.hpp`, given that this is purely mesh-dependent?

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,228 @@
...
+
+  return _currentOptimum;
+}
+
+inline double RBFParameterTuner::estimateMeshResolution(mesh::Mesh &inputMesh)
+{
+  size_t sampleSize = std::min((size_t) 3, inputMesh.vertices().size());
+
+  const auto         i0 = inputMesh.vertices().size() / 2;
+  const mesh::Vertex x0 = inputMesh.vertices().at(i0);
```
fsimonis: Let's not copy the vertex here here.

```suggestion
  const mesh::Vertex& x0 = inputMesh.vertices().at(i0);
```

src/mapping/impl/RBFParameterTuner.hpp
```diff
@@ -0,0 +1,228 @@
...
+  mutable logging::Logger _log{"mapping::RBFParameterTuner"};
+
+  /// Factor by which the radius is allowed to change before stopping
+  static constexpr double POS_TOLERANCE = 1.5;
+  /// Factor by which the error is allowed to change before stopping
+  static constexpr double ERR_TOLERANCE = 0.5;
+  /// Which reciprocal condition number should be considered too low for a good support radius
+  static constexpr double RCOND_TOLERANCE = 1e-12;
+  /// Number of iterations during the "bisection" step. After finding initial samples
+  static constexpr int MAX_BISEC_ITERATIONS = 6;
```
fsimonis, line 49: Should these constants be configurable by the user?

src/mapping/impl/BasisFunctions.hpp
```diff
@@ -60,6 +60,11 @@ struct CompactSupportBase {
   {
     return true;
   }
+
+  static constexpr double requiresRadiusToShapeConversion()
```
fsimonis, line 66: Can you add a Doxygen description of what this is and why it is needed?

src/mapping/RadialBasisFctSolver.hpp
fsimonis: Is there actually a reason why `RBFSolver::solveConsistent/Conservative` need to be `const`, while `RadialBasisFctMapping::mapConservative()` is not?

It could be way cleaner to make these two function non-const. Alternative would be to create a AutoTunedSolver class which derives from the RBFSolver and overwrites the const solveConsistent with a non-const version. This could work as the Mappings use the RBFSolver as Template argument.

I think there is no more I can suggest here. @davidscn has a significantly better overview over the inner workings of the RBF mapping.

src/mapping/impl/RBFParameterTuner.hpp
fsimonis: You could 
1. gather this information and return it to the solver, who has the information to generate a statement.
2. store the mesh name in the constructor and use it here.

src/mapping/impl/RBFParameterTuner.hpp
FabioTucciarone: Ideally not, since the point of the tuner is to remove configuration complexity.
These are simply constants that produce good results without performing to many iterations.

src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -449,16 +456,60 @@ Eigen::MatrixXd RadialBasisFctSolver<RADIAL_BASIS_FUNCTION_T>::solveConsistent(E
...
+  Eigen::MatrixXd out;
+  if (_rbfTunerConfig.autotuneShape) {
+    if constexpr (RADIAL_BASIS_FUNCTION_T::hasCompactSupport()) {
+      const Eigen::Index outSize    = _matrixA.rows();
+      const Eigen::Index inSize     = inputData.size();
+      const Eigen::Index polyParams = _matrixA.cols() - inSize;
+
+      // not yet necessary: integrated polynomial is only supported for SPD functions
+      out = applyKernelToMatrix<RADIAL_BASIS_FUNCTION_T>(_matrixA.block(0, 0, outSize, inSize), _tuner->getLastOptimizedRadius()) * p.segment(0, inSize);
+      out += _matrixA.block(0, inSize, outSize, polyParams) * p.segment(inSize, polyParams);
```
davidscn: @FabioTucciarone after having #2389 merged, the `p.segment` access doesn't work anymore. I was actually assuming that this PR incorporates the matrix solver interface. Can you check that the merge commit 748c2af4bc55ee912b9cb7f98fa114e283aa46a6 is correct?
Should it then be something like
```suggestion
      out = applyKernelToMatrix<RADIAL_BASIS_FUNCTION_T>(_matrixA.block(0, 0, outSize, inSize), _tuner->getLastOptimizedRadius()) * p.block(0, p.rows(), 0, inSize);
      out += _matrixA.block(0, inSize, outSize, polyParams) * p.block(0, p.rows(), inSize, polyParams);
```
?

davidscn:
```suggestion
      out = applyKernelToMatrix<RADIAL_BASIS_FUNCTION_T>(_matrixA.block(0, 0, outSize, inSize), _tuner->getLastOptimizedRadius()) * p.topRows(inSize);
      out += _matrixA.block(0, inSize, outSize, polyParams) * p.middleRows(inSize, polyParams);
```

PR #2346, opened 2025-08-13
PU-RBF interpolation on GPUs or OpenMP

davidscn:
## Main changes of this PR

Adds a PU-RBF variant which runs on GPUs/OpenMP using Kokkos and Kokkos-kernels.

Quick summary:

- uses the batched infrastructure of Kokkos-kernels to process all PU clusters concurrently
- works for consistent mappings at the moment (mapData kernel for conservative mappings is still missing)
- may be used with distributed memory, i.e., potentially multi-GPU + MPI
- enables support for SYCL by fixing a recursive function pattern when computing the fixed pow
- makes Ginkgo an optional dependency for the PU-RBF configuration
- adds tests, but the CI will fail due to missing dependencies (Kokkos kernels)
- particularly useful for large vertex counts per rank

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/2346.md
+++ b/docs/changelog/2346.md
@@ -1,2 +1,2 @@
-- Added support for a batched PU-RBF solver for consistent mappings using Kokkos and Kokkos-Kernels. The solver can leverage OpenMP, CUDA, HIP, and SYCL as executors.
+- Added support for a batched PU-RBF solver for consistent mappings using Kokkos and Kokkos-Kernels, which can be enabled using the CMake option `PRECICE_FEATURE_KOKKOS_KERNELS_MAPPING`. The solver can leverage OpenMP, CUDA, HIP, and SYCL as executors.
 - Added compatibility with SYCL.


CMakeLists.txt
```diff
@@ -63,6 +63,7 @@ option(PRECICE_FEATURE_PYTHON_ACTIONS "Enable Python support for preCICE actions
 option(PRECICE_FEATURE_PROFILING_COMPRESSION "Enable profiling compression with LZMA." OFF)
 option(PRECICE_CONFIGURE_PACKAGE_GENERATION "Configure package generation." ON)
 option(PRECICE_FEATURE_GINKGO_MAPPING "Enable use of Ginkgo for accelerated data mapping." OFF)
+option(PRECICE_FEATURE_KOKKOS_KERNELS_MAPPING "Enable use of Kokkos kernels for accelerated PU-RBF data mapping." OFF)
```
fsimonis, line 66: This depends on `PRECICE_FEATURE_GINKGO_MAPPING`, right?

We should consider using https://cmake.org/cmake/help/latest/module/CMakeDependentOption.html#command:cmake_dependent_option here.

CMakeLists.txt
```diff
@@ -122,6 +123,16 @@ add_feature_info(PRECICE_FEATURE_PETSC_MAPPING PRECICE_FEATURE_PETSC_MAPPING
    Requires PRECICE_FEATURE_MPI_COMMUNICATION.
   ")
 
+  add_feature_info(PRECICE_FEATURE_KOKKOS_KERNELS_MAPPING PRECICE_FEATURE_KOKKOS_KERNELS_MAPPING
+  "Enables the Kokkos and Kokkos-kernels backend which supports the usage of GPUs or OpenMP for pum-rbf mappings.
+
+   The executor can be specified as a subtag of the mapping configuration.
+
+   This feature can be enabled/disabled by setting the PRECICE_FEATURE_GINKGO_MAPPING CMake option.
+   Requires Kokkos-kernels, Kokkos (v4.1 or higher) and Eigen 3.4.
```
fsimonis: These are already checked, not sure if there is a lot of value mentioning it again here.

CMakeLists.txt
```diff
@@ -307,20 +318,7 @@ endif()
...
-       set(HIP_PATH $ENV{HIP_PATH} CACHE PATH "Path to which HIP has been installed")
-     endif()
-    endif()
-
-    if(NOT DEFINED ROCM_PATH)
-     set(ROCM_PATH "/opt/rocm" CACHE PATH "Path to which ROCM has been installed")
-     set(ENV{ROCM_PATH} ${ROCM_PATH})
-    endif()
-
+    find_package(hip REQUIRED)
```
fsimonis: Who provides this non-standard package?

CMakeLists.txt
```diff
@@ -334,6 +332,32 @@ else()
...
+
+  find_package(KokkosKernels 4.1 REQUIRED)
+  message(STATUS "Found KokkosKernels version ${KokkosKernels_VERSION} at ${KokkosKernels_DIR}")
+
+  include(CMakeDependentOption)
+  # The kokkos config will be the same as above in case we enable both features
+  cmake_dependent_option(PRECICE_WITH_CUDA "Enable Kokkos Cuda support in preCICE" ON "PRECICE_FEATURE_KOKKOS_KERNELS_MAPPING;Kokkos_ENABLE_CUDA" OFF)
+  cmake_dependent_option(PRECICE_WITH_HIP "Enable Kokkos HIP support in preCICE" ON "PRECICE_FEATURE_KOKKOS_KERNELS_MAPPING;Kokkos_ENABLE_HIP" OFF)
+  cmake_dependent_option(PRECICE_WITH_SYCL "Enable Kokkos SYCL support in preCICE" ON "PRECICE_FEATURE_KOKKOS_KERNELS_MAPPING;Kokkos_ENABLE_SYCL" OFF)
+  cmake_dependent_option(PRECICE_WITH_OPENMP "Enable Kokkos OpenMP support in preCICE" ON "PRECICE_FEATURE_KOKKOS_KERNELS_MAPPING;Kokkos_ENABLE_OPENMP" OFF)
```
fsimonis: These have already been defined in the `PRECICE_FEATURE_GINKGO_MAPPING` I'm not 100% sure, but I think they conflict.

Can we combine the definition of both to reduce the size of the CMake? CMake 3.22 added support for a full `if(` style condition in the condition block of the dependant option.

CMakeLists.txt
```diff
@@ -334,6 +332,32 @@ else()
...
+
+  if(PRECICE_WITH_HIP)
+    include(CheckLanguage)
+    check_language(HIP)
+    enable_language(HIP)
+    set(CMAKE_HIP_STANDARD 17)
+    set(CMAKE_HIP_STANDARD_REQUIRED ON)
+    find_package(hip REQUIRED) # required to link against hip::device
+  endif()
```
fsimonis: We have the exact code a bit further above. Let's combine them to keep the CMakeLists short.

CMakeLists.txt
```diff
@@ -463,45 +487,77 @@ else()
...
+  set_target_properties(preciceCore PROPERTIES CXX_COMPILER_LAUNCHER "")
   if(PRECICE_WITH_CUDA)
     target_compile_definitions(preciceCore PUBLIC PRECICE_WITH_CUDA)
     # Required to make nvcc happy
     target_compile_definitions(preciceCore PRIVATE BOOST_PP_VARIADICS=1)
     target_compile_definitions(preciceCore PRIVATE BOOST_ASIO_USE_TS_EXECUTOR_AS_DEFAULT=1)
+    message(STATUS "Enabled CUDA support for preCICE via Kokkos.")
+  endif()
+  if(PRECICE_WITH_HIP)
+    target_compile_definitions(preciceCore PUBLIC PRECICE_WITH_HIP)
```
fsimonis, line 501: Both features use the same internal defines `PRECICE_WITH_HIP` etc. This will lead to issues. Either we need to unify them by enabling/disabling both features at the same time or we need to use different defines in the code.

fsimonis (commented):
While the duplication of the options are not pretty, I think the duplicate use of the compile definition will lead to problems when only one feature is enabled/disabled. I think it would make more sense to either unify the feature toggle, or to unify some common configurations in the CMakeLists.

CMakeLists.txt
davidscn: No it doesnt

CMakeLists.txt
davidscn: What do you mean by already checked? By CMake?

CMakeLists.txt
davidscn: The installation of HIP?

CMakeLists.txt
davidscn: Do you have an example scenario in mind? I think I tested all combinations of on/off with Ginkgo/Kokkos

CMakeLists.txt
davidscn: I thought it gives the CMakeLists a more logical flow to read through. Since Ginkgo also needs the hipblas libraries. it will then become an extra if...if below, but can certainly do that

CMakeLists.txt
davidscn: Can you explain the issue? The define is mostly used as a PP define for example to enable or disable tests.

CMakeLists.txt
fsimonis: Exactly. CMake already checks these versions. So a user that enabled this feature should see an error in the configuration stage.

Such comments get outdated easily. I would remove them, but its up to you.

CMakeLists.txt
fsimonis: > I think I tested all combinations of on/off with Ginkgo/Kokkos

That's the scenario I have in mind. ON OFF and OFF ON.

So no conflict

CMakeLists.txt
davidscn: I see your point, removed it. Should probably put it on the website then

CMakeLists.txt
fsimonis: > Can you explain the issue?

I missed this define

```cmake
  target_compile_definitions(preciceCore PUBLIC PRECICE_NO_KOKKOS_KERNELS)
```

so `This will lead to issues.` is wrong as long as both the Kokkos kernels and Ginkgo mapping don't conflict at some point for given `PRECICE_WITH_X`.

CMakeLists.txt
fsimonis: This is the only part I really think a restructure would help a lot, especially to maintain this in the long run.

My suggestion is to check step by step:
* Check kokkos if either kokkos kernels or ginkgo mappings are activated. Set options for Cuda Hip etc and centrally handle their activation.
* Find ginkgo mappings if required (If needed display a note that SYCL isn't supported for the ginkgo mappings)
* Find kokkos kernels if required

Then try to use the same structure when modifying the `preciceCore` target.
This should increase the readability significantly.

davidscn: For `preciceCore`, that's already the case: I first have the branch affecting both features, and then the specializations. Note that some backend-specific choices are not the same for both features. Let me see what I can do here

davidscn: Done in [0087347](https://github.com/precice/precice/pull/2346/commits/008734784c11e0c68efd101538a5e3ca4af92d40)

is this what you had in mind?

fsimonis: Yes. Thanks!

fsimonis (approved):
I only reviewed the CMake perspective though

docs/changelog/2346.md
```diff
@@ -0,0 +1,2 @@
+- Added support for a batched PU-RBF solver for consistent mappings using Kokkos and Kokkos-Kernels. The solver can leverage OpenMP, CUDA, HIP, and SYCL as executors.
```
davidscn:
```suggestion
- Added support for a batched PU-RBF solver for consistent mappings using Kokkos and Kokkos-Kernels, which can be enabled using the CMake option `PRECICE_FEATURE_KOKKOS_KERNELS_MAPPING`. The solver can leverage OpenMP, CUDA, HIP, and SYCL as executors.
```

davidscn:
> I only reviewed the CMake perspective though

Feel free to review more, if you like. The majority of changes is in the kernel implementation, which are (1) kokkos-specific and (2) already quite tuned. But maybe you want to check the integration into the main PU-RBF class and the BatchedSolver overflow detection, where I couldn't find an optimal solution.

PR #2349, opened 2025-08-20
Add documentation for QR3 filter

Fujikawas:
## Main changes of this PR
Added documentation for QR3 filter

## Author's checklist

* [ ] ~I added a changelog file with `make changelog` if there are user-observable changes since the last release.~
* [ ] ~I added a test to cover the proposed changes in our test suite.~
* [ ] ~For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).~
* [ ] ~I stuck to CMake version 3.22.1.~
* [ ] ~I ran the systemtests by adding the label `trigger-system-tests` (may be skipped if minor change)~

## Reviewers' checklist

### DIFF
--- a/src/acceleration/config/AccelerationConfiguration.cpp
+++ b/src/acceleration/config/AccelerationConfiguration.cpp
@@ -381,14 +381,14 @@ void AccelerationConfiguration::addCommonIQNSubtags(xml::XMLTag &tag)
   XMLTag tagFilter(*this, TAG_FILTER, XMLTag::OCCUR_NOT_OR_ONCE);
   tagFilter.setDocumentation("Type of filtering technique that is used to "
                              "maintain good conditioning in the least-squares system. Possible filters:\n"
-                             " - `QR1-filter`: updateQR-dec with (relative) test \\\\(R(i,i) < \\epsilon *\\lVert R\\rVert_F\\\\)\n"
-                             " - `QR1_absolute-filter`: updateQR-dec with (absolute) test \\\\(R(i, i) < \\epsilon\\\\)\n"
+                             " - `QR1-filter`: update QR-dec with (relative) test \\\\(R(i,i) < \\epsilon *\\lVert R\\rVert_F\\\\)\n"
+                             " - `QR1_absolute-filter`: update QR-dec with (absolute) test \\\\(R(i, i) < \\epsilon\\\\)\n"
                              " - `QR2-filter`: en-block QR-dec with test \\\\(\\lVert v_\\text{orth} \\rVert_2 < \\epsilon * \\lVert v \\rVert_2\\\\)\n\n"
+                             " - `QR3-filter`: update QR-dec only when the pre-scaling weights have changed or there is one or more columns are to be removed with test \\\\(\\lVert v_\\text{orth} \\rVert_2 < \\epsilon * \\lVert v \\rVert_2\\\\)\n"
                              "Please note that a QR1 is based on Given's rotations whereas QR2 uses "
                              "modified Gram-Schmidt. This can give different results even when no columns "
-                             "are filtered out.\n"
-                             "When this tag is not provided, the QR2-filter with the limit value 1e-2 is used.");
-  XMLAttribute<double> attrSingularityLimit(ATTR_SINGULARITYLIMIT, 1e-16);
+                             "are filtered out.");
+  XMLAttribute<double> attrSingularityLimit(ATTR_SINGULARITYLIMIT, 1e-2);
   attrSingularityLimit.setDocumentation("Limit eps of the filter.");
   tagFilter.addAttribute(attrSingularityLimit);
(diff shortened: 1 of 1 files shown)


src/acceleration/config/AccelerationConfiguration.cpp
```diff
@@ -381,21 +381,22 @@ void AccelerationConfiguration::addCommonIQNSubtags(xml::XMLTag &tag)
...
+                             " - `QR1_absolute-filter`: update QR-dec with (absolute) test \\\\(R(i, i) < \\epsilon\\\\)\n"
                              " - `QR2-filter`: en-block QR-dec with test \\\\(\\lVert v_\\text{orth} \\rVert_2 < \\epsilon * \\lVert v \\rVert_2\\\\)\n\n"
+                             " - `QR3-filter`: update QR-dec only when the pre-scaling weights have changed or there is one or more columns are to be removed with test \\\\(\\lVert v_\\text{orth} \\rVert_2 < \\epsilon * \\lVert v \\rVert_2\\\\)\n"
                              "Please note that a QR1 is based on Given's rotations whereas QR2 uses "
                              "modified Gram-Schmidt. This can give different results even when no columns "
-                             "are filtered out.\n"
-                             "When this tag is not provided, the QR2-filter with the limit value 1e-2 is used.");
-  XMLAttribute<double> attrSingularityLimit(ATTR_SINGULARITYLIMIT, 1e-16);
+                             "are filtered out.");
+  XMLAttribute<double> attrSingularityLimit(ATTR_SINGULARITYLIMIT, 1e-2);
```
davidscn: Doesnt this also change the default filter threshold?

fsimonis: Interesting.
The limit is different when defining no filer to defining a filter without limit.

Defining nothing defaults to `1e-2`.
https://github.com/precice/precice/blob/790b2782d0502927f830638d88196c40e6c2b80e/src/acceleration/config/AccelerationConfiguration.hpp#L125-L148

Defining `<filter type="..." />` defaults to `1e-16`.

This could even warrant a fix entry in the changelog.

uekerman: `1e-2` is a good value for QR2 and QR3, but not for QR1. My tendency would be to not change it. In any case, this should be a separate PR, no?

src/acceleration/config/AccelerationConfiguration.cpp
```diff
@@ -381,13 +381,13 @@ void AccelerationConfiguration::addCommonIQNSubtags(xml::XMLTag &tag)
   XMLTag tagFilter(*this, TAG_FILTER, XMLTag::OCCUR_NOT_OR_ONCE);
   tagFilter.setDocumentation("Type of filtering technique that is used to "
                              "maintain good conditioning in the least-squares system. Possible filters:\n"
-                             " - `QR1-filter`: updateQR-dec with (relative) test \\\\(R(i,i) < \\epsilon *\\lVert R\\rVert_F\\\\)\n"
-                             " - `QR1_absolute-filter`: updateQR-dec with (absolute) test \\\\(R(i, i) < \\epsilon\\\\)\n"
+                             " - `QR1-filter`: update QR-dec with (relative) test \\\\(R(i,i) < \\epsilon *\\lVert R\\rVert_F\\\\)\n"
+                             " - `QR1_absolute-filter`: update QR-dec with (absolute) test \\\\(R(i, i) < \\epsilon\\\\)\n"
                              " - `QR2-filter`: en-block QR-dec with test \\\\(\\lVert v_\\text{orth} \\rVert_2 < \\epsilon * \\lVert v \\rVert_2\\\\)\n\n"
+                             " - `QR3-filter`: update QR-dec only when the pre-scaling weights have changed or there is one or more columns are to be removed with test \\\\(\\lVert v_\\text{orth} \\rVert_2 < \\epsilon * \\lVert v \\rVert_2\\\\)\n"
```
fsimonis: Given that we are touching this anyway now, let's remove the `-filter` suffix here to make the list correspond to the `filter` attribute.

```suggestion
                             " - `QR1`: update QR-dec with (relative) test \\\\(R(i,i) < \\epsilon *\\lVert R\\rVert_F\\\\)\n"
                             " - `QR1-absolute`: update QR-dec with (absolute) test \\\\(R(i, i) < \\epsilon\\\\)\n"
                             " - `QR2`: en-block QR-dec with test \\\\(\\lVert v_\\text{orth} \\rVert_2 < \\epsilon * \\lVert v \\rVert_2\\\\)\n\n"
                             " - `QR3`: update QR-dec only when the pre-scaling weights have changed or there is one or more columns are to be removed with test \\\\(\\lVert v_\\text{orth} \\rVert_2 < \\epsilon * \\lVert v \\rVert_2\\\\)\n"
```

PR #2352, opened 2025-08-21
Fix CouplingScheme finalize aborting in DTOR

fsimonis:
## Main changes of this PR

This PR changes BaseCouplingScheme finalize to not throw an error (and thus exception) in the finalize.

This crashes the participant descructor when a requirement isn't fulfilled.
This can easily happen when an error is triggered between initialize and the first advance.

## Author's checklist


fsimonis:
Systemtests are failing because the OpenFOAM servers are currently down.

PR #2357, opened 2025-09-11
FORTRAN Example: Update README.md

cysanghavi2:
## Author's checklist

## Reviewers' checklist


MakisH (commented):
Thanks for the contribution, @cysanghavi2!

Could you please elaborate on what you were trying to do and what issues you had with the procedure that was already there?

The "alternatively" is currently not very specific: Alternatively to what? One definitely needs to have preCICE installed, and CMake does not try to find the preCICE module, but to link directly to the intrinsic Fortran bindings.

fsimonis:
Closing due to inactivity

Issue #479, opened 2019-08-06
Add option to pass custom MPI_Comm to bindings


MakisH:
https://github.com/precice/precice/pull/382 added the option to use a user-defined MPI communicator in the `SolverInterface` constructor.

We still need to port this to our language bindings for:
- [x] C
- [x] Fortran/C https://github.com/precice/precice/pull/2363 
- [x] Fortran 2003 https://github.com/precice/fortran-module/pull/44 
- [x] Python https://github.com/precice/python-bindings/issues/19
- [x] MATLAB https://github.com/precice/matlab-bindings/issues/13 `wontfix` now
- [x] Julia

BenjaminRueth:
Let's deal with this issue at our Hackathon.

BenjaminRueth:
In #382 an appropriate test case is mentioned for the elastictube (see https://github.com/precice/elastictube1d/tree/user-defined-comm). We will have to define similar test cases for the other language bindings, as well. I will try to do so for the solverdummies, since they exist for all the languages where bindings exist. The elastictube only exists in C++ and python. 

*Note:* If we already have the test cases, we should probably also run them somewhere on a regular basis. Otherwise we are to some degree wasting the effort put into developing the tests.

BenjaminRueth:
I wrote a solverdummy for python that uses communicator splitting to test the optional `communicator` argument of `Interface`. See https://github.com/precice/python-bindings/commit/168a70ffd75497d63885d76c012e7b9e906bbeea. Still open question: Do we want to use the customized solverdummy (Or the [modified elastic tube](https://github.com/precice/elastictube1d/tree/user-defined-comm) provided by @MakisH ) for testing?

BenjaminRueth:
For the matlab bindings, we will not deal with the issue yet. I opened an issue in the matlab bindings repository: https://github.com/precice/matlab-bindings/issues/13

kcooley-cvd:
I have Fortran bindings that enable this and work with our in-house adapter using preCICE 2.5.1. I'll work on updating them to preCICE 3.x and report back once ready.

fsimonis:
Closed the Fortran part by https://github.com/precice/precice/pull/2363
Thanks a lot @kcooley-cvd

kcooley-cvd:
Fortran 2003 is done now after [this PR](https://github.com/precice/fortran-module/pull/44) on the separate bindings module.

MakisH:
[...]

PR #2363, opened 2025-09-16
Add function to create Fortran participant using custom communicator

kcooley-cvd:
Use MPI_Comm_f2c to convert the Fortran communicator handle to a C handle, which is then used to create the participant as usual.

## Main changes of this PR

Add a Fortran binding to create a participant from Fortran using a specified MPI communicator.

See issue #479

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/extras/bindings/fortran/include/precice/preciceFortran.hpp
+++ b/extras/bindings/fortran/include/precice/preciceFortran.hpp
@@ -40,6 +40,31 @@ PRECICE_API void precicef_create_(
     int         participantNameLength,
     int         configFileNameLength);
 
+/**
+ * Fortran syntax:
+ * precicef_create_with_communicator_(
+ *   CHARACTER participantName(*),
+ *   CHARACTER configFileName(*),
+ *   INTEGER   solverProcessIndex,
+ *   INTEGER   solverProcessSize,
+ *   INTEGER   communicator )
+ *
+ * IN:  participantName, configFileName, solverProcessIndex, solverProcessSize,
+ *      communicator
+ * OUT: -
+ *
+ * @copydoc precice::Participant::Participant()
+ *
+ */
+PRECICE_API void precicef_create_with_communicator_(
+    const char *participantName,
+    const char *configFileName,
+    const int  *solverProcessIndex,
+    const int  *solverProcessSize,
+    const int  *communicator,
+    int         participantNameLength,
+    int         configFileNameLength);
+
 ///@}
 
 /// @name Steering Methods
--- a/extras/bindings/fortran/src/preciceFortran.cpp
+++ b/extras/bindings/fortran/src/preciceFortran.cpp
@@ -72,7 +72,7 @@ try {
   // However, the Fortran standard guarantees interoperability of the int type,
   // so that is what is passed. Do the conversion in case int is not identical
   // to MPI_Fint.
-  MPI_Fint f_communicator = (MPI_Fint) *communicator;
+  MPI_Fint f_communicator = static_cast<MPI_Fint>(*communicator);
   auto     c_communicator = MPI_Comm_f2c(f_communicator);
   impl                    = std::make_unique<precice::Participant>(
       precice::impl::strippedStringView(participantName, participantNameLength),
--- a/extras/bindings/fortran/src/preciceFortran.cpp
+++ b/extras/bindings/fortran/src/preciceFortran.cpp
@@ -2,6 +2,9 @@
 #include <cstddef>
 #include <iostream>
 #include <memory>
+#ifndef PRECICE_NO_MPI
+#include <mpi.h>
+#endif
 #include <stdexcept>
 #include <string>
 #include <string_view>
@@ -51,6 +54,40 @@ try {
       precice::impl::strippedStringView(configFileName, configFileNameLength),
       *solverProcessIndex,
       *solverProcessSize);
+} catch (::precice::Error &e) {
+  std::abort();
+}
+
+void precicef_create_with_communicator_(
+    const char *participantName,
+    const char *configFileName,
+    const int  *solverProcessIndex,
+    const int  *solverProcessSize,
+    const int  *communicator,
+    int         participantNameLength,
+    int         configFileNameLength)
+try {
+#ifndef PRECICE_NO_MPI
+  // MPI_Comm_f2c requires an MPI_Fint argument, which may differ from int.
+  // However, the Fortran standard guarantees interoperability of the int type,
+  // so that is what is passed. Do the conversion in case int is not identical
+  // to MPI_Fint.
+  MPI_Fint f_communicator = (MPI_Fint) *communicator;
+  auto     c_communicator = MPI_Comm_f2c(f_communicator);
+  impl                    = std::make_unique<precice::Participant>(
+      precice::impl::strippedStringView(participantName, participantNameLength),
+      precice::impl::strippedStringView(configFileName, configFileNameLength),
+      *solverProcessIndex,
+      *solverProcessSize,
+      &c_communicator);
+#else
+  PRECICE_WARN("preCICE was configured without MPI but you passed an MPI communicator. preCICE ignores the communicator and continues.");
+  impl = std::make_unique<precice::Participant>(
+      precice::impl::strippedStringView(participantName, participantNameLength),
+      precice::impl::strippedStringView(configFileName, configFileNameLength),
+      *solverProcessIndex,
+      *solverProcessSize);
+#endif
 } catch (::precice::Error &e) {
   std::abort();
 }
--- a/docs/changelog/2363.md
+++ b/docs/changelog/2363.md
@@ -1,0 +1,1 @@
+- Added function to create a participant from Fortran with a custom communicator.


extras/bindings/fortran/src/preciceFortran.cpp
```diff
@@ -55,6 +56,26 @@ try {
...
+void precicef_create_with_communicator_(
+    const char *participantName,
+    const char *configFileName,
+    const int  *solverProcessIndex,
+    const int  *solverProcessSize,
+    const int  *communicator,
+    int         participantNameLength,
+    int         configFileNameLength)
+try {
+  MPI_Fint c_communicator = MPI_Comm_f2c(*communicator);
```
kcooley-cvd: MPI_Comm_f2c takes an `MPI_Fint` argument. `MPI_Fint` is [defined by the MPI standard](https://www.mpi-forum.org/docs/mpi-2.2/mpi22-report/node361.htm) to be the same as a Fortran `INTEGER` of default kind, which is "usually" equivalent to `int`. Some Fortran compilers will warn if an integer argument passed to a bound C function is not of kind `C_INT`, which may or may not be equivalent to default kind. So there is a possibility that `MPI_Fint` == `INTEGER` != `INTEGER(KIND=C_INT)` == `int`. If so, a user might get anything from a compiler warning to a mysterious MPI failure.

I can add a `static_assert` to stop compilation if this situation is detected, but perhaps there is another preferred way?

fsimonis:
This is an interesting problem:

How do we handle this consistently if preCICE isn't compiled with MPI?
In C++, we ignore the communicator (we probably should warn at this point).
In C, we simply forward to C++.
In Fortran, we need MPI to convert the communicator before we can forward it using MPI. If we don't have MPI, we could either pass a dummy value (as passing nullptr is considered an error) or warn and call the constructor without the communicator.
I think the last one would make most sense here as dummy values are pretty tricky to get right.

> Some Fortran compilers will warn if an integer argument passed to a bound C function is not of kind C_INT, which may or may not be equivalent to default kind. So there is a possibility that MPI_Fint == INTEGER != INTEGER(KIND=C_INT) == int. If so, a user might get anything from a compiler warning to a mysterious MPI failure.

MPI implementations must face the same issue. How do they support the case where MPI is compiled normally but the calling FORTRAN code is compiled with different `INTEGER` size?

MakisH:
Thank you for the contribution! This was haunting me for a while, and I previously attempted to address it in https://github.com/precice/precice/pull/651.

How do you test this? Back in the day, I used a [modified elastic-tube-1d](https://github.com/precice/elastictube1d/commit/0a41ea15154e1f7c6aee0f807317fd7f80d17934). Would be great to have an integration test, instead.

Note to self: Maybe we could provide such an example as a tutorial to showcase the feature. We could indeed use the elastic-tube-1d again.

extras/bindings/fortran/include/precice/preciceFortran.hpp
```diff
@@ -40,6 +40,31 @@ PRECICE_API void precicef_create_(
     int         participantNameLength,
     int         configFileNameLength);
 
+/**
+ * Fortran syntax:
+ * precicef_create(
```
MakisH:
```suggestion
 * precicef_create_with_communicator(
```

extras/bindings/fortran/src/preciceFortran.cpp
```diff
@@ -2,6 +2,7 @@
 #include <cstddef>
 #include <iostream>
 #include <memory>
+#include <mpi.h>
```
MakisH, line 6: We still need to adjust either:

- the Fortran bindings to also work without MPI, or
- the bare CI workflows to not build them

kcooley-cvd: Added `#ifndef PRECICE_NO_MPI` to fall back to creating the participant with no communicator if preCICE was built without MPI.

extras/bindings/fortran/include/precice/preciceFortran.hpp
kcooley-cvd: Fixed

kcooley-cvd:
> How do you test this? Back in the day, I used a [modified elastic-tube-1d](https://github.com/precice/elastictube1d/commit/0a41ea15154e1f7c6aee0f807317fd7f80d17934). Would be great to have an integration test, instead.

I made this improvement because my code has an overset grid assembler that uses some of the available processors, so it splits the communicator into overset and flow-solver communicators and passes the flow-solver communicator when creating the participant. So I have seen it work, but with large test cases using in-house proprietary software.

Any case could become a test case for this feature, if it's OK to just pass `MPI_COMM_WORLD` as the communicator.

extras/bindings/fortran/src/preciceFortran.cpp
kcooley-cvd: After some reading I believe the correct path is the following. From the [MPI standard](https://www.mpi-forum.org/docs/mpi-5.0/mpi50-report/node473.htm#Node473):

> MPI-3.0 (and later) compliant Fortran bindings are not only a property of the MPI library itself, but rather a property of an MPI library together with the Fortran compiler suite for which it is compiled.

I understand this to mean that if you build your MPI with/for a particular Fortran compiler, that library will use the same default `INTEGER`/`REAL`/`COMPLEX` kinds as that Fortran compiler. To answer the question [here](https://github.com/precice/precice/pull/2363#issuecomment-3302096905):

> MPI implementations must face the same issue. How do they support the case where MPI is compiled normally but the calling FORTRAN code is compiled with different INTEGER size?

If you later build some application with a different Fortran compiler, or on a different system, with different default kinds then MPI won't work and it's your fault, not MPI's. So to get a C handle for a communicator in a safe, standard-compliant way, one *has* to use `MPI_Comm_f2c` which *has* to take a Fortran default `INTEGER` argument. The [binding interface](https://github.com/precice/fortran-module/blob/develop/precice.f90) should then be

```fortran
SUBROUTINE precicef_create_with_communicator(participantName, configFileName, &
                              solverProcessIndex, solverProcessSize, &
                              communicator, &
                              participantNameLength, configFileNameLength) &
   BIND(C, NAME='precicef_create_with_communicator_')
   IMPORT :: C_INT, C_CHAR
   CHARACTER(KIND=C_CHAR), DIMENSION(*), INTENT(IN) :: participantName
   CHARACTER(KIND=C_CHAR), DIMENSION(*), INTENT(IN) :: configFileName
   INTEGER(KIND=C_INT), INTENT(IN) :: solverProcessIndex
   INTEGER(KIND=C_INT), INTENT(IN) :: solverProcessSize
   INTEGER, INTENT(IN) :: communicator  ! Note default kind here instead of C_INT
   INTEGER(KIND=C_INT), VALUE, INTENT(IN) :: participantNameLength
   INTEGER(KIND=C_INT), VALUE, INTENT(IN) :: configFileNameLength
END SUBROUTINE precicef_create_with_communicator
```
Depending on the system and the compiler, a user might get a compiler warning when building their adapter (not when building preCICE itself) about the default `INTEGER` not necessarily being C-interoperable. It would be better to check for that on the Fortran side, i.e. to assert that `KIND(1) == C_INT`.

extras/bindings/fortran/include/precice/preciceFortran.hpp
```diff
@@ -40,6 +43,42 @@ PRECICE_API void precicef_create_(
...
 
+/**
+ * Fortran specifies the communicator as an integer of default kind which is
+ * equivalent to the C type MPI_Fint, but if preCICE is built without MPI then
+ * the MPI header is not available so a different type must be used.
+ */
+#ifndef PRECICE_NO_MPI
+typedef MPI_Fint comm_int_type;
+#else
+typedef int comm_int_type;
```
fsimonis: We prefer `using` over `typedef` and use scoped global names:

```suggestion
using PRECICE_COMM_INT_TYPE = MPI_Fint;
#else
using PRECICE_COMM_INT_TYPE = int;
```

fsimonis (changes requested):

fsimonis:
Lets this ready and include it in the 3.3.0 release.

extras/bindings/fortran/include/precice/preciceFortran.hpp
kcooley-cvd: After some more thought I decided it is better to pass `int` to `precicef_create_with_communicator_`, to benefit from the interoperability between `int` and Fortran `INTEGER(KIND=C_INT)` guaranteed by the Fortran standard, and then explicitly cast that argument to an `MPI_Fint` inside that function before passing it to `MPI_Comm_f2c`. Then this typedef is not necessary. There should also be no compiler warnings unless the user tries to pass an `INTEGER` of different kind from their Fortran code, which would not have worked anyway and which can be easily fixed in their Fortran code.

extras/bindings/fortran/src/preciceFortran.cpp
```diff
@@ -55,6 +58,40 @@ try {
...
+    const int  *communicator,
+    int         participantNameLength,
+    int         configFileNameLength)
+try {
+#ifndef PRECICE_NO_MPI
+  // MPI_Comm_f2c requires an MPI_Fint argument, which may differ from int.
+  // However, the Fortran standard guarantees interoperability of the int type,
+  // so that is what is passed. Do the conversion in case int is not identical
+  // to MPI_Fint.
+  MPI_Fint f_communicator = (MPI_Fint) *communicator;
```
fsimonis: Let's use a static cast here.
~~Also, we still need to check if the communicator is not a nullptr with `PRECICE_CHECK(communicator, "")`.~~

_edit: nonsense, it shouldn't be null on the fortran side_

```suggestion
  MPI_Fint f_communicator = static_cast<MPI_Fint>(*communicator);
```

extras/bindings/fortran/include/precice/preciceFortran.hpp
fsimonis: This sounds like a very reasonable approach. Thanks!

extras/bindings/fortran/src/preciceFortran.cpp
fsimonis: I'm marking this as resolved as the conversation continues below.

fsimonis:
@kcooley-cvd I'm taking this PR over and add the last bits before merging as I aim to get the release candidate ready today.

fsimonis:
I tested the change with the following patch of the solverdummy: [fortrandummy.patch](https://github.com/user-attachments/files/22513326/fortrandummy.patch)

It seems to work fine.

The Windows CI is failing due to the Eigen 5 release.

fsimonis:
@kcooley-cvd Thank you very much for the research and contribution!

PR #2364, opened 2025-09-17
Add quads

### LINKED ISSUE #2296: Quads aren't supported in 2D
**Describe your setup**

Operating system (e.g. Linux distribution and version): all
preCICE Version: 3.2.0 and before

**Describe the problem**

When using LCI in 2D, the user needs to define triangles or quads.
The latter has an internal check for convexity, which only supports 3D coordinates.
This leads to an error and triangles need to be used as a fall-back.

**Step To Reproduce**
Based on https://precice.discourse.group/t/different-behaviou-of-setmeshquads-between-v3-1-2-and-v3-2-0-in-setting-2d-mesh-connectivity/2451
 
1. Define 2d meshes and a linear-cell-interpolation mapping between them. 
2. Define quads using `setMeshQuads`
3. Observe the assertion `This only works in 3D.` in Debug.
 
**Expected behaviour**

Either:
* the Quads should be set correctly for 2D meshes. `setMeshQuad()` should not error due to dimensionality and `setMeshQuads()` should not run into an assertion 
* `setMeshQuads()` should error due to dimensionality

**Additional context**

[Reported on discourse](https://precice.discourse.group/t/different-behaviou-of-setmeshquads-between-v3-1-2-and-v3-2-0-in-setting-2d-mesh-connectivity/2451)

carme-hp:
## Main changes of this PR

Fix bug that prevented quads to be set for 2d meshes. 

This PR closes #2296

## Author's checklist

## Reviewers' checklist


fsimonis:
The interface still doesn't allow setting quads for a 2D mesh.

carme-hp:
Thank you, it should be fine now

PR #2377, opened 2025-10-13
Fix scalar handling in AxialGeoMultiscaleMapping

franiqui:
Added a dedicated code path for scalar data (`inDataDimensions = 1`) 
to prevent out-of-bounds access caused by using `effectiveCoordinate` 
for single-component fields.

- **SPREAD:** replicate single input pressure to all output vertices  
- **COLLECT:** average all input pressures into one output vertex

---

## Main changes of this PR
- Introduced explicit scalar handling in `mapConsistent()`  
- Prevented invalid indexing for scalar data (`dims == 1`)  
- Preserved existing behavior for vector (velocity) mappings  

---

The previous implementation assumed vector data for all mappings.  
When scalar quantities (like pressure) were used with 
`effectiveCoordinate = 1` or `2`, the code accessed 
indices beyond the vector bounds, causing Eigen assertion errors.

This PR adds a safe and clear scalar path.

## Author's checklist

### DIFF
--- a/docs/changelog/2377.md
+++ b/docs/changelog/2377.md
@@ -1,1 +1,1 @@
-Fixed invalid indexing for scalar data in axial geometric multiscale mapping by introducing explicit scalar handling. (#2377)
+- Fixed invalid indexing for scalar data in axial geometric multiscale mapping by introducing explicit scalar handling.


franiqui:
Hi @MakisH 
Here is my proposed fix for scalar handling in `AxialGeoMultiscaleMapping.cpp`.

src/mapping/AxialGeoMultiscaleMapping.cpp
```diff
@@ -117,6 +117,30 @@ void AxialGeoMultiscaleMapping::mapConsistent(const time::Sample &inData, Eigen:
   PRECICE_ASSERT(inDataDimensions == outDataDimensions);
 
   PRECICE_DEBUG("Map consistent");
+
+  // ---------- NEW: scalar (pressure) handling ----------
```
MakisH: In the "Files changed" view, I can see the diff of your branch. No need to highlight the changes with comments (which I understand you added for the draft you previously sent me).

Image

src/mapping/AxialGeoMultiscaleMapping.cpp
```diff
@@ -117,6 +117,30 @@ void AxialGeoMultiscaleMapping::mapConsistent(const time::Sample &inData, Eigen:
...
+  if (inDataDimensions == 1 && outDataDimensions == 1) {
+    if (_type == MultiscaleType::SPREAD) {
+      // 1D -> 3D: replicate the single input pressure to every output vertex
+      PRECICE_ASSERT(input()->nVertices() == 1, "Pressure SPREAD expects one input vertex.");
+      const double p = inputValues(0);
```
MakisH: We should not hard-code "pressure" into our assumptions. The mapping (and the library itself) is meant to be application-agnostic.

I do see some comments in already existing parts of the code.
Comments are helpful as examples/intention, but error messages and variable names should be more general.

For the error message: `Axial spread mapping` should do.

src/mapping/AxialGeoMultiscaleMapping.cpp
```diff
@@ -117,6 +117,30 @@ void AxialGeoMultiscaleMapping::mapConsistent(const time::Sample &inData, Eigen:
...
+      for (size_t i = 0; i < output()->nVertices(); ++i) {
+        outputValues(static_cast<Eigen::Index>(i)) = p;
```
MakisH: Instead of the `static_cast` to `Eigen::Index`, why not defining the loop index itself as `Eigen::Index`?

src/mapping/AxialGeoMultiscaleMapping.cpp
```diff
@@ -117,6 +117,30 @@ void AxialGeoMultiscaleMapping::mapConsistent(const time::Sample &inData, Eigen:
   PRECICE_ASSERT(inDataDimensions == outDataDimensions);
 
   PRECICE_DEBUG("Map consistent");
+
+  // ---------- NEW: scalar (pressure) handling ----------
+  if (inDataDimensions == 1 && outDataDimensions == 1) {
+    if (_type == MultiscaleType::SPREAD) {
```
MakisH: I would try to merge the scalar and the non-scalar case, to avoid these duplications. Avoiding such duplication would not only make the code shorter, but also easier to maintain and debug (it would be easy to forget handling an issue in both cases).

src/mapping/AxialGeoMultiscaleMapping.cpp
MakisH, line 1: A test case is needed as well (this is caught by CodeCov, which notices that the new lines of code are not covered by any test).

Related files:

- Unit tests: `src/mapping/tests/AxialGeoMultiscaleMappingTest.cpp`
- Integration tests: `tests/geometric-multiscale/AxialGeoMultiscale.cpp` (needs an XML file as well, let's name it `AxialGeoMultiscaleReverse`)

src/mapping/tests/AxialGeoMultiscaleMappingTest.cpp
```diff
@@ -243,3 +243,93 @@ BOOST_AUTO_TEST_CASE(ConsistentCollectZ)
 
 BOOST_AUTO_TEST_SUITE_END()
 BOOST_AUTO_TEST_SUITE_END()
+
+PRECICE_TEST_SETUP(1_rank);
+BOOST_AUTO_TEST_CASE(ConsistentSpreadXScalar)
+{
+  PRECICE_TEST();
+  // 1D -> 3D, scalar data: the same value is copied to all output vertices, along X axis
```
MakisH: This case needs to be updated to consider the spread profile, as this one would be uniform: https://github.com/precice/precice/pull/2386

When that gets merged, the unique characteristic of this tutorial would be the scalar data, not that the same value is copied to all output vertices. The comment could better reflect this:


```suggestion
  // 1D -> 3D, scalar data, uniform spread along the X axis
```

src/mapping/tests/AxialGeoMultiscaleMappingTest.cpp
```diff
@@ -243,3 +243,93 @@ BOOST_AUTO_TEST_CASE(ConsistentCollectZ)
...
+
+  double radius = 1.0; // radius of the "tube" from or to which the data is mapped, i.e., radius of the circular interface between the two participants
+
+  // Setup mapping with mapping coordinates and geometry used
+  precice::mapping::AxialGeoMultiscaleMapping mapping(mapping::Mapping::CONSISTENT, dimensions, mapping::AxialGeoMultiscaleMapping::MultiscaleType::SPREAD, mapping::AxialGeoMultiscaleMapping::MultiscaleAxis::X, radius);
+  mapping.setMeshes(inMesh, outMesh);
+  BOOST_TEST(mapping.hasComputedMapping() == false);
+
+  // Create data to map
+  Eigen::VectorXd inValues(1);
```
MakisH, line 275: Do we actually need an `Eigen::VectorXd` here, if we want a scalar? From the user perspective, that would be strange, and maybe it hides more issues.

Even if it should work, maybe we can use a `double` in this case, to test both (assuming both should work).

src/mapping/tests/AxialGeoMultiscaleMappingTest.cpp
```diff
@@ -243,3 +243,93 @@ BOOST_AUTO_TEST_CASE(ConsistentCollectZ)
...
+  outValues = Eigen::VectorXd::Zero(1);
+
+  // Map data
+  mapping.computeMapping();
+  mapping.map(inSample, outValues);
+
+  BOOST_TEST(mapping.hasComputedMapping() == true);
+
+  // All outputs equal to the scalar input
+  BOOST_TEST(outValues(0) == (3.0 + 7.0 + 5.0) / 3.0);
```
MakisH: Dividing by `3.0` hides the fact that this is an average. It should also not be needed, given that the enumerator is already `double`.

```suggestion
  BOOST_TEST(outValues(0) == (3.0 + 7.0 + 5.0) / 3);
```

src/mapping/AxialGeoMultiscaleMapping.cpp
```diff
@@ -116,31 +110,48 @@ void AxialGeoMultiscaleMapping::mapConsistent(const time::Sample &inData, Eigen:
...
+        outputValues(i) = s;
+      }
+    } else {
+      // Vector: Currently, a Hagen-Poiseuille profile determines the velocity value.
+      for (size_t i = 0; i < outSize; i++) {
+        PRECICE_ASSERT(static_cast<size_t>((i * outDataDimensions) + effectiveCoordinate) < static_cast<size_t>(outputValues.size()),
+                       ((i * outDataDimensions) + effectiveCoordinate), outputValues.size());
+        // When adding support for 2D, remember that this should be 1.5 * inputValues(effectiveCoordinate) * (1 - (_vertexDistances[i] * _vertexDistances[i]));
+        outputValues((i * outDataDimensions) + effectiveCoordinate) = 2 * inputValues(effectiveCoordinate) * (1 - (_vertexDistances[i] * _vertexDistances[i]));
+      }
```
MakisH, line 144: We probably should first merge #2386 and then reconsider the different profiles here.

I think that distributing a scalar from the 1D to the 3D with a profile should also be a valid option. We can add a "not yet implemented" error, but these should be independent options.

Otherwise, we should check and throw an error earlier if a user tries to set a profile for a scalar spread, and document it.

tests/geometric-multiscale/AxialGeoMultiscaleReverse.cpp
MakisH, line 1: Self-reminder: Have another look at this test (but I think it's fine as-is).

src/mapping/AxialGeoMultiscaleMapping.cpp
MakisH: Since it looks like the tests #2386 are failing without this fix, let's first merge this one, and handle the special cases in #2386.

docs/changelog/2377.md
```diff
@@ -0,0 +1 @@
+Fixed invalid indexing for scalar data in axial geometric multiscale mapping by introducing explicit scalar handling. (#2377)
```
MakisH: 1. We need a bullet in the beginning (assumed when preparing the total changelog)
2. I was wrong, we don't need the PR number in here


```suggestion
- Fixed invalid indexing for scalar data in axial geometric multiscale mapping by introducing explicit scalar handling.
```

PR #2379, opened 2025-10-14
Overhaul docker releases

fsimonis:
## Main changes of this PR

This PR attempts to 
- allow the nightly docker image and the docker release to be called from other workflows
- allows versions with v prefix in the release workflow
- adds the ability to pass a branch or tag to the nightly release with default of develop (to work with cron jobs)
- ~~reuses both docker release and docker nightly as part of the release pipeline~~ moved to #2380

I tested using workflow dispatch:
- release can be called with v prefix
- nightly can be called without branch argument (cron job) 

## Author's checklist

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -1,0 +1,21 @@
+name: Test
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        type: string
+
+jobs:
+
+  docker-release:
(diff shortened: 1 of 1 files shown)


fsimonis:
I split the release workflow into a separate PR as I really want to test the workflow calling and for this I need to move the test to the develop branch.

PR #2385, opened 2025-10-23
Add mocking prototype

JulianHarrison-Wirth:
## Main changes of this PR

Adds a **mock of the preCICE Participant API** in `extras/mock/`, so adapters can be developed and tested without running real preCICE and without a coupling partner.

* **`libpreciceMocked.so`** (CMake target `preciceMocked`): a drop-in replacement for `libprecice.so` that re-implements the complete `precice::Participant` API in `extras/mock/preciceMocked.cpp`. It does not link against `libprecice` and only needs a subset of its dependencies — a C++17 compiler, Eigen, libxml2, the bundled fmt, and optionally MPI. No Boost, no PETSc, no communication backends.
* **Two ways to use it**: preload it into an unmodified solver (`LD_PRELOAD=/path/to/libpreciceMocked.so ./run.sh`, which also covers solvers using the language bindings), or link `preciceMocked` instead of `precice`.
* **Configuration-aware validation**: the mock reads your regular preCICE configuration and validates every API call against it — participant, mesh and data names, read/write direction, buffer sizes, and the legal order of steering calls — and reports violations as `precice::Error` with an actionable message. Time stepping, subcycling, implicit iterations and checkpoint requirements are derived from the configured coupling scheme.
* **Optional `precice-mock-config.xml`** (placed next to the preCICE configuration) controls what `readData()` returns — `buffer` (echoes what was written, default), `scaled` (scalar or element-wise vector multiplier), or `random` (with bounds and seed) — per mesh/data pair or globally, plus the logging mode and a `max-iterations` override for implicit coupling.
* **C and Fortran bindings** are compiled into the mock library as well. To make that possible they no longer depend on preCICE internals (`logging/Logger`, `logging/LogMacros`, `utils/assertion`): the `PRECICE_CHECK` guards for "participant not created (yet)" now print to `std::cerr` and `std::abort()`, which matches the previous observable behaviour without pulling in Boost.Log.
* **Tests** (`extras/mock/tests/`, see `extras/mock/tests/README.md`):
  * *Golden API tests* — 26 ctest cases (`ctest -L mock`). A driver (`tests/api/driver.cpp`) exercises the API scenario by scenario (explicit, implicit, subcycling, max-time, first-participant, error paths) and its output is diffed against committed expectations. Since the mock is deterministic, any diff is a behaviour change.
  * *Differential harness (manual)* — `tests/api/run.sh` runs the same driver against the real `libprecice` as a coupled pair and against the mock, and diffs both. This is how the golden files are (re)validated whenever real preCICE behaviour changes.
  * *Tutorial smoke tests (manual, ~4.5 min)* — `tests/tutorials/smoke.sh` runs real adapters (OpenFOAM, python bindings, FMI, nutils, C++) from a curated set of tutorials plus the four preCICE course solution sets against the mock, and compares normalized output against committed baselines. `smoke_update.sh` regenerates those baselines, first validating every case against real preCICE.
* **Docs**: `extras/mock/README.md` (building, usage, configuration, behaviour, limitations) and `extras/mock/tests/README.md`.
* `.pre-commit-config.yaml`: excludes the committed tutorial baselines from codespell.

### Known limitations (documented in the README)

* **No real communication**: each rank mocks the coupling independently and reads back rank-local buffers. An MPI communicator passed to the constructor is validated but not used, so `getMeshVertexSize()` on a received mesh without API access returns 0.
* **No convergence measures**: implicit schemes converge after a fixed number of iterations (default 2, overridable) instead of evaluating the configured measures.
* **Connectivity and gradient data** are validated but not stored (there is no read counterpart for gradients in the API).
* If a configuration bounds neither `max-time` nor `max-time-windows`, the mock adds `max-time-windows="100"` and warns — valid for real preCICE, but the mock has no partner that could end the simulation.

Testing an adapter today requires either a full preCICE installation plus a second participant, or a hand-written mock maintained inside the adapter (as the Python bindings had in `test/Participant.cpp`). Both are slow to set up, awkward in CI, and expose tests to failures outside the code under test.

A single mock maintained next to preCICE gives adapter developers a fast, dependency-light way to exercise API usage, error paths and checkpoint handling, and lets bindings and adapters drop their private mocks. A companion PR for the Python bindings (precice/python-bindings#XXX) does exactly that: it removes the hand-written mock and runs the full test suite against `libpreciceMocked.so`, which also grew the suite from 65 to 76 tests, since the mock allows testing behaviour the old stub could not express.

This is a first complete-but-simple implementation and deliberately does not try to be a preCICE simulator — feedback on scope and on the mock configuration format is very welcome.

## Author's checklist

## Reviewers' checklist

* [ ] Is `extras/mock/` the right place for this, and should `preciceMocked` be installed/exported or stay a build-tree-only target?
* [ ] Are the C/Fortran bindings acceptable without preCICE's logging and assertion headers (`std::cerr` + `std::abort()` instead of `PRECICE_CHECK`)?
* [ ] Is the `precice-mock-config.xml` format (modes, defaults, iteration override) a good long-term interface for adapter developers?
* [ ] Are the committed tutorial baselines (~75k lines) acceptable in this repository, or should the smoke tests live elsewhere?
* [ ] Should the golden API tests (`ctest -L mock`) run in CI by default?


MakisH:
Following @fsimonis' suggestion, I changed the target branch to a new feature branch `mocked-api`. Let's work on smaller PRs to that feature branch. Later on, you can open a feature PR to develop.

MakisH:
@JulianHarrison-Wirth looking at the failing CI output ([example](https://github.com/precice/precice/actions/runs/19235225535/job/54983301950?pr=2385)), it looks like the tests running the C and Fortran solver dummies are the ones failing:

```
The following tests FAILED:
Errors while running CTest
	1201 - precice.solverdummy.run.c-c (Failed)              solverdummy
	1202 - precice.solverdummy.run.fortran-fortran (Failed)  solverdummy
	1203 - precice.solverdummy.run.cpp-c (Timeout)           solverdummy
	1204 - precice.solverdummy.run.cpp-fortran (Timeout)     solverdummy
	1205 - precice.solverdummy.run.c-fortran (Failed)        solverdummy
```

But the C++ version (`precice.solverdummy.run.cpp-cpp`) passes.

Example:

```
1201/1235 Test #1201: precice.solverdummy.run.c-c ............................................................................................................................................***Failed    0.46 sec
DUMMY: Running solver dummy with preCICE config file "/__w/precice/precice/examples/solverdummies/precice-config.xml" and participant name "SolverTwo".
preCICE has been created already! Be sure to call either "precicec_createParticipant" or "precicec_createParticipant_withCommunicator" exactly once.DUMMY: Running solver dummy with preCICE config file "/__w/precice/precice/examples/solverdummies/precice-config.xml" and participant name "SolverOne".
preCICE has been created already! Be sure to call either "precicec_createParticipant" or "precicec_createParticipant_withCommunicator" exactly once./__w/precice/precice/cmake/runsolverdummies.sh: line 9: 13977 Aborted                    (core dumped) $2 $3 SolverTwo
/__w/precice/precice/cmake/runsolverdummies.sh: line 10: 13976 Aborted                    (core dumped) $1 $3 SolverOne
CMake Error at /__w/precice/precice/cmake/runsolverdummies.cmake:37 (message):
  An error occurred running the solverdummies! Return code : "134"
```

The CI passes in the base branch. If I understand correctly, there is not anything forcing the `solverdummy` tests to use the mock at the moment. However, the C tests stumble upon this check (`errormsgCreate`):

```cpp
void precicec_createParticipant_withCommunicator(
    const char *participantName,
    const char *configFileName,
    int         solverProcessIndex,
    int         solverProcessSize,
    void       *communicator)
try {
  if (impl == nullptr) {
    std::cerr << errormsgCreate;
    std::abort();
  }
  impl = std::make_unique<precice::Participant>(participantName,
                                                configFileName,
                                                solverProcessIndex,
                                                solverProcessSize,
                                                communicator);
} catch (::precice::Error &e) {
  std::abort();
}
```

@fsimonis any ideas?

fsimonis:
@MakisH I just realized that I converted the checks straight to if conditions. So they are all flipped...

fsimonis:
I pushed a fix to the `add-mocking-prototype` branch. You can also cherry pick it 8373edf2afbb11ad6cbc42934461817069037744

MakisH:
@JulianHarrison-Wirth I triggered the CI, the pre-commit check is failing citing formatting issues. Let's get that fixed and see if the build actually works now.

JulianHarrison-Wirth:
@MakisH can you please run the run_all_tutorials.sh file and send me the log via email
On my machine I get 32 passed but most errors are missing packages... and I would like a result on a second machine as well

JulianHarrison-Wirth:
correction: I got 40 passes
all of the ones requiring pyprecice in a venv failed due to the newest version of pyprecice throwing an error on install
for manual install I was able to use version 3.2.1 but I don't know how that works with venv
the error was: Failed building wheel for pyprecice
and It said something about a missing/wrong licence

examples/solverdummies/cpp/CMakeLists.txt
```diff
@@ -7,7 +7,6 @@ add_executable(solverdummy solverdummy.cpp)
 set_target_properties(solverdummy PROPERTIES CXX_STANDARD 11)
 target_link_libraries(solverdummy PRIVATE precice::precice)
 
-
```
MakisH, line 10: Left-over from debugging / unrelated change.

extras/mock/README.md
```diff
@@ -0,0 +1,172 @@
+# preCICE Mock Participant
+
+A lightweight mock implementation of the preCICE Participant API for testing and development.
+
+## Features
+
+- **Full API Coverage**: Implements the complete preCICE Participant API
+- **Configuration Validation**: Parses preCICE XML configs and validates API calls
```
MakisH: Sounds easy to confuse with the actual config checker. Maybe we find a better name.

extras/mock/README.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Data Exchange Modes
+
+The mock participant supports three modes for `readData()` operations, configured via an optional `mock-config.xml` file:
+
+### 1. Random Mode
+Returns random seeded data (useful for testing error handling).
+
+```xml
+<data-mapping mesh="MeshName" data="DataName" mode="random" />
```
MakisH: For a moment, I got confused seeing a `data-mapping` tag here, thinking that this is the preCICE config.

Why not call this `read-data`? Or `mock-data` or similar?

If this is not related to `precice-config.xml`, what is the reason for XML?

It is clear that things can be improved, but some argumentation (and outlook) would be nice in your thesis text.

extras/mock/README.md
```diff
@@ -0,0 +1,172 @@
...
+```
+
+### 2. Buffer Mode (Default)
+Returns the data previously written via `writeData()`. This mimics the behavior of the Python adapter mock and is the default if no mock-config is provided.
```
MakisH:
```suggestion
Returns the data previously written via `writeData()`. This mimics the behavior of the Python bindings mock and is the default if no mock-config is provided.
```

Even though I would only mention the relation to the Python bindings mock in the thesis text, not in the README file.

extras/mock/README.md
```diff
@@ -0,0 +1,172 @@
...
+**Element-wise vector multiplication:**
+```xml
+<data-mapping mesh="MeshName" data="DataName" mode="scaled">
+  <vector-multiplier values="1.0;2.0;3.0" />
+</data-mapping>
+```
+
+## Configuration Files
+
+### preCICE Configuration
```
MakisH: To discuss: Is there a use case for using the mock without any preCICE configuration file? I.e., would it make sense to have that optional as well?

Probably not, but since that was the first prototype, let's confirm it.

Imagine the following scenario: I am a developer of a simulation code, and I want to add calls to the preCICE API, but someone else will build the actual application. Being able to run the mock without having to write a config file (which is at the moment a tedious process), would be beneficial.

In the end, I would say it depends on the effort needed to implement the defaults.

extras/mock/README.md
```diff
@@ -0,0 +1,172 @@
...
+
+## Building
+
+The mock is built as part of the preCICE library. Link against `libpreciceMocked.so` instead of `libprecice.so`:
+
+```cmake
+target_link_libraries(your_solver preciceMocked)
+```
+
+Alternatively build normally and use LD_PRELOAD to load the mock at runtime
```
MakisH: I would argue that this should be the default, to showcase that no changes are needed on the adapter side.

extras/mock/README.md
```diff
@@ -0,0 +1,172 @@
...
+- Does not perform actual mesh mapping or data interpolation
+- Does not communicate between participants (single-process only)
+- Simplified convergence logic for implicit coupling (converges after N iterations)
+- No gradient data support (returns zeros)
+- No mesh connectivity validation
+
+## Testing Your Adapter
+
+1. Create a simple preCICE config with your participant
+2. Optionally create a mock-config.xml for specific data behavior
```
MakisH:
```suggestion
2. Optionally, create a `mock-config.xml` for specific data behavior
```

extras/mock/run_all_tutorials.sh
MakisH, line 1: Is this script meant for development purposes only? It generally goes into the direction of reimplementing the [system tests](https://precice.org/dev-docs-system-tests.html). For the long-term CI, I would rather add one test case there, based on the elastic-tube-1d tutorial.

What tutorials are you trying to run here? How are you installing the dependencies? Aiming to execute all tutorials sounds a bit too much to me (but amazing for your thesis if that works).

extras/mock/run_all_tutorials.sh
```diff
@@ -0,0 +1,94 @@
+#!/bin/bash
+
+# SET CORRECT PATHS BEFORE RUNNING
+# recommended to run from the extras/mock/ directory with the command:
+# ./run_all_tutorials.sh &> mock_all_tutorials.log
+MOCK_LIB="/home/julian/Documents/precice/build/libpreciceMocked.so.3.3.0"
+TUTORIALS_DIR="/home/julian/Documents/precice/examples/tutorials"
+TIMEOUT=300  # 5 minutes timeout per tutorial
```
MakisH: Interesting approach to check the timeout (if it still runs after some time).
What is the purpose with the 5min? Do you want to check that the simulation starts, or that it does not diverge so much to make the solver exit?
Some documentation on the intention is needed.

extras/mock/mock-config.xml.example
MakisH, line 1: Why this almost duplicate file, if `mock-config.xml` already exists?
Documentation needed.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1450 @@
...
+    if (solverProcessSize <= 0) {
+      throw precice::Error(precice::utils::format_or_error("solverProcessSize must be > 0, but is {}.", solverProcessSize));
+    }
+    if (solverProcessIndex < 0) {
+      throw precice::Error(precice::utils::format_or_error("solverProcessIndex must be >= 0, but is {}.", solverProcessIndex));
+    }
+    if (solverProcessIndex >= solverProcessSize) {
+      throw precice::Error(precice::utils::format_or_error("solverProcessIndex={} must be smaller than solverProcessSize={}.", solverProcessIndex, solverProcessSize));
+    }
+    seed = static_cast<uint32_t>(rank) ^ 0x9e3779b9u;
```
MakisH: What is this for? Add a comment.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1450 @@
...
+  bool       configParsed = false;
+
+  MockConfig mockConfig;
+  bool       mockConfigParsed = false;
+
+  // Write data buffers: Key is "meshName:dataName"
+  std::map<std::string, std::vector<double>> writeBuffers;
+
+  // std::unique_ptr<FileDoubleReader> fileReader;
+
```
MakisH:
```suggestion
```

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1450 @@
...
+  void parseConfig()
```
MakisH: I wonder if we could simplify the parsing, reusing existing implementations (e.g., libxml2).

On the other hand, I see that this is only extracting minimal information.

MakisH (commented):
Some first comments, mostly on the high level so far. I have only gone through ~half the `preciceMocked.cpp`. Looks very nice so far!

Note that, to make the review easier, I cherry-picked the skeleton-related commits into the `mocked-api` branch, merged `develop` into both branches (since you had already merged as well), and merged `mocked-api` to your branch, resolving a conflict in `CMakeLists.txt`. The skeleton-related commits are still shown here, but the `files changed` view is clearer. Feel free to rebase later.

extras/mock/README.md
JulianHarrison-Wirth: updated to be mocked-data

extras/mock/README.md
JulianHarrison-Wirth: rewrote the explanation if you think the name needs updating as well let me know

extras/mock/README.md
JulianHarrison-Wirth: removed mention of the python bindings

extras/mock/README.md
JulianHarrison-Wirth: Implementing would not be a huge effort, but since most callers will have the config anyway I am not sure if it is necessary 
I can add it as an option later if wanted

extras/mock/README.md
JulianHarrison-Wirth: swaped the order

extras/mock/README.md
JulianHarrison-Wirth: added

extras/mock/run_all_tutorials.sh
JulianHarrison-Wirth: that was just a quick benchmark for dev purposes to see if I was breaking stuff

extras/mock/mock-config.xml.example
JulianHarrison-Wirth: pre-commit didnt let me have comments in an xml file

extras/mock/preciceMocked.cpp
JulianHarrison-Wirth: should be gone now

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1635 @@
...
+
+// --include precise errors instead of assertions !! check when error/exception should be used
+// test for exception handling (exceptions/errors when meant to be thrown)
+// --mimick implicit coupling with checkpointing
+// catch errors from course !! do course against mock
+// replace python (and fenics) adapter mocks
+// check if vector components are written correctly (later/put in issues)
+// --read config (parts) to see if adapter does what config expects from it !!
+// --names, dims...
+// --libxml2 wrapper in utils or xml !! Not feasible due to requiring linking the full precice library
```
fsimonis: I think you can get away with only implementing the tag callback:

https://github.com/precice/precice/blob/0085b0b04ea74bee0cc63602c48f79eda144521b/src/xml/ConfigParser.cpp#L44

Then you can setup a minimal SAXHandler like this one, but only one defined handler:
https://github.com/precice/precice/blob/0085b0b04ea74bee0cc63602c48f79eda144521b/src/xml/ConfigParser.cpp#L168-L199

Then you can detect `data:scalar` as prefix "data" and name "scalar".

This should be enough to detect Data and Mesh names and dimensions as well as participant names.
Also the `coupling-scheme:X-explicit` should be easy to detect.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1635 @@
...
+  Currently, data reading returns seeded generate random data,
+  using data from a user-specified file if provided at initialization is currently being implemented.
+  */
+namespace precice {
+
+namespace impl {
+
+class ParticipantImpl {
+public:
+  ParticipantImpl(::precice::string_view participantName,
```
fsimonis, line 217: Let's define this outside the class definition to make this more digestible.

extras/mock/preciceMocked.cpp
fsimonis: Using libxml2 would save a lot of code.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1635 @@
...
+
+Participant::~Participant() = default;
+
+// Steering methods
+
+void Participant::initialize()
+{
+  if (!_impl) {
+    throw precice::Error("Participant implementation missing in initialize().");
+  }
```
fsimonis: You initialize `_impl` in the constructor, so this shouldn't happen.
Therefore, this should be an [`assert()`](https://en.cppreference.com/w/cpp/error/assert.html) instead.

Same for the other API functions.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1635 @@
...
+  if (_impl->initialized) {
+    throw precice::Error("initialize() may only be called once.");
+  }
+
+  // Parse configuration
+  _impl->parseConfig();
+
+  // Parse mock configuration
+  _impl->parseMockConfig();
```
fsimonis: This should be done only in the constructor. Users define the meshes between the constructor and `initialize()`.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1635 @@
...
+#include <regex>
+#include <sstream>
+#include <string>
+#include <thread>
+
+#include <precice/Tooling.hpp>
+#include <precice/Types.hpp>
+#include <precice/span.hpp>
+#include "precice/Exceptions.hpp"
+#include "utils/fmt.hpp"
```
fsimonis, line 30: We are about to move to C++20, so you could directly use [`std::format`](https://en.cppreference.com/w/cpp/utility/format/format.html) in the mock.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1635 @@
+#include <chrono>
+#include <filesystem>
+#include <fstream>
+#include <future>
```
fsimonis: What are your future plans?

_pun intended_

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1635 @@
...
+  }
+
+  // Parse configuration
+  _impl->parseConfig();
+
+  // Parse mock configuration
+  _impl->parseMockConfig();
+
+  // Communication setup prints
+  if (_impl->rank == 0) {
```
fsimonis: This will become very frequent. I recommend 
```suggestion
  if (_impl->primary) {
```

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1635 @@
...
+    _impl->configData.iterationConverged = false;
+  }
+}
+
+void Participant::advance(double computedTimeStepSize)
+{
+  if (!_impl) {
+    throw precice::Error("Participant implementation missing in advance().");
+  }
+  std::lock_guard<std::mutex> lk(_impl->mtx);
```
fsimonis: For interest. Why did you add locking?
preCICE itself isn't threadsafe yet. #2407

extras/mock/preciceMocked.cpp
JulianHarrison-Wirth: what would the advantage of abstracting it to a new member variable  (primary = (rank == 0))?

extras/mock/preciceMocked.cpp
JulianHarrison-Wirth: was used in a in-between version, I have removed it now

extras/mock/preciceMocked.cpp
JulianHarrison-Wirth: I updated it to use std::format, but just fyi format is not implemented in gcc11 used in ubuntu 22.04 and only got implemented in gcc13
https://en.cppreference.com/w/cpp/compiler_support/20.html

extras/mock/preciceMocked.cpp
JulianHarrison-Wirth: removed duplicate calls, parseConfig() is still guarded against multiple calls

extras/mock/preciceMocked.cpp
JulianHarrison-Wirth: I changed the c++ version in the makefile 
@fsimonis do you prefer it only be changed for the mock? If so could you please adjust that in the makefile (ill push it later) If I adjusted it, it threw lots of warnings

extras/mock/preciceMocked.cpp
JulianHarrison-Wirth: there was a moment in time where the mock had more than one thread, but that feature (where future was used) has since been removed. But unless there is a reason to remove it I don't think there is a reason to remove the mutexes

extras/mock/preciceMocked.cpp
JulianHarrison-Wirth: Because asserts are removed at compile-time and the mock is built to catch any edge cases is there any harm in keeping it in?

extras/mock/preciceMocked.cpp
JulianHarrison-Wirth: I have implemented it using a SAX handler now

extras/mock/preciceMocked.cpp
JulianHarrison-Wirth: I refactored the code to hopefully be more readable

JulianHarrison-Wirth:
This is the final big version unless you can find any problems or have any suggestions @MakisH @fsimonis 
@fsimonis I answered all of your suggestions so far, in some I still have open questions, if you could please let me know if any of those still need addressing that would be amazing

JulianHarrison-Wirth:
I will have one more large update cleaning up and expanding the readme and the comments in the mock, but i will do that while I work through the code for my thesis

extras/mock/preciceMocked.cpp
fsimonis: Change it for the mock only for now. preCICE compiled with C++20 after the recent fixes on develop.

fsimonis: Change it for the mock only for now. preCICE compiled with C++20 after the recent fixes on develop.

JulianHarrison-Wirth: Ok, I think I was able to avoid the warnings by disabling some optimizations for the compilation of the mock, so now only the mock runs on c++20

extras/mock/preciceMocked.cpp
fsimonis: Less to type and easier to follow. We also do this in the main library.

extras/mock/preciceMocked.cpp
fsimonis: > But unless there is a reason to remove it I don't think there is a reason to remove the mutexes

As there is a reason to add it to the main library, there is no reason to remove it from the mock :wink:

extras/mock/preciceMocked.cpp
fsimonis: Very nice

extras/mock/preciceMocked.cpp
fsimonis: The type system already requires the constructor to be called before the member functions. So there is no valid case where this could become a problem.

extras/mock/preciceMocked.cpp
MakisH: In my perspective, it is confusing having some mutexes here and there, but not having a consistently thread-safe implementation. It does not harm anyone in the mock, but a comment on the intention would be helpful (guarding against what possibility? Why is this here and not in the main API?).

Since both @fsimonis and me stumbled upon this, a comment might be helpful.

extras/mock/run_course_solutions.sh
MakisH, line 1: In the final code, these two files (`run_course_solitions.sh` and `run_all_tutorials.sh`) should not be there. Happy to have them submitted as part of your thesis, but I would not relate this repository with the training or tutorials.

CMakeLists.txt
```diff
@@ -78,7 +78,7 @@ option(PRECICE_RELEASE_WITH_DEBUG_LOG "Enable debug logging in release builds" O
 option(PRECICE_RELEASE_WITH_TRACE_LOG "Enable trace logging in release builds" OFF)
 option(PRECICE_RELEASE_WITH_ASSERTIONS "Enable assertions in release builds" OFF)
 
-set(PRECICE_CXX_STANDARD "17" CACHE STRING "CXX Standard for all preCICE targets" )
+set(PRECICE_CXX_STANDARD "20" CACHE STRING "CXX Standard for all preCICE targets" )
```
MakisH: Reminder: This PR increases the C++ standard version for all targets.
We should probably do this independently on develop first.

Edit: Our baseline is currently Ubuntu 22.04. According to release schedules, the baseline bump to 26.04 would happen after preCICE v3.4.0.

MakisH: Building currently fails on Ubuntu 22.04 - GCC 12.3 (both in the CI and on my system) with:

```
 [2/66] Building CXX object CMakeFiles/preciceMocked.dir/extras/mock/preciceMocked.cpp.o
FAILED: CMakeFiles/preciceMocked.dir/extras/mock/preciceMocked.cpp.o 
/usr/bin/g++ -DEIGEN_INITIALIZE_MATRICES_BY_NAN -DPRECICE_NO_MPI -D_GLIBCXX_ASSERTIONS -DpreciceMocked_EXPORTS -I/__w/precice/precice/src -I/__w/precice/precice/build/src -I/__w/precice/precice/extras/bindings/c/include -I/__w/precice/precice/extras/bindings/fortran/include -isystem /usr/include/eigen3 -isystem /usr/include/libxml2 -Wall -Wextra -Wno-unused-parameter --coverage -g -fPIC -std=gnu++20 -MD -MT CMakeFiles/preciceMocked.dir/extras/mock/preciceMocked.cpp.o -MF CMakeFiles/preciceMocked.dir/extras/mock/preciceMocked.cpp.o.d -o CMakeFiles/preciceMocked.dir/extras/mock/preciceMocked.cpp.o -c /__w/precice/precice/extras/mock/preciceMocked.cpp
/__w/precice/precice/extras/mock/preciceMocked.cpp:3:10: fatal error: format: No such file or directory
    3 | #include <format>
      |          ^~~~~~~~
```

I think we need at least GCC 13 for that: https://en.cppreference.com/w/cpp/compiler_support/20.html

extras/mock/preciceMocked.cpp
JulianHarrison-Wirth: ok updated

extras/mock/preciceMocked.cpp
JulianHarrison-Wirth: removed them

MakisH:
The CI currently fails with:

```
 /__w/precice/precice/extras/bindings/fortran/src/preciceFortran.cpp:79:3: error: 'PRECICE_WARN' was not declared in this scope; did you mean 'PRECICE_API'?
   79 |   PRECICE_WARN("preCICE was configured without MPI but you passed an MPI communicator. preCICE ignores the communicator and continues.");
      |   ^~~~~~~~~~~~
      |   PRECICE_API
```

This seems to be originating from https://github.com/precice/precice/commit/c85140962b6ed0f34b832a94975d4d34cae41f67 (skeleton). We simply need to convert the `PRECICE_WARN` similarly to the `PRECICE_CHECK`.

extras/mock/precice-mock-config.xml
MakisH, line 1: Hopefully trivial change, but useful for clarity: let's rename `mock-config.xml` to `precice-mock-config.xml`.

This is for consistency with all other output files, for example.

JulianHarrison-Wirth:
@fsimonis If you could look at the build issues on mac please, it seems to be a cmake issue

extras/mock/README.md
```diff
@@ -0,0 +1,276 @@
+# preCICE Mock Participant
```
MakisH: Let's integrate this page to the website.
I can prepare the technicalities.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+    const auto &meshInfo = meshEntry.second;
+    if (!meshInfo.provided) {
+      continue;
+    }
+
+    if (_impl->primary) {
+      std::cout << "---[precice]  Prepare partition for mesh " << meshInfo.name << std::endl;
+      std::cout << "---[precice]  Gather mesh " << meshInfo.name << std::endl;
+      std::cout << "---[precice]  Send global mesh " << meshInfo.name << std::endl;
+    }
```
uekerman: IMO, we don't need such logging. Better specify what the mock is doing.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+double Participant::getMaxTimeStepSize() const
+{
+  std::lock_guard<std::recursive_mutex> lk(_impl->mtx);
+  if (_impl->finalized) {
+    throw precice::Error("getMaxTimeStepSize() cannot be called after finalize().");
+  }
+  if (!_impl->initialized) {
+    throw precice::Error("initialize() has to be called before getMaxTimeStepSize() can be evaluated.");
+  }
+  return _impl->timeWindowSize;
```
uekerman: Here, we should not return the complete time window size, but the remainder in the time window.
In `advance`, you need to sum up `computedTimeStepSize`.
Note that in an implicit coupling scheme, we are repeating the complete time window (not only the current time step).
This page could help: https://precice.org/couple-your-code-time-step-sizes.html

uekerman (commented):
Some quick comments already. I still want to try out the mock actually.

uekerman:
I am getting building errors like:

```
/home/uekermbn/Repos/precice/extras/bindings/fortran/src/preciceFortran.cpp:35:6: error: ‘string_view’ in namespace ‘std’ does not name a type
   35 | std::string_view strippedStringView(const char *string, int length);
      |      ^~~~~~~~~~~
/home/uekermbn/Repos/precice/extras/bindings/fortran/src/preciceFortran.cpp:35:1: note: ‘std::string_view’ is only available from C++17 onwards
   35 | std::string_view strippedStringView(const char *string, int length);
      | ^~~
```

This does not happen on the current state of `develop`.
I am admittedly on an old system, using gcc 9.4, which should however support C++-17.

extras/mock/precice-mock-config.xml
```diff
@@ -0,0 +1,40 @@
...
+      If specified, this value will be used instead of the max-iterations from the coupling-scheme.
+      Use -1 to respect the preCICE config value. If the override is higher than the config
+      value, a warning will be printed but the override will still be used. -->
+  <max-iterations-override value="2" />
+
+  <!-- GLOBAL DEFAULT (optional) -->
+  <!-- If specified, this mode applies to all data items not explicitly configured.
+       If omitted, defaults to buffer mode. -->
+  <mocked-data-default mode="scaled">
+    <!-- Optional: default scalar multiplier for scaled mode -->
```
uekerman:
```suggestion
    <!-- Optional: scalar multiplier for scaled mode -->
```
The "default" here confused me for a moment. Maybe fewer misunderstandings like this.

extras/mock/precice-mock-config.xml
```diff
@@ -0,0 +1,40 @@
...
+  <mocked-data mesh="MeshTwo" data="Velocity" mode="scaled">
+    <!-- For 3D vector data: multiply component-wise -->
+    <vector-multiplier values="1.0;2.0;3.0" />
+  </mocked-data>
+
+  <!-- Example 5: Multiple scalar values can use semicolons too -->
+  <mocked-data mesh="MeshThree" data="Heat-Flux" mode="scaled">
+    <!-- Will cycle through multipliers for each value -->
+    <vector-multiplier values="0.5;1.0;1.5;2.0" />
+  </mocked-data>
```
uekerman: I'm not getting the main comment here.
And I am neither sure whether this feature is really needed. Is there a use case I am not seeing?
I would simply expect an error if the vector multiplier has the wrong length.

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+
+### 1. Random Mode
+Returns random seeded data (useful for testing error handling). Optional bounds and seed can be specified with nested elements (bounds defaults: 0.0 to 1.0, seed defaults to rank-based value).
+
+```xml
+<mocked-data mesh="MeshName" data="DataName" mode="random">
+  <bounds lower="-1.0" upper="1.0" />
+  <seed value="42" />
+</mocked-data>
+```
```
uekerman, line 170: Could be worth integrating in the example config.

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+
+### 2. Buffer Mode (Default)
+Returns the data previously written via `writeData()`. This is the default if no mock-config is provided.
+
+```xml
+<mocked-data mesh="MeshName" data="DataName" mode="buffer" />
+```
```
uekerman: Maybe make this the first item in the list.

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+</mocked-data>
+```
+
+### Global Default Configuration
+
+You can set a global default mode and multipliers that apply to all data items not explicitly configured:
+
+```xml
+<mock-config>
+  <!-- Set default mode for all unmapped data -->
```
uekerman: Why "unmapped"?

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+  <mocked-data mesh="MeshOne" data="SpecialData" mode="random">
+    <bounds lower="0.0" upper="5.0" />
+    <seed value="123" />
+  </mocked-data>
+</mock-config>
+```
+
+In this example:
+- `SpecialData` uses random mode with custom bounds and seed
+- Random mode honors optional `bounds` element (defaults to 0.0-1.0) and optional `seed` element (defaults to rank-based)
```
uekerman: What do you mean with "rank-based"?

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+
+**Explicit Coupling:**
+- Termination is checked after each time window
+- Simulation ends when max-time or max-time-windows is reached
+
+**Implicit Coupling:**
+- `max-time-windows`: Terminates after completing the specified number of time windows (each time window may involve multiple sub-iterations)
+- `max-time`: Terminates at the end of a time window when the accumulated time reaches or exceeds the limit (does not terminate during sub-cycling)
+
+Both coupling schemes will throw an error if neither termination criterion is provided.
```
uekerman: I think that all of this is not needed. This is normal preCICE behavior.

Maybe one overall h2 section "Configuration" could be a better structure?

extras/mock/README.md
uekerman: I see the point, but I don't think we need such a feature.
While developing an adapter, I want to mainly check whether I call set_mesh_vertices or read/write_data correctly. For all of these, I need names.

We already now have enough tooling to easily generate a first config.

And developing an adapter without targeting a specific application is nothing we have seen so far.

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+- Validate that the participant name exists
+- Check mesh and data declarations
+- Enforce read/write permissions
+- Detect implicit vs explicit coupling
+- Return correct dimensions for meshes and data
+
+### Mock Configuration (Optional)
+Place `precice-mock-config.xml` in the same directory as your preCICE config file.
+
+**Example precice-mock-config.xml:**
```
uekerman: The config is also included here. No need to duplicate. We only have to document where to find the example.

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+}
+```
+
+## Implicit Coupling
+
+For implicit coupling schemes, the mock automatically:
+- Returns `true` from `requiresWritingCheckpoint()` at the start of each iteration
+- Returns `true` from `requiresReadingCheckpoint()` when iterating (not converged)
+- Tracks iteration count and convergence status
+- Reports time window completion correctly
```
uekerman: That's all expected behavior, does not need documentation IMO.

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Building
```
uekerman: I would move this to the very top.

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+
+The mock is built as part of the preCICE library.
```
uekerman: Can I only build the mock?

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+
+Build normally and use LD_PRELOAD to load the mock at runtime
+
+```
+LD_PRELOAD=/path/to/precicedir/precice/build/libpreciceMocked.so.3.3.0
+```
```
uekerman: Good could be some examples here. Running FEniCS as a Python code and running an OpenFOAM case via a run.sh script.

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+target_link_libraries(your_solver preciceMocked)
+```
+## Error Handling
+
+The mock validates all API calls against the configuration and throws `precice::Error` with descriptive messages:
+
+```
+Error: Data 'Temperature' on mesh 'FluidMesh' is not configured for writing by participant 'SolverOne'.
+Please add <write-data name="Temperature" mesh="FluidMesh" /> to the configuration.
+```
```
uekerman: All such things could go into a "Behavior" or "Features" h2 section.

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+
+This helps catch configuration mistakes early in development.
+
+## Limitations
+
+- Does not perform actual mesh mapping or data interpolation
```
uekerman: Not a limitation IMO. I would not expect this from a mock.

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+- Does not communicate between participants (single-process only)
```
uekerman: "between participants" will lead to confusions.
"single-process only" is also not true, right? You have tried to run with a parallel participant. Let's maybe try to boil it down here to the actual limitation (cannot handle MPI communicator in constructor? or maybe there is none in  the end?).

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+- Simplified convergence logic for implicit coupling (converges after N iterations)
```
uekerman: Also not a limitation IMO

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+- No gradient data support (returns zeros)
+- No mesh connectivity validation
```
uekerman: Maybe those are actual limitations. Could you add more details? set_edges etc are nops? Why "returns zero"? There is no read_gradient_data in preCICE (yet).

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+
+## Testing Your Adapter
+
+1. Create a simple preCICE config with your participant
+2. Optionally, create a 'precice-mock-config.xml' for specific data behavior
+3. Run your adapter against the mock
+4. Verify correct API usage patterns
+5. Test error handling by introducing configuration mistakes
+6. Validate checkpoint handling for implicit coupling
```
uekerman: This could be a "how to use" section, but there is a duplication with the "Building" above.

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+
+The mock is ideal for:
+- Unit testing adapters without running full simulations
+- Testing error handling code paths
+- Developing adapters without a coupling partner
+- CI/CD integration testing
```
uekerman: I am not sure if this helps much. Better could be to have examples here (at some point).

extras/mock/README.md
```diff
@@ -0,0 +1,286 @@
...
+
+
+## Recommendations
+
+For practitioners adopting the mock:
+  - Use the mock as the default testing environment during development for fast feedback loops.
+  - Integrate mock-based tests into CI pipelines to enable tight feedback on commits.
+  - Augment mock tests with full-stack validation before production deployment.
+  - Exploit mock configuration files for scenario-based robustness testing.
+  - Keep the mock synchronized with preCICE releases.
```
uekerman: Duplication or anyway clear IMO

.gitignore
```diff
@@ -25,3 +25,9 @@ src/precice/Version.h
 src/testing/SourceLocation.hpp
 .idea
 CMakeUserPresets.json
+examples/tutorials
+examples/course
+.gitignore
+examples/solverdummies/cpp/precice-profiling
+examples/solverdummies/cpp/Testing/Temporary
+.venv
```
uekerman: Probably not sth we want to merge?

CMakeLists.txt
```diff
@@ -576,6 +576,18 @@ target_include_directories(preciceMocked PUBLIC
   $<INSTALL_INTERFACE:include>
   )
 
+target_link_libraries(preciceMocked PUBLIC Eigen3::Eigen)
```
uekerman, line 579: Should we document the dependencies of the mock somewhere. All are dependencies of preCICE as well, but it's a proper subset. You could link to the dependencies of preCICE and then state which ones are needed.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+#include <random>
+#include <string>
+
+#include <libxml/SAX2.h>
+
+#include <precice/Exceptions.hpp>
+#include <precice/Tooling.hpp>
+#include <precice/Types.hpp>
+#include <precice/span.hpp>
+#include "utils/fmt.hpp"
```
uekerman, line 30: A good point for the documentation could be to write down what from preCICE the mock uses.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+  struct ConfigData {
+    std::map<std::string, MeshInfo> meshes;
+    std::vector<DataInfo>           dataItems;
+    bool                            isImplicitCoupling    = false;
+    bool                            hasConvergenceMeasure = false;
+    int                             maxIterations         = 1;
+    int                             configMaxIterations   = 1; // Stores the value from preCICE config
+    int                             maxIterationsOverride = 2; // Default override (2), -1 use config, >0 override
+    int                             currentIteration      = 0;
+    bool                            iterationConverged    = false;
```
uekerman: Do these two really fit here? Aren't these part of the "internal state" below?

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+    std::map<std::string, MockDataConfig> dataConfigs; // Key: "meshName:dataName"
+    DataMode                              defaultMode             = DataMode::Buffer;
+    double                                defaultScalarMultiplier = 1.0;
+    double                                defaultRandomLower      = 0.0;
+    double                                defaultRandomUpper      = 1.0;
+    uint32_t                              defaultRandomSeed       = 0; // 0 means use default seed from rank
+    std::vector<double>                   defaultVectorMultiplier;     // Empty means use scalar
+  };
+
+  // Minimal internal state
```
uekerman:
```suggestion
  // Internal state
```

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+  std::string                  name;
```
uekerman:
```suggestion
  std::string                  participantName;
```
?

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+  std::string                  config;
+  std::string                  mockConfigPath;
+  int                          rank            = 0;
+  bool                         primary         = false;
```
uekerman: Could you document here why this is needed?

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+  MockConfig mockConfig;
+  bool       mockConfigParsed = false;
+  bool       mockConfigExists = false;
+
+  // Write data buffer shared across all meshes/data names
+  std::vector<double> writeBuffer;
+  // Track last written value count per mesh/data for read-size validation
+  std::map<std::string, std::size_t> lastWriteSizes;
+  // Vertex counts per mesh to detect empty provided meshes
+  std::map<std::string, std::size_t> meshVertexCounts;
```
uekerman: Why is the first one needed? Shouldn't the second one be enough?

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+
+  enum class DataType { Scalar = 1,
+                        Vector = 3 };
+
+  // SAX parsing state for config
+  struct ConfigParseState {
+    bool                            inParticipant = false;
```
uekerman:
```suggestion
    bool                            inParticipant = false; // Track if we're inside a participant element
```

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+  }
+  if (_impl->initialized) {
+    throw precice::Error("initialize() may only be called once.");
+  }
+
+  // Print startup banner (only on rank 0)
+  if (_impl->primary) {
+    std::cout << "---[precice]  This is preCICE version 3.3.0 (mock)" << std::endl;
+    std::cout << "---[precice]  Revision info: mock-implementation" << std::endl;
+    std::cout << "---[precice]  Build type: Release (mock)" << std::endl;
```
uekerman: Let's not hard-code the preCICE version and the build type here. Get from cmake configuration?

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+    try {
+      std::cout << "---[precice]  Working directory \"" << std::filesystem::current_path().string() << "\"" << std::endl;
+    } catch (std::filesystem::filesystem_error &fse) {
+      std::cout << "---[precice]  Working directory unknown due to error \"" << fse.what() << "\"" << std::endl;
+    }
+    if (_impl->mockConfigExists) {
+      std::cout << "---[precice]  Configuring preCICE (mock) with configuration \"" << std::filesystem::absolute(_impl->config).string() << "\" and mock configuration \"" << std::filesystem::absolute(_impl->mockConfigPath).string() << "\"" << std::endl;
+    } else {
+      std::cout << "---[precice]  Configuring preCICE (mock) with configuration \"" << std::filesystem::absolute(_impl->config).string() << "\" and default mock configuration. See the mock-README about how to adjust this." << std::endl;
```
uekerman: mock README will become the documentation of the mock. https://precice.org/tooling-mock.html maybe?

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+  }
+
+  // Validate that termination criteria are provided
+  if (_impl->configData.maxTime <= 0 && _impl->configData.maxTimeWindows <= 0) {
+    throw precice::Error(
+        "The preCICE configuration must specify either max-time or max-time-windows in the coupling-scheme \n"
+        "to prevent infinite execution. Please add one of these attributes:\n"
+        "  <max-time value=\"...\"/>\n"
+        "  <max-time-windows value=\"...\"/>");
+  }
```
uekerman: Does `precice-cli config check` also catch this? If yes, you could also simply state such things as a prerequisite, i.e. the config needs to be valid.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+  if (!_impl->initialized) {
+    throw precice::Error("initialize() has to be called before requiresReadingCheckpoint().");
+  }
+
+  // For implicit coupling, require reading checkpoint when not converged
+  if (_impl->configData.isImplicitCoupling) {
+    // Read checkpoint needed when iteration hasn't converged and we're past iteration 0
+    return _impl->configData.currentIteration > 0 && !_impl->configData.iterationConverged;
+  }
+
```
uekerman, line 1783: will become trickier when supporting subcycling

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+            return meshes.empty() ? "none" : meshes;
+          }()));
+    }
+  }
+
+  // If configuration wasn't parsed or mesh wasn't found in parsed config,
+  // try to return a previously inferred mesh dimension (e.g. from earlier
+  // setMeshVertex/setMeshVertices calls). If none is available, return 0
+  // to indicate unknown dimensionality so higher-level bindings can infer
+  // dimensions from the provided vertex data (matching real preCICE behavior).
```
uekerman: I am not sure I understand this part. If the mesh exists, there should also be a dimension available?

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+    }
+
+    // Data not found - throw error
+    throw precice::Error(precice::utils::format_or_error(
+        "Data '{}' on mesh '{}' is not used by participant '{}'",
+        dataNameStr, meshNameStr, _impl->name));
+  }
+
+  // Fallback: mock as scalar
+  return 1;
```
uekerman: There should be an actual dimension, no need for a fallback?

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+  if (_impl->initialized) {
+    throw precice::Error("setMeshVertex() cannot be called after initialize(). Mesh modification is only allowed before calling initialize().");
+  }
+  std::string meshNameStr(meshName.data(), meshName.size());
+  // Determine expected dimensionality from configuration if available
+  int  expectedDims = 3;
+  auto meshIt       = _impl->configData.meshes.find(meshNameStr);
+  if (meshIt != _impl->configData.meshes.end()) {
+    expectedDims = meshIt->second.dimensions;
+  }
```
uekerman: Why not throwing an error here?

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+
+int Participant::getMeshVertexSize(::precice::string_view meshName) const
+{
+  std::lock_guard<std::recursive_mutex> lk(_impl->mtx);
+  std::string                           meshNameStr(meshName.data(), meshName.size());
+  auto                                  it = _impl->meshVertexCounts.find(meshNameStr);
+  if (it != _impl->meshVertexCounts.end()) {
+    return static_cast<int>(it->second);
+  }
+  return 0;
```
uekerman: I would also throw an error here.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+  auto meshIt       = _impl->configData.meshes.find(meshNameStr);
+  if (meshIt != _impl->configData.meshes.end()) {
+    expectedDims = meshIt->second.dimensions;
+  }
+  // Expect coordinates as expectedDims * N entries for N VertexIDs
+  if (coordinates.size() != ids.size() * expectedDims) {
+    throw precice::Error(precice::utils::format_or_error("setMeshVertices() was called with {} vertices and {} coordinates ({}D), but needs {} coordinates ({} x {}).", ids.size(), coordinates.size(), expectedDims, ids.size() * expectedDims, ids.size(), expectedDims));
+  }
+  _impl->meshVertexCounts[meshNameStr] += ids.size();
+}
```
uekerman, line 1967: ids is not filled here. I think this could be helpful. And then check in read/write_data.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+// Data access
+
+bool Participant::requiresInitialData()
+{
+  std::lock_guard<std::recursive_mutex> lk(_impl->mtx);
+  if (_impl->initialized) {
+    throw precice::Error("requiresInitialData() has to be called before initialize().");
+  }
+  return false;
+}
```
uekerman, line 2258: This could be sth to read from the preCICE config. Otherwise, document as limitation.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+  if (!_impl->initialized) {
+    throw precice::Error("initialize() has to be called before writeAndMapData().");
+  }
+  // coordinates are 3*N, values are N for scalar data
+  if (coordinates.size() % 3 != 0) {
+    throw precice::Error(precice::utils::format_or_error("writeAndMapData() was called with {} coordinates, but needs a multiple of 3 coordinates.", coordinates.size()));
+  }
+  if (values.size() != coordinates.size() / 3) {
+    throw precice::Error(precice::utils::format_or_error("writeAndMapData() was called with {} coordinates and {} values, but needs {} values ({} / 3).", coordinates.size(), values.size(), coordinates.size() / 3, coordinates.size()));
+  }
```
uekerman: the "3" should actually be the mesh size. If 2D is configured, you will errors here.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+// Direct access
+
+void Participant::setMeshAccessRegion(
+    ::precice::string_view /*meshName*/,
+    precice::span<const double> boundingBox) const
+{
+  std::lock_guard<std::recursive_mutex> lk(_impl->mtx);
+  if (!_impl->initialized) {
+    throw precice::Error("initialize() has to be called before setMeshAccessRegion().");
+  }
```
uekerman: No, exactly the other way around. https://precice.org/couple-your-code-direct-access.html

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+  // bounding box: min/max for each dimension -> 3D: 6 entries
+  if (boundingBox.size() != 6) {
+    throw precice::Error(precice::utils::format_or_error("setMeshAccessRegion() was called with {} bounding box coordinates, but expects 6 coordinates (min and max for each dimension in 3D).", boundingBox.size()));
+  }
```
uekerman: depends on the mesh dimension. more similar issues below.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+    precice::span<const double>   gradients)
+{
+  std::lock_guard<std::recursive_mutex> lk(_impl->mtx);
+  if (!_impl->initialized) {
+    throw precice::Error("initialize() has to be called before writeGradientData().");
+  }
+  // gradients expected as 3 components per id for vector gradient in 3D
+  if (gradients.size() != ids.size() * 3) {
+    throw precice::Error(precice::utils::format_or_error("writeGradientData() was called with {} vertices and {} gradient components, but needs {} components ({} x 3).", ids.size(), gradients.size(), ids.size() * 3, ids.size()));
+  }
```
uekerman: depends on mesh and data size

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+  // no-op
+}
+
+// Profiling
+
+void Participant::startProfilingSection(::precice::string_view /*sectionName*/)
+{
+  // no-op
+}
```
uekerman: add to limitations. actually, it would not be complicated to check a few things here. e.g. that `stopLastProfilingSection` cannot be called without calling `startProfilingSection` first.

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+{
+  out << "preCICE mock config reference; type=" << static_cast<int>(reftype) << "\n";
+}
+
+void checkConfiguration(const std::string &filename, const std::string &participant, int size)
+{
+  throw precice::Error(precice::utils::format_or_error(
+      "---[precice]  precice mock: checkConfiguration called for file '{}' participant '{}' size {}",
+      filename, participant, size));
+}
```
uekerman, line 2839: Add to limitations? I guess you need these for ABI compatibility, right?

extras/mock/preciceMocked.cpp
```diff
@@ -0,0 +1,1796 @@
...
+
+} // namespace tooling
+
+std::string getVersionInformation()
+{
+  return std::string("precice-mock;revision=mock;features=MPI:OFF,PETSC:OFF,PYTHON:OFF");
```
uekerman: Let's not hard-code such things. Could still be interesting to have the actual data here.

uekerman (changes requested):
I had closer look from the user perspective. I did not look at the implementation of the configuration.

Besides the points below:
- Let's try to condense the documentation to the important things.
- We are building the mock as part of the CI already, right? Could be good to also add some smoke test (e.g. solver dummies)
- As mentioned at the presentation: An interesting test would be to run all preCICE integration tests and see what happens (crash vs. expected errors vs. unexpected errors). For example, the missing support for subcycling should be caught like this. 
- To what extend does the mock support [remeshing](https://precice.org/couple-your-code-moving-or-changing-meshes.html#remeshing-using-precice)?

fsimonis:
FYI, this is still on my TODO list. I'll get to it.

uekerman:
@JulianHarrison-Wirth Thank you for the update. As this is a huge PR, which becomes hard to keep an overview of. To mitigate, I suggest resolving "done" comments and keeping commits small: one commit per non-trivial change.

In the current state, I don't believe that subcycling already works. The handling in `advance` still seems wrong. You should be able to easily test this with a solverdummy that subcycles. Maybe unclear: With subcycling we simply mean that a participant uses a time step size smaller than the time window size. More information in the [docs on time stepping](https://precice.org/couple-your-code-time-step-sizes.html). Please ask for help if unclear.

extras/mock/README.md
JulianHarrison-Wirth: do you need me to do anything here or is this just waiting for the final merge?

extras/mock/README.md
JulianHarrison-Wirth: do you mean rebuild the mock after changes or build the mock without any precice components available?

uekerman: To only build the mock, but not libprecice, testprecice, ...
```
make mock
```

extras/mock/preciceMocked.cpp
JulianHarrison-Wirth: this is the only comment I have not addressed as I don't know where it will end up

MakisH: @JulianHarrison-Wirth I will integrate this repository to the preCICE website, and the `README.md` will be rendered. For this, you will need to add a preamble, similar to https://github.com/precice/precice.github.io/blob/master/content/docs/tooling/tooling-cli.md?plain=1

I agree with `tooling-mock.html` as the target.

Once the preamble is there, you don't need to do anything else - I will take care of the integration on the website side.

extras/mock/README.md
MakisH: The file mainly needs a preamble/frontmatter. See https://github.com/precice/precice/pull/2385#discussion_r2863478044

I can do the integration with the website after merging.

extras/mock/tests/tutorials/baselines/courses/B.out
MakisH, line 1: I need to catch up with the latest state, but I think we should separate here between test cases that are available in this repository, and test cases that are maintained separately. For this reason, I would remove the tutorials and the course outputs from this PR. They are nice showcases for your thesis, but they will immediately become outdated.

Later on, some of these could be moved to the system tests (tutorials repository), if there is an archive.

MakisH: @JulianHarrison-Wirth if you removed the extra files from here, we could more easily continue with the review. If there are any further changes locally, please don't hesitate to push.

The feature freeze for v3.5.0 is on Friday.

JulianHarrison-Wirth:
All comments should be resolved

fsimonis:
I'll integrate this when I find time mid November

Issue #1607, opened 2023-03-23
Change the local Data/Gradient representation


fsimonis:
**Please describe the problem you are trying to solve.**

Currently, we save the values and gradients of a data in a component-oriented way.

Let 
* `msize` be the mesh vertex count
* `mdim` be the mesh dimensionality 2/3-D
* `ddim` be the data dimensionality 1D (Scalar) 2/3-D Vectorial

We currently save:
* values as a vector of components `msize * ddim`
* gradients as a matrix of gradients per component `(msize * ddim) x (mdim)`

This requires explicitly handling the various dimensionalities dynamically using loops for components.

Also, it isn't possible to deduce mesh size or data dimension based on the values alone.
We always need a `Mesh` where we want to process `Data::values`.

**Describe the solution you propose.**

Save using the following layout:
* values as a matrix of data vectors per vertex `msize x ddim` 
* gradients as a matrix of gradient/Jacobian per vertex `msize x (ddim * mdim)`

Note that the physical data layout doesn't change, only the logical layout/access pattern.

This has the following upsides:
* Values and Gradient now explicitly encode the amount of mesh vertices, as well as the dimensionality of the mesh and data, removing the need to pass Mesh or Data pointers to many places.
* We can get rid of many loops and `std::copy_n`, which handle dimensionality. Eigen handles and checks assignments for us!
* The interface of read/write data can now use passed buffers as `Eigen::Matrix` using an `Eigen::Map` and directly copy data vertex by vertex, simplifying the API a lot.
* Removing the need to pass Data and Mesh items around, allows simplifying function interfaces, which reduces boilerplate to test them.

**Describe alternatives you've considered**
_none_

**Additional context**

fsimonis:
You are right, I got the dimensions wrong…

We could store
* values as `MatrixXd(ddata, msize)`
* gradients/jacobians as `MatrixXd(ddata*dmesh, msize)`

This way
```cpp
values.row(vertexID).size()    == ddata
gradients.row(vertexID).size() == ddata * dmesh
```

davidscn:
> You are right, I got the dimensions wrong…

I don't think so. Values would be stored as a matrix with msize rows and ddata columns, hence `MatrixXd(msize, ddata)` (corresponding to the `column-major` scheme, we would use a `ddata-major` scheme).

fsimonis:
True, data/gradients/Jacobians go into columns.

PR #2389, opened 2025-10-30
Matrix solver interface

FabioTucciarone:
## Main changes of this PR

I changed the interface of the solver methods so that they use `Eigen::MatrixXd` instead of `Eigen::VectorXd` as mentioned in the [todo comment](https://github.com/precice/precice/blob/1bef5d979e84cd350195caca8de0a1b3b666cad7/src/mapping/RadialBasisFctSolver.hpp#L440) in `RadialBasisFctSolver.cpp`:
```
// @todo: change the signature to Eigen::MatrixXd and process all components at once, the solve function of eigen can handle that
```

The main motivation for doing so is the RBF Tuner from https://github.com/precice/precice/pull/2340. Having a matrix interface would open the possibility to optimize for all vector components at the same time.

Some problems I encountered while implementing this change that might need to be discussed:

1. There is currently a lot of copying going on, mainly because of the row-major storage format used by the mapping functions, while the the solver methods use column-major.
```c++
RadialBasisFctMapping<SOLVER_T, Args...>::mapConsistent():

// copy of input data to Eigen matrix format
MatrixXd in = Eigen::Map<RowMatrixXd, ...>(globalInValues.data(), ...);
// copy col-major to row-major format
RowMatrixXd out = _rbfSolver->solveConsistent(in, _polynomial); 
// copy mapped data at correct position to output data values
outData = Eigen::Map<Eigen::VectorXd>(out.data(), outValuesSize.at(0));
```
This seems unnecessary, but I had difficulties finding a sensible solution that works in the distributed case. Maybe we could try using row-major in the solver?

2. For the Ginkgo modifications I simply moved the loop, looping over the data dimensions into the solver function.
   This is because I had difficulties getting Ginkgo to work on my machine and was thus unable to test more complex modifications, as I had to rely on the `_solveRBFSystem()` function.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/2389.md
+++ b/docs/changelog/2389.md
@@ -1,0 +1,1 @@
+- Refactored the RadialBasisFctSolver interface from a vector to a matrix interface, such that all vector components are computed in one function call.


src/mapping/impl/SphericalVertexCluster.hpp
```diff
@@ -313,20 +313,22 @@ void SphericalVertexCluster<RADIAL_BASIS_FUNCTION_T>::mapConsistent(const time::
...
-    // Step 2: solve the system using a consistent constraint
-    auto result = _rbfSolver.solveConsistent(in, _polynomial);
-    PRECICE_ASSERT(static_cast<Eigen::Index>(_outputIDs.size()) == result.size());
+  // Step 2: solve the system using a consistent constraint
+  auto result = _rbfSolver.solveConsistent(in, _polynomial);
+  PRECICE_ASSERT(static_cast<Eigen::Index>(_outputIDs.size() * nComponents) == result.size());
 
-    // Step 3: now accumulate the result into our global output data
+  // Step 3: now accumulate the result into our global output data
+  for (unsigned int c = 0; c < nComponents; ++c) {
```
davidscn: Could you move the components loop onto the most inner level? Then it should only one line be in the most inner loop.This would make more sense from the way the data is processed.

src/mapping/RadialBasisFctMapping.hpp
```diff
@@ -248,28 +248,16 @@ void RadialBasisFctMapping<SOLVER_T, Args...>::mapConservative(const time::Sampl
...
+    using MapToMatrix = Eigen::Map<RowMatrixXd, Eigen::Unaligned, Eigen::OuterStride<Eigen::Dynamic>>;
 
-    Eigen::VectorXd in;                     // rows == outputSize
-    in.resize(_rbfSolver->getOutputSize()); // rows == outputSize
+    // copy of input data to Eigen matrix format, rows == outputSize
+    Eigen::MatrixXd in = MapToMatrix(globalInValues.data(), _rbfSolver->getOutputSize(), valueDim, Eigen::OuterStride(valueDim));
 
-    outputValues.setZero();
+    // copy all output values without polynomial entries from col-major to row-major format
+    RowMatrixXd out = _rbfSolver->solveConservative(in, _polynomial).block(0, 0, this->output()->getGlobalNumberOfVertices(), valueDim);
```
davidscn, line 258: Why the extra copy here? Wouldnt it be possible to create the `out` Map first and directly assign the return value to this map? Could also require an Eigen::Ref

tests/serial/mapping-rbf-gaussian/helpers.cpp
```diff
@@ -3,38 +3,129 @@
...
+      expectedValues = evaluateFunction(meshTwo);
+    }
+
+    double maxDt = interfaceB.getMaxTimeStepSize();
+
+    Eigen::Vector<double, numOut * dim> values = Eigen::Vector<double, numOut * dim>::Zero();
+    interfaceB.readData(meshTwoID, dataOneID, idsTwo, maxDt, values);
+    interfaceB.advance(maxDt);
+
+    for (size_t i = 0; i < values.size(); i++) {
```
davidscn:
```suggestion
    for (Eigen::Index i = 0; i < values.size(); i++) {
```

src/mapping/RadialBasisFctMapping.hpp
FabioTucciarone: The reason for that is the column major format of the return type of `_rbfSolver->solveConservative`.
This copy was also done in the original code.

I assume it would be possible to only copy the relevant parts into smaller arrays when scattering the data to the other ranks in:

https://github.com/FabioTucciarone/precice/blob/c55ec16adc563f8dec060cfbba211d3b60b77394/src/mapping/RadialBasisFctMapping.hpp#L279-L280

fsimonis:
> There is currently a lot of copying going on, mainly because of the row-major storage format used by the mapping functions, while the the solver methods use column-major.

Can't you use Eigen to transpose the matrix before calling the solver and then transpose the result before assigning? Assuming this information can be passed between functions.

In general, I think documentation is key here. We generally need to make the handling of Data less error-prone. This could possibly include: https://github.com/precice/precice/issues/1607

src/mapping/impl/SphericalVertexCluster.hpp
```diff
@@ -313,20 +313,22 @@ void SphericalVertexCluster<RADIAL_BASIS_FUNCTION_T>::mapConsistent(const time::
     for (unsigned int i = 0; i < _inputIDs.size(); i++) {
```
davidscn:
```suggestion
for (unsigned int i = 0; i < _inputIDs.size(); i++) {
  // Step 1: extract the relevant input data from the global input data and store
  // it in a contiguous array, which is required for the RBF solver (last polyparams entries remain zero)
  for (unsigned int c = 0; c < nComponents; ++c) {
```

davidscn: Putting the components loop as the inner-most loop makes more sense. Before, the entire block was in a for loop over all components.

src/mapping/RadialBasisFctMapping.hpp
davidscn: Ok let's leave it at that. The RBFMapping class is the right place to do this transformation. Eventually, it should happen in the `Participant` class, but that's then future.

PR #2410, opened 2025-12-05
Fix error with Ginkgo mapping and Kokkos 5.0.0

bajfi:
1. Lambda Parameter Binding Errors

- Problem: Using non-const lvalue references (auto &acc) in lambda expressions within std::accumulate calls
- Fix: Changed to pass-by-value (auto acc) or const reference (const auto &val)

2. Ambiguous format_to Function Calls

- Problem: Conflict between fmt::format_to and C++20's std::format_to
- Fix: Explicitly qualified all calls as fmt::format_to

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/2410.md
+++ b/docs/changelog/2410.md
@@ -1,0 +1,1 @@
+- Fixed compilation with C++20 required by Kokkos 5.0.0


src/mapping/RadialBasisFctSolver.hpp
```diff
@@ -109,7 +109,7 @@ inline double computeSquaredDifference(
     v[d] = (u[d] - v[d]) * static_cast<int>(activeAxis[d]);
   }
   // @todo: this can be replaced by std::hypot when moving to C++17
-  return std::accumulate(v.begin(), v.end(), static_cast<double>(0.), [](auto &res, auto &val) { return res + val * val; });
+  return std::accumulate(v.begin(), v.end(), static_cast<double>(0.), [](auto res, auto &val) { return res + val * val; });
```
davidscn: What's the error message the compiler emits here? According to [cppreference](https://en.cppreference.com/w/cpp/algorithm/accumulate.html) the previous signature should be valid

> The signature does not need to have const &.

fsimonis: This means that passing by reference or passing by const ref are allowed.

It allows passing const lvalues and rvalues.

fsimonis:
@bajfi hi. Can you please upload the cmake output of your failing build?
We need to understand how these building failures were triggered to assess whether this requires a hotfix or not.

Currently, we are patching in the darkness.

bajfi:
@fsimonis sure 
[build.log](https://github.com/user-attachments/files/23957692/build.log)

src/mapping/RadialBasisFctSolver.hpp
davidscn: the non-const lvalue should also be possible, I don't see why this is not possible

fsimonis: The spec says `const T&` with optional `const&` so this is what STL implementations rely on. In C++ 20 accumulate calls the binary operation with an `std::move`d argument, which cannot bind against `auto&`, which is why it fails now.

Not following the spec can work, but sometimes it doesn't. The second parameter type is also `auto&` which is also incorrect, but doesn't fail _yet_.

Ref aside, this accumulates `double`s, which fit in registers, so let's pass them by value.
The comment above also suggests that this can be replaced as we are now using C++17.

src/mapping/PartitionOfUnityMapping.hpp
```diff
@@ -214,7 +214,7 @@ void PartitionOfUnityMapping<RADIAL_BASIS_FUNCTION_T>::computeMapping()
   PRECICE_DEBUG("Partition of unity data mapping between mesh \"{}\" and mesh \"{}\": mesh \"{}\" on rank {} was decomposed into {} clusters.", this->input()->getName(), this->output()->getName(), inMesh->getName(), utils::IntraComm::getRank(), _clusters.size());
 
   if (_clusters.size() > 0) {
-    PRECICE_DEBUG("Average number of vertices per cluster {}", std::accumulate(_clusters.begin(), _clusters.end(), static_cast<unsigned int>(0), [](auto &acc, auto &val) { return acc += val.getNumberOfInputVertices(); }) / _clusters.size());
+    PRECICE_DEBUG("Average number of vertices per cluster {}", std::accumulate(_clusters.begin(), _clusters.end(), static_cast<unsigned int>(0), [](auto acc, auto &val) { return acc += val.getNumberOfInputVertices(); }) / _clusters.size());
```
fsimonis: This is still incorrect.

```suggestion
    PRECICE_DEBUG("Average number of vertices per cluster {}", std::accumulate(_clusters.begin(), _clusters.end(), static_cast<unsigned int>(0), [](auto acc, auto val) { return acc + val.getNumberOfInputVertices(); }) / _clusters.size());
```

src/mapping/RadialBasisFctSolver.hpp
fsimonis: Let's fully fix the signature:

```suggestion
  return std::accumulate(v.begin(), v.end(), static_cast<double>(0.), [](auto res, auto val) { return res + val * val; });
```

This could also be a call to `std::inner_product(v.begin(), v.end(), v.begin(), static_cast<double>(0.))` if you want to avoid the lambda.

src/mapping/PartitionOfUnityMapping.hpp
bajfi: Is something error ? Seems it can compiled successfully

src/mapping/RadialBasisFctSolver.hpp
bajfi: This seems better

fsimonis:
I additionally compiled preCICE manually with C++20 and GCC 15.2 as well as Clang 20.1.8.
This PR fixes all compilation problems seen on develop.

I'll merge once the tests run through.

fsimonis:
This will be part of the next hotfix.

PR #2411, opened 2025-12-05
Improve baricentric map and tag

fsimonis:
## Main changes of this PR

This PR optimizes the barycentric map and tagging based on results of a recent benchmarking session.

The map loop now uses a direct loop for 1D, templatized segment for 2D, and MatrixX3d with row assignment for 3D.
These turned out to be the optimal choices for meshes of 1M in (~3M triangles) and 1M out vertices. The 3D case could also use the templates segment version, which avoids the map. This we can still revert if not worth the runtime difference.

image

The tagging builds and then uses a `std::vector<bool>` as a dynamic bitmap instead of a set, which is the fastest of all variants.

image

Keep the mapping kernels fast, especially as we now exchange substeps by default in implicit schemes.

The tagging itself is dwarfed by the computation of the mapping, but this allows us to focus on the computation itself in the future.
For a simple NP case using ASTE I got the following timings in ms:

| Tagging Method | tagging total | compute | tag |
--- | --- | --- | ---
current |2 284 | 2 206 | 78
simple loop |2 236 | 2 203 | 32
vector bool| 2 220 | 2 213 | 6.79

## Author's checklist

### DIFF
--- a/docs/changelog/1952.md
+++ b/docs/changelog/1952.md
@@ -1,0 +1,2 @@
+- Added configurable lower and upper bounds for scalar and vector data.
+- Added post-processing methods to quasi-Newton acceleration to handle steps that would violate defined data bounds. Options for `on-bound-violation` are `discard`, `clamp` or `scale` QN steps. Default is to `ignore` violations.
--- a/docs/changelog/2448.md
+++ b/docs/changelog/2448.md
@@ -1,0 +1,1 @@
+- Improved config reading to use a single call if possible. This avoids the partially multi-second latency on slow filesystems used on clusters and supercomputers.
--- a/src/acceleration/Acceleration.hpp
+++ b/src/acceleration/Acceleration.hpp
@@ -24,6 +24,14 @@ public:
   static const int PODFILTER     = 4;
   static const int QR3FILTER     = 5;
 
+  /// Options for handling bound violations
+  enum struct OnBoundViolation {
+    Ignore,
+    Clamp,
+    Discard,
+    ScaleToBound
+  };
+
   /// Map from data ID to data values.
   using DataMap = std::map<int, cplscheme::PtrCouplingData>;
 
--- a/src/acceleration/BaseQNAcceleration.cpp
+++ b/src/acceleration/BaseQNAcceleration.cpp
@@ -40,6 +40,7 @@ BaseQNAcceleration::BaseQNAcceleration(
     int                     filter,
     double                  singularityLimit,
     std::vector<int>        dataIDs,
+    OnBoundViolation        onBoundViolation,
     impl::PtrPreconditioner preconditioner,
     bool                    reducedTimeGrid)
     : _preconditioner(std::move(preconditioner)),
@@ -47,6 +48,7 @@ BaseQNAcceleration::BaseQNAcceleration(
       _maxIterationsUsed(maxIterationsUsed),
       _timeWindowsReused(timeWindowsReused),
       _primaryDataIDs(std::move(dataIDs)),
+      _onBoundViolation(onBoundViolation),
       _forceInitialRelaxation(forceInitialRelaxation),
       _reducedTimeGrid(reducedTimeGrid),
       _qrV(filter),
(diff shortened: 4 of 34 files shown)


src/mapping/BarycentricBaseMapping.cpp
```diff
@@ -70,15 +83,28 @@ void BarycentricBaseMapping::mapConsistent(const time::Sample &inData, Eigen::Ve
...
+  case 1: // Use non-strided access for 1D data
     for (const auto &op : _operations) {
       outValues[op.out] += op.weight * inValues[op.in];
     }
-  } else {
+    return;
+  case 2: // Use templated segment access for 2D
     for (const auto &op : _operations) {
-      outValues.segment(op.out * dimensions, dimensions) += op.weight * inValues.segment(op.in * dimensions, dimensions);
+      outValues.segment<2>(op.out * 2).noalias() += op.weight * inValues.segment<2>(op.in * 2);
```
davidscn: I think an interesting optimization here would be to eliminate this outer for loop altogether. Eigen provides dedicated support for operating on index sets (see [Indexing in the Eigen docs](https://libeigen.gitlab.io/eigen/docs-nightly/group__TutorialSlicingIndexing.html)). This would require shattering the `WeightedElement` into structures of arrays, but would at the same time enable natural processing using Eigen.

Other than that, I would not be in favor of this three-case distinction duplicating the code. The segment seems to be the most efficient alternative, while not requiring a manual implementation of all three cases.

fsimonis: > Eigen provides dedicated support for operating on index sets

I couldn't figure out how to implement this in our case where multiple operations map to the same output.

> The segment seems to be the most efficient alternative, while not requiring a manual implementation of all three cases.

If we want to get the best performance, then this duplication may be worth it.
However, if the (rather verbose) matrix variant isn't worth it (we need more scenarios to conclude that), then we could use the templated segment variant for the 3d case too.

This code is still pretty clean. It's not like we are writing inline assembly or similar.

davidscn: > I couldn't figure out how to implement this in our case where multiple operations map to the same output.

why is that a problem? I don't think the indices in the array must be unique. The final result will look somewhat like Python or Matlab syntax.

>   If we want to get the best performance, then this duplication may be worth it.
However, if the (rather verbose) matrix variant isn't worth it (we need more scenarios to conclude that), then we could use the templated segment variant for the 3d case too.

In the end, that's a prioritization. For the three components, the differences are barely visible in your plots. For the 1d version, more but I am not sure that's really worth it. We also have the duplication in multiple places

fsimonis: > I don't think the indices in the array must be unique.

This was my assumption.

I implemented three indexing variants based on two `vector<int> outIds, inIds` and a `vector<double>` which is mapped to an `VectorXd eigenW`
- `dindexing` is a scalar version, which performs `out(outIds) += eigenW * in(inIds);`.
- `indexing` is a dynamic version that maps to a MatrixXd and then used the above to map rows `outMap(outIds, Eigen::all) += eigenW * inMap(inIds, Eigen::all)`
- `tindexing` is a version with templated ncols of the Matrix map.

I don't know what is going on here, but this is dog slow for some reason.
 
image

> For the three components, the differences are barely visible in your plots.

Actually, the templated row version is somewhat fragile and fluctuates. The templated segment version seems very solid.

_edit:updated plot with higher iteration count_

fsimonis: I extracted the maps into an anonymous namespace and simplified it to using templated segments for n>1

davidscn: > I implemented three indexing variants based on two vector<int> outIds, inIds and a vector<double> which is mapped to an VectorXd eigenW

where are these implementations?

>  The templated segment

I really dislike calling things "templated" as this is misleading in C++ (it is not a templated function in this case) . What you would like to say is that you lifted the parameter into the type system. Since it is an integer, the evaluation is done at compile-time rather than at runtime. 

With the above measurements, I would still argue that using `segment<N>`, is the best compromise for code simplicity and performance.

fsimonis: > I really dislike calling things "templated" as this is misleading in C++ (it is not a templated function in this case).

The function `template<int SIZE> segment<SIZE>(start)` is templated on the size of the segment. There is also dynamic variant of this `segment(start, size)`.
I am comparing the templated version with the dynamic version. Not sure what there is to dislike about this, nor why this would be misleading.

> What you would like to say is that you lifted the parameter into the type system. Since it is an integer, the evaluation is done at compile-time rather than at runtime.

Yes, I measure the performance for concrete sizes (1,2,3).
The segment size is known to the compiler at compile time for the templated versions, and it is unknown in the non-templated versions.

> With the above measurements, I would still argue that using segment<N>, is the best compromise for code simplicity and performance.

You must be talking about the n==1 branch as I already eliminated the `MatrixX3d` version in [b5d1ace](https://github.com/precice/precice/pull/2411/commits/b5d1ace5529fd8929a203360eedbaef1b5b7f16d).

I manually set the iterations to 1000 to give us tighter results here. The difference for this problem size is small but measurable.

```
-----------------------------------------------------------------------------------------
Benchmark             Time             CPU   Iterations bytes_per_second items_per_second
-----------------------------------------------------------------------------------------
map_direct/1      13286 us        13263 us         1000      1.68414Gi/s       226.041M/s
map_tsegment/1    13873 us        13850 us         1000      1.61276Gi/s       216.461M/s
```

The extra code that the direct indexing branch takes is trivial in complexity and length. I see absolutely no issue integrating this into preCICE as long as this difference is observable across other cases and machines. (Which is why I tagged you in this PR)

davidscn: >> I really dislike calling things "templated" as this is misleading in C++ (it is not a templated function in this case).

> The function template<int SIZE> segment<SIZE>(start) is templated on the size of the segment.

I know what you want to say, and my original point remains: it is a function template and not a templated function. The function has a type argument `int`. 

> size is small but measurable.

Yes, this was already mentioned above. My point was that the difference you measure is not worth another code branch.

fsimonis: [templated function](https://www.cppreference.com/w/cpp/language/templates.html) is an umbrella term.

> A templated function is a function template or a function that is templated.

> The function has a type argument int.

This is called a [constant template parameter](https://www.cppreference.com/w/cpp/language/template_parameters.html#Constant_template_parameter) in contrast to a [type template parameter](https://www.cppreference.com/w/cpp/language/template_parameters.html#Type_template_parameter). The argument is 2/3/4.

> My point was that the difference you measure is not worth another code branch.

I'll create a small case to make sure this is true for smaller n. The current case is pretty large.

fsimonis: > With the above measurements, I would still argue that using segment<N>, is the best compromise for code simplicity and performance.

With higher iteration counts, this is becomming more evident.
I updated the PR to the use of segments in b5d1ace

PR #2415, opened 2025-12-08
Force linking against Boost dynamically

hmartinez82:
## Main changes of this PR

Explicitly set `Boost_USE_STATIC_LIBS`, so Boost linkage behavior is consistent.

Fixes https://github.com/precice/precice/issues/2413 .

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/2415.md
+++ b/docs/changelog/2415.md
@@ -1,0 +1,1 @@
+- Added CMake option `Boost_USE_STATIC_LIBS` defaulting to OFF which forces Boost to find shared library by default.


CMakeLists.txt
```diff
@@ -204,6 +204,7 @@ if(POLICY CMP0167)
   cmake_policy(SET CMP0167 NEW)
 endif()
 
+set(Boost_USE_STATIC_LIBS OFF)
```
fsimonis: I would feel safer to give users/package maintainers an override here:

```suggestion
option(Boost_USE_STATIC_LIBS "Force BoostConfig.cmake to find static libraries." OFF)
```

fsimonis:
@hmartinez82 Thanks a lot for the patch!

PR #2416, opened 2025-12-09
Refactor CXX standard and symbol visibility into CMake variables

fsimonis:
## Main changes of this PR

This PR

- refactors CXX standard into single cache variable
- refactors symbol visibility into single variable

This allows to bump the standard version easier, which can be interesting in testing compatibility with future standards.
It gives the user the option to build preCICE with visible API symbols, which is interesting for testing and compatibility

## Author's checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,6 +79,8 @@ option(PRECICE_RELEASE_WITH_TRACE_LOG "Enable trace logging in release builds" O
 option(PRECICE_RELEASE_WITH_ASSERTIONS "Enable assertions in release builds" OFF)
 
 set(PRECICE_CXX_STANDARD "17" CACHE STRING "CXX Standard for all preCICE targets" )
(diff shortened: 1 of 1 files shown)


fsimonis:
I still need to test if the Kokkos override works as expected

fsimonis:
Tested with Kokkos 4.3.01 and Kokkos 5.0.0.

Using Kokkos 4.3.01 < 4.4.0 results in:

```
-- Found Ginkgo 1.11.0 at /home/simonifc/Documents/precice/ginkgo/build
-- Enabled Kokkos devices: SERIAL
-- Found Kokkos version 4.3.1 at /home/simonifc/Documents/precice/kokkos/build
-- preCICE symbol visibility turned back ON due to Kokkos < 4.4
```

The visibility flags are set correctly.

I won't add a changelog as these options are dev-centric.

PR #2420, opened 2025-12-24
Add acceleration hooks for custom iteration log columns

### LINKED ISSUE #2110: Remove QN-related columns from non-QN schemes in `*-iterations.log`
We have some acceleration schemes that are no quasi-Newton schemes. However, they still write the QN-related columns `QNColumns`, `DeletedQNColumns`, and `DroppedQNColumns` in `*-iterations.log`. It would be better to drop these columns. However, this would be a breaking change.

**ToDo**

If we decide to introduce this change, we could already create the corresponding PR now, since it should be quite simple:

* Update `BaseCouplingScheme::initializeTXTWriters()`: Only `addData("QNColumns")` for QN-related columns if `qnAcceleration && not doesFirstStep()`
* Update `BaseCouplingScheme::advanceTXTWriters()`: Only `writeData` for QN-related columns if `qnAcceleration && not doesFirstStep()`

Compare to 92ae987daf70bc4f3ba5548208b12130329d9d71

**Context**

In #2107 I already proposed this change, but we decided to move the breaking changes into an issue since this would break post-processing scripts if they depend on the order of the columns. In #2107 I introduced some changes that simplify the `Acceleration` class hierarchy without user-observable changes.

BobSolyman:
## Main changes of this PR

- Added two virtual hooks to the Acceleration base class:
  - `addIterationColumns(io::TXTTableWriter&) const`
  - `writeIterationColumns(io::TXTTableWriter&) const`
- BaseCouplingScheme now calls these hooks instead of hardcoding QN-specific columns
- BaseQNAcceleration overrides the hooks to provide QN-specific columns (QNColumns, DeletedQNColumns, DroppedQNColumns)
- Removed dynamic_pointer_cast to BaseQNAcceleration from BaseCouplingScheme
- Non-QN accelerations no longer write QN columns to iteration log files

Closes #2110

This PR implements the extensible design suggested in the issue: instead of BaseCouplingScheme hardcoding knowledge about QN-specific columns, we introduce virtual hooks that allow acceleration implementations to declare and write custom columns.

**Changes:**
- Default hook implementations are no-ops, preserving behavior for existing accelerations
- QN accelerations override the hooks to register their specific columns
- This ensures non-QN accelerations (Constant, Aitken) no longer produce unnecessary QN columns in `precice-*-iterations.log`
- Future accelerations can register their own custom columns by overriding the hooks

**Testing:**
- All 1241 tests pass, including 33 coupling scheme tests
- Verified with IQN-ILS and Constant Relaxation accelerations:
  - IQN: iteration logs contain QNColumns, DeletedQNColumns, DroppedQNColumns with actual values
  - Constant: iteration logs do NOT contain QN columns

## Author's checklist

## Reviewers' checklist

Note: Changelog file will be added after PR number is assigned (per contribution guidelines).
All required tests pass (1241/1241). No breaking changes - default hooks are no-ops.

### DIFF
--- a/src/acceleration/Acceleration.hpp
+++ b/src/acceleration/Acceleration.hpp
@@ -42,16 +42,10 @@ public:
   virtual void importState(io::TXTReader &reader) {}
 
   /// Allows the acceleration to add custom columns to the iteration log file
-  virtual void addLogEntries(io::TXTTableWriter &writer) const
-  {
-    // Default: no extra iteration columns
-  }
+  virtual void addLogEntries(io::TXTTableWriter &writer) const {}
 
   /// Allows the acceleration to write values to custom iteration log columns
-  virtual void writeLogEntries(io::TXTTableWriter &writer) const
-  {
-    // Default: nothing to write
-  }
+  virtual void writeLogEntries(io::TXTTableWriter &writer) const {}
 
 protected:
   /// Checks if all dataIDs are contained in cplData
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -790,7 +790,7 @@ void BaseCouplingScheme::advanceTXTWriters()
     _iterationsWriter->writeData("Convergence", converged ? 1 : 0);
 
     if (not doesFirstStep() && _acceleration) {
-      // Let the acceleration write its custom iteration column values (no-op by default)
+      // Let the acceleration write its custom iteration column values
       _acceleration->writeLogEntries(*_iterationsWriter);
     }
   }


docs/changelog/2420.md
```diff
@@ -0,0 +1 @@
+- Added acceleration hooks for custom iteration log columns; implement for QN acceleration.
```
fsimonis: The changelog is a description of user-observable behavior rather than implementation.

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -772,9 +772,8 @@ void BaseCouplingScheme::initializeTXTWriters()
         _convergenceWriter->addData(convMeasure.logHeader(), io::TXTTableWriter::DOUBLE);
       }
       if (_acceleration) {
-        _iterationsWriter->addData("QNColumns", io::TXTTableWriter::INT);
-        _iterationsWriter->addData("DeletedQNColumns", io::TXTTableWriter::INT);
-        _iterationsWriter->addData("DroppedQNColumns", io::TXTTableWriter::INT);
+        // Ask acceleration to add any custom iteration columns (default no-op)
```
fsimonis:
```suggestion
        // Ask acceleration to add any custom iteration columns
```

Such comments go quickly out of date. Same for the comment below.

src/acceleration/BaseQNAcceleration.hpp
```diff
@@ -152,6 +152,12 @@ class BaseQNAcceleration : public Acceleration {
    */
   int getMaxUsedTimeWindows() const;
 
+  /// Adds QN-specific columns to the iteration log file
+  void addIterationColumns(io::TXTTableWriter &writer) const override;
+
+  /// Writes QN-specific values to the iteration log columns
+  void writeIterationColumns(io::TXTTableWriter &writer) const override;
```
fsimonis: QN acceleration schemes already use the terms iteration and column extensively in its implementation.
We should avoid even more uses by renaming these functions to something more log-oriented like `addLogEntries` and `writeLogEntries`.

fsimonis (changes requested):
Hi, thanks for the very clean contribution.

I have a few minor comments. Afterwards this PR is ready to merge.

BobSolyman:
Thank you for your helpful feedback.
Please find the required updates implemented in the above commits

src/acceleration/Acceleration.hpp
```diff
@@ -40,6 +41,18 @@ class Acceleration {
 
   virtual void importState(io::TXTReader &reader) {}
 
+  /// Allows the acceleration to add custom columns to the iteration log file
+  virtual void addLogEntries(io::TXTTableWriter &writer) const
+  {
+    // Default: no extra iteration columns
+  }
```
fsimonis: IMHO the default is clear and we can shorten this.

```suggestion
  virtual void addLogEntries(io::TXTTableWriter &writer) const {}
```

src/acceleration/Acceleration.hpp
```diff
@@ -40,6 +41,18 @@ class Acceleration {
...
+
+  /// Allows the acceleration to write values to custom iteration log columns
+  virtual void writeLogEntries(io::TXTTableWriter &writer) const
+  {
+    // Default: nothing to write
+  }
```
fsimonis: Same here

```suggestion
  virtual void writeLogEntries(io::TXTTableWriter &writer) const {}
```

src/cplscheme/BaseCouplingScheme.cpp
```diff
@@ -791,18 +790,8 @@ void BaseCouplingScheme::advanceTXTWriters()
...
-        _iterationsWriter->writeData("QNColumns", qnAcceleration->getLSSystemCols());
-        _iterationsWriter->writeData("DeletedQNColumns", qnAcceleration->getDeletedColumns());
-        _iterationsWriter->writeData("DroppedQNColumns", qnAcceleration->getDroppedColumns());
-      } else {
-        // non-breaking implementation uses "0" for these columns (delete columns in the future?)
-        _iterationsWriter->writeData("QNColumns", 0);
-        _iterationsWriter->writeData("DeletedQNColumns", 0);
-        _iterationsWriter->writeData("DroppedQNColumns", 0);
-      }
+      // Let the acceleration write its custom iteration column values (no-op by default)
```
fsimonis: We should remove the "no-op by default" here as well.

fsimonis:
@MakisH This could impact the system tests if they rely on the iterations logs.

MakisH:
> @MakisH This could impact the system tests if they rely on the iterations logs.

Thanks for the heads-up! We don't yet check anything else than VTU exports, so touching the `iterations.log` should be fine for now.

Related issue: https://github.com/precice/tutorials/issues/440

BobSolyman:
@fsimonis Comments fixed 👍🏽

PR #2424, opened 2026-01-05
Fix msvc build errors

### LINKED ISSUE #2423: Non-standard C++ and small errors preventing MSVC compilation
**Describe your setup**

preCICE Version: develop (b86d874258d077fab68b36726f31b9b882289e22)

**Describe the problem**
I am trying to compile preCICE using MSVC, because MSYS2 uses msmpi as the mpi library, but our software already uses intel's  impi and I am worried that this will cause runtime issues. I managed to compile the library with a few small changes to the code and I am creating this issue to ask if you would like to merge those changes back. I am aware that MSVC is not supported as a compiler, but I believe that some of these issues are real issues in your code and should be merged back.

Specific issues:

1. ~~CMakeLists.txt:233: I needed to find package libXml2 as a CONFIG on windows, since a custom build only provided the config.cmake files. This may also prove problematic for users who require a custom build of libXml2.~~
2. src/mapping/NearestNeighborGradientMapping.cpp:8: strings.h only exists on POSIX. The header is actually unused in this file.
3. src/mapping/RadialBasisFctMapping.hpp:181: auto param = std::get<0>(optionalArgs); is not strictly a dependent expression in this if constexpr branch, since optionalArgs is a concrete type that does not depend on the condition. MSVC /permissive- checks the code and hits a static_assert when optionalArgs is an empty tuple, because you cannot access the first element of an empty tuple. This has to do with two-phase lookup and different interpretations of the standard by different compilers as I understand it.
4. src/mapping/impl/BasisFunctions.hpp: There are a bunch of std::invalid_argument that are [[nodiscard]] but that are discarded here. It looks like the conditions are wrong (the not is missing) and then the exception instances are not thrown but simply initialised without a throw statement. MSVC rightly errors out here. EDIT: Is a warning, not an error. The code still appears wrong though
5. src/precice/impl/ParticipantImpl.cpp:185: A preprocessor if occurs inside a function-like macro call.
6. src/utils/networking.cpp:11: __WIN32 does not seem to be the best macro to check on windows, since it is not defined by MSVC. _WIN32 seems more appropriate.

**Step To Reproduce**

1. Checkout boost, libxml2 and eigen in their own folders in precice\thirdparty
2. Build boost:
bootstrap vc143
b2 --prefix=<BASE_PATH>\precice\thirdparty\boost\install --build-dir=<BASE_PATH>\precice\thirdparty\boost\build --with-log --with-thread --with-system --with-filesystem --with-program_options --with-test variant=release link=shared
b2 install --prefix=<BASE_PATH>\precice\thirdparty\boost\install --build-dir=<BASE_PATH>\precice\thirdparty\boost\build --with-log --with-thread --with-system --with-filesystem --with-program_options --with-test variant=release link=shared

3. Build libxml2:
cmake -S . -B build -G "Ninja" -D CMAKE_C_COMPILER=cl -D CMAKE_CXX_COMPILER=cl -D CMAKE_INSTALL_PREFIX=<BASE_PATH>\precice\thirdparty\libxml2\install -D LIBXML2_WITH_ICONV=OFF -D LIBXML2_WITH_LZMA=OFF -D LIBXML2_WITH_PYTHON=OFF
cmake --build build --config Release --parallel
cmake --install build --config Release

4. Build eigen:
cmake -S . -B build -G "Ninja" -D CMAKE_C_COMPILER=cl -D CMAKE_CXX_COMPILER=cl -D CMAKE_INSTALL_PREFIX=<BASE_PATH>\precice\thirdparty\eigen\install -D EIGEN_BUILD_TESTING=OFF -D EIGEN_BUILD_BLAS=OFF -D EIGEN_BUILD_LAPACK=OFF -D EIGEN_BUILD_DOC=OFF -D EIGEN_BUILD_DEMOS=OFF
cmake --install build

5. Build preCICE within an intel oneAPI command prompt for Intel 64 and Visual Studio 2022 (for impi support):
set CXXFLAGS=/permissive-
set FFLAGS=/names:lowercase /assume:underscore
cmake -S . -B build -G "Ninja" -D CMAKE_CXX_COMPILER=cl -D CMAKE_PREFIX_PATH=<THIRD_PARTY_PATH>\eigen\eigen-5.0.1\share\eigen3\cmake;<THIRD_PARTY_PATH>\libxml2\libxml2-v2.15.1\lib\cmake\libxml2;<THIRD_PARTY_PATH>\boost\boost_1_90_0\lib\cmake\Boost-1.90.0;<THIRD_PARTY_PATH>\petsc\petsc-3.21.3 -D CMAKE_INSTALL_PREFIX=install -D CMAKE_BUILD_TYPE=Release -D BUILD_SHARED_LIBS=ON -D BUILD_TESTING=ON -D PRECICE_FEATURE_MPI_COMMUNICATION=ON -D PRECICE_FEATURE_PYTHON_ACTIONS=OFF -D PRECICE_FEATURE_GINKGO_MAPPING=OFF -D PRECICE_FEATURE_PETSC_MAPPING=OFF -D CMAKE_FIND_PACKAGE_PREFER_CONFIG=ON

cmake --build build --config Release --parallel
cmake --install build --config Release

**Expected behaviour**
The build succeeds

**Additional context**
I will create a pull request with my changes to get it to build.

harmenwierenga:
## Main changes of this PR
1. CMakeLists.txt:233: I needed to find package libXml2 as a CONFIG on windows, since a custom build only provided the config.cmake files. This may also prove problematic for users who require a custom build of libXml2. **Made more robust, fall back to the MODULE method to not break existing builds.**
2. src/mapping/NearestNeighborGradientMapping.cpp:8: strings.h only exists on POSIX. The header is actually unused in this file. **Removed.**
3. src/mapping/RadialBasisFctMapping.hpp:181: auto param = std::get<0>(optionalArgs); is not strictly a dependent expression in this if constexpr branch, since optionalArgs is a concrete type that does not depend on the condition. MSVC /permissive- checks the code and hits a static_assert when optionalArgs is an empty tuple, because you cannot access the first element of an empty tuple. This has to do with two-phase lookup and different interpretations of the standard by different compilers as I understand it. **I have refactored the code into a free function template, so that the local optionalArgs is a dependent expression there.**
4. src/mapping/impl/BasisFunctions.hpp: There are a bunch of std::invalid_argument that are [[no_discard]] but that are discarded here. It looks like the conditions are wrong (the not is missing) and then the exception instances are not thrown but simply initialised without a throw statement. MSVC rightly errors out here. **I have added throw statements and inverted the conditions (added not).** 
5. src/precice/impl/ParticipantImpl.cpp:185: A preprocessor if occurs inside a function-like macro call. **I created a separate string variable to extract the different uses of the preprocessor.**
6. src/utils/networking.cpp:11: __WIN32 does not seem to be the best macro to check on windows, since it is not defined by MSVC. _WIN32 seems more appropriate. **Switched to _WIN32.**

closes #2423 

## Author's checklist

## Reviewers' checklist

Could you run the test pipelines for me please?

### DIFF
--- a/src/mapping/impl/BasisFunctions.hpp
+++ b/src/mapping/impl/BasisFunctions.hpp
@@ -241,12 +241,12 @@ public:
       : _shape(shape),
         _supportRadius(supportRadius)
   {
-    if (math::greater(_shape, 0.0)) {
-      std::invalid_argument{
+    if (!math::greater(_shape, 0.0)) {
+      throw std::invalid_argument{
           "Shape parameter for radial-basis-function gaussian has to be larger than zero. Please update the \"shape-parameter\" attribute."};
     }
-    if (math::greater(_supportRadius, 0.0)) {
-      std::invalid_argument{
+    if (!math::greater(_supportRadius, 0.0)) {
+      throw std::invalid_argument{
           "Support radius for radial-basis-function gaussian has to be larger than zero. Please update the \"support-radius\" attribute."};
     }
 
@@ -318,8 +318,8 @@ class CompactThinPlateSplinesC2 : public CompactSupportBase,
 public:
   explicit CompactThinPlateSplinesC2(double supportRadius)
   {
-    if (math::greater(supportRadius, 0.0)) {
-      std::invalid_argument{
+    if (!math::greater(supportRadius, 0.0)) {
+      throw std::invalid_argument{
           "Support radius for radial-basis-function compact thin-plate-splines c2 has to be larger than zero. Please update the \"support-radius\" attribute."};
     }
     _r_inv             = 1. / supportRadius;
@@ -370,8 +370,8 @@ class CompactPolynomialC0 : public CompactSupportBase,
 public:
   explicit CompactPolynomialC0(double supportRadius)
   {
-    if (math::greater(supportRadius, 0.0)) {
-      std::invalid_argument{
+    if (!math::greater(supportRadius, 0.0)) {
+      throw std::invalid_argument{
           "Support radius for radial-basis-function compact polynomial c0 has to be larger than zero. Please update the \"support-radius\" attribute."};
     }
     _r_inv             = 1. / supportRadius;
@@ -422,8 +422,8 @@ class CompactPolynomialC2 : public CompactSupportBase,
 public:
   explicit CompactPolynomialC2(double supportRadius)
   {
-    if (math::greater(supportRadius, 0.0)) {
-      std::invalid_argument{
+    if (!math::greater(supportRadius, 0.0)) {
+      throw std::invalid_argument{
           "Support radius for radial-basis-function compact polynomial c2 has to be larger than zero. Please update the \"support-radius\" attribute."};
     }
 
@@ -475,8 +475,8 @@ class CompactPolynomialC4 : public CompactSupportBase,
 public:
   explicit CompactPolynomialC4(double supportRadius)
   {
-    if (math::greater(supportRadius, 0.0)) {
-      std::invalid_argument{
+    if (!math::greater(supportRadius, 0.0)) {
+      throw std::invalid_argument{
           "Support radius for radial-basis-function compact polynomial c4 has to be larger than zero. Please update the \"support-radius\" attribute."};
     }
 
@@ -528,8 +528,8 @@ class CompactPolynomialC6 : public CompactSupportBase,
 public:
   explicit CompactPolynomialC6(double supportRadius)
   {
-    if (math::greater(supportRadius, 0.0)) {
-      std::invalid_argument{
+    if (!math::greater(supportRadius, 0.0)) {
+      throw std::invalid_argument{
           "Support radius for radial-basis-function compact polynomial c6 has to be larger than zero. Please update the \"support-radius\" attribute."};
     }
     _r_inv             = 1. / supportRadius;
@@ -580,8 +580,8 @@ class CompactPolynomialC8 : public CompactSupportBase,
 public:
   explicit CompactPolynomialC8(double supportRadius)
   {
-    if (math::greater(supportRadius, 0.0)) {
-      std::invalid_argument{
+    if (!math::greater(supportRadius, 0.0)) {
+      throw std::invalid_argument{
           "Support radius for radial-basis-function compact polynomial c6 has to be larger than zero. Please update the \"support-radius\" attribute."};
     }
     _r_inv             = 1. / supportRadius;


CMakeLists.txt
```diff
@@ -230,7 +230,11 @@ message(STATUS "Found Eigen ${Eigen3_VERSION}")
 precice_validate_eigen()
 
 # LibXML2
-find_package(LibXml2 REQUIRED)
+# Try CONFIG mode first for custom builds, fallback to MODULE mode for Linux system packages
+find_package(LibXml2 CONFIG QUIET)
+if(NOT LibXml2_FOUND)
+  find_package(LibXml2 MODULE REQUIRED)
+endif()
```
fsimonis: This style is common, but not very user-friendly. The QUIET hides information, and not using QUIET leads to misleading output.

Have you tried `CMAKE_FIND_PACKAGE_PREFER_CONFIG`?
It exists exactly for your use-case.

You can also overwrite `LIBXML2_INCLUDE_DIR` and `LIBXML2_LIBRARY`

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -182,24 +183,24 @@ void ParticipantImpl::configure(
...
-                 " + trace log"
+    buildTypeStr += " + trace log";
 #endif
 #ifndef PRECICE_NO_ASSERTIONS
-                 " + assertions"
+    buildTypeStr += " + assertions";
 #endif
 #endif // NDEBUG
-    );
+    PRECICE_INFO(buildTypeStr);
```
fsimonis: This used string literal concatenation on purpose.

You could assign the concatenated literal to a `const char*` and then pass that to the `PRECICE_INFO`.

So similar to what you did, but without the `std::string::operator+=`

src/utils/networking.cpp
```diff
@@ -8,7 +8,7 @@ std::string loopbackInterfaceName()
   return "lo";
 #elif defined(__APPLE__) || defined(BSD)
   return "lo0";
-#elif defined(__WIN32)
+#elif defined(_WIN32)
```
fsimonis, line 11: This looks like a typo, which makes me wonder why the below warning isn't shown during MSYS2 builds.

CMakeLists.txt
harmenwierenga: I am using CMAKE_FIND_PACKAGE_PREFER_CONFIG now, thank you

src/precice/impl/ParticipantImpl.cpp
harmenwierenga: I have changed it to compile-time string literal concatenation. Since you use c++17, I used a string_view instead of const char*.

harmenwierenga (commented):
I have responded to your comments

src/utils/networking.cpp
harmenwierenga: I do not know if it prints a warning with MinGW. The problem for me was that #warning is not defined by MSVC, so I got a build error instead of a warning

harmenwierenga:
I have managed to run the unit tests locally on windows, and they succeed:
image

I will create a separate pull request for the one CMake change that was necessary to get them running.

src/utils/networking.cpp
fsimonis: Interesting info. Thanks!

fsimonis:
I'll take over the PR, cleanup the commits and merge.

fsimonis:
@harmenwierenga Thanks for the contribution!

PR #2426, opened 2026-01-09
Copy test executable dependencies to the build folder on windows

### LINKED ISSUE #2425: Run ctest on windows
**Describe your setup**

preCICE Version: develop b86d874258d077fab68b36726f31b9b882289e22

**Describe the problem**
See issue #2423 for details on how to build with MSVC. After building, I attempt to run the unit tests using ctest:

**Step To Reproduce**
1. cd build
2. ctest .

This fails because of missing DLLs. On linux, the runpath ensures that shared objects can be found in the build tree or on the system path, but on windows a best practice appears to be to copy over the DLLs of test dependencies to the build folder of the tests.

**Expected behaviour**
The tests succeed.

harmenwierenga:
Windows lacks the runpath, but it looks in the executable folder for the dependencies. This way, ctest can run after a build and the tests do not require installation

Closes #2425 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -676,6 +676,12 @@ IF (BUILD_TESTING)
     endforeach()
   endif()
 
+  if(WIN32)
+    add_custom_command(TARGET testprecice POST_BUILD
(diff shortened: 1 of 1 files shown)


fsimonis:
Let's discuss in the issue. This shouldn't be necessary.

PR #2430, opened 2026-01-14
Add runtime dll dirs to test path in windows

### LINKED ISSUE #2425: Run ctest on windows

fsimonis:
## Main changes of this PR

This PR appends the DLL locations to the PATH env when running tests on Windows.

This version requires CMake 3.27, which currently doesn't meet our baseline.

This is necessary when preCICE is build against other dynamic libraries on Windows and these are not exposed in PATH.

@harmenwierenga does this work on Windows?

Closes #2425

## Author's checklist

### DIFF
--- a/cmake/discover_tests.py
+++ b/cmake/discover_tests.py
@@ -2,11 +2,19 @@ import argparse
 import pathlib
 import subprocess
 import re
+import os
 
 
-def get_tests():
+def get_tests(extra_paths):
+    env = os.environ.copy()
+    if extra_paths:
+        if "PATH" in env:
+            env["PATH"] = os.pathsep.join([env["PATH"], extra_paths])
+        else:
+            env["PATH"] = extra_paths
+
     out = subprocess.run(
-        [args.executable, "--list_units"], stdout=subprocess.PIPE, check=True
+        [args.executable, "--list_units"], env=env, stdout=subprocess.PIPE, check=True
     ).stdout
     for line in out.strip().splitlines():
         name, ranks = line.decode().split()
@@ -55,7 +63,7 @@ parser.add_argument("--paths")
 
 args = parser.parse_args()
 
-for name, ranks in get_tests():
+for name, ranks in get_tests(args.paths):
 
     print(f"# {name} {ranks}", file=args.output)
 


cmake/discover_tests.py
```diff
@@ -116,5 +117,13 @@ def add_command(out, cmd: str, cargs: list[str]):
         "LABELS",
         f'"{";".join(labels)}"',
     ]
+
+    if args.dlls:
+        paths = [str(pathlib.Path(dll).parent) for dll in args.dlls.split(";")]
+        pargs += [
+            "ENVIRONMENT_MODIFICATION",
+            f"PATH=path_list_append:{';'.join(paths)}",
```
fsimonis:
```suggestion
            f'"PATH=path_list_append:{';'.join(paths)}"',
```

cmake/discover_tests.py
```diff
@@ -116,5 +117,13 @@ def add_command(out, cmd: str, cargs: list[str]):
         "LABELS",
         f'"{";".join(labels)}"',
     ]
+
+    if args.dlls:
+        paths = [str(pathlib.Path(dll).parent) for dll in args.dlls.split(";")]
```
fsimonis:
```suggestion
        paths = { str(pathlib.Path(dll).parent) for dll in args.dlls.split(";") }
```

cmake/discover_tests.py
```diff
@@ -116,5 +117,13 @@ def add_command(out, cmd: str, cargs: list[str]):
         "LABELS",
         f'"{";".join(labels)}"',
     ]
+
+    if args.dlls:
+        paths = { str(pathlib.Path(dll).parent) for dll in args.dlls.split(";") }
```
fsimonis:
```suggestion
        paths = {str(pathlib.Path(dll).parent) for dll in args.dlls.split(";")}
```

harmenwierenga:
I have tried it with msvc, and I got an error during the build that I forgot to mention before. Without the copying, the test discovery already fails because of missing DLLs:
```
FAILED: ctest_tests.cmake precice/build/ctest_tests.cmake
...
  File "precice\cmake\discover_tests.py", line 58, in <module>
    for name, ranks in get_tests():
                       ~~~~~~~~~^^
  File "precice\cmake\discover_tests.py", line 8, in get_tests
    out = subprocess.run(
          ~~~~~~~~~~~~~~^
        [args.executable, "--list_units"], stdout=subprocess.PIPE, check=True
        ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    ).stdout
    ^
  File "Python\Python313\Lib\subprocess.py", line 577, in run
    raise CalledProcessError(retcode, process.args,
                             output=stdout, stderr=stderr)
subprocess.CalledProcessError: Command '[WindowsPath('precice/build/testprecice.exe'), '--list_units']' returned non-zero exit status 3221225781.
ninja: build stopped: subcommand failed.
```
This means that the environment of this command also has to be changed at build time. Copying the dlls to the build folder does also solve this issue, but again, it is a questionable practice

fsimonis:
@harmenwierenga Sorry for the long delay.

This should fix the test detection.

PR #2433, opened 2026-01-14
Increase standard to C++20

fsimonis:
## Main changes of this PR

This PR increases the C++ standard to 20.

Test PR to see how the baselines handle this bump

## Author's checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -78,7 +78,7 @@ option(PRECICE_RELEASE_WITH_DEBUG_LOG "Enable debug logging in release builds" O
 option(PRECICE_RELEASE_WITH_TRACE_LOG "Enable trace logging in release builds" OFF)
 option(PRECICE_RELEASE_WITH_ASSERTIONS "Enable assertions in release builds" OFF)
 
-set(PRECICE_CXX_STANDARD "17" CACHE STRING "CXX Standard for all preCICE targets" )
+set(PRECICE_CXX_STANDARD "20" CACHE STRING "CXX Standard for all preCICE targets" )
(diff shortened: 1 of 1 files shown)


nots1dd:
@fsimonis 

As part of this draft, will you be open to using concepts and ranges (or any other cpp20 features)?

fsimonis:
@nots1dd 

> As part of this draft, will you be open to using concepts and ranges (or any other cpp20 features)?

The pragmatic answer is that we are limited by the compilers and STL version of the baseline, which is Ubuntu 22.04 LTS. We also need to compile the codebase with nvcc when using the GPU mappings.

The former is checked by our CI, the latter needs manual checking.
If this all works, then there is no reason not to use modern features.

Note that the baseline will be lifted to Ubuntu 24.04 LTS with the release of Ubuntu 26.04 LTS in April.

The feature I am looking forward to is `source_location`, which should allow us to significantly simplify the logging framework.

nots1dd:
@fsimonis 

> We also need to compile the codebase with nvcc when using the GPU mappings.

I doubt i can check this unless there are some docs on how to do this perhaps? Figured most of the features could have incorporated without too much inertia but now am not sure.

> The feature I am looking forward to is source_location, which should allow us to significantly simplify the logging framework.

It does, logging in >= cpp20 is actually a bliss

davidscn:
> I doubt i can check this unless there are some docs on how to do this perhaps? Figured most of the features could have incorporated without too much inertia but now am not sure.

https://docs.nvidia.com/cuda/cuda-programming-guide/05-appendices/cpp-language-support.html

(Kokkos -- our device support -- v5 also requires C++20)

I could also test for support.

PR #2444, opened 2026-02-02
Communicate connection info via primaries

s3n-w6i:
This is a draft pull request and is not yet ready to merge. Currently, it removes useful functionality, as it was easier to develop with without introducing new configuration options.

## Main changes of this PR

When using Two-Level initialization, this change adds a third level, where the secondary ranks exchange their connection info through the already connected primaries instead of using the file system for this. This improves performance if the file system is slow (esp. for many processes), but can also hinder performance is the file system is fast:

Pre-Connect Secondary Ranks (max)

- "Original" uses the default WORK file system of SuperMUC
- "DAOS" is a faster alternative

Notice how using the default file system of SuperMUC results in very unpredictable performance, but using the DAOS file system can even be faster than this approach. So this should be a configurable option.

## Next steps

- Define what the configuration options should look like

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/com/Communication.cpp
+++ b/src/com/Communication.cpp
@@ -7,7 +7,10 @@
 #include "Request.hpp"
 #include "logging/LogMacros.hpp"
 #include "precice/impl/Types.hpp"
+#include "profiling/Event.hpp"
 #include "utils/assertion.hpp"
+
+using precice::profiling::Event;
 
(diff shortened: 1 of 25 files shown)


s3n-w6i:
Closing in favor of #2556

PR #2445, opened 2026-02-04
Extend exporter test

HenryCai-st:
## Main changes of this PR
Extend the export tests to have additional checks for file existence, empty data skipping and content verification

Issue [Test output of Exporter #2060](https://github.com/precice/precice/issues/2060)

I keep it as draft and will switch to PR once I feel it is complete.

Currently I have finished the implementation and will go through the checklist. I forgot to format code after inital commit and made a new commit after `pre-commit` hook, would that be a problem?

4 test files: csv/-vtk/-vtp/-vtu export test have been extended, and the testing.cpp/hpp with additional help functions to remove files, read files and examine vtu/vtp files.

Additional checks for file existence is done by removing files before execution.

Check for skipping missing data is currently in `ExportCSVTest.cpp`, that the data columns are left out.

For content verification, I currently check the syntax of VTP and VTU files using libxml2. And CSV using string comparison.

I used `xmlFreeDoc(doc)` to clear the memory when reading xml using libxml2, is that necessary?

For VTK I used string comparison because it is not xml file, which I do not find it good. I am looking for an alternative.

For removal of output files before execution, it might be good to ensure that there is no file before execution.

And I feel this pull request is a bit much, should it be splitted into 2 PRs, check removal + missing and check content?

This currently works on my VM. I need to reproduce it elsewhere.

## Author's checklist

* [✔] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [✔] I added a test to cover the proposed changes in our test suite.

## Reviewers' checklist

### DIFF
--- a/src/io/tests/ExportCSVTest.cpp
+++ b/src/io/tests/ExportCSVTest.cpp
@@ -99,6 +99,8 @@ PRECICE_TEST_SETUP(""_on(2_ranks).setupIntraComm())
 BOOST_AUTO_TEST_CASE(ExportScalarParallel)
 {
   PRECICE_TEST();
+  const std::string filename = fmt::format("Mesh-io-CSVExport.{}_init.csv", context.rank);
+  testing::removeFile(filename);
   mesh::Mesh mesh("Mesh", 2, testing::nextMeshID());
   if (context.isPrimary()) {
     mesh.createVertex(Eigen::Vector2d::Zero());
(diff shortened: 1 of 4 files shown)


src/testing/Testing.cpp
```diff
@@ -139,6 +145,82 @@ void expectFile(std::string_view name)
...
+                                XML_PARSE_NOWARNING | XML_PARSE_NOERROR);
+  bool      valid = (doc != nullptr);
+  if (doc)
+    xmlFreeDoc(doc);
+  return valid;
+}
+
+static bool validateVTKFile(std::string_view filename, const char *typeName)
+{
+  xmlSetGenericErrorFunc(nullptr, xmlErrorSuppress);
```
fsimonis: This looks like a global setter. We should use a local version instead.

src/testing/Testing.cpp
```diff
@@ -139,6 +145,82 @@ void expectFile(std::string_view name)
...
+  xmlDocPtr doc = xmlReadFile(std::string(filename).c_str(), nullptr,
```
fsimonis: Let's wrapt this into a unique_ptr with a custom deleter.

This saves a lot of the trickery below.

PR #2446, opened 2026-02-04
fix a crash in two-level initialization

dabele:
## Main changes of this PR

keep alive buffers used in aSend during two-way intialization.

the buffer passed to `aSend` needs to be kept alive until the send is complete, i.e., after `request->wait()`. Here, a local variable was used that is overwritten in a loop. When the send actually happens, the value at the memory location has changed and the wrong value is sent. Task A sends X to B, B receives Y. Y is either random uninitialized bits or the value that is sent to some other task later in the loop depending on timing. Can be diagnosed by logging the size that is sent and received.

In my tests, depending on the received wrong value, this manifests either as an error message from MPI_Recv about a truncated message or a bad_alloc exception because the changed value is randomly very large.

I would appreciate advice about how you would like this tested, since it is a race condition that depends on the problem size and probably also MPI implementation/config (internal buffering, scheduling, etc). Not sure how large the problem needs to be, I used 2 ASTE instances with 512 MPI tasks each.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/partition/ReceivedPartition.cpp
+++ b/src/partition/ReceivedPartition.cpp
@@ -688,12 +688,16 @@ void ReceivedPartition::createOwnerInformation()
     // Exchange list of shared vertices with the neighbor
     // to store send requests.
     std::vector<com::PtrRequest> vertexListRequests;
+    vertexListRequests.reserve(2 * sharedVerticesSendMap.size());
+    // to keep all send buffers alive until the async sends are complete
+    std::vector<int> vertexListSizeSendBuffers;
+    vertexListSizeSendBuffers.reserve(sharedVerticesSendMap.size()); // make sure there are no reallocations
 
     for (auto &receivingRank : sharedVerticesSendMap) {
-      int  sendSize = receivingRank.second.size();
-      auto request  = utils::IntraComm::getCommunication()->aSend(sendSize, receivingRank.first);
+      vertexListSizeSendBuffers.push_back(static_cast<int>(receivingRank.second.size()));
+      auto request = utils::IntraComm::getCommunication()->aSend(vertexListSizeSendBuffers.back(), receivingRank.first);
       vertexListRequests.push_back(request);
-      if (sendSize != 0) {
+      if (vertexListSizeSendBuffers.back() != 0) {
         auto request = utils::IntraComm::getCommunication()->aSend(span<const int>{receivingRank.second}, receivingRank.first);
         vertexListRequests.push_back(request);
       }
--- a/docs/changelog/2446.md
+++ b/docs/changelog/2446.md
@@ -1,0 +1,1 @@
+- Fixed a race condition in the two-way initialization process that can cause a crash


dabele:
one test times out, unrelated as far as I can tell, doesn't use 2LI

src/partition/ReceivedPartition.cpp
```diff
@@ -688,12 +688,16 @@ void ReceivedPartition::createOwnerInformation()
     // Exchange list of shared vertices with the neighbor
     // to store send requests.
     std::vector<com::PtrRequest> vertexListRequests;
+    vertexListRequests.reserve(sharedVerticesSendMap.size());
```
fsimonis: This stores the requests for n sizes and up to n vertex id buffers.

```suggestion
    vertexListRequests.reserve(2*sharedVerticesSendMap.size());
```

fsimonis:
Thank you very much for the fix!
This is obviously a bug. This really looks like we should do a sweep over all aSends and aReceives again.

> In my tests, depending on the received wrong value, this manifests either as an error message from MPI_Recv about a truncated message or a bad_alloc exception because the changed value is randomly very large.

It is a bit puzzling why we never ran into this. We have 2LI tests, and we also run larger cases with 2LI on clusters, including the SuperMUC.

> one test times out, unrelated as far as I can tell, doesn't use 2LI

True. We need to investigate the recent timeouts on macOS.

> I would appreciate advice about how you would like this tested [...] I used 2 ASTE instances with 512 MPI tasks each.

I think there is not a lot you can do. This info already helps us to kind of figure out what is going on.

We have our project meeting soon, @davidscn and I could try to design a test that triggers this.

I would say, let's merge this now and include it in the Spring release.

fsimonis:
Additional info. This implementation hasn't been touched since its introduction in https://github.com/precice/precice/pull/921, released with preCICE version 2.4.0.

src/partition/ReceivedPartition.cpp
dabele: done

dabele:
> It is a bit puzzling why we never ran into this. We have 2LI tests, and we also run larger cases with 2LI on clusters, including the SuperMUC.

wild speculation: 
- I think some MPIs buffer an unspecified amount of small sends internally, but not unlimited.
- if the sizes are all equal, the buffer may be overwritten with the same value. I encountered it in a probably very badly distributed setup (https://precice.discourse.group/t/problems-with-rbf-on-large-unstructured-meshes/2708/10)

fsimonis:
@dabele Which MPI implementation were you using when these errors appeared?
We normally use OpenMPI or Intel MPI when running large cases.

dabele:
we are using OpenMPI version 4.1.3

fsimonis:
> we are using OpenMPI version 4.1.3

We have used this version excessively in the past. 

Maybe we just need some torture tests. With very wild mesh distributions.

We have our project meeting next week. I'll set this on the agenda and merge afterwards. 

Thanks for your time!

src/partition/ReceivedPartition.cpp
```diff
@@ -688,12 +688,16 @@ void ReceivedPartition::createOwnerInformation()
...
+    // to keep all send buffers alive until the async sends are complete
+    std::vector<int> vertexListSizeSendBuffers;
+    vertexListSizeSendBuffers.reserve(2 * sharedVerticesSendMap.size()); // make sure there are no reallocations
```
davidscn:
```suggestion
    vertexListSizeSendBuffers.reserve(sharedVerticesSendMap.size()); // make sure there are no reallocations
```

src/partition/ReceivedPartition.cpp
davidscn:
```suggestion
    vertexListRequests.reserve(2 * sharedVerticesSendMap.size());
```

src/partition/ReceivedPartition.cpp
dabele: I'm confused, this is what @fsimonis suggested, would like change it back?

davidscn: `vertexListRequests` vs `vertexListSizeSendBuffers` ?

dabele: ah, yes, you are right, I put the `2 *` in the wrong place

dabele: fixed and rebased

fsimonis:
@davidscn and I tried a range of different ways of reproducing this bug, but we cannot.

That said, it is obviously wrong. Merging

fsimonis:
@dabele Thanks again for the fix!

PR #2455, opened 2026-02-22
GSoC entry test draft pr

Amanyd:
## Main changes of this PR
Updated the PRECICE_CHECK macro in MeshConfiguration.cpp to include the XML tag context when a configured data mesh is missing. Used tag.getFullName() instead of getName() to ensure full context.

This is for the entry test of multi step configuration project(gsoc)

before:
bf

after:
af

Could've done same for other configs, but how tightly coupled parsing and object creation are, we definitely need an ast layer and will be overwritten anyways then.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mesh/config/MeshConfiguration.cpp
+++ b/src/mesh/config/MeshConfiguration.cpp
@@ -80,9 +80,9 @@ void MeshConfiguration::xmlTagCallback(
       }
     }
     PRECICE_CHECK(found,
-                  "Data with name \"{}\" used by mesh \"{}\" is not defined. "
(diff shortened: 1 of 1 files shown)


fsimonis:
Low-effort PR

Amanyd:
Hi @fsimonis , thanks for the review. This was a draft PR submitted only as my GSoC entry test, sorry for the confusion. If draft PRs aren’t desired for entries I’ll avoid this in future. Do i delete the branch or leave it closed?

PR #2458, opened 2026-02-23
changing the error message as an entry task for gsoc

Sujanian1304:
## Main changes of this PR
- Added a loop that collects all currently defined data names into a comma-separated string. This iterates over
 `_dataConfig->data()`, which holds all `<data:...> `tags that were parsed from the config file.

- If no data is defined at all, sets the string to `"(none)"` so the message is still clear.

- Enhanced the PRECICE_CHECK message in two ways:
   - Shows the exact XML syntax needed (`<data:scalar .../> `or `<data:vector .../>`), so the user knows the tag format
   - Lists all currently defined data names, so the user can see what's available and spot typos or missing definitions

before:
Screenshot 2026-02-23 052552
after:
Screenshot 2026-02-23 052852

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/mesh/config/MeshConfiguration.cpp
+++ b/src/mesh/config/MeshConfiguration.cpp
@@ -79,10 +79,20 @@ void MeshConfiguration::xmlTagCallback(
         break;
       }
     }
+    std::string availableData;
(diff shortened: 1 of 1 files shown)


fsimonis:
No activity, so I assume this PR was abandonned

Issue #2407, opened 2025-12-02
Thread-safe Participant


fsimonis:
**Please describe the problem you are trying to solve.**

Some users may want to use preCICE in a multi-threaded environment.
Currently, there is no issue when the participant is used only on the primary thread.
This is the case, when the solver internally uses fan-out-fan-in technologies such as OpenMP, [parallel execution policy](https://en.cppreference.com/w/cpp/algorithm/execution_policy_tag_t.html), or [Intel TBB](https://www.intel.com/content/www/us/en/developer/tools/oneapi/onetbb.html).

It is however a problem if each thread maintains its own list of vertices and data blocks. In such a case, each thread may want to `readData()` and `writeData()` from preCICE.

This currently leads to race conditions.

**Describe the solution you propose.**

Make the preCICE participant thread-safe in two steps:
 
3. lock down the complete API with an internal lock
4. try to prevent locking of `readData` and `writeData` as much as possible.

**Describe alternatives you've considered**

Flatout forbid this and require the user from locking as they currently need to.

**Additional context**

We need a [`mutable std::mutex`](https://en.cppreference.com/w/cpp/thread/mutex.html) inside the `Participant` and use [`std::scoped_lock`](https://en.cppreference.com/w/cpp/thread/scoped_lock.html) in the API functions.
The finalization requires special thought.

davidscn:
> It is however a problem if each thread maintains its own list of vertices and data blocks. In such a case, each thread may want to readData() and writeData() from preCICE.

For reading data, it's, however, not an issue if the coupling mesh is shared across all threads and just the operation of reading data is being threaded. This may be a relevant use case.

fsimonis:
Conceptually all non-interpolating reads directly read from the data buffer, so they should be fine. Interpolating reads are a bit complicated. 

Logging and profiling also need some thought.

davidscn:
> Interpolating reads are a bit complicated.

The memory allocations should be thread-local though, so I don't see them as more complicated.

fsimonis:
You are right that the allocation shouldn't be a problem.

The Bspline used for the interpolation gets computed on the first interpolation. This needs to be locked.

fsimonis:
In a `readData()` call, this complete branch is the only part that I think needs locking (including the condition):

https://github.com/precice/precice/blob/18a381cf66fb97a81f6f3bedd23d8713ff5ba2ba/src/time/Waveform.cpp#L222-L225

SKM2227229725:
[...]

PR #2460, opened 2026-02-23
Add thread safety to Participant API with mutex synchronization

devSuryansh:
addresses: #2407

## Main changes of this PR

- Added `std::mutex _apiMutex` member variable to the `Participant` class for thread-safe API access
- Protected all public API methods with `std::scoped_lock` to ensure mutual exclusion during concurrent access
- Added `#include <mutex>` headers to `Participant.cpp` and `Participant.hpp`
- Applied consistent code formatting using clang-format across modified files
- Updated `Waveform` class files with consistent indentation and formatting

This change addresses the need for thread-safe access to the preCICE Participant API when multiple threads in an application need to interact with preCICE simultaneously. Without proper synchronization, concurrent calls to the Participant API could lead to race conditions and undefined behavior.

The implementation uses a single mutex to protect all public API methods, ensuring that only one thread can execute Participant API calls at a time. This approach:

- Provides thread safety without requiring changes to user code
- Is simple and maintainable
- Has minimal performance impact for typical single-threaded use cases
- Enables new multi-threaded coupling scenarios

## Author's checklist

- [x] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
- [x] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
- [ ] I added a test to cover the proposed changes in our test suite.
- [ ] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html). (N/A - not a breaking change)
- [ ] I ran the systemtests by adding the label `trigger-system-tests` (may be skipped if minor change)

## Reviewers' checklist

- [ ] Does the changelog entry make sense? Is it formatted correctly?
- [ ] Do you understand the code changes?
- [ ] Is the mutex implementation appropriate for the use case?
- [ ] Should we add thread-safety tests for this feature?
- [ ] Is the performance impact acceptable?

### DIFF
--- a/docs/changelog/thread-safety.md
+++ b/docs/changelog/thread-safety.md
@@ -1,0 +1,1 @@
+- Added thread safety to the Participant API by introducing mutex-based synchronization for all public methods, enabling safe concurrent access from multiple threads.


davidscn:
> Hi, Thanks for the contribution!
> 
> You added the API lock on the wrong layer of the API. We use a compilation firewall using a pimpl idiom. The actual API is in ParticipantImpl.cpp.
> 
> As a first step, I recommend protecting the complete API with a single lock. We can add fine-grained access in follow-up PRs.

This would exclude https://github.com/precice/precice/issues/2407#issuecomment-3602277784, no?

fsimonis:
@davidscn Yes exactly.

The global mutex is something we can add and test today.
Everything else will run into problems with at least profiling and logging.
I'll add a roadmap to the issue: https://github.com/precice/precice/issues/2407#issuecomment-3966853057

fsimonis:
No more activity, I assume the PR was abandoned.

Issue #2028, opened 2024-05-23
Investigate BSpline control point computation


fsimonis:
**Please describe the problem you are trying to solve.**

We implement time-interpolation using a b-spline. This requires us to generate a knot vector and compute control points.

Computing the control points results in a system with $2 + (n-2) * (degree+1)$ entries in a $n \times n$ matrix, where $n$ is the amount of samples in time.

For large $n$ of multiple hundred, the fill-factor is far smaller than $0.5$ and a sparse matrix representation makes sense.
For smaller $n$ the matrix has a fill-factor of $\approx 0.5$ and a dense representation may be the superior choice due to the overhead of sparse matrices.

**Describe the solution you propose.**

1. Investigate how computing the control points of sparse and dense variants compare based on $n$ for various spline degrees and vertex counts.
2. Does it make sense to dynamically switch between a dense and sparse representation based on the amount of samples $n$, the amount of vertices, and the spline degree?

**Describe alternatives you've considered**

Assume that the sparse variant is good enough until we run into an example which show-cases performance issues.

**Additional context**

See #2021

BenjaminRodenberg:
> Assume that the sparse variant is good enough until we run into an example which show-cases performance issues.

From a practical point of view we should wait for this condition before we start working on this issue. It's also much easier to develop a feature to adaptively shift from the sparse to non-sparse implementation, if you have a benchmark.

From the perspective of research it's naturally an interesting question to find out where this point is. Might be a nice topic for a thesis to find the limits and come up with a solution.

NiklasKotarsky:
This is a really neat idea. I am a little bit unsure of were the current bottle neck is in the current bspline algorithm. One thing I am a little bit suspicious of are the following lines in the function 'interpolateAt'
https://github.com/precice/precice/blob/db0cf012f5cb9e90d701cf5cedd052ef01131644/src/math/Bspline.cpp#L70C1-L75C1
I have not measured it yet nor thought much about how to make it faster, since it seems to be fast enough for me. However, a speedup in 'interpolateAt' should be more noticeable than in the construction of the bsplines, since this function is normally executed multiple times per bspline construction.

fsimonis:
> I am a little bit unsure of were the current bottle neck is in the current bspline algorithm.

You can define some `Events` and then use the profiling tools to get more insight.
This is all it takes:

```cpp
profiling::Event e("my.event");
```

SKM2227229725:
[...]

PR #2461, opened 2026-02-23
Add Event to measure BSpline computation

SKM2227229725:
**Main changes**

- Adds optional profiling to BSpline control point computation.
- Measures matrix construction, QR factorization, solve step, total time, and fill factor.
- Enabled via PRECICE_BSPLINE_PROFILE=1. 
- No functional or numerical changes.
- Zero overhead when disabled.

**Motivation**

The current implementation uses an n × n sparse matrix.
For small n, sparse overhead may dominate due to higher fill factor.

This PR provides runtime data to:

1. Identify bottlenecks
2. Verify fill factor
3. Determine if/when a dense solver would be beneficial

 No solver change is introduced.
**Author's checklist**

-  Used pre-commit
-  Changelog (not needed)
-  Tests (investigation only)
-  Breaking changes (N/A)
-  C++17
-  CMake 3.22.1
-  Squashed commits
-  System tests (optional)

### DIFF
--- a/src/math/Bspline.cpp
+++ b/src/math/Bspline.cpp
@@ -1,13 +1,13 @@
 #include "math/Bspline.hpp"
-#include "math/differences.hpp"
-#include "utils/assertion.hpp"
 
 #include <Eigen/Core>
 #include <Eigen/Sparse>
 #include <algorithm>
 #include <cstdlib>
 #include <unsupported/Eigen/Splines>
+
 #include "math/differences.hpp"
+#include "profiling/Event.hpp"
 #include "utils/assertion.hpp"
 
 namespace precice::math {
@@ -26,6 +26,8 @@ Bspline::Bspline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree
   auto relativeTime = [tsMin = _tsMin, tsMax = _tsMax](double t) -> double { return (t - tsMin) / (tsMax - tsMin); };
   ts                = ts.unaryExpr(relativeTime);
 
+  profiling::Event e("bspline.compute");
+
   // The code for computing the knots and the control points is copied from Eigens bspline interpolation with some modifications
   // https://gitlab.com/libeigen/eigen/-/blob/master/unsupported/Eigen/src/Splines/SplineFitting.h
 
@@ -38,6 +40,7 @@ Bspline::Bspline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree
   std::vector<Eigen::Triplet<double>> matrixEntries;
   matrixEntries.reserve(2 + (n - 2) * (splineDegree + 1));
 
+  // Build matrix entries
   matrixEntries.emplace_back(0, 0, 1.0);
   for (Eigen::DenseIndex i = 1; i < n - 1; ++i) {
     const Eigen::DenseIndex span      = Eigen::Spline<double, 1>::Span(ts[i], splineDegree, _knots);
@@ -50,10 +53,12 @@ Bspline::Bspline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree
   matrixEntries.emplace_back(n - 1, n - 1, 1.0);
   PRECICE_ASSERT(matrixEntries.capacity() == matrixEntries.size(), matrixEntries.capacity(), matrixEntries.size(), n, splineDegree);
 
+  // Create sparse matrix
   Eigen::SparseMatrix<double> A(n, n);
   A.setFromTriplets(matrixEntries.begin(), matrixEntries.end());
   A.makeCompressed();
 
+  // Solve system
   Eigen::SparseQR<Eigen::SparseMatrix<double>, Eigen::COLAMDOrdering<int>> qr;
   qr.analyzePattern(A);
   qr.factorize(A);


src/math/Bspline.cpp
```diff
@@ -29,15 +58,24 @@ Bspline::Bspline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree
...
+  auto knotEnd = std::chrono::high_resolution_clock::now();
+  double knotTime = std::chrono::duration<double, std::milli>(knotEnd - knotStart).count();
 
   // 2. Compute the control points
   // We use a nxn sparse matrix with 2 + (n-2) * (d+1) entries and thus a fill-factor < 0.5.
   Eigen::DenseIndex                   n = xs.cols();
   std::vector<Eigen::Triplet<double>> matrixEntries;
   matrixEntries.reserve(2 + (n - 2) * (splineDegree + 1));
 
+  auto constructStart = std::chrono::high_resolution_clock::now();
```
davidscn: Note that we have internally our own profiling framework (see `src/profiling/Event.hpp`), introducing a new layer via `chrono` is redundant.

SKM2227229725: Thank you for the pointer!

I wasn’t aware that preCICE already has an internal profiling framework via src/profiling/Event.hpp. That makes sense — introducing std::chrono would indeed be redundant.

I will rework the PR to integrate the measurements using the existing profiling infrastructure instead of adding a separate timing layer.

Thanks for the guidance!

SKM2227229725:
The PR has now been updated to use the internal profiling framework instead of std::chrono.
Please let me know if further refinements are needed.

src/math/Bspline.cpp
```diff
@@ -1,20 +1,18 @@
...
 #include "utils/assertion.hpp"
+#include "profiling/Event.hpp"
 
 #include <Eigen/Core>
 #include <Eigen/Sparse>
 #include <algorithm>
 #include <cstdlib>
 #include <unsupported/Eigen/Splines>
-#include "math/differences.hpp"
-#include "utils/assertion.hpp"
```
fsimonis, line 11: I'm surprised that clang-format didn't deal with this

src/math/Bspline.cpp
```diff
@@ -26,39 +24,56 @@ Bspline::Bspline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree
   auto relativeTime = [tsMin = _tsMin, tsMax = _tsMax](double t) -> double { return (t - tsMin) / (tsMax - tsMin); };
   ts                = ts.unaryExpr(relativeTime);
 
-  // The code for computing the knots and the control points is copied from Eigens bspline interpolation with some modifications
-  // https://gitlab.com/libeigen/eigen/-/blob/master/unsupported/Eigen/src/Splines/SplineFitting.h
```
fsimonis, line 30: Please restore theses comments

src/math/Bspline.cpp
```diff
@@ -75,4 +90,5 @@ Eigen::VectorXd Bspline::interpolateAt(double t) const
 
   return interpolated;
 }
-} // namespace precice::math
+
```
fsimonis: ?

src/math/Bspline.cpp
```diff
@@ -26,39 +24,56 @@ Bspline::Bspline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree
...
+    matrixEntries.emplace_back(n - 1, n - 1, 1.0);
+    PRECICE_ASSERT(matrixEntries.capacity() == matrixEntries.size(), matrixEntries.capacity(), matrixEntries.size(), n, splineDegree);
   }
-  matrixEntries.emplace_back(n - 1, n - 1, 1.0);
-  PRECICE_ASSERT(matrixEntries.capacity() == matrixEntries.size(), matrixEntries.capacity(), matrixEntries.size(), n, splineDegree);
 
+  // Create sparse matrix
   Eigen::SparseMatrix<double> A(n, n);
   A.setFromTriplets(matrixEntries.begin(), matrixEntries.end());
   A.makeCompressed();
```
fsimonis, line 59: This should be part of the construction event.
Use `eConstruct.stop()` here instead of the extra scope.

src/math/Bspline.cpp
```diff
@@ -26,39 +24,56 @@ Bspline::Bspline(Eigen::VectorXd ts, const Eigen::MatrixXd &xs, int splineDegree
...
 
+  // Solve system
   Eigen::SparseQR<Eigen::SparseMatrix<double>, Eigen::COLAMDOrdering<int>> qr;
-  qr.analyzePattern(A);
-  qr.factorize(A);
-
-  _ctrls = qr.solve(xs.transpose());
+  
+  {
+    profiling::Event eFactor("bspline.factorization");
```
fsimonis: Please remove the scopes and use a single event `bspline.control` to measure the rest of the function.

fsimonis (changes requested):
Hi,
this is a good start into making this measurable.

Please also add a debug output of the fill-ratio, which seemed to have disappeared.

fsimonis:
Some notes for the future:
- There is a reason why we provide a PR template. Please stick to it.
- The Original Post is severely outdated now. Could you please update it?

If you are interested into this topic, you could use the [oscillator tutorial](https://precice.org/tutorials-oscillator.html) to test some scenarios. For larger meshes and more substeps, you could build your own test case.

Related to #2028

src/math/Bspline.cpp
SKM2227229725: Understood. I'll restore the original comments regarding the B-spline interpolation logic and the Eigen reference link."

src/math/Bspline.cpp
SKM2227229725: "I'll double-check my local .clang-format settings and manually sort these includes to match the project style. Thanks for catching that."

src/math/Bspline.cpp
SKM2227229725: Got it. I will remove the extra scope and use the suggested approach for the construction event."

src/math/Bspline.cpp
SKM2227229725: That makes sense. I'll consolidate the profiling into a single bspline_control event as requested

src/math/Bspline.cpp
SKM2227229725: I'll clean up the whitespace/formatting here. It seems like an accidental addition during the refactor.

SKM2227229725:
@fsimonis Hi, I’ve addressed the review comments and updated the PR accordingly.
Could you please take another look when you have time?
Thank you for your feedback so far!

fsimonis:
@SKM2227229725 I don't see answers to the review comments nor changes in the PR.

SKM2227229725:
@fsimonis Thanks for your feedback! I’ve now addressed the review comments and pushed the updates. Could you please take another look when you have time?

SKM2227229725:
@fsimonis Thanks for your feedback! I’ve now addressed the review comments and pushed the updates.

- Restored missing comments and Eigen reference
- Consolidated profiling into a single event (bspline.control)
- Removed unnecessary scopes and cleaned up formatting
- Ensured construction is part of the profiling event
- Adjusted includes to match project style

Could you please take another look when you have time?

fsimonis:
I did the final cleanup for this PR.

PR #2462, opened 2026-02-24
Add optional mesh attribute to export tags to allow selective mesh export

### LINKED ISSUE #1888: Select mesh to export

Exports can be necessary for debugging adapters, but in some cases the exports can be too large.
This is especially a problem when combining large meshes with exporting many timestamps.

Allow to select meshes to export.

Export everything
```xml
<export:vtu />
```
Export only a specific mesh
```xml
<export:vtu mesh="SolidMesh" />
```
Export multiple meshes
```xml
<export:vtu meshes="SolidMesh:FluidMesh" />
```
or
```xml
<export:vtu mesh="SolidMesh" />
<export:vtu mesh="FluidMesh" />
```

**Additional context**

Sujanian1304:
## Main changes of this PR

_edit by @fsimonis_

This PR adds an attribute to restrict an export to a given mesh.
`<export:X mesh="Name" />` now only exports the mesh `Name`.

Closes #1888 

## Author's checklist

### DIFF
--- a/src/io/ExportContext.hpp
+++ b/src/io/ExportContext.hpp
@@ -23,6 +23,9 @@ struct ConfiguredExport {
 
   // @brief type of the exporter (e.g. vtk).
   std::string type;
+
+  // @brief Configured mesh name to export (empty means export all meshes).
+  std::string configuredMeshName;
 };
 
 struct ExportContext : public ConfiguredExport {
--- a/src/precice/config/ParticipantConfiguration.cpp
+++ b/src/precice/config/ParticipantConfiguration.cpp
@@ -658,10 +658,26 @@ void ParticipantConfiguration::finishParticipantConfiguration(
 
   // Add export contexts
   for (const io::ConfiguredExport &exportConfig : _exportConfig->exportContexts()) {
-    auto kind = exportConfig.everyIteration ? io::Export::ExportKind::Iterations : io::Export::ExportKind::TimeWindows;
+    // Check if meshes to export even exist
+    if (!exportConfig.configuredMeshName.empty()) {
+      PRECICE_CHECK(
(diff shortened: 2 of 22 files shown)


src/io/ExportContext.hpp
```diff
@@ -27,6 +27,9 @@ struct ExportContext {
 
   // @brief type of the exporter (e.g. vtk).
   std::string type;
+
+  // @brief Configured mesh name to export (empty means export all meshes).
+  std::string configuredMeshName;
```
fsimonis, line 28: We already have a mesh member that keeps the future name.
You can use that instead and simplify a bunch of logic below.

src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -729,15 +734,31 @@ void ParticipantConfiguration::finishParticipantConfiguration(
     };
 
     // Create one exporter per provided mesh
+    bool meshFound = exportContext.configuredMeshName.empty();
     for (const auto &meshContext : participant->providedMeshContexts()) {
+      if (!exportContext.configuredMeshName.empty() &&
+          exportContext.configuredMeshName == meshContext.mesh->getName()) {
+        meshFound = true;
+      }
```
fsimonis: Instead of doing this twice, please add a separate validation step where you check if the defined meshes in exports actually exist on this participant.
The error message you added is good.

fsimonis (changes requested):

aadhil2k4:
Hi @Sujanian1304 , We both are working on the same issue. If you are comfortable, we both can collaborate in this PR.

fsimonis:
@Sujanian1304  Are you still interested in completing this?

fsimonis:
I cleaned up the PR, added tests and will merge after the CI passes

Issue #1927, opened 2024-01-10
Ownership cleanup


fsimonis:
**Please describe the problem you are trying to solve.**

Our code-base is full of `std::shared_ptr` as this has always been used by default.
We pass a lot of these things around, even if a normal non-owning pointer or reference would suffice.

This leads to a lot of unnecessary code, more complicated function signatures, and code linters complaining about missing moves everywhere.

**Describe the solution you propose.**

Ideally avoid ownership altogether and use a pure value-based system where possible.
For the rest start by converting functions to using references instead of shared pointers where owning isn't necessary.
Then find components which have clear ownership and change the owner to using a `std::unique_ptr`.

**Describe alternatives you've considered**

Live with shared ownership.

**Additional context**

alihan-tadal:
Hi! I would be happy to contribute to the project. I thought I could start with this issue but I need some guidance. In your opinion, what path should I follow and where should I start cleanup? Thank you very much in advance.

fsimonis:
Hi there, sure!
Given that this touches essentially every aspect of the code, it may be a good idea to open multiple smaller PRs to keep friction to a minimum.

Start with the individual packages and have a look at the `SharedPointer` headers, try to follow the usage of the defined pointer types. Then inspect and clean-up the callsides. Something that takes a shared point to a Mesh often only needs to take a `const Mesh&` for example. This changes can sometimes propagate upwards in the call chain.

Cleaning this up for all the major packages will be a huge step in the right direction.

Once all of this has been clarified, ownership will be much easier to deduce.

Wdhanashri:
Submitted io ownership-cleanup slice fix in #2463  cleaned up ownership in the io module (using single-owner pointers where sharing was unnecessary), removed the old io pointer-alias header, and updated related callsites/includes without changing behavior; build and pre-commit pass. If this approach looks right, I’ll continue for the remaining modules.

Wdhanashri:
Follow-up, Submitted #2468 it applies the ownership cleanup pattern to the mapping module.

PR #2463, opened 2026-02-24
Change exporters to be uniquely owned by the Participant

Wdhanashri:
- replace shared ownership of io::Export with unique ownership
- make ExportContext move-only and propagate move semantics
- change ParticipantConfiguration export config ownership to unique_ptr
- remove io SharedPointer alias header and remaining alias include usage
- keep behavior unchanged while clarifying ownership boundaries"

## Main changes of this PR
* Replaced shared ownership of `io::Export` with unique ownership in the export pipeline.
  * `io::ExportContext::exporter` changed from shared pointer alias to `std::unique_ptr<io::Export>`.
  
* Made `io::ExportContext` effectively move-only and propagated move semantics through call sites.
  * `ExportConfiguration` now inserts contexts with `std::move`.
  * `ParticipantState::addExportContext` now takes by value and moves into storage.
  * `ParticipantConfiguration` now builds per-mesh `ExportContext` instances and moves them into participant state.

* Changed ownership of `ParticipantConfiguration::_exportConfig` from shared to unique (`std::unique_ptr<io::ExportConfiguration>`), since it has a single owner.

* Removed `src/io/SharedPointer.hpp` and all remaining includes/usages in touched code.

* Replaced legacy `PtrExport(new ...)` construction with `std::make_unique<...>`.

Part of #1927.
This PR addresses the io package by reducing unnecessary shared ownership where ownership is clear and local.
The changes keep behavior intact while making ownership boundaries explicit and simplifying type usage and callsites.

## Author's checklist

### Validation performed

  - Full build succeeded:
      - cmake --build precice/build-full -j
  - Formatting/lint checks passed:
      - pre-commit run -a

  ### Notes on scope

  - This PR is intentionally scoped to io ownership cleanup.
  - It does not attempt cross-package shared_ptr cleanup; that will be handled in follow-up PRs.


src/precice/config/ParticipantConfiguration.cpp
```diff
@@ -665,14 +665,20 @@ void ParticipantConfiguration::finishParticipantConfiguration(
...
-      exportContext.meshName = meshContext.mesh->getName();
-
-      io::PtrExport exporter;
+      io::ExportContext participantExportContext;
+      participantExportContext.location          = exportContext.location;
+      participantExportContext.everyNTimeWindows = exportContext.everyNTimeWindows;
+      participantExportContext.everyIteration    = exportContext.everyIteration;
+      participantExportContext.updateSeries      = exportContext.updateSeries;
+      participantExportContext.type              = exportContext.type;
+      participantExportContext.meshName          = meshContext.mesh->getName();
```
fsimonis: At this point, it may be cleaner to split the configuration members off into a base class/struct `ConfiguredExport`.
Then make the ExportContext inherit from it and the export pointer and the meshName. Then save the latter in the ParticipantState.

Wdhanashri: Done. I split the configuration fields into ConfiguredExport, made ExportContext inherit it, and now construct ExportContext from the ConfiguredExport directly. Please take another look.

src/io/ExportContext.hpp
```diff
@@ -29,6 +25,24 @@ struct ExportContext {
...
+struct ExportContext : public ConfiguredExport {
+  ExportContext() = default;
+  explicit ExportContext(const ConfiguredExport &config)
+  : ConfiguredExport(config)
+  {
+  }
+  explicit ExportContext(ConfiguredExport &&config)
+  : ConfiguredExport(std::move(config))
+  {
+  }
```
fsimonis: If you define a ctor, then you need to define the move ctor and assignment too.
It's probably easier to either move this logic into a standalone function `ConfiguredExport + name -> ExportContext`, where you then later overwrite the exporter.

fsimonis:
@Wdhanashri Are you still interested in completing this?

src/io/ExportContext.hpp
Wdhanashri: I removed the explicit ExportContextconstructors and added a standalone helpermakeExportContext(ConfiguredExport, meshName) function

Wdhanashri:
> @Wdhanashri Are you still interested in completing this?

Hi @fsimonis , sorry for the late reply I was a bit under the weather along with exams over the past few weeks.

PR #2465, opened 2026-02-26
Add performance regression test workflow

Eshaan-byte:
## Main changes of this PR

I added the github workflow for performance regression testing which runs test on precise-bench on base and PR branch, and also compares the result and reports them in Github action  jobs summary

Addressed this issue based on feedback from @MakisH :
-Trigers manually by label or from weekly schedule
-Reused the existing precise bench  (Google Benchmark) suite
-Compares base vs head on same runner for fair comparson
-Reports job summary result only, no pass/fail output

## Author's checklist

* [x] I used the `pre-commit` hook
* [ ] I added a changelog file — N/A (CI-only change, not user-observable)
* [ ] I added a test — N/A (workflow file)
* [ ] For breaking changes — N/A
* [x] I stuck to C++17 features
* [x] I stuck to CMake version 3.22.1
* [x] I squashed all commits
* [ ] I ran the systemtests — skipping (CI-only change)

### DIFF
--- a/.github/workflows/performance-regression.yml
+++ b/.github/workflows/performance-regression.yml
@@ -1,6 +1,6 @@
 name: Performance Regression Test
 # Compare precice-bench results between base and PR branch.
-# Based on the approach from build-and-test.yml (archlinux Release config).
+# Uses the production CMakePreset and Google Benchmark's compare.py.
 
 on:
   pull_request:
@@ -8,6 +8,11 @@ on:
   schedule:
     - cron: '0 4 * * 1'  # weekly on Monday
   workflow_dispatch:
+    inputs:
+      base_ref:
+        description: 'Base ref to compare against (branch, tag, or SHA)'
+        required: false
+        default: 'develop'
 
 concurrency:
   group: perf-${{ github.event.pull_request.number || github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/performance-regression.yml
```diff
@@ -0,0 +1,109 @@
...
+jobs:
+  benchmark:
+    if: >
+      github.event_name == 'workflow_dispatch' ||
+      github.event_name == 'schedule' ||
+      github.event.label.name == 'trigger-perf-test'
+    runs-on: ubuntu-latest
+    container:
+      image: 'precice/ci-archlinux:latest'
+      options: --shm-size=2gb
```
fsimonis, line 37: I'm not sure if arch is the best choice for performance regressions as we need to build our own petsc. Maybe the fedora one is better suited for that, but let's start with arch.

.github/workflows/performance-regression.yml
```diff
@@ -0,0 +1,109 @@
...
+
+      - name: Build and benchmark head
+        run: |
+          git checkout ${{ github.event.pull_request.head.sha || github.sha }}
+          cmake -B build-head -DCMAKE_BUILD_TYPE=Release -DPRECICE_BUILD_BENCHMARKS=ON -DPRECICE_FEATURE_MPI_COMMUNICATION=ON -DPRECICE_FEATURE_PETSC_MAPPING=ON
+          cmake --build build-head -j$(nproc)
+          cd build-head
+          ./precice-bench --benchmark_out=head.json --benchmark_out_format=json --benchmark_repetitions=3
+
+      - name: Compare results
```
fsimonis, line 81: Google benchmark includes a [compare script](https://github.com/google/benchmark/blob/main/tools/compare.py), please use that. The arch package installs it in `/usr/share/googlebenchmark/tools/compare.py`.

.github/workflows/performance-regression.yml
```diff
@@ -0,0 +1,109 @@
...
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+
+      # Build and run benchmarks for the base branch first, then HEAD.
+      # Both run on the same machine so the comparison is fair.
+      - name: Build and benchmark base
+        run: |
+          git checkout ${{ github.event.pull_request.base.sha || 'HEAD~1' }}
+          cmake -B build-base -DCMAKE_BUILD_TYPE=Release -DPRECICE_BUILD_BENCHMARKS=ON -DPRECICE_FEATURE_MPI_COMMUNICATION=ON -DPRECICE_FEATURE_PETSC_MAPPING=ON
```
fsimonis: You should probably use the production CMakePreset and disable both tests and tools to speed up the build. Same below

.github/workflows/performance-regression.yml
```diff
@@ -0,0 +1,109 @@
...
+          cmake --build build-base -j$(nproc)
```
fsimonis: No need to pass `-j` with ninja. Same below

fsimonis (commented):
Hi, thanks for the effort.

How does this work when it is triggered as a workflow dispatch or via the schedule? Which versions are compared then?

.github/workflows/performance-regression.yml
Eshaan-byte: Got it, sticking with arch for now.

Eshaan-byte:
> Hi, thanks for the effort.
> 
> How does this work when it is triggered as a workflow dispatch or via the schedule? Which versions are compared then?

Good point. I added a step that sets the refs explicitly for each trigger type — PRs compare base vs head sha, schedule compares develop~1 vs develop, and manual dispatch lets you specify a base ref

.github/workflows/performance-regression.yml
Eshaan-byte: Switched to the compare.py that ships with the arch benchmark package. Much better because it also does statistical testing.

.github/workflows/performance-regression.yml
Eshaan-byte: Right, ninja parallelizes by default.

fsimonis:
For the schedule I'm not sure that comparing develop with develop~1 is very useful.
Comparing develop to main/last release on the other hand would be very interesting.

Eshaan-byte:
> For the schedule I'm not sure that comparing develop with develop~1 is very useful. Comparing develop to main/last release on the other hand would be very interesting.

Makes sense, changed it to compare develop against main for scheduled runs.

.github/workflows/performance-regression.yml
```diff
@@ -35,75 +40,58 @@ jobs:
...
-              f.write("| Benchmark | Base (ns) | Head (ns) | Change |\n")
-              f.write("|-----------|-----------|-----------|--------|\n")
-              for name in sorted(base.keys()):
-                  if name not in head:
-                      continue
-                  b, h = base[name], head[name]
-                  pct = ((h - b) / b) * 100 if b > 0 else 0
-                  f.write(f"| {name} | {b:.1f} | {h:.1f} | {pct:+.1f}% |\n")
-          EOF
+          pip install scipy
```
fsimonis: This shouldn't work as arch disallows installing system packages using pip.

Can you please create a venv and install scipy and numpy in there?

.github/workflows/performance-regression.yml
```diff
@@ -35,75 +40,58 @@ jobs:
...
+          python3 /usr/share/googlebenchmark/tools/compare.py benchmarks build/base.json build/head.json | tee comparison.txt
```
fsimonis: We should enable to only show aggregates, disable color to prevent the output from getting ugly in the markdown summary.

Also, neat would be to export the summary as json and upload it too.

fsimonis:
@Nitish-bot no need for another PR. This is in a good enough state that I rather clean it up myself instead of review an entire new one.

fsimonis:
@Eshaan-byte are you still working on this?

fsimonis:
I did a cleanup pass over this PR. Merging.

fsimonis:
@Eshaan-byte Thanks for the contribution!

PR #2466, opened 2026-02-26
Remove obsolete Time.cpp

nots1dd:
## Main changes of this PR

I was just reading through the source tree and managed to read `Time.cpp` which served no other purpose than to initialize the `Time` struct's static members.

There is no need to create a separate cpp file just to initialize `Time::DEFAULT_WAVEFORM_DEGREE` and `Time::MIN_WAVEFORM_DEGREE`, can simply inline it in the header itself and decl the init vals in `Time.hpp` itself.

Can confirm after the changes the build and all tests are passing

nil (rational is explained in commit description and explanation for the changes made.)

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/sources.cmake
+++ b/src/sources.cmake
@@ -299,7 +299,6 @@ target_sources(preciceCore
     src/time/Sample.hpp
     src/time/SampleResult.hpp
     src/time/Stample.hpp
-    src/time/Time.cpp
     src/time/Time.hpp
     src/time/TimeGrids.cpp
     src/time/TimeGrids.hpp
--- a/src/time/Time.cpp
+++ b/src/time/Time.cpp
@@ -1,9 +1,0 @@
-#include "Time.hpp"
-
-namespace precice::time {
-
-const int Time::DEFAULT_WAVEFORM_DEGREE = 1;
-
-const int Time::MIN_WAVEFORM_DEGREE = 0;
-
-} // namespace precice::time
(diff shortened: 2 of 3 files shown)


fsimonis:
Hi,
Thanks, good catch!
This is what happens after many years of redesign.

Not sure why your motivation is not in list :wink:

Our commit messages normally start with a capital verb according to the rule:

> Applying this commit will X

nots1dd:
> Hi, Thanks, good catch! This is what happens after many years of redesign.
> 
> Not sure why your motivation is not in list 😉
> 
> Our commit messages normally start with a capital verb according to the rule:
> 
> > Applying this commit will X

ah my bad dont seem to recall that from your contrib guidelines

as for the motivation well i did initially find your org via gsoc but thought i might be too late to make my mark so
i just thought y not just help in refactoring code here and there (helps me learn about the code structure and logic more so thats another plus)

so my motivation just boils down to helping another foss project ig :)

fsimonis:
> thought i might be too late to make my mark 

The deadline is end of March. Plenty of time.

nots1dd:
> > thought i might be too late to make my mark 
> 
> The deadline is end of March. Plenty of time. 

Oh yea just checked it is from 16th to 31st dk y i assumed 16th was the last date

nots1dd:
Regarding the commit name @fsimonis would u like me to rename it? Or is it alright for now

fsimonis:
Thanks!

Issue #751, opened 2020-05-07
Context for XML error messages


fsimonis:
**Please describe the problem you are trying to solve.**
Providing context is important for error messages. The use must be able to see:
* what went wrong
* why it went wrong
* where it went wrong

We cover the first two points in #698 and provide the module raising the error e.g. `mesh::MeshConfiguration`.
Especially for configuration-related errors, there is currently no way of displaying detailed location to the user. 

**Describe the solution you propose.**

Ideally we would display compiler-style messages:
```
preCICE ERROR in Configuration: Time window size has to be larger than zero.
40 |         <time-window-size value="1.0" />
   |                           ^~~~~~~~~~~
```

The information about the line and the position of the final character of the tag are relatively easy to fetch from libxml2 using:
```cpp
auto l = xmlSAX2GetLineNumber(parserContext);
auto c  = xmlSAX2GetColumnNumber(parserContext);
```

This requires to save the parsercontext as a member of `ConfigParser` so that it can be accessed during call-backs. Extracting the line of the configuration file requires to save the file content as a member.
All of this has to happen here:

https://github.com/precice/precice/blob/47747abf9babb977a6b45edbe16ac69ec8d6948b/src/xml/ConfigParser.cpp#L128-L135

As the callbacks can only happen during [`xmlParseChunk()`](https://github.com/precice/precice/blob/47747abf9babb977a6b45edbe16ac69ec8d6948b/src/xml/ConfigParser.cpp#L133), we can reset these members after it has been called.

All of this allows us to create error messages like this:
```
preCICE ERROR in Configuration: Time window size has to be larger than zero.
40 |         <time-window-size value="-2" />
```
or 
```
preCICE ERROR in Configuration: Time window size has to be larger than zero.
40 |         <time-window-size value="-2" />
   |                                       ^
```
Based on the length of the name and attribute key-value pairs, we could approximate the start of the tag, but this will be cumbersome.

Alternatively, we could ignore the column and give more vertical some context:
```
preCICE ERROR in Configuration: Time window size has to be larger than zero.
39 |         <max-time-windows value="1" />
40 >         <time-window-size value="-2" />
41 |         <exchange data="Data1" mesh="MeshOne" from="SolverOne" to="SolverTwo" />
```

A better but more involved solution would be to use the information of the libxml2 [node information](http://www.xmlsoft.org/html/libxml-parser.html#xmlParserNodeInfo) to get the start and end of the actual node.

**Open Questions**
[...]

PR #2467, opened 2026-02-26
Improve XML error messages with file location context

SriShruti24:
## Main changes of this PR
Adds parser-level storage of the parsed file content and parser context, capturing line/column positions during SAX callbacks and propagating them into tag objects **ConfigParser.cpp.**
Extends **XMLTag, getLine, getLocationContext, formatMessage** and exposes the current tag to attribute parsers so errors can include file snippets **XMLTag.hpp, XMLTag.cpp.**
Enhances error reporting so libxml2 parse errors, unknown-tag errors and attribute validation errors include compiler-style context (line/column + snippet); updates attribute handling to attach location info **XMLAttribute.hpp**.

Improves usability: configuration errors now point directly to the offending location in the XML file, letting users find and fix problems far faster
Related to the issue #751 .

## Author's checklist

## Reviewers' checklist


src/xml/XMLTag.cpp
```diff
@@ -8,6 +8,10 @@
 
 namespace precice::xml {
 
+namespace {
+thread_local const XMLTag *s_currentTag = nullptr;
```
fsimonis: Random globals are a nogo for maintainability.

src/xml/XMLTag.cpp
```diff
@@ -312,4 +322,57 @@ std::string_view XMLTag::getOccurrenceString(XMLTag::Occurrence occurrence)
...
+  for (int i = 1; i <= _line && std::getline(iss, lineStr); ++i) {
+  }
+  std::string result = std::to_string(_line) + " | " + lineStr + "\n";
+  if (_column > 0) {
+    result += std::string(_column - 1, ' ') + "^";
+  }
+  return result;
+}
+
+std::string XMLTag::formatMessage(std::string_view message) const
```
fsimonis, line 338: What is the motivation for appending to the message inside the function?
If the context contains braces, then the formatting library will return runtime errors that are very tedious to debug.

src/xml/XMLTag.hpp
```diff
@@ -235,8 +242,14 @@ class XMLTag {
   void areAllSubtagsConfigured() const;
 
   void resetAttributes();
+
+  int _line = -1;
+  int _column = -1;
```
fsimonis: I mentioned the problem with this inside the issue.

fsimonis (changes requested):

fsimonis:
Also, this PR doesn't close 751

fsimonis:
@SriShruti24 hi, are you still working on this?

src/xml/XMLTag.hpp
SriShruti24: Hiii @fsimonis could you please confirm this do you want me to refactor this to use std::optional<int> for _line and _column instead of using -1 as a sentinel value?

src/xml/XMLTag.cpp
SriShruti24: Hiii @fsimonis thanks for pointing out, I now pass the XMLTag pointer directly as a parameter to readValue() methods. so when a function needs the tag context, it gets it as a parameter.

src/xml/XMLTag.cpp
SriShruti24: that is true actually when I was implementing it I overlooked it and missed this and for now what I did is I escape curly braces in formatMessage by duplicating them to prevent misinterpretation by the formatter.

SriShruti24 (commented):
Hiii @fsimonis could you please now look the changes and confirm for the third one. Thank you very much for the review !!!! and really sorry for the late actually I was busy with the midterm exams.

SriShruti24:
Once the review is completed and finalized, I will make sure to resolve the conflicts.

fsimonis (changes requested):
I have reviewed the changes and think they need complete overhaul. There is also enough value here to keep the PR.

A collection of thoughts:
- Pointing XMLTags back to the ConfigParser will lead to trouble. Let's keep them separated.
- The ConfigParser instance is already handed to the xml parser as context, no need to create more statics or so.
- As the XML files are generally small, I think we could create a `std::vector<std::string>` that hold the lines from the file for direct lookup. If we do this before the xml parsing step, and we can use them as early as the parser error messages.
- The error messages are raised via the `OnError(etc)Func` callbacks and then passed to the message proxy. The line numbers need to be extracted on the callback side.
- When a XMLTag gets "connected", you added the `setLocation`, which could also just pass the content of the line to the XMLTag, instead of handling references again. Not super efficient, but the usability boost will be worth it. The line number can then be rendered in the error message. This way, tag locations are available everywhere without further changes and without relying on an existing ConfigParser instance.
- Let's concentrate to adding context to the libxml2 parser errors first.

@SriShruti24 Are you still up to the challenge?

SriShruti24:
Thanks for the detailed feedback — I've reworked the implementation along the lines you described. XMLTag and ConfigParser are now fully separated, the static parser-instance state has been removed, and the XML file lines are stored up front so they can be used for libxml2 parser errors and later tag-location handling. setLocation() now passes the relevant source-line content directly to XMLTag. As requested, I focused first on adding context to libxml2 parser errors and added a test covering malformed XML. Happy to make further adjustments if needed.

fsimonis:
Hi, please read our updated contributing.md.

PR #2468, opened 2026-02-26
Refactor mapping ownership: unique_ptr for Mapping, non-owning pointers for borrowers

Wdhanashri:
## Main changes of this PR

- `ParticipantState` becomes the single owner of all `Mapping` objects via `vector<unique_ptr<Mapping>>` and a new `addMapping()` method.
- `MappingContext`, `DataContext`, and `Partition` change from `shared_ptr<Mapping>` to non-owning `Mapping*`.
- `ConfiguredMapping::mapping` changes from `shared_ptr` to `unique_ptr<Mapping>`, with `make_unique` at all creation sites.
- `_mappingConfig` in `ParticipantConfiguration` changes from `shared_ptr` to `unique_ptr`.
- Removes dead includes of `mapping/SharedPointer.hpp` and dead `getMapping` declaration.

Addresses #1927 for the mapping module. Replaces `shared_ptr<Mapping>` with explicit ownership (`unique_ptr` for the single owner, raw pointers for borrowers) to make the ownership model clear and reduce unnecessary reference counting.

## Author's checklist

- [x] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
- [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
- [ ] I added a test to cover the proposed changes in our test suite.
- [ ] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).
- [ ] I ran the system tests by adding the label `trigger-system-tests` (may be skipped if minor change).


Divinesoumyadip:
Solid work driving #1927 forward. Replacing shared_ptr with strict unique ownership and raw observers is exactly the right architectural move to kill the atomic ref-counting overhead and clarify lifecycle intent.
One critical detail we need to verify before merging: strict topological destruction.
By moving to non-owning raw pointers (Mapping*) in DataContext, MappingContext, and Partition, we are implicitly trusting that ParticipantState (the sole owner) will strictly outlive all borrowers. If the teardown sequence is ever inverted, or if an observer attempts to dereference its Mapping* during its own destructor (e.g., for logging or local cleanup) after ParticipantState has purged its vector<unique_ptr<Mapping>>, we introduce a silent use-after-free vulnerability.
Does the current global object graph strictly guarantee this lifecycle boundary? If there is any ambiguity in the teardown order, we should either explicitly nullify these observer pointers during ParticipantState teardown, or drop in some debug assertions to validate the destruction sequence.

Otherwise, the implementation and diff look exceptionally clean. Appreciate the refactor.

fsimonis:
@Wdhanashri Looks like there are some formatting issues which shouldn't happen with pre-commit installed.
You can run `pre-commit run -va` to fix them.

fsimonis:
@Wdhanashri Hi, are you still working on this?

Wdhanashri:
> @Wdhanashri Looks like there are some formatting issues which shouldn't happen with pre-commit installed. You can run `pre-commit run -va` to fix them.

I fixed the formatting/source-index issues locally, ran pre-commit run -a successfully, and pushed the changes.

PR #2469, opened 2026-02-27
Prevent duplicate coupling schemes between identical participants

### LINKED ISSUE #1695: Don't allow duplicate coupling schemes in compositional coupling scheme
The following compositional coupling scheme is currently allowed:
```
  <coupling-scheme:parallel-implicit>
    ...
  </coupling-scheme:parallel-implicit>

  <coupling-scheme:parallel-explicit>
    ...
  </coupling-scheme:parallel-explicit>

  <coupling-scheme:parallel-explicit>
    ...
  </coupling-scheme:parallel-explicit>
</precice-configuration>
```

I think one of the two parallel explicit schemes is not needed and we should raise an error. We could easily check for duplicates of coupling schemes between two identical solvers.

Again: If we reduce the number of legal configurations, this will reduce the number of required tests, simplify implementation and make preCICE more user-friendly.

Sujanian1304:
## Main changes of this PR
Adds a check in CompositionalCouplingScheme::addCouplingScheme() that rejects 
coupling schemes whose coupling partners already exist in the composition, 
raising a clear error instead of silently allowing misconfiguration.

Closes #1695

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/cplscheme/CompositionalCouplingScheme.cpp
+++ b/src/cplscheme/CompositionalCouplingScheme.cpp
@@ -54,15 +54,13 @@ void CompositionalCouplingScheme::addCouplingScheme(
   PRECICE_TRACE();
 
   // Check for duplicate coupling partners
-  auto newPartners = couplingScheme->getCouplingPartners();
-  for (const auto &scheme : allSchemes()) {
-    auto existingPartners = scheme->getCouplingPartners();
-    for (const auto &newPartner : newPartners) {
-      PRECICE_CHECK(std::find(existingPartners.begin(), existingPartners.end(), newPartner) == existingPartners.end(),
-                    "A coupling scheme with participant \"{}\" already exists in this compositional coupling scheme. "
-                    "Please make sure that only one coupling scheme exists between any two participants.",
-                    newPartner);
-    }
+  auto existingPartners = getCouplingPartners();
+  auto newPartners      = couplingScheme->getCouplingPartners();
+  for (const auto &newPartner : newPartners) {
+    PRECICE_CHECK(std::find(existingPartners.begin(), existingPartners.end(), newPartner) == existingPartners.end(),
+                  "A coupling scheme between participants \"{}\" and \"{}\" already exists in this compositional coupling scheme. "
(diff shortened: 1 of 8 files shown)


src/cplscheme/CompositionalCouplingScheme.cpp
```diff
@@ -52,6 +52,19 @@ void CompositionalCouplingScheme::addCouplingScheme(
     const PtrCouplingScheme &couplingScheme)
 {
   PRECICE_TRACE();
+
+  // Check for duplicate coupling partners
+  auto newPartners = couplingScheme->getCouplingPartners();
+  for (const auto &scheme : allSchemes()) {
+    auto existingPartners = scheme->getCouplingPartners();
+    for (const auto &newPartner : newPartners) {
```
fsimonis: The compositional scheme also implements `getCouplingPartners()`.
You can use that to get the existing partners.

src/cplscheme/CompositionalCouplingScheme.cpp
```diff
@@ -52,6 +52,19 @@ void CompositionalCouplingScheme::addCouplingScheme(
...
+      PRECICE_CHECK(std::find(existingPartners.begin(), existingPartners.end(), newPartner) == existingPartners.end(),
+                    "A coupling scheme with participant \"{}\" already exists in this compositional coupling scheme. "
```
fsimonis: It's best to name both conflicting partners here.

fsimonis (changes requested):
Hi,
thanks for the effort!

Could you create some simple violating configurations and add them to the config tests `tests/config/tests.cmake`?
This is a quick way of ensuring that some configs raise an error. Use a simple parallel-explicit scheme to keep the configs short.

You can test many combinations like swapping first and second, or using a multi scheme with three participants and a parallel-explicit scheme with two.

fsimonis:
Looks like these changes are untested.

I suggest building the code you submit locally and run the tests.

fsimonis:
@Sujanian1304 are you still working on this?

PR #2470, opened 2026-02-27
Refactor includes and utils code structure

nots1dd:
## Main changes of this PR

Honestly I contemplated whether I should open multiple PRs for this but it doesnt warranty it, so ill explain every change here.

1. All instances of including `stddef.h` have been replaced with cpp's `cstddef` in the codebase

Note that all tests pertaining to these changes are passing as they should, because `cstddef` only adds confirmation of certain types being under std namespace - its functionality remains identical to stddef

2. `utils/` overhaul

Quite a lot of changes made in `precice::utils`:

2.1 - General headers that were not required to be included are omitted (validated by successful compilation and CI)

2.2 - `utils/traits.hpp` - concatenated namespace

2.3 - `stacktrace`, `networking` and `Helpers` did not require a separate cpp file considering that only one function with minimal logic was declared in each one of them, they could simply be inlined.

2.4 - `TypeNames.hpp` refactor involved in replacing the old `getTypeName` logic that used string allocs to a more modern solution involving templates, constexpr and string_view. Considering the usage of this function is within `xml/Printer` (particularly in `toMarkdown`) the previous impl simply brought unnecessary allocs that can be avoided. Also note that if we invoke explicitly `getTypeName<int>()` etc, due to constexpr it is evaluated in compile time unlike the former.

There is a generic forwarder of `getTypeName` for API compatibilty and passing any parameter like so:

```cpp
utils::precice::getTypeName(myVar);
```

To validate 2.4's changes, there is already a test that validates the ostream of `toMarkdown` that uses `TypeNames` (it passes)

3. `mesh/` extra includes were removed. (tests and builds pass)

We can validate all changes by CI + ctest

To refactor the codebase and move towards modern cpp design choices

We should wait for the CI to complete builds to ensure that there are no missing include problems (even though I did not face this problem). Other commit info is mentioned in main changes section.

No new tests needed to be added as these were mostly modification changes, no new features were added. (no breaking changes)

## Author's checklist

## Reviewers' checklist


nots1dd:
@fsimonis or @MakisH initial thoughts on this? (ive seen that it seems to be a busy week for yall so just an inital review would be great)

hopefully future work i can actually gear towards some proposal work once i understand the logical flow better

fsimonis:
Hi,
multiple commits here would have been really nice. At the moment, this PR is a lot of noise. With multiple commits, I could filter out all the include changes for example.

Include cleanups are nice, but also a trap. Please test locally that the project continues to build with the unity build disabled (yes, we should have this in the CI, but it's too slow). There is a PRECICE CMake variable to configure this.

> did not require a separate cpp file

That's by design. Please revert.

> TypeNames.hpp refactor involved in replacing the old getTypeName logic that used string allocs to a more modern solution involving templates

More modern yes, but not cleaner. A clean and modern replacement is a traits-based system with a value variable template. Example being [`std::is_scalar_v<>`](https://en.cppreference.com/w/cpp/types/is_scalar.html) but instead of bool, you return a `constexpr std::string_view`.

nots1dd:
Hi @fsimonis,

> Example being [std::is_scalar_v<>](https://en.cppreference.com/w/cpp/types/is_scalar.html) but instead of bool, you return a constexpr std::string_view.

Ok i did not know this was a thing, however considering that this approach directly allows us to be put in the fmt ostream isnt this cleaner? if i were to use a boolean to validate i would either need a helper or ternary operator to validate then print it. Regardless `std::string_view` will be RO data only so there are tradeoffs to both in my eyes. (with this method being less readable but the comments should explain it)

>  multiple commits here would have been really nice. At the moment, this PR is a lot of noise. With multiple commits, I could > filter out all the include changes for example.

i am sorry for that i misread the docs regarding commits thats on me.

> test locally that the project continues to build with the unity build disabled

yes i disabled unity build and it still compiles perfectly fine

i am willing to just scrap this PR and perhaps add a new one with multiple commits that can then be squashed before merge

fsimonis:
> i am sorry for that i misread the docs regarding commits thats on me.

It's not in the docs, it's good practise.

> i am willing to just scrap this PR and perhaps add a new one with multiple commits that can then be squashed before merge

Let's do this. Please open a PR for the include cleanup and a separate PR for the typename refactoring.

Closing

PR #2473, opened 2026-03-01
Include the data name in preconditioner warnings

### LINKED ISSUE #1515: Warning for sub-vector becoming numerically zero missing data context
I have a case exchanging multiple data values, some of which don't change over time. This leads to multiple instances of [this](https://github.com/precice/precice/blob/fded29bfb7e07e3b945cc21c3838434354b4750c/src/acceleration/impl/ResidualSumPreconditioner.cpp#L60-L65) warning, without telling me which data that is. It would be helpful including that information in the error message (I already guessed now).

Example:

```
---[precice] WARNING: A sub-vector in the residual-sum preconditioner became numerically zero ( sub-vector = 0). If this occurred in the second iteration and the initial-relaxation factor is equal to 1.0, check if the coupling data values of one solver is zero in the first iteration. The preconditioner scaling factors were not updated for this iteration and the scaling factors determined in the previous iteration were used.
---[precice] WARNING: A sub-vector in the residual-sum preconditioner became numerically zero ( sub-vector = 0). If this occurred in the second iteration and the initial-relaxation factor is equal to 1.0, check if the coupling data values of one solver is zero in the first iteration. The preconditioner scaling factors were not updated for this iteration and the scaling factors determined in the previous iteration were used.
```

Amanyd:
## Main changes of this PR
Include the data names in precondition warnings (applied across all three types)

When exchanging multiple data values, the preconditioner warning fires once per zero sub-vector but doesn't tell the name. closes #1515 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/acceleration/AitkenAcceleration.cpp
+++ b/src/acceleration/AitkenAcceleration.cpp
@@ -52,7 +52,9 @@ void AitkenAcceleration::initialize(const DataMap &cplData)
   _values       = Eigen::VectorXd::Zero(entries);
   _oldValues    = Eigen::VectorXd::Zero(entries);
   if (_primaryDataIDs.size() > 1) {
-    _preconditioner->initialize(subVectorSizes);
+    std::vector<std::string> subVectorNames;
+    std::transform(_primaryDataIDs.cbegin(), _primaryDataIDs.cend(), std::back_inserter(subVectorNames), [&cplData](const auto &d) { return cplData.at(d)->getDataName(); });
+    _preconditioner->initialize(std::move(subVectorSizes), std::move(subVectorNames));
   }
 }
 
--- a/src/acceleration/BaseQNAcceleration.cpp
+++ b/src/acceleration/BaseQNAcceleration.cpp
@@ -746,7 +746,9 @@ void BaseQNAcceleration::initializeVectorsAndPreconditioner(const DataMap &cplDa
 
   std::vector<size_t> subVectorSizes; // needed for preconditioner
   std::transform(_primaryDataIDs.cbegin(), _primaryDataIDs.cend(), std::back_inserter(subVectorSizes), [&cplData, windowStart, this](const auto &d) { return _primaryTimeGrids.value().getTimeGridAfter(d, windowStart).size() * cplData.at(d)->getSize(); });
-  _preconditioner->initialize(subVectorSizes);
+  std::vector<std::string> subVectorNames;
+  std::transform(_primaryDataIDs.cbegin(), _primaryDataIDs.cend(), std::back_inserter(subVectorNames), [&cplData](const auto &d) { return cplData.at(d)->getDataName(); });
+  _preconditioner->initialize(std::move(subVectorSizes), std::move(subVectorNames));
 
   specializedInitializeVectorsAndPreconditioner(cplData);
 }
--- a/src/acceleration/impl/Preconditioner.hpp
+++ b/src/acceleration/impl/Preconditioner.hpp
@@ -2,6 +2,7 @@
 
 #include <Eigen/Core>
 #include <numeric>
+#include <string>
 #include <vector>
 
 #include "cplscheme/SharedPointer.hpp"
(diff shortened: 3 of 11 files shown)


Amanyd:
Hi @MakisH , would u mind taking a look.

src/acceleration/impl/Preconditioner.hpp
```diff
@@ -32,13 +33,15 @@ class Preconditioner {
 
   /**
    * @brief initialize the preconditioner
-   * @param size of the pp system (e.g. rows of V)
+   * @param svs sizes of each sub-vector
+   * @param svNames names of the coupling data corresponding to each sub-vector
    */
-  virtual void initialize(std::vector<size_t> &svs)
+  virtual void initialize(std::vector<size_t> &svs, std::vector<std::string> &svNames)
```
fsimonis: We already touch the initialization, we can also cleanup this old interface by passing by value.

```suggestion
  virtual void initialize(std::vector<size_t> svs, std::vector<std::string> svNames)
```

src/acceleration/impl/Preconditioner.hpp
```diff
@@ -32,13 +33,15 @@ class Preconditioner {
...
   {
     PRECICE_TRACE();
 
     _subVectorSizes = svs;
+    _subVectorNames = svNames;
```
fsimonis: We should assert that both have the same size.

src/acceleration/test/PreconditionerTest.cpp
```diff
@@ -152,7 +152,8 @@ BOOST_AUTO_TEST_CASE(testResPreconditioner)
 
   ResidualPreconditioner precond(-1);
 
-  precond.initialize(svs);
+  std::vector<std::string> svNames{"Data0", "Data1", "Data2"};
+  precond.initialize(svs, svNames);
```
fsimonis: With the above change, you can construct these in-place and cleanup the tests.


```suggestion
  precond.initialize({2,4,2}, {"Data0", "Data1", "Data2"});
```

src/acceleration/AitkenAcceleration.cpp
```diff
@@ -52,7 +52,9 @@ void AitkenAcceleration::initialize(const DataMap &cplData)
   _values       = Eigen::VectorXd::Zero(entries);
   _oldValues    = Eigen::VectorXd::Zero(entries);
   if (_primaryDataIDs.size() > 1) {
-    _preconditioner->initialize(subVectorSizes);
+    std::vector<std::string> subVectorNames;
+    std::transform(_primaryDataIDs.cbegin(), _primaryDataIDs.cend(), std::back_inserter(subVectorNames), [&cplData](const auto &d) { return cplData.at(d)->getDataName(); });
+    _preconditioner->initialize(subVectorSizes, subVectorNames);
```
fsimonis: And then move them here

```suggestion
    _preconditioner->initialize(std::move(subVectorSizes), std::move(subVectorNames));
```

src/acceleration/BaseQNAcceleration.cpp
```diff
@@ -746,7 +746,9 @@ void BaseQNAcceleration::initializeVectorsAndPreconditioner(const DataMap &cplDa
 
   std::vector<size_t> subVectorSizes; // needed for preconditioner
   std::transform(_primaryDataIDs.cbegin(), _primaryDataIDs.cend(), std::back_inserter(subVectorSizes), [&cplData, windowStart, this](const auto &d) { return _primaryTimeGrids.value().getTimeGridAfter(d, windowStart).size() * cplData.at(d)->getSize(); });
-  _preconditioner->initialize(subVectorSizes);
+  std::vector<std::string> subVectorNames;
+  std::transform(_primaryDataIDs.cbegin(), _primaryDataIDs.cend(), std::back_inserter(subVectorNames), [&cplData](const auto &d) { return cplData.at(d)->getDataName(); });
+  _preconditioner->initialize(subVectorSizes, subVectorNames);
```
fsimonis: Same here.

Amanyd:
@fsimonis Thanks for the review, changes applied.

fsimonis:
> Thanks for the review, changes applied.

For the future, please answer to the review comments in some way. Can be an explanation or :+1:

fsimonis:
Last thing: please add a changelog file

Amanyd:
@fsimonis Thanks for the tip. I'll make sure to answer individual review comments moving forward. I've just added the changelog file.

fsimonis:
Thanks for the contribution!

PR #2476, opened 2026-03-02
2454 : m2n: detect large data transfers for future compression

### LINKED ISSUE #2454: Lossless data compression for large data-sets in time-interpolation

Combining time-interpolation with volume coupling and high-dimensional data can result in a lot of data to exchange.

This can overburden the communication backend and/or the network between participants.

Inspired by [python-blosc2](https://blosc.org/python-blosc2/reference/storage.html) who provides compression of numpy arrays, preCICE could implement a compression step for exceptionally large amounts of data.
This compression threshold could be hard-coded or configurable, maybe even per m2n.

blosc2 defaults to the modern zstd.
We implemented profiling #2328 compression using lzma as it's supported in the python version used by our users. Otherwise, zstd would have been a better fit.

The alternative could be a lossy method, which reduces the amount of samples needed to compute the bspline on the receiving side.

**Additional context**

ChehakTrehan:
This PR adds a small preparatory change in `M2N::send` to detect exceptionally large data transfers.

It computes the exact size of the data being sent, compares it against a temporary hard-coded threshold,
and logs whether the data would be sent raw or marked for compression.

No compression is applied yet. The goal of this change is only to establish the correct location
and decision logic before introducing actual compression.

If this approach looks okay, I will proceed with implementing the compression step in a follow-up change.
Fixes #2454

### DIFF
--- a/src/m2n/M2N.cpp
+++ b/src/m2n/M2N.cpp
@@ -265,6 +265,25 @@ void M2N::send(
     int                         meshID,
     int                         valueDimension)
 {
+  // compute data size in bytes (exact)
+  const std::size_t dataSizeBytes =
+      itemsToSend.size() * sizeof(double);
+
+  // temporary compression threshold (1 MB)
+  const std::size_t compressionThreshold =
+      1 * 1024 * 1024;
+
+  // compression decision (NO compression yet)
+  const bool shouldCompress =
+      dataSizeBytes >= compressionThreshold;
+
+  // log decision (for verification only)
+  PRECICE_DEBUG(
+      "M2N::send decision: size={} bytes, threshold={} → {}",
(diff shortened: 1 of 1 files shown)


fsimonis:
Hi some notes:

- We have a PR template, please follow it.
- This doesn't fix the issue, please don't use closing markers such as fixes or closes, but use something like "Related to X" instead.
- This PR doesn't add value and the message is misleading for everyone using preCICE from develop.

If you want to investigate this properly, then try to implement a compression in this place and measure the compression delay using the profiling library. Then display the real compression ratio in the logs.
After that head to the tutorials and run real cases to see how this compression ratio looks in practice. Then head back to the issue and report your findings.
These steps would be super useful, but this PR isn't.

Closing

PR #2479, opened 2026-03-03
Precompute hwloc topology

fsimonis:
## Main changes of this PR

This PR precomputes the hwloc topology in the CI and defines it using HWLOC_XMLFILE.

It should reduce the startup time per test by around 100ms.

## Author's checklist

### DIFF
--- a/.github/workflows/build-and-test.yml
+++ b/.github/workflows/build-and-test.yml
@@ -251,6 +251,15 @@ jobs:
       run: brew install --overwrite --quiet eigen libxml2 boost petsc openmpi ninja
     - name: Generate build directory
       run: mkdir -p build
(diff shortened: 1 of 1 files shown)


fsimonis:
We need to add hwloc to all base images to make this easier to handle.

fsimonis:
The precomputation is now enabled for linux and mac CIs.
Bottleneck is still the Intel CI, which may need some more inspection.

The reduction in overhead is significantly lower than expected, overall we save around 10 minutes. Ubuntu 2004 doesn't show a difference, but this baseline will shift very soon.
Archlinux also doesn't show a noticeable change even though one is expected.

I'll merge now and check again after the baseline bump.

PR #2481, opened 2026-03-04
Add expression templates for default filter and formatter 

### LINKED ISSUE #1428: Expression templates for default log filters and formatters

preCICE uses boost log to parse log formatters and filters at runtime.
This includes the default formatters and filters.
According to the docs of boost log, filters and formatters written as expression templates, tend to perform better than the ones parsed from strings, as types are unknown when parsing.

As the defaults are fixed and logging is a big part of preCICE, we are potentially leaving performance on the table.

Add expression templates for the default filters and formatters.
These two, however need to stay in sync.

Exclusively rely on parsing.

aadhil2k4:
## Main changes of this PR
- Added createDefaultFormatter(), createDefaultFilter() using Boost.Log expression templates to replicate existing formatter, filter strings.
- Updated sink to use expression based implementations when configured filter, formatter equals default.
- Runtime parsing for all custom user defined filter, formatters is unchanged.

Expression templates perform better than string parsed filters, formatters, since type are known at compile time and no runtime parsing is required.
Closes #1428 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/logging/LogConfiguration.cpp
+++ b/src/logging/LogConfiguration.cpp
@@ -166,6 +166,38 @@ const std::string BackendConfiguration::default_formatter = "(%Rank%) %TimeStamp
 const std::string BackendConfiguration::default_type      = "stream";
 const std::string BackendConfiguration::default_output    = "stdout";
 
+static boost::log::formatter createDefaultFormatter()
+{
+  namespace expr    = boost::log::expressions;
+  namespace trivial = boost::log::trivial;
+  auto severity     = expr::attr<trivial::severity_level>("Severity");
+  return expr::stream
+         << "("
+         << expr::attr<int>("Rank")
+         << ") "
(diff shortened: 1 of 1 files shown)


aadhil2k4:
Hi @fsimonis, can you please review this PR.

src/logging/LogConfiguration.cpp
```diff
@@ -192,6 +192,38 @@ void BackendConfiguration::setEnabled(bool enabled)
   this->enabled = enabled;
 }
 
+static boost::log::formatter createDefaultFormatter()
```
fsimonis, line 169: Could you move these up so that they are next to the string definitions? This should make it more difficult for them to get our of sync.

fsimonis:
Looking good. One last move and I'll approve the workflows.

aadhil2k4:
Thanks for the review!! I have made the change.

Issue #2093, opened 2024-10-02
Allow remeshing directly after initialization


fsimonis:
**Please describe the problem you are trying to solve.**

The current implementation of remeshing doesn't allow remeshing after initialization.
This can be useful if an ALE solver needs to check straight away if remeshing is required/benefitial based on the initial displacements.

This is applicable in scenarios where meshes are not optimized or cannot be optimized before the simulation due to availability, format, or similar.

**Describe the solution you propose.**

Allow remeshing after between initialize and the first advance.

**Describe alternatives you've considered**

None

**Additional context**

#2070

PR #2484, opened 2026-03-04
Allow remeshing directly after initialization (#2093)

prakhar0608shrma:
## Main changes of this PR

Allow `resetMesh()` to be called directly after `initialize()`, before any `advance()` call.

Related to #2093.
 Previously, calling `resetMesh()` right after `initialize()` failed because `isTimeWindowComplete()` returns false before the first advance. The fix adds a `_numberAdvanceCalls == 0` check so remeshing is permitted immediately after initialization. Two new integration tests cover parallel-explicit and parallel-implicit coupling schemes.

## Author's checklist

- [ ] I used the pre-commit hook to prevent dirty commits and used `pre-commit run --all` to format old commits.
- [x] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
- [x] I added a test to cover the proposed changes in our test suite.
- [ ] For breaking changes: I documented the changes in the appropriate porting guide.
- [ ] I ran the systemtests by adding the label `trigger-system-tests` (may be skipped if minor change)

## Reviewers' checklist

- [ ] Does the changelog entry make sense? Is it formatted correctly?
- [ ] Do you understand the code changes?

### DIFF
--- a/docs/changelog/2093.md
+++ b/docs/changelog/2093.md
@@ -1,0 +1,1 @@
+- Allow remeshing directly after initialization, before the first call to advance(). This is useful for ALE solvers that need to check and adjust mesh configuration based on initial displacements. See #2093.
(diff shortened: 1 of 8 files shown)


fsimonis:
@prakhar0608shrma This will take some time to review.

PR #2485, opened 2026-03-04
Make Participant API thread-safe via a single coarse-grained mutex

### LINKED ISSUE #2407: Thread-safe Participant

VanshKharb:
## Summary

This PR addresses issue #2407 by introducing a coarse-grained `std::mutex` at the public `Participant` API boundary, making all API calls safe to call from multiple threads without risking data races or undefined behaviour.

## Motivation

preCICE is often embedded in parallel solver codes that may call the `Participant` API from multiple threads — for example, when different solver components (e.g. fluid and thermal) are managed on separate threads but share a single `Participant` instance. Without synchronisation, concurrent calls to methods like `writeData`, `readData`, `advance`, or any mesh-manipulation routines can produce data races on the internal `ParticipantImpl` state.

## What changed

**`src/precice/Participant.hpp`**
- Added `#include <mutex>` to the header.
- Added a `mutable std::mutex _mutex` to the private section of the `Participant` class.
  Using `mutable` allows const-qualified methods (e.g. `readData`, `isCouplingOngoing`) to acquire the lock without removing their `const` qualification.

**`src/precice/Participant.cpp`**
- Added `std::scoped_lock lock{_mutex};` as the first line of every non-constructor, non-destructor public method.
  `std::scoped_lock` is RAII, exception-safe, and available since C++17 (which preCICE already requires).

## Design notes

- **Granularity**: A single mutex at the API boundary is the simplest correct approach. It serialises all concurrent API calls, which is safe and has negligible overhead in practice — the actual computation is in the solver, not here. Finer-grained locking (e.g. per-mesh or per-data locks) would require invasive changes across `ParticipantImpl`, `ReadDataContext`, `WriteDataContext`, and `Mesh`, and would be a larger, separate effort.
- **`readData` / `writeData`**: These methods are locked just like the rest. They access internal waveform storage and coupling-scheme time state, which can be modified concurrently by `advance()`. Leaving them unprotected would introduce subtle time-dependent bugs.
- **Constructors / Destructor**: Not locked — it is the caller's responsibility to ensure a `Participant` object is not used from multiple threads while still being constructed or destroyed (standard C++ rule).

## Testing

Compiled and ran the full test suite locally (1290 tests, covering unit, integration, and parallel MPI scenarios). All tests pass.

```
100% tests passed, 0 tests failed out of 1290
```

Closes #2407

### DIFF
--- a/src/precice/Participant.cpp
+++ b/src/precice/Participant.cpp
@@ -45,73 +45,87 @@ Participant::~Participant() = default;
 
 void Participant::initialize()
 {
+  std::scoped_lock lock{_mutex};
   _impl->initialize();
 }
 
 void Participant::advance(
     double computedTimeStepSize)
 {
+  std::scoped_lock lock{_mutex};
   _impl->advance(computedTimeStepSize);
 }
 
 void Participant::finalize()
 {
+  std::scoped_lock lock{_mutex};
   return _impl->finalize();
 }
 
 int Participant::getMeshDimensions(::precice::string_view meshName) const
 {
+  std::scoped_lock lock{_mutex};
   return _impl->getMeshDimensions(toSV(meshName));
 }
 
 int Participant::getDataDimensions(::precice::string_view meshName, ::precice::string_view dataName) const
 {
+  std::scoped_lock lock{_mutex};
   return _impl->getDataDimensions(toSV(meshName), toSV(dataName));
 }
 
 bool Participant::isCouplingOngoing() const
 {
+  std::scoped_lock lock{_mutex};
   return _impl->isCouplingOngoing();
 }
 
 bool Participant::isTimeWindowComplete() const
 {
+  std::scoped_lock lock{_mutex};
   return _impl->isTimeWindowComplete();
 }
 
 double Participant::getMaxTimeStepSize() const
 {
+  std::scoped_lock lock{_mutex};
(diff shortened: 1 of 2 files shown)


fsimonis:
Doesn't follow the PR template

PR #2486, opened 2026-03-04
Simplify mapping sanity checks in ParticipantConfiguration

### LINKED ISSUE #2156: Dead code in ParticipantConfiguration
Most of the following code is dead

https://github.com/precice/precice/blob/f239c3c69efc8718216da296cc0ccc7cc46c4f5a/src/precice/config/ParticipantConfiguration.cpp#L415-L441

because the first check (in both if branches) includes already the subsequent two. Seems to be a consequence of #2124.
We should remove the dead part or refine the error message in a suitable way.

PranjalManhgaye:
## Main changes of this PR

- Removed the redundant "swapped mapping" sanity checks in `ParticipantConfiguration::finishParticipantConfiguration` (the ones that said "needs to map from received to provided / provided to received, but they are swapped").
- Left the specific checks as-is — they already fail with clearer messages like "without receiving it" / "without providing it" and tell you which tag to add.

Closes #2156.

The issue pointed out that after #2124 the first check in each branch (read vs write) was effectively dead: if that check would fail, one of the two following checks would already have failed with a better message. So we're just deleting the redundant part. Behavior is unchanged — invalid configs still fail, same messages for the user, less code to maintain.

## Author's checklist

* [x] I used the pre-commit hook and ran `pre-commit run --all` on the branch.
* [ ] I added a changelog file with `make changelog` if there are user-observable changes.
* [ ] I added a test to cover the proposed changes.
* [ ] For breaking changes: I documented them in the porting guide.
* [x] I stuck to C++17 and CMake 3.22.1.
* [x] Single commit, squashed.
* [ ] I ran systemtests (or skipped for this minor change).

## Reviewers' checklist

* [ ] Changelog makes sense (if any).
* [ ] Code changes are clear.

### DIFF
--- a/src/precice/config/ParticipantConfiguration.cpp
+++ b/src/precice/config/ParticipantConfiguration.cpp
@@ -441,10 +441,6 @@ void ParticipantConfiguration::finishParticipantConfiguration(
     // sanity checks
     if (confMapping.direction == mapping::MappingConfiguration::Direction::READ) {
       // A read mapping maps from received to provided
-      PRECICE_CHECK(participant->isMeshReceived(fromMesh) || confMapping.toMesh->isJustInTime() || participant->isMeshProvided(toMesh),
-                    "A read mapping of participant \"{}\" needs to map from a received to a provided mesh, but in this case they are swapped. "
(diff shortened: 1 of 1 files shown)


fsimonis:
Thanks! Merging

PR #2488, opened 2026-03-05
Add XML tag context to attribute error messages

### LINKED ISSUE #2441: Missing XML tag context

```xml
```

```bash
```
 
from this code

VanshKharb:
## Description

Closes #2441

When a required XML attribute is missing or has an invalid value in the preCICE configuration file, the error message now includes the name of the parent XML tag. This makes it significantly easier for users to find and fix configuration errors.

## Problem

Currently, when a required attribute is missing, the error reads:

```
ERROR: Attribute "from" is required, but was not defined.
```

For large configuration files with multiple tags of the same type (e.g., multiple `<receive-mesh>` entries), it is very difficult to identify which tag has the missing attribute.

## Solution

Pass the parent tag's full name to `XMLAttribute::readValue()` and include it in both the missing-attribute and invalid-value error messages.

**Before:**
```
ERROR: Attribute "from" is required, but was not defined.
```

**After:**
```
ERROR: Tag <receive-mesh>: Attribute "from" is required, but was not defined.
```

## Changes

- `src/xml/XMLAttribute.hpp`: Updated `readValue()` to accept a `std::string_view tagName` parameter and include it in both the missing-attribute and invalid-value error messages.
- `src/xml/XMLTag.cpp`: Updated `XMLTag::readAttributes()` to pass `_fullName` to each attribute's `readValue()` call.

## Testing

- All existing XML tests pass (12/12).
- Full test suite passes — build and all tests exit with code 0.
- Manually verified with `precice-tools check` on a config with a missing `from` attribute — the new error message correctly shows `Tag <receive-mesh>: Attribute "from" is required, but was not defined.`


fsimonis:
Not following PR template and untidy history

PR #2490, opened 2026-03-07
[XML] Add tag context to attribute error messages

VanshKharb:
## Description
This PR improves the developer/user experience when debugging malformed XML configuration files by adding the exact XML tag name to attribute error messages. Addresses a core part of #751.

**Before:**
If a user forgets a required attribute like `name` on a `<mesh>` tag, or mistypes `dimensions="3"`, preCICE would output:
> `Attribute "name" is required, but was not defined.`
> `Invalid value "3D" of attribute "dimensions": ...`
*The user is forced to manually `grep` their entire `precice-config.xml` file to figure out which tag caused the error.*

**After:**
The error message now explicitly includes the owning tag name:
> `Tag <mesh>: Attribute "name" is required, but was not defined.`
> `Tag <mesh>: Invalid value "3D" of attribute "dimensions": ...`

## Implementation Details
This is achieved with a highly targeted, surgical change:
* Modified `XMLAttribute::readValue()` to accept the `tagName` and prepend `Tag <tagName>:` to missing/invalid attribute exceptions.
* Modified `XMLTag::readAttributes()` to capture `_fullName` in its lambda and pass it down.
* Added `MissingRequiredAttribute` and `InvalidAttributeOption` test cases to `XMLTest.cpp` utilizing `BOOST_CHECK_EXCEPTION` and `precice::testing::errorContains`.

## Context vs #2467
PR #2467 addresses #751 by injecting full libxml2 line numbers into all errors. This PR (#) takes a different, complimentary approach:
1. **Low complexity:** It requires only a 5-line change in core logic.
2. **Immediate usability:** It fully resolves the pain of locating attribute errors.
3. **Complimentary:** Tag names are highly useful context even when line numbers are present (e.g., when configurations are generated dynamically or concatenated). 

Happy to make any adjustments! (Working towards a GSoC 2026 application).


fsimonis:
Not following PR template

PR #2491, opened 2026-03-08
Enhance preconditioner warnings with sub-vector names

### LINKED ISSUE #1515: Warning for sub-vector becoming numerically zero missing data context

VanshKharb:
## Main changes of this PR

This PR enhances the warnings produced by `ResidualSumPreconditioner`, `ResidualPreconditioner`, and `ValuePreconditioner` when a sub-vector becomes numerically zero. It modifies the `Preconditioner` interface to recursively pass down and utilize sub-vector names, and updates `AitkenAcceleration` and `BaseQNAcceleration` to pass these data names directly from the `CouplingData`.

This provides users with better context regarding which data is numerically zero (e.g. `data: "Forces"`), rather than an obscure error. 
Closes #1515.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/acceleration/AitkenAcceleration.cpp
+++ b/src/acceleration/AitkenAcceleration.cpp
@@ -42,8 +42,12 @@ void AitkenAcceleration::initialize(const DataMap &cplData)
   // Accumulate number of entries
   // Size for each subvector needed for preconditioner
   std::vector<std::size_t> subVectorSizes;
-  // Gather sizes
-  std::transform(_primaryDataIDs.cbegin(), _primaryDataIDs.cend(), std::back_inserter(subVectorSizes), [&cplData](const auto &d) { return cplData.at(d)->getSize(); });
+  // Gather sizes and names
+  std::vector<std::string> subVectorNames;
+  for (const auto &d : _primaryDataIDs) {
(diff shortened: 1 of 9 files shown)


fsimonis:
Already implemented in #2473

PR #2495, opened 2026-03-09
Refactor typename getter logic

nots1dd:
## Main changes of this PR

Previous implementation had unnecessary string allocs and logically for a logging sense (used in xml printers) string_view makes more sense here.

Like I mentioned in #2470, there are 2 ways to invoke this that can utilize constexpr nicely:

-> we invoke getTypeName(): will go thru a template forwarder that will evalute at RT

-> if getTypeName<T>() is invoked this is a compile time evaluation so explicit instantiations are now possible with this.

We were discussing about using something like `std::is_scalar` however there are a few problems with that:

1. It may look decent within the function scope itself, but we want to print this to ostream and we need the specific dtype, if we pass is_scalar to int, float, double, bool, etc. they all will give true

so it really doesnt help here and we would need logic to write that specific dtype to string or string_view anyways.

2. Other custom types like eigen vectorxd will return false so we are back to square one. is_scalar may look idiomatic but in this case where we simply need to return the string_view for printing only, it isnt ideal.

Mentioned in #2470 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/2459.md
+++ b/docs/changelog/2459.md
@@ -1,0 +1,1 @@
+- Added config validation error when neither max-time nor max-time-windows is defined in a coupling scheme.
--- a/src/cplscheme/config/CouplingSchemeConfiguration.cpp
+++ b/src/cplscheme/config/CouplingSchemeConfiguration.cpp
@@ -321,6 +321,13 @@ void CouplingSchemeConfiguration::xmlEndTagCallback(
 {
   PRECICE_TRACE(tag.getFullName());
   if (tag.getNamespace() == TAG) {
+    PRECICE_CHECK(_config.maxTime != CouplingScheme::UNDEFINED_MAX_TIME ||
+                      _config.maxTimeWindows != CouplingScheme::UNDEFINED_TIME_WINDOWS,
+                  "At least one termination condition is required "
+                  "for the coupling scheme of type \"{}\". "
+                  "Please add a <max-time value=\"...\"/> or <max-time-windows value=\"...\"/> tag "
+                  "inside your <coupling-scheme:{}> configuration.",
+                  _config.type, _config.type);
     if (_config.type == VALUE_SERIAL_EXPLICIT) {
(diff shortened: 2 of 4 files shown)


nots1dd:
@fsimonis I have separated the PR contents as requested and ive stuck with my original logic for getTypeName and gave valid justification, do look at it and give ur opinion, and thanks for the same

fsimonis:
> We were discussing about using something like std::is_scalar however there are a few problems with that:
> It may look decent within the function scope itself, but we want to print this to ostream and we need the specific dtype, if we pass is_scalar to int, float, double, bool, etc. they all will give true

This looks like you misunderstood the comparison I was trying to present. `is_scalar_v` is only an example of this technique. Let's put that aside for a moment.

What you do now is to deduce the argument type as a template parameter, pass that to a trait and return its specialized string_view.
The modern and less verbose version of this is to use a single function with auto argument to remove the explicit template, then extract its type, remove cv and use it in a chain of constexpr if where you check is_same_v.

That's a lot of C++ details, but implementing this will be a valuable learning experience for you.

nots1dd:
Hi @fsimonis ,

> The modern and less verbose version of this is to use a single function with auto argument to remove the explicit template, > then extract its type, remove cv and use it in a chain of constexpr if where you check is_same_v.

Hm seems interesting, however i mentioned that explicitly invoking a template is not necessary but an added qol per se

if u notice there is a forwarder that has the same function signature as the previous logic:

```cpp
// value based forwarder
template <typename T>
constexpr std::string_view getTypeName(const T &)
{
  return getTypeName<T>();
}
```

so all calls to getTypeName have remained the same and internally it gets the type and returns the right string_view

and although i understand where u are coming from, i am still confused on how Eigen::VectorXd will be handled

nots1dd:
oh wait am dumb nvm sorry i was overcomplicating it, will change this and squash (didnt understand what u meant i got it now)

nots1dd:
ok i messed up the branch will reopen another PR

PR #2496, opened 2026-03-09
Make Participant API thread-safe

### LINKED ISSUE #2407: Thread-safe Participant

VanshKharb:
## Main changes of this PR

This PR makes the preCICE C++ API thread-safe by implementing mutex locks in the `Participant` class. It adds a `mutable std::mutex _mutex` to the private section of the `Participant` class and wraps every non-constructor, non-destructor public method call in a `std::scoped_lock`.

Currently, concurrent calls to the preCICE API from multiple threads can lead to race conditions. This PR serializes all API calls at the boundary level to ensure safe internal state modifications without adding significant overhead to the solvers. 
Closes #2407.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/precice/Participant.cpp
+++ b/src/precice/Participant.cpp
@@ -45,73 +45,87 @@ Participant::~Participant() = default;
 
 void Participant::initialize()
 {
+  std::scoped_lock lock{_mutex};
   _impl->initialize();
 }
 
 void Participant::advance(
     double computedTimeStepSize)
 {
+  std::scoped_lock lock{_mutex};
   _impl->advance(computedTimeStepSize);
 }
 
 void Participant::finalize()
 {
+  std::scoped_lock lock{_mutex};
   return _impl->finalize();
 }
 
 int Participant::getMeshDimensions(::precice::string_view meshName) const
 {
+  std::scoped_lock lock{_mutex};
   return _impl->getMeshDimensions(toSV(meshName));
 }
 
 int Participant::getDataDimensions(::precice::string_view meshName, ::precice::string_view dataName) const
 {
+  std::scoped_lock lock{_mutex};
   return _impl->getDataDimensions(toSV(meshName), toSV(dataName));
 }
 
 bool Participant::isCouplingOngoing() const
 {
+  std::scoped_lock lock{_mutex};
   return _impl->isCouplingOngoing();
 }
 
 bool Participant::isTimeWindowComplete() const
 {
+  std::scoped_lock lock{_mutex};
   return _impl->isTimeWindowComplete();
 }
 
 double Participant::getMaxTimeStepSize() const
 {
+  std::scoped_lock lock{_mutex};
   return _impl->getMaxTimeStepSize();
 }
 
 bool Participant::requiresInitialData()
 {
+  std::scoped_lock lock{_mutex};
   return _impl->requiresInitialData();
 }
 
 bool Participant::requiresReadingCheckpoint()
 {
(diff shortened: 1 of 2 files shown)


fsimonis:
Hi, I see that you want to merge to main.
The default is develop, so you must have manually changed this.
What is your motivation to do so?

VanshKharb:
Ah my bad, didn't realize `develop` was the default. Just changed it back, let me know if I need to rebase anything!

nots1dd:
shudnt synchronization happen in the impl (src/precice/impl/ParticipantImpl.cpp)? seems that the repo uses pimpl, and normally all threading and locking logic is contained in the impl afaik

VanshKharb:
Good catch! Yeah, the pImpl pattern here is pretty clear — but the reason I locked in `Participant.cpp` (the public facade) rather than in `ParticipantImpl` is that this is the actual entry point users are calling from multiple threads. Putting the lock there means every public API call is protected before it even touches the impl, so we don't have to worry about any future method additions slipping through without a lock.

If the maintainers prefer it in `ParticipantImpl`, I'm totally fine moving it — just wanted to make sure the public boundary was the guarded one.

nots1dd:
> Good catch! Yeah, the pImpl pattern here is pretty clear — but the reason I locked in `Participant.cpp` (the public facade) rather than in `ParticipantImpl` is that this is the actual entry point users are calling from multiple threads. Putting the lock there means every public API call is protected before it even touches the impl, so we don't have to worry about any future method additions slipping through without a lock.
> 
> If the maintainers prefer it in `ParticipantImpl`, I'm totally fine moving it — just wanted to make sure the public boundary was the guarded one.

Thats not what pimpl means structurally, all "internal" implementations (including concurrency and synchronization) is supposed to be in `ParticipantImpl`, if that is implemented correctly u dont have to worry about the public api

take a look at any pimpl refs online ([example](https://medium.com/@weidagang/modern-c-the-pimpl-idiom-53173b16a60a)) and u will understand what am tryna say

VanshKharb:
yeah fair point, i was thinking about it wrong. if the impl is properly locked then the public api is protected anyway, no need to put it in the facade. will move it into ParticipantImpl, thanks for the link

fsimonis:
@VanshKharb Hi, are you still working on this?

VanshKharb:
not really, apologies for not updating you

fsimonis:
Ok. I'll close it then.

PR #2478, opened 2026-03-03
Improve XML missing-attribute error context

PranjalManhgaye:
## Main changes of this PR

- **include tag name in required-attribute errors**: extend `precice::xml::XMLAttribute::readValue` to take the tag name and use it when reporting missing required attributes.
- **Wire tag context through `XMLTag`**: update `precice::xml::XMLTag::readAttributes` to pass the tag’s `_fullName` to all attributes so error messages can mention the concrete tag (e.g. `<receive-mesh>`).
- **Add regression test**: introduce a minimal XML configuration (`src/xml/tests/config_xmltest_missing_attribute.xml`) and a new unit test `MissingRequiredAttributeIncludesTagName` in `ParserTest.cpp` to verify that missing-attribute errors now include the tag name.

## Motivation and additional information

This PR addresses the usability issue described in **#2441 (Missing XML tag context)**.

Previously, when a required attribute was missing, preCICE emitted an error like:

> `Attribute "from" is required, but was not defined.`

While technically correct, this gave no indication of **where** in the XML the problem occurred, which is particularly confusing when a configuration contains multiple tags that accept the same attribute name.

With this change, the error includes the tag name, for example:

> `The tag <receive-mesh> in the configuration is missing required attribute "from".`

This makes it much easier for users to locate and fix configuration mistakes, and is consistent with other XML configuration errors that already include tag context (e.g. for unknown attributes). The change is implemented in a generic way, so all missing-required-attribute errors now include their tag name, not just `<receive-mesh>`.

The new unit test ensures that this behavior stays stable.

## Author's checklist

* [x] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [ ] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [x] I added a test to cover the proposed changes in our test suite.
* [ ] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).
* [x] I stuck to C++17 features.
* [x] I stuck to CMake version 3.22.1.
* [ ] I squashed / am about to squash all commits that should be seen as one.
* [ ] I ran the systemtests by adding the label `trigger-system-tests` (may be skipped if minor change)

## Reviewers' checklist

* [ ] Does the changelog entry make sense? Is it formatted correctly?
* [x] Do you understand the code changes?

PR #2497, opened 2026-03-09
Add XML tag context for missing attributes

### LINKED ISSUE #2441: Missing XML tag context

nots1dd:
## Main changes of this PR

This adds more context for missing attributes and ideally closes #2441 

Most of the changes revolve around config parser, xmltag then finally xmlattr classes.

Logical flow of the new changes:

ConfigParser::connectTags - new param (fullXMLCtx) => generated recursively within

connectTags then invokes XMLTag::readAttributes - new param (fullXMLCtx) => directly passed onto XMLAttribute's readAttribute method

XMLAttribute::readAttribute - new param (fullXMLCtx) if any error occurs, it will print _name and fullXMLCtx that is much more detailed

New info that gets printed are the existing attributes, participant name where the error occurred and the tag where the tag is missing.

Unsure on whether this change requires to be mentioned in changelog or add a new test for since this new param is solely being used to print to stdout (all tests pass so no errors or UB occurs in xml logic with this change)

New message:

image

Resolves #2441 and in the future similar logic can be utilized for any other error msgs that require more context.

A cleaner method could be to store the context in a struct and pass that around (and access its members to then print), however pertaining to the specific issue it did not seem necessary.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/xml/XMLAttribute.hpp
+++ b/src/xml/XMLAttribute.hpp
@@ -78,7 +78,7 @@ public:
     return _hasValidation;
   };
 
-  void readValue(const std::map<std::string, std::string> &aAttributes);
+  void readValue(std::string_view tagName, const std::map<std::string, std::string> &aAttributes);
 
   const std::string &getName() const
   {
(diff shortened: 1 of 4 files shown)


nots1dd:
@fsimonis @davidscn (since david originally made the issue i figured i can get ur view) how does this look? i debated on whether adding a new struct for this but didnt seem necessary but i believe the general way of adding more ctx to the xml errors would follow this.

If there are any other cases where there is a requirement of adding more xml ctx perhaps i can adopt a similar approach to them as well

nots1dd:
ah i see that #2478 has alr been merged however this pr also includes more details like the participant name as well so this might still be worth looking into

fsimonis:
@nots1dd If you think that there is extra value here after #2478 has been merged, then please rebase this PR and update the title.

PR #2499, opened 2026-03-09
Raise error if receive data is declared but not exchanged (#1836)

### LINKED ISSUE #1836: Raise error, if configuration contains data that is not exchanged

#1803 describes an assertion hit resulting from an invalid configuration. The corresponding invalid configuration and fix can be seen in https://github.com/precice/tutorials/pull/384.

Raise an error, if data is declared, but not exchanged. Note: There are some valid use-cases (mappings, actions) that lead to similar, but valid configurations.

Just raise the assertion as it is currently the case.

**Additional context**

We don't have a system in place that allows us to test for errors (see #280)

PranjalManhgaye:
- Add hasSamples check in readData and mapAndReadData before reading
- Add PRECICE_CHECK in DataContext::mapData when mapping receive data with no samples
- Suppress destructor exceptions during cleanup so init errors propagate
- Add regression test ReceiveDataNotExchanged

Closes #1836

## Main changes of this PR

- Add checks using `ReadDataContext::hasSamples()` in `ParticipantImpl::readData` and `ParticipantImpl::mapAndReadData`. If receive data has no samples, a clear `precice::Error` is raised instead of continuing and failing later.

- Use the existing `PRECICE_CHECK` in `DataContext::mapData` so that mapping receive data which was declared but never exchanged fails with a clear error message rather than an internal assertion.

- Wrap the implicit `finalize()` call in `ParticipantImpl::~ParticipantImpl()` in a try/catch block. This avoids destructor-time exceptions masking the actual error thrown during `initialize()`.

- Add a regression test (`ReceiveDataNotExchanged.cpp` and `ReceiveDataNotExchanged.xml`) reproducing the configuration from #1836 and verifying that the correct error is raised.

- Add a changelog entry in `docs/changelog/1836.md`.

The configuration in #1836 declares `<read-data>` for a participant, but the corresponding data is never exchanged.

Before this change, this situation could lead to an assertion failure somewhere in the mapping or initialization logic, which made the problem hard to understand from a user perspective.

With this PR, preCICE now detects that the receive data contains no samples and throws a clear `precice::Error` explaining that the data was declared but never exchanged. The error message points users to check the `<exchange>` configuration.

The regression test ensures this behavior remains covered, and suppressing destructor exceptions ensures that initialization errors are not hidden during cleanup.
   **Local testing**

   `ctest -R "Integration/Serial/Configuration/ReceiveDataNotExchanged" --output-on-failure`

image

## Author's checklist

* [ ] For breaking changes: I documented the changes in the appropriate porting guide.

## Reviewers' checklist

### DIFF
--- a/docs/changelog/1836.md
+++ b/docs/changelog/1836.md
@@ -1,0 +1,1 @@
+- Raised an error when receive data is declared but not exchanged in the coupling configuration.
--- a/src/precice/impl/ParticipantImpl.cpp
+++ b/src/precice/impl/ParticipantImpl.cpp
@@ -166,7 +166,12 @@ ParticipantImpl::~ParticipantImpl()
 {
   if (_state != State::Finalized) {
     PRECICE_INFO("Implicitly finalizing in destructor");
-    finalize();
+    try {
+      finalize();
+    } catch (...) {
+      // Suppress exceptions during cleanup so an in-flight exception (e.g. from
+      // initialize()) can propagate to the caller
+    }
   }
 }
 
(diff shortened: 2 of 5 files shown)


PranjalManhgaye:
@fsimonis , is there anything you want to suggest , i'm happy to update it , thanks

fsimonis:
I cannot review this as the error message and the test are not connected.

The error condition is already handled in initialize and can only be triggered by manually calling the API.
The test doesn't use JIT and already fails in initialize.

PR #2501, opened 2026-03-10
Implement RAII for xmlParserCtxt in ConfigParser

brightyorcerf:
## Main changes of this PR

- Wrapped the raw `xmlParserCtxtPtr` in `std::unique_ptr` with `xmlFreeParserCtxt` as a custom deleter, ensuring the context is freed during stack unwinding if an exception occurs.
- Added a `PRECICE_CHECK` to guard against a null return from `xmlCreatePushParserCtxt` before parsing begins.

`ConfigParser::readXmlFile` relied on manual cleanup of the libxml2 parser context. If `xmlParseChunk` or a SAX callback threw an exception, `xmlFreeParserCtxt` would be skipped, leaking the context. Additionally, the return value of `xmlCreatePushParserCtxt` was not checked, risking a null pointer dereference on failed initialization.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/xml/ConfigParser.cpp
+++ b/src/xml/ConfigParser.cpp
@@ -188,12 +188,14 @@ int ConfigParser::readXmlFile(std::string const &filePath)
 
   _hash = utils::preciceHash(content);
 
-  xmlParserCtxtPtr ctxt = xmlCreatePushParserCtxt(&SAXHandler, static_cast<void *>(this),
-                                                  content.c_str(), content.size(), nullptr);
+  auto ctxt = std::unique_ptr<xmlParserCtxt, void (*)(xmlParserCtxtPtr)>(
+      xmlCreatePushParserCtxt(&SAXHandler, static_cast<void *>(this),
+                              content.c_str(), content.size(), nullptr),
+      xmlFreeParserCtxt);
 
-  xmlParseChunk(ctxt, nullptr, 0, 1);
-  xmlFreeParserCtxt(ctxt);
-  xmlCleanupParser();
+  PRECICE_CHECK(ctxt != nullptr, "XML parser was unable to create a push parser context for file \"{}\"", filePath);
+
+  xmlParseChunk(ctxt.get(), nullptr, 0, 1);
 
   return 0;
 }


brightyorcerf:
Hi everyone! This is my first contribution to preCICE. I wanted to start with a small surgical fix for a latent resource management issue I noticed in the XML parser.

I also had a quick question regarding workflow: I noticed the issue tracker is silent; is the standard practice here to submit PRs directly for improvements rather than opening an issue first? I’ve read the contribution guidelines and want to make sure I’m respecting the team’s preferred communication style. 

Thanks for the guidance!

fsimonis:
Hi,
we don't use the prefixing of PR titles nor commits in preCICE. 
Also, your markdown isn't formatted correctly. Same problem as in the issue.
You also altered the PR template for no reason. Please restore the checklists.

> I noticed the issue tracker is silent; is the standard practice here to submit PRs directly for improvements rather than opening an issue first?

Depends. The issue tracker is for larger TODOs, bugs and feature requests.
For some tinkering from the maintainer side, the friction is too big to open an issue first.

In general avoid working on something that other people are already working on. There is enough to cleanup in the repo.

brightyorcerf:
@fsimonis 

Fixed (restored the checklist format, removed the prefix from the title and commit, and cleaned up the markdown). 
Thanks for the clarification.

src/xml/ConfigParser.cpp
```diff
@@ -188,11 +188,14 @@ int ConfigParser::readXmlFile(std::string const &filePath)
...
+      xmlCreatePushParserCtxt(&SAXHandler, static_cast<void *>(this),
+                              content.c_str(), content.size(), nullptr),
+      xmlFreeParserCtxt);
 
-  xmlParseChunk(ctxt, nullptr, 0, 1);
-  xmlFreeParserCtxt(ctxt);
+  PRECICE_CHECK(ctxt != nullptr, "XML parser was unable to create a push parser context for file \"{}\"", filePath);
+
+  xmlParseChunk(ctxt.get(), nullptr, 0, 1);
   xmlCleanupParser();
```
fsimonis: According to the docs, libxml2 calls after `xmlCleanupParser()` lead to memory corruptions.

So, by fixing a potential memory corruption, you added a definite memory corruption. 😅

The suggestion from the library is to not call this manually. So, let's remove this call and see what happens.

brightyorcerf: Good catch! I’ve removed the xmlCleanupParser() call.

I was looking into the Windows/Clang CI failures yesterday and identified the xmlCleanupParser() ordering and a potential function pointer mismatch as the primary candidates.

After removing the manual cleanup, all local tests pass cleanly. This confirms the ordering issue (where the RAII wrapper attempted to destroy the context after the global library cleanup) was the root cause.

fsimonis:
Thanks for the contribution.

PR #2503, opened 2026-03-10
Add tag context to XML attribute error messages

### LINKED ISSUE #2441: Missing XML tag context

VanshKharb:
## Main changes of this PR

When a required XML attribute is missing, or has an invalid value, the error message now includes the name of the parent XML tag to help users quickly identify where the missing attribute is expected.
Also adds a test to cover this.

Closes #2441. 
The previous error messages were lacking context, making it hard to find the problematic XML tag.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/2490.md
+++ b/docs/changelog/2490.md
@@ -1,0 +1,1 @@
+- Add tag context to XML attribute error messages to improve debugging (#2441).


fsimonis:
Superseded by #2478

Please take a look if other people are working on the same issue. There is no value in adding multiple PRS.

PR #2504, opened 2026-03-10
Modify XML missing attribute context logic

nots1dd:
## Main changes of this PR

This change mainly reflects on #2478 , while it did add more context, this commit adds more info about the parent xml context (like the participant name)

If multiple missing attributes exist with same snippet spread across multiple participants, the current logic would still provide a vague output, meanwhile this change can pinpoint the issue directly

This builds upon the work I made in #2497  and is linked to issue #2441 

Here is how the final output looks like:

image

image

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/xml/ConfigParser.cpp
+++ b/src/xml/ConfigParser.cpp
@@ -132,7 +132,8 @@ ConfigParser::ConfigParser(std::string_view filePath, const ConfigurationContext
     SubTags.push_back(m_AllTags[0]);
 
   try {
-    connectTags(context, DefTags, SubTags);
+    // parent xml context here is nil, constructed later on.
+    connectTags(context, DefTags, SubTags, "");
   } catch (::precice::Error &) {
     throw;
   } catch (const std::exception &e) {
@@ -222,7 +223,7 @@ auto gatherCandidates(const std::vector<std::shared_ptr<XMLTag>> &DefTags, std::
 }
 } // namespace
 
-void ConfigParser::connectTags(const ConfigurationContext &context, std::vector<std::shared_ptr<XMLTag>> &DefTags, CTagPtrVec &SubTags)
+void ConfigParser::connectTags(const ConfigurationContext &context, std::vector<std::shared_ptr<XMLTag>> &DefTags, CTagPtrVec &SubTags, const std::string &parentXMLCtx)
 {
   std::unordered_set<std::string> usedTags;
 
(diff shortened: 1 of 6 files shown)


nots1dd:
@fsimonis hi i have rebased and opened a new pr that explains what this adds on top of the previous PR (more on #2497)

tldr we can fetch parent xml context thru config parser and pass it down to xmlattribute (or xmltag itself)

we propogate this context to wherever we require to print an error or log with more context of where the problem is

note that my commit description also has this rationale to explain why this is better than the current implementation

PranjalManhgaye:
Good build on #2478 ,nice follow up — the parent context in the message will help a lot. one note: if the missing-attribute tag is at root, there’s no parent, so the message won’t contain "in inside" and the test might fail. Worth adjusting the test or adding a root-level case, thanks @nots1dd

nots1dd:
> Good build on #2478 ,nice follow up — the parent context in the message will help a lot. one note: if the missing-attribute tag is at root, there’s no parent, so the message won’t contain "in inside" and the test might fail. Worth adjusting the test or adding a root-level case, thanks @nots1dd

can u give some example xml snippets where u think this might happen? i have tested it with `examples/solverdummies` and the general missing attr test and did not find a scenario where `inside {}` is not being printed. Regardless of the context, logically this should work under certain pretense:

-> the xml file is formatted correctly (with defined participant and no missing tags)

so for example say we omit `<participant name="SolverTwo"> ` in solverdummies (or dont define it properly) i will get an error like so:

```text
Opening and ending tag mismatch: precice-configuration line 2 and participant
```

so only after config parser is able to parse xml properly will xml attributes and tags be checked and thats where this logic resides.

~i could be wrong tho but i havent found any cases where this doesnt apply~

just went thru the code and flow again, the only scenario i can think of where we cant get the context properly is if the root tag itself has missing attributes which is unlikely since they seem to be optional afaik 

And if some unexpected attribute is put in the root or any tag it throws:

```text
(0) 17:12:40 [xml::XMLTag]:209 in readAttributes: ERROR: The tag <precice-configuration> in t
he configuration contains an unknown attribute "version". Expected attributes are experimenta
l, allow-remeshing, wait-in-finalize.
```

an error like so, yea i highly doubt there is an edge case here.

PranjalManhgaye:
thanks for the detailed explanation , i went through the flow again and it makes sense to me now , previously  ig 
the case i thought was somewhat hypothetical , Your point about the parser only running the attribute checks after the structure is valid,  root attributes being either optional or handled by the existing “unknown attribute” error in readAttributes, convinces me that a missing-attribute case on the root without a parent context is effectively excluded in practice , thanks @nots1dd

nots1dd:
Hi @fsimonis, can you review the changes? pranjal and I seem to be in agreement that this change adds more context than the previous version that was merged

PR #2513, opened 2026-03-10
Allow convergence measures using written but non-exchanged data

### LINKED ISSUE #2442: Allow convergence measures using written but non-exchanged data

In #1678, we list the option to provide custom data for the sole purpose of steering the convergence measure.

For a serial implicit scheme, this data could be created directly in the second participant.
As this participant computes the convergence measure, it doesn't need to be exchanged.

I think we currently require data to be exchanged in order to use it in convergence measures, meaning that the currently proposed workaround doesn't work.

We should create a test for this and use it as a base to implement this feature.

We could allow written data on provided meshes by the second participant even if it is not exchanged.

none

**Additional context**

ankitv06:
## Main changes of this PR
Relax the restriction that the convergence measure data must be exchanged. Data written by the convergence-checking participant (on a mesh it provides) can now be used in convergence measures without exchange.

- Add `checkConvergenceMeasureDataAvailable()` - a validation function that allows convergence measure data to either be exchanged (existing behavior) or written locally by the convergence-checking participant on a provided mesh (new behavior), implying the locally written data doesn't need to be exchanged.
- Update `addConvergenceMeasures()` to use the new check instead of `checkIfDataIsExchanged()`. Narrow `checkIfDataIsExchanged()` error message to only mention acceleration data.
- Add Integration test for a serial-implicit scheme where `Data3` is used in a convergence measure but is only written locally by the second participant — never exchanged.
- Registered the new test in `tests.cmake`.

Closes #2442

The code change applies to all implicit coupling schemes (serial, parallel, multi) since they all share `addConvergenceMeasures()`. The test although covers only the serial implicit case. I am willing to add tests for parallel implicit / multi coupling if desired.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/src/cplscheme/config/CouplingSchemeConfiguration.cpp
+++ b/src/cplscheme/config/CouplingSchemeConfiguration.cpp
@@ -13,6 +13,7 @@
 #include "cplscheme/BaseCouplingScheme.hpp"
 #include "cplscheme/BiCouplingScheme.hpp"
 #include "cplscheme/CompositionalCouplingScheme.hpp"
+#include "cplscheme/CouplingData.hpp"
 #include "cplscheme/CouplingScheme.hpp"
 #include "cplscheme/MultiCouplingScheme.hpp"
 #include "cplscheme/ParallelCouplingScheme.hpp"
@@ -1198,6 +1199,33 @@ void CouplingSchemeConfiguration::addConvergenceMeasures(
   for (auto &elem : convergenceMeasureDefinitions) {
     _meshConfig->addNeededMesh(participant, elem.meshName);
     checkConvergenceMeasureDataAvailable(elem.data->getID(), participant, elem.data->getName(), elem.meshName);
+
+    // checkConvergenceMeasureDataAvailable verified that the participant writes data.
+    // If data is not exchanged, add it to the scheme's _allData (memory bank for convergence iterations).
+    const DataID dataID = elem.data->getID();
+
+    // Check if data is exchanged (existing behavior)
+    const auto match = std::find_if(_config.exchanges.begin(),
+                                    _config.exchanges.end(),
+                                    [dataID, &participant](const Config::Exchange &exchange) {
+                                      if (exchange.from != participant && exchange.to != participant) {
+                                        return false;
+                                      }
+                                      return exchange.data->getID() == dataID;
+                                    });
+
+    if (match == _config.exchanges.end()) {
+      // Data is not exchanged.
+      if (scheme->localParticipant() != participant) {
+        // If current participant is not the one computing this convergence measure, skip it safely.
+        continue;
+      }
+
+      // Add it to the scheme's _allData.
+      const mesh::PtrMesh &mesh = _meshConfig->getMesh(elem.meshName);
+      scheme->addCouplingData(elem.data, mesh, false, false, CouplingData::Direction::Send);
+    }
+
     scheme->addConvergenceMeasure(elem.data->getID(), elem.suffices, elem.strict, elem.measure);
   }
 }
--- a/src/cplscheme/BaseCouplingScheme.hpp
+++ b/src/cplscheme/BaseCouplingScheme.hpp
@@ -231,6 +231,24 @@ public:
    */
   bool hasConverged() const override;
 
+  /// @copydoc cplscheme::CouplingScheme::localParticipant()
+  std::string localParticipant() const final override;
+
+  /**
+   * @brief Adds CouplingData with given properties to this BaseCouplingScheme and returns a pointer to the CouplingData
+   *
+   * If CouplingData with ID of provided data already exists in coupling scheme, no duplicate is created but a pointer to the already existing CouplingData is returned.
+   *
+   * @param data data the CouplingData is associated with
+   * @param mesh mesh the CouplingData is associated with
+   * @param requiresInitialization true, if CouplingData requires initialization
+   * @param exchangeSubsteps true, if CouplingData exchanges all substeps in send/recv
+   * @param direction is the coupling data send or received?
+   *
+   * @return PtrCouplingData pointer to CouplingData owned by the CouplingScheme
+   */
+  PtrCouplingData addCouplingData(const mesh::PtrData &data, mesh::PtrMesh mesh, bool requiresInitialization, bool exchangeSubsteps, CouplingData::Direction direction);
+
 protected:
   /// All send and receive data as a map "data ID -> data"
   DataMap _allData;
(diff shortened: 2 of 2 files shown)


fsimonis:
Hi, please install and run the pre-commit hooks.

tests/serial/convergence-measures/testConvergenceMeasures5.xml
```diff
@@ -0,0 +1,55 @@
...
+
+  <coupling-scheme:serial-implicit>
+    <participants first="SolverOne" second="SolverTwo" />
+    <max-time-windows value="2" />
+    <time-window-size value="1.0" />
+    <exchange data="Data1" mesh="MeshOne" from="SolverOne" to="SolverTwo" />
+    <exchange data="Data2" mesh="MeshOne" from="SolverTwo" to="SolverOne" />
+    <max-iterations value="100" />
+    <!-- Data3 is written by SolverTwo on its provided MeshTwo but NOT exchanged.
+         This tests issue #2442: convergence measures using non-exchanged data. -->
```
fsimonis: Could you please move this information to the cpp file?

tests/serial/convergence-measures/testConvergenceMeasures5.cpp
```diff
@@ -0,0 +1,64 @@
...
+    if (interface.requiresWritingCheckpoint()) {
+      numberOfIterations = 0;
+    }
+
+    if (context.isNamed("SolverTwo")) {
+      // Write Data3 (non-exchanged, used for convergence measure)
+      interface.writeData(meshName, "Data3", {&vertexID, 1}, {&writeValuesData3.at(numberOfAdvanceCalls), 1});
+    }
+
+    interface.advance(1.0);
```
fsimonis: The error-proof way is to use `getMaxTimeStepSize()` here.

tests/serial/convergence-measures/testConvergenceMeasures5.cpp
```diff
@@ -0,0 +1,64 @@
...
+BOOST_AUTO_TEST_SUITE(ConvergenceMeasures)
+PRECICE_TEST_SETUP("SolverOne"_on(1_rank), "SolverTwo"_on(1_rank))
+BOOST_AUTO_TEST_CASE(testConvergenceMeasures5)
+{
+  PRECICE_TEST();
+  using Eigen::Vector2d;
+
+  std::string meshName = context.isNamed("SolverOne") ? "MeshOne" : "MeshTwo";
+
+  precice::Participant interface(context.name, context.config(), 0, 1);
```
fsimonis: This is an old name we don't use anymore, you can use p, part, or participant instead.

```suggestion
  precice::Participant p(context.name, context.config(), 0, 1);
```

tests/serial/convergence-measures/testConvergenceMeasures5.cpp
```diff
@@ -0,0 +1,64 @@
...
+
+  // Data3 converges on iteration 2 for time window 1, and iteration 3 for time window 2.
+  // These values are written by SolverTwo on its provided MeshTwo.
+  // Data3 is NOT exchanged, but used in the convergence measure.
+  std::vector<double> writeValuesData3 = {1.0, 1.01, 2.0, 2.5, 2.51};
+
+  // Expected iterations per time window based on the convergence of Data3
+  std::vector<int> expectedIterations = {2, 3};
+
+  while (interface.isCouplingOngoing()) {
```
fsimonis: Ensure that this loop cannot exceed the hardcoded amount of values.

Use `BOOST_REQUIRE` to stop the test before you see out of bound exceptions.

tests/serial/convergence-measures/testConvergenceMeasures5.xml
ankitv06: sure. moved comment to testConvergenceMeasures5.cpp

tests/serial/convergence-measures/testConvergenceMeasures5.cpp
ankitv06: changed `p.advance(p.getMaxTimeStepSize());` instead of hardcoded advance

tests/serial/convergence-measures/testConvergenceMeasures5.cpp
ankitv06: updated old `interface` var name to new `p` across the file.

tests/serial/convergence-measures/testConvergenceMeasures5.cpp
ankitv06: used `BOOST_REQUIRE` before accessing array values to avoid out-of-bound exceptions

fsimonis:
@ankitv06 Tests are failing. Please run `ctest -j` locally to validate your changes.

fsimonis:
If this can be fixed within the next two days, then I am happy to include it into the upcoming feature release.

ankitv06:
@fsimonis Hi,
upon running ctest -j locally, I found that `testConvergenceMeasures5` test was throwing errors. 
`addConvergenceMeasure` requires Data3 to be added into the memory bank (`_allData`) in order to be used for iterative convergence (prev val - cur val).
**Fix:** Inside `addConvergenceMeasures()` if the data is not being exchanged and the current participant is the one computing the convergence, insert the non-exchanged data in memory bank `_allData`.
Also made some changes in `BaseCouplingScheme.hpp` to move some methods from protected to public to allow using them in the `CouplingSchemeConfiguration` file.

Passed all tests successfully upon ctest -j.
Screenshot 2026-03-18 003250

src/cplscheme/config/CouplingSchemeConfiguration.cpp
```diff
@@ -1172,7 +1198,34 @@ void CouplingSchemeConfiguration::addConvergenceMeasures(
...
+    if (match == _config.exchanges.end()) {
+      // Data is not exchanged.
+      if (scheme->localParticipant() != participant) {
+        // If current participant is not the one computing this convergence measure, skip it safely.
+        continue;
+      }
+
+      // Add it to the scheme's _allData.
+      const mesh::PtrMesh &mesh = _meshConfig->getMesh(elem.meshName);
+      scheme->addCouplingData(elem.data, mesh, false, false, CouplingData::Direction::Send);
```
fsimonis, line 1226: This is a lie 😉

This data is neither send nor received.

fsimonis (changes requested):
If you add the data to _allData, then it cannot be `Send` nor `Received` (that's the point to this PR).
So you need to add the state of local data: `Direction::None`.

On top of that, you don't want to accelerate this data as it won't be exchanged and may result in some weirdness, so it should be filtered out in `getAccelerationData()`.

ankitv06:
@fsimonis 
> So you need to add the state of local data: `Direction::None`

Because the non-exchanged data requires a participant to locally write it and `Direction` enum only having two values: Send or Recieve. Send aligned closest to writing data locally, bypassing any crash. So, I set `Direction::Send`.

Now to make `Direction::None`, i will have to add new state for local data. and update `getAccelerationData()` to filter out any data marked as `Direction::None`. right? is this what you meant?

fsimonis:
> Now to make Direction::None, i will have to add new state for local data. 

Correct, there needs to be a third variant.

> and update getAccelerationData() to filter out any data marked as Direction::None. right? is this what you meant?

That's what I meant. I confirmed this briefly with @uekerman and this data should not be part of the acceleration.

I'll remove this from the 3.4 release as this will need more testing, especially wrt multi coupling.

fsimonis:
@ankitv06 After some further discussion, it may be cleaner to handle this non-exchanged data separately.

Meaning:
- Store local convergence data in `_localConvergenceData` instead of `_allData`
- Add local convergence data there if the data is not exchanged (requiresInitialization and exchanges substeps are false, use any direction).
- Handle this local data separately in initialize, reinitialize, storeIteration, moveToNextWindow (may be more)

This way, you don't need to filter `_allData`.

Issue #2199, opened 2025-02-17
Add error code for different preCICE errors


fsimonis:
**Please describe the problem you are trying to solve.**

The `precice::Error` is raised when preCICE encounters an error.
Currently, the only way to distinguish errors is to check the error message.
It would be helpful to be able to identify an error using a fixed error code.

**Describe the solution you propose.**

Add an error code member to the `precice::Error` similar to [`std::system_error`](https://en.cppreference.com/w/cpp/error/system_error).

Then create a file with all error codes inside preCICE.
Maybe directly in the `precice/error.hpp` header.

**Describe alternatives you've considered**

Stick to using the message only, possibly by prefixing the message with a number.
This could be useful to enumerate warnings and info logs too.

**Additional context**

Extension of #280

BenjaminRodenberg:
Do we also want to be able to test for warnings? This just came up while discussing https://github.com/precice/precice/issues/531 with @Fujikawas. (maybe testing for warnings is also a bad idea in the first place?)

uekerman:
> We could add another parameter for the error code

makes sense to me.

> Do we also want to be able to test for warnings?

I think this would indeed be helpful.

ankitv06:
Hi @fsimonis @uekerman,
I've been looking into it.

1. What should the error code categories look like? This is what I'm thinking - broad and module level.
Each directory represents a logical subsystem, so most of them could have unique error codes, like xml directory has config parsing errors, but the precice/ could have wrong mesh name, bad state, etc. 
``` cpp
enum class ErrorCode : int {
  Unknown = 0,
  Configuration,    // XML parsing, invalid config values -> xml/ (directory)
  Communication,    // MPI, socket, m2n connection errors -> com/, m2n/
  Mapping,          // RBF divergence, mapping setup failures -> mapping/
  CouplingScheme,   // Time stepping, convergence issues -> cplscheme/
  Mesh,             // Mesh definition, vertex/edge errors -> mesh/
  InvalidArgument,  // Bad API arguments (wrong mesh name, etc.) -> precice/
  InvalidState,     // Calling API functions in wrong order -> precice/
  IO,               // File read/write errors -> io/
  // ...
};
```
or should they be more fine-grained per error case (e.g., `MappingDivergence`, `MappingInvalidBasisFunction`, ...)

2. When a C++ preCICE function throws `precice::Error`, each binding (Fortran, Python, cpp) handles it differently, so it will need binding updates as well, but i guess that should be a follow up issue.

[...]

PR #2517, opened 2026-03-11
Prevent Participant usage after non recoverable exception

### LINKED ISSUE #2142: Participant is usable after a non-recoverable exception.
**Describe your setup**

Operating system (e.g. Linux distribution and version): _Linux_
preCICE Version: dev after 3.1.2 0ba3f3a95

**Describe the problem**

When a Participant encounters a non-recoverable exception, it can still be used without problems or even any warnings.

Let's say that the communication buildup fails in `initialize()`, then an exception is thrown which needs to be handled by the user.
If the user chooses to ignore it, then other functions like `advance()` or `readData()` can still be called even though they are pretty much undefined at that point.

**Expected behaviour**

It should directly throw an exception:

`Error: Participant "Fluid" triggered an unrecoverable exception and entered an erroneous state.`

**Additional context**

We could catch exceptions in critical functions and set an `bool erroneousState` flag.

ankitv06:
## Main changes of this PR
This PR adds `_hasError` flag to `ParticipantImpl` to prevent participant usage after a non recoverable exception (e.g., a communication setup failure in `initialize()`).
- Wrapped `initialize()` and `advance()` in try-catch block - whenever exception occurs, it gets caught and `_hasError = true` and the exception is re-thrown
- Added a `PRECICE_CHECK(!_hasError, ...)` guard at the top of every public API method (except `finalize()`) - if the flag is set, the method throws a clear error message.
- instead of throwing exception, `finalize()` quits gracefully by releasing ownership, closing connections and finalizing MPI without attempting operations that depend on internal state (e.g., coupling scheme).

Closes #2142 

As mentioned in the issue, Further refinements can be done like partitioning errors into critical vs non-critical when error categories are established. (Related to #2199 )

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/tests/serial/lifecycle/ErrorInAdvance.cpp
+++ b/tests/serial/lifecycle/ErrorInAdvance.cpp
@@ -14,36 +14,36 @@ PRECICE_TEST_SETUP("SolverOne"_on(1_rank), "SolverTwo"_on(1_rank))
 BOOST_AUTO_TEST_CASE(ErrorInAdvance)
 {
   PRECICE_TEST();
-  precice::Participant interface(context.name, context.config(), context.rank, context.size);
+  precice::Participant p(context.name, context.config(), context.rank, context.size);
 
   if (context.isNamed("SolverOne")) {
     auto   meshName = "MeshOne";
     double coords[] = {0.1, 1.2, 2.3};
-    auto   vertexid = interface.setMeshVertex(meshName, coords);
+    auto   vertexid = p.setMeshVertex(meshName, coords);
 
     auto   dataName = "DataOne";
     double data[]   = {3.4, 4.5, 5.6};
-    interface.writeData(meshName, dataName, {&vertexid, 1}, data);
+    p.writeData(meshName, dataName, {&vertexid, 1}, data);
   } else {
     auto   meshName = "MeshTwo";
     double coords[] = {0.12, 1.21, 2.2};
-    auto   vertexid = interface.setMeshVertex(meshName, coords);
+    auto   vertexid = p.setMeshVertex(meshName, coords);
 
     auto   dataName = "DataTwo";
     double data[]   = {7.8};
-    interface.writeData(meshName, dataName, {&vertexid, 1}, data);
+    p.writeData(meshName, dataName, {&vertexid, 1}, data);
   }
 
-  interface.initialize();
-  BOOST_TEST(interface.isCouplingOngoing());
+  p.initialize();
+  BOOST_TEST(p.isCouplingOngoing());
 
   // Call advance() with a negative time step size to trigger an error
-  BOOST_CHECK_THROW(interface.advance(-1.0), ::precice::Error);
+  BOOST_CHECK_THROW(p.advance(-1.0), ::precice::Error);
 
   // After the error, further API calls should be blocked
-  BOOST_CHECK_THROW(interface.isCouplingOngoing(), ::precice::Error);
+  BOOST_CHECK_THROW(p.isCouplingOngoing(), ::precice::Error);
 
-  interface.finalize();
+  p.finalize();
 }
 
 BOOST_AUTO_TEST_SUITE_END() // Lifecycle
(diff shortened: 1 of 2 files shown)


src/precice/impl/ParticipantImpl.cpp
```diff
@@ -589,6 +605,34 @@ void ParticipantImpl::finalize()
...
+    device::Device::finalize();
+#endif
+    profiling::EventRegistry::instance().finalize();
+
+    // Clear events and finalize MPI
+    utils::Parallel::finalizeOrCleanupMPI();
+    _state = State::Finalized;
+    return;
+  }
+
```
fsimonis: This should be handled fine below.

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -260,43 +265,48 @@ void ParticipantImpl::initialize()
   // Enforce that all user-created events are stopped to prevent incorrect nesting.
   PRECICE_CHECK(_userEvents.empty(), "There are unstopped user defined events. Please stop them using stopLastProfilingSection() before calling initialize().");
 
-  _solverInitEvent.reset();
-  Event e("initialize", profiling::Fundamental, profiling::Synchronize);
+  try {
```
fsimonis: Try using a function try catch block to keep this simple.

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -415,54 +426,59 @@ void ParticipantImpl::advance(
   PRECICE_CHECK(!math::equals(computedTimeStepSize, 0.0), "advance() cannot be called with a time step size of 0.");
   PRECICE_CHECK(computedTimeStepSize > 0.0, "advance() cannot be called with a negative time step size {}.", computedTimeStepSize);
 
-  _numberAdvanceCalls++;
+  try {
```
fsimonis: Try using a function try catch block to keep this simple.

src/precice/impl/ParticipantImpl.cpp
```diff
@@ -639,13 +683,15 @@ void ParticipantImpl::finalize()
 int ParticipantImpl::getMeshDimensions(std::string_view meshName) const
 {
   PRECICE_TRACE(meshName);
+  PRECICE_CHECK(!_hasError, "getMeshDimensions() cannot be called after an unrecoverable error. Participant \"{}\" is in an erroneous state.", _accessorName);
```
fsimonis: The name of the function is already included in the error log.

Please make the error message uniform.
The move it to a separate macro in along the other validation macros named `PRECICE_VALIDATE_ERROR_STATE()`. Assume the `_hasError` name.

src/precice/impl/ParticipantImpl.cpp
ankitv06: sure. modified internal try-catch block to function try-catch for initialize()

src/precice/impl/ParticipantImpl.cpp
ankitv06: sure. modified internal try-catch block to function try-catch for advance()

src/precice/impl/ParticipantImpl.cpp
ankitv06: 1. Created a validation macro `PRECICE_VALIDATE_ERROR_STATE()` in `precice\impl\ValidationMacros.hpp` to keep the error messages uniform.
2. Removed name of function from error check as it is already included in the error log.

src/precice/impl/ParticipantImpl.cpp
ankitv06: Replaced this code with just a `PRECICE_WARN_IF()` to announce erroneous state, as exit is already handled fine in the lower segment of `finalize()`

ankitv06:
Created 4 new files `ErrorInInitialize.cpp, .xml, ErrorInAdvance.cpp, .xml` as integration tests for `advance()` and `initialize()` to verify that API calls are non-usable after an exception occurence in either of these files.
1. `ErrorInInitialize` skips initialization of data in order to invoke exception
2. `ErrorInAdvance` uses negative time step to invoke exception
Local build tests for these tests were successful.

PR #2533, opened 2026-03-24
Compute and report solver imbalance at runtime 

valentinaschueller:
## Main changes of this PR

This PR introduces a `SolverImbalance` object in the participant.
It compares how long each participant takes to complete a time window, excluding coupling overhead.
At the end of the time window, a solver imbalance is computed.
This is exposed through the participant API and can be accessed at runtime.

Simulations coupled through preCICE can be subject to load imbalances inside each, but also between the participants.
Participants in preCICE live in different MPI communicators, so this can be hard to catch with standard tooling.
From preCICE, this information is easy to obtain: 
We compare how long the solvers take to complete time steps (`solverAdvanceTime` is the corresponding profiling event).
Two kinds of information are exposed:
1. How large is the imbalance of all participants? (Same value for all participants)
2. How much faster than the slowest participant is this participant? (Different value for the different participants)

## Author's checklist

## Reviewers' checklist


fsimonis:
@valentinaschueller Could you please open an issue for this feature so we can discuss architectural aspects independent of code?
This would also be a good place to show the user-perspective of the feature.

Issue #2361, opened 2025-09-16
pre-commit not executed in docs PRs


MakisH:
#2359 made a trivial change to the `docs/README.md` (which, as trivial, was done from the GitHub web interface), passing all CI tests. After merging, the pre-commit failed due to a trailing space in the respective Markdown file (which the GitHub web interface seems to always add).

Looking at the [workflow](https://github.com/precice/precice/blob/develop/.github/workflows/build-and-test.yml), changes in `docs/` are excluded from triggers to the workflow in a pull request, which makes sense.

We should also run the pre-commit in pull requests, independently of where the changes are.

Since the motivation here is to reduce the number of workflows shown in a PR, I wonder if having the pre-commit as a separate workflow actually makes sense. @fsimonis opinion?

MakisH:
~~Same on the website:~~

~~- Workflow: https://github.com/precice/precice.github.io/blob/master/.github/workflows/build.yml~~
~~- Recent example: https://github.com/precice/precice.github.io/pull/597 (in this case, it seems to have failed in the PR. Why did auto-merge work, then?)~~

Edit: This seems to have been just an unlucky coincidence, as the workflow file was updated to exclude the respective directory when I updated the branch. The issue does not apply on the website CI workflows.

PR #2541, opened 2026-03-30
Add DuMux adapter ref to system-tests.yaml

MakisH:
## Main changes of this PR

Modifies the [system-tests.yaml workflow](https://github.com/precice/precice/blob/develop/.github/workflows/system-tests.yaml) to get the latest state of the DuMux adapter `develop` branch before triggering the system tests.

@Fujikawas @IshaanDesai As far as I understand, you wanted the Micro-Manager version to be fixed. There is no `MICRO_MANAGER_REF` variable to set.

~~Depends on https://github.com/precice/tutorials/pull/758, see that for previous work.~~ Actually, it does not, it only passes additional arguments.

Required to actually get the latest version of the DuMux adapter in the system tests. I know that we need to offload this step to the `tutorials` repository, but this is the state we have right now.

## Author's checklist

* I added a changelog file with `make changelog` if there are user-observable changes since the last release. -> I don't think one is needed.
* [x] I ran the systemtests by adding the label `trigger-system-tests` (may be skipped if minor change) -> see [this run](https://github.com/precice/precice/actions/runs/23812435427?pr=2541)

### DIFF
--- a/.github/workflows/system-tests.yaml
+++ b/.github/workflows/system-tests.yaml
@@ -12,6 +12,7 @@ jobs:
     outputs:
       ref-python-bindings: ${{ steps.ref-python-bindings.outputs.shorthash }}
       ref-calculix-adapter: ${{ steps.ref-calculix-adapter.outputs.shorthash }}
+      ref-dumux-adapter: ${{ steps.ref-dumux-adapter.outputs.shorthash }}
       ref-fenics-adapter: ${{ steps.ref-fenics-adapter.outputs.shorthash }}
       ref-openfoam-adapter: ${{ steps.ref-openfoam-adapter.outputs.shorthash }}
       ref-su2-adapter: ${{ steps.ref-su2-adapter.outputs.shorthash }}
@@ -30,6 +31,13 @@ jobs:
         with:
           owner: precice
           repo: calculix-adapter
+          branch: develop
+      - id: ref-dumux-adapter
+        name: Get DuMux adapter ref
+        uses: nmbgeek/github-action-get-latest-commit@main
+        with:
+          owner: precice
+          repo: dumux-adapter
           branch: develop
       - id: ref-fenics-adapter
         name: Get FEniCS adapter ref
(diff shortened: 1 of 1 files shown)


IshaanDesai (commented):
I am a bit confused about why the system tests workflow is located in the `precice/precice` repository. Is this related to which tests run when a preCICE release candidate is created?

MakisH:
> I am a bit confused about why the system tests workflow is located in the `precice/precice` repository. Is this related to which tests run when a preCICE release candidate is created?

This workflow is for allowing PRs to trigger the system tests (reusable workflow in the tutorials repository). We have such a workflow in several repositories, mainly changing which arguments we pass.

Overview of this workflow:

```yaml
on:
  pull_request:
    types: [labeled]

jobs:
  gather-refs:
     ...

  run-system-tests:
    name: Trigger system tests
    if: ${{ github.event.label.name == 'trigger-system-tests' }}
    needs: gather-refs
    uses: precice/tutorials/.github/workflows/run_testsuite_workflow.yml@develop
    with:
      suites: release_test
      build_args: "PLATFORM:ubuntu_2404,\
      ...
```

IshaanDesai:
> This workflow is for allowing PRs to trigger the system tests (reusable workflow in the tutorials repository). We have such a workflow in several repositories, mainly changing which arguments we pass.

Okay, got it. So could I add such a workflow in the Micro Manager repo to trigger the system test when certain things are done (for example, a release candidate)?

MakisH:
> > This workflow is for allowing PRs to trigger the system tests (reusable workflow in the tutorials repository). We have such a workflow in several repositories, mainly changing which arguments we pass.
> 
> Okay, got it. So could I add such a workflow in the Micro Manager repo to trigger the system test when certain things are done (for example, a release candidate)?

Yes, exactly. I prepared PRs for the Micro-Manager and the DuMux adapter:

- https://github.com/precice/micro-manager/pull/237 (depends on https://github.com/precice/tutorials/pull/758)
- https://github.com/precice/dumux-adapter/pull/80

MakisH:
This is actually independent of the tutorials PR. The system tests pass.

Fujikawas:
As I understand here,  the PRs from this repo can trigger system tests including the suite `dumux_test`, but the `REF` of MM and `dumux-adapter` were the default in `latest-components` so far. With this `PR` the latest commit in `develop` will be used as `REF`. Then these changes make sense for me.

Fujikawas (approved):

IshaanDesai (approved):

MakisH:
As an another instance of https://github.com/precice/precice/issues/2361, the pre-commit check also did not run on this PR, but ran on the main branch. If I recall correctly, shellcheck was previously not failing for the same syntax.

I will work on that today and fix the CI.

PR #2549, opened 2026-04-13
Fix Ginkgo includes and improve HIP/CUDA dependency in RBF TU

davidscn:
## Main changes of this PR

Fixes the erroneous includes pointing to Ginkgo and provides a portable way for compiling with CUDA/HIP.

While the previous PRs fixed the CMake handling #2451 correctly, the problem is that I had Ginkgo installed on all my systems. Thus headers were still discoverable in my include paths, so no error was triggered. Reproduced it with spack. 

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/docs/changelog/2549.md
+++ b/docs/changelog/2549.md
@@ -1,0 +1,1 @@
+- Fixed include bug where compiling with Kokkos-Kernels depends on Ginkgo.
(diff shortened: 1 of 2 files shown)


fsimonis:
> Thus headers were still discoverable in my include paths, so no error was triggered.

I guess this is also the reason why the CI worked.

I'll mark it for a hotfix.

davidscn:
Tested it on CUDA, SYCL and OpenMP. HIP is currently not available, but it is also a rather obvious bug, once you see it.

PR #2550, opened 2026-04-13
Enforce preCICE compatibility with `CMAKE_CXX_STANDARD`

davidscn:
## Main changes of this PR

Enforce preCICE compatibility for [`CMAKE_CXX_STANDARD`](https://cmake.org/cmake/help/latest/variable/CMAKE_CXX_STANDARD.html).

The standard way to define a CXX standard in CMake is `CMAKE_CXX_STANDARD`.  At the moment, this option is silently ignored, and the preCICE variable is well hidden in the CMakeCache. This makes it hard for users who compile preCICE with other tools like CUDA, HIP, etc, which have requirements in this regard. 

The PR also prints the defined standard to the user as a status message, as this is incredibly helpful for debugging and in terms of transparency.

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -79,7 +79,13 @@ option(PRECICE_RELEASE_WITH_DEBUG_LOG "Enable debug logging in release builds" O
 option(PRECICE_RELEASE_WITH_TRACE_LOG "Enable trace logging in release builds" OFF)
 option(PRECICE_RELEASE_WITH_ASSERTIONS "Enable assertions in release builds" OFF)
 
-set(PRECICE_CXX_STANDARD "17" CACHE STRING "CXX Standard for all preCICE targets" )
+if(NOT CMAKE_CXX_STANDARD)
+  set(PRECICE_CXX_STANDARD "17" CACHE STRING "CXX Standard for all preCICE targets" )
+else()
+  set(PRECICE_CXX_STANDARD ${CMAKE_CXX_STANDARD})
+endif()
+message(STATUS "Setting preCICE CXX standard to ${PRECICE_CXX_STANDARD}")
+
 option(PRECICE_HIDE_SYMBOLS "Hide non-API symbols in preCICE library" ON)
 mark_as_advanced(PRECICE_HIDE_SYMBOLS PRECICE_CXX_STANDARD)
 


fsimonis:
> The standard way to define a CXX standard in CMake is CMAKE_CXX_STANDARD. At the moment, this option is silently ignored, and the preCICE variable is well hidden in the CMakeCache.

My comment to this is that CMake has a complicated relationship with defining standards.

The point of `PRECICE_CXX_STANDARD` is to easily control the standard used across preCICE. If you have the special requirement of overriding it, then you can do it as such:

```
cmake -DPRECICE_CXX_STANDARD=20 .
```

If you really want to respect the `CMAKE_CXX_STANDARD`, then we need to implement our own way of computing the max of `CMAKE_CXX_STANDARD` and the default in preCICE. This however opens the door to forward-compatibility issues as we have already seen with C++20.

```cmake
if(DEFINED CMAKE_CXX_STANDARD and (CMAKE_CXX_STANDARD GREATER PRECICE_CXX_STANDARD))
  set(_precice_cxx_std ${CMAKE_CXX_STANDARD})
  message(INFO "Using newer C++ default standard ${_precice_cxx_std}")
else()
  set(_precice_cxx_std ${PRECICE_CXX_STANDARD})
endif()
# Then use _precice_cxx_std
```

For debugging properties, you can use: https://cmake.org/cmake/help/latest/module/CMakePrintHelpers.html

```
cmake_print_properties(TARGETS preciceCore PROPERTIES CXX_STANDARD)
```

Sadly, there is no property equivalent to [`variable_watch`](https://cmake.org/cmake/help/latest/command/variable_watch.html).

davidscn:
> If you really want to respect the CMAKE_CXX_STANDARD, then we need to implement our own way of computing the max of CMAKE_CXX_STANDARD and the default in preCICE. This however opens the door to forward-compatibility issues as we have already seen with C++20.

But why making it so complicated? The way I see it: the `PRECICE_CXX_STANDARD` is our internal convenience variable to handle things. It is not documented anywhere and well hidden inside CMake. So, a user may not use it in the first place.

The `CMAKE_CXX_STANDRD` is then simply the user interface to control our internal implementation. The point here is that people who know CMake can use it without further ado, i.e., no reading of documentation (which we dont have) or other package-specific handling (because it follows the standard).

fsimonis:
> But why making it so complicated?

If we rely on CMAKE_CXX_STANDARD, then we rely on the user to respect a minimum. And if we require a minimum, then we impose a minimum on packages that use preCICE as subproject. Furthermore, not defining this default, will pass no flag, falling back to the compiler default. This will lead to confusing errors for both developers and users. 

>The CMAKE_CXX_STANDRD is then simply the user interface to control our internal implementation.

The user has no say in what standard preCICE uses, which is why the current variable is marked as advanced.

The fact that we need to adjust the standard for Kokkos or Ginkgo is something we should handle internally.

davidscn:
> If we rely on CMAKE_CXX_STANDARD, then we rely on the user to respect a minimum. And if we require a minimum, then we impose a minimum on packages that use preCICE as subproject. Furthermore, not defining this default, will pass no flag, falling back to the compiler default. This will lead to confusing errors for both developers and users.

Not in the current PR implementation. 

> The user has no say in what standard preCICE uses, which is why the current variable is marked as advanced

Well, this is your interpretation and the point of the entire discussion: users should be able to configure the standard, particularly useful for CUDA etc.

> The fact that we need to adjust the standard for Kokkos or Ginkgo is something we should handle internally.

Trying to cover every edge case is close to impossible. I don't see a reason not to pick up a defined CMAKE_CXX_STANDRD.

fsimonis:
> Well, this is your interpretation and the point of the entire discussion: users should be able to configure the standard, particularly useful for CUDA etc.

Users can override the standard right now using `PRECICE_CXX_STANDARD`.

This is a PR. Such a change is a discussion that belongs into an issue.

davidscn:
> Users can override the standard right now using PRECICE_CXX_STANDARD.

The point of this PR is to make the behavior standard-compliant with the CMake.

https://github.com/precice/precice/pull/2550#issue-4254757251

fsimonis:
> The point of this PR is to make the behavior standard-compliant with the CMake.

I did some more digging.
We _could_ use `target_compile_features(T PRIVATE cxx_std_20)` to impose a minimum standard to the preCICE targets. Then we could remove the `CXX_STANDARD` property from the main `CMakeLists.txt`. This allows CMake to pick a higher standard if `CMAKE_CXX_STANDARD` is set to a higher one.

https://cmake.org/cmake/help/latest/prop_gbl/CMAKE_CXX_KNOWN_FEATURES.html#high-level-meta-features-indicating-c-standard-support

This is also what Kokkos uses in its generated CMake config. It sets `INTERFACE_COMPILE_FEATURES "cxx_std_20"`, so that preCICE, which links against that target, requires C++20. Meaning, this is already correctly handled.

We could port the standard definition from `CXX_STANDARD` to the compile feature and let CMake handle this. However, I am still opposed to letting the user freely upgrade the standard. This further increases our testing matrix.

davidscn:
> I am still opposed to letting the user freely upgrade the standard. This further increases our testing matrix.

How is this related to our testing matrix? As you noted already before, users _can_ already put a different standard for which we do not test either.  The purpose of this change is pure standard compliance and thereby convenience in the _how_.

fsimonis:
> How is this related to our testing matrix? As you noted already before, users can already put a different standard for which we do not test either.

There is a reason that this setting is internal/marked as advanced. If the user (or developer) has a technical reason to override it, then they can do this comfortably right now. But without any guarantees.

If we make the version respect the user choice (or the choice of the super-project), then we should actively test it too. At this point, it becomes just another flexible dependency.

> The purpose of this change is pure standard compliance and thereby convenience in the how.

It is not a "standard" to comply with; it sets the default for a property. This is only relevant for projects that don't pin their standard requirement.
We, however, can, and IMO should pin the version to avoid surprises.

[Example](https://github.com/precice/precice/compare/develop...fsimonis:precice:cmake-min-standard) how this could look like.

davidscn:
>  There is a reason that this setting is internal/marked as advanced. If the user (or developer) has a technical reason to override it, then they can do this comfortably right now. But without any guarantees.

> If we make the version respect the user choice (or the choice of the super-project), then we should actively test it too. At this point, it becomes just another flexible dependency.

So you say: "users can do this comfortably right now" and "If we make the version respect the user choice...we should actively test it".
This doesn't add up: so right now, changing it is possible without guarantees, but user choices need to be tested, if we changing is possible?

> It is not a "standard" to comply with; it sets the default for a property

The standard way to define a CXX standard in CMake is `CMAKE_CXX_STANDARD`. I am not aware of any larger scientific software project that does not follow this, except for preCICE.

> This is only relevant for projects that don't pin their standard requirement.
We, however, can, and IMO should pin the version to avoid surprises.

So right now, the "pinning" can be changed by a project-specific variable. Using "CMAKE_CXX_STANDARD" for that exact same purpose is more standard. I dont understand where we pin more or less when changing this behavior. Of course I am open in the way we implement it.

MakisH:
I see the unexpected behavior, and I see that the user should not arbitrarily change the standard. Since this is mainly for special situations, maybe something in this direction could help:

Why not throw a warning whenever `CMAKE_CXX_STANDARD` is set (but still ignore it), pointing to the project-specific variable? Permutations of this would be throwing an error or accepting but throwing a warning.

For an admin/user who is likely in a difficult situation trying to make a complex stack of dependencies with conflicting requirements work, this would at least guide them through the options but also let them take the responsibility if they really want to override it. Silently ignoring it sounds indeed confusing, so some kind of user information would be helpful in that case.

davidscn:
Some final thoughts on this:

> I see that the user should not arbitrarily change the standard

- with GPUs and device compilers, compiling becomes more fragile, so while this sounds reasonable for our standard use cases, it becomes more valuable in such scenarios.

> pointing to the project-specific variable

- other projects had such project-specific variables, for example Kokkos had "Kokkos_CXX_STANDARD", which now gives

> [[DEPRECATED - USE CMAKE_CXX_STANDARD INSTEAD]] The C++ standard for Kokkos to use...

also deal.II once had such a variable. It seems that all larger projects first have to go through this "project-specific solution" before reverting later on. But ok, I understand that both of you are not in favor of making CMake accept the CMAKE_CXX_STANDARD variable.

MakisH:
> I understand that both of you are not in favor of making CMake accept the CMAKE_CXX_STANDARD variable

From my side, I am in favor of reading it, and I would be happy with accepting it with a warning.

In any case, I have no strong opinion on this; I just think we should still communicate what is happening when this is set.

Looking into other projects, linking to their respective discussions, would be very interesting. But that's indeed more complicated.

fsimonis:
> So you say: "users can do this comfortably right now" and "If we make the version respect the user choice...we should actively test it".
This doesn't add up: so right now, changing it is possible without guarantees, but user choices need to be tested, if we changing is possible?

It is simply possible. The user needs to explicitly tell preCICE to use a different standard.
Using `CMAKE_CXX_STANDARD` makes this a side effect that the user may not be aware of, and hence needs to be tested.

It is common to set the standard at the top of a CMake project so that all executables follow the same C++ standard. If such a user uses preCICE as subproject, then this C++ change is propagated. This will lead to confusion at some point.

> So right now, the "pinning" can be changed by a project-specific variable. Using "CMAKE_CXX_STANDARD" for that exact same purpose is more standard.

Again. CMAKE_CXX_STANDARD is global and propagates to subprojects. The point of pinning is to prevent it to change despite global setting.

> It seems that all larger projects first have to go through this "project-specific solution" before reverting later on. But ok, I understand that both of you are not in favor of making CMake accept the CMAKE_CXX_STANDARD variable.

The world of scientific software is not a world of shining standards. Look at versioning for example.
The reason for adopting this could just be a subgroup of projects deciding to do it this way. This is, for example, what xSDK was all about, especially before they adopted spack as common ground.

Instead of blindly following others, I would be highly interested in the reasoning behind this decision.

> Why not throw a warning whenever CMAKE_CXX_STANDARD is set (but still ignore it), pointing to the project-specific variable?

This would interfere with the super-projects. I would avoid anything that requires workarounds for silencing warnings.

> For an admin/user who is likely in a difficult situation trying to make a complex stack of dependencies with conflicting requirements work, this would at least guide them through the options but also let them take the responsibility if they really want to override it.

That's what the pinning is for. This takes one more decision away from the admin/user.

PR #2552, opened 2026-04-14
Minor portability updates on Kokkos implementation

davidscn:
## Main changes of this PR

Minor portability updates on Kokkos implementation.

No bug, but sticking closer to the Kokkos syntax, see also https://kokkos.org/kokkos-core-wiki/API/core/stl-compat/pair.html

Using `std::pair` is explicitly declared in the docs as valid argument and use case. Still, I guess this may prevent future issues.

## Author's checklist

## Reviewers' checklist


davidscn:
I guess a changelog entry is not strictly necessary. Merging here, such that I can test all updates together.

PR #2557, opened 2026-04-27
Add a pre-commit workflow for all files

### LINKED ISSUE #2361: pre-commit not executed in docs PRs

MakisH:
## Main changes of this PR

Adds a pre-commit workflow that is independent of the `build-and-test.yml`.

The latter is only invoked for a subset of files, to save resources on the slow steps of building and testing preCICE. The `pre-commit` step only takes a few seconds to run.

While we could further expand the filters in the `build-and-test.yml`, I think this is the simplest solution that ensures that pre-commit is always being executed, without any significant disadvantage.

Closes #2361

The pre-commit hook that was triggered after merging https://github.com/precice/precice/pull/2541 was added in https://github.com/precice/precice/pull/2373, and I had not updated the pre-commit hooks on my system meanwhile. With the workflow suggested here, which also runs on every pull request, the CI would have stopped me.

## Author's checklist


.github/workflows/pre-commit.yml
```diff
@@ -0,0 +1,26 @@
...
+  pull_request:
+  workflow_dispatch:
+
+concurrency:
+  group: ${ {github.event_name }}-${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: ${{github.event_name == 'pull_request'}}
+
+jobs:
+  pre-commit:
+    runs-on: ubuntu-latest
```
fsimonis: Let's try the ubuntu-slim, which should start much faster.

.github/workflows/system-tests.yaml
fsimonis, line 1: Unrelated changes?

fsimonis:
This is a good step, which has been long overdue. It should also allow us to check the format of changelog files in the future.

fsimonis (approved):

.github/workflows/system-tests.yaml
MakisH: These are intended, as they fix an issue that should have been caught by pre-commit. I can split the PR. -> Moved to https://github.com/precice/precice/pull/2558

.github/workflows/pre-commit.yml
MakisH: I had not noticed the new `ubuntu-slim`, thanks! ([blog post](https://github.blog/changelog/2025-10-28-1-vcpu-linux-runner-now-available-in-github-actions-in-public-preview/)) Looks like this runner is made for this kind of small jobs, and runs in a container instead of a VM.

In the first run, the job took a long longer (almost 5min). Second run was almost 2min. Third was 4min. A second attempt of the [same job](https://github.com/precice/precice/actions/runs/25099885365) dropped to 37s.

Previous runs with `ubuntu-latest` took <20s ([example](https://github.com/precice/precice/actions/runs/24994382611)).

I also had to upgrade actionlint to recognize the `ubuntu-slim` runner.

I am not sure if the `ubuntu-slim` helps in this case, but, in any case, the total times are small.

fsimonis: Hmm, the single core also hurts the performance.

Let's use it here to gather some experience with this slim runner.

PR #2565, opened 2026-05-07
Experimental: Migrate to C++23 format

fsimonis:
## Main changes of this PR

This experimental PR attempts to migrate preCICE to C++23 format.
This PR primarily exists to see how the support on all distros looks like.

## Author's checklist


fsimonis:
The C++23 support on Ubuntu 24.04s stdlibc++ 13.2 is pretty flaky.

Missing:
- print header
- range formatting (affects STL container)

fsimonis:
It is also weird that this somehow prevents CMake from using unity builds, which makes the build time explode to 34 minutes.

This is too flaky. I will close the PR but keep the branch as reference.

PR #2567, opened 2026-05-18
Preallocate buffers rbf solver

aadityajoshi01:
## Author's checklist

## Reviewers' checklist


davidscn (changes requested):
In case this PR originates from the TODOs: The point here was to reduce the memory footprint as such, e.g., using in-place operations or reusing already-existing memory buffers from the caller side. The suggested solution is not well-suited, because it permanently blocks memory instead of re-allocating memory if needed. Having the member variables `mutable` is not a good idea here either, as the declared member variables belong to the class.

fsimonis:
Closing due to a lack of activity and the proposed solution being unsuitable.

PR #2578, opened 2026-06-18
Add workflow notifications

fsimonis:
## Main changes of this PR

This PR adds workflow notifications for workflows:

- Docker nightly release (only on failure): scheduled, manual, via release
- Coverity Scan: scheduled
- Performance Regression Test: scheduled, manual, labelled

Have a central place to see what is going on

## Author's checklist

### DIFF
--- a/.github/workflows/coverity-scan.yml
+++ b/.github/workflows/coverity-scan.yml
@@ -49,3 +49,10 @@ jobs:
         env:
           TOKEN: ${{ secrets.COVERITY_SCAN_TOKEN }}
           EMAIL: ${{ secrets.COVERITY_SUBMITTER_EMAIL }}
+      - name: Notification
+        if: always()
+        uses: precice/notify-action@v0
+        with:
+          homeserver: ${{ secrets.MATRIX_BOT_HOMESERVER }}
+          access_token: ${{ secrets.MATRIX_BOT_TOKEN }}
+          room_id: ${{ secrets.MATRIX_BOT_ROOMID }}
--- a/.github/workflows/coverity-scan.yml
+++ b/.github/workflows/coverity-scan.yml
@@ -53,6 +53,6 @@ jobs:
         if: always()
         uses: precice/notify-action@v0
         with:
-          homeserver: ${{ secrets.MATRIX_BOT_HOMESERVER }}
+          homeserver: ${{ vars.MATRIX_BOT_HOMESERVER }}
           access_token: ${{ secrets.MATRIX_BOT_TOKEN }}
-          room_id: ${{ secrets.MATRIX_BOT_ROOMID }}
+          room_id: ${{ vars.MATRIX_BOT_ROOMID }}
--- a/.github/workflows/performance-regression.yml
+++ b/.github/workflows/performance-regression.yml
@@ -107,9 +107,9 @@ jobs:
             build/target.json
             comparison.txt
       - name: Notification
-        if: if: ${{ always() && github.event_name == 'schedule' }}
+        if: ${{ always() && github.event_name == 'schedule' }}
         uses: precice/notify-action@v0
         with:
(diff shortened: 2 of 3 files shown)


.github/workflows/coverity-scan.yml
```diff
@@ -49,3 +49,10 @@ jobs:
         env:
           TOKEN: ${{ secrets.COVERITY_SCAN_TOKEN }}
           EMAIL: ${{ secrets.COVERITY_SUBMITTER_EMAIL }}
+      - name: Notification
+        if: always()
+        uses: precice/notify-action@v0
+        with:
+          homeserver: ${{ secrets.MATRIX_BOT_HOMESERVER }}
+          access_token: ${{ secrets.MATRIX_BOT_TOKEN }}
+          room_id: ${{ secrets.MATRIX_BOT_ROOMID }}
```
MakisH: Any particular reason to set the `homeserver` and `room_id` as secrets and not variables? I cannot just see their values in the settings for the review.

.github/workflows/coverity-scan.yml
```diff
@@ -49,3 +49,10 @@ jobs:
         env:
           TOKEN: ${{ secrets.COVERITY_SCAN_TOKEN }}
           EMAIL: ${{ secrets.COVERITY_SUBMITTER_EMAIL }}
+      - name: Notification
+        if: always()
+        uses: precice/notify-action@v0
```
MakisH, line 54: Since this action is new and does not have any documentation yet: how does the output look like?

.github/workflows/performance-regression.yml
```diff
@@ -106,3 +106,10 @@ jobs:
             build/base.json
             build/target.json
             comparison.txt
+      - name: Notification
+        if: always()
```
MakisH: This workflow is executed also manually and on PRs. I think that restricting such reports to the scheduled runs would make them less frequent and thus more useful.

Something like this could work:


```suggestion
        if: ${{ github.event_name == 'schedule' }}
```

MakisH (commented):
I like the idea, and all relevant workflows in this repository are covered. I am not yet sure how this will look like in practice.

.github/workflows/coverity-scan.yml
fsimonis: Mostly to have them in one place. The token needs to be an org secret, so I just added the others there too.

.github/workflows/performance-regression.yml
fsimonis: True. Does this run the step also in case of failure though?

.github/workflows/coverity-scan.yml
MakisH: In that case, some comment/documentation would help, in case we want to change these in the future. At the moment, I cannot just see the current values and update them, I have to guess how they looked like when the secret was created, to get the equivalent ones for the new room.

.github/workflows/performance-regression.yml
MakisH: Why not? The `always()` was not distinguishing between failure or success. I would not expect the schedule to also not distinguish them.

.github/workflows/coverity-scan.yml
fsimonis: I added some docs to the readme https://github.com/precice/notify-action

.github/workflows/coverity-scan.yml
fsimonis: Ok. Makes sense to change them to envs.

.github/workflows/performance-regression.yml
fsimonis: Turns out that we need this:

```suggestion
        if: if: ${{ always() && github.event_name == 'schedule' }}
```

Meaning: run even if failing but only on schedule

.github/workflows/coverity-scan.yml
fsimonis: Done

PR #2586, opened 2026-07-29
Fix ValueParser for empty strings and malformed fractions (#2584)

### LINKED ISSUE #2584: xml::ValueParser returns an uninitialized double for empty/whitespace-only attribute values instead of raising a config error
**Describe your setup**

Operating system (e.g. Linux distribution and version): _Windows 11_ — the issue is platform-independent, it follows from standard `std::istream` semantics
preCICE Version: 3.4.1 (`develop`, commit `124ee9bf`)

No adapter or bindings involved — this is in the XML configuration parser itself.

**Describe the problem**

`parseDouble()` in [`src/xml/ValueParser.cpp`](https://github.com/precice/precice/blob/develop/src/xml/ValueParser.cpp#L14-L27) uses `!iss.eof()` as its "did the value parse completely?" check:

```cpp
double parseDouble(const std::string &rawValue)
{
  std::istringstream iss{rawValue};
  try {
    iss.imbue(std::locale(PARSING_LOCALE));
  } catch (...) {
  }
  double value;              // <-- never initialized
  iss >> value;
  if (!iss.eof()) {          // <-- does not detect failure
    throw std::runtime_error{"Could not fully parse value \"" + rawValue + "\" as a double."};
  }
  return value;
}
```

For an empty or whitespace-only `rawValue` this check does not fire, for two compounding reasons:

1. The `sentry` for `operator>>` skips leading whitespace, hits end-of-input, and sets **both** `failbit` and `eofbit`. Since `eofbit` is set, `!iss.eof()` is `false` and **no exception is thrown**.
2. Because the sentry evaluated to `false`, `operator>>` returns *without touching* `value` ([istream.formatted.reqmts]). The uninitialized local `double value` is therefore returned — **undefined behaviour**, and in practice whatever was left on the stack.

The guard only catches the case where extraction stops early with input remaining (e.g. `"abc"`, `"1.0foo"`). It cannot catch outright extraction failure.

This is reachable straight from a user's XML config. `XMLAttribute<T>::readValue()` passes the raw attribute string through with no emptiness check ([`src/xml/XMLAttribute.hpp:172`](https://github.com/precice/precice/blob/develop/src/xml/XMLAttribute.hpp#L172)):

```cpp
try {
  readValueSpecific(position->second, _value);
} catch (const std::exception &e) {
  PRECICE_ERROR(e.what());
}
```

and `ConfigParser` stores attribute values verbatim (`attributesMap[attributeName] = decodeXML(value);`, [`src/xml/ConfigParser.cpp:67`](https://github.com/precice/precice/blob/develop/src/xml/ConfigParser.cpp#L67)), so `value=""` arrives as an empty string.

The `a/b` fraction syntax in `readValueSpecific(const std::string &, double &)` widens the exposure, because it feeds `parseDouble()` substrings that can be empty by construction — `value="1/"`, `value="/2"` and `value="/"` all split into an empty operand and are accepted.

Two related, lower-severity variants of the same root cause:

- `readValueSpecific(const std::string &, int &)` ([line 42](https://github.com/precice/precice/blob/develop/src/xml/ValueParser.cpp#L42-L53)) has the identical `!iss.eof()` guard. It is not UB — `_value` is a member and thus already initialized — but an empty attribute is silently accepted and the attribute keeps its prior value rather than producing a config error.
- `readValueSpecific(const std::string &, Eigen::VectorXd &)` ([line 55](https://github.com/precice/precice/blob/develop/src/xml/ValueParser.cpp#L55-L67)) calls `parseDouble()` per `;`-separated component, so any empty component inherits the same behaviour.

**Step To Reproduce**

I do not have a preCICE build on this machine, so I verified the parser functions directly rather than end-to-end. The two functions below are copied **verbatim** from `src/xml/ValueParser.cpp` (only the anonymous namespace was dropped). The `-12345.0` sentinel shows whether the parser wrote anything at all, and `dirtyStack()` makes an uninitialized read visible instead of incidentally zero.

1. Save as `repro.cpp`:

```cpp
#include <iostream>
#include <locale>
#include <sstream>
#include <stdexcept>
#include <string>

constexpr static const char *PARSING_LOCALE = "en_US.UTF-8";

// verbatim from src/xml/ValueParser.cpp
double parseDouble(const std::string &rawValue)
{
  std::istringstream iss{rawValue};
  try {
    iss.imbue(std::locale(PARSING_LOCALE));
  } catch (...) {
  }
  double value;
  iss >> value;
  if (!iss.eof()) {
    throw std::runtime_error{"Could not fully parse value \"" + rawValue + "\" as a double."};
  }
  return value;
}

// verbatim from src/xml/ValueParser.cpp
void readValueSpecific(const std::string &rawValue, double &value)
{
  if (rawValue.find('/') != std::string::npos) {
    std::string left  = rawValue.substr(0, rawValue.find('/'));
    std::string right = rawValue.substr(rawValue.find('/') + 1, rawValue.size() - rawValue.find('/') - 1);
    value = parseDouble(left) / parseDouble(right);
  } else {
    value = parseDouble(rawValue);
  }
}

__attribute__((noinline)) void dirtyStack()
{
  volatile unsigned char buf[4096];
  for (int i = 0; i < 4096; ++i) buf[i] = 0xAB;
}

void tryDouble(const char *raw)
{
  dirtyStack();
  double v = -12345.0; // sentinel
  try {
    readValueSpecific(std::string(raw), v);
    std::cout << "  raw=\"" << raw << "\"  -> NO THROW, value = " << v << "\n";
  } catch (const std::exception &e) {
    std::cout << "  raw=\"" << raw << "\"  -> threw: " << e.what() << "\n";
  }
}

int main()
{
  std::istringstream iss{""};
  double d = -12345.0;
  iss >> d;
  std::cout << "istringstream(\"\") >> double : fail=" << iss.fail()
            << " eof=" << iss.eof() << " value=" << d << "\n\n";

  tryDouble("1.0");   // control
  tryDouble("1/2");   // control
  tryDouble("");
  tryDouble("   ");
  tryDouble("1/");
  tryDouble("/2");
  tryDouble("/");
  tryDouble("abc");   // control
}
```

2. Build and run:

```
g++ -std=c++17 -O2 -Wall -o repro repro.cpp && ./repro
```

Output (g++ 6.3.0):

```
istringstream("") >> double : fail=1 eof=1 value=-12345

  raw="1.0"  -> NO THROW, value = 1
  raw="1/2"  -> NO THROW, value = 0.5
  raw=""     -> NO THROW, value = -2.53017e-098
  raw="   "  -> NO THROW, value = -2.53017e-098
  raw="1/"   -> NO THROW, value = 1
  raw="/2"   -> NO THROW, value = -1.26509e-098
  raw="/"    -> NO THROW, value = 1
  raw="abc"  -> threw: Could not fully parse value "abc" as a double.
```

Two things to note:

- The first line confirms the root cause: on empty input the stream sets **both** `fail` and `eof`, and leaves the target variable untouched (the `-12345` sentinel survives).
- `"1/"` and `"/"` yielding exactly `1` is the signature of a stale stack slot being reused from the preceding `parseDouble()` call — i.e. the returned value depends on unrelated prior execution, not on the input.

Note that `-Wall` does **not** warn here, because `value` is passed by reference to `operator>>` and the compiler cannot see that the sentry path leaves it unwritten.

At the config level the same inputs are reachable via any `XMLAttribute<double>` — for example `<max-time value=""/>` or `<max-time value="1/"/>` inside a coupling scheme (`attrValueMaxTime` is an `XMLAttribute<double>`, [`src/cplscheme/config/CouplingSchemeConfiguration.cpp:488`](https://github.com/precice/precice/blob/develop/src/cplscheme/config/CouplingSchemeConfiguration.cpp#L488)). If a maintainer with a build handy can confirm end-to-end, that would be worth adding here.

**Expected behaviour**

An empty, whitespace-only, or malformed-fraction attribute value should produce the usual `PRECICE_ERROR` config error naming the tag and attribute — the same way `value="abc"` already does. It should never return an indeterminate value, and a config typo should not depend on stack contents.

**Additional context**

The check needs to test `fail()`, not just `eof()`, and the local should be initialized. Something along these lines:

```cpp
double parseDouble(const std::string &rawValue)
{
  std::istringstream iss{rawValue};
  try {
    iss.imbue(std::locale(PARSING_LOCALE));
  } catch (...) {
  }
  double value = 0.0;
  iss >> value;
  if (iss.fail() || !iss.eof()) {
    throw std::runtime_error{"Could not fully parse value \"" + rawValue + "\" as a double."};
  }
  return value;
}
```

The same `iss.fail() || !iss.eof()` change applies to the `int` overload. The fraction path in `readValueSpecific(const std::string &, double &)` would then reject `"1/"`, `"/2"` and `"/"` automatically, since each produces an empty operand. It may still be worth rejecting a zero denominator explicitly, since `value="1/0"` currently parses to `inf` without complaint.

I could not find an existing issue covering this. There is currently no test in `src/xml/tests/` exercising empty or malformed numeric attribute values; happy to open a PR with the fix plus regression tests for `""`, `"   "`, `"1/"`, `"/2"` and `"/"` if that is useful.

anushkagupta200615-jpg:
## Main changes of this PR

- Refactored `parseDouble()` and the inline int-parsing block in `src/xml/ValueParser.cpp` into a single templated `parseArithmetic<T>()` helper
- Changed the failure check from `!iss.eof()` to `iss.fail() || !iss.eof()` — this catches empty/whitespace-only strings where the stream sentry sets both `failbit` and `eofbit` but leaves the target variable untouched
- Value-initialized the result variable (`T value{}`) to eliminate undefined behaviour when extraction fails
- Added an explicit `denominator == 0.0` guard in the `a/b` fraction path so `"1/0"` and `"0/0"` throw instead of returning `inf`/`nan`
- Added `src/xml/tests/ValueParserTest.cpp` with regression tests covering all affected cases

Closes #2584

`xml::ValueParser` silently returned an uninitialized/garbage `double` when an XML attribute value was empty (`value=""`), whitespace-only (`value=" "`), or a malformed fraction (`value="1/"`, `value="/2"`, `value="/"`). The root cause was that `operator>>` sets both `failbit` and `eofbit` on empty input, so the old `!iss.eof()` guard evaluated to `false` and no exception was thrown — undefined behaviour per [istream.formatted.reqmts].

---

## Author's checklist

* [ ] I used the `pre-commit` hook to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [ ] For breaking changes: I documented the changes in the appropriate porting guide.

### DIFF
--- a/src/tests.cmake
+++ b/src/tests.cmake
@@ -114,5 +114,6 @@ target_sources(testprecice
     src/utils/tests/StringTest.cpp
     src/xml/tests/ParserTest.cpp
     src/xml/tests/PrinterTest.cpp
+    src/xml/tests/ValueParserTest.cpp
     src/xml/tests/XMLTest.cpp
     )
--- a/src/xml/ValueParser.cpp
+++ b/src/xml/ValueParser.cpp
@@ -1,5 +1,4 @@
 #include "xml/ValueParser.hpp"
-#include <boost/algorithm/string/constants.hpp>
 #include <boost/algorithm/string/split.hpp>
 #include <ostream>
 #include <sstream>
@@ -68,12 +67,14 @@ void readValueSpecific(const std::string &rawValue, int &value)
 void readValueSpecific(const std::string &rawValue, Eigen::VectorXd &value)
 {
   std::vector<std::string> components;
+  // Do NOT use token_compress_on: it would silently drop an empty-string input,
+  // producing zero components and bypassing the error check in parseDouble.
   boost::split(
-      components, rawValue, [](char c) { return c == ';'; }, boost::algorithm::token_compress_on);
-  const int size = components.size();
+      components, rawValue, [](char c) { return c == ';'; });
+  const Eigen::Index size = static_cast<Eigen::Index>(components.size());
 
   Eigen::VectorXd vec(size);
-  for (int i = 0; i != size; ++i) {
+  for (Eigen::Index i = 0; i != size; ++i) {
     vec(i) = parseDouble(components[i]);
   }
   value = vec;
--- a/src/xml/ValueParser.cpp
+++ b/src/xml/ValueParser.cpp
@@ -11,45 +11,58 @@ namespace precice::xml {
 namespace {
 constexpr static const char *PARSING_LOCALE = "en_US.UTF-8";
 
-double parseDouble(const std::string &rawValue)
+/// Extracts a single arithmetic value, requiring the entire string to be consumed.
+template <typename T>
+T parseArithmetic(const std::string &rawValue, const char *typeName)
 {
   std::istringstream iss{rawValue};
   try {
     iss.imbue(std::locale(PARSING_LOCALE));
   } catch (...) {
   }
-  double value;
+  T value{};
   iss >> value;
-  if (!iss.eof()) {
-    throw std::runtime_error{"Could not fully parse value \"" + rawValue + "\" as a double."};
+  // An empty or whitespace-only string makes the sentry fail, which sets eofbit alongside
+  // failbit and leaves value untouched. Checking eof() alone therefore misses these.
+  if (iss.fail() || !iss.eof()) {
+    throw std::runtime_error{"Could not fully parse value \"" + rawValue + "\" as " + typeName + "."};
   }
   return value;
+}
+
+double parseDouble(const std::string &rawValue)
+{
+  return parseArithmetic<double>(rawValue, "a double");
 }
 } // namespace
 
 void readValueSpecific(const std::string &rawValue, double &value)
 {
-  if (rawValue.find('/') != std::string::npos) {
-    std::string left  = rawValue.substr(0, rawValue.find('/'));
-    std::string right = rawValue.substr(rawValue.find('/') + 1, rawValue.size() - rawValue.find('/') - 1);
-
-    value = parseDouble(left) / parseDouble(right);
-  } else {
+  const auto pos = rawValue.find('/');
+  if (pos == std::string::npos) {
     value = parseDouble(rawValue);
+    return;
   }
+
+  double numerator   = 0.0;
+  double denominator = 0.0;
+  try {
+    numerator   = parseDouble(rawValue.substr(0, pos));
+    denominator = parseDouble(rawValue.substr(pos + 1));
+  } catch (const std::runtime_error &) {
+    // Report the value the user actually wrote, not the offending operand.
+    throw std::runtime_error{"Could not fully parse value \"" + rawValue + "\" as a fraction of two doubles."};
+  }
+
+  if (denominator == 0.0) {
+    throw std::runtime_error{"Could not parse value \"" + rawValue + "\" as a double: the denominator is zero."};
+  }
+  value = numerator / denominator;
 }
 
 void readValueSpecific(const std::string &rawValue, int &value)
 {
-  std::istringstream iss{rawValue};
-  try {
-    iss.imbue(std::locale(PARSING_LOCALE));
-  } catch (...) {
-  }
-  iss >> value;
-  if (!iss.eof()) {
-    throw std::runtime_error{"Could not fully parse value \"" + rawValue + "\" as an int."};
-  }
+  value = parseArithmetic<int>(rawValue, "an int");
 }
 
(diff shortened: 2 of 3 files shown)


fsimonis:
Thanks for the PR!
Sorry but it will take some time until we get around to review your PR.

PR #2593, opened 2026-08-20
fix: scale scalar time tolerances by operand magnitude (fixes #2592)

### LINKED ISSUE #2592: Implicit coupling aborts at t = 64 s with "contains no samples" when the solver reports its own dt
## Describe your setup

- **Operating system:** Ubuntu 24.04.4 LTS
- **preCICE Version:** 3.3.0 (Ubuntu package `libprecice3`). Source inspected on `develop` at `v3.3.1-69-g8b4546cc`; the relevant code is unchanged there.
- **Compiler / MPI:** GCC 13, Open MPI 4.1.6
- Both participants single-rank, C++ API.

## Describe the problem

An implicit coupling aborts deterministically once the coupling time reaches **t ≈ 64 s = 2^6**:

```
ERROR: Data "Data-Two" cannot be read from mesh "SolverOne-Mesh" as it contains
no samples. This is typically a configuration issue of the data flow. Check if
the data is correctly exchanged to this participant "SolverOne" and mapped to
mesh "SolverOne-Mesh".
```

The partner then dies on the broken socket. In our production runs (a different
code path — the read happens through a mapping) the same condition surfaces as
`Data q on mesh hegel_wall didn't contain any data samples while attempting to
map to mesh Solid-Mesh`.

## Step to reproduce

**Correction to my original report:** I previously wrote that "a trivial
dummy-solver pair suffices". That is **wrong**, and I apologise if it cost
anyone time. The stock `examples/solverdummies` does **not** reproduce this: it
sets `dt = getMaxTimeStepSize()` and passes that same value back to `advance()`.

The failure needs the solver to keep **its own clock** and report the
difference — what our solver, the OpenFOAM adapter, and generally any code with
its own time integration does:

```cpp
const double tOld = t;
t += participant.getMaxTimeStepSize();   // solver integrates its own clock
participant.advance(t - tOld);           // ... and reports the difference
```

Reproducer: the stock C++ solverdummy with exactly that three-line change
(attached as `precice-2592-reproducer.zip` on the comment below). Config:
`time-window-size 5e-3`,
`min/max-iterations 2`, `acceleration:constant 0.5`, scalar data,
`nearest-neighbor` mapping on the received mesh. Our production cases that hit
this are **parallel**-implicit with `time-window-size 5e-3`; rows B and G below
show the coupling scheme makes no difference.

### Measured results

| # | solver `dt` | window | max-time | library | outcome |
|---|-------------|--------|----------|---------|---------|
| A | `getMaxTimeStepSize()` echoed | 5e-3 | 65 s | 3.3.0 stock | **clean**, ends t = 65.000000000013372, 26000 advances |
| B | own clock, `t - t_old` | 5e-3 | 65 s | 3.3.0 stock | **ABORT at t = 64.010000000014273**, advance 25604 (window 12802) |
| C | own clock, `t - t_old` | 5e-3 | **33 s** | 3.3.0 stock | **clean**, ends t = 32.99999999999843, 13200 advances |
| D | own clock, `t - t_old` | 5e-3 | 65 s | 3.3.1 + patch below | **clean**, ends t = 65.000000000013372, 26000 advances |
| E | own clock, `t - t_old` | 2.5e-3 | 65 s | 3.3.0 stock | **ABORT at t = 64.0075** |
| F | own clock, `t - t_old` | **1e-2** | 65 s | 3.3.0 stock | **clean** — see caveat below |
| G | own clock, `t - t_old` | 5e-3 | 65 s | 3.3.0 stock | **ABORT at t = 64.010000000014273** — *parallel*-implicit, identical to B |
| H | own clock, `t - t_old` | **1e-2** | 65 s | 3.3.0 stock | **clean** — *parallel*-implicit, identical to F |

- **B vs C** is the discriminating pair: identical solver, identical `dt`,
  identical window size; only the absolute time reached differs. Below the 2^6
  crossover it survives, above it it dies.
- **B vs A** isolates the trigger to the solver reporting its own `dt`.
- **B vs E**: halving the window doubles the window count but keeps the abort at
  the same *simulation time* — so it is not a window-count or buffer limit.
- **B vs G and F vs H**: serial- and parallel-implicit behave identically, so
  the coupling scheme is not part of the trigger.
- **F/H are an honest caveat:** with `time-window-size 1e-2` the same solver runs
  clean past 64 s under both schemes. So the failure depends on the *specific*
  `dt` as well as on the magnitude of `t` — it is not "always at 64 s for every
  configuration".

## Root cause (confirmed by instrumenting preCICE)

I built `develop` (`8b4546cc`) with `fprintf` tracing in `Waveform::setSampleAtTime`,
`Waveform::trimBefore` and `ParticipantImpl::handleDataAfterAdvance`
(the patch is in the attached zip). **preCICE carries two different clocks, and they drift apart
by exactly one ULP:**

- `_couplingScheme->getTimeWindowStart()` — accumulated **exactly**, Kahan-compensated
  (`TimeHandler` uses `utils::DoubleAggregator`, i.e. boost `sum_kahan`, adding
  `timeWindowSize` per window).
- **sample timestamps** — written at `timeSteppedTo`, which follows the
  *participant-reported* `dt`.

Traced at the failing window (`time-window-size 5e-3`, solver reports its own `dt`):

```
[INSTR] setSampleAtTime(64.009999999999991) wf=0x...cc8 nBefore=1
[INSTR] handleDataAfterAdvance: windowStart=64.010000000000005
                               timeSteppedTo=64.009999999999991
[INSTR] trimBefore(64.010000000000005) wf=0x...cc8 nBefore=1 ulp(time)=1.421085e-14
[INSTR]    stamp=64.009999999999991 diff=-1.421085e-14 (-1.000 ulp) smaller=1  <== TRIMMED
[INSTR] trimBefore(64.010000000000005) wf=0x...cc8 nAfter=0   *** WAVEFORM NOW EMPTY ***
```

The sample sits **exactly one ULP** below `windowStart`. At this magnitude
one ULP is `1.421085e-14`, which is larger than `NUMERICAL_ZERO_DIFFERENCE = 1e-14`,
so `math::smaller(stamp, windowStart)` returns **true** and
`Waveform::trimBefore` deletes the only sample. The next `readData` then aborts.

The control run, same solver and same window size, traced at t ≈ 32 s:

```
[INSTR] trimBefore(31.990000000000002) wf=0x...768 nBefore=1 ulp(time)=3.552714e-15
[INSTR]    stamp=31.989999999999998 diff=-3.552714e-15 (-1.000 ulp) smaller=0
```

**Identical** one-ULP gap — but at 32 s one ULP is `3.55e-15`, *below* the fixed
`1e-14`, so the sample is kept and the run proceeds. The bug is therefore not
"time drifts"; the gap is a constant one ULP. What changes at 2^6 is only
whether the fixed absolute tolerance is still bigger than that ULP:

| time range      | ULP(t)       | vs. 1e-14 | one-ULP-low sample |
|-----------------|--------------|-----------|--------------------|
| [16, 32) s      | 3.55e-15     | ok        | kept               |
| [32, 64) s      | 7.11e-15     | ok        | kept               |
| **[64, 128) s** | **1.42e-14** | exceeds   | **trimmed**        |
| [128, 256) s    | 2.84e-14     | exceeds   | trimmed            |

This also explains row F: whether the participant-reported `dt` produces a
one-ULP-low stamp (rather than an exact hit) depends on the specific `dt`, so
some window sizes never open the gap at all.

## Relevant code

Sample timestamps are **absolute** simulation times, and
`Waveform::trimBefore` (`src/time/Waveform.cpp:139`) drops samples using a
**fixed absolute** tolerance:

```cpp
void Waveform::trimBefore(double time)
{
  auto beforeTime = [time](const auto &s) { return math::smaller(s.timestamp, time); };
  ...
}
```

```cpp
constexpr double NUMERICAL_ZERO_DIFFERENCE = 1.0e-14;   // differences.hpp:8
... smaller(Scalar A, Scalar B, Scalar tolerance = NUMERICAL_ZERO_DIFFERENCE)
{ return A - B < -tolerance; }                          // differences.hpp:85
```

## Expected behaviour

Comparisons of absolute simulation-time stamps stay robust at any run length;
coupling proceeds past 64 s (and past 256 s, below).

### A second, probably related abort at t = 256 s = 2^8

With the first problem patched, our runs then aborted at t ≈ 256 s from
`BaseCouplingScheme::addComputedTime`:

> The time step size given to preCICE in 'advance' 0.005000000000023874 exceeds the maximum allowed time step size 0.005

Here the solver-side `t - t_old` carries a ULP(t) error (2.8e-14 at 256 s) which
trips `math::greaterEquals(getNextTimeStepMaxSize(), timeToAdd)`. The compared
values are the *small* `dt`, so scaling by the operands would not help — the
relevant magnitude is `getTime()`. (This one is arguably a solver-side bug: your
error message's own advice, clamping to `min(solverDt, getMaxTimeStepSize())`,
fixes it.)

### The patch we run in production

Scale the tolerance by operand magnitude in the scalar comparators of
`src/math/differences.hpp`:

```cpp
const Scalar scale = std::max(Scalar(1), std::max(std::abs(A), std::abs(B)));
// e.g. smaller:  return A - B < -tolerance * scale;
```

For `|A|,|B| <= 1` (e.g. O(1) mesh coordinates) behaviour is unchanged. Run D
above is this patch. We have run it on v3.3.1 in production since 2026-07-30;
two independent coupled cases that previously died reproducibly at t = 64.010
now run past it to their 150 s end time with no other changes.

I recognise this is a blunt instrument and may not be the fix you want — a
targeted alternative is to compare waveform timestamps **window-relative**
(`equals(s.timestamp - windowStart, time - windowStart)`), keeping operands
small, which is effectively what `TimeHandler` already does. Happy to prepare
either as a PR, with runs A–F as regression tests.

Lightningbolt935:
## Main changes of this PR

* Fixes the scalar comparators in `math/differences.hpp` to scale the tolerance 
  by `max(1, |a|, |b|)` instead of the fixed `1e-14` constant.
* Values with magnitude ≤ 1 (mesh coordinates, etc.) are completely unaffected.
* Added a clamp in `BaseCouplingScheme::addComputedTime` so that sub-ULP 
  rounding in `timeToAdd` cannot make it exceed `getNextTimeStepMaxSize()`.
* Added `ScalarLargeTimestamps` regression test covering the 64 s and 256 s crash cases.
* Fixes #2592

Reported in #2592 — coupled runs abort at exactly t = 64 s with 
`"didn't contain any data samples"`. I traced this to the scalar comparators 
in `differences.hpp`, which use a fixed absolute tolerance of `1e-14` for all 
values, including large simulation timestamps.

IEEE 754 double spacing (ULP) grows with magnitude. At t ≥ 64 s, one ULP 
already exceeds `1e-14`, so timestamps from repeated `dt` additions can end up 
a single ULP below `windowStart`. `Storage::trimBefore()` then classifies them 
as strictly smaller and drops the sample, causing the abort.

A second crash at 256 s hits `addComputedTime` for the same reason: 
floating-point subtraction error makes `timeToAdd` look like it exceeds the 
window max by a sub-ULP amount.

The reporter in #2592 had already verified a similar patch works. I cleaned it 
up and added a regression test covering both cases.

## Author's checklist

* [ ] I used the `pre-commit` hook and ran `pre-commit run --all`.
* [x] I added a changelog file with `make changelog`.   ← do this after you get the PR#
* [x] For breaking changes: I documented the changes in the porting guide.  ← N/A here
* [ ] I ran the systemtests by adding the label `trigger-system-tests`.

## Reviewers' checklist

### DIFF
--- a/docs/changelog/2593.md
+++ b/docs/changelog/2593.md
@@ -1,0 +1,1 @@
+* Fixed scalar time comparators in `math/differences.hpp` to use magnitude-scaled tolerance, preventing coupling aborts at large simulation times (t ≥ 64 s and t ≥ 256 s).
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -463,15 +463,19 @@ bool BaseCouplingScheme::addComputedTime(
                   "Make sure to pass your own desired time-step size or use the recommended limiting \"dt = min(solver_dt, getMaxTimeStepSize())\".");
   }
 
-  // Clamp timeToAdd to guard against sub-ULP floating-point rounding errors
-  // that can make timeToAdd spuriously exceed getNextTimeStepMaxSize() at large
-  // absolute simulation times (t >= 256 s). This is a belt-and-suspenders fix
-  // that works in tandem with the magnitude-scaled tolerances in differences.hpp.
-  const double maxStepSize = getNextTimeStepMaxSize();
-  const double clampedTimeToAdd = std::min(timeToAdd, maxStepSize);
+  // Clamp timeToAdd to guard against sub-ULP floating-point rounding errors.
+  // At large absolute simulation times (t >= 256 s), the subtraction
+  // timeToAdd = time - time_old carries ~ULP(t) error. The step sizes
+  // themselves are small (e.g. 5e-3), so the magnitude-based scaling in
+  // differences.hpp would use scale=1 and not help here. We therefore scale
+  // the tolerance explicitly by the current simulation time, matching the
+  // magnitude of the rounding error, and clamp the value used for progressBy().
+  const double maxStepSize       = getNextTimeStepMaxSize();
+  const double timeTolerance     = math::NUMERICAL_ZERO_DIFFERENCE * std::max(1.0, getTime());
+  const double clampedTimeToAdd  = std::min(timeToAdd, maxStepSize);
 
-  // Check validness
-  bool valid = math::greaterEquals(maxStepSize, timeToAdd);
+  // Check validness using the time-scaled tolerance
+  bool valid = math::greaterEquals(maxStepSize, timeToAdd, timeTolerance);
   PRECICE_CHECK(valid,
                 "The time step size given to preCICE in \"advance\" {} exceeds the maximum allowed time step size {} "
                 "in the remaining of this time window. "
--- a/src/cplscheme/BaseCouplingScheme.cpp
+++ b/src/cplscheme/BaseCouplingScheme.cpp
@@ -463,17 +463,24 @@ bool BaseCouplingScheme::addComputedTime(
                   "Make sure to pass your own desired time-step size or use the recommended limiting \"dt = min(solver_dt, getMaxTimeStepSize())\".");
   }
 
+  // Clamp timeToAdd to guard against sub-ULP floating-point rounding errors
+  // that can make timeToAdd spuriously exceed getNextTimeStepMaxSize() at large
+  // absolute simulation times (t >= 256 s). This is a belt-and-suspenders fix
+  // that works in tandem with the magnitude-scaled tolerances in differences.hpp.
+  const double maxStepSize = getNextTimeStepMaxSize();
+  const double clampedTimeToAdd = std::min(timeToAdd, maxStepSize);
+
   // Check validness
-  bool valid = math::greaterEquals(getNextTimeStepMaxSize(), timeToAdd);
+  bool valid = math::greaterEquals(maxStepSize, timeToAdd);
   PRECICE_CHECK(valid,
                 "The time step size given to preCICE in \"advance\" {} exceeds the maximum allowed time step size {} "
                 "in the remaining of this time window. "
                 "Did you restrict your time step size, \"dt = min(preciceDt, solverDt)\"? "
                 "For more information, consult the adapter example in the preCICE documentation.",
-                timeToAdd, getNextTimeStepMaxSize());
+                timeToAdd, maxStepSize);
 
   // add time interval that has been computed in the solver to get the correct time remainder
-  _time.progressBy(timeToAdd);
+  _time.progressBy(clampedTimeToAdd);
 
   return reachedEndOfTimeWindow();
 }
--- a/src/math/differences.hpp
+++ b/src/math/differences.hpp
@@ -17,10 +17,14 @@ constexpr bool equals(const Eigen::MatrixBase<DerivedA> &A,
 }
 
 /// Compares two scalar (arithmetic) types
+/// The tolerance is scaled by the magnitude of the operands so that comparisons
+/// remain robust for large absolute values (e.g. long simulation times).
+/// For |a|, |b| <= 1 (e.g. mesh coordinates) the behaviour is unchanged.
 template <class Scalar>
 typename std::enable_if<std::is_arithmetic<Scalar>::value, bool>::type equals(const Scalar a, const Scalar b, const Scalar tolerance = NUMERICAL_ZERO_DIFFERENCE)
 {
-  return std::abs(a - b) <= tolerance;
+  const Scalar scale = std::max(Scalar(1), std::max(std::abs(a), std::abs(b)));
+  return std::abs(a - b) <= tolerance * scale;
 }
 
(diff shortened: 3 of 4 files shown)


fsimonis:
Hi, I'm going to close this PR, as the source of the problem is not well understood and changing these parts of the code has far-reaching consequences.
I'll comment in the issue #2592 for more information.

Issue #2408, opened 2025-12-03
Investigate spacial indexing speed


fsimonis:
**Please describe the problem you are trying to solve.**

Projection-based mappings as well as PoU mapping all heavily rely on spacial-index trees of mesh primitives, which we currently use the boost.geometry rtree for.

For sufficiently large meshes, this becomes noticeable. For fully triangulated 1000x1000 meshes, it takes around 8 seconds to compute a NP mapping.
This mapping computation is dominated by the vertex and triangle lookups. This was analyzed during https://github.com/precice/precice/pull/2406

**Describe the solution you propose.**

We should investigate if we can improve query speeds by:
- tuning the existing [rtree parameters](https://www.boost.org/doc/libs/latest/libs/geometry/doc/html/geometry/reference/spatial_indexes/parameters/boost__geometry__index__rstar.html) on a per-primitive basis.
- minimizing indirection by directly using 3D `Vertex::rawCoords()` instead of adapting `mesh::Vertex` and `Eigen::VectorXd` (important for 2D meshes, 1D meshes #1669 can be implemented as an ordered sequence of ids with binary search.)
- using indices that use the dimensionality of the mesh instead of mapping everything to 3D
- customizing distance strategies, box and index types
- exploring other implementations of rtrees or [spacial index data structures](https://en.wikipedia.org/wiki/Spatial_database#Spatial_index)

**Additional context**

Currently, we [use these indices](https://github.com/precice/precice/blob/develop/src/query/Index.hpp) as follows:
- the vertex index is used for box-intersection, 1-nearest-neighbour and k-nearest-neighbour queries
- the edge and triangle indices are used for k-nearest-neighbour lookups where k is generally fixed at compile time
- the tetrahedron index is used for 1-nearest-neighbour lookups only

Our indices are non-mutable. We create them once and use them only for lookups. We never insert or delete individual entries.

This allows us to always use the [packing constructor](https://www.boost.org/doc/libs/latest/libs/geometry/doc/html/geometry/reference/spatial_indexes/boost__geometry__index__rtree/rtree_ran_cr_par_cr_ind_cr_val_cr_all_cr_pac_cr.html) of the boost rtree. This should lead to [best-case build and query times](https://www.boost.org/doc/libs/latest/libs/geometry/doc/html/geometry/spatial_indexes/introduction.html).

We have [some benchmarks](https://github.com/precice/precice/blob/develop/benchmarks/mesh-index.cpp) that could be used to investigate such changes.

[...]

PR #2594, opened 2026-08-21
Query the spatial index with raw coordinates

mohitt31:
## Main changes of this PR

The rtree queries in `query::Index` now take raw coordinates instead of an `Eigen::VectorXd`.

Why the old way costs something:

- boost.geometry reads an `Eigen::VectorXd` through this accessor:

```cpp
static double get(Eigen::VectorXd const &p)
{
  if (Dimension >= static_cast<size_t>(p.rows())) {
    return 0;
  } else {
    return p(Dimension);
  }
}
```

- So every coordinate read does a bounds check and goes through the vector's heap pointer.
- A query repeats that for every primitive it visits, so the cost grows with the size of the search space rather than with the number of queries.
- `eigenToRaw` already exists and pads to three dimensions the same way, so converting the query point once per query gives the same results.

Affected calls: `getClosestVertex`, `getClosestVertices`, `getClosestEdges`, `getClosestTriangles`, `getEnclosingTetrahedra`.

- Found while looking into #2408, which lists "minimizing indirection by directly using 3D `Vertex::rawCoords()` instead of adapting `mesh::Vertex` and `Eigen::VectorXd`" as one direction.
- This only moves the conversion out of the per-primitive adapter and into the query boundary. As @fsimonis points out below, the Index API itself still takes `Eigen::VectorXd`, so the conversions aren't gone, just cheaper. Discussion on that is in the comments.

Measurements, using the existing `benchmarks/mesh-index.cpp`:

- Redone with @fsimonis's settings: AC adapter connected, macOS Low Power Mode off, `--benchmark_repetitions=11 --benchmark_report_aggregates_only=true` on a filtered subset.
- Low Power Mode was the main source of noise in my first attempt, so please ignore the numbers I posted earlier. These are lower.
- Four independent runs. `Index vertices` isn't touched by this PR, so it's in as a control.

| benchmark | median of 4 runs | per-run |
|---|---|---|
| Closest vertex to center/65536 | -7.3% | -6.5, -7.3, -7.3, -8.7 |
| Closest vertex to center/1048576 | -6.6% | -6.0, -7.1, -6.8, -6.3 |
| 10 closest vertices to center/65536 | -5.7% | -6.1, -5.3, -5.1, -6.9 |
| Index vertices/65536 (control) | -0.6% | +3.4, -0.2, -1.1, -1.2 |

Within-run CV is under 1.2% now, against 20% and worse before.

Rebased onto current `develop`, so the recent mesh changes are in both sides.

**Disclosure per CONTRIBUTING.md:** the code and benchmark scripts in this PR were prepared with an AI assistant.

## Author's checklist

On the test checkbox: this doesn't change behaviour, and the existing query and mapping tests already cover the affected calls. Can add something specific if you'd like.

### DIFF
--- a/benchmarks/mapping-nn.cpp
+++ b/benchmarks/mapping-nn.cpp
@@ -1,0 +1,52 @@
+#include <benchmark/benchmark.h>
+
+#include "helper.hpp"
+#include "mapping/Mapping.hpp"
+#include "mapping/NearestNeighborMapping.hpp"
+#include "mesh/Mesh.hpp"
+#include "mesh/SharedPointer.hpp"
+
+using namespace precice;
+
+namespace {
+
+/// Two 3d Halton meshes of n vertices, offset so no vertices coincide
+std::pair<mesh::PtrMesh, mesh::PtrMesh> makeMeshPair(int n)
+{
+  mesh::PtrMesh inMesh(new mesh::Mesh("InMesh", 3, 0));
+  auto          inPos = generate3DHalton(n);
+  for (auto row : inPos.rowwise()) {
+    inMesh->createVertex(row);
+  }
+
+  mesh::PtrMesh outMesh(new mesh::Mesh("OutMesh", 3, 1));
+  auto          outPos = generate3DHalton(n);
+  outPos.array() += 1e-3;
+  for (auto row : outPos.rowwise()) {
+    outMesh->createVertex(row);
+  }
+  return {inMesh, outMesh};
+}
+
+} // namespace
+
+/// Benchmarks computeMapping of a consistent nearest-neighbour mapping between two 3d meshes
+static void computeNearestNeighborMapping(benchmark::State &state)
+{
+  int n          = state.range(0);
+  auto [in, out] = makeMeshPair(n);
+
+  for (auto _ : state) {
+    state.PauseTiming();
+    mapping::NearestNeighborMapping mapping(mapping::Mapping::CONSISTENT, 3);
+    mapping.setMeshes(in, out);
+    in->index().clear();
+    state.ResumeTiming();
+
+    mapping.computeMapping();
+    benchmark::DoNotOptimize(mapping);
+  }
+  state.SetComplexityN(state.range(0));
+}
+
+BENCHMARK(computeNearestNeighborMapping)->Name("NN computeMapping")->RangeMultiplier(4)->Range(1 << 10, 1 << 18)->Complexity();
--- a/benchmarks/sources.cmake
+++ b/benchmarks/sources.cmake
@@ -7,6 +7,7 @@ target_sources(precice-bench
     benchmarks/bb.cpp
     benchmarks/helper.hpp
     benchmarks/main.cpp
+    benchmarks/mapping-nn.cpp
     benchmarks/mesh-index.cpp
     benchmarks/mesh-tagging.cpp
     benchmarks/rbf-assembly-kernels.cpp
--- a/docs/changelog/2594.md
+++ b/docs/changelog/2594.md
@@ -1,0 +1,1 @@
+- Improved the speed of nearest-neighbour and enclosing-tetrahedron index queries by converting the query point to raw coordinates once per query.
--- a/src/mapping/NearestNeighborBaseMapping.cpp
+++ b/src/mapping/NearestNeighborBaseMapping.cpp
@@ -1,6 +1,7 @@
 #include "NearestNeighborBaseMapping.hpp"
 
 #include <boost/container/flat_set.hpp>
+#include <cmath>
 #include <functional>
 #include <iostream>
 #include "logging/LogMacros.hpp"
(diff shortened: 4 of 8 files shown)


fsimonis:
> M4, macOS 26.5,
> A laptop is a noisy machine.

CPU frequency fixing is problematic on macOS. What you can do is to connect the AC adapter, use a high power mode if there is such a thing, and use `--benchmark_repetitions=11 --benchmark_report_aggregates_only=true`. You may want to run only a subset of benchmarks using `--benchmark_filter=`

mohitt31:
Thanks, that is useful. You were right that the machine was the problem, and more so than I realised: I checked and it was running on battery with macOS Low Power Mode enabled, which throttles the CPU. That explains the drift I was trying to work around with the pairing.

I will redo the measurements on AC with Low Power Mode off and your settings (`--benchmark_repetitions=11 --benchmark_report_aggregates_only=true` on a filtered subset) and replace the table with those numbers. I would rather post one clean run than more numbers from a throttled laptop, so it may take me a day or two to get to a machine I can keep on mains.

Given that, the CI workflow from #2465 is still the number I would trust for this change.

mohitt31:
Redone with your settings and the numbers are cleaner, so thanks for pushing back on it.

The main thing was Low Power Mode. It was on, and it was throttling the CPU, which is where most of my noise came from. On AC with it off and `--benchmark_repetitions=11 --benchmark_report_aggregates_only=true` on a subset, within-run CV dropped from over 20% to under 1.2%, so I did not need the A/B pairing trick at all.

Four independent runs, with `Index vertices` as an untouched control:

| benchmark | median of 4 runs | per-run |
|---|---|---|
| Closest vertex to center/65536 | -7.3% | -6.5, -7.3, -7.3, -8.7 |
| Closest vertex to center/1048576 | -6.6% | -6.0, -7.1, -6.8, -6.3 |
| 10 closest vertices to center/65536 | -5.7% | -6.1, -5.3, -5.1, -6.9 |
| Index vertices/65536 (control) | -0.6% | +3.4, -0.2, -1.1, -1.2 |

Worth saying plainly: this is smaller than the 7-8% I claimed before. The effect is about 6-7%, and the earlier figure came from a throttled machine. I have replaced the table in the PR description.

I also rebased onto current `develop` so the mesh changes from last week are included in both sides.

fsimonis:
So far, these are interesting results.

> This PR is only the query-point half of that bullet. I left out the stored-primitive half and explained why in

Not entirely, as the Index API still uses `Eigen::VectorXd`. If the API were using RawCoords instead, then the mapping and the querying could avoid any conversion.
Have a look at `index::findNearestProjection` for 3D and count the conversions between `Eigen::VectorXd` and `RawCoords`.

mohitt31:
Counted it. For the 3D path, per queried vertex:

```
NearestProjectionMapping.cpp:93
  fVertex.getCoords()                      RawCoords -> Eigen::VectorXd   (1)
findNearestProjection(location, n)
  findVertexProjection(location)
    getClosestVertex(location)             Eigen -> RawCoords             (2)
    Polation{location, vertex}             stays Eigen
  findTriangleProjection(location, n, ...)
    getClosestTriangles(location, n)       Eigen -> RawCoords             (3)
    Polation(location, triangle)           stays Eigen
    [fallback] findEdgeProjection(...)
      getClosestEdges(location, n)         Eigen -> RawCoords             (4)
```

So two conversions on the common path, three when it falls back to edges, and all of them convert the same value. Plus the one on the way in.

The one on the way in is the one I had not looked at properly. `Vertex::getCoords()` is

```cpp
Eigen::VectorXd v(_dim);
std::copy_n(_coords.data(), _dim, v.data());
return v;
```

which is a heap allocation per output vertex, purely to hand the index a value it already has as `RawCoords` and that my patch then converts straight back.

So you are right that calling this the query-point half was wrong. What I did was move the conversion out of the per-primitive adapter and into the query boundary, which is why it shows up as a win, but the conversion count is still non-zero and the allocation is untouched. `RawCoords` on the Index API would take all four to zero, and `fVertex.rawCoords()` returns a reference, so the allocation goes too.

Before I write any of it, how would you like it split?

- Extend this PR to change the Index API and update the callers, so it lands as one change.
- Keep this PR as the narrow win and do the API change as a follow-up, given it touches `NearestProjectionMapping`, `LinearCellInterpolationMapping` and `WatchPoint`.

I lean towards the second, mostly because the numbers here are clean and I would rather not make them harder to review. But it is your call, and I am happy either way.

One thing I would want your opinion on for the API change: `Polation` also takes `const Eigen::VectorXd &`, so unless that changes too there is still one conversion left before constructing it. Is `Polation` in scope for you, or would you rather leave that alone for now?

fsimonis:
Hi, please first read our updated contributing.md.

Please add the mapping updates to this PR and see what the net benefit is for the mappings.

You can use an ASTE https://precice.org/tooling-aste.html example to profile the mapping performance.

mohitt31:
I pushed the mapping updates. `getClosestVertex` and `getClosestVertices` now take `RawCoords` (the Eigen overloads just forward to them).

The main fix was in `NearestNeighborBaseMapping::computeMapping`. It was doing 2 heap allocations per vertex because `getCoords()` created a new Eigen vector every time. They're references now. I computed the distance manually so it matches `.norm()` exactly. All tests pass.

Performance numbers (4 runs, your settings):

* `computeMapping` (16k): -9.6%
* `computeMapping` (65k): -7.9%

For a control, the untouched `Index vertices` benchmark shifted +1.6% (likely just code layout). That puts the precision at roughly ±2%, so the gains are real.

I couldn't get ASTE to install (VTK kept pulling in Qt). Instead, I added `benchmarks/mapping-nn.cpp`. This plugs straight into the CI (#2465) so you can test it directly on your hardware. Does this work for you, or do you still want ASTE numbers?

One last thing: the projection path still uses Eigen because `Polation` requires it. Do we want to change `Polation` too?

Issue #1540, opened 2023-01-16
Check API symbols


fsimonis:
**Please describe the problem you are trying to solve.**

Adding or changing the API normally needs to happen in multiple places for C++, C and Fortran.
So, there is a lot that can go wrong or that can be forgotten.

**Describe the solution you propose.**

Create a CSV file with symbol names for each binding.

```
precice::SolverInterface::isCouplingOngoing();precicec_isCouplingOngoing;precicef_is_coupling_ongoing_
```

Then use a test to check if they are present.

CMake already offers functionality for this:
* [C and Fortran](https://cmake.org/cmake/help/latest/module/CheckSymbolExists.html)
* [C++](https://cmake.org/cmake/help/latest/module/CheckCXXSymbolExists.html#module:CheckCXXSymbolExists)

So we could implement the test as a CMake script that checks these symbols.

**Describe alternatives you've considered**

Rely on future systemtests

**Additional context**

fsimonis:
The script added in #2245 now creates a report for these symbols.
There are still mismatches though. So this cannot really be added into the CI atm.

mohitt31:
I had a look at why the report still has mismatches. On current `develop` there are 13, and only one of them is something you would actually have to decide about. They fall into four groups.

**1. Constructors and the destructor (6 entries)**

- C++ side: `Participant` and `~Participant` have no C or Fortran counterpart by name.
- C/Fortran side: `precicec_createParticipant`, `precicec_createParticipant_withCommunicator`, `precicef_create_`, `precicef_create_with_communicator_`.
- These are matched pairs. The name mapping just cannot see that a constructor is what `create*` corresponds to.

**2. Free functions (2 entries)**

- `precicec_getVersionInformation` and `precicef_get_version_information_`.
- `load_precice_symbols` only keeps symbols starting with `precice::Participant::`, and `precice::getVersionInformation` is a free function, so the C++ side is never collected.

**3. `to_cpp` limitations (2 entries)**

- `precicec_createParticipant_withCommunicator` maps to `createParticipantWithcommunicator`. Splitting on `_` capitalises only the first letter of a part that is already camel-cased.
- `precicef_get_mesh_vertex_ids_and_coordinates_` maps to `getMeshVertexIdsAndCoordinates`, but the C++ name is `getMeshVertexIDsAndCoordinates`. The C binding matches fine, so this is only a snake-case-to-camel-case problem with acronyms.

**4. Deprecated Fortran aliases (6 entries)**

[...]

PR #2487, opened 2026-03-05
Fix inconsistent naming in mesh Fortran API

ankitv06:
## Main changes of this PR
This PR renames the Fortran functions for consistency with C/C++ naming scheme. The following functions have been changed to follow the C/C++ naming scheme to include 'mesh' in the name. 

precicef_set_vertex_, precicef_set_edge_, precicef_set_triangle_, precicef_set_quad_, and precicef_set_tetrahedron 

Old names are kept as deprecated wrappers for backward compatibility.
## Motivation and additional information

Closes #2248

## Author's checklist

* [x] I used the [`pre-commit` hook](https://precice.org/dev-docs-dev-tooling.html#setting-up-pre-commit) to prevent dirty commits and used `pre-commit run --all` to format old commits.
* [x] I added a changelog file with `make changelog` if there are user-observable changes since the last release.
* [ ] I added a test to cover the proposed changes in our test suite.
* [ ] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html).
* [x] I stuck to C++17 features.
* [x] I stuck to CMake version 3.22.1.
* [ ] I squashed / am about to squash all commits that should be seen as one.
* [ ] I ran the systemtests by adding the label `trigger-system-tests` (may be skipped if minor change)

## Reviewers' checklist

* [x] Does the changelog entry make sense? Is it formatted correctly?
* [x] Do you understand the code changes?

PR #2595, opened 2026-08-23
Rename precicef_set_tetrahedron to precicef_set_tetrahedron_

mohitt31:
## Main changes of this PR

Adds the Fortran function `precicef_set_tetrahedron_`.

The problem:

- Fortran compilers append a trailing underscore to symbol names.
- Five of the six deprecated `precicef_set_*` aliases have that underscore. `precicef_set_tetrahedron` does not.

| symbol | trailing `_` |
| --- | --- |
| `precicef_set_vertex_` | yes |
| `precicef_set_vertices_` | yes |
| `precicef_set_edge_` | yes |
| `precicef_set_triangle_` | yes |
| `precicef_set_quad_` | yes |
| `precicef_set_tetrahedron` | **no** |

- So `call precicef_set_tetrahedron(...)` looks for `precicef_set_tetrahedron_`, which is not in the library, and the link fails.
- The other five deprecated aliases work fine.

- I found this while looking at #1540, running the `tools/building/checkSymbols` logic against a build. `precicef_set_tetrahedron` showed up expecting a C++ counterpart called `setTetrahedron` instead of `setMeshTetrahedron`, which is what made me look at the name.
- The name has been wrong since the binding was added in #1382, and CHANGELOG.md records it under that name, so no Fortran code can be relying on it today.
- I did not simply rename it. `preCICE_SOVERSION` is the major version, so `libprecice.so.3` promises a stable ABI, and renaming would drop an exported symbol from it.
- So this PR keeps the old symbol and adds the correctly named one next to it, with a note that the old one can be removed in the next major release.
- If you would rather rename it outright, or want a different deprecation note, tell me and I will change it.

On tests:

- The deprecated Fortran aliases have no test coverage today. The only Fortran tests are `precice.solverdummy.build.fortran` and `precice.solverdummy.run.*`, and they use the current API.
- Catching this kind of bug is really a symbol-level check rather than a functional test, which is what #1540 is about. So I did not add Fortran test infrastructure inside a bugfix PR.
- Happy to add either if you want it.

Checked locally on macOS 26.5, arm64, Clang 21, Boost 1.92, Eigen 5.0.1:

- both `precicef_set_tetrahedron` and `precicef_set_tetrahedron_` are exported
- full test suite passes, 1273/1273, including the six `solverdummy` Fortran tests

## Author's checklist

* [ ] I added a test to cover the proposed changes in our test suite. (see above)
* [ ] For breaking changes: I documented the changes in the appropriate [porting guide](https://precice.org/couple-your-code-porting-overview.html). (not breaking, this only adds a symbol)

### DIFF
--- a/extras/bindings/fortran/include/precice/preciceFortran.hpp
+++ b/extras/bindings/fortran/include/precice/preciceFortran.hpp
@@ -505,7 +505,7 @@ PRECICE_API void precicef_set_mesh_tetrahedron_(
  * @brief @deprecated Use precicef_set_mesh_tetrahedron_() instead.
  * @see precicef_set_mesh_tetrahedron_
  */
-PRECICE_API void precicef_set_tetrahedron(
+PRECICE_API void precicef_set_tetrahedron_(
     const char *meshName,
     const int  *firstVertexID,
     const int  *secondVertexID,
--- a/docs/changelog/2595.md
+++ b/docs/changelog/2595.md
@@ -1,0 +1,1 @@
+* Fixed the deprecated Fortran function `precicef_set_tetrahedron`, which was missing the trailing underscore that Fortran compilers append and could therefore not be called from Fortran. It is now `precicef_set_tetrahedron_`.
--- a/extras/bindings/fortran/src/preciceFortran.cpp
+++ b/extras/bindings/fortran/src/preciceFortran.cpp
@@ -616,7 +616,7 @@ void precicef_set_quad_(
   precicef_set_mesh_quad_(meshName, firstVertexID, secondVertexID, thirdVertexID, fourthVertexID, meshNameLength);
 }
 
-void precicef_set_tetrahedron(
+void precicef_set_tetrahedron_(
     const char *meshName,
     const int  *firstVertexID,
     const int  *secondVertexID,


extras/bindings/fortran/include/precice/preciceFortran.hpp
```diff
@@ -505,6 +505,23 @@ PRECICE_API void precicef_set_mesh_tetrahedron_(
...
+    int         meshNameLength);
+
+/**
+ * @brief @deprecated Use precicef_set_mesh_tetrahedron_() instead.
+ *
+ * This name lacks the trailing underscore that Fortran compilers append to
+ * symbol names, so it cannot be called from Fortran. It only remains to keep
+ * the ABI of preCICE 3 intact and can be removed in the next major release.
+ *
+ * @see precicef_set_mesh_tetrahedron_
```
MakisH: No need for such motivation comments when the reason is immediately clear. Let's follow the same format as already in the file:


```suggestion
 * @brief @deprecated Use precicef_set_mesh_tetrahedron_() instead.
 * @see precicef_set_mesh_tetrahedron_
```

fsimonis:
> I did not simply rename it. preCICE_SOVERSION is the major version, so libprecice.so.3 promises a stable ABI, and renaming would drop an exported symbol from it.
So this PR keeps the old symbol and adds the correctly named one next to it, with a note that the old one can be removed in the next major release.

@MakisH To my understanding, this symbol is not callable in Fortran. So no one can use it right now, and we can just rename it. Correct?

MakisH:
> > I did not simply rename it. preCICE_SOVERSION is the major version, so libprecice.so.3 promises a stable ABI, and renaming would drop an exported symbol from it.
> > So this PR keeps the old symbol and adds the correctly named one next to it, with a note that the old one can be removed in the next major release.
> 
> @MakisH To my understanding, this symbol is not callable in Fortran. So no one can use it right now, and we can just rename it. Correct?

Not sure what happened to my review summary: while I would also simply rename it, I am fine with either, since we already do the same for similar cases. Especially in this case, I agree that directly renaming makes more sense.

extras/bindings/fortran/include/precice/preciceFortran.hpp
mohitt31: Done. Following your and @fsimonis's preference I dropped the old symbol instead of keeping both, so that block is gone entirely and the remaining declaration keeps the existing two-line format.

MakisH (approved):

MakisH (commented):
I just noticed that this is already a deprecated function (there is the `precicef_set_mesh_tetrahedron_`), and the renaming was already done in https://github.com/precice/precice/pull/2487, which has been released.

@fsimonis merge or not merge? I would suggest to leave the state as-is (not merge).

In any case, the effort is very welcome, and first PRs often overlook such side aspects.

mohitt31:
You are right, and thanks for catching it. I checked #2487 and `precicef_set_mesh_tetrahedron_` is there and works, so anyone setting tetrahedra from Fortran already has a working function. The broken alias only mattered to code migrating off the old name, and since the name never worked from Fortran there is nobody in that position. So there is no user-visible bug to fix here, and I agree with leaving it as-is.

The only thing that still reads a bit odd is that `preciceFortran.hpp` keeps declaring a function that Fortran cannot call, so `checkSymbols` will keep reporting it. If you ever want that tidied, deleting the declaration and definition is a one-liner, but that is cosmetic and your call.

Happy to close this myself, just say the word. I will take the same look at the remaining `checkSymbols` mismatches and put what I find on #1540 instead, since that is where it is actually useful.

fsimonis:
@MakisH thanks for checking this!
This symbol is actually kept for backward compatibility. I'll close this PR

PR #2597, opened 2026-09-01
Fix Boost linking of unit_test_framework and disallow static

### LINKED ISSUE #2596: Boost linking issues
**Describe your setup**

Operating system (e.g. Linux distribution and version): Alma Linux 8, Windows 11
preCICE Version: develop 01b0aa5469ca0a8d5534880d6b364cd135ac3ec8

**Describe the problem**
We are moving the third-party dependencies of Delft3D to Conan. We also moved Boost into Conan, but then ran into a linker issue on Linux. There is a separate issue in the Boost Conan recipe when using Boost as a shared library, overlinking some dependencies of Boost Test and causing multiple copies of some global string symbols in the linked libraries. This issue was presented to us because the preciceCore CMake target links to Boost::unit_test_framework even if BUILD_TESTING is OFF.

We tried a workaround, where we link Boost statically: this is likely the most common method of consuming Boost. However, preCICE's CMakeLists.txt unconditionally defines the macro BOOST_ALL_DYN_LINK causing issues on Windows. There is the Boost_USE_STATIC_LIBS option, but it is ignored in this case.

**Additional context**
I will create a PR with proposed fixes for these two issues.

harmenwierenga:
…nking unit_test_framework to testprecice

## Main changes of this PR

1. Raise error when Boost_USE_STATIC_LIBS is ON.
2. Link Boost::unit_test_framework into testprecice target, not into preciceCore.

closes #2596

## Author's checklist

## Reviewers' checklist

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -220,9 +220,12 @@ if(POLICY CMP0167)
 endif()
 
 option(Boost_USE_STATIC_LIBS "Force BoostConfig.cmake to find static libraries." OFF)
-find_package(Boost 1.83.0 REQUIRED CONFIG
-  COMPONENTS log log_setup program_options thread unit_test_framework
-  )
+set(_precice_boost_components log log_setup program_options thread)
+if(BUILD_TESTING)
+  list(APPEND _precice_boost_components unit_test_framework)
+endif()
+find_package(Boost 1.83.0 REQUIRED CONFIG COMPONENTS ${_precice_boost_components})
+unset(_precice_boost_components)
 mark_as_advanced(Boost_INCLUDE_DIR Boost_LOG_LIBRARY_RELEASE Boost_LOG_SETUP_LIBRARY_RELEASE Boost_PROGRAM_OPTIONS_LIBRARY_RELEASE Boost_SYSTEM_LIBRARY_RELEASE Boost_THREAD_LIBRARY_RELEASE Boost_UNIT_TEST_FRAMEWORK_LIBRARY_RELEASE)
 mark_as_advanced(Boost_INCLUDE_DIR Boost_LOG_LIBRARY_DEBUG Boost_LOG_SETUP_LIBRARY_DEBUG Boost_PROGRAM_OPTIONS_LIBRARY_DEBUG Boost_SYSTEM_LIBRARY_DEBUG Boost_THREAD_LIBRARY_DEBUG Boost_UNIT_TEST_FRAMEWORK_LIBRARY_DEBUG)
 message(STATUS "Found Boost ${Boost_VERSION}")
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -427,14 +427,16 @@ endif()
 
 
 # Setup Boost
-target_compile_definitions(preciceCore PUBLIC BOOST_ALL_DYN_LINK BOOST_ASIO_ENABLE_OLD_SERVICES BOOST_GEOMETRY_DISABLE_DEPRECATED_03_WARNING)
+target_compile_definitions(preciceCore PUBLIC BOOST_ASIO_ENABLE_OLD_SERVICES BOOST_GEOMETRY_DISABLE_DEPRECATED_03_WARNING)
+if(NOT Boost_USE_STATIC_LIBS)
+  target_compile_definitions(preciceCore PUBLIC BOOST_ALL_DYN_LINK)
+endif()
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
fsimonis: To be absolutely correct, we should not request the unit tests if tests are disabled.

harmenwierenga: I have added some code to remove the unit_test_framework component when BUILD_TESTING is OFF

fsimonis:
This doesn't work on an up2date Arch system. I added a commit to use static boost in the CI.

When I configure with `cmake --preset=development -DBoost_USE_STATIC_LIBS=ON`, then I get build errors.

```
/src/testing/main.cpp:77:20: error: use of undeclared identifier 'init_unit_test'
   77 |   framework::init(&init_unit_test, 1, argv);
      |                    ^~~~~~~~~~~~~~
```

When I configure with `cmake --preset=development -DBUILD_TESTING=OFF -DBoost_USE_STATIC_LIBS=ON`, then the build succeeds, but the executables SEGAULT.

```
$ gdb -ex run ./precice-version
...
#0  0x00007ffff7d94d9b in boost::log::v2s_mt_posix::attributes::named_scope::named_scope() () from build/libprecice.so.3
#1  0x00007ffff7932ddd in precice::logging::Logger::LoggerImpl::LoggerImpl (this=0x55555555d420, module="mesh::BoundingBox")
    at src/logging/Logger.cpp:125
#2  0x00007ffff7933092 in precice::logging::Logger::Logger (this=0x7ffff7e4d7d8 <precice::mesh::BoundingBox::_log>, module="mesh::BoundingBox")
    at src/logging/Logger.cpp:129
#3  0x00007ffff727624d in __cxx_global_var_init () at src/mesh/BoundingBox.cpp:14
#4  0x00007ffff7276299 in _GLOBAL__sub_I_unity_mesh_cxx.cxx () from build/libprecice.so.3
```

harmenwierenga:
> This doesn't work on an up2date Arch system. I added a commit to use static boost in the CI.
> 
> When I configure with `cmake --preset=development -DBoost_USE_STATIC_LIBS=ON`, then I get build errors.
> 
> ```
> /src/testing/main.cpp:77:20: error: use of undeclared identifier 'init_unit_test'
>    77 |   framework::init(&init_unit_test, 1, argv);
>       |                    ^~~~~~~~~~~~~~
> ```
> 
> When I configure with `cmake --preset=development -DBUILD_TESTING=OFF -DBoost_USE_STATIC_LIBS=ON`, then the build succeeds, but the executables SEGAULT.
> 
> ```
> $ gdb -ex run ./precice-version
> ...
> #0  0x00007ffff7d94d9b in boost::log::v2s_mt_posix::attributes::named_scope::named_scope() () from build/libprecice.so.3
> #1  0x00007ffff7932ddd in precice::logging::Logger::LoggerImpl::LoggerImpl (this=0x55555555d420, module="mesh::BoundingBox")
>     at src/logging/Logger.cpp:125
> #2  0x00007ffff7933092 in precice::logging::Logger::Logger (this=0x7ffff7e4d7d8 <precice::mesh::BoundingBox::_log>, module="mesh::BoundingBox")
>     at src/logging/Logger.cpp:129
> #3  0x00007ffff727624d in __cxx_global_var_init () at src/mesh/BoundingBox.cpp:14
> #4  0x00007ffff7276299 in _GLOBAL__sub_I_unity_mesh_cxx.cxx () from build/libprecice.so.3
> ```

Interesting. I now see the following in [the documentation for boost log](https://www.boost.org/doc/libs/latest/libs/log/doc/html/log/installation/config.html):
"If your application consists of more than one module (e.g. an exe and one or several dll's) that use Boost.Log, the library must be built as a shared object."

Since preCICE is shipped as a shared library, I think that this means that only shared Boost is supported. Would you agree if I removed the static boost option altogether?

fsimonis:
Thanks for looking into this!

> Since preCICE is shipped as a shared library, I think that this means that only shared Boost is supported.

That's a fair assessment. The reasons for switching to another logging backend are really piling up.

> Would you agree if I removed the static boost option altogether?

You mean the support you added? Sure.

From this PR, we can keep the conditional discovery and linking of the unit test framework.

harmenwierenga:
Thanks @fsimonis, I have kept the unit test framework changes. I have also added an error if boost is linked statically, please check if you agree with that.

I also removed your test commit

PR #2570, opened 2026-05-27
Release PR template: Remove tests covered by the system tests

MakisH:
## Main changes of this PR

Removes from the release pull request template the cases now covered by the system tests. See recent runs in https://github.com/precice/tutorials/actions/workflows/system-tests-latest-components.yml and https://github.com/precice/tutorials/issues/448

It also removes the code_aster tutorial, since the respective adapter is currently unmaintained.

## Motivation and additional information

See the complete list in the `release_test` in [`tests.yaml`](https://github.com/precice/tutorials/blob/develop/tools/tests/tests.yaml) and https://github.com/precice/tutorials/issues/448

In particular, the OpenFOAM parallel cases that this PR removes are also now executed in parallel by the system tests.

The `perpendicular-flap` with `fluid-nutils` - `solid-calculix` takes a few minutes to even complete the first time window (without converging). I added that to an additional `extra` test suite, which can be triggered manually.

### Solver dummies

To test the bindings dummies, we would need to first have a mechanism to integrate cases from other repositories than the tutorials (https://github.com/precice/tutorials/issues/538). However, we already test the elastic-tube-1d with Python (and C++), and we could add more of these.

- elastic-tube-1d C++ > C++ solverdummy
- elastic-tube-1d Fortran > Fortran solverdummy
- elastic-tube-1d Fortran module > Fortran module solverdummy
- elastic-tube-1d Python > Python solverdummy
- elastic-tube-1d Julia > Julia bindings solverdummy

That's why I remove these solverdummies, which should be covered by the CI of each repository. The above list excludes the:

- MATLAB bindings solverdummy

To test Alya and MATLAB, similarly + licenses.

Testing SuperMUC as part of the system tests will not be in scope any time soon.

PR #2599, opened 2026-09-12
Always-on system tests

MakisH:
## Main changes of this PR

We currently have a workflow that runs the basic `precice` test suite on demand (using the label).

This PR:

1. Changes the label-based workflow back to triggering the `release` test suite (useful for releases, on-demand).
2. Adds a system-tests job in the `build-and-test.yml` workflow, with dependency on the `pre-commit` job.

It also removes the instruction regarding the `trigger-system-test` label from the pull request template, as that is now only needed for releases (addressed in https://github.com/precice/precice/pull/2570).

Screenshot from 2026-09-12 17-26-04

The `system-tests` job with the given test suite takes 13min to run (faster than some current jobs), with ~5min being spent in the first test.

It could also be configured to depend on the `linux` job to save the runner in case the code does not build, but that would extend the total time of the CI, without much benefit.

We often forget to trigger the on-demand workflow. This should make sure that a basic set of tests always runs.

The test suite is defined in the [tutorials repository](https://github.com/precice/tutorials/blob/develop/tests/tests.yaml) and is currently:

https://github.com/precice/tutorials/blob/e79a58a6df3e4c781337f5660b462ae29a6c60d7/tests/tests.yaml#L1074-L1084

These include just a few components that typically don't introduce dependency-related regressions, but cover a wide set of features and work complementary to the unit and integration tests.

More tests anyway run nightly (`release`) and weekly (`extra`), but this should make it easier to locate regressions stemming from changes in preCICE.

**Side-effects:**

1. Whenever we introduce breaking changes, this step will need to be adjusted to use the tutorials (or other components) from another branch. This can be done by adding `TUTORIALS_REF` to the `build_args:` (by default, the `develop` branch of all components is being picked).
2. The tests might fail for unrelated reasons (e.g., Sourceforge being down, which serves the OpenFOAM packages). This has not happened in the past several months, and is the same issue we already have in the nightly tests.

## Author's checklist


fsimonis (approved):
Very nice that it now takes the same amount of time than our larger suites

